PR #102, opened 2016-07-16
Doc fixes

somus:
- Replace block static methods with inline static methods in inline doc file
- Fix hash link refs for static methods in various model doc files.

### DIFF
--- a/docs/reference/models/block.md
+++ b/docs/reference/models/block.md
@@ -20,8 +20,8 @@ Block nodes may contain nested block nodes, inline nodes, and text nodes—just 
   - [`length`](#length)
   - [`text`](#text)
 - [Static Methods](#static-methods)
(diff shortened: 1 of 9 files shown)


somus:
I guess there is some problem with tests. I was unable to run tests on my machine too. But, I got a different error. It says `EDGE_METHODS.includes is not a function`.

ianstormtaylor:
Thank you @somus you are awesome!

No worries about the tests, I hadn't quite gotten them passing properly on Travis CI on Friday before I went away for the weekend. I think the `.includes` error might be because I was developing on Node 6, and hadn't yet made sure it also works on 4. I'll switch that out for a compatible method, opened an issue for it here: https://github.com/ianstormtaylor/slate/issues/110

PR #105, opened 2016-07-16
Fix "initialState" declaration in "Installing Slate" doc

ctrlplusb:
I tried to use the example however I received errors saying that `map` could not be executed over `undefined`.

It looks like the deserialiser was trying to map over the `nodes` prop of an object, but the example passes in an array of objects.  I removed the wrapping array and all was well again.


ianstormtaylor:
Ah shoot, it should actually contain an extra wrapping `{ nodes: [...] }`. Thanks for finding this. I'll merge and add the wrapper!

PR #117, opened 2016-07-18
Join texts with a space

MaxGfeller:
The computed property `text` is incredibly useful. But when it's used in a document to join all texts, those should be separated by a space instead of just concatenated.

### DIFF
--- a/lib/models/document.js
+++ b/lib/models/document.js
@@ -70,7 +70,7 @@ class Document extends new Record(DEFAULTS) {
   get text() {
     return this.nodes
       .map(node => node.text)
-      .join('')
+      .join(' ')
   }
 
 }


ianstormtaylor:
Hey @MaxGfeller thanks for opening this! I'm not sure we want to do this, since the `text` property is "technically" correct, as far as calculating offsets and such, in not adding the spaces between blocks. Could I ask what your use case was?

You might want to check out the [`Plain`](https://github.com/ianstormtaylor/slate/blob/master/docs/reference/serializers/plain.md) serializer if you're looking to get a text string out of the document.

Let me know!

MaxGfeller:
@ianstormtaylor `Plain.serialize()` was exactly what i was looking for, thanks a lot!

ianstormtaylor:
Nice, glad to hear it!

PR #121, opened 2016-07-18
Fixing broken links in reference of docs

jessep:
Looks like there was a refactor at some point where this went from a flat directory to split into categories, and some of the links never got changed. Just fixing those up.


ianstormtaylor:
Thank you @jessep! Awesome :)

Issue #83, opened 2016-07-14
Type definitions


mitermayer:
Great job, and great effort! looking forward to see how the project will shape it self!

Would be great if this could export some type definitions for either typescript or flow.

ianstormtaylor:
Thanks :) glad you're interested! I'd definitely be down for this. I haven't used either myself, and would want to wait a little bit more until API's stabilize, but at that point I'd be +1 someone contributing them!

ianstormtaylor:
Closing in favor of @joehua87's mention of `flow-typed` in #148.

arekkas:
@ianstormtaylor there are still no type definitions, right? Maybe re-open this issue?

frederickfogerty:
Please re-open this issue - there are no type definitions on `flow-typed`.

PR #148, opened 2016-07-21
Try add Flow Annotation

joehua87:
I've try to add Flow Annotation as proposed in https://github.com/ianstormtaylor/slate/issues/83
I just worked with all the files in models & components/content & waiting for **feedbacks** before make it further
Please let me know if you need anything that make it works

Command:
```
npm run flow
```


ianstormtaylor:
Hey @joehua87, this is awesome thank you for taking the time to do this!

I have one worry, which is that the methods right now are still not super-nailed down, since I'm still working on a few of the "core" requirements, so things may change a decent amount. For that reason, I'm thinking that we should keep this PR here but not merge it until things settle down a bit more.

I haven't used Flow before myself, so I'm not really sure what the workflow looks like, or what I'd be getting myself into by committing to maintain the flow annotations. What exactly would the commitment be?

ctrlplusb:
This would be great. I'm a flow user, and have been creating definitions too. Happy to help out in any way when things have settled.

joehua87:
@ianstormtaylor great to hear that. I will maintain this to keep it sync with the main repo. :)

joehua87:
I think that we should make a separate type annotation in https://github.com/flowtype/flow-typed in order to not pollute the code

frederickfogerty:
Hey @ctrlplusb, would you still have your definitions lying around by any chance? I'm trying to create some Typescript definitions, and having a flow version would make it a lot easier than creating the Typescript ones by scratch

PR #157, opened 2016-07-21
modifyNode transform

### LINKED ISSUE #123: add modifyNode Transform

mathieudutour:
fix #123

### DIFF
--- a/lib/models/transform.js
+++ b/lib/models/transform.js
@@ -50,7 +50,7 @@ const DOCUMENT_RANGE_TRANSFORMS = [
  */
 
 const DOCUMENT_NODE_TRANSFORMS = [
-  'modifyNode',
+  'setNodeByKey',
 ]
 
 /**
--- a/lib/models/transform.js
+++ b/lib/models/transform.js
@@ -42,7 +42,8 @@ const DOCUMENT_TRANSFORMS = [
   'unwrapBlockAtRange',
   'unwrapInlineAtRange',
   'wrapBlockAtRange',
-  'wrapInlineAtRange'
+  'wrapInlineAtRange',
+  'modifyNode',
 ]
 
 /**
@@ -211,9 +212,13 @@ class Transform extends new Record(DEFAULT_PROPERTIES) {
 
     if (includes(DOCUMENT_TRANSFORMS, type)) {
       let { document, selection } = state
-      let [ range, ...rest ] = args
-      range = range.normalize(document)
-      document = document[type](range, ...rest)
+      try {
+        let [ range, ...rest ] = args
+        range = range.normalize(document)
+        document = document[type](range, ...rest)
+      } catch(e) {
+        document = document[type](...args)
+      }
       selection = selection.normalize(document)
       state = state.merge({ document, selection })
       return state
--- a/lib/models/transform.js
+++ b/lib/models/transform.js
@@ -23,10 +23,10 @@ const Step = new Record({
 })
 
 /**
- * Document transforms.
+ * Document range transforms.
  */
 
-const DOCUMENT_TRANSFORMS = [
+const DOCUMENT_RANGE_TRANSFORMS = [
   'deleteAtRange',
   'deleteBackwardAtRange',
   'deleteForwardAtRange',
@@ -43,6 +43,13 @@ const DOCUMENT_TRANSFORMS = [
   'unwrapInlineAtRange',
   'wrapBlockAtRange',
   'wrapInlineAtRange',
+]
+
+/**
+ * Document node transforms.
+ */
+
+const DOCUMENT_NODE_TRANSFORMS = [
   'modifyNode',
 ]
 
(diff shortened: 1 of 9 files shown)


ianstormtaylor:
Hey @mathieudutour this is awesome! Thank you so much. 

I'm going to read through and add a few comments. Before that, one request, could we actually change it to `setNode`? That way it's parallel with `setInline` and `setBlock`. I just realized if we do that we can actually change the `setInline/Block`methods to use `setNode` under the covers too, moving us one step closer to something like collaboration (and it'll be cleaner too).

lib/models/transform.js
```diff
@@ -211,9 +212,13 @@ class Transform extends new Record(DEFAULT_PROPERTIES) {
 
     if (includes(DOCUMENT_TRANSFORMS, type)) {
       let { document, selection } = state
-      let [ range, ...rest ] = args
-      range = range.normalize(document)
-      document = document[type](range, ...rest)
+      try {
```
ianstormtaylor: Let's create another array constant here so that we end up with `DOCUMENT_RANGE_TRANSFORMS` and `DOCUMENT_NODE_TRANSFORMS` so that we don't have to `try/catch`.

lib/models/transforms.js
```diff
@@ -899,6 +899,30 @@ const Transforms = {
...
+   *
+   * @param {String} key
+   * @param {Object or String} properties
+   * @return {Node} node
+   */
+
+  modifyNode(key, properties) {
+    const node = this
+
+    node.assertDescendant(key)
```
ianstormtaylor: `assertDescendant` will actually return the match, since it has to loop the tree anyways, so we can have one less lookup here.

ianstormtaylor:
Looks good except for those few things!

mathieudutour:
Regarding the renaming to `setNode`, wouldn't it be weird since the `setBock` and `setNode` won't have the same signature? 
One is `key, props`, the other is `props`.

Maybe we can change the signature of `setBlock` to accept both. If there is a key, then it's using `setNode`, otherwise it's using `setBlocksAtRange`.
What do you think?

mathieudutour:
Or maybe change `setBlock` to `key, props` and create `setBlockInSelection`?

ianstormtaylor:
Good question!

So i debated this for while, whether to prefix all of the "operate on the current selection" methods with something—I had been thinking `current`. So it'd be `setCurrentBlock()`, `setCurrentLine()`, `addCurrentMark()`, but it felt a bit verbose and didn't sound right grammar-wise everywhere. If we added something like `inCurrentSelection` so that they'd be `setBlockInCurrentSelection` they'd always sound right grammatically, but then it's really verbose.

I just noticed actually that `setBlock/setInline` should actually be pluralized everywhere to match the other properties, especially since they operate on all the nodes that match, so they'd instead become `setBlocks` and `setInlines`.

So we'd end up with:

- `setNode`
- `setBlocks` (which is really just convenience for `document.setNode` on all `state.blocks`)
- `setInlines` (which is really just convenience for `document.setNode` on all `state.inlines`)

ianstormtaylor:
Another thought while working on inserting node transforms. If we added a `deleteNode`, it might be more clear to call it `removeNodeByKey`. And then `setNodeByKey` could work as well. Such that we might end up with a group of them:

- `setNodeByKey(key, properties)`
- `deleteNodeByKey(key)`
- `insertNodeAfterByKey(key, properties || node)`
- `insertNodeBeforeByKey(key, properties || node)`

Which makes the "current-acting" transforms more distinguished:

- `setBlocks(properties)`
- `setInlines(properties)`
- `insertFragment(fragment)`
- `insertNode(properties || node)`

I could also be convinced to change `set*` to `update*` and `delete*` to `remove*`.

mathieudutour:
The first time I used slate, I had a hard time understanding that `setBlocks` was working on the current selection, it's not clear from its name. I like `setCurrentBlocks`, it's not too long and it's clearer

ianstormtaylor:
Yeah you're right. Not quite clear enough, I've just gotten used to it and forgotten what it's like to learn. What do you think about `setSelectedBlocks`? I think it might be even clearer than "current" I just realized. 

It gets slightly awkward with `addSelectedMarks` but maybe not enough to matter. Although I'm not sure what to do with delete's, since `deleteSelected` is slightly awkward, but `deleteSelectedForward` is very awkward. I was hoping to have a single naming scheme that applies to all the stateful ones.

mathieudutour:
Yeah I like `setSelected`. `deleteSelected` is fine, but `addSelectedMarks` is really weird, same as `deleteSelectedForward` (but `deleteForward` is not great anyway, maybe `deleteNextCharacters`?). 

I think it's better to have longer names. It's a bit annoying when developing but really helps when reviewing and maintaining.

So I would lean more on the side of `setBlocksInSelection` or someting

ianstormtaylor:
Nice, this looks good! I'm gonna think/use some more before renaming the current selection ones, but thanks for the ideas!

ianstormtaylor:
Added in `0.6.7`!

PR #158, opened 2016-07-22
fix performance issue in Leaf

mathieudutour:


ianstormtaylor:
Ah good catch, thanks!

PR #165, opened 2016-07-23
typeof key is string (not Key)

joehua87:


ianstormtaylor:
@joehua87 thank you, you're awesome! I love these types of fixes 😄

PR #170, opened 2016-07-24
Fix error in Hello World of installation guide

adamlong5:
Looks like the `deserialize()` function expects to be passed an object with a `nodes` key.


ianstormtaylor:
Ah, thanks @adamlong5!

Issue #174, opened 2016-07-25
make `onSelect` handled by the "core" plugin


ianstormtaylor:
Relating to https://github.com/ianstormtaylor/slate/issues/173

ianstormtaylor:
Done by @tyler-johnson in `0.7.2`.

Issue #173, opened 2016-07-25
Question - accessing the native browser event that triggered a state change


tyler-johnson:
Hello, I am attempting to make a toolbar that shows itself whenever a non-collapsed selection is made, similar to your `hovering-menu`. To do this, I have a plugin that defines an `onChange` method that updates the toolbars position as the selection changes. I would like the position of the toolbar to be based on the mouse position, specifically where the `mouseup` event happens after a selection.

I notice that when the `onSelect` event fires in the [`Content` component](https://github.com/ianstormtaylor/slate/blob/c7fac3456eb0fccc5e7aa039a2b4bea0cf11174b/lib/components/content.js#L554), if the event was triggered by the mouse, `e.nativeEvent` is a MouseEvent and contains the details I need (positioning). How can I access this event from within the plugins `onChange`?

Ideally, any state change would have the associated event that triggered it (if it was triggered by an event). Is it possible to attach this information?

Edit: I should mention that I have attempted to access the native event directly through `window.event`, but react seems to dispatch its own custom event that doesn't contain the original native event.

ianstormtaylor:
Hmm interesting! Good question. I haven't had this use case yet, but seems like it should be accounted for. From what I can tell, it seems like it would be solved if we made `onSelect` actually handled by the plugin stack, with the "core" plugin handling the default behavior last. So you could observe the events from the plugin, or if you really wanted to, change the selection behavior.

Just opened an issue for this: https://github.com/ianstormtaylor/slate/issues/174

tyler-johnson:
Yes, perfect! Exposing `onSelect` as a plugin method would solve my issue.

Is this something that I could perhaps open a pull request for? This seems like it would be fairly straight forward to implement.

PR #175, opened 2016-07-25
moved onSelect event handling to core plugin

tyler-johnson:
My attempt at #174.

I added an `onSelect` method to the core plugin with the Content component `onSelect` logic and then added an `onSelect` prop to the Editor component.

Tests and linting pass on my machine. I also did a local install into my project and it solves my problem in #173.


ianstormtaylor:
@tyler-johnson you are seriously amazing! Thank you!

ianstormtaylor:
Added in `0.7.2`!

PR #189, opened 2016-07-28
undo and redo transform does not return Transform

SamyPesse:
This PR fixes the documentation for `undo` and `redo` transforms, which return a new `State` instead of a `Transform`.


ianstormtaylor:
Ah nice, thanks dude!

PR #193, opened 2016-07-28
Add properties hasUndo and hasRedo to State

SamyPesse:
This PR adds two computed properties  to `State`: `hasUndo` and `hasRedo`.

These properties are shortcuts when for example rendering Undo/Redo buttons in a toolbar.

I’ve documented the properties, but I’m not an english native so let me know if the descriptions are not good enough.


SamyPesse:
Concerning tests, I don't know where to add unit tests for these properties.

ianstormtaylor:
Nice, I dig this. I'm going to change them to be plural, so `hasUndos` and `hasRedos`.  Thanks!

ianstormtaylor:
Published in `0.8.1`.

PR #195, opened 2016-07-28
fix isBackward in onSelect event selection

tyler-johnson:
I think at some point in the `onSelect` changes the `isBackward` value broke. I mostly just copied how `moveTo` managed to update it, which was the method used before `onSelect` was moved.

Hopefully this is the correct way to do it?


ianstormtaylor:
Nice, perfect! Thanks!

ianstormtaylor:
Published in `0.8.1`.

PR #196, opened 2016-07-28
Fix #194: fix possilbe deletion of "data" property with setBlock transforms

### LINKED ISSUE #194: node.data can be undefined?
I'm working on building a POC of Slate in our application.

When switching a header into a code block, an exception is thrown by our code block node component (which is using `node.data.get`): `Uncaught TypeError: Cannot read property 'get' of undefined`

It looks like the data property can be undefined. Shouldn't it default to an empty `Data()`?

screen shot 2016-07-28 at 18 22 22

SamyPesse:
This PR prevents the deletion of `data` by passing `{ data: null | undefined }` to `setBlock`.

### DIFF
--- a/lib/models/transforms.js
+++ b/lib/models/transforms.js
@@ -527,11 +527,15 @@ const Transforms = {
     if (typeof properties == 'string') {
       properties = { type: properties }
     }
+    if (properties.data) {
+        properties.data = Data.create(properties.data)
+    } else {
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Good call!

ianstormtaylor:
Published in `0.8.1`.

PR #197, opened 2016-07-28
add tests for calling wrapInline twice

### LINKED ISSUE #185: calling .wrapInline() twice causes breaking returns on both sides of selection
I have a simple transform like so:

```js
state.transform()
  .wrapInline("inner")
  .wrapInline("outer")
  .apply();
```

If before the transform I have a block that looks like this:

```
<div>
  Before Text After
</div>
```

I was expecting result of the transform to be this:

```
<div>
  Before
  <span class="outer">
    <span class="inner">Text</span>
  </span>
  After
</div>
```

However what I got looks like this:

```
<div>
  Before
  <span class="inner"><br /></span>
  <span class="outer">
    <span class="inner">Text</span>
  </span>
  <span class="inner"><br /></span>
  After
</div>
```

Am I doing something wrong here?

tyler-johnson:
So unfortunately it looks like the `0.8.0` release did not fix #185. I added a small test to show this.

I hope calling the folder `twice` is clear enough. Please let know if I should change anything.

### DIFF
--- a/test/transforms/fixtures/wrap-inline/twice/index.js
+++ b/test/transforms/fixtures/wrap-inline/twice/index.js
@@ -1,0 +1,23 @@
+
+import assert from 'assert'
+
+export default function (state) {
+  const { document, selection } = state
+  const texts = document.getTexts()
+  const text = texts.first()
+  const range = selection.merge({
+    anchorKey: text.key,
+    anchorOffset: 7,
+    focusKey: text.key,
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Thanks, that looks perfect! I'll figure out what's going on.

ianstormtaylor:
Fixed in `0.8.1`, thanks for the test case!

PR #202, opened 2016-07-29
Add ability for `renderMark` to return a React componen

### LINKED ISSUE #201: Add ability for `renderMark` to return a React component
Following https://github.com/ianstormtaylor/slate/issues/188, it would be nice to be able to use full-fledged components as mark renderers. This would make re-using styles a lot easier, but also allow for things like styling marks on `:hover`, etc.

SamyPesse:
This PR fixes #201 

Basically, it allows `renderMarks` to return React components:

```js
const BoldSpan = (props => <b>{this.props.children}</b>)

renderMarks(mark) {
     if (mark.type == 'bold') return BoldSpan
}
```

Currently, if a leaf has 2 marks (`bold` and `italic`), it will result in one `span` with `style={ { fontWight: 'bold', fontStyle: 'italic' }}`; with this PR it will result in `<b><em>{text}</em></b>`.

### TODOs

- [x] Change rendering of marks to use a component
- [x] Backward compatibility if `renderMark` returns an Object, build it as `<span style={style}>{props.children}</span>
- [x] Adapt unit tests
- [x] Ensure that examples are working

--- 

@ianstormtaylor I'm posting this PR while it's still a WIP so that you can review it and tell if it looks like the right implementation for you.

### DIFF
--- a/lib/components/editor.js
+++ b/lib/components/editor.js
@@ -252,9 +252,10 @@ class Editor extends React.Component {
 
       // plugin.renderMak can return a map of style
       if (!(typeof component === 'function')) {
-          component = (props => <span style={component}>{props.children}</span>)
+          const style = component
+          component = (props => <span style={style}>{props.children}</span>)
       }
-      
+
       if (component) return component
     }
   }
--- a/lib/components/editor.js
+++ b/lib/components/editor.js
@@ -242,17 +242,15 @@ class Editor extends React.Component {
    *
    * @param {Mark} mark
    * @param {Set} marks
-   * @return {Object}
+   * @return {Element}
    */
 
   renderMark = (mark, marks) => {
     for (const plugin of this.state.plugins) {
       if (!plugin.renderMark) continue
-      const style = plugin.renderMark(mark, marks, this.state.state, this)
-      if (style) return style
+      const component = plugin.renderMark(mark, marks, this.state.state, this)
+      if (component) return component
     }
-
-    return {}
   }
 
   /**
--- a/lib/components/editor.js
+++ b/lib/components/editor.js
@@ -248,7 +248,13 @@ class Editor extends React.Component {
   renderMark = (mark, marks) => {
     for (const plugin of this.state.plugins) {
       if (!plugin.renderMark) continue
-      const component = plugin.renderMark(mark, marks, this.state.state, this)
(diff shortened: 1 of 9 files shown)


SamyPesse:
@ianstormtaylor It's working but there is one important change (concerning the unit test `rendering/custom-mark-multiple`):

Since before Slate was combining style, it was possible in `renderMark` to return a style matching multiple marks (example: [test/rendering/fixtures/custom-mark-multiple/index.js#L17](https://github.com/ianstormtaylor/slate/blob/master/test/rendering/fixtures/custom-mark-multiple/index.js#L17).
With this PR, it's not possible anymore, since the `renderMark` created by `Editor` for the leafs only returns components.

Do you think we should keep this feature? I personally believe that it's not a good idea to return a component or style matching multiple marks in one `renderMark` call.

ianstormtaylor:
Nice @SamyPesse this is great! 

I think we should preserve the feature allowing for rendering based on multiple marks, yup. It's useful for cases like where you have a `MyFontBoldItalic.ttf` and need to change `font-family` for instance. And since that one seems valid, there's probably other cases I've yet to discover too!

SamyPesse:
@ianstormtaylor In fact, my suggestion was a bad idea, we can't easily mix `styles` and "auto-combine" marks; and the rendering based on multiple marks can still work (https://github.com/SamyPesse/slate/tree/feature/renderMarkComponent/test/rendering/fixtures/custom-mark-multiple)

```js
function renderMark(mark, marks) {
  if (
    marks.size > 1 &&
    mark.type == 'bold' &&
    marks.some(m => m.type == 'italic')
  ) {
    return BOLD_ITALIC
  }

  if (mark.type == 'bold') return BOLD
  if (mark.type == 'italic') return ITALIC
}
```

SamyPesse:
It's now passing all tests! 🎉 and combining marks is possible (see change in [test/rendering/fixtures/custom-mark-multiple/index.js](https://github.com/ianstormtaylor/slate/pull/202/commits/0720c2e3be108f2113916bab9d472a197c82e171))

Let mw know if you need me to do more changes

ianstormtaylor:
Awesome, thanks man!

SamyPesse:
Thanks 🍻

ianstormtaylor:
Published in `0.10.0`! Thanks so much man!

PR #204, opened 2016-07-29
RTL support

### LINKED ISSUE #134: handle both LTR and RTL text direction

SamyPesse:
This PR fixes #134 by adding support for RTL editing. it support editor contents with a mixture of LTR and RTL text.

The functionality is based on DraftJS implementation:

```
To that end, Draft uses a bidi algorithm to determine appropriate text alignment and direction on a per-block basis.

Text is rendered with an LTR or RTL direction automatically as the user types. You should not need to do anything to set direction yourself.
```

### TODOs

- [x] Accept `textAlignment` prop in `Editor`
- [x] Get text direction on a per-block basis (using [direction](https://github.com/wooorm/direction))
- [x] Update `directionMap` when content changed
- [x] Add `dir="rtl"` when required.
- [x] Fix tests

### DIFF
--- a/lib/utils/bidi.js
+++ b/lib/utils/bidi.js
@@ -1,14 +1,20 @@
 import direction from 'direction'
-import { OrderedMap, is } from 'immutable'
+import { Map } from 'immutable'
+
+let cache = {}
 
 /**
- * Get text direction for a block
- * @param  {Node} node
+ * Get direction of a text
+ * @param  {String} text
  * @return {String} dir?
  */
 
-function getDirForNode(node) {
-    let dir = direction(node.text)
+function getDirection(text) {
+    if (cache[text]) {
+        return cache[text]
+    }
+
+    let dir = direction(text)
     if (dir === 'neutral') {
         return
     }
@@ -16,36 +22,9 @@ function getDirForNode(node) {
     return dir
 }
 
-/**
- * Calcul direction map for all blocks
- * @param  {Node} node
- * @param {OrderedMap} prevBidiMap?
- * @return {OrderedMap}
- */
-
-function getDirectionMap(node, prevBidiMap) {
-    let map = {}
-    let lastValue = 'ltr'
-
-    node.filterDescendants(child => {
-        if (child.kind !== 'block') return
-
-        let dir = getDirForNode(child) || lastValue
-        map[child.key] = dir
-        lastValue = dir
-    })
-
-    let bidiMap = new OrderedMap(map)
-
-    if (prevBidiMap != null && is(prevBidiMap, bidiMap)) {
-        return prevBidiMap
-    }
-
-    return bidiMap
-}
 
 const bidi = {
-    getDirectionMap
+    getDirection
 }
 
 export default bidi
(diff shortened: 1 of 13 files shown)


SamyPesse:
@ianstormtaylor This is working, but here are some notes:

- The implementation is based on the same idea as Draft, where `directionMap` is [updated](https://github.com/facebook/draft-js/blob/1ce48f67728681fcb7c24a73ac26fa8b97a6e909/src/model/immutable/EditorBidiService.js#L27) at every [EditorState.push](https://github.com/facebook/draft-js/blob/1ce48f67728681fcb7c24a73ac26fa8b97a6e909/src/model/immutable/EditorState.js#L348); I think the perfs can be improved by:
    - Memoize the calls to `direction`
    - Only update the entries in `directionMap` for blocks that changed (by comparing a the new document with the old one)
- I've had to implement a `State.setDocument` so that all updates of document trigger an update of `directionMap`, Let me know if you still prefer to use `.merge({ document })` directly and I'll find another solution

ianstormtaylor:
@SamyPesse this is awesome!!! Going through now and reviewing!

ianstormtaylor:
I think we can make a few changes that will simplify things:

1. I think we should change `getDirectionMap()` to be a method on the `Document` model (and memoized). Ideally it would be recursive too, checking `getDirection(default)` on the `Block` model (which is passed the `lastValue`). That way we'd eliminate the need to keep it updated on `State`, but still have performance handled due to memoization.

2. As currently written is seems like this `textAlignment` property is just convenience for setting the `text-align` CSS property? What's the use case when that would be used? Off the top of my head, it seems like we're going to be determining the direction on a per-block basis, but if they wanted to prescribe that text alignment should always be right-aligned, they could add that themselves? Basically, no need for us to special case it, when they can just use the `style=` property instead.

3. Should the direction changing be handled in `shouldComponentUpdate` for nodes?

lib/utils/bidi.js
```diff
@@ -0,0 +1,51 @@
...
+        if (child.kind !== 'block') return
+
+        let dir = getDirForNode(child) || lastValue
+        map[child.key] = dir
+        lastValue = dir
+    })
+
+    let bidiMap = new OrderedMap(map)
+
+    if (prevBidiMap != null && is(prevBidiMap, bidiMap)) {
```
ianstormtaylor: We might not need this after memoizing, not sure.

SamyPesse:
@ianstormtaylor I've made two changes:

- I've removed the `textAlignment` prop
- I've removed the `directionMap` on State and replaced it by a `node.textDir`, the call to `direction(text)` is memoized (Do I need to implement a limitation on the memoized ? I've seen that there is none on `utils/memoize.js`)

ianstormtaylor:
Nice! Could you remove the old `setDocument`-related logic, and then I can help merge it in. I think I'll change `.textDir` to be `.direction`, and I'll figure out what we need to do for memoization logic!

SamyPesse:
@ianstormtaylor Done!

ianstormtaylor:
Awesome, thanks!

ianstormtaylor:
Published in `0.10.0` as well, thanks!

PR #205, opened 2016-07-29
Fix HTML serializer to only throw 'No serializer defined' when all rules have been tried

darobin:
With the current code, if the HTML serializer gets multiple rules, it will correctly skip those that do not have a `serializer` method but the first that has one that returns nothing (to signal it is not handling this case) will cause the code to throw. Instead, it now throws after the loop has completed.


ianstormtaylor:
Ah, good call! Thanks @darobin!

ianstormtaylor:
Published in `0.9.2`!

Issue #215, opened 2016-08-01
Rendering to IFrame


vleletko:
There is a problem in current version of Slate related to IFrame rendering. 

You can check it here: [https://jsfiddle.net/tq6n2jf8/2/](https://jsfiddle.net/tq6n2jf8/2/)

### Use case: 
* You want to have different styles for editor content & parent page. For instance apply global typography styles
* Page builders (responsive breakpoints)

### Problem 
Since code is evaluated in parent document context, Slate is using wrong `document` instance. 
It directly references `window.document` in 

1. `components/content.js`
2. `components/leaf.js` 
3. `plugins/core.js`
4. `utils/find-dom-node.js`
5. something else ?
 
### Options 
1, 2 - We can use React context mechanism or direct property propogation (did not investigated property option) with fallback to `window.document`. 

3 - Could be solved  by configuring `plugins/core` from `Editor` component props or by context value. 

4 - Add document as additional parameter. Maybe show console warning if parameter is not set (and use `window.document` as fallback).

ianstormtaylor:
Hey @vleletko interesting use case! How do other editors like Draft and Prosemirror handle this case?

It seems like an easy solution would be to render the iframe using a separate route in your application?

vleletko:
Hi, thanks for fast reply. 

Draft has same problem. 
Prosemirror don't use React so I guess it is quite hard to model such situation. 

The goal is to make detached css context for editor itself, but still have single react component tree. So iframe is always part of the editor. 

I guess the best example is a page builder app. 

Consider having property sheet and palette in top level document, They will have own styles.
And iframe wrapped editor with pure bootsrtap(or other) styles, not poluted by global styles.

This will guarantee that generated document will have same appearence as document in editor. 

I guess it is possible to keep all top level styles under some root css class. 
But it would be more natural to split css context. 

About separate root: 
 Can you please explain a bit, how separate application route will help ?

### Option 2
Actually we can make it one abstraction level deeper & introduce some kind of factory/service for Slate components to access `window.document` api

vleletko:
I will make PR if we find appropriate solution.

ianstormtaylor:
Interesting, great use case to try to solve for. I could see myself having a similar need. I'd be down for a PR that adds support for this!

[...]

PR #216, opened 2016-08-03
Add/render to iframe

vleletko:
As discussed in #215 , added support for Slate IFrame rendering, but in quite different way. 

There is actually no need to pass `dom={...}` becasue valid window could be deducted by the `event.native.target` transparently. 

One more problem found in React `SelectionEventPlugin`: it also has hardcoded `window` references.
Basicly version of `SelectionEventPlugin` swallowing all onSelection events, and this leads to incosistent slection state in slate.

To handle this I created small project [react-frame-aware-selection-plugin](https://github.com/vleletko/react-frame-aware-selection-plugin)

It is used in provided example.

### DIFF
--- a/examples/iframe-rendering/Readme.md
+++ b/examples/iframe-rendering/Readme.md
@@ -1,7 +1,14 @@
-
 # IFrame rendering example
 
 This example shows how to render Slate into IFrame, preserving single react component tree.
 You may need this if you want to have separate styles for editor content & application.
+
+In example this exmaple you can see, 
+that editor is using bootstrap styles, while they are not included to parent page. 
+
+## React onSelect problem
+Current react version has a problem with onSelect event handling, if input is rendered from parent component tree to iframe.
+
+This problem is solved by custom SelectEventPlugin - [react-frame-aware-selection-plugin](https://www.npmjs.com/package/react-frame-aware-selection-plugin)
 
 Check out the [Examples readme](..) to see how to run it!
--- a/examples/iframe-rendering/Readme.md
+++ b/examples/iframe-rendering/Readme.md
@@ -1,0 +1,7 @@
+
+# IFrame rendering example
+
+This example shows how to render Slate into IFrame, preserving single react component tree.
+You may need this if you want to have separate styles for editor content & application.
+
+Check out the [Examples readme](..) to see how to run it!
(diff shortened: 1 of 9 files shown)


ianstormtaylor:
@vleletko very cool! I like how minimal the changes are.

Question for you: can you open a pull request on React with the injector functionality proposed? I'm a little wary of having to use the `injector`, and would love to see that logic added natively to React if they agree with it.

ianstormtaylor:
Hey @vleletko also, I've just added the ability to "scroll to the cursor when a change is made" in `0.10.11`, but I think this logic may also need to be reconsidered for iframes? Can you take a look and see what you think? Thanks!

vleletko:
@ianstormtaylor, checked code related to "scroll to the cursor when a change is made", and yes, to support iframes it should be changed. 

Problem in [element-scroll-to](https://github.com/webmodules/element-scroll-to) code. 
It references to `window` and to `document.body`. 
Opened issue [#2](https://github.com/webmodules/element-scroll-to/issues/2)

About custom injector - I'll open issue in react repo, but I think they will not change it, at least [react-tap-event-plugin](https://github.com/zilverline/react-tap-event-plugin) still exists. 

btw, injector is used only in examples. We can remove Frame example from slate, and maybe make separate repo for it.

ianstormtaylor:
Added in `0.11.2`, thanks so much @vleletko for the help on this!

Updated the example to render with Bootstrap buttons and typography based on the rich text example. If you run into any issues while using it for your live preview let me know.

vleletko:
@ianstormtaylor, Thanks for great project!!

entiendoNull:
This may be the wrong place to ask, but I have been looking through documentation and also read through this post (that caught my attention to begin with): https://github.com/ianstormtaylor/slate/issues/215

My hope is to create a small page builder. The template included lives within an iframe so that I can isolate for example styling from my builder. I am doing this in React and my initial hope was using React-Quill as my rich text editor... but I stumbled over the exact same issue: Quill was instantiated in the parent document and text selection (- and surely other events too) were lost due to this.

I am certainly willing to perform tests my self, but still am eager to find out more about what this turned out like. I read somewhere that the `react-frame-aware-selection-plugin` is not maintained as much as it should... so was this functionality added directly to Slate? Or was interest lost somewhere along the way? Is there a straight forward way to achieve this?

Thanks for your time.
Your project looks amazing.

PR #217, opened 2016-08-03
Drop deprecated `escape` and `unescape`

w00fz:
As the title suggests, it'd be best to drop [`escape`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape) and [`unescape`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape) since they are both deprecated methods that serve no real purpose when in combination with `decodeURIComponent` and `encodeURIComponent`.

In addition, a quick [benchmark](https://jsfiddle.net/w00fz/3mgLe6bf/) shows a ~50% boost in speed.

### DIFF
--- a/lib/serializers/base-64.js
+++ b/lib/serializers/base-64.js
@@ -10,7 +10,7 @@ import Raw from './raw'
 
 function encode(object) {
   const string = JSON.stringify(object)
-  const encoded = window.btoa(window.unescape(window.encodeURIComponent(string)))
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Well then haha, that's a sweet diff. Thanks @w00fz !

ianstormtaylor:
Added in `0.10.10`!

Issue #218, opened 2016-08-03
Add an event handler for plugins when the editor receive a new state


SamyPesse:
I'm making progress on the schema validation ( [slate-schema](https://github.com/SamyPesse/slate-schema)).

But something is missing in the API to build a "validation" plugin: When the editor receive a new state modified by another part of the application (for example a Redux reducer), there is no event handler that can be called to process this new state before rendering.

Then a wrong state will be display before the user type something or move the selection, and the `onChange` is called to validate the state.

I propose to add a `onRender` or `onNewState` or `onReceiveState` event handler that will be chained like `onChange`.

----

PS-1: I know that it's possible to do:

```js
const myValidationPlugin = MyPlugin(...);

render() {
     return <Slate.Editor
          onChange={...}
          state={myValidationPlugin.onChange(this.state.state}
          plugins={[myValidationPlugin]}
    />;
}
```

But I think Slate can provide a better API.

----

PS-2: I understand that in theory it should not needed since the developer should not build transforms that can break its state, but when building a large rich text edition, preventing all the cases during transform can easily become a huge pain.

ianstormtaylor:
Interesting, I definitely understand the use case, but I also agree that it feels hackish from the theory standpoint. Do you have a few use cases in mind where this has come up already? Would be helpful to make sure I'm thinking about it correctly.

Also, I wonder if a solution might be to call it `onBeforeChange` and have it fire before any `onChange` handler, but including when passing a new state, so that "normalization" would only need to be implemented in that handler alone.

SamyPesse:
Closed by #219

PR #219, opened 2016-08-03
Add handler "onBeforeChange"

SamyPesse:
This PR is related to #218.

It adds an handler `onBeforeChange` that can be used to do validation before rendering:

Here is an example for a plugin to enforce that the document contains only one block:

```js
function MyPlugin(blockType) {
    function onBeforeChange(newState) {
        let { document } = newState;
        let { nodes } = document;

        if (nodes.size > 1) {
            nodes = nodes.slice(0, 1);
        }
        document = document.merge({ nodes });
        return newState.merge({ document });
    }

    return {
        onBeforeChange
    };
}
````

### DIFF
--- a/lib/components/editor.js
+++ b/lib/components/editor.js
@@ -199,8 +199,7 @@ class Editor extends React.Component {
       state = newState
     }
 
-    if (!this.props.onBeforeChange) return state
-    return this.props.onBeforeChange(state)
+    return state
   }
 
   /**
--- a/lib/components/editor.js
+++ b/lib/components/editor.js
@@ -53,6 +53,7 @@ class Editor extends React.Component {
     className: React.PropTypes.string,
     onChange: React.PropTypes.func.isRequired,
     onDocumentChange: React.PropTypes.func,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/components/editor.js [hunk 1]
```diff
@@ -180,6 +181,27 @@ class Editor extends React.Component {
...
+
+    for (const plugin of this.state.plugins) {
+      if (!plugin.onReceiveState) continue
+      const newState = plugin.onReceiveState(state, this)
+      if (newState == null) continue
+      state = newState
+    }
+
+    if (!this.props.onReceiveState) return state
+    return this.props.onReceiveState(state)
```
ianstormtaylor: This call should be unnecessary because of the logic in `resolvePlugins` creating an "editor plugin", I believe.

SamyPesse: Ok, I'll move it to the `CorePlugin`

SamyPesse:
@ianstormtaylor I've updated with your suggestion from #218

PR #227, opened 2016-08-06
WIP: Add transform wrapText

SamyPesse:
This goal of this PR is to implement a `wrapText` transform, this transform is a requirement for my application and I think it can be useful to other devs.

```js
state.transform()
    .wrapText('Hello, ', ', how are you?')
    .apply()
```

@ianstormtaylor I'm posting this PR early so that I can get your feedback and requirements on implementing a new transform (what are the required unit tests? etc).


ianstormtaylor:
Sweet! What's your use case for this? Just curious. On my phone now but I can do a better review later!

SamyPesse:
Well, it's not a very common use case of slate 😄: in our application, we have 2 modes for editing markdown: Document mode (WYSIWYG) and Raw mode (like a code editor), This transformation is useful in Raw mode to insert markup (like `**<selection>**`).

The raw mode is using the Prism highlighting per line to make it looks like a code editor.

At the end the goal is to remove the Raw mode, but we need first to support some complex block (raw html in markdown and templating syntax).

ianstormtaylor:
Nice, I think this is a great idea for a transform actually! Perfect for markdown editing. 

I think I'd like to make one small API addition, to make `after` optional, in which case it'll be the same as before. So you can do `wrapText('**')` for simplicity. And that's just cosmetic, other than that the API is perfect.

Question: how do you think the wrapping should perform across blocks? What does Markdown support?

Another question: should we add an `unwrapText`? We can do this later, but might be nice for Markdown use cases too.

Off the top of my head, I think we'll need a series of tests for this that look like:

- `start-of-block`
- `middle-of-block`
- `end-of-block`
- `whole-block`
- `across-blocks` (I think this one will fail as currently written!)
- `empty-block`
- `across-inlines`
- `empty-inline`
- `inside-void` (Same with this one, should just return early.)

SamyPesse:
@ianstormtaylor Thank you for your feedback.

I've updated the PR:

- The prefix defaults to suffix value
- I've added [tests](https://github.com/SamyPesse/slate/tree/add/transformWrapText/test/transforms/fixtures/wrap-text)

> Question: how do you think the wrapping should perform across blocks? What does Markdown support?

It depends, for example styling (bold `**...**`, italic `_..._`, etc) can only be applied in the same block, but wrapping for code blocks (with fences) are multilines.

Right now my implementation supports wrapping across multiple blocks (it only prepend the `prefix` and append the `suffix`). Do you think I should change it?

> Another question: should we add an unwrapText? We can do this later, but might be nice for Markdown use cases too.

Yes it can be useful, for example to implement a toolbar to toggle bold style (it will required to know if the current selection is wrapped by some text).

---

Do you want me to make more changes? This basic implementation seems enough for my use-case, but I'll understand if you prefer not to merge it since it's a edge-case transformation.

docs/reference/models/transform.md
```diff
@@ -32,6 +32,7 @@ Transform methods can either operate on the [`Document`](./document.md), the [`S
   - [`unwrapInline`](#unwrapinline)
   - [`wrapBlock`](#wrapblock)
   - [`wrapInline`](#wrapinline)
+  - [`wrapText`](#wraptext)
```
ianstormtaylor: Can you also add the equivalent `wrapTextAtRange` doc here as well? Thanks!

lib/models/state.js
```diff
@@ -1138,6 +1138,27 @@ class State extends new Record(DEFAULTS) {
   }
 
   /**
+   * Wrap the current selection with prefix/suffix.
+   * Cursor will be moved before the suffix
```
ianstormtaylor: I think we should make the cursor not change places, similar to how the other wrapping ones maintain the selection as it appears to the user. So if it starts as:

```
|text|
```

And is wrapped with `**` it would become:

```
**|text|**
```

Does that make sense?

ianstormtaylor:
@SamyPesse nice man! Just added a few comments for changes, but it looks almost perfect! I'm definitely down to have this transform in core though, no worries there. I agree that it's slightly more edge-case, but seems like a valid use case.

SamyPesse:
@ianstormtaylor I've made the changes, but maybe the solution to preserve the selection is not the right one.

cassiozen:
@SamyPesse @ianstormtaylor: I'm also working on a project with markdown and started working on a markdown serializer for slate https://github.com/netlify/slate-markdown-serializer

It's still in very early stages, but may be useful.

SamyPesse:
@cassiozen Our markdown parser for GitBook supports exporting JSON for Slate: https://github.com/GitbookIO/markup-it

It's not really a Slate serializer:

```js
const Slate = require('slate');
const MarkupIt = require('markup-it');
const markdownSyntax = require('markup-it/syntaxes/markdown');

const markdown = new MarkupIt(markdownSyntax);

// Parse the markdown
const content = markdown.toContent('Hello **World**');

// Convert the content to JSON
const rawJson = MarkupIt.SlateUtils.encode(content);

// Deserialize it using Slate.Raw
const state = Slate.Raw.deserialize(rawJson);
```

ianstormtaylor:
@SamyPesse amazing! Thank you so much for this!

cassiozen:
Very cool @SamyPesse.

ianstormtaylor:
@cassiozen thanks for sharing the serializer! I'll keep an eye on it 😄  (Also if you add `slate` to the `package.json` keywords it'll show up with the other plugins/tools!)

cassiozen:
Thanks, @ianstormtaylor, but I'm looking at @SamyPesse's solution and I might just start to use it.

PR #230, opened 2016-08-06
Fix typo in serialize methods

DevAlien:
While trying to serialize it to HTML I had an issue while copying from the docs/guide and now it should be fixed, was just a small typo.


ianstormtaylor:
Ah, thank you @DevAlien!

PR #232, opened 2016-08-08
Change concat into insertChildrenAfter and insertChildrenBefore

aunsuwijak:


ianstormtaylor:
@aunsuwijak nice I like this kind of diff :) is this just cleanup or are you noticing it fix a specific use case? If it's a fix, can you add a regression test to make sure it doesn't break again?

aunsuwijak:
@ianstormtaylor Sure.

ianstormtaylor:
Awesome, thanks!!

PR #233, opened 2016-08-08
Fix examples watch command

aunsuwijak:


ianstormtaylor:
Ah that was dumb on my part, thank you!

PR #236, opened 2016-08-10
change react-frame-aware-selection-plugin to 0.0.2, with safari fix.

vleletko:
just dependency version update


ianstormtaylor:
Awesome, thanks!!

PR #244, opened 2016-08-14
Add schema

ianstormtaylor:
Pushing this up as a pull request to get feedback that anyone might have.

This is an attempt to solve some of the issues that @SamyPesse has smartly brought up around "schema handling". After discussing it, I started to realize that it was really a consideration that should be handled as first-class by the core library, since it has the potential to add so much complexity if not.

The core library wouldn't itself enforce specific `types` of nodes, or enforce `data` about them, but it would provide the tools to do that, so that plugins can build on top of a standardized way of doing it.

For an example of what a schema looks like in the most basic case, the rich text example changes from:

```js
/**
 * Define a set of node renderers.
 *
 * @type {Object}
 */

const NODES = {
  'block-quote': (props) => <blockquote {...props.attributes}>{props.children}</blockquote>,
  'bulleted-list': props => <ul {...props.attributes}>{props.children}</ul>,
  'heading-one': props => <h1 {...props.attributes}>{props.children}</h1>,
  'heading-two': props => <h2 {...props.attributes}>{props.children}</h2>,
  'list-item': props => <li {...props.attributes}>{props.children}</li>,
  'numbered-list': props => <ol {...props.attributes}>{props.children}</ol>
}

/**
 * Define a set of mark renderers.
 *
 * @type {Object}
 */

const MARKS = {
  bold: {
    fontWeight: 'bold'
  },
  code: {
    fontFamily: 'monospace',
    backgroundColor: '#eee',
    padding: '3px',
    borderRadius: '4px'
  },
  italic: {
    fontStyle: 'italic'
  },
  underlined: {
    textDecoration: 'underline'
  }
}

  /**
   * Render the Slate editor.
   *
   * @return {Element}
   */

  renderEditor = () => {
    return (
      <div className="editor">
        <Editor
          placeholder={'Enter some rich text...'}
          state={this.state.state}
          renderNode={this.renderNode}
          renderMark={this.renderMark}
          onChange={this.onChange}
          onKeyDown={this.onKeyDown}
        />
      </div>
    )
  }

  /**
   * Return a node renderer for a Slate `node`.
   *
   * @param {Node} node
   * @return {Component or Void}
   */

  renderNode = (node) => {
    return NODES[node.type]
  }

  /**
   * Return a mark renderer for a Slate `mark`.
   *
   * @param {Mark} mark
   * @return {Object or Void}
   */

  renderMark = (mark) => {
    return MARKS[mark.type]
  }
```

Into:

```js
/**
 * Define a schema.
 *
 * @type {Object}
 */

const schema = {
  nodes: {
    'block-quote': props => <blockquote {...props.attributes}>{props.children}</blockquote>,
    'bulleted-list': props => <ul {...props.attributes}>{props.children}</ul>,
    'heading-one': props => <h1 {...props.attributes}>{props.children}</h1>,
    'heading-two': props => <h2 {...props.attributes}>{props.children}</h2>,
    'list-item': props => <li {...props.attributes}>{props.children}</li>,
    'numbered-list': props => <ol {...props.attributes}>{props.children}</ol>,
  },
  marks: {
    bold: {
      fontWeight: 'bold'
    },
    code: {
      fontFamily: 'monospace',
      backgroundColor: '#eee',
      padding: '3px',
      borderRadius: '4px'
    },
    italic: {
      fontStyle: 'italic'
    },
    underlined: {
      textDecoration: 'underline'
    }
  }
}

  /**
   * Render the Slate editor.
   *
   * @return {Element}
   */

  renderEditor = () => {
    return (
      <div className="editor">
        <Editor
          placeholder={'Enter some rich text...'}
          schema={schema}
          state={this.state.state}
          onChange={this.onChange}
          onKeyDown={this.onKeyDown}
        />
      </div>
    )
  }
```

Which makes the default case of "types are unique, and my components map exactly to one type" very simple.

But the `schema.nodes` and `schema.marks` syntaxes are actually just shorthands for a more flexible underlying `schema.rules` implementation. Such that a simple `paragraph` node implemented as:

```js
const schema = {
  nodes: {
    paragraph: props => <p {...props.attributes}>{props.children}</p>
  }
}
```

Is actually equivalent to `rules` of:

```js
const schema = {
  rules: [
    {
      match: (object) => {
        return (
          (object.kind == 'block' || object.kind == 'inline') &&
          object.type == 'paragraph'
        )
      },
      render: (props) => {
        return <p {...props.attributes}>{props.children}</p>
      }
    }
  ]
}
```

So the whole system is really flexible.

In addition to matching for component rendering, it allows you to make "validations" about the editor's content. For example, by default Slate requires that the `document` only contain `Block` node children, which previously was not enforced anywhere, but now is expressed as:

```js
const schema = {
  rules: [
    {
      match: (object) => {
        return object.kind == 'document'
      },
      validate: (document) => {
        const { nodes } = document
        const invalids = nodes.filter(n => n.kind != 'block')
        return invalids.size ? invalids : null
      },
      transform: (transform, document, invalids) => {
        return invalids.reduce((t, n) => t.removeNodeByKey(n.key), transform)
      }
    },
    ...
  ]
}
```

Which is nice because I think it will let us move all of the `Selection.normalize` and `Node.normalize` logic into the schema itself, in a much more performant way.

Let me know if you have thoughts! I'll probably merge this pretty soon.


SamyPesse:
Awesome 🎉 (One less repository to maintain 😄 ).

The API looks good and flexible. Can plugins extend the schema ? (Not just replace it)

ianstormtaylor:
@SamyPesse yup! The core schema is defined by the "core" plugin, so it is added last in the plugins stack can thus can be overridden by any plugin before it.

I've actually just pushed an update that simplifies the schema logic a bit for Slate's side (examples updated above to reflect it). With that change, now all of the schema rule arguments are _always_ functions. I realized that trying to accommodate things like:

```js
{
  validate: {
    nodes: {
      anyOf: [
        { kind: 'block' }
      ]
    }
  }
}
```

Was adding a lot of extra maintenance complexity and only accidentally restricting things the schema could do (like validate `Data` values), without tons of benefit. Instead, it now has to be a function, so the above would be expressed as:

```js
{
  validate: (node) => {
    const invalids = node.nodes.filterNot(child => child.kind == 'block')
    return invalids.size ? invalids : null
  }
}
```

But it would still be possible for a non-core library to be written for declarative schemas if anyone wanted to, which would just map to functions before passing them to Slate.

SamyPesse:
Yes using functions is much better (a library can provide composable helper functions for this).

SamyPesse:
@ianstormtaylor  it might be great to update the table example to showcase a validation "same count of columns per row", and allow pressing enter to split block which should create new columns because of the validation.

PR #248, opened 2016-08-17
terse text can contain marks

cassiozen:
Text nodes in terse form can contain marks - but currently they are being ignored when "untersifying Text".

Referenced in https://github.com/GitbookIO/markup-it/pull/23


ianstormtaylor:
Nice! Thanks @cassiozen!

Issue #178, opened 2016-07-26
Question - embedding editors within editors (editor inception)


tyler-johnson:
I am wondering the best approach for placing editors within editors. On the surface this seems very trivial with Slate compared to other WYSIWYGs, however I need the histories (undo/redo) of the editors to be connected.

My use case goes like this:

1. User types text into editor.
2. User highlights inline text and adds a "popup" (ie. annotation). This action pauses the original editor and opens a second editor for the popup's contents.
3. User decides after creating the popup that they don't want the popup, so they press Cmd-z to undo. Ideally this would reverse the previous state, destroying the popup and returning to the original text.
4. Then the user decides they really did want the popup, so within the original editor the press Cmd-Shift-z thus redoing the previous action and returning to the popup mode.

After a brief reading of the Transform, State and Document code this seems rather doable by sharing the History object between editor states. What is tripping me up is how to let the states know which editor the history was made in so it can smartly switch between editors as the user undoes and redoes changes.

I would like to mention that I have implemented this scenario in many, many different WYSIWYGs with varying degrees of success. The closest I have come so far is with Quill, which our production application is currently running. Not a single editor manages to come close to having undo/redo support with embed editors and all of them had issues with switching between two editors. To me this is the true power of Slate and IMO is the absolutely the future of WYSIWYG on the web. So kudos for doing what everyone else has fallen short at.

ianstormtaylor:
Thanks 😄  glad you like it!

Question, since you have more context on this than me, what changes would you make to the architecture to make this possible?

tyler-johnson:
Okay after giving this a lot of thought and a thorough reading of Slate's code, I guess what I am asking for is **multiple document/editor support**.

I'm not totally sure how this manifests itself in Slate, but in my mind documents/editors would share a couple of features:

- Undo/redo history. Ideally typing cmd-z in one editor would undo a change in different editor (if that was the last change in history). I know with immutability this may be difficult to do transparently, however I think there is a real use case here. Our users have been asking for it for a while now, but it has been incredibly complex to achieve in Quill.
- *Some, but not all* plugins. Obviously this already doable, however I'm thinking more that there is a shared state of sorts so that plugins could be aware of all the connected documents/editors.

ianstormtaylor:
The history one makes sense to me. What is the "some, but not all plugins" need for?

tyler-johnson:
[...]

PR #249, opened 2016-08-18
Add operations

ianstormtaylor:
This is an ongoing pull request that I'm working on to refactor all of the transforms to be built up from smaller "operations", as a step towards collaborative editing.

Eventually I think it will contain a breaking change that takes us from:

```js
state
  .transform()
  .delete()
  .insertText('Something...')
  ...
  .apply()
```

To something like;

```js
editor
  .change()
  .delete()
  .insertText('Something...')
  ...
  .apply()
```

Which will give move `History` to be editor-level, rename the current `Transform` to `Change`, and which will make `History` and `Change` mutable. It will also allow for defining your own transform functions, and all of the core ones will be defined in the same way, and loaded last via the "core" plugin.

All of the transform functions actually end up making smaller operations on the state. The current plan for operations will be:

```
- insert text
- remove text
- add mark
- remove mark
- update mark
- insert node
- remove node
- update node
- move node
- split node
- merge node
- update selection
```


SamyPesse:
@ianstormtaylor does it means that transforms/change will move from the `State` to the react component?

What will be the workflow for a FLUX/Redux application?

ianstormtaylor:
@SamyPesse yup, that was what I had been throwing around at least. But I haven't thought it through all the way yet (working in stages), so if that screws up your workflow I'd love to hear how and figure out what to change about the idea!

Could you explain how your currently using it and how that change would mess it up? I used Redux in the past but haven't used it in a long while now. Thanks man!

ianstormtaylor:
Ah @SamyPesse sorry for the confusion, you're totally right. Just realized while making breakfast 😄 . I was trying to solve for the history issues in https://github.com/ianstormtaylor/slate/issues/178, but just realized that only needs to have `History` become editor-level, and keep the transforms state-level as would make sense.

The only difference is that `.undo()` and `.redo()` will no longer be transforms, instead _they'd_ be called on the editor, which seems like how it should have been already!

(Editing that top post to convey this!)

SamyPesse:
Why do we need History to be editor-level? In our application, it's also better to have the history state-level (Basically pressing `Ctrl-Z` in the application dispatch the action `editor.undo` which emit a new state for the application with a slate state which is `state.transform().undo().apply()`.

I truly believe that such operations on data should be kept at the state-level. But maybe I'm missing something important.

Concerning transforms and plugins, the main issue is that plugins have to deal with state-level and editor-level.

----

### Idea

One solution can be to have "state-level" plugins or extend the state with an `operations: Map(String:Function)` property that is updated when plugins are updated on the react component:

A plugin can be define like this:

```js
module.exports = function TablePlugin() {
     ...

     function insertRow(transform, ...args) {
        ...
     }

     return {
        // Classic plugin props
        onKeyDown,
        ...

        // Provides a map of operations
        operations: {
            insertRow,
            insertColumn
        }
     };
}
```

Then when the editor is initialised (mount or receiving new plugins by props), it can update the `operations` map in the state:

```js
componentDidMount: function() {
    const { state, plugins, onChange } = this.props;

    // Extract all operations from plugins
    const newOps = fromPlugins(plugins);

    let { operations } = state;
    operations = operations.merge(newOps);
    
    state = state.merge({ operations });

    // Emit the state binded to plugins
    onChange(state);
}
```

Then calling `state.transform()` will basically be the creation of a `Transform` with operations binded to it:

```js
transform() {
    const state = this;
    const { operations } = state;

    const tr = new Transform(state);

    operations.forEach((fn, name) => {
        tr[name] = function(...args) {
            return tr.dispatch(name, args)
        }

        tr.on(name, fn);
    });

    return tr;
}
```

I'm imaging `transform.dispatch(name: String, args: Array)` as the creation of a serializable operation that can be transferred for real-time edition.

And `transform.on(name: String, fn: Function)` as the binding of an operation to a transformation function.

Maybe I don't really understand how you are imaging operations for real-time edition (I'm not personally interested in this feature right now), but from my understanding it requires actions to be serializable and reproducible when receive in a specific order on other editors (basically add a separation between the creation/dispatch of the operation and its application on the state).

---

CC @soreine

cobbweb:
Is there going to be a way to insert a block before a void block with the new Transforms?

ianstormtaylor:
@cobbweb yup! The lower level `insertNodeByKey` takes an index, so you can insert anywhere!

Soreine:
Hey @ianstormtaylor,

This PR brings very useful features, which will greatly help us in our current work. We are working on a several tasks in parallel and need to schedule them. Our schedule could depend on the estimated completion of this PR. Do you have an idea of how much is left to do ? And do you know when will you have time to finalize it ? Any estimation you can give will prove valuable for us.
Not meaning to be blunt or pushy, just genuinely concerned.

As always, of course, we would be glad to assist you or advise you on on that matter. Let us know if we can help!

Cheers :)

ianstormtaylor:
Hey @Soreine! Totally valid question and not pushy at all. I had kind of run into a few different roadblocks while trying to get all of the operations completed. I actually just started working again on what _I think_ is the last one:

I needed to figure a way to handle selections in the history. Since the current way of setting a selection after updating the document tree doesn't work for undo/redo where operations are inverted. (For undo'ing it will try to set the selection back to a state that doesn't exist yet!)

I think I'm actually really close to having it completed. Maybe even today. When I do, I'm going to publish it as `@next` on npm, and it would be awesome if you could test it out to ensure that nothing crazy is broken. There might be one or two edge cases to fix, and probably one or two edge cases that were broken previously but fixed now 😄

ianstormtaylor:
I published on `next` and already checked it out on my own codebase actually, so now it's on `latest` as version `0.14.0`! If you notice anything strange, open an issue as usual! Thanks so much

lib/transforms/on-selection.js
```diff
@@ -364,6 +364,7 @@ export function moveForward(transform, n) {
  */
 
 export function moveTo(transform, properties) {
+  debugger
```
Soreine: Happens to me all the time :)

Issue #245, opened 2016-08-15
make `isVoid` determined by schema


ianstormtaylor:
Now that we have the concept of a schema, we no longer need to make `isVoid` a property of each node, and we can instead have it determined by the schema.

cobbweb:
I wonder if this should be defined by the Block itself? Is there a use-case for a single block type that can operate as both void node _or_ having children?

Or if for some reason you did have a node that could be either, you could make a getter function for `block.isVoid`?

```
function ImageBlock({ node }) {
  return <img src={node.data.src} />;
}

MyBlock.isVoid = true;

const schema = {
  nodes: {
    image: ImageBlock
  }
};
```

If people were to start distributing blocks for others to use, this style would make them easier to consume?

ianstormtaylor:
I think the issue with putting it purely in the component land is that transforms need to know about whether it's void or not so they can know to use different deleting/etc. behavior.

I think the simplest for now would actually be to keep it the way it is, but allow for an easy way for the schema to ensure that certain nodes are always void. So that you wouldn't have to care about it anymore.

cobbweb:
Can the transform not pull it off the Component through the schema though?
I guess in my example `MyBlock.isVoid` is a static prop, so my getter idea wouldn't work as cleanly as I originally thought (if someone needed such functionality). 

Assuming it can be access from the Component, that still feels like the right place to me. 
Renderers are written with a specific `isVoid` value in mind, but don't really have a way to enforce this. This leads to edge-cases like if you call `Transform.setBlock('paragraph')` when the current selection is a void block, you end up with a [void paragraph](https://jsfiddle.net/f6L257bs/3/) (which cannot be edited). 

I suppose having `isVoid` in the schema might fix that, but you've still got an implicit dependency in the renderer that the Node's `isVoid` is a particular value that it cannot enforce.

ianstormtaylor:
Yeah definitely tricky. I think it needs to either be in the schema or in the state itself though, and not in the renderer, because you shouldn't need to be rendering something to be able to transform it properly (eg. imagine runnning transforms server-side).

cobbweb:
Ah yep, gotcha. Schema it is!

ianstormtaylor:
Closing since this is a userland schema choice after our decision. There could probably be a bit better docs for this use case (or adding to images example) if someone wants to PR!

PR #276, opened 2016-08-30
Add failing test showing splitBlock with isVoid bug

cobbweb:
Not sure what I actually expected calling `splitBlock()` on void block, but I would expect `splitBlock()` to respect a given block type's isVoid property.
This failing test demonstrates that `splitBlock()` always creates a the new block with `isVoid: false`.

Edit: Sorry, here's the test runner output... the second block has `isVoid: false` and some extra child nodes:

image

### DIFF
--- a/lib/models/transforms.js
+++ b/lib/models/transforms.js
@@ -612,7 +612,8 @@ const Transforms = {
       secondChild = Block.create({
         nodes: secondChildren,
         type: parent.type,
-        data: parent.data
+        data: parent.data,
+        isVoid: parent.isVoid
       })
 
       // Add the new children.
--- a/test/transforms/fixtures/split-block/with-void/index.js
+++ b/test/transforms/fixtures/split-block/with-void/index.js
@@ -1,0 +1,24 @@
+
+import assert from 'assert'
+
+export default function (state) {
+  const { document, selection } = state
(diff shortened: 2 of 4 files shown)


cobbweb:
If this is a valid bug, I can have a quick go at fixing it.

cobbweb:
Could be potentially fixed by #245.

cobbweb:
Potentially related to my real issue: it looks like you can't insert a block before a void node according to [insertBlockAtRange](https://github.com/ianstormtaylor/slate/blob/0413d93305e13b3baa38741cd51cf5cbe369ba03/lib/models/transforms.js#L252)?

ianstormtaylor:
Sorry for slowness on this @cobbweb, I'd say this is definitely valid. If you're down to have a go at fixing it that would be awesome!

cobbweb:
Ok, turned out to be a super easy fix (changes pushed). I think there will be a similar bug in `splitInlineAtRange`, as there's a similar [chunk of code](https://github.com/ianstormtaylor/slate/blob/3c46d0544c90d7e84adbc6701e195a40cc0d5e2d/lib/models/transforms.js#L668). I left it alone though as I don't know much about Inline's yet.

ianstormtaylor:
Beautiful, thanks man!

PR #280, opened 2016-08-31
Fixes a small error in a code example in HTML walkthrough.

nickpassarella:
This is a tiny thing, but I caught this when working through the walkthrough.


ianstormtaylor:
Thanks!

Issue #286, opened 2016-09-01
Copy-paste of multiple blocks in a nested block fails


Soreine:
Hello @ianstormtaylor,

I think I have found a bug in `insertFragmentAtRange`, when pasting a fragment of several blocks into a nested block.

You can see it in action here https://github.com/GitbookIO/slate-edit-list/issues/6

[Pull Request](https://github.com/ianstormtaylor/slate/pull/287) that adds a fixture for this case.

PR #287, opened 2016-09-01
Fixture for insertFragmentAtRange with multiple blocks in nested block

Soreine:
This adds a fixture related to this bug https://github.com/ianstormtaylor/slate/issues/286

I might have made a mistake in the `expected.yaml`. But the bug happens before that, because `insertFragmentAtRange` simply throws an error.

I think at this point in the code, for this failing test, `node` is the document, whereas it should be the block where we are pasting: https://github.com/ianstormtaylor/slate/blob/master/lib/models/transforms.js#L386


ianstormtaylor:
Thanks @Soreine! Would you be down to have a go at writing the fix?

Soreine:
I have fiddled with the code, but I have not found a solution with the time I have. The only feeling I have, is that [here](https://github.com/ianstormtaylor/slate/blob/d6901aa416120dc86b8fa9f6bbfff17f47f70c58/lib/models/transforms.js#L386) the `fragment.nodes` are added directly to the node, whereas it should be added to some other parent block (need to figure that out). The assumption that `block` is a direct child of `node` is wrong here.

I still don't understand everything happening in `insertFragmentAtRange`. There is a lot of cases taken care of, and all three variables `node`, `fragment` and `block` are updated throughout the function, I find it hard to follow. The comments really help to understand what each part of the code does, but it does not help to understand why from a higher level. It might be worth to add a comment at the top of `insertFragmentAtRange`, to explain what edge cases this function deals with and how.

I'd be happy to help, or answer questions.

Soreine:
I rebased the PR for 0.14.2. The code does not throws anymore. But the test is still failing. I adapted it a bit.

Input
```yaml

nodes:
  - kind: block
    type: list-item
    nodes:
      - kind: block
        type: paragraph
        nodes:
          - kind: text
            ranges:
              - text: First child
      - kind: block
        type: paragraph
        nodes:
          - kind: text
            ranges:
              - text: Second child
```

Fragment
```yaml
nodes:
  - kind: block
    type: paragraph
    nodes:
      - kind: text
        ranges:
          - text: fragment one
  - kind: block
    type: paragraph
    nodes:
      - kind: text
        ranges:
- text: fragment two
```

Output
```yaml

nodes:
  - kind: block
    type: list-item
    nodes:
      - kind: block
        type: paragraph
        nodes:
          - kind: text
            text: Fifragment one
      - kind: block
        type: paragraph
        nodes:
          - kind: text
            text: fragment tworst child
      - kind: block
        type: paragraph
        nodes:
          - kind: text
            text: Second child
```

It seems that the two last blocks are in wrong order.

```
  1) transforms at-range insertFragmentAtRange() fragment-multiple-blocks-in-nested-block:

      AssertionError: { nodes: [ { kind: 'block', type: 'list-item', nodes: [Object] } ] } === { nodes: [ { kind: 'block', type: 'list-item', nodes: [Object] } ] }
      + expected - actual

                 "kind": "block"
                 "nodes": [
                   {
                     "kind": "text"
      -              "text": "Second child"
      +              "text": "fragment tworst child"
                   }
                 ]
                 "type": "paragraph"
               }
                 "kind": "block"
                 "nodes": [
                   {
                     "kind": "text"
      -              "text": "fragment tworst child"
      +              "text": "Second child"
                   }
                 ]
                 "type": "paragraph"
               }
```

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/309

PR #292, opened 2016-09-04
Fix bug with non-immediate updating of readOnly state

### LINKED ISSUE #262: read only mode doesn't update immediately
As reported via Slack, enabling read-only will sometimes leave the cursor in a block, which allows a single extra edit to be made before the contents is actually read-only.

Fiddle to demonstrate: https://jsfiddle.net/u0tfxf7u/3/

I think this is probably because `shouldComponentUpdate` in `Node` doesn't take into account read-only for blocks, so the cursor isn't removed. Might be able to add that logic to `Content`, or could pass down the `readOnly` prop to leaves?

edencore:
Fixes #262

### DIFF
--- a/lib/components/content.js
+++ b/lib/components/content.js
@@ -93,6 +93,8 @@ class Content extends React.Component {
    */
 
   shouldComponentUpdate = (props, state) => {
+    // If the readOnly state has changed, we need to re-render so that
+    // the cursor will be added or removed again.
     if (props.readOnly != this.props.readOnly) return true
 
     // If the state has been transformed natively, never re-render, or else we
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Awesome, can you add a comment above it like so: (for future us)

```js
// If the readOnly state has changed, we need to re-render so that
// the cursor will be re-rendered.
if (props.readOnly != this.props.readOnly) return true
```

edencore:
Yeah, sure. Done

ianstormtaylor:
Awesome, thanks for the PR!

PR #293, opened 2016-09-04
fixed a few typos

MatthewKosloski:
There were two occurrences of the state attribute in the Editor element; I believe one was for the schema.


ianstormtaylor:
Awesome, thank you!

PR #294, opened 2016-09-04
fixed a typo where the method for focus was blur()

MatthewKosloski:
In the Editor reference documentation, there was a typo.  Blur() is listed twice; once under "blur" and again under "focus."


ianstormtaylor:
Thanks @MatthewKosloski !

PR #296, opened 2016-09-07
Fix exception thrown when normalizing selection on empty documents

mull:
I wasn't entirely sure on where to put the test so I added a new folder. Unless I'm blind there are no unit tests anyhow. Maybe I'm wrong but I also couldn't find a library to do regular asserts :-)

### DIFF
--- a/lib/models/selection.js
+++ b/lib/models/selection.js
@@ -288,6 +288,11 @@ class Selection extends new Record(DEFAULTS) {
     const { isCollapsed } = selection
     let { anchorKey, anchorOffset, focusKey, focusOffset, isBackward } = selection
 
+
+    if (node.getTexts().size === 0) {
+      return this
+    }
+
     // If the selection isn't formed yet or is malformed, set it to the
     // beginning of the node.
     if (
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "react-dom": "^0.14.0 || ^15.0.0"
   },
   "devDependencies": {
+    "assert": "^1.4.1",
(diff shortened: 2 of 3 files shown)


mull:
It rarely if ever makes sense to have an empty document, because AFAIK then there's nothing to focus on or write into. Less convinced about this PR now.

The bigger issue is probably what to do when you end up with an empty document and how the API can give the serializer a default node to use instead. If that makes sense. :-)

ianstormtaylor:
Mmm yep, I agree with you. For your use case, what would a solution look like? Potentially could be a schema-specific validation?

mull:
Here's a more clear use case that doesn't involve serializing:

We're supporting LaTeX editing through MathJax. These are custom void blocks. To edit the formula for a math block you have to click on it, which renders an overlay with an editor over the rest of the editor.

This editor has buttons to Save/Cancel/Remove. The Remove button calls `transform.removeNodeByKey(key)`. Without additional logic here I would get an error when calling `apply()` if the math block was the only one in the tree.

This does make sense, I guess. I'm drifting between whether I think the editor should be smart enough to insert a default block (specified through props) if there is none available or not.

So my solution in pseudo-code looks like:

```
if onlyBlock
  convertBlockToAnotherType
else
  removeBlock
```

ianstormtaylor:
Hey @mull inadvertently I needed to allow for `anchorKey` and `focusKey` to be null for the `add-operations` changes, so in `0.14.0` I believe your use case should no longer error? Let me know if that's so and we can close this out! Otherwise we can find a solution still.

PR #309, opened 2016-09-13
Fix insert fragment multiple nested

ianstormtaylor:
cc @Soreine

In the newly refactor codebase this is an embarrassingly simple fix 😄


Soreine:
Haha, well I'm glad this ended up fixing something in the end

Issue #302, opened 2016-09-12
In image example, selected image cannot be drag n drop


aunsuwijak:
![rrgqahja88](https://cloud.githubusercontent.com/assets/5433223/18431405/f09b609c-7906-11e6-8b89-ec0a9066283f.gif)

When drag and drop selected image (blue border), image does not move and throws an error.

ianstormtaylor:
Thanks for finding the fix @aunsuwijak !

PR #316, opened 2016-09-14
Fix issue #302

aunsuwijak:
Fix this https://github.com/ianstormtaylor/slate/issues/302

@ianstormtaylor Not really sure, but I think to prevent this kind of problem should `removeNodeByKey` unset selection before doing it's job?

### DIFF
--- a/examples/images/index.js
+++ b/examples/images/index.js
@@ -178,6 +178,7 @@ class Images extends React.Component {
   onDropNode = (e, data, state) => {
     return state
       .transform()
+      .unsetSelection()
       .removeNodeByKey(data.node.key)
       .moveTo(data.target)
       .insertBlock(data.node)


ianstormtaylor:
Ah good call! Thanks @aunsuwijak 

I guess thinking about it now... the super technically correct way would be instead to actually split and then move the node instead of removing and inserting again... but this sounds like a good fix since it's just an example!

aunsuwijak:
@ianstormtaylor Agree with that. FYI, this occurred many places for example I put the similar fix to the table plugin. May be we could put `unsetSelection` in the transform document, what do you think?

ianstormtaylor:
@aunsuwijak yup, we totally should. If you PR it I'll merge!

PR #319, opened 2016-09-14
Fix Text.updateMark

Soreine:
I was testing the `setMarkByKey` transform, and I think there is a bug here, because I got `marks.indexOf is not a function` :
https://github.com/ianstormtaylor/slate/blob/8fc3631328481b3c56c48a56e1a9bad2eb24bd5a/lib/models/text.js#L294

`char.marks` is an `Immutable.Set` which does not have `.indexOf` method.

### DIFF
--- a/lib/models/text.js
+++ b/lib/models/text.js
@@ -3,7 +3,7 @@ import Character from './character'
 import Mark from './mark'
 import memoize from '../utils/memoize'
 import uid from '../utils/uid'
-import { List, Record, Set } from 'immutable'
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @Soreine thanks for finding this! I just pushed up a change that fixes it.

I realized that I never added tests for all of the new `*ByKey` transforms yet, so some of them are probably non-functional. Totally slipped my mind. If you come across any other, I'd be happy to accept fixes/tests! Otherwise I'll get to them eventually 😄 

Thanks!

PR #322, opened 2016-09-15
change toSet to toOrderedSet

### LINKED ISSUE #321: Images example: document.getBlocks doesn't work as expected
In the Images example [document.getBlocks is used](https://github.com/ianstormtaylor/slate/blob/b7f317a/examples/images/index.js#L117-L119), but as far as I can see [getBlocks](https://github.com/ianstormtaylor/slate/blob/88f6ee5e12aa76b95a329ae075358f0e48b0e836/lib/models/node.js#L159) doesn't give ordering guarantees, so checking for the last returned value won't always work.

We're working around this like this:

```javascript
const lastText = document.getTexts().last()
const lastBlock = document.getClosestBlock(lastText)
```

Is that sensible? Thanks!

somus:
Fixes #321

### DIFF
--- a/src/models/node.js
+++ b/src/models/node.js
@@ -160,7 +160,7 @@ const Node = {
     return this
       .getTexts()
       .map(text => this.getClosestBlock(text))
-      .toSet()
+      .toOrderedSet()
       .toList()
   },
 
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thanks @somus !!

PR #324, opened 2016-09-15
Remove dist files

bkniffler:
The dist files currently lurk around the repo, hereby they be gone!


ianstormtaylor:
Beautiful, thanks!

Issue #317, opened 2016-09-14
Inline node bugs


hector-esv:
On the Links demo:

1. Inserting a link and then undoing throws Uncaught TypeError: Cannot read property 'size' of undefined.
2. Inserting a link at the start of a line, and then backspacing from the end of the link to its beginning: as the first character is deleted it throws Uncaught Error: Could not find a descendant node with key "xinw".
3. Adding text immediately after the link styles the text as part of the link. In other words it doesn't seem easy to add normal text after the link.

I'd be prepared to contribute fixes, but please provide some pointers on what would need to be done. Thanks!

kommen:
One more issue I just ran into in our app and can reproduce in the Links demo:

1) Delete the `!` after `hyperlinks` in the first line
2) Press Enter

This results in `TypeError: undefined is not an object (evaluating 'node.nodes.first().kind')`

ianstormtaylor:
Thanks for reporting this @hector-esv! I just split this up into multiple errors, since they are solved independently. I'm going to close this one, and tackle them in those smaller issues instead!

Issue #167, opened 2016-07-23
Collapse to end of parent block? (try add new table-row)


joehua87:
I've try to add a new table-row to table but end up like this: table-row & table-cell is at same level
```
[
  {
    "kind": "block",
    "type": "table-cell",
    "nodes": [
      {
        "kind": "text",
        "ranges": [
          {
            "text": "9"
          }
        ]
      }
    ]
  },
  {
    "kind": "block",
    "type": "table-row",
    "nodes": [
      {
        "kind": "text",
        "ranges": [
          {
            "text": ""
          }
        ]
      }
    ]
  }
]
```
I've try difference ways but cannot insert new table-row as the same level
```js
    // Assume that we're focus on an tabel-cell
    // Get the first block (should be table-cell)
    const cellNode = state.blocks.first()

    // Get the Table
    const tableBlock = state.document.getClosest(cellNode, (block) => block.type == 'table')
    if(!tableBlock) return

    // Get the last row
    const lastRow = tableBlock.nodes.last()
    console.log('Last Child', lastRow.toJS()) // Show table-row here

    state = tranform
      .collapseToEndOf(tableBlock)
      // .collapseToEndOf(lastRow)
      .splitBlock()
      .setBlock('table-row')
      .apply()
```

ianstormtaylor:
Ah shoot yup. This is because selections are always "relative to" the leaf text nodes. So that `collapseToEndOf` actually ends up setting the cursor at the end of the text node in the last table cell of the block. And then that one gets split and the newly split one changed to a row. 

You could try using `insertFragment` as a stopgap for now, but it will feel a little clunky until we can get the `insertNode` transform (and others) that operate at the node level regardless of selection. I was working in that yesterday but it'll still be a few more days at least until it's ready I think.

bkniffler:
Hey, 
I'm trying this now with insertNodeByKey, but i can't seem to get this to work.

If I use 

```jsx
.insertNodeByKey(
    tableBlock.key,
    tableBlock.nodes.length,
    Block.create({ type: 'table-row', data: { } }) // or simply 'table-row'
)
```

I get 
```jsx
// components/node.js:347 Uncaught TypeError: Cannot read property 'kind' of undefined, node.nodes.first().kind

if (node.kind == 'block' && node.nodes.first().kind != 'block') {
      var direction = node.getTextDirection();
      if (direction == 'rtl') attributes.dir = 'rtl';
}
```

[...]

PR #329, opened 2016-09-16
Check if .first() exists in component/node.js functions, check if startkey set in models/node.js

bkniffler:
Might fix issues posted https://github.com/ianstormtaylor/slate/issues/317 and https://github.com/ianstormtaylor/slate/issues/167

Is that okay? Or would you rather make sure there always is at least one node within a nodes' nodes for the .first() issue?


ianstormtaylor:
Hey @bkniffler I believe we've solved these issues now via other means, so we don't need to have the extra checks. I'm going to close this, but if you still run into issues let me know!

PR #336, opened 2016-09-17
Fix #318: Implement "splitBlockAtRange" as a node method

### LINKED ISSUE #318: Copy & paste doesn't work as expected
In `0.14.3` copying of fragments inside the editor gives an error:

```
Uncaught TypeError: node.splitBlockAtRange is not a function
```

Now if i paste this inside the same editor, it is being handled as type `html` which leads to the loss of all formatting and custom nodes etc.

SamyPesse:
This PR is intended to fix #318.

### Current Bug

Since upgrading to 0.14, `splitBlockAtRange` is no longer a method of `Node` (before it was a transform that was also set as a Node's method).

### Fix Implementation

I was not sure what kind of solution to apply to this problem, so feel free to reject this PR.

1. I've moved the code from the `splitNode` operation into a Node method
2. I've modified the `split_node` operation to call `node. splitNode`
3. I've implemented `node. splitBlockAtRange` (which require the transform of method `splitNode`)

### Problems remaining:

- The code for `splitBlockAtRange` is duplicated between the transform and the node method, I'm not sure how to fix this.

### DIFF
--- a/src/models/node.js
+++ b/src/models/node.js
@@ -459,6 +459,85 @@ const Node = {
   },
 
   /**
+   * Split a node by `path` at `offset`.
+   *
+   * @param {String} path
+   * @param {Number} offset
+   * @return {Node}
+   */
+
+   splitNode(path, offset) {
+       let base = this
+       let node = base.assertPath(path)
+       let parent = base.getParent(node)
+       const isParent = base == parent
+       const index = parent.nodes.indexOf(node)
+
+       let child = node
+       let one
+       let two
+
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Hey @SamyPesse this looks awesome! Creating a `splitNode` method on `Node` is exactly what I would have done. I think in a later cleanup we could actually remove `splitBlockAtRange` hopefullly altogether and have the `getFragment` code use just `splitNode`, but that's not necessary right now to get it working.

Thank you for this!

PR #338, opened 2016-09-19
Focus and blur should not clear marks on selection

Soreine:
In our editor, we have buttons to toggle style, and to ensure that clicking on it does not remove focus, we `.focus()` after toggling marks. But the [current implementation](https://github.com/ianstormtaylor/slate/blob/master/src/transforms/operations.js#L374-L378) of the `setSelectionOperation` clear the marks on focus/blur. So our style buttons do not work on a collapsed selection.

I see two options:

1. Implement a different `focus` transform (or modify it) to not clear the marks.
2. Adapt the condition in `setSelectionOperation`

This PR opts for 2.

### DIFF
--- a/src/transforms/operations.js
+++ b/src/transforms/operations.js
@@ -371,9 +371,18 @@ export function setSelectionOperation(transform, properties) {
     prevProps[k] = selection[k]
   }
 
-  // If the current selection has marks, and the new selection doesn't change
-  // them in some way, they are old and should be removed.
-  if (selection.marks && properties.marks == selection.marks) {
+  // If the selection moves, clear any marks, unless the new selection
+  // does change the marks in some way
+  const moved = [
+      'anchorKey',
+      'anchorOffset',
+      'focusKey',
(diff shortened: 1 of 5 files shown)


ianstormtaylor:
@Soreine good use case, seems valid to me. Question for you, if you call `e.preventDefault()` inside the toggle button's `onClick` handler, won't that prevent the selection from being unfocused? I think that's what I had been doing to prevent that from happening.

Soreine:
Yes, I tested and you can prevent the blur during `onMouseDown`.
```js
    onMouseDown(e) {
        e.preventDefault();
    }
```

In our app we have a lot of places that would need to be modified to become aware of that. For example we can toggle bold with a button, from a menu, with a shortcut... In the end, they are all tied to the same action: toggle bold. So we found it simpler to tie the focus logic to the action itself, rather than to every piece of UI that can trigger it :)

ianstormtaylor:
Nice, that makes sense. Sounds good to me!

PR #339, opened 2016-09-20
Fix #234: fix typing text near inline void nodes

### LINKED ISSUE #234: Single node inline void in a block prevent edition of text
When a block (for example a `paragraph`) contains a single inline void node, it's not possible to insert text before or after.
It's working as expected, if some text nodes are already existing.

JSFiddle to reproduce the issue: https://jsfiddle.net/rhmkxsd5/

SamyPesse:
This PR fixes #234, it extends the normalization to add empty text nodes around void nodes.

### TODOs

- [x] POC implementation
- [ ] Fix and adapt tests
- [ ] Optmized implementation
- [x] New example with inline void nodes: emojis

### Optional

- [ ] Adapt serialize to remove the empty text nodes from the result (with option `terse`)


SamyPesse:
@ianstormtaylor I'd be awesome if you can check out the `Emojis` example that I've added, the edition is "working" but still buggy (removing an emoji, crash the editor; and empty text nodes are hard to navigate).

I'm still working on tests and improving the implementation of the normalization.

ianstormtaylor:
Hey @SamyPesse thanks for working on this! I love the example toolbar 😄 

Just pulled and checked it out. It's making me question whether going the route of "normalizing by adding empty text nodes around single inline nodes" is actually a good approach. 

The reason I say that is because while messing with the example, I remembered that browsers don't really handle empty text nodes very well. We get it to work for empty blocks because we swap it out for a `<br>` tag, but with inline nodes this isn't an option.

This is making me wonder whether we might need to completely encapsulate the extra weird inline void handling logic inside the `<Void>` component instead. Where it could render extra spans that have `&nbsp;` inside of them maybe, so that they actually contain "content" and the browser will navigate into them with the keyboard. And then any typing that happens inside those spans would be converted into `.insertText()` transforms on the state.

I think that might have to be the way to solve it instead, since I'm not sure trying to represent empty inline text nodes in the DOM and have the native browser's `contenteditable` editing handle it will work 😢 

What do you think?

(Either way, this emoji example is a great test case for getting the behavior right, so we should keep it regardless!)

ianstormtaylor:
Hmm, might be wrong on that actually.

Since selection coercion will always kick the cursor into the void inline node itself. Might be best to stick with the normalization tack and see if there are any browser hacks we can do or something. Not sure...

Hard problem.

SamyPesse:
@ianstormtaylor Yes, this is an hard problem, it's the main issue blocking us from releasing a slate-powered version of our editor.

I have to go to sleep, but I'll keep investigate solutions tomorrow.

It may be interesting to investigate how Draft-js is dealing with such an issue (it may be a good source of inspiration).

ianstormtaylor:
Yeah that's a good call. Do you know if Draft/Prosemirror currently solve this at all? or do they not have this issue for some reason because of the differences in structure?

SamyPesse:
Both Prosemirror and DraftJS handle it correctly, but I was not able to find any code directly related to this issue.

In the case of an inline image in a paragraph, Prosemirror is keeping a dom with the strick minimum ( no span before the image, etc):

screen shot 2016-09-22 at 00 30 01

Draft is also keeping a dom almost empty ([tested with the emoji plugin, and an empty emoji in a paragraph](https://www.draft-js-plugins.com/plugin/emoji)) (no spans before or after the emoji):

screen shot 2016-09-22 at 00 33 49

----

It really is a frustrating issue 😟  , I think void inline nodes are important for a lot of use cases (inline images, emoji, immutable mentions, preview, etc), I'll keep investigate.

ianstormtaylor:
Agreed it's definitely important. One thing to investigate is what is it about their data model (either in the DOM itself, or in how the selections are described) that makes it work. It might be two things, but not totally sure, for example:

- In Draft, there is no concept of "inline elements", so the inline entities just work as regular text, so it makes it simpler to handle the inline void case.
- In Prosemirror, the selection data model is different and might have a way to distinguish between positions on either side of a "inline node's barrier".

But based on Draft's emoji example working it definitely seems like there should be a DOM solution that works, so it's probably something we need to figure out at the data model level.

Just guesses off the top of my head.

ianstormtaylor:
Looking into this more, looks like we'll need to also add `props.parent` to nodes, to figure out when to render `<br>` inside spans of text. If inside of a block node, yes, otherwise no. I've got that change working locally but need to push it up to your branch still.

SamyPesse:
I've added you to the fork, but the easier solution will probably to start over with a new branch and I can refactor this PR to only be about the Emojis example.

ianstormtaylor:
Another update! 

Looking into the Draft emoji example to see how they've got it working, it seems like they aren't using inline `contenteditable={false}` nodes at all, they're just using emoji unicode characters in regular, editable text. Which is also easily possible with Slate (and works in any example really, although there might be one or two unicode length bugs to find).

There is a bug in Webkit (and I can't quite figure out if it's still in Chrome or not) that prevents selections from being placed inside empty elements:

- https://bugs.webkit.org/show_bug.cgi?id=15256
- https://bugs.webkit.org/show_bug.cgi?id=23189

So if we want to render empty text nodes, we'll need to solve that somehow. It sounds like zero-width spaces can solve it, but then we need to be careful about removing them when the copy/pastes (and potentially in other cases?)

I've added two commits:

- One that fixes the unnecessary `<br />` being rendered in empty text nodes.
- One that adds zero-width spaces to allow the cursor to get on either side of inline voids!

Working on copy-paste handling now to make sure that the zero-width spaces don't leave the editor when you try to copy text out.

@SamyPesse let me know how you find that behavior. I believe there's still a bug when backspacing, since it's probably trying to delete the empty text node, and the selection is no longer being set correctly afterwards.

Soreine:
@ianstormtaylor I have tested the Emojis example at a78caf030c241e138acf1c9ad9a0b218e45e3eb1 . I think this is the behavior I would expect from void nodes 👍 Good work @SamyPesse and @ianstormtaylor

The copy-paste looks OK to me :D I have not encountered the bug with backspacing you mentioned.
However, I have noticed a few others. I'll list them here, if it can help. I am using Chrome `Version 53.0.2785.116 (64-bit)`

* From the anywhere on the last line, <kbd>←</kbd> brings directly to the right of 🍔
* Focus 😃 and backspace. It is removed. Then undo. It is restored, but at the wrong place, right after the `!`
* From the first line, <kbd>↓</kbd> focuses 😃 instead of moving to 🍔 
* When focusing 😃, <kbd>↓</kbd> brings back at the beginning of the line, instead of next line

Cheers 🍻

ianstormtaylor:
Thanks @Soreine! That was super helpful.

I've just pushed a fix for the strange selection behaviors that actually works cross-browser (Chrome, Safari, Firefox) after investigating what the heck was going on 😮 

The undo bug where it puts it in the wrong place is still there, but I have a feeling this is fixed on `master`, or already exists in `master`, so it's at least better than what we have, so I'm going to merge!

The undo bugs that we are seeing in a few different places are caused by `.normalize()` calls. Since the extra calls to `document.normalize()` that are leftover in the transform codebase actually edit the state without using operations to do it, when reversing the old operations from the history sometimes they don't match up. 

The fix to these is that we need to remove all `document.normalize() && node.normalize()` calls and replace them with their operation equivalents. For example, in `removeNodeByKey` instead of calling `document.normalize()` it should figure out the exact invalid states that could result, and handle those specifically with more operations. I'll open a new issue for this!

Thanks @SamyPesse for kickstarting this PR!

Issue #340, opened 2016-09-20
wrapBlockAtRange() logic seems slightly off


tyler-johnson:
It seems to me that the way the `wrapBlockAtRange()` transform works is a little off of what would be expected, however I might be totally wrong.

The logic is currently attempting to find all the sibling block nodes in the selection that are closest to root and wraps those. The basis of this logic is what is expected, but I believe there is a little bug that stems from the use of `getBlocksAtRange()` and `getDepth()`.

Consider the following document that has two `paragraph` nodes inside two `list-item` nodes:

```yaml
document:
  type: document
  nodes:
    - type: list-item
      nodes:
        - type: paragraph
    - type: list-item
      nodes:
        - type: paragraph
```

I want to wrap both `list-item` nodes in a `bulleted-list` node, so I select everything in the document and run `transform.wrapBlock('bulleted-list')`. What I am expecting to get back is this:

```yaml
document:
  type: document
  nodes:
    - type: bulleted-list
      nodes:
        - type: list-item
          nodes:
            - type: paragraph
        - type: list-item
          nodes:
            - type: paragraph
```

However what I end up getting is this:

```yaml
document:
  type: document
  nodes:
    - type: list-item
      nodes:
        - type: bulleted-list
          nodes:
            - type: paragraph
            - type: paragraph
    - type: list-item
```

This is happening because `getBlocksAtRange()` return the bottom-most nodes and the use of `getDepth()` to determine the best siblings. Since both of the `paragraph` nodes in the example are at the same depth, the logic treats them as siblings, even though they are not siblings.

I believe the correct logic would look for the closest shared parent of the bottom-most nodes and wrap all the children nodes to that parent that fall within the selection.

ianstormtaylor:
Wow, good find! The proper result should definitely be your expected YAML. If you're down to look into this, I'd love a PR with the fix! If not, a PR with a failing test case would be awesome too! Thanks man

tyler-johnson:
fixed with #343

Issue #320, opened 2016-09-14
how to unwrap a block by key?


Soreine:
Ideally, I would like to make a transform that unwrap a block by key.
My use case is to unwrap items in a list that have sublists inside them. If my understanding is correct, `unwrapBlock` and `unwrapBlockAtRange` both unwrap all blocks at range that match a type, or certain data. My problem is to unwrap just that parent list, not every children list too.
Would it be a solution to add a test on a `key` property [here](https://github.com/ianstormtaylor/slate/blob/8fc3631328481b3c56c48a56e1a9bad2eb24bd5a/lib/transforms/at-range.js#L599-L602) ? So that we can just unwrap a single block by key ?

---

On a side note, I'll explain my related issue. I am [struggling](https://github.com/GitbookIO/slate-edit-list/pull/16) to implement something that unwrap a block by two levels. Something that would extract a a list item from a list, and then unwrap that list item's children.

It would go from that (short notation):

``` yaml
- type: list
  nodes:
    - type: item
      nodes:
        - text: 1
    - type: item # Item with sublist to unwrap
      nodes:
        - text: 2
        - type: list
          nodes:
            - type: item
              nodes:
                - text: 2.1
            - type: item
              nodes:
                - text: 2.2
    - type: item
      nodes:
        - text: 3
```

To this:

``` yaml
- type: list
  nodes:
    - type: item
      nodes:
        - text: 1
# Unwrapped item's children
- text: 2
- type: list
  nodes:
    - type: item
      nodes:
        - text: 2.1
    - type: item
      nodes:
        - text: 2.2
# Notice that the root list was split
- type: list
  nodes:
    - type: item
      nodes:
        - text: 3
```

I think I need to make it in 2 steps :
1.  Extract the list item from its parent list. If the parent list becomes empty, delete it, else, just split it in two lists, and put the list item in between.
2.  Replace the extracted list item with its children.

I am not sure of what "unwrapping" means exactly. But I think we could say that 1. is unwrapping the list item from it's parent list, and 2. is unwrapping the list item's children from the list item.

I would greatly appreciate if anyone could confirm the meaning of "unwrapping", or provide advice on how to implement this transform neatly.

Thank you all

Soreine:
@tyler-johnson I noticed you had problems with wrapping/unwrapping too. I saw you guys improved the wrap transforms, which was awesome. I still have not found a solution to my unwrapping problem. Is it an issue for you too? Have you found a workaround since then?

[...]

PR #343, opened 2016-09-20
improvements to wrapBlockAtRange()

tyler-johnson:
Fix for #340.

This seems to fix the issue for me. Hopefully the way I implemented it is okay. There are a few less-than-performant spots like the nested `getClosest()` but I figure most nodes won't have more than a few parents.

### DIFF
--- a/src/transforms/at-range.js
+++ b/src/transforms/at-range.js
@@ -721,24 +721,53 @@ export function wrapBlockAtRange(transform, range, block) {
 
   const { state } = transform
   const { document } = state
+
   const blocks = document.getBlocksAtRange(range)
-  const depth = blocks.map(n => document.getDepth(n)).min()
+  const firstblock = blocks.first()
+  const lastblock = blocks.last()
+  let parent, siblings, index
 
-  const siblings = blocks
-    .map((node) => {
-      const d = document.getDepth(node)
-      if (d == depth) return node
-      return document.getClosest(node, p => document.getDepth(p) == depth)
+  // if there is only one block in the selection then we know the parent and siblings
+  if (blocks.length === 1) {
+    parent = document.getParent(firstblock)
+    siblings = blocks
+  }
+
+  // determine closest shared parent to all blocks in selection
+  else {
+    parent = document.getClosest(firstblock, p1 => {
+      return !!document.getClosest(lastblock, p2 => p1 == p2)
     })
-    .toOrderedSet()
-    .toList()
+  }
 
-  const first = siblings.first()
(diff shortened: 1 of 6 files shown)


tyler-johnson:
So upon using this with my code, I noticed an edge case that revealed an error in my implementation. This is wrapping the bottom-most shared parent, however now I realize that may be incorrect. Should all blocks, all the way up to the document, be wrapped after this command? If so that vastly simplifies this logic.

ianstormtaylor:
@tyler-johnson awesome man!

As for the second question, wrapping the bottom-most shared parent is the expected behavior here I believe, yup.

tyler-johnson:
So I actually think the expected logic here is the top-most parent (i.e. the document). The issue with the bottom-most parent is that the wrap is "sandwiched" between two block nodes which is unexpected at levels deeper than 1.

Consider this document:

```yaml
nodes:
  - types: quote
    nodes:
      - types: paragraph
```

If we wrap it using the bottom-most parent, this document is produced:

```yaml
nodes:
  - types: quote
    nodes:
      - types: header1
        nodes:
          - types: paragraph
```

Where as if we wrap the top-most parent, this document is produced:

```yaml
nodes:
  - types: header1
    nodes:
      - types: quote
        nodes:
          - types: paragraph
```

I think what it comes down to is consistency and clarity. When multiple blocks are selected it won't always be known _where_ the block is going to be wrapped with bottom-most. Top-most on the other hand is very consistent.

Additionally, the current way wrapBlock is documented makes it seem that it wraps everything in the selection, all the way to the outermost node. 

Perhaps there needs to be two different methods for each? This definitely relates to the use case behind #320.

tyler-johnson:
I have committed both top and bottom-most implementations. Let me know which one you desire to keep.

ianstormtaylor:
Wrapping at the bottom-most block allows for you to "not care" about parent nesting when running wrap commands, which is useful for cases where an entire section is wrapped in a `blockquote` for instance. With bottom-most nesting you'd still be able to wrap in a list for example inside a `blockquote`. Otherwise things tend to get weird I think where you'd have to specify the depth.

In general I've been following an approach of: block actions act on the bottom-most block, and inline actions act on the top-most inline. This seems to handle things closest to expected.

So for now let's keep bottom-most only.

tyler-johnson:
okay sounds good, I've reverted those changes! I guess that means this new `wrapBlock()` will probably work for #320 too.

ianstormtaylor:
Boom, awesome, thanks man!!

Issue #344, opened 2016-09-22
Wrong behavior when press backspace from paragraph below the table


aunsuwijak:
![tyzl6sffqm](https://cloud.githubusercontent.com/assets/5433223/18736649/442127c0-80b3-11e6-9820-2f48e069f30e.gif)

PR #345, opened 2016-09-22
Prevent backspace to move selection into table-cell issue #344

aunsuwijak:
https://github.com/ianstormtaylor/slate/issues/344


ianstormtaylor:
Awesome, thanks @aunsuwijak !

Issue #357, opened 2016-09-25
Backspace the image in first line failed 


kiliwalk:
Assume an image in the first line of document, focus it, then tap `Backspace` or `Delete` key, image is not removed but an error occurs: `node.js:107 Uncaught Error: Could not find a descendant node with key "759l".`

After debug this, I found a solution to fix the backspace problem.
Replace the [line in at-current-range.js#L109](https://github.com/ianstormtaylor/slate/blob/master/src/transforms/at-current-range.js#L109): 

```
    after = selection
```

with 

```
    let parentNode = document.getClosestBlock(startNode)
    if(parentNode && parentNode.isVoid) { // first node is like image etc
      after = selection.collapseToStartOf(document.getNextBlock(startNode))
    } else {
      after = selection
    }
```

I'd like to see a more accurate method to fix these problems.
Thanks for this great framework!

arekkas:
This yields now the following warning: `Warning: Selection was invalid and reset to start of the document`

ianstormtaylor:
This should be fixed now!

PR #360, opened 2016-09-28
Fix backspace first void block and delete last void block (#357)

kiliwalk:
- Fix backspace first void block, see #357
- Fix delete last void block in empty document

### DIFF
--- a/src/transforms/at-current-range.js
+++ b/src/transforms/at-current-range.js
@@ -1,5 +1,6 @@
 
 import Normalize from '../utils/normalize'
+import Block from '../models/block'
 
 /**
  * Add a `mark` to the characters in the current selection.
@@ -106,7 +107,16 @@ export function deleteBackward(transform, n = 1) {
   }
 
   else if (selection.isAtStartOf(document)) {
-    after = selection
+    if(document.hasVoidParent(startNode)) { // first node is void, like image etc
+      let nextBlock = document.getNextBlock(startNode)
+      if(!nextBlock) { // document is empty, so create an empty paragraph as placeholder
+        nextBlock = Block.create({type: 'paragraph'})
+        transform = transform.insertBlock(nextBlock)
+      }
+      after = selection.collapseToStartOf(nextBlock)
+    } else {
+      after = selection
+    }
   }
 
   else if (selection.isAtStartOf(startNode)) {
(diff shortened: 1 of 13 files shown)


ianstormtaylor:
Hey @kiliwalk thanks for submitting this! Can you add a test that ensures that we don't accidentally break this functionality in the future? Thanks!

skogsmaskin:
I think this also need patching for ``deleteForward``. It has the same symptoms - if it is a void-node at the end of the document.

kiliwalk:
@ianstormtaylor That's ok, I will append tests later.
@skogsmaskin Already included in this fix

skogsmaskin:
@kiliwalk - i have to applied your patch (thank you btw), it works well for void-nodes in the beginning of the document, but it seems like we still have issues for void-nodes in the end of the document. Also void-nodes after each other in the middle of the document.

kiliwalk:
@skogsmaskin Could you give a detail case? As I tested, there is no problem with delete-forwad last void block in document

skogsmaskin:
@kiliwalk try inserting three void-nodes (only) in the document. Focus last or second node and press backspace. I get ``node.js:107 Uncaught Error: Could not find a descendant node with key "nm02".``.

Deleting the first void-node always works fine.

kiliwalk:
@skogsmaskin Oh yeah, you are right, but it's delete-backward issue. I'd like to try to fix these together

kiliwalk:
The [fix backspace the void block continuously](https://github.com/ianstormtaylor/slate/pull/360/commits/1b5f8e4835978c1dca62617d16ddc10d2a1154fe) commit tries to fix this problem:
1. Drop two images to the empty editor, then the document only has three image blocks, no others like paragraph
2. Now current selection is on the last image, and the offset is 1(not 0)
3. Backspace will emit error
4. Click the last image, now the selection offset is 0
5. Backspace successfully
6. Now selection is on the first image, and the offset is 1
7. Since backspace will fail on offset=1, so click then backspace, but error occurs since there is nothing in document. Accutually, we should create an empty paragraph.

With this commit, backspace will be ok when offset=1, and will insert an empty paragraph when document is empty

skogsmaskin:
It works fine now, but one thing - the use of paragraph node type. As @ianstormtaylor commented on my pull request (https://github.com/ianstormtaylor/slate/pull/371#issuecomment-252321491), should not be acknowledged by slate core.

Not sure how fix this otherwise though.

ianstormtaylor:
Ah yup, as @skogsmaskin mentioned, we should avoid anything that requires Slate being opinionated about a specific `node.type` value. If that gets fixed this sounds good to go!

kiliwalk:
@ianstormtaylor So is there any method like `document.createEmpty()` to insert a empty node? I have no idea how to replace specifying paragraph type, please give some advices  if possible

ianstormtaylor:
@kiliwalk good question. I think maybe the error that is thrown when the document is empty should be fixed, to make that a valid state, and then when deleting it can just remove the node and not insert anything new. What is that error by the way?

kiliwalk:
The error is like `Uncaught Error: Could not find a descendant node with the key "xxxx"`, and thrown at `moveTo` method since deleting the only one void node there is no other nodes in the document

ianstormtaylor:
Hey @kiliwalk sorry for the long delay here. The error when deleting the first void block in a document has been fixed now thanks to some other refactoring.

There is still the issue that you cannot delete the last void node in a document if it's the only node left, and I think the core of Slate alone can't solve this problem for you, other than leaving you with an empty document which isn't helpful. Instead, if you're dealing with this situation, I'd recommend adding your own custom schema rule that enforces that the void nodes in question are always followed by another node.

I'm going to close this pull request since we can't solve it this way without forcing Slate to care about the nodes in your schema. Thank you for the help you, and for prompting the discussion!

erquhart:
@ianstormtaylor it seems that rules are only run for a node when the node itself is changed - if I delete all of the nodes around the void node, the rule is never triggered. Is this expected?

erquhart:
Update for anyone else who stumbles here: the document itself can be validated, and that happens on every change.

PR #361, opened 2016-09-29
Remove old `snapshot` parameter for Transform.apply

Soreine:
I simply updated any reference to `snapshot` that I could find and replaced them with the new `save` parameter. The documentation still refers to the concept of snapshot though for clarity. Just the name of the parameter changed.

This also fixes a bug: that normalization was pushing a to the undo/redo history.


ianstormtaylor:
Ah, awesome. Thank you @Soreine !

Issue #180, opened 2016-07-27
optimize `Node` methods


ianstormtaylor:
**If anyone wants to help, these are great, self-contained places to start!**

A big area of performance low-hanging fruit is that many of the `Node` methods are not very optimized (I've been focused on getting the logic right, and worrying about optimization later). When working with large documents this is very often what slows things down.

Examples:
- [ ] Many methods do `assertDescendant()` at the beginning to ensure that the node in question exists, which traverses the whole tree. They might be able to be rewritten to throw only after not finding the child.
- [ ] New methods `getFirstText()`and `getLastText()` would save cycles instead of currently using `getTexts().first()` or `getTexts().last()` in lots of places, because they will avoid traversing the whole document tree.
- [ ] Many places use immutable objects for a series of updates, when using `asMutable()` and then `asImmutable()` might result in quicker operations.
- [ ] Many places use `mapDescendants` when they really could be aborting much sooner after having found the node in question.

Some of the hottest code paths to optimize first are:
- `getTexts`
- `getBlocks`
- `getDescendant`
- `getParent`
- `getClosest`
- `updateDescendant`
- `removeDescendant`
- `mapDescendants`

As I find more I'll add them to this list. 

_Note: some of these might be tricky, because we're currently memoizing return values on immutable retrieval methods, which means that the second `getText()` call might be faster than two methods that have different traversal logic._

Soreine:
I believe having a way to measure the execution time of those functions is key to solve performance problems:

1. Measuring actually shows which operations are slow and need work first.
2. Measuring allows to validate a performance improvement after optimizing.
3. Measuring helps preventing regression when optimizing other parts (thinking of memoizing benefits).

Now on how satisfy each point:

1. Profiling would be the best way to make educated guess about performance bottlenecks (which part needs to be optimized in priority). But it's hard to do and not easily reproducible. Alternatively, simply measuring common operations or functions could highlight the most obvious places of interest.
2. Adding performance scripts to measure the execution time, on average, of common operations or functions, would allow to compare two implementations.
3. Adding performance tests (requirements on average execution time) could do the trick. But it's hard to do since it's hardware dependent.

Now on how necessary they are:

1. This could be necessary, depending on how confident you are on your analysis @ianstormtaylor. 
[...]

PR #364, opened 2016-10-04
Node methods optimizations

Soreine:
Working on https://github.com/ianstormtaylor/slate/issues/180

I open this PR as a work in progress to gather reviews as I work on it.
I cannot measure the impact of the changes, so I'm not sure each point actually improves performances (tricky case of memoization).

So far:

- Optimize `Node.getParents` to return as soon as the child is found
- Add `Node.getAncestors` method (used in `Node.getClosest`)
- Optimize `Node.getClosest` to not use successive `getParent`, by using `Node.getAncestors` instead. I don't know if there is a benefit to this, since `getParent` is recursive and memoized.
- Remove extra use of `assertDescendant` (but keep throwing when not found) in the following implementations:
  - `Node.getPath`
  - `Node.updateDescendant`
  - `Node.removeDescendant`

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "http-server": "^0.9.0",
     "is-image": "^1.0.1",
     "is-url": "^1.2.2",
+    "jsdom": "9.8.0",
     "jsdom": "9.6.0",
     "jsdom-global": "2.1.0",
     "microtime": "2.1.1",
--- a/src/components/void.js
+++ b/src/components/void.js
@@ -130,7 +130,7 @@ class Void extends React.Component {
 
   renderLeaf = () => {
     const { node, schema, state } = this.props
-    const child = node.getTexts().first()
+    const child = node.getFirstText()
     const ranges = child.getRanges()
     const text = ''
     const marks = Mark.createSet()
--- a/src/models/node.js
+++ b/src/models/node.js
@@ -105,19 +105,27 @@ const Node = {
   },
 
   /**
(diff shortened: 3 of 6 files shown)


ianstormtaylor:
@Soreine this is awesome!

Agreed that it's hard without a way to measure the impacts. Maybe there's a way we can get a benchmark-like test suite going so that we can know if we're making things more performant? I don't know much about benchmarking unfortunately.

SamyPesse:
@ianstormtaylor Yes, we were talking with @Soreine today about making a PR to add benchmarks to Slate.
I think we can have a simple solution that prints the result and compare between versions using Travis build's logs.

We'll probably open another PR for the benchmarks. This PR still contains minors but required optimizations (ex: `X to return as soon as the child is found`, etc) that do not require benchmarks.

Soreine:
@ianstormtaylor I think you can review/merge this PR. The changes are self-contained. There are still things to do, but I can always open a new PR.

Soreine:
- Fixed `Node.findDescendant` which would not return nodes deeper than the first level.
- Implemented `Node.getFirstText` and `Node.getLastText` and used it where fit. I think even with memoization, these implementations still compete with a memoized call to `Node.getTexts`. It's still a constant complexity.

Soreine:
Note to self: When we merge it, review if there are new places where we could use `node.getFirstText`

ianstormtaylor:
Sweet, @Soreine if you can rebase against master I can get this merged!

Soreine:
Done.

PR #365, opened 2016-10-05
fix #342, Uncaught Error: Invalid `key` argument when `removeNodeByKey`

### LINKED ISSUE #342: operation to replace document entirely?
My editor is partially-collaborative, in the sense that multiple people can edit documents at the same time, but they can't see what the other people are doing. This mostly works in practice since we have a limited number of people working in the same document and even fewer working in the same Slate editor.

When changes happen, the new complete document (stored in HTML) is distributed to other clients and the other clients silently swap it out with the old state. The issue I am facing now is that the undo/redo history is removed since it is a brand new state. In the pre-0.14 transform I could simply save a new version of the document into the history, however with this new operation-based transform that is not so easy.

So is it possible to get an operation that totally overwrites the state's document with a new document?

chilijung:
I found out causing error 

```
Uncaught Error: Invalid `key` argument! It must be either a block, an inline, a text, or a string. You passed: "null".
```
which mentioned in issue #342 is that when the node is removed `this.selection.anchorKey` and `this.selection.focusKey` are both `null` and are asked to get `getDescendant(key)`!

### DIFF
--- a/src/models/state.js
+++ b/src/models/state.js
@@ -5,7 +5,7 @@ import Mark from './mark'
 import Selection from './selection'
 import Transform from './transform'
 import uid from '../utils/uid'
-import { Record, Set, Stack } from 'immutable'
+import { Record, Set, Stack, List } from 'immutable'
 
 /**
  * History.
@@ -361,9 +361,9 @@ class State extends new Record(DEFAULTS) {
    */
 
   get marks() {
-    if (this.selection.anchorKey || this.selection.focusKey) {
-      return this.selection.marks || this.document.getMarksAtRange(this.selection)
-    }
+    return this.selection.isUnset
+      ? new List()
+      : this.selection.marks || this.document.getMarksAtRange(this.selection)
   }
 
   /**
@@ -373,9 +373,9 @@ class State extends new Record(DEFAULTS) {
    */
 
   get blocks() {
-    if (this.selection.anchorKey || this.selection.focusKey) {
-      return this.document.getBlocksAtRange(this.selection)
-    }
+    return this.selection.isUnset
+      ? new List()
+      : this.document.getBlocksAtRange(this.selection)
   }
 
   /**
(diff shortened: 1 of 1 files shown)


chilijung:
not sure if it should return an empty array?

ianstormtaylor:
Ah, interesting case! Good find. I think the code could be simplified a little bit, and I think those methods should probably always return the same type of object, but in the case of the selection being set to `null`, they can return an empty set/list.

And then we can also make use of the `selection.isUnset` property.

So for example `get blocks()` might be implemented as:

```js
get blocks() {
  return this.selection.isUnset
    ? new List()
    : this.document.getBlocksAtRange(this.selection)
}
```

@chilijung would you be down to make those changes?

chilijung:
@ianstormtaylor fixed!

src/models/state.js
```diff
@@ -361,9 +361,9 @@ class State extends new Record(DEFAULTS) {
    */
 
   get marks() {
-    if (this.selection.anchorKey || this.selection.focusKey) {
-      return this.selection.marks || this.document.getMarksAtRange(this.selection)
-    }
+    return this.selection.isUnset
+      ? new List()
```
ianstormtaylor: This one should be a `new Set()` I believe.

src/models/state.js
```diff
@@ -385,9 +385,9 @@ class State extends new Record(DEFAULTS) {
    */
 
   get fragment() {
-    if (this.selection.anchorKey || this.selection.focusKey) {
-      return this.document.getFragmentAtRange(this.selection)
-    }
+    return this.selection.isUnset
+      ? new List()
```
ianstormtaylor: This one should be a `Document.create()` to get a new document.

chilijung:
@ianstormtaylor Thanks for review the code! fixed in the latest commit.

ianstormtaylor:
Awesome, thanks @chilijung!

PR #368, opened 2016-10-05
Add benchmarks

Soreine:
Following-up on some suggestions (https://github.com/ianstormtaylor/slate/issues/180 and https://github.com/ianstormtaylor/slate/pull/364)

# Goal

This PR aims to add benchmarks to Slate. The objective is to provide a reliable way to compare the efficiency of different Slate's implementations.

# To do

- [x] Add a script to run benchmarks, and a structure to easily define new benchmarks.

  I used [BenchmarkJS](https://benchmarkjs.com/) since it seems popular and precise (used by ImmutableJS). I made sure that every execution is made with a brand new `State` instance to not measure memoized functions. I faced a few issues with the BenchmarkJS though. I did not manage to use it along with babel. And its internal magic plays very bad with scopes, so I had to do ugly workarounds using globals. In the end, it might not matter that the script itself is ugly, but if anyone want to have a try at finding a cleaner way, I can give a hand.

- [x] Run benchmarks in Travis, to keep a written track of the progress

- [x] Write relevant benchmarks. 

    For example, to benchmark `normalizeDocument`, one could write a couple of benchmarks, executing the `normalizeDocument` transform on states of increasing complexity.

- [x] Implement exporting the results of the benchmarks to a JSON file, so that the script can later pick it up and compare it with new results. One could checkout `master`, run benchmarks once, keep the JSON output, checkout new code, and have the script run the new benchmarks and do the comparison.

- [x] Add README explaining how to run and interpret the benchmarks.


SamyPesse:
👍 

Benchmarks for normalize and most calls should be applied to different set of documents: small doc, normal and large; it'll let us see what is impacted by the nesting and size of the doc.

And `npm run benchmarks` should be added to the `.travis.yml` file

ianstormtaylor:
@Soreine you are awesome! Thank you!

Soreine:
I just implemented the serialization of the results to a JSON file. Now you can checkout an implementation, `npm run perf:save` (store results in `./perf/reference.json`), checkout another implementation, and `npm run perf` to see results compared with `./perf/reference.json`.

I have hit out of memory issues though, when benchmarking fast operations. For fast operations, BenchmarkJS has to do more runs to get accurate results. Hence, the test setup takes a lot of memory, because we store beforehand a unique `Slate.State` for each run (to avoid memoization). So we might need to find a smarter way to bypass memoization.

SamyPesse:
@Soreine BenchmarkJS works fine on Node v4, the errors you are getting is ES6 not supported on Node v4. You should use `babel-node` or a pre-compilation to babel and it'll work fine.

Soreine:
@SamyPesse Oh yeah right. Now I remember that I had tried to use babel, but after noticing errors I did not dig that much, since it was not a priority. I'll give it another try.

Soreine:
Just pasting some early comparison results. 

# Comparing #364 with master

Current is #364 and Reference is master. We can see that `normalize` itself has not changed, but the basic `delete` and `splitBlock` transforms are between 50% and 200% faster.

```
Benchmarks

   delete-backward-deep
     Current:	 132 ops/sec (70 runs sampled)
     Reference:	 84.08 ops/sec (70 runs sampled)
     comparison: Faster
     diff: +57.56%
     rme: ±5.85%

   delete-backward-normal
     Current:	 212 ops/sec (81 runs sampled)
     Reference:	 164 ops/sec (80 runs sampled)
     comparison: Faster
     diff: +28.98%
     rme: ±3.16%

   delete-backward-wide
     Current:	 120 ops/sec (54 runs sampled)
     Reference:	 59.16 ops/sec (58 runs sampled)
     comparison: Faster
     diff: +102.66%
     rme: ±4.06%

   normalize-document-deep
     Current:	 189 ops/sec (71 runs sampled)
     Reference:	 179 ops/sec (74 runs sampled)
     comparison: Indeterminate
     diff: +5.69%
     rme: ±4.32%

   normalize-document-normal
     Current:	 617 ops/sec (71 runs sampled)
     Reference:	 549 ops/sec (64 runs sampled)
     comparison: Faster
     diff: +12.52%
     rme: ±4.09%

   normalize-document-wide
     Current:	 118 ops/sec (53 runs sampled)
     Reference:	 107 ops/sec (54 runs sampled)
     comparison: Faster
     diff: +10.84%
     rme: ±6.67%

   split-block-deep
     Current:	 144 ops/sec (68 runs sampled)
     Reference:	 96.06 ops/sec (72 runs sampled)
     comparison: Faster
     diff: +49.41%
     rme: ±3.12%

   split-block-normal
     Current:	 358 ops/sec (72 runs sampled)
     Reference:	 191 ops/sec (81 runs sampled)
     comparison: Faster
     diff: +86.88%
     rme: ±2.79%

   split-block-wide
     Current:	 174 ops/sec (52 runs sampled)
     Reference:	 57.85 ops/sec (56 runs sampled)
     comparison: Faster
     diff: +201.38%
     rme: ±5.88%
```

# Comparing #354 with master

Current is #354 and Reference is master. We can notice slight improvement overall. The `split-block-normal` benchmark is shown as slower, but it's not significant enough to come to a conclusion.

```
Benchmarks

   delete-backward-deep
     Current:	 114 ops/sec (81 runs sampled)
     Reference:	 84.08 ops/sec (70 runs sampled)
     comparison: Faster
     diff: +35.22%
     rme: ±5.87%

   delete-backward-normal
     Current:	 208 ops/sec (84 runs sampled)
     Reference:	 164 ops/sec (80 runs sampled)
     comparison: Faster
     diff: +26.35%
     rme: ±3.23%

   delete-backward-wide
     Current:	 87.65 ops/sec (61 runs sampled)
     Reference:	 59.16 ops/sec (58 runs sampled)
     comparison: Faster
     diff: +48.17%
     rme: ±4.96%

   normalize-document-deep
     Current:	 275 ops/sec (70 runs sampled)
     Reference:	 179 ops/sec (74 runs sampled)
     comparison: Faster
     diff: +53.49%
     rme: ±4.47%

   normalize-document-normal
     Current:	 697 ops/sec (66 runs sampled)
     Reference:	 549 ops/sec (64 runs sampled)
     comparison: Faster
     diff: +27.08%
     rme: ±4.59%

   normalize-document-wide
     Current:	 149 ops/sec (58 runs sampled)
     Reference:	 107 ops/sec (54 runs sampled)
     comparison: Faster
     diff: +39.05%
     rme: ±5.99%

   split-block-deep
     Current:	 118 ops/sec (80 runs sampled)
     Reference:	 96.06 ops/sec (72 runs sampled)
     comparison: Faster
     diff: +23.02%
     rme: ±2.98%

   split-block-normal
     Current:	 173 ops/sec (77 runs sampled)
     Reference:	 191 ops/sec (81 runs sampled)
     comparison: Slower
     diff: -9.88%
     rme: ±3.46%

   split-block-wide
     Current:	 93.39 ops/sec (61 runs sampled)
     Reference:	 57.85 ops/sec (56 runs sampled)
     comparison: Faster
     diff: +61.42%
     rme: ±5.45%
```

ianstormtaylor:
Awesome @Soreine! Let me know when you think it's ready to merge and I can look deeper and do a review.

Is there a way to hook up these benchmarks so that we can see the effects of React rendering, since that seems to be one of the bigger cases of slowness?

Side note: I'd suggest to avoid time writing benchmarks for `normalizeDocument`, since that will be removed soon by my other PR. And we could instead focus on the more generic integration-test-style cases.

ianstormtaylor:
@Soreine I'd also be curious about getting benchmarks that don't try to bypass the memoization logic, for a more real-world use case, since that logic should ideally be saving lots of cycles.

Soreine:
🏄‍♀️ Ready for review.

I'll be happy to answer any question, or to add comments. Make sure to read `perf/Readme.md` first.

Here is my own review of the PR:

- We still need to add more "generic integration-test-style" benchmarks. @ianstormtaylor if you have suggestion about what to benchmark, I can write the benchmark for it.

- I tried to make the code as clean as a script can be.

- I managed to implement two React rendering benchmarks. They will at least serve as reference. But I have not pushed the concept very far (it does not always measure exactly the rendering, but some overhead too).

- The benchmarks are always given separate State instances, to avoid unwanted memoization between runs. It should be easy though to add an option for benchmark to disable that mechanism. (out of curiosity, I ran the benchmarks without it, and they were often 3 times faster).

- Initially, I was not using babel, so I was not using `import/export` statements. It can be changed. There's just a catch: we need to make sure `jsdom` is setup after we require BenchmarkJS, and before we require React (see comment in the code).

- BenchmarkJS does a good job to run the code and make relevant statistical analysis. But it does so in a very restricting manner (doing a lot of magic), which breaks scoping and such (that's why I had to use globals..). It was quite a hassle to use, but I don't know any comparable library in term of stats quality.

Soreine:
Fixed Travis checks. 😏

ianstormtaylor:
Thank you so much @Soreine this is amazing. 😄

PR #371, opened 2016-10-07
Insert empty paragraph where appropriate when hitting enter on void node

skogsmaskin:
Adds a empty paragraph in the bottom of the document if the last node is a void node.
This allows us to move the cursor below the node, and add more stuff.

### DIFF
--- a/src/plugins/core.js
+++ b/src/plugins/core.js
@@ -358,9 +358,16 @@ function Plugin(options = {}) {
 
     // For void blocks, we don't want to split. Instead we just move to the
     // start of the next text node if one exists.
+    // If it doesn't exist, we create an empty paragraph.
     if (startBlock && startBlock.isVoid) {
       const text = document.getNextText(startKey)
-      if (!text) return
+      if (!text) {
+        return state
+          .transform()
+          .collapseToEndOf(startBlock)
+          .insertBlock('paragraph')
+          .apply()
+      }
       return state
         .transform()
         .collapseToStartOf(text)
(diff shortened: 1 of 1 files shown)


skogsmaskin:
Implemented this the wrong place first. Now implemented in the core plugin's `onKeyDownEnter`function.

skogsmaskin:
Added more code to handle hitting enter on void nodes in the start of the document, between other nodes and in the end of the document.

ianstormtaylor:
Hey @skogsmaskin thanks for this!

I think this should actually live in a plugin (for now at least) because Slate shouldn't have a knowledge of the `'paragraph'` block type, since that would be part of the developer's schema choices. But it seems like a very useful case to have as a simple plugin that ensures that a certain block type is created when <kbd>enter</kbd> is pressed on a void node!

skogsmaskin:
@ianstormtaylor, thank you for your feedback. Yes, you are probably right, I will make it into a plugin instead.

skogsmaskin:
Ok, first stab at this plugin:
https://github.com/skogsmaskin/slate-insert-block-on-enter

The initialisation of the plugin (block properties) is not yet as I want, because there are some bugs in Slate when Block.create is used in combination with ``insertNodeByKey``.

If I do ``Block.create({type: 'paragraph'})``

``insertBlock`` inserts a block with one node (empty text node).
``insertNodeByKey`` inserts a block with empty nodes - causing side effects in the rendering of node.

So what I have to do for now is ``Raw.deserializeNode(blockInputProps)`` where `blockInputProps` is like ``{kind: 'block', type: 'paragraph', nodes: [{kind: 'text', text: '', ranges: []}]}``

PR #373, opened 2016-10-11
Node shouldComponentUpdate use Immutable.is

Soreine:
This tiny PR fixes the false positives for `Node.shouldComponentUpdate` when the new `node` prop was a different object, but its value did not actually change.

Instead of a reference comparison `newNode != oldNode`, we use `Immutable.is` that is optimized to compare properties values if needed. It cannot be worst in term of performance. 

However it highlights the fact that we sometimes return new Immutable objects without actually changing their values. We have noticed it in the given situation: Put cursor in the middle of a paragraph of the Rich Text example, hit Return. Then, all blocks in the document are re-rendered, and fail the equality check because `node.nodes` is a different, cloned `Immutable.List`.

### DIFF
--- a/src/components/node.js
+++ b/src/components/node.js
@@ -1,5 +1,6 @@
 
 import Base64 from '../serializers/base-64'
+import Immutable from 'immutable'
 import Debug from 'debug'
 import React from 'react'
 import ReactDOM from 'react-dom'
@@ -97,7 +98,7 @@ class Node extends React.Component {
     }
 
     // If the node has changed, update.
-    if (props.node != this.props.node) {
+    if (!Immutable.is(props.node, this.props.node)) {
       return true
     }
 


ianstormtaylor:
Hey @Soreine good catch! Do you know why we are getting new instances of the `Node` without any changes? Ideally that shouldn't be happening, it might be a normalization problem? If we could fix it at the source that would be best. Although I might be misunderstanding the issue.

SamyPesse:
@ianstormtaylor I think we should always use `Immutable.is` when comparing `Node` and other immutable data; 

Performances of `Immutable.is(valueA, valueB)` compared to `valueA === valueB` should be good  enough since the first test done by `Immutable.is` is [`valueA === valueB`](https://github.com/facebook/immutable-js/blob/master/src/is.js#L65), and it'll avoid potential false positives.

The source of the issue is coming from `Node.normalize` at line [1130](https://github.com/ianstormtaylor/slate/blob/master/src/models/node.js#L1130) (my fault :fearful:):

```js
// Create a record similar to Node with an inner list
var R = Immutable.Record({ list: Immutable.List() })

// Create an instance with a list
var a = R({ list: Immutable.List([1,2,3]) })

// Update the list by setting a new list identical
var b = a.set('list', Immutable.List([1,2,3]))

b === a
> false
```

Soreine:
Maybe it's not worth fixing this if we plan to remove `normalize` calls.

ianstormtaylor:
Nice, thanks for that info @SamyPesse! I also agree that it's not worth fixing @Soreine.

My only remaining thought to run by you two: 

I was thinking that `Immutable.is` should never need to be used, since the nodes shouldn't be mutated in most cases, and we want the fast equality checking, not the slower value-based checking. So my thought was that if we switch to using `Immutable.is`, we're kind of patching a problem that shouldn't occur and making it harder to catch later (since it won't trigger re-renders, just slower `Immutable.is` checks).

But that could also be wrong, and you could argue that we should always do `Immutable.is` because there are valid use cases where you'll end up creating an exact duplicate of the node, and it's going to be faster than re-rendering in the case that someone replaces a node with an exact duplicate. The only thing there is I'm not sure of what the cases are where duplicates will be valid… although that might not matter.

What do you think? Does it make it harder to catch the true bugs? Or is it actually something that the system should account for, and those bugs should be caught somewhere else?

Soreine:
I always ask myself that same question. It's a conflict between wanting the flaws in your code to be as obvious as possible when developing, and wanting them to be forgiving (more robust) in production.

It's not much easier to notice slow `Immutable.is` checks than to notice a few unnecessary re-renders, anyway. So I'd go for using `Immutable.is`.

What about having a different behavior in dev and production mode ?:
- Dev: Check strict equality, if it fails, also check `Immutable.is` equality and if both results differ, log a warning.
- Production: Just use `Immutable.is` all the time.

Soreine:
I close this PR. In the schema normalization PR we opted to keep a check in DEV mode only, for performance reason.

PR #383, opened 2016-10-13
Slow renderer causes double-click selection to break

arekkas:
At the moment, [onSelect](https://github.com/arekkas/slate/blob/e367326618ec62f64ee914e802f0fdf2a1d58886/src/components/content.js#L559-L566) is aborting if `this.tmp.isRendering` is true. This results in buggy behaviour if the rendering method is slow. For demonstration purposes, I created this PR. To get the buggy behaviour, uncomment [this line](https://github.com/arekkas/slate/blob/e367326618ec62f64ee914e802f0fdf2a1d58886/src/components/content.js#L563), re-build everything (libs and examples) and double-click on some word in the editing area.

Additionally, I created two screencasts which show the problem.

With `this.tmp.isRendering` check in `onSelect`:

Without `this.tmp.isRendering` check in `onSelect`:

I don't know what the reasoning behind this check was. If it is ok to remove it, I will clean up this PR and propose a fix.

### DIFF
--- a/examples/hovering-menu/index.js
+++ b/examples/hovering-menu/index.js
@@ -1,4 +1,3 @@
-
 import { Editor, Mark, Raw } from '../..'
 import Portal from 'react-portal'
 import React from 'react'
@@ -69,7 +68,7 @@ class HoveringMenu extends React.Component {
    */
 
   onChange = (state) => {
-    this.setState({ state })
+      this.setState({ state })
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
The reason for it being there is that the selection is updated in `<Leaf>` nodes when they are rendered, to ensure it's in the right spot. But that updating will trigger `onSelect`, so we throw out the selection changes while rendering. Open to another way of handling this though. I got that technique from Draft.js originally.

arekkas:
> But that updating will trigger onSelect

Is there no other way of preventing the event from propagation? This causes some really weird selection issues.

arekkas:
After seeing how Leaf works, the current approach makes sense, I think. Closing for now.

arekkas:
Unfortunately, this is an issue still, but I don't have good ideas how to fix this.

ianstormtaylor:
@arekkas sorry for the trouble here. 

I've pushed up a change that removes the `isRendering` block, since I can't find a case that it impacts now that we have the separate check to ensure that the selection isn't the same. If one does crop up though we may have to revert it, but hopefully it turned out to be unnecessary!

PR #387, opened 2016-10-16
Fix #385: Re-render when selection state changed for a node

### LINKED ISSUE #385: Node not re-render for selection changed just after insertion
In our application, we render a popover around images when the node is selected.

But after inserting a new image using `transform.insertInline(...)`, the popover is display, but clicking somewhere else on the document doesn't re-render the image without the popover.

If we click again on the image, then click back on the document it works as expected; it also works as expected for image already inserted in the document.

---

JSFiddle that reproduced the bug: https://jsfiddle.net/ad61x74c/

---

I'll investigate this bug and publish a PR as soon as I find the cause (probably a condition [`shouldComponentUpdate `](https://github.com/ianstormtaylor/slate/blob/master/src/components/node.js#L89).

SamyPesse:
This PR fixes #385, basically it extends the condition in the `shouldComponentUpdate` of `Node` to re-render when selection has changed for an `inline` or `block` node:

- selection is in the node != selection was in the node
- was focused != is focused

We'd be great to have unit tests for this, but I'm not sure how to write unit tests for `shouldComponentUpdate`.

### DIFF
--- a/src/components/node.js
+++ b/src/components/node.js
@@ -125,8 +125,8 @@ class Node extends React.Component {
     // the node, to allow for intuitive selection-based rendering.
     if (
       this.props.node.kind != 'text' &&
-      this.props.state.isFocused &&
-      this.props.state.selection.hasEdgeIn(this.props.node)
+      props.state.isFocused != this.props.state.isFocused &&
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Awesome, thanks for this! 

Agreed that we need better tests for `shouldComponentUpdate`, but since we don't have them yet I'm happy to merge for now.

Issue #379, opened 2016-10-12
Deleting an inline node: Could not find a descendant node with key `x`


roadscape:
https://jsfiddle.net/bga2f328/4/

To reproduce: select the whole link and press backspace (or backspace until there is only one character left in the link), and you'll get the following error:

![image](https://cloud.githubusercontent.com/assets/5168676/19324735/fd37ed1a-9090-11e6-97bd-69552a30288c.png)

---

I put logging into getDescendant -- at first the node is found, but the next time it's not (which would make sense if it was deleted, but why is it still asserting it exists afterwards?) 

Note the asterisks--

```
{
  "document": {
    "kind": "document",
    "nodes": [
      {
        "data": {},
        "key": "yygx",
        "kind": "block",
        "isVoid": false,
        "type": "paragraph",
        "nodes": [
          {
            "data": {
              "href": "https://en.wikipedia.org/wiki/Hypertext"
            },
            "key": "2aki",
            "kind": "inline",
            "isVoid": false,
            "type": "link",
            "nodes": [
              {
                "key": "n3iy",
                "kind": "text",
                "ranges": [
                  {
                    "kind": "range",
                    "text": "deleteme",
                    "marks": []
                  }
                ]
              }
            ]
          }
        ]
      }
    ]
  },
  "kind": "state"
}
node.js:494 % getDescendant called on ehoy looking for n3iy not found, recursing...
node.js:504 % getDescendant called on ehoy looking for n3iy found in tree! ***********
node.js:494 % getDescendant called on yygx looking for 2aki found!
node.js:494 % getDescendant called on yygx looking for 2aki found!
node.js:494 % getDescendant called on ehoy looking for n3iy not found, recursing...
node.js:504 % getDescendant called on ehoy looking for n3iy found in tree!
node.js:494 % getDescendant called on ehoy looking for 2aki not found, recursing...
node.js:494 % getDescendant called on yygx looking for 2aki found!
node.js:504 % getDescendant called on ehoy looking for 2aki found in tree!
node.js:494 % getDescendant called on ehoy looking for yygx found!
node.js:494 % getDescendant called on ehoy looking for n3iy not found, recursing...
node.js:494 % getDescendant called on yygx looking for n3iy not found, recursing...
node.js:504 % getDescendant called on yygx looking for n3iy not found in tree
node.js:504 % getDescendant called on ehoy looking for n3iy not found in tree ***********
[...]

Issue #347, opened 2016-09-22
backspacing to the start of an inline node errors


ianstormtaylor:
From #317:

> Inserting a link at the start of a line, and then backspacing from the end of the link to its beginning: as the first character is deleted it throws Uncaught Error: Could not find a descendant node with key "xinw".

joelburget:
I'd like this fixed and could take a look, but would prefer guidance on how you'd tackle it first.

arekkas:
This is resolved in the 0.15 branch!

ianstormtaylor:
Woot!

PR #389, opened 2016-10-16
Fix #388: prevent selection from being unset

### LINKED ISSUE #388: Selection should never be unset
@ianstormtaylor A few months ago, you opened [#36 `get rid of the concept of an "unset" selection, normalizing instead`](https://github.com/ianstormtaylor/slate/issues/36), but since then, it has been rollbacked by the commit [`fix selection setting for history`](https://github.com/ianstormtaylor/slate/commit/521f0d9f1d7d820389c8681c96429c02b35eb98a).

This is causing issues in a lot of places, and requires to maintain conditions like `if (!selection.startKey) { ...` everywhere. For example @bkniffler opened a PR to handle these cases for some methods: [#329](https://github.com/ianstormtaylor/slate/pull/329).

It's causing bugs in our application with some weird scenarios where the editor is focused but the selection is unset, and pressing enter or backspace throw critical errors.

Our current solution will be to add conditions like `if (!selection.startKey) return` everytime we have to work with the selection, even in `onKeyDown` where it seems illogical to get a keyboard event without a valid selection (we can reproduced this bug).

I think the selection can be focused or blurred; but **should never** be unset, and we should always normalize to keep a valid selection.

What are the reasons for having potential unset selections ?  
Should we go with adding conditions to handle this case and merge #329 ?

SamyPesse:
The goal of this PR is to fix #388. It basically revert the changes of [`521f0d9f1d7d820389c8681c96429c02b35eb98a`](https://github.com/ianstormtaylor/slate/commit/521f0d9f1d7d820389c8681c96429c02b35eb98a).

The issues for unit tests was the transform `setSelection` which build its inverse operation, but fails to do it when the nodes has been removed because we were doing `document.getPath(prevProps.focusKey)` (looking in the current document for a key from the previous document) which throws `Could not find a descendant node with key "ia36".`.  
My current solution is to always use keys instead of paths (so avoid resolving it using `getPath`), but I guess I'm missing something about why we need paths instead of keys.

I'm opening this Pr, but I'll keep investigate the best solution and try to have a better understanding of the issue.


src/transforms/operations.js
```diff
@@ -387,7 +387,7 @@ export function setSelectionOperation(transform, properties) {
   }
 
   // Resolve the selection keys into paths.
-  if (props.anchorKey) {
+  /* if (props.anchorKey) {
```
SamyPesse: I've commented this part since the `getPath` are failing because we are resolving a removed key in the new document. `getPath` throws `Could not find a descendant node with key "ia36".`.

But I'm not removing it till I understand what is the exact reason you were using paths instead of keys.

SamyPesse: BTW: this is could also be the source of error for these issues: #379 and #347

SamyPesse:
It also fixes the `Link` example, which is currently broken when trying to remove a link.

src/transforms/operations.js
SamyPesse: It can be fixed combine with #396

SamyPesse:
Closing in favour of #396

PR #390, opened 2016-10-16
Add transform "unwrapBlockByKey" and "unwrapInlineByKey"

SamyPesse:
This PR adds 2 more simple transforms: `unwrapBlockByKey` and `unwrapInlineByKey`. These transforms are only utilities when for example adding a popover button to unwrap a link.

⚠️ **Important:** I've found an important bug, `unwrap[Inline|Block]` unwrap all the nodes in the document, not only the one in the given range. (posted as #391)

- [x] Documentation
- [x] Tests
- [x] Transforms
- [x] Fix unwrap unwrapping all nodes in the document
- [x] Add more tests to avoid #391 to happen again


SamyPesse:
@ianstormtaylor This PR is ready for review :)

PR #392, opened 2016-10-18
Fix normalize with schema

ianstormtaylor:
@SamyPesse this was the other approach I was trying to use by normalizing against a `Schema`, which would be simpler. But I didn't get super far. I think these are the things to do:

- [ ] Remove `transform.normalize()` calls in the existing transforms, since it happens in one place now
- [ ] Port the new tests from `remove-normalize` branch
- [ ] Add more schema rules to `transforms/normalize.js` to account for everything in `Node.normalize`
- [ ] Optimize `Node.validate` to recurse down the tree, so that it's fast when memoized

### DIFF
--- a/src/models/block.js
+++ b/src/models/block.js
@@ -53,6 +53,10 @@ class Block extends new Record(DEFAULTS) {
     properties.isVoid = !!properties.isVoid
     properties.nodes = Block.createList(properties.nodes)
 
+    // if (properties.nodes.size == 0) {
+    //   properties.nodes = properties.nodes.push(Text.create())
+    // }
+
     return new Block(properties).normalize()
   }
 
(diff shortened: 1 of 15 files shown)


src/transforms/at-range.js
```diff
@@ -94,7 +94,7 @@ export function deleteAtRange(transform, range) {
 
   const lonely = document.getFurthest(endBlock, p => p.nodes.size == 1) || endBlock
   transform.removeNodeByKey(lonely.key)
-  transform.normalizeDocument()
+  transform.normalize()
```
ianstormtaylor: These can be removed, since it's in `models/transform.js#apply` once now.

src/models/block.js
```diff
@@ -53,6 +53,10 @@ class Block extends new Record(DEFAULTS) {
     properties.isVoid = !!properties.isVoid
     properties.nodes = Block.createList(properties.nodes)
 
+    // if (properties.nodes.size == 0) {
+    //   properties.nodes = properties.nodes.push(Text.create())
+    // }
```
ianstormtaylor: Not sure why I commented this out

ianstormtaylor:
Closing in favor of https://github.com/ianstormtaylor/slate/pull/396

PR #384, opened 2016-10-14
fix for deleting across an inline

tyler-johnson:
Noticed an issue with Slate deleting around an inline node, where the selection start and end are in text nodes. This test shows the issue. I will attempt a bug fix next week.

PR #393, opened 2016-10-18
fix deleteing across inlines for remove-normalize

tyler-johnson:
This is the same fix for #384, but since it conflicts with the remove-normalize branch (#354), I figured I would make the changes directly.


ianstormtaylor:
@tyler-johnson I think this is good to close now that the other one was merged! Let me know if not though

PR #396, opened 2016-10-19
Normalization done by core schema

### LINKED ISSUE #397: change selection manipulation methods to always use leaf text nodes
Right now methods on the selection like `collapseToStartOf` and `moveToRangeOf` take in a node, and set the selection's `anchorKey` and `focusKey` to that node's key. Instead, they should be smarter and find the proper descendant text nodes's keys to use if the passed in node is an `Inline` or `Block`

### LINKED ISSUE #403: replace uid() calls with node.regenerateKey()
Previously we'd been using `uid` to generate new keys in a few places, which would be better to switch out for using the new `regenerateKey()` method everywhere I think

SamyPesse:
The goal is to move all normalization to a core schema using the same API as plugins.

### APIs
- [x] Normalization is now done through `Schema` and transforms only
- [x] Transform `transform.normalize()` to normalize the whole state using core schema
- [x] Transform `transform.normalizeNodeByKey(key)` to normalize a specific node and its children using core schema
- [x] Transform `transform.normalizeParentsByKey(key)` to normalize a specific node and its parents using core schema
- [x] Merged #354 in it (it adds `transform.joinNodeByKey`, etc)
- [x] Fix #403: use `regenerateKey` instead of `uid`
- [x] Fix #397: Resolve to leaf texts when using selection methods like `collapseToEndOf`

### Misc
- [x] Add more benchmarks:
  - [x] Deserialize large document
  - [x] Render large document
- [x] Add example for a large document
- [x] Improve unicity of keys, previous `uid()` was causing issues with large documents

### Normalization in Core Schema

All rules are implemented in [`src/plugins/schema.js`](https://github.com/GitbookIO/slate/blob/schema-normalize/src/plugins/schema.js)
- [x] No duplicate keys
- [x] Only allow block nodes in documents
- [x] Only allow block, inline and text nodes in blocks
- [x] Only allow inline and text nodes in inlines.
- [x] At least one text node in blocks/inlines
- [x] Void nodes contain a single space of content
- [x] Inline void nodes are surrounded with text nodes
- [x] No adjacent text nodes
- [x] No inline node is empty
- [x] No extra empty text nodes
### Update for `Schema` normalization process
- [x] All rules matching a node are applied (previously only the first one was applied)
- [x] Transform returned by rules are applied in series
- [x] The children are normalized before the node itself
- [x] **Performances:** Replace calls to `transform.normalize()` by calls to `transform.normalizeNodeByKey(node)`
- [x] **Performances:** Normalize only the right part of the document, don't run `.normalize()` in `transform.apply()`
- [x] Don't duplicate normalize logic between transforms and schema

### Unit tests / Benchmarks
- [x] Adapt all tests
- [x] Add more tests for normalize rules
- [x] Add tests to ensure that duplicate keys are regenerated
- [x] Add more test for `deleteAtRange`
- [x] Add tests for undo of `splitNode`

### Cleanup
- [x] Remove `test.js` file


tyler-johnson:
I am a little curious how this relates to #354? I was under the assumption that normalization was to be removed altogether, however this seems to move normalization to the core plugin. Is this an replacement for #354, or is this to be used in conjunction with those changes?

ianstormtaylor:
@tyler-johnson good question. This is a replacement (that happens to use the same commits 😄) that @SamyPesse has awesomely decided to work on. Instead of trying to do the normalization afterwards in the transforms themselves, we want to see if we can get things working with the `Schema` doing the normalization in a way that's just as performant (maybe more, who knows) but also much simpler and less error-prone for future tweaks.

This will also force `Schema` to be abstracted well for when developers want to layer in additional normalization logic (eg. only allowing certain block `type`s in a document).

The critical piece regardless of how normalization is performed is that it is performed as a series of transforms (and thus operations) on the state, so that it works in history and collaboration.

Hope that makes sense!

SamyPesse:
Since the PR is now working (passing all previous tests and new ones), I'm working on some performance improvements:

- **Selection:** When clicking / moving the cursor, previously the normalization was called 3 times. With [`112b7c8`](https://github.com/GitbookIO/slate/commit/112b7c84069f57cf8e69fd4a59fc5866953dd6cd), moving the cursor doesn't normalize the document, only the selection.
- **onBeforeChange/onChange:** Since the normalization is performed on the `transform.apply()` and explicitly in the `onBeforeChange`, it may be better to try to remove the `onBeforeChange` in the `onChange` (so that normalization will be called only once instead of twice when the user is typing).

src/plugins/schema.js
```diff
@@ -0,0 +1,283 @@
...
+  }
+}
+
+/**
+ * A default schema rule to ensure that inline nodes are not empty
+ *
+ * @type {Object}
+ */
+
+const INLINE_NO_EMPTY = {
```
Soreine: Just a note: This rule should act on the parent block, since it removes the inline. This is fixed in an upcoming branch.

src/plugins/schema.js
```diff
@@ -0,0 +1,283 @@
...
+      }
+
+      const prevNode = index > 0 ? block.nodes.get(index - 1) : null
+      const nextNode = block.nodes.get(index + 1)
+
+      const prev = (!prevNode || isInlineVoid(prevNode))
+      const next = (!nextNode || isInlineVoid(nextNode.kind))
+
+      if (next || prev) {
+        accu.push({ next, prev, index })
```
Soreine: I think there's a bug here. The index passed should be shifted for every previously inserted text. Have a counter, incremented by 1 each time `next` is true, and each time `prev` is true.

Then in normalization, the insertion index of `prev` is always the passed `index`, but the insertion index of `next` should be `index + 1` if `prev = false` or `index + 2` if `prev = true`

src/plugins/schema.js
```diff
@@ -0,0 +1,283 @@
...
+
+        return [n, next]
+      })
+      .filter(Boolean)
+
+    return invalids.size ? invalids : null
+  },
+  normalize: (transform, node, pairs) => {
+    return pairs
+        // We reverse the list since we want to handle 3 consecutive text nodes
```
Soreine: Because `joinNodeByKey` merges the second node into the first node, if I understand well.

src/transforms/normalize.js
```diff
@@ -1,31 +1,188 @@
...
   let { document, selection } = state
   selection = selection.normalize(document)
+
+  // If the selection is nulled (not normal)
+  if (
+    selection.isUnset ||
+    !document.hasDescendant(selection.anchorKey) ||
+    !document.hasDescendant(selection.focusKey)
+  ) {
+      warning('Selection was invalid and reset to start of the document')
```
Soreine: The dev warning is a great idea

SamyPesse:
Update:

We still a few works to do on fixing the history (we are using a custom history in our application, and this PR breaks Slate's one).

And also improve the perfs, our latest changes slow down Slate compared to our first changes, we are investigating this.

This PR will be ready for middle of next-week.

Currently, this PR adds 43 tests (476 -> 519)

Soreine:
I have uploaded the benchmarks comparisons between

1. The last commit on this PR and master : https://gist.github.com/Soreine/76b91c21dde9f4e744c12dbc518d2508

2. The last commit on this PR and before we start normalizing the selection at the operation level https://gist.github.com/Soreine/f5ec25b073db5893a6cc0f3d3ae1c575

Overall, this PR is largely faster than master. But the selection logic in operations we added made things a bit slower. We are working on improving that now.

src/utils/memoize.js
```diff
@@ -36,24 +41,79 @@ function memoize(object, properties) {
...
+  value = value === undefined ? UNDEFINED : value
+
+  let parentMap = map
+  let childMap
+  for (const key of keys) {
+    childMap = parentMap.get(key)
+
+    if (childMap === UNSET) {
+      // This path was not created yet
+      childMap = new Map()
```
SamyPesse: Eslint was failing because of Map, do we need a polyfill ?

src/transforms/normalize.js
```diff
@@ -1,31 +1,288 @@
...
+const MAX_CALLS = 50
+
+/**
+ * Refresh a reference to a node that have been modified in a transform.
+ * @param  {Transform} transform
+ * @param  {Node} node
+ * @return {Node} newNode
+ */
+
+function _refreshNode(transform, node) {
```
ianstormtaylor: Can you change these private functions to not be prefixed with `_`, and instead just have them at the end of the file? (For consistency with other places we do this)

src/transforms/operations.js
```diff
@@ -106,14 +105,27 @@ export function insertTextOperation(transform, path, offset, text, marks) {
...
+      offset,
+    }]
+  } else {
+    // The number of children after which we split
+    const count = node.nodes.count()
+
+    inverse = [{
+      type: 'split_node',
+      path: withPath,
+      count,
```
ianstormtaylor: Why change this to `count` here? It feels like it makes the `split_node` operation harder to reason about and standardize. Instead could we not calculate the offset of the nodes and use that?

src/utils/get-leaf-text.js
```diff
@@ -0,0 +1,24 @@
...
+ * @return {Text} text
+ */
+
+function getLeafText(node) {
+  if (node.kind == 'text') {
+    return node
+  }
+
+  const texts = node.getTexts()
+  return texts.first()
```
ianstormtaylor: Should this be `node.getFirstText()` now instead?

src/utils/is-dev.js
```diff
@@ -0,0 +1,12 @@
...
+  process.env &&
+  process.env.NODE_ENV !== 'production'
+)
+
+/**
+ * Return true if running slate in development
+ * @type {Boolean}
+ */
+
+export default IS_DEV
```
ianstormtaylor: Let's put this inside `src/constants` instead

src/utils/uid.js
```diff
@@ -1,14 +1,16 @@
...
+
 /**
  * Create a unique identifier.
  *
  * @return {String} uid
  */
 
 function uid() {
-  return generate(4)
+  return (N++) + generate(4)
```
ianstormtaylor: What's this for?

SamyPesse: We noticed a lot of conflicts in keys when the document is large; for example with a document containing 2000/3000 nodes, we were getting key conflict each time.

It reduces the probability of conflicts.

src/index.js
```diff
@@ -20,6 +20,7 @@ import Schema from './models/schema'
 import Selection from './models/selection'
 import State from './models/state'
 import Text from './models/text'
+import Range from './models/range'
```
SamyPesse: We are working on a new version of [`markup-it`](https://github.com/GitbookIO/markup-it/pull/32) which is using Slate for its data models, and we wanted to use `Range` when building `Text`.

Do you think we should not expose it?

src/transforms/normalize.js
Soreine: ✅

src/transforms/normalize.js
```diff
@@ -1,31 +1,288 @@
+import warning from '../utils/warning'
+import { default as defaultSchema } from '../plugins/schema'
+import Normalize from '../utils/normalize'
+import { Map } from 'immutable'
+
+// Maximum recursive calls for normalization
+const MAX_CALLS = 50
```
Soreine: A certain rule can normalize a node for a certain aspect. But a single rule is not aware of the aspects treated by other rules. So we decided to call normalization rules on a node until all rules pass. `MAX_CALLS` is the maximum number of pass. Without this guard, if you have a buggy rule that does not fix the node to pass its own validation, it will be called indefinitely. This means tests will flood the terminal with stack overflows, and the browser will simply freeze and need to be killed (very annoying when developing).

src/transforms/at-range.js
```diff
@@ -862,31 +970,40 @@ export function wrapInlineAtRange(transform, range, inline) {
...
+      const node = inline.regenerateKey()
+      transform.insertNodeByKey(block.key, 0, node, { normalize: false })
 
       block.nodes.forEach((child, i) => {
-        transform.moveNodeByKey(child.key, node.key, i)
+        transform.moveNodeByKey(child.key, node.key, i, { normalize: false })
       })
+
+      if (normalize) {
+        transform = transform.normalizeNodeByKey(block.key)
```
Soreine: I have removed them :)

I think we kept this syntax out of nostalgia. Deep in our heart, we have always wished that transform had stayed immutable. Keeping this syntax would have made it easier to change it back :)

src/transforms/at-current-range.js
```diff
@@ -95,70 +95,11 @@ export function _delete(transform) {
...
-        after = selection.collapseToEndOf(last)
-      }
-    } else {
-      after = selection.moveBackward(n)
-    }
-  }
-
-  else {
-    after = selection.moveBackward(n)
-  }
```
Soreine: Yes, it is because most of the work is already done in the operations themselves. It may be a bit slower in the end, but it's much more convenient to write transforms this way. One of the great idea from @SamyPesse

src/transforms/at-current-range.js
```diff
@@ -95,70 +95,11 @@ export function _delete(transform) {
...
+  const { selection } = state
 
   return transform
-    .unsetSelection()
     .deleteBackwardAtRange(selection, n)
-    .moveTo(after)
+    .collapseToEnd()
```
Soreine: No collapsing should be necessary indeed. Actually, if we remove this, all the unit tests still pass. The problem is related to `isNative` maybe, because if you remove the `collapse` and in the examples press `Ctrl + D`, you can see the cursor moving backward by one character whereas our code never changed the selection. This is something we should investigate further.

src/transforms/apply-operation.js
```diff
@@ -187,14 +233,55 @@ function removeMark(state, operation) {
...
+  const startDesc = startKey ? getRemoved(startKey) : null
+  const endDesc = endKey ? getRemoved(endKey) : null
+
+  if (startDesc) {
+    const prevText = prevDocument.getTexts()
+      .takeUntil(text => text.key == startKey)
+      .filter(text => !getRemoved(text.key))
+      .last()
+    if (!prevText) selection = selection.unset()
+    else selection = selection.moveStartTo(prevText.key, prevText.length)
```
Soreine: ✅

src/transforms/apply-operation.js
```diff
@@ -187,14 +233,55 @@ function removeMark(state, operation) {
...
+      selection = selection.collapseToStart()
+    } else {
+      const nextText = prevDocument.getTexts()
+        .skipUntil(text => text.key == startKey)
+        .slice(1)
+        .filter(text => !getRemoved(text.key))
+        .first()
+
+      if (!nextText) selection = selection.unset()
+      else selection = selection.moveEndTo(nextText.key, 0)
```
Soreine: ✅

src/models/schema.js
```diff
@@ -138,6 +45,28 @@ class Schema extends new Record(DEFAULTS) {
   }
 
   /**
+   * Return true if one rule can normalize the document
+   *
+   * @return {Boolean} isNormalization
+   */
+
+  get isNormalization() {
```
Soreine: ✅

src/models/selection.js
```diff
@@ -621,23 +704,6 @@ EDGE_METHODS.forEach((pattern) => {
 })
 
 /**
- * Memoize read methods.
- */
-
-memoize(Selection.prototype, [
-  'hasAnchorAtStartOf',
```
Soreine: All these methods rely directly on methods from `Node` which are already memoized. This removes the overhead of memoization for `Selection`. It's rare to reuse the same `Selection` object, selections change often, whereas nodes don't.
In practice, I removed them when I was optimizing the rendering. `shouldComponentUpdate` does a lot of `hasEdgeIn`. The memoization was slowing everything down with no benefit. Without memoization, the rendering was faster.

src/plugins/schema.js
```diff
@@ -0,0 +1,304 @@
...
+    const invalids = nodes.filter(n => n.kind != 'inline' && n.kind != 'text')
+    return invalids.size ? invalids : null
+},
+  normalize: (transform, inline, invalids) => {
+    return invalids.reduce((t, n) => t.removeNodeByKey(n.key, { normalize: false }), transform)
+  }
+}
+
+/**
+ * A default schema rule to ensure that inline nodes are not empty
```
Soreine: I added a comment, which is true for any normalization rule that remove nodes.
> This rule is applied to all blocks, because when they contain an
empty inline, we need to remove the inline from that parent
block. If `validate` was to be memoized, it should be against the
parent node, not the inline themselves.

src/plugins/schema.js
```diff
@@ -0,0 +1,304 @@
...
+  }
+}
+
+/**
+ * A default schema rule to ensure that void nodes contain a single space of content.
+ *
+ * @type {Object}
+ */
+
+const INLINE_VOID_TEXT_RULE = {
```
Soreine: ✅

src/transforms/apply-operation.js
```diff
@@ -120,14 +132,46 @@ function insertText(state, operation) {
...
+ *   respective last node and first node of the nodes' children. Like a zipper :)
  * @return {State}
  */
 
 function joinNode(state, operation) {
-  const { path, withPath } = operation
-  let { document } = state
-  document = document.joinNode(path, withPath)
-  state = state.merge({ document })
+  const { path, withPath, deep = false } = operation
```
Soreine: We need it to implement correctly the inverse of the `splitNode` operation.

Here is a case, as tested in `fixtures/on-history/undo/split-node-block`

```yaml
nodes:
  - kind: block
    type: paragraph
    nodes:
      - kind: text
        text: SplitHere
```

We `splitNode` the paragraph, at an `offset` of 3, we get this:

```yaml
nodes:
  - kind: block
    type: paragraph
    nodes:
      - kind: text
        text: Split
  - kind: block
    type: paragraph
    nodes:
      - kind: text
        text: Here
```

Now, to revert that, we need to `joinNode` both paragraph.

```yaml
nodes:
  - kind: block
    type: paragraph
    nodes:
      - kind: text
        text: Split
      - kind: text
        text: Here
```

And then `joinNode` again the children text:

```yaml
nodes:
  - kind: block
    type: paragraph
    nodes:
      - kind: text
        text: SplitHere

```

`splitNode` splits all the way down, we need a way to join all the way down too :). The `deep` option of `joinNode` does that, by joining nodes like a zipper. 

By default, `deep=false`, and `joinNode` only join two nodes, no further, as before.

src/transforms/apply-operation.js
```diff
@@ -226,10 +326,10 @@ function removeText(state, operation) {
  */
 
 function setMark(state, operation) {
-  const { path, offset, length, mark, properties } = operation
+  const { path, offset, length, mark, newMark } = operation
```
Soreine: To have some context here, the lowest level method `Text.updateMark` was the one responsible of [updating the mark](https://github.com/ianstormtaylor/slate/blob/master/src/models/text.js#L290). It needed to have the existing `Mark` value, would then create a new `Mark` by merging properties with `Immutable.Map.merge`, and would replace the previous `Mark` value everywhere with the new value.
Since `Text.updateMark` creates the new `Mark`, we could not provide the new `Mark` yet when we needed to [create the inverse operation](https://github.com/ianstormtaylor/slate/blob/master/src/transforms/operations.js#L297). So the inverse operation was not updating anything.
I decided to move the update logic of the mark higher in the stack. It's more flexible, and allowed to fix the issue.

I think a `Mark` can be serialized to JSON easily. If not, then the existing `mark` argument was already a problem.

src/transforms/operations.js
Soreine: See my comment on `splitNodeAfter` :)

src/utils/get-leaf-text.js
Soreine: ✅

src/utils/is-dev.js
Soreine: ✅

src/utils/memoize.js
Soreine: I used [`es6-map`](https://www.npmjs.com/package/es6-map)

Soreine:
I answered and applied most of your suggestions =D. I'm just waiting for this last one about `onBeforeChange`. I'll write a detailed summary tomorrow.

src/index.js
ianstormtaylor: Nope, I don't have an opinion. Just curious since it's always better to export less, but if you're using it for a valid use case that sounds good to me to keep it exported!

src/models/node.js
```diff
@@ -930,11 +978,26 @@ const Node = {
    */
 
   getTexts() {
+    return List(this._getTexts())
```
ianstormtaylor: Why are the arrays faster than the list?

src/models/node.js
```diff
@@ -1329,6 +1321,40 @@ const Node = {
...
   /**
+   * Split a node by `path` after 'count' children.
+   * Does not work on Text nodes. Use `Node.splitNode` to split text nodes as well.
+   *
+   * @param {Array} path
+   * @param {Number} count
+   * @return {Node}
+   */
+
+  splitNodeAfter(path, count) {
```
ianstormtaylor: What if the original `joinNode` operation "zippered" all the way down to the text nodes? That seems like a reasonable thing for it to do when joining, which would then make splitting simple?

src/models/node.js
```diff
@@ -1380,54 +1405,30 @@ const Node = {
...
 /**
  * Memoize read methods.
  */
 
 memoize(Node, [
-  'assertChild',
-  'assertDescendant',
-  'filterDescendants',
-  'filterDescendantsDeep',
-  'findDescendant',
```
ianstormtaylor: Nice, that makes total sense!

src/transforms/apply-operation.js
ianstormtaylor: Continuing the idea from above, why not always have `joinNode` be deep and "zipper" the nodes together?

Or, if that doesn't work... what about having `joinNode` and `splitNode` take a `depth` argument, which defaults to `Infinity`, but can be passed in for cases where we need the control instead?

src/transforms/apply-operation.js
ianstormtaylor: Boom, yes you are totally right. Sounds good!

src/transforms/normalize.js
ianstormtaylor: Awesome, that makes sense, good call

ianstormtaylor:
Awesome, just commented on the remaining ones. Looking great!

src/components/editor.js
```diff
@@ -202,8 +202,6 @@ class Editor extends React.Component {
   onChange = (state) => {
     if (state == this.state.state) return
 
-    state = this.onBeforeChange(state)
```
SamyPesse: Yes we can probably remove the `onBeforeChange` since we are no longer using it (the schema is doing the job).

If we remove it, we have to move the normalization with the plugins schemas to the react component itself.

src/models/node.js
Soreine: Good question. There is a benchmark for `document.getTexts()`. On my machine this is the results:
- Array 52,000 ops/sec
- List 16,000 ops/sec

I would expect it to be faster to concat List (linked list) than Arrays (copied values)... So I don't know why.

src/models/node.js
Soreine: We need to have a way to simply join two nodes. So instead, we could keep `joinNode` that way, and have a `mergeNode` that does what you say. We would have 4 operations, 2 of them being the reverse of the two others:

```
splitNode <-> mergeNode
splitNodeAfter <-> joinNode
```

The names could be better

src/models/node.js
```diff
@@ -1380,54 +1405,30 @@ const Node = {
...
-  },
-
-  /**
-   * Validate the node against a `schema`.
-   *
-   * @param {Schema} schema
-   * @return {Object || Void}
-   */
-
-  validate(schema) {
```
Soreine: Here is the branch https://github.com/GitbookIO/slate/pull/6

src/utils/uid.js
```diff
@@ -8,7 +8,7 @@ import generate from 'uid'
  */
 
 function uid() {
-  return generate(4)
+  return String(N++)
```
kommen: We're running into issues with this, where the document and the first text node both have the id `"0"`. Can this be some sort of race condition?

kommen: No race condition. I guess N must not be initialized to 0. If one loads a persisted slate state into a new slate instance, the state of N is lost and starts creating nodes from 0 again, causing duplicate uids.

Soreine: hmm...
Well the persisted state does not store the keys. So you could have two instances of the Slate library loaded. There's only one place where we re-keys the nodes as needed, that's in the `Node.insertNode` method. So maybe you are bypassing that method somehow too ?

Soreine:
# High level summary of the changes and motivations

- **Normalization now uses exclusively transforms** (and therefore operations). Normalization therefore inherits the properties of operations, and so **becomes always serializable and reversable (undo/redo)**. 

- **Normalization is now implemented using Schemas** that were already in core. It becomes a lot simpler to write the normalization logic, in terms of rules. We have now two normalization Schemas, the core Schema, and optionally the plugins' Schema.
The validation rules from **the plugins' Schema are still applied on the whole document on `Editor.onChange` only**. However, for convenience, **the core Schema is applied after each core transforms by default** (most transforms `*ByKey` now take an optional `normalize` flag). This makes writing transforms easier, since we have a control on the normalization, so we can skip it at will, or use it at will, depending on what we want to achieve. It also makes core normalization faster, since transforms can restrict normalization to nodes that have changed only.

- In Schemas, **to normalize a given Node, the validation rules are called repeatedly, until all of them pass**. Previously, the first rule that would match and normalize a node, had to somehow normalize it in a way that should be valid for all remaining rules. This is not feasible in practice.
 
- **Operations now do their best to adapt the selection**. This makes it even easier to write  transforms, since in most cases, the underlying operations will update the selection as you would expect.

- **Memoization was optimized**. It is slightly faster in implementation, using native `Map` instead of `Immutable.Map`. But most importantly, methods have been rewritten to take full advantage of memoization.
  - Within Slate, all methods that coerce argument, are passed the coerced value directly to optimize cache hits (for example we always pass a key string directly, where a `Node` or a `String` are both accepted).
  - All methods that take function as argument (`Node.findDescendant` etc.) are no more memoized, since they seldom receive the same function parameter twice. Instead, canonical uses of these methods are available through dedicated methods which are memoized on their own (`Node.getDescendant` etc.).
  - A lot of methods are no longer memoized, either because they provided no real benefit, or they themselves relied on already memoized methods (for example `Selection` methods).

- And various other performance improvements

ianstormtaylor:
@Soreine, amazing! Thank you! All of that sounds great. Is there anything else were waiting for? I'm ready to merge if you two are happy with it!

@kommen let us know what you find, and we can figure out what the issue is and fix it.

SamyPesse:
@ianstormtaylor I think we are good to merge 🚀

src/utils/uid.js
kommen: The keys are persisted when when using the `Raw` serializer.

kommen:
As per my code comment above, we're running into issues with the new uid generation in this PR. We serialize slate states with the `Raw` serializer, and later load that into a new slate instance. As the serialized state also contains the keys beginning from 0, but new nodes now will also get keys beginning from 0 again, leading to duplicate keys. Or am I missing something here?

If you're interested why draft.js doesn't have incrementing keys, you can find some discussion about it over there: https://github.com/facebook/draft-js/pull/268#issuecomment-208394793

Soreine:
I merged master for the last time.
I wanted to `git rebase -i` to squash a lot of commits... but I think it's too late because we merged master a few times already (and other sub-PR), and this [generates conflicts during `git rebase -i`](http://stackoverflow.com/questions/3133449/why-does-git-rebase-give-me-merge-conflicts-when-all-im-doing-is-squashing-comm).

Here is [the squashes I wanted to do](https://gist.github.com/Soreine/70391ceec8739add6592c18d7543df49), if you ever happen to be a git wizard and manage to do it without conflicts...

Most of the commits make sense though...

ianstormtaylor:
@kommen ah thank you, I forgot that the non-terse `Raw` serialization does in fact keep the keys intact across serializations. 

Question for you: is there any reason you need the keys? Asking because I forgot that was the case, and I think it would be good for us to not actually serialize the keys in general, since they're not really intended to be used for uniquely identifying forever, and more intended as a convenient way to uniquely identify for the single lifespan of a state in memory.

I propose that we remove the keys from the serialized ouput, and then they will properly get regenerated when serializing and deserializing, so then the new system shouldn't be a problem.

(If someone comes up with a valid use case for preserving them, we can add a `preserveKeys` option to the serializer in the future. At which point ensuring that keys aren't duplicated would be a user-level concern if they chose to enable it.)

---

@Soreine I'll check out the rebasing/merging options! Thanks for that list. Worst case I'll just merge them all and preserve the history completely. (Seems better than squashing to a single huge commit, yeah?)

kommen:
@ianstormtaylor Yeah right now we use the keys for cross-referencing nodes, i.e a node can point to other nodes on which it depends on. We use this then on the backend to evaluate the document.

We'd prefer to have an option to serialize the keys and maybe provide our own implementation of a key generator.

I'm sure we can come up with a different solution to handle this though, if you absolutely want to get rid of the keys.

kommen:
@SamyPesse & @Soreine: also thanks very much for this PR, really appreciate your work. Using it in production already and the performance boost we get from it is awesome :)

kommen:
@SamyPesse & @Soreine: also thank you very much for this PR, really appreciate your work. We're using this branch in production already and the performance boost we got was awesome. :)

ianstormtaylor:
@kommen okay gotcha that makes sense, and sounds like a totally valid use case. I think (for now at least) Slate should stick with the "dumb" key generation of just incrementing an integer by default. This ensures that people don't rely on the keys for uniqueness (since as demonstrated by the need for the change, they aren't very unique).

Similarly, the default should be that when serializing state, the keys are stripped, since they are local to Slate's memory implementation, and really only used as a convenience for referring to nodes for transforms.

But, I think you make a good point that it seems valid to want to set the specific key generator used, so I've added an exported function called `setKeyGenerator` that allows you to do that. And I've changed `Raw` to omit keys by default, but added a `preserveKeys: true` option that will preserve them.

---

So for your use case, once this is merged, you'll want to supply your own unique key generator, and also ensure that you are using `preserveKeys: true` when serializing state to the backend. (Deserialization will use keys if supplied, as before, no option needed.)

If you ever decide you want to do it in a way that doesn't require figuring out unique keys, you could instead write some logic that converts the metadata key references to paths when exported, and use that to denote the dependencies between nodes instead.

Let me know if I've forgotten anything!

PR #401, opened 2016-10-22
Editor: track latest state from props.

joelburget:
I was getting unexpected calls to `onDocumentChange` after passing in a
new state and then changing the /selection/ only. Since the new document
had not been set in `componentWillReceiveProps`, the selection change
triggered `onChange`, which found `state.document != this.tmp.document`,
which then triggered the spurious `onDocumentChange`.

After this change the same selection change does not trigger
`onDocumentChange`.

### DIFF
--- a/src/components/editor.js
+++ b/src/components/editor.js
@@ -122,9 +122,11 @@ class Editor extends React.Component {
       this.setState({ schema })
     }
 
-    this.setState({
-      state: this.onBeforeChange(props.state)
-    })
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @joelburget thank you for catching this and submitting a pull request! And sorry for the slow response. I've just pushed up a commit that fixes this issue in a slightly different way, so it should be solved for you in the latest release!

PR #405, opened 2016-10-25
Fix inserting fragment at the end of current text node

zyulyaev:
Previously inserting fragment at the end of current text node may lead to inconsistent selection, where offset was greater than text length.


ianstormtaylor:
Hey @zyulyaev thanks for this! Could you include a test case with this PR so that we can be sure it doesn't have a regression in the future? Thanks!

zyulyaev:
No problem.

ianstormtaylor:
Awesome, thank you!

PR #406, opened 2016-10-25
Break after applying first matching rule in HTML Serializer

### LINKED ISSUE #404: Html Serializer continues to apply rules even though a valid rule has applied
It's expected that the serializer use the returned object `ret` when a rule is successfully applied to the element; however, subsequent applicable rules are applied to the element and override the previous rules with higher priority.

At https://github.com/ianstormtaylor/slate/blob/f1a5d6f3b400c1b576ab4bcf6db97a6330d1d529/src/serializers/html.js#L179:

```
   for (const rule of this.rules) {
      if (!rule.deserialize) continue
      const ret = rule.deserialize(element, next)
      if (!ret) continue
      node = ret.kind == 'mark'
        ? this.deserializeMark(ret)
        : ret
    }
```

The for loop should break once the node is deserialized the first time.

LeeKevin:
Fixes #404

### DIFF
--- a/src/serializers/html.js
+++ b/src/serializers/html.js
@@ -183,6 +183,7 @@ class Html {
       node = ret.kind == 'mark'
         ? this.deserializeMark(ret)
         : ret
+      break
     }
 
     return node || next(element.children)


ianstormtaylor:
Thanks @LeeKevin! Could you add a test case for this so we don't have a regression later? Happy to merge it once that's there!

ianstormtaylor:
Weird, I merged it but it didn't auto-close this or show up. Must have done it in a weird way. Either way, this is merged in now!

PR #408, opened 2016-10-26
Html serializer: optionally return React elements

jasonphillips:
This PR adds an options parameter to the Html serializer so that it can be asked to return a list of React elements, instead of automatically rendering these to static markup. 

I see a number of use cases in my own application where I'd love to use the existing serializer and rules but handle placing the resulting tree of React elements into my UI directly instead of receiving an HTML string.

### DIFF
--- a/docs/reference/serializers/html.md
+++ b/docs/reference/serializers/html.md
@@ -53,7 +53,7 @@ Deserialize an HTML `string` into a [`State`](../models/state.md). How the strin
 ### `Html.serialize`
 `Html.serialize(state: State, [options: Object]) => String || Array`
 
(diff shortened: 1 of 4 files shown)


ianstormtaylor:
Hey @jasonphillips thanks for this! Sounds like a good use case. Could you change the option name from `returnElements: true` to `render: false`?

I think eventually it might be better to actually split the `Html` renderer into a `React` and `Html` renderer potentially, but doing it as an option at first sounds like a good call, so I like this!

Thanks

jasonphillips:
Updated the option name as requested; thanks. As for splitting the renderer at some point, that probably makes sense as well; I mainly handled this via an argument for now on account of the fact that the Html serializer essentially is a React renderer already, until the very last step when it calls `renderToStaticMarkup` on the results.

Issue #410, opened 2016-10-28
Selection issue with editable elements nested inside noneditable.


AlbertHilb:
I'm writing an app based on Slate to take math notes.
It uses few custom blocks: `definition`, `proposition`, `proof`; they are containers for other blocks like text, lists, images.
Definition blocks are, essentially, rendered as

```
<section className="definition" contentEditable={false}>
  <h3> Definition.</h3>
  <div className="content" contentEditable={true}>
    {props.children}
  </div>
</section>
```

The others are rendered in a similar manner.

Unfortunately there are issues editing those nodes.

It happens because Slate uses `isNonEditable` function from `content.js` to check if the target of a `select` event is a descendant of some `contenteditable="false"` element and in such a case it ignores the event even if the target itself is actually editable.

I wonder why not using [`isContentEditable` html element property](https://developer.mozilla.org/it/docs/Web/API/HTMLElement/isContentEditable) to make the check.

The only reason I can think of is to construct a node non-editable from the point of view of Slate but still editable, that is a node, the editable parts of which are not handled by Slate.

Anyway that behavior can be achieved attaching to the root element of the node a special attribute, something like `data-noneditable`, and then ignoring `select` events targeting elements with ancestors having such an attribute.

If you agree with this proposal I will happy to make  a PR.

ianstormtaylor:
Hey @AlbertHilb, good question. I had no idea the `isContentEditable` property existed, but that does sound like a better alternative. I'd be open to a PR if you were to make it. Just double-check that the current behavior for void nodes (like in the images and embeds examples) still works! Thank you :D

ianstormtaylor:
Thanks for the pull request @AlbertHilb!

PR #413, opened 2016-11-03
Use `isContentEditable` HTML element property to check if an element is editable

AlbertHilb:
Use `isContentEditable` HTML element property to check if an element is editable.
This does not change how Slate handles its content except for `contenteditable="true"` elements nested inside `contenteditable="false"`. Those elements, noneditable by Slate before, now are editable.
At the moment, as far as I could see, there should be no components with such a structure, so this PR shouldn't in fact alter Slate behavior.
Anyway, I manually tested few `isVoid: true` components: images, embeds and emojis, and they still work.

This PR fixes issue #410.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -710,10 +710,8 @@ class Content extends React.Component {
  */
 
 function isNonEditable(event) {
-  const { target, currentTarget } = event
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @AlbertHilb!

Issue #297, opened 2016-09-07
transforms blur and focus do not work properly


arekkas:
The state transforms blur and focus do not work properly in both firefox and chrome. I uploaded a demo here: https://jsfiddle.net/q2mmegtj/3/

I believe that the problem is that

```
state.transform()
    .focus() // or
    .blur()
    .apply();  
```

transforms the state, but does not do anything in the DOM. Thus, the transforms don't work properly in neither firefox nor chrome.

In Chrome, when the focus is blurred, you can still continue typing:

![slate-blur](https://cloud.githubusercontent.com/assets/3372410/20219924/c84516e8-a82b-11e6-93d4-22faf0145fec.gif)

In firefox, neither blur nor focus have *any* effect:

![slate-undo-firefox](https://cloud.githubusercontent.com/assets/3372410/20219967/ff5acdda-a82b-11e6-8cc4-073f57995c4e.gif)

I have no idea how to approach this issue. I experimented with DOMElement.focus() and .blur on the contenteditable with little effect. `window.getSelection().removeAllRanges()` has the desired effect but does not work when the next focused element has a range.

arekkas:
I think what's missing is that there is nothing that focuses the dom node, when `isFocused` changes. This could probably be done in `willReceiveProps` or something like that

arekkas:
So it looks like this is a bit platform specific:

* Firefox - blur: no, focus: no
* Chrome - blur: no, focus: yes
* Edge - blur: no, focus: no
* Safari - blur: no, focus: yes

arekkas:
this issue persists on the 0.14.14 release

arekkas:
this issue persists on the 0.14.15 release. @ianstormtaylor do you have any idea where I should start looking? I would really like to fix this, but I don't know where to start

arekkas:
indeed it looks like the contenteditable DOM node is missing `focus()` or `blur()`. I was able to get the example above working with:

https://jsfiddle.net/he8exq3L/1/

Obviously, `setTimeout` and `removeAllRanges()` is weird but I think it should give you a good idea why this is not working at the moment.

arekkas:
Here is a better demo, which by the way works fine in chrome, but not in firefox: https://jsfiddle.net/q2mmegtj/1/

Looks like a x-browser issue?

arekkas:
So, apparently `Leaf.updateSelection()` is being called on the focus toggle in chrome, but not in firefox!

PR #417, opened 2016-11-13
Resolve focus issue with firefox

arekkas:
This PR resolves issues with the focus transform in firefox. A similar solution was suggested on [stackoverflow](http://stackoverflow.com/questions/2388164/set-focus-on-div-contenteditable-element).

I don't know how to write a test for this.


arekkas:
This fix is related to #297

arekkas:
whoops messed something up

PR #418, opened 2016-11-13
Resolve issues with focus and blur transformations in the DOM

### LINKED ISSUE #297: transforms blur and focus do not work properly

arekkas:
Things to note:

1. This won't work in IE / Edge because it uses [`closest()`](http://caniuse.com/#search=closest). However, slate uses `closest()` in several places so I just rolled with it.
2. I have no idea how to write tests for this.
3. I'm not sure if it resolves all issues with focus/blur, but I tested a few cases where it worked (and previously didn't).
4. Previously, `document.activeElement` didn't change. Now, after executing the blur transform, it will be the body element.
5. We can't trust `state.selection` in isLeaf because it might be out of date in environments that are a bit slow. Because it is not possible to get the real steat I added a DOM-based check that detects if our contenteditable is still responsible for handling the selection.

Closes #297 and maybe #418

### DIFF
--- a/src/components/leaf.js
+++ b/src/components/leaf.js
@@ -6,24 +6,6 @@ import ReactDOM from 'react-dom'
 import getWindow from 'get-window'
 
 /**
- * Tests if the child is a descendant of parent.
- *
- * @param parent
- * @param child
- * @returns {boolean}
- */
-
-function isDescendant(parent, child) {
-  if (!child || !parent) return false
-  let node = child.parentNode
-  while (node != null) {
-    if (node == parent) return true
-    node = node.parentNode
-  }
-  return false
-}
-
-/**
  * Debugger.
  *
  * @type {Function}
@@ -154,6 +136,9 @@ class Leaf extends React.Component {
     const { state, ranges, isVoid } = this.props
     const { selection } = state
 
+    // If the selection is blurred we have nothing to do.
+    if (selection.isBlurred) return
+
     let { anchorOffset, focusOffset } = selection
     const { node, index } = this.props
     const { start, end } = OffsetKey.findBounds(index, ranges)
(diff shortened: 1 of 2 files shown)


arekkas:
Don't merge this just yet, I may have found an issue here

arekkas:
Yes indeed, I think there is an issue when more than one editor is included on the page.

src/components/leaf.js
```diff
@@ -148,17 +145,41 @@ class Leaf extends React.Component {
...
+
+    // If no selection is defined we're probably in a node environment and we can skip the selection magic.
+    if (!window || !window.getSelection) return
+
+    const native = window.getSelection()
+    const parent = ref.closest('[contenteditable]')
+
+    // If the selection is blurred but the native selection has still edges in this leaf,
+    // we need to clean up the ranges and blur the contenteditable.
+    if (selection.isBlurred && (hasAnchor || hasFocus)) {
```
arekkas: The problem is that I thought that `hasAnchor || hasFocus` is being computed by the native selection, but it's computed using the document selection. this obviously yields true multiple times if multiple slate instances exist and then causes the removeallranges thing to remove the selection of the second slate instance.

The idea behind this was that by checking if the current selection is still in the contenteditable although it should be blurred, we blur the selection using the DOM.

src/components/leaf.js
```diff
@@ -148,17 +163,47 @@ class Leaf extends React.Component {
...
+    const parent = ref.closest('[contenteditable]')
+
+    // If the selection is blurred but the DOM selection still focuses this leaf,
+    // we need to clean up the ranges and blur the contenteditable.
+    if (selection.isBlurred && (hasAnchor || hasFocus)) {
+      // We need to make sure that the selection from our state is up-to-date with the native selection.
+      // We can do this by checking if native.anchorNode is a descendant of our contenteditable parent, which is this
+      // slate instance. If it's not, the selection is no longer under this instance's responsibility and not further
+      // action is required.
+      if (!isDescendant(parent, native.anchorNode)) return
```
ianstormtaylor: Could we do this inside `componentDidMount/Update` inside `src/content.js` instead? That way it could be a more generic case, where we only need to check "was it focused before, blurred now, and still have a range inside the `contenteditable`?".

src/components/leaf.js
```diff
@@ -148,17 +163,47 @@ class Leaf extends React.Component {
...
+
+      // Apparently our selection is blurred, but the DOM still has ranges in the nodes we manage. To fix this,
+      // we blur the contenteditable and remove all ranges.
+      native.removeAllRanges()
+      if (parent) parent.blur()
```
ianstormtaylor: Why the need to both `blur()` and `removeAllRanges()`? I would have assumed that either one would do?

ianstormtaylor:
Hey @arekkas thanks for looking into this! The `focus()` treatment for Firefox is a good find. I added comments about the blurring logic for you to check out. Happy to get this merged in once we've sorted a few things!

src/components/leaf.js
```diff
@@ -148,17 +163,47 @@ class Leaf extends React.Component {
     const hasFocus = selection.hasFocusBetween(node, start, end)
     if (!hasAnchor && !hasFocus) return
 
+    // We have a selection to render, so prepare a few things...
+    const ref = ReactDOM.findDOMNode(this)
+    const el = findDeepestNode(ref)
+    const window = getWindow(el)
+
+    // If no selection is defined we're probably in a node environment and we can skip the selection magic.
+    if (!window || !window.getSelection) return
```
arekkas: I think the issue was `ReactDOM.renderToStaticMarkup()` if I recall correctly. I will verify the issue

src/components/leaf.js
arekkas: No, that's exactly the point of this PR, it's unfortunately not enough to just removeAllRanges, you also need to blur the contenteditable or otherwise key strokes will be processed by the onInput event in all browsers I tested

src/components/leaf.js
arekkas: Sure, I'll give it a try!

arekkas:
Thanks for reviewing @ianstormtaylor , I've added some comments and will work on the things you mentioned today.

src/components/leaf.js
arekkas: Nevermind, you're right. This code should not be reachable if window is undefined.

arekkas:
Hey @ianstormtaylor I have resolved the comments you posted. The blur magic now happens in `content.js` and the additional focus call is in `leaf.js`. As I mentioned before, blur and removeAllRanges are required to properly set the DOM.activeElement. I think this should be good to merge now :)

arekkas:
Oh and I tested a couple of cases by hand, they all seem to work.

src/components/leaf.js
ianstormtaylor: Gotcha, sounds good! Thanks for figuring that out

arekkas:
By the way, travis failed because of a timeout which apparently happens randomly :)

ianstormtaylor:
Awesome, thank you @arekkas!

I've added a `focus-blur` example that will hopefully be useful to continue ensuring the right behavior, until we figure out better tests for the DOM-level stuff

arekkas:
Awesome, thanks for the merge and the example!

PR #421, opened 2016-11-14
Do not process onInput events when selection is lost

arekkas:
Resolves an issue where `this.getPoint(anchorNode, anchorOffset)` throws an error if a key event is triggered although not in focus.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -418,6 +418,7 @@ class Content extends React.Component {
   onInput = (e) => {
     if (this.tmp.isRendering) return
     if (this.tmp.isComposing) return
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Ah interesting, thank you!

Weird that `onInput` can even be triggered when not in focus... might be some underlying issue we're not accounting for. Even if so, this sounds good in the meantime!

arekkas:
Cool, thanks for the merge @ianstormtaylor . I don't believe that there's an underlying issue here, the problem is simply that contenteditable events propagate quite quickly, while render loops with redux and similar jazz are not that fast. I think the root cause is how react handles events, as well.

PR #429, opened 2016-11-17
Upgrade debug package

aehlke:
Fixes an issue where debug may clear localStorage.debug on start, making it impossible to use. See https://github.com/visionmedia/debug/issues/330


aehlke:
Gr, sorry, screwed this one up - they haven't cut a release of `debug` yet with the fix. Will follow up once they do...

PR #431, opened 2016-11-17
resolve issue with line breaks before anchor nodes - closes #430

### LINKED ISSUE #430: Inserting new-line before anchor renders two inlines
JSFiddle: https://jsfiddle.net/aeneas/te8ow3nh/2/

To reprodue use `shift+enter` somewhere in the text, and then just before the link as you can see in this gif:

broken-newline

I checked the transform, it's not the culprit. Also it isn't solved with #396 unfortunately. Probably a bug in the leaf or something?

arekkas:
Because white-space: pre-wrap is set on the contenteditable root, the newline check is not needed any more.

### DIFF
--- a/src/components/leaf.js
+++ b/src/components/leaf.js
@@ -267,12 +267,6 @@ class Leaf extends React.Component {
     // selection can be inserted next to it still.
     if (text == '') return <span className="slate-zero-width-space">{'\u200B'}</span>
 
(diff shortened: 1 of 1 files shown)


arekkas:
Resolved by https://github.com/ianstormtaylor/slate/commit/a2be745bede67060be31d9a3e93b925d89b32f23

PR #435, opened 2016-11-18
Added test cases fore insert-newline and split-block

arekkas:


ianstormtaylor:
Hey @arekkas, could you change the `insert-newline/before-mark` to `insert-text/newline-before-mark`, so that it's in one of the transform folders. (Since `insertNewline` isn't a specific transform.) Thanks!

ianstormtaylor:
Actually, I ended up adding a `split-block/before-inline` one when refactoring your other PR, so we don't need that one anymore. And I think the newline one isn't strictly needed, since it's just the same as any text-inserting text. But if you still want it, if you refactor this or open a new PR with just that one I can merge. Closing otherwise!

arekkas:
I just wanted to add another failing test case and use the code from here but it doesn't seem you merged it to master? At least my code-search can't find it

ianstormtaylor:
Not sure what you mean. I didn't merge this pill request, nope.

arekkas:
ok, nevermind I might have mixed up something

Issue #432, opened 2016-11-17
Links break selection


arekkas:
Go to http://slatejs.org/#/links?_k=kigr00 and place the selection before `hyperlinks`, then press enter, delete, enter and start typing. You should see something like this:

![slate-link](https://cloud.githubusercontent.com/assets/3372410/20407594/500be6ca-ad12-11e6-91b5-a8befc502b9b.gif)

ianstormtaylor:
Fixed now, thanks for the test case!

PR #436, opened 2016-11-18
Added failing test case for inlines causing selection issue #432

arekkas:
I'm not sure if this is really the cause for #432 but it might, so I gave it a shot.


ianstormtaylor:
Awesome, thank you! I simplified it a bit and added a fix.

arekkas:
Cool thanks!

Von meinem iPhone gesendet

PR #439, opened 2016-11-19
added breaking tests for inline elements with leading/trailing newline

arekkas:
I observed some issues with newlines and inline nodes, so I wrote tests for it. Some of them are failing, and they reflect what I see in my editor.

### DIFF
--- a/test/transforms/fixtures/at-current-range/delete-backward/inside-inline-with-trailing-newline/index.js
+++ b/test/transforms/fixtures/at-current-range/delete-backward/inside-inline-with-trailing-newline/index.js
@@ -1,0 +1,28 @@
+
+import assert from 'assert'
+
+export default function (state) {
+  const { document, selection } = state
+  const texts = document.getTexts()
+  const first = texts.get(1)
+  const range = selection.merge({
+    anchorKey: first.key,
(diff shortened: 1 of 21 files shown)


test/transforms/fixtures/at-current-range/delete-backward/inside-inline-with-trailing-newline/index.js
```diff
@@ -0,0 +1,28 @@
...
+  const next = state
+    .transform()
+    .moveTo(range)
+    .deleteBackward()
+    .apply()
+
+  // Not sure what this should look like
+  // assert.deepEqual(
+  //   next.selection.toJS(),
+  //   range.toJS()
```
arekkas: Either the selection has a bug here too, or I made a mistake. Please let me know what should be here

ianstormtaylor:
Hey @arekkas I'm sorry for this but I'm going to close this just because the PR is so old. Back in the day I was pretty good about merging and fixing failing test cases like this, but these days I really need folks to pre-fix the test case with the PR I think for me to have the time. If you want to do that in a new PR that would be awesome! Thanks for understanding 😄

PR #443, opened 2016-11-21
add support for nested content-editables

tyler-johnson:
I am attempting to embed a Slate editor inside a void node inside another Slate editor. In v0.14 this actually worked alright, but in v0.15 the events started bubbling up improperly. This is probably a more extreme use case for Slate, however adding support was very straight forward and on the surface this appears to work consistently.

The original code was just checking that bubbled events happen inside of a content-editable node, not specifically the one created by the content node. My changes add a React ref for the root content-editable and then all events make sure the event bubbled from the correct element.

I wanted to add tests for this, however Slate appears to be missing integration tests currently. I am happy to add tests for this if/when those are in place.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -149,19 +149,22 @@ class Content extends React.Component {
    * @param {Element} n
    */
 
-  ref = (n) => this.element = n
+  ref = (element) => {
+    this.element = element
+  }
 
   /**
-   * Check if an `event` is being fired from inside a non-contentediable child
-   * element, in which case we'll want to ignore it.
+   * Check if an `event` is being fired from within the contenteditable element.
+   * This will return false for edits happening in non-contenteditable children,
+   * such as void nodes and other nested Slate editors.
    *
    * @param {Event} event
    * @return {Boolean}
    */
 
-  isNonEditable = (event) => {
+  isInContentEditable = (event) => {
     const { target } = event
-    return !target.isContentEditable || target !== this.element
+    return target.isContentEditable && target === this.element
   }
 
   /**
@@ -172,7 +175,7 @@ class Content extends React.Component {
 
   onBeforeInput = (e) => {
     if (this.props.readOnly) return
-    if (this.isNonEditable(e)) return
+    if (!this.isInContentEditable(e)) return
 
     const data = {}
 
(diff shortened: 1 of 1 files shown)


src/components/content.js
```diff
@@ -144,14 +144,35 @@ class Content extends React.Component {
   }
 
   /**
+   * The React ref method to set the root content element locally.
+   *
+   * @param {Element} n
+   */
+
+  ref = (n) => this.element = n
```
ianstormtaylor: Can you change this to be multiple lines, eg.

```js
ref = (element) => {
  this.element = element
}
```

Thanks!

src/components/content.js
```diff
@@ -144,14 +144,35 @@ class Content extends React.Component {
...
+
+  /**
+   * Check if an `event` is being fired from inside a non-contentediable child
+   * element, in which case we'll want to ignore it.
```
ianstormtaylor: Can you update this comment to reflect the new heuristic it's using too?

src/components/content.js
```diff
@@ -144,14 +144,35 @@ class Content extends React.Component {
...
+   *
+   * @param {Event} event
+   * @return {Boolean}
+   */
+
+  isNonEditable = (event) => {
```
ianstormtaylor: This is a slightly bigger ask, but can you change this method to be called something more clear and positive (like `isInContentEditable`) so that the logic isn't double-negative anymore which is hard to read?

This was my fault for setting it up that way to begin with, but as we add more logic it's just getting less understandable haha

src/components/content.js
```diff
@@ -144,14 +144,35 @@ class Content extends React.Component {
...
+    const { target } = event
+    return !target.isContentEditable || target !== this.element
```
ianstormtaylor: Does this need to check for whether the target element is **inside** `this.element`? Or is it only ever going to fire with the `contenteditable` element as the target. Asking because I can't remember, but just want to double check we're not leaving a gap here.

tyler-johnson: Yes, I thought about this too. My understanding is that this should only be used on events which are only triggered on input elements, including content-editables. The events which use `isInContentEditable()` are: `onBeforeInput`, `onBlur`, `onCompositionEnd`, `onCompositionStart`, `onCopy`, `onCut`, `onDragEnd`, `onDragOver`, `onDragStart`, `onDrop`, `onInput`, `onKeyDown`, `onKeyUp`, `onPaste`, `onSelect`.

Most of those will only target the input element, the only ones that may not be the case is the onDrag/Drop events? So perhaps we should be traversing the target's parent instead of only checking the target. It will definitely be safer.

ianstormtaylor:
Thanks!

PR #450, opened 2016-11-23
add the option to normalize the state on creation, defaulting to `false`

ianstormtaylor:


Soreine:
I'm thinking about rules that surround inline void nodes with empty texts for example. As long as these empty texts are kept after serializing to Raw, this option can be useful. More generally, as long as the Raw serializer can keep the state normalized, this option is useful.

ianstormtaylor:
Fixed with https://github.com/ianstormtaylor/slate/commit/eb4854006a3c6e4cda1a4328d770d455eac2c44c by doing the inverse, defaulting to `true` instead as the current behavior.

PR #453, opened 2016-11-24
Avoid deprecation warnings in node.getNextBlock and node.getPreviousBlock

skogsmaskin:
node.getPreviousText and node.getNextText is called with a node parameter, and not node.key, in node.getNextBlock and node.getPreviousBlock. It triggers a deprecation warning.

### DIFF
--- a/src/models/node.js
+++ b/src/models/node.js
@@ -747,7 +747,7 @@ const Node = {
       last = block.getLastText()
     }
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Oops, sorry @skogsmaskin I accidentally forgot I had marked this as read! Thank you for this!

PR #462, opened 2016-11-30
add resetKeyGenerator util

### LINKED ISSUE #53: add server-side rendering
This might already be possible, in which case all we need to do is port the examples to be rendered server-side. But if not, while porting there might need to be a few changes made.

ianstormtaylor:
Closes #53

### DIFF
--- a/docs/reference/utils/utils.md
+++ b/docs/reference/utils/utils.md
@@ -11,6 +11,7 @@ import {
 Utility functions that ship with Slate that may be useful for certain use cases.
 
 - [`findDOMNode`](#finddomnode)
+- [`resetKeyGenerator`](#resetkeygenerator)
 - [`setKeyGenerator`](#setkeygenerator)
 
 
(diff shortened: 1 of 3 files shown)


efleurine:
HI everyone.
I use the `resetKeyGenerator` as you can see in my code below. I still have the same react warning about a checksum difference between the server and the client with major difference that now  I always get the same number for data-key.
```
(client) or="true" data-key="1" contenteditable="
(server) or="true" data-key="0" contenteditable="
```

here my full custom component code.
```
import React, { Component } from 'react'
import { resetKeyGenerator } from 'slate'
import { Editor } from 'slate-react'
import Plain from 'slate-plain-serializer'

// 1 they initialize the counter
// return the first slate value


class CustomEditor extends Component {
  lastSerializedState = undefined;
  constructor(props) {
    super(props)
    if (!props.value) {
      throw Error(
        'You must pass an initial value to the editor. You can pass "" if there is no content.'
      )
    }
    this.state = {
      editorState: Plain.deserialize(props.value),
      focus: false
    }
    this.lastSerializedState = props.value
    resetKeyGenerator()
  }

  onChange = ({ state }) => {
    this.setState({ editorState: state })
  }

  render () {
    const { editorState } = this.state;

    return (
      <Editor
        placeholder="Enter some rich text..."
        state={editorState}
        onChange={this.onChange}
      />
    )
  }
}

export default CustomEditor
```

efleurine:
While writing about it I got the idea to add a call to `resetKeyGenerator()` 
before my customEditor class and suddenly the warming disappears.

Issue #464, opened 2016-11-30
Opt-Shift-Up and Opt-Shift-Down don't follow the Mac standard


thesunny:
According to the "Mac Keyboard Shortcuts" listing here:

https://support.apple.com/en-ca/HT201236

And also confirmed in the Pages app on Mac OS X, these are expected behaviours:

**Option–Shift–Up Arrow** Extend text selection to the beginning of the current paragraph, then to the beginning of the following paragraph if pressed again.

**Option–Shift–Down Arrow** Extend text selection to the end of the current paragraph, then to the end of the following paragraph if pressed again.

In Slate, using Chrome, it does not go the beginning or end of each additional paragraph selected. Instead, it goes to a position that is based on the offset (within a block) of the anchor position. This damages the utility of using the shortcut for selecting complete blocks.

thesunny:
Just confirmed this bug also appears on Safari.

A different selection bug appears on Firefox where the selection refuses to grow passed a single block. I believe this attached fix should resolve this as well (I'm at another computer so can't verify).

Firefox also appears to have a bug on Option-Up and Option-Down movement beyond a block. Will open a new issue for this.

thesunny:
Confirmed that the pull request fixes the Firefox bug as well.

PR #465, opened 2016-11-30
Fix for OPTION-SHIFT-UP and OPTION-SHIFT-DOWN

thesunny:
This is a fix for the following issue:

https://github.com/ianstormtaylor/slate/issues/464

Reposted the details here:

According to the "Mac Keyboard Shortcuts" listing here:

https://support.apple.com/en-ca/HT201236

And also confirmed in the Pages app on Mac OS X, these are expected behaviours:

**Option–Shift–Up Arrow** Extend text selection to the beginning of the current paragraph, then to the beginning of the following paragraph if pressed again.

**Option–Shift–Down Arrow** Extend text selection to the end of the current paragraph, then to the end of the following paragraph if pressed again.

In Slate, using Chrome, it does not go the beginning or end of each additional paragraph selected. Instead, it goes to a position that is based on the offset (within a block) of the anchor position. This damages the utility of using the shortcut for selecting complete blocks.

### DIFF
--- a/src/plugins/core.js
+++ b/src/plugins/core.js
@@ -530,17 +530,17 @@ function Plugin(options = {}) {
    */
 
   function onKeyDownUp(e, data, state) {
-    if (!IS_MAC || data.isCtrl || !data.isAlt || !data.isShift) return;
-    e.preventDefault();
+    if (!IS_MAC || data.isCtrl || !data.isAlt || !data.isShift) return
+    e.preventDefault()
     const {selection, document, focusBlock} = state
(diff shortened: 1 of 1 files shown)


thesunny:
Turns out the easiest way to implement the fix for the following issue was two small changes in this pull request:

https://github.com/ianstormtaylor/slate/issues/469

ianstormtaylor:
Thanks @thesunny!

thesunny:
This was my first pull request.

Thanks for accepting.

Issue #468, opened 2016-11-30
`onKeyDown` not fired for editable elements nested inside non-editables


AlbertHilb:
In my app I use container blocks rendered as

```
<section className="definition" contentEditable={false}>
  <header>Definition.</header>
  <div className="content" contentEditable={true}>
    {props.children}
  </div>
</section>
```
Unfortunately editing their children does not trigger the 'onKeyDown' event handler.
This happens because the `content` component use the test

```
(
  (target.isContentEditable) &&
  (target === element || target.closest('[contenteditable]') == element)
)
```
to decide if an 'onKeyDown' handler must be called.
According to the comment in the source code, the test is intended to:
 1. prevent 'onKeyDown' handlers being called for non-editable elements 
     ```target.isContentEditable```
 1. ignore events coming from nested slate editors
     ```target === element || target.closest('[contenteditable]') == element```

In the second check it is assumed that the only ancestors of an element with `contenteditable` attribute are slate `content` component instances.
That assumption, of course, is wrong in my case and the test fails even if it should not.

My proposal to solve the problem is to add an attribute to slate `content`elements, let's say `data-slate-root`, to uniquely identify them, and then to rewrite the test as

```
(
  (target.isContentEditable) &&
  (target === element || target.closest('[data-slate-root]') === element)
)
 ```

If you agree I will make a PR.

ianstormtaylor:
Hey @AlbertHilb, cool use case, I totally agree with your conclusion! I'd be happy to merge that PR. Thank you!

tyler-johnson:
Just noticed this issue and seeing as I submitted #443 which changed the functionality to this, I feel like I should weigh in.

Perhaps the issue here is that we are treating all the events the same? For my nested use case, the issue wasn't that all events were bubbling, but rather that the React `onInput` event was bubbling. This was causing both Slate instances to interpret the changes resulting in undesired behavior. Maybe only the `onInput` event should use `[content-editable]` and the other events use `[data-slate-content]`?

tyler-johnson:
Actually ignore my "solution" as I see it still runs into the same issue of the nested content-editables. Still I could see there being nested content-editables that should not be a part of the Slate instance and therefore should not have their `onInput` bubble. #470 would not solve that usecase.

AlbertHilb:
Hi @tyler-johnson. :blush: 
What exactly is the use-case not fixed by #470?
Thanks!

tyler-johnson:
[...]

PR #470, opened 2016-12-01
Add an attribute to uniquely identify divs used to render `content` component

AlbertHilb:
Add `data-slate-content` attribute to the divs used to render `content` components.
Use that attribute to uniquely identify them.
Update test expected results to reflect the presence of the new attribute.

This fixes issue #468.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -167,7 +167,7 @@ class Content extends React.Component {
     const { target } = event
     return (
       (target.isContentEditable) &&
-      (target === element || target.closest('[contenteditable]') == element)
+      (target === element || target.closest('[data-slate-content]') === element)
     )
   }
 
@@ -670,6 +670,7 @@ class Content extends React.Component {
 
     return (
       <div
+        data-slate-content={true}
         key={this.forces}
         ref={this.ref}
         contentEditable={!readOnly}
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -670,7 +670,7 @@ class Content extends React.Component {
 
     return (
       <div
-        data-slate-content={true}
+        data-slate-content
         key={this.forces}
         ref={this.ref}
         contentEditable={!readOnly}
--- a/test/rendering/fixtures/custom-block-void/output.html
+++ b/test/rendering/fixtures/custom-block-void/output.html
@@ -1,5 +1,5 @@
 
-<div contenteditable="true">
+<div data-slate-content="true" contenteditable="true">
   <div style="position:relative;">
     <span style="position:absolute;top:0px;left:-9999px;text-indent:-9999px;">
       <span>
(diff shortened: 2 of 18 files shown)


ianstormtaylor:
Hey @AlbertHilb thank you for this! I like the change to `-content` instead of `-root` too 😄 

I just had a thought, to make it even more "seamless" in terms of number of concepts we're using. Instead of adding `data-slate-content="true"` I think we should add `data-key="X"` with the key of the `Document` node itself, since all other nodes keys are in the DOM already. This would keep that parallelism, and then we can check for the "root" node by using it's key, similar to how we'd check for any other node!

Are you down to make that change? Thank you!

AlbertHilb:
Hi @ianstormtaylor.

I totally agree with the principle of using the fewest possible ingredients to solve a problem. :blush:

However, I have doubts on the use of `data-key="X"` attribute in this particular case, maybe it's because I don't known deeply enough the Slate source code. :pensive:

The problem to solve here is this.
In a scenario in which there are two different instances of slate editor nested one inside the other, how to prevent a `keydown` event targeting an element in the nested editor render tree, from being handled by the other editor.
This corresponds to establish if the event target DOM element is a descendant of the nested editor element or not.

Since the slate node keys are numbered starting from 0, when there are two (or more) editors in the same page, there are collisions between node keys of different instances. For example, there are two nodes with key equal to 0 and therefore two DOM elements with attribute `date-key="0"`.
In other words, it's impossible just knowing the `data-key` attribute of a DOM element to determine which editors the element belongs to.

`findDomNode`, the function used to retrieve DOM element associated to a Slate node, uses the code

```
window.document.querySelector('[data-key="${node.key}"]')
```

hence it returns the first element in the DOM with a given `data-key="X"` attribute independently which instance is invoked from.

If the node keys were unique then we could

1. find the closest ancestor of `target` with a `data-key="X"` attribute
2. check if in the document of the slate instance that is being handling the `keydown` event there is a node with key "X"
3. if there is, continue handling the event else stop.

But, unfortunately, this in not the case.

To sum up, I think it is safer to use `data-slate-content` attribute, until the Slate internal structure does not change, and more generally to avoid at all the use of more Slate instances in the same page.

If we want to keep the possibility to use two or more editor instances per page we must change the way Slate generates node keys. The simpler way, probably, is to use a static variable in the `Editor` class, let's say `instanceCount`, that counts the number of instances created, then use `instanceCount` to generate an unique key for each editor in the page and prefix to each node keys the key of the editor it belongs to.

What is your opinion?

ianstormtaylor:
@AlbertHilb I think you make a very good point, especially relating to the recent issue we're having in https://github.com/ianstormtaylor/slate/issues/493 which will probably need this exact distinction. I'm going to figure that issue and this PR out soon!

PR #485, opened 2016-12-02
Remove lodash dependency

Soreine:
This removes `lodash` from the dependencies. I think it's for the best. If we need some utilities, we can always include them individually instead.


ianstormtaylor:
Wow I had no idea we were using it so little 😄  sweet!

There's an issue with `cheerio` btw that it loads a ton of `lodash` modules (in the latest version on npm) which causes bloat, if that's something you're after. It was fixed in `cheerio#master` but they haven't released a new version yet.

PR #488, opened 2016-12-02
Fix #484: implement wrapInlineByKey and allow wrap of inline void in wrapInline

### LINKED ISSUE #484: transform.wrapInline not working when selection is an inline void
In our application, we are using inline void for images, when the user select an image and click to insert a link, we are doing:

```js
transform
.wrapInline({
  type: MarkupIt.INLINES.LINK,
  data: {
    href:  action.href,
    title: action.title
  }
})
.collapseToEnd()
.focus()
.apply()
```

It looks like the `wrapInline` doesn't work when the selection is an inline void.

SamyPesse:
This PR fixes #484.

I think it's `wrapInlineAtRange`'s responsability to handle case where inline void are selected, even if it means a collapsed selection.

- [x] Add tests for wrapping inline voids with an inline node
- [x] Implement `wrapInlineByKey` (similar to `wrapBlockByKey`)
- [x] Document `wrapInlineByKey`
- [x] Adapt `wrapInlineAtRange` to handle the case of a selected inline void


test/transforms/fixtures/at-current-range/wrap-inline/inline-void/input.yaml
```diff
@@ -0,0 +1,9 @@
+
+nodes:
+  - kind: block
+    type: paragraph
+    nodes:
+      - kind: inline
```
ianstormtaylor: Can you add the extra empty text nodes around the inline here for clarity?

test/transforms/fixtures/at-range/wrap-inline-at-range/wrap-inline-void/index.js
```diff
@@ -0,0 +1,22 @@
+
+export default function (state) {
+  const { document, selection } = state
+  const texts = document.getTexts()
+  // The first text is the one arround the inline void
```
ianstormtaylor: And then we can remove this comment to 😄

test/transforms/fixtures/at-current-range/wrap-inline/inline-void/index.js
```diff
@@ -0,0 +1,30 @@
+
+import assert from 'assert'
+
+export default function (state) {
+  const { document, selection } = state
+  const texts = document.getTexts()
+  // The first text is the one arround the inline void
```
ianstormtaylor: And this comment too

SamyPesse:
@ianstormtaylor updated!

ianstormtaylor:
Thanks!

Issue #482, opened 2016-12-02
suggestion for non-core transform syntax


thesunny:
`slate-edit-table` is an example of a Slate plugin that exports its own transforms.

The problem is that the transform syntax is rather ugly when using non-core transforms.

Here is an example to illustrate:

```js
const {insertRow, moveSelectionBy} = slateEditTableInstance.transforms;
let transform = state.transform();

insertRow(transform);
transform.moveTo(state.selection);
moveSelectionBy(transform, 1, 0);
return transform.apply();
```

What I don't like about the syntax is that

* because core transforms are chained, it helps to show that they are all part of the same transform
* non-core transforms lose this clarity

I would like to add a method, maybe something like `call` that is just syntactic sugar for calling a non-core transform with the first argument being the `transform` object and the remaining arguments being being the remaining arguments in the call.

The code below would then be equivalent to the code above but more readable.

```js
const {insertRow, moveSelectionBy} = slateEditTableInstance.transforms;

return state.transform()
  .call(insertRow)
  .moveTo(state.selection)
  .call(moveSelectionBy, 1, 0)
  .apply();
```

What do you (a) think of the idea and (b) the name of the method `call`. I can create a pull request for this if you think this is a good idea.

ianstormtaylor:
Hey @thesunny, interesting idea. I definitely feel the pain you're talking about with not being able to chain. 

@samypesse @soreine what do you think?

I think it's an interesting solution. Another would be to be able to define transforms upfront somewhere, like on `State` object but I don't think that's flexible enough. 

Another might be to be able to pass in extra transforms when invoking `state.transform`. 

But the `.call()` approach might actually be the lest boilerplate-y that still preserves chaining. I just don't know if it's too weird or not.

thesunny:
Just to go through my entire thought process, to potentially help evaluate the proposal, my initial instinct was to extend the native transforms by adding `transforms` to the plugin architecture.

The idea would be then that you could call the methods directly off of the Transform object like this:

```js
// original idea was this
return transform.insertRow().apply();

// instead of this
return transform.call(insertRow).apply();
```

I ended up with some issues which caused me to reject this:

- Plugins may come to rely on transforms from other plugins and then you have this issue where the end-developer needs to handle dependencies themselves.
- High probability of name conflicts
- Probably the most important, some plugins instantiate with options. If you add more than one instance of a plugin and they have different options, which transform do you run? They will have the same name but different options.

[...]

PR #503, opened 2016-12-07
Added `call` to Transforms

thesunny:
This pull request enables calling of custom transforms as per the discussion in this issue:

https://github.com/ianstormtaylor/slate/issues/482

A few notes:

- The new method `call` didn't fall neatly into a pre-existing category of Transforms. As a result, I put it into a category of its own. If you prefer to create a new category or to lump it in with an existing one, let me know.
- I also created a new top level test directory for `call`. Because it was a single method on its own in a category of its own, I had one less directory structure level for the tests but otherwise kept to the standards in the other tests
- Testing was fairly simple but I did create two tests. One for `call` with no arguments and one for `call` with arguments
- Just noticed that I didn't add any documentation. After this PR is accepted, I'll add some documentation.


thesunny:
Fixing the conflicts...

thesunny:
Wow, quite a few changes in a short period of time there!

Should be conflict free now.

PR #509, opened 2016-12-08
Add unwrapNodeByKey

Soreine:
This PR adds a transform named `unwrapNodeByKey`, to unwrap a particular node from its parent.
Calling `unwrapNodeByKey` on a block is not equivalent to calling `unwrapBlockByKey` on that same block though, because the latter will unwrap the block's children, whereas the former will unwrap the block itself from its parent. I'm open to renaming that transform for that reason, but I could not find another name.

I needed that one to fix an issue with unwrapping lists https://github.com/GitbookIO/slate-edit-list/issues/13

I have added tests for all the cases covered by this transform, as well as documentation.


ianstormtaylor:
Nice, thank you @Soreine! I think I need this one too. 😄

PR #513, opened 2016-12-09
Add the `Stack` concept

ianstormtaylor:
This adds a new concept to Slate called a `Stack`, which reduces some of the work that the `<Editor>` was doing previously, specifically around managing plugins.

Previously the editor was responsible for creating a stack of plugins, and sending events through them as things were triggered. Because the plugins stack was coupled with the editor, it made it hard to test since you'd need to render the React tree to test plugins's behavior correctly—and rendering the React tree in a pseudo-browser environment is near impossible due to all of the DOM features we use.

Specifically the previous setup made it hard to test **multiple** plugins at once—their interaction with other third-party plugins and schemas and the "core" plugin. Testing a single plugin was possible [like so](https://github.com/GitbookIO/slate-edit-list/blob/master/tests/split-item-start/transform.js).

With this PR, all of that plugin-stack-specific logic is managed by a `Stack`, which knows how to send events through the list of plugins. This makes it a lot easier to test without the DOM, as you can see in the new [`behavior`](https://github.com/ianstormtaylor/slate/tree/add-stack/test/behavior/fixtures) tests!

I've also added the first stab at a [`Simulate`](https://github.com/ianstormtaylor/slate/blob/add-stack/test/behavior/fixtures/on-key-down/split-empty-block/index.js#L5) test helper, which we can extract out into a separate module for third-party plugin testing.

@SamyPesse @Soreine would love for you two to check this out and let me know if you see anything that looks off.


ianstormtaylor:
Sweet, thanks for reviewing!

Yup, we could totally figure out a way to make it opt-in for the tests. Right now I've only enabled it in the `behavior` tests, since the others will require refactoring. But it just uses the new `preserveSelection` option to the `Raw` serializer, so it wouldn't be hard to make it opt-in in cases where we don't want the duplication.

Soreine:
I started using it, and to make selection opt-in in the tests, it was quite easy:

```js
const input = readYaml(...);
const stateInput = Slate.Raw.deserialize(input, { terse: true }); // already opt-in

const stateOutput = runTest(stateInput, ...);

const expected = readYaml(...);
const output = Slate.Raw.serialize(newState, {
  terse: true,
  preserveSelection: Boolean(expected.selection) // Only preserve selection if the output.yaml specifies one
});

// ... compare `output` and `expected`
```

PR #525, opened 2016-12-17
Added documentation to transform.call

thesunny:
I added documentation to the transform.call from an earlier pull request.


ianstormtaylor:
Awesome, thanks!

PR #526, opened 2016-12-17
Added `state` as a property to Transform in the documentation

thesunny:
I noticed that in the code, it's pretty common to pull the `state` out of the `transform`.

The only way I was able to learn about this was by looking through the code and then I wasn't sure if it was a supported property. Considering that it seems to be fairly commonly used and looks like it will be officially supported, I made a pull request to add it to the documentation.

### DIFF
--- a/docs/reference/models/transform.md
+++ b/docs/reference/models/transform.md
@@ -11,6 +11,8 @@ All changes are performed through `Transform` objects, so that a history of chan
 
 Transform methods can either operate on the [`Document`](./document.md), the [`Selection`](./selection.md), or both at once.
 
+- [Properties](#properties)
+  - [`state`](#state)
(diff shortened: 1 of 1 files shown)


docs/reference/models/transform.md
```diff
@@ -87,6 +89,13 @@ Transform methods can either operate on the [`Document`](./document.md), the [`S
   - [`undo`](#undo)
 
 
+## Properties
+
+### `state`
+
+An instance of [`State`](./state.md) where the transform was created from.
```
ianstormtaylor: Can you change this to say:

> A [`State` ](./state.md) with the transforms current operations applied. Each time you run a new transform function this property will be updated.

ianstormtaylor:
Thanks @thesunny!

Issue #337, opened 2016-09-17
Expose constants for keyboard events


SamyPesse:
⚠️ This is **not** an important issue.

In all the plugins we are working on at GitBook (`slate-edit-code`, `slate-edit-list`, `slate-edit-blockquote`, etc), we are adding [these lines](https://github.com/GitbookIO/slate-edit-code/blob/master/lib/index.js#L6):

``` js
const KEY_ENTER     = 'enter';
const KEY_TAB          = 'tab';
const KEY_BACKSPACE = 'backspace';
```

It might be great if slate can expose these constants:

``` js
const Slate = require('slate');

Slate.KEY_ENTER, ...
```

thesunny:
Since it's an easy one, decided to make a pull request for it.

Will add documentation if PR accepted.

ianstormtaylor:
Hey sorry for the delay here. I'm not yet sure I want to do this, because I'm not yet sure how useful it would be compared to people who want more control using something like Flow or Typescripts. Since I'm unsure, I'm currently defaulting to not adding it, just because it adds API scope.

ianstormtaylor:
I'm going to close this for now, since I'd rather not do this I think for bloat reasons. But if someone wants to implement this as another library that works! And if it gets popular we can consider merging it into core.

PR #530, opened 2016-12-17
Added key constants that are exported

thesunny:
For this issue:

https://github.com/ianstormtaylor/slate/issues/337

### DIFF
--- a/src/constants/keys.js
+++ b/src/constants/keys.js
@@ -1,0 +1,19 @@
+/**
+ * Constants for special keys used in the `onKeyDown` handler.
+ *
+ * @type {Object}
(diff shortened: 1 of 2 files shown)


thesunny:
@ianstormtaylor just wanted to say that I put the PR in case you wanted to accept it realizing that it wasn't clear (from the issue) whether this was worthwhile or not.

ianstormtaylor:
Hey @thesunny I'm going to close this for now for cleanup reasons, because I'm not sure we want to take this on yet, and might not implement exactly the same way when/if we do. But thank you for submitting it! (And for all your other submissions!)

Issue #362, opened 2016-09-30
Include link to github repo on slatejs.org


arekkas:

thesunny:
Made a pull request here:

https://github.com/ianstormtaylor/slate/pull/531

PR #531, opened 2016-12-18
Updated examples to add GitHub and Docs link + some styling.

thesunny:
Added GitHub link to examples as per this issue:

https://github.com/ianstormtaylor/slate/issues/362

Also added a link to the Documentation.

In order to make the changes work, I updated some of the HTML/CSS. Notably:

* Added a top bar that displays "Slate Examples" + the links at the upper right.
* Added a background rounded box behind the current example. This was to distinguish the stateful tab from the non-stateful links to GitHub/Docs
* Put rounded box on tabs in hover (but not the white font colour) just to provide some nice UI hints about its tab-like nature. The white font colour is reserved for the active tab.

### DIFF
--- a/examples/index.css
+++ b/examples/index.css
@@ -86,8 +86,33 @@ input:focus {
  * App.
  */
 
+.topbar {
+  padding: 10px 15px;
+  color: #AAA;
+  background: #000;
+}
+
+.topbar-title {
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thank you! This is awesome

PR #532, opened 2016-12-19
Reloading the page now also reloads the CSS.

thesunny:
When reloading a page in `/examples/dev.html` it will now also reload the CSS.

Also updated the JavaScript from last time as I realized I wasn't using the JavaScript style guidelines in Slate.


ianstormtaylor:
Thanks @thesunny!

Issue #543, opened 2016-12-29
Add a way to propagate `readOnly` state to nested editable content blocks


AlbertHilb:
Nested editable content blocks like:
```
<section className="definition" contentEditable={false}>
  <header>Definition.</header>
  <div className="content" contentEditable={true}>
    {props.children}
  </div>
</section>
```
remain editable even when the editor enters `readOnly` state.

A natural way to solve this is rewriting the block as
```
<section className="definition" contentEditable={false}>
  <header>Definition.</header>
  <div className="content" contentEditable={!editor.props.readOnly}>
    {props.children}
  </div>
</section>
```
Unfortunately this does not work because when `readOnly` editor property changes, only the nodes containing the selection edges are updated.

A possible solution is to update all the document nodes.
Another possible solution is to add an handler to the editor (and to the plugins), let's say `onReadOnlyChange`, that can be used by components to react to `readOnly` editor state mutations.

ianstormtaylor:
Hey @AlbertHilb, I agree with you. It makes sense that `readOnly` would trigger re-renders across all nodes. Should be fairly easy to do by passing `readOnly` into the nodes themselves and tweaking `shouldComponentUpdate`. If you're down to submit a pull request for that, I'd be happy to merge it!

PR #544, opened 2016-12-31
Update nodes on `readOnly` editor property change

AlbertHilb:
Propagate `readOnly` editor property to node components.
Update nodes when `readOnly` editor property changes.

This fixes issue #543

### DIFF
--- a/src/components/node.js
+++ b/src/components/node.js
@@ -88,6 +88,8 @@ class Node extends React.Component {
    */
 
   shouldComponentUpdate = (nextProps) => {
+    // If the `readOnly` status has changed, we need to re-render in case there is
+    // any user-land logic that depends on it, like nested editable contents.
     if (nextProps.readOnly !== this.props.readOnly) return true
 
     const { Component } = this.state
--- a/src/components/node.js
+++ b/src/components/node.js
@@ -88,10 +88,6 @@ class Node extends React.Component {
    */
 
   shouldComponentUpdate = (nextProps) => {
-    // If the `readOnly` status has changed, we need to re-render in case there is
-    // any user-land logic that depends on it, like nested editable contents.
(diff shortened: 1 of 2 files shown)


src/components/node.js
```diff
@@ -87,6 +88,8 @@ class Node extends React.Component {
    */
 
   shouldComponentUpdate = (nextProps) => {
+    if (nextProps.readOnly !== this.props.readOnly) return true
```
ianstormtaylor: I think this line should go below the one a few down that says: `Component && Component.suppressShouldComponentUpdate`, since that one should be able to override all `shouldComponentUpdate` logic.

Could you add a comment above it of something like:

```js
// If the `readOnly` status has changed, we need to re-render in case there is
// any user-land logic that depends on it, like nested editors.
```

AlbertHilb: Ok for the comment.
About the down shift of the row that checks read-only status changes, I think it's not strictly needed.
The value returned by `shouldComponentUpdate` is not affected by the shift.
If `readOnly` property has changed, it is, in any case, `true`.
The same if `readOnly` property hasn't changed but the component rendering the node has `suppressShouldComponentUpdate` set to true.
Finally if `readOnly` property hasn't changed and the component does not have `suppressShouldComponentUpdate` set to true, the returned value depends, in any case, on the rows below.
Anyway, if you prefer I'll move down the row. :blush:

ianstormtaylor: I'd like to move it down in the event that someone wants to have full control and not actually force a re-render should the `readOnly` value change, since many cases would be able to optimize that re-render away if there is no nesting. I'd rather opt for that escape hatch to be given complete override power instead of guessing what they might want.

AlbertHilb: Ok!

ianstormtaylor:
Thanks @AlbertHilb!

PR #563, opened 2017-01-18
Single state for multiple editors

Mangatt:


ianstormtaylor:
@Mangatt could you put together an example test case for this "split" editor usage? It's hard to get an idea for how it works otherwise. I'm worried that there are lots of internal assumptions about the document structure that might not work across editors, although I've just thought about it by looking at the code and not actually running it so far.

How does it handle things like pressing <kbd>Backspace</kbd> at the start of an editor that isn't actually at the start of the document. Usually Slate would do nothing, but in this case will it think there is "room to delete" and delete past the editor itself?

Trying to wrap my head around what the failure cases might be.

arekkas:
We run multiple slate instances by connecting them via redux. I don't think this should be handled by Slate itself.

lxcid:
Seconding @arekkas use case. We use it similarly and felt this shouldn't be handle by slate too.

Mangatt:
@arekkas @lxcid That was the way that we used as well. But codebase became much more complicated and less efficient. A lot of things that Slate have figured out you have to do again:

- history
- moving nodes
- keeping focus / blurring

And for no good reason imo. React DND is good example - provides context and then do with it whatever you want, not forcing you to use his internal rendering logic or assumptions about structure. Slate could do that as well, this PR demonstrate exactly that. And with minimal changes to codebase.

@ianstormtaylor You are right, when I press backspace at beginning of NodeEditor, behavior is strange or buggy. But since that can be handled by quite simple plugin, I guess that it would be that hard to add to core.

Despite that PR I guess that it could be best way to not doing (every)NodeEditor, but DocumentNodeEditor instead. That would minimeze changes of internal logic to simple check if editor has children, work with document nodes in same way as you normally handle document node.

Mangatt:
Just thought - there are probably a lot of people, that are connecting multiple editors through redux since they can cooperate. Despite that this is definitely less performant, we all are probably writing similar code that is mostly included already in Slate code.

Even though there are voices against this approach, this might be undocumented feature for time beeing and we could make this work eventually. Allowing Editor to have children and setting up context (or just returning renderNode function if you are not comfortable with context) are not that big deal in core code, but allows much more flexible use cases. And problems could solve custom plugins.

Mangatt:
One more thought. If you don't wanna add this functionality to the core and deal with the context, best way would IMO be to add render event to the plugin options.

```javascript
render(state, editor, renderNode)
````

That way there's no need to modify any current functionality. This is in line with my other opinion about elevating plugins capabilities, when some plugin wants to change way how Editor works, it would be his responsibility to make sure that it will work.

It should work as other events, only when any plugin does not sent rendered content, it will be handled by core.

Mangatt:
Is there any new development on that? Little summary of proposed solutions:

### ```Context``` and ```NodeEditor``` components
- pros: backward compatibility, great flexibility
- cons: needs to handle delete differently at beginning of NodeEditor

### Allow children nodes and function of Editor / add render method for plugins
- pros: minimal changes to codebase, backward compatibility
- cons: none that I think of for Slate itself, a lot of work to do for developer

### Shared state between editors - defining used slice
- pros: easy to use, quite possibly widespread usage
- cons: I guess that a lot of changes in Slate codebase

### Keep Slate as is
- pros: no changes to codebase at all
- cons: document structure is rigid
- cons for working multiple cooperating editors: a lot of Slate code has to be written again in app itself, history is unusable, performance suffers and managing focus is not that simple

I my opinion the best way to go is with Context, Slate could possibly benefit heavily from that approach in future. Every node could instead of passing props down read from context, updating Nodes could be simpler when parent component implements shouldComponentUpdate.

In case that you aren't willing to change code that much, my vote goes for allowing children of editor, it means changing 3 lines of code. That could be good first step.

ianstormtaylor:
Closing, since https://github.com/ianstormtaylor/slate/pull/621 is newer for the same issue.

PR #569, opened 2017-01-26
Add significant whitespace to fix headers on npm

TehShrike:
Right now the `###`-prepended lines aren't being converted to header tags when displayed on npm.


ianstormtaylor:
Thanks @TehShrike!

PR #571, opened 2017-01-27
Removing unnecessary word

TehShrike:


ianstormtaylor:
Thank you @TehShrike!

PR #572, opened 2017-01-28
Add data property to `document`.

AlbertHilb:
Add data property to `document`. It is useful, for example, to store document level metadata: creation date, author, keywords, etc.

Of course, I'm open to any change you think needed.

### DIFF
--- a/src/transforms/apply-operation.js
+++ b/src/transforms/apply-operation.js
@@ -32,7 +32,8 @@ const OPERATIONS = {
   set_node: setNode,
   split_node: splitNode,
   // Selection operations.
-  set_selection: setSelection
+  set_selection: setSelection,
+  set_document_data: setDocumentData
 }
 
 /**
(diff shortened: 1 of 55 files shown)


src/transforms/apply-operation.js
```diff
@@ -32,7 +32,8 @@ const OPERATIONS = {
   set_node: setNode,
   split_node: splitNode,
   // Selection operations.
-  set_selection: setSelection
+  set_selection: setSelection,
+  set_document_data: setDocumentData
```
ianstormtaylor: Does the `set_node` operation now work here already for the document? I think it should, or we should make it work.

AlbertHilb: No, it does not! But it can be made working with little effort.

AlbertHilb:
Now `setNode` works also with document node and `setDocumentDataOperation` uses it internally.
Or does you prefer to eliminate `setDocumentDataOperation` too?

AlbertHilb:
With the last commit also `setNodeByKey` does work.

ianstormtaylor:
@AlbertHilb awesome! Yeah let's get rid of `setDocumentDataOperation` and then I think we're good to merge! Thank you!

AlbertHilb:
I removed `setDocumentDataOperation`. :blush:

ianstormtaylor:
Awesome, thank you for this @AlbertHilb!

PR #577, opened 2017-02-01
Fix deleteCharBackwardAtRange

kib357:
When we have nested inlines in startBlock, startOffset points to the wrong character. So, i think that using startText will be correct here.


kib357:
You can test by trying to delete some 2-byte emoji after inline sibling

ianstormtaylor:
@kib357 thank you for this! Sorry for the delay. When investigating this change I realized that I had completely incorrectly implemented those `delete{Char,Word,Line}BackwardAtRange` transforms in the first place, so I made a few more changes to get it working. But delete around/across emojis in inlines should work now! Thank you.

PR #580, opened 2017-02-03
F/upstream inline selection

sashless:
I am trying to fix here a problem which occurs when i click into a `inline` node having `isVoid: true`.

My plugin gets a selection in `onSelect` which is located somewhere before my inline node instead of afterwards.

In my manual testing this didnt broke anything. Ideas ?

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -659,7 +659,7 @@ class Content extends React.Component {
       const anchorInline = document.getClosestInline(anchor.key)
       const focusInline = document.getClosestInline(focus.key)
 
-      if (anchorInline && !anchorInline.isVoid && anchor.offset == anchorText.length) {
+      if (anchorInline && anchorInline.isVoid && anchor.offset == anchorText.length) {
         const block = document.getClosestBlock(anchor.key)
         const next = block.getNextText(anchor.key)
         if (next) {
@@ -668,7 +668,7 @@ class Content extends React.Component {
         }
       }
 
-      if (focusInline && !focusInline.isVoid && focus.offset == focusText.length) {
+      if (focusInline && focusInline.isVoid && focus.offset == focusText.length) {
         const block = document.getClosestBlock(focus.key)
         const next = block.getNextText(focus.key)
         if (next) {


src/components/content.js
```diff
@@ -659,7 +659,7 @@ class Content extends React.Component {
       const anchorInline = document.getClosestInline(anchor.key)
       const focusInline = document.getClosestInline(focus.key)
 
-      if (anchorInline && !anchorInline.isVoid && anchor.offset == anchorText.length) {
+      if (anchorInline && anchorInline.isVoid && anchor.offset == anchorText.length) {
```
sashless: Nah, i was wrong. In some cases it is not working like expected.

What i would like to fix here is to place the selection __directly__ after the inline node and not after the next text node

ianstormtaylor:
Hey @sashless can you upload a GIF of the problem behavior? I'm not sure I understand what you're going for. Shouldn't it be placing the selection inside the inline void node instead of before or after it?

sashless:
I created a file to show the problem @ianstormtaylor 

https://jsfiddle.net/less/8uuq8c14/5/

When you click on the text an inlide void node will be created at the end of the anchorKey node.

You can now check the console, it is showing key 0

When you click now on the inline void now nothing will be created because the anchorKey is somewhere in node with key 6 .

ianstormtaylor:
Hey @sashless thanks for the fiddle. I still don't quite understand. The pull requested code looks like it just changes a boolean flip, which is going to break the existing logic? I might be missing something though.

ianstormtaylor:
Hey @sashless, I'm going to close out this PR because it's been open for a while now without changes. This is purely because there's lots of things to keep track of in Slate and I'm trying to reduce the maintenance burden. If you think this still applies and should be re-opened or re-written, feel free to respond and we can discuss! Thanks for understanding.

sashless:
hey @ianstormtaylor, sorry for my late reply. Re the PR. Yea it is just a boolean flip which is going to break the exisiting logic. It was just an attempt to achieve an expected behavior in regards to the fiddle. I did not expect to get this merged though :) But anyway, no worries we already handled that. Thank you !

PR #587, opened 2017-02-07
Set the `Void` `Inline` wrapper css `display` property to `inline-block`

AlbertHilb:
In Chrome, when it is positioned right before a `Void ` `Inline` node, the caret is not visible.
That can be verified, for example, by using the [Emojiis example](http://slatejs.org/#/emojis).
This PR fixes that issue setting to `inline-block` the css `display` property of the `Void` `Inline` wrappers.

As always I'm open to any change you think needed. :blush:

### DIFF
--- a/src/components/void.js
+++ b/src/components/void.js
@@ -86,11 +86,20 @@ class Void extends React.Component {
   render = () => {
     const { props } = this
     const { children, node } = props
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thank you @AlbertHilb this is awesome!

Issue #473, opened 2016-12-01
Some performance tests are broken


kommen:
The following per tests error with ```Current: Error: You must pass a `schema` object.``` for me right now on `$ npm run perf`:

- normalize-document-deep
- normalize-document-normal
- normalize-document-twice
- normalize-document-wide

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/675

PR #591, opened 2017-02-09
export SCHEMA to run normalize properly

sashless:
adressing https://github.com/ianstormtaylor/slate/issues/473

I took the easy way and just export it. 
A better solution might be to provide it magically to normalize ?

### DIFF
--- a/perf/benchmarks/normalize-document-deep/index.js
+++ b/perf/benchmarks/normalize-document-deep/index.js
@@ -1,6 +1,6 @@
 
 module.exports = {
-  run(state) {
-    return state.transform({ normalized: false }).normalize().apply()
+  run(state, schema) {
+    return state.transform({ normalized: false }).normalize(schema).apply()
   }
 }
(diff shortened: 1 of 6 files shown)


ianstormtaylor:
Hey @sashless thanks for working on this! I think for now, instead of exporting it for public use, we can just require it by its path for the tests here. That way we don't need to commit to an extra export forever. Are you down to make that change?

sashless:
yea i will push that later but the problem how to use normalize without the `SCHEMA` still persists in that case

Soreine:
Having no way to normalize using the core schema is a problem for us too. Maybe the `normalize` transforms should default to using the core schema?

Soreine:
I am using a workaround to normalize a whole document

```
// Normalize the document
const normalizedDocument = Slate.State.create({
        document
    }, { normalize: true })
    .document;

ianstormtaylor:
What do you need to normalize with the core schema for? Just curious. 

I'm definitely down to default the normalize functions to use the core schema if you wanna PR! Thanks

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/675

PR #592, opened 2017-02-10
Fix case in spellCheck

kennethtruong:


ianstormtaylor:
Ah, thanks @kennethtruong!

PR #594, opened 2017-02-10
Fix deleteAtRange with nested blocks

Soreine:
Continuation of https://github.com/ianstormtaylor/slate/pull/524

This fixes deleteAtRange, when the end block's highest ancestor only have a single child, but some of its ancestor have several childs.


ianstormtaylor:
Awesome, thank you @Soreine!

PR #595, opened 2017-02-10
Allow setting a defaultBlockType as object for HTML serializer

skogsmaskin:
We need to be able to set the data on the defaultBlockType.

### DIFF
--- a/src/serializers/html.js
+++ b/src/serializers/html.js
@@ -70,6 +70,7 @@ class Html {
    * @param {Object} options
    *   @property {Array} rules
    *   @property {String} defaultBlockType
+   *   @property {String|Object} defaultBlockType
    */
 
   constructor(options = {}) {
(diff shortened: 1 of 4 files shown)


skogsmaskin:
@ianstormtaylor The idea is to have the Object|String type constructor for the node as supported elsewhere in Slate.

skogsmaskin:
When adding the test, I realised that the parameter name is ``defaultBlockType`` (indicating string). It should now maybe be more accurately named  ``defaultBlock``. But that would be a breaking change.

src/serializers/html.js
```diff
@@ -108,10 +109,21 @@ class Html {
         return memo
       }
 
-      const block = {
-        kind: 'block',
-        type: this.defaultBlockType,
-        nodes: [node]
+      let block
+      if (this.defaultBlockType.constructor.name === 'String') {
```
ianstormtaylor: Can you change this to use `typeof === 'string'` instead of `.constructor.name === 'String'`? Thanks

src/serializers/html.js
```diff
@@ -108,10 +109,21 @@ class Html {
...
+        block = {
+          kind: 'block',
+          type: this.defaultBlockType,
+          nodes: [node]
+        }
+      } else {
+        block = Object.assign(
```
ianstormtaylor: Can you change this to use the object spread syntax (`{ ...this.defaultBlockType }`) instead? Thanks

skogsmaskin:
Thanks for the feedback. I've changed the code in regard of your suggestions, and made it even more compact.

ianstormtaylor:
Awesome, thanks!

Issue #540, opened 2016-12-28
Predictive text input on iOS


julienp:
Slate seem to work really well on iOS apart from predictive text input (the suggested words above the keyboard). When tapping on one of the suggestions, the text is simply appended instead of replacing the current word.

The event triggered by tapping on the suggestions is first an event of type `textInput` with the word in `data` followed by an event of type `input`. This is then followed by another `textInput` with a single space character in `data` and finally another `input` event. 

So far I've been unable to determine from the information in the events what data should be removed before inserting the suggestion. Unfortunately you can't either have a simple rule like deleting backwards to the first space as you could backspace over the previous space, which doesn't change the suggestions offered by the keyboard.

A workaround is to bail out early in onBeforeInput: https://github.com/julienp/slate/commit/85886812e4789ce5dcf529bb5834b4339d76786e

Of course this skips the 'native' check in core.js onBeforeInput and forces a rerender. The fix doesn't feel quite right to me, maybe it should go in core.js instead. I also haven't exhaustively verified if the condition doesn't have any side effects in other browsers, but it seems to not have an adverse effects in Safari, Chrome or Firefox.

I was wondering if anyone had a better idea or any input on dealing with this, or maybe any knowledge on when textInput events happen with data that's more than a single character.

ianstormtaylor:
Hey @julienp thanks for opening this! Good research. I haven't personally investigate iOS support yet, so I don't really have much to add, but hopefully some others can chime in. If you come up with a decision on what we should do to make this work without adverse effects let me know and I can help get a pull request merge.

lxcid:
I on another attempt to try to resolve this issue and I think I might find something of interest.

It seems that when iOS autocorrected you, the selection will wrap around the text. (`isCollapsed == false`)

If you try to retrieve the DOM selection during `onBeforeInput`, you be able to query it.

My hypothesis on why predictive text is duplicating is because the selection is not updated to overwrite the new changes.

I have to sign off for the day but I'll try to research more tomorrow to confirm and see if I can come up with a fix.

lxcid:
An update on my findings. I realised that `Slate.Contents.onInput` already catered for selection before updating the text.

So I dig deeper and found out that selection are changed between `Slate.Contents.onBeforeInput` and `Slate.Contents.onInput`.

It seems that we need to update the selection during `onBeforeInput` and make changes to the text after that.

I'm still trying to study the `onInput` on slate and see what I can do.

[...]

PR #596, opened 2017-02-10
Add props “autoCorrect” to editor and content.

lxcid:
This allow user to turn off predictive text input on iOS thus mitigating #540…


ianstormtaylor:
Awesome thank you @lxcid!

PR #600, opened 2017-02-13
Eliminate duplicates in getBlocksAtRange

Soreine:
Looking at `getBlocks` implementation confirma my thoughts that `getBlocksAtRange` is broken and can return the same block multiple times.


ianstormtaylor:
Thanks @Soreine!

PR #605, opened 2017-02-15
To hide caret inside `Void` `Inline` nodes, set spacer CSS `color` property to `transparent`

AlbertHilb:
Trying to hide text cursor inside `Void` `Inline` nodes, commit a836a347861d7f0d3cb452bfe0f5b6e82cf04e31 has introduced some weird bugs in caret movement in Chrome. To understand what I mean go to [emoji example](http://slatejs.org/#/emojis), place the cursor on the first line of text (even far from emoji) then press down arrow key. The cursor should jump on the second line, but it doesn't.
After some investigations, I found the best way to hide the caret avoiding that sort of problems is to make it invisible setting to `transparent` the spacer CSS `color` property, instead of moving the spacer out of screen.
This PR implements that idea.
Among other things, to avoid graphics glitches, in Chrome, it replaces the zero-width space inside spacer with an hair space.

As always l'm open to any change you think needed. :blush:

### DIFF
--- a/src/components/leaf.js
+++ b/src/components/leaf.js
@@ -264,7 +264,12 @@ class Leaf extends React.Component {
     // COMPAT: If the text is empty otherwise, it's because it's on the edge of
     // an inline void node, so we render a zero-width space so that the
     // selection can be inserted next to it still.
-    if (text == '') return <span data-slate-zero-width>{'\u200B'}</span>
+    if (text == '') {
+      // COMPAT: In Chrome, zero-width space produces graphics glitches, so use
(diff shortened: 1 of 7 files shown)


ianstormtaylor:
@AlbertHilb this is genius! I completely overlooked that setting `color: transparent` would hide the cursor, I really like this idea. I'm on my way out to a camping trip for the long weekend, but I'm hoping to merge this when I get back on Tuesday. Thank you!!

ianstormtaylor:
Nevermind, I had a few extra minutes! Thank you!

PR #607, opened 2017-02-15
Fix jumpy scroll behaviour

### LINKED ISSUE #279: Code Highlighting Example has jumpy scroll behaviour
Scrolling goes all over the place once the editor height exceeds the viewport height.

Steps to reproduce:
1. Click into the Code Block (anywhere will do)
2. Hold Enter to insert _heaps_ of blank lines
3. After enough line breaks are added, the scrolling position jumps with every added/removed line break

<div align="center"></div>

Mainly raising this because I'm implementing my own code highlighting block and I've got the same issue.

### LINKED ISSUE #589: Scroll jumping when editor scrollable with soft brakes
It's reproduced when height of block with soft-breakes is more than height of container.
https://monosnap.com/file/uipuOSUQPmX44kGF7SKOThNOMMb7sm

I fixed this issue by using selection range rectangle in `scroll-to.js`:
```
function scrollWrapperElements(element, options) {
  let window = getWindow(element)
  const sel = window.getSelection()
  let elementRect = sel.getRangeAt(sel.rangeCount - 1).cloneRange().getBoundingClientRect()
  if (elementRect.top === 0 && elementRect.bottom === 0) {
    return null
  }
```
but it seems to be empty in some cases, so i don't process scrolling when selection returns empty rectangle, and it's not nice.

kennethtruong:
I don't think this is the best fix for the issue (I would have liked it to allow the browser to deal with moving the cursor into view.) but for now I think this is better than what it was before. I tried to imitate what chrome was doing when the cursor was out of view. (Except when it was partially in view but I decided to just keep it simple) 

This basically checks the cursor's position and if it isn't in view it'll move the cursor into the middle of the view. 

Fixes #279
Fixes #589

### DIFF
--- a/src/utils/scroll-to.js
+++ b/src/utils/scroll-to.js
@@ -1,67 +1,58 @@
 
 /**
- * Copied from `element-scroll-to`, but edited to account for `window` being
- * inside of iframes.
- *
- * https://github.com/webmodules/element-scroll-to
+ * Helps scroll the cursor into the middle of view if it isn't in view
  */
 
 import getWindow from 'get-window'
 
-function scrollWrapperElements(element, options) {
-  let window = getWindow(element)
-  let elementRect = element.getBoundingClientRect()
-  let wrapper = element.parentNode
-
-  while (wrapper != window.document.body) {
-    let wrapperRect = wrapper.getBoundingClientRect()
-    let margin = options.margin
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thank you @kennethtruong this is awesome! Sorry for the delay on merging.

Issue #597, opened 2017-02-11
Using selection transform methods lead to warning


oyeanuj:
Hi @ianstormtaylor, I ran into this warning while using `collapseToStartOfNextBlock` on `onKeyDown`.

```
Warning: An object was passed to a Node method instead of a `key` string. This was previously supported, but is being deprecated because it can have a negative impact on performance. The object in question was: Block {key: "1", nodes: List[1], type: "big-heading"}

warn	@	warn.js:32
key	@	normalize.js:95
getDescendant	@	node.js:535
assertDescendant	@	node.js:115
getNextBlock	@	node.js:852
object.(anonymous function)	@	memoize.js:109
collapseToStartOfNextBlock	@	on-selection.js:137
Transform.(anonymous function)	@	transform.js:134
onKeyDown	@	HeadingPlugin.js:36
```

It doesn't show up in production version of Slate so I wasn't able to reproduce on JSFiddle with the same code. But here is that code anyway, if it helps - https://jsfiddle.net/cxcx81jh/1/

Thank you!

oyeanuj:
Also, seeing this warning in the [Auto-Markdown](http://slatejs.org/#/auto-markdown) example.

ianstormtaylor:
Should be fixed now by #610!

Issue #598, opened 2017-02-12
Selection focus warning on auto-markdown example


oyeanuj:
On the[ Auto-Markdown example](http://slatejs.org/#/auto-markdown), seeing the following warning being invoked [here](https://github.com/ianstormtaylor/slate/blob/master/examples/auto-markdown/index.js#L140) on the use of `extendToStartOf` - 

```
Warning: The selection focus was set to a Node that is not a Text node. This should not happen and can degrade performance. The node in question was: Block {key: "12", nodes: List[1], type: "heading-one"}
warn @ build.prod.js:22519
normalize @ build.prod.js:11745
(anonymous) @ build.prod.js:20485
Transform.(anonymous function) @ build.prod.js:13942
AutoMarkdown._this.onSpace @ build.prod.js:185
```

The functionality works, but is there a better way of doing this instead?

ianstormtaylor:
Fixed now!

PR #610, opened 2017-02-17
Various fixes for selections

skogsmaskin:
Use key, not node, with ``document.getNextBlock``  and ``document.getPreviousBlock``to avoid deprecation warnings.

Also, in the core plugin onCutCopy, one must check for a selection with ``isCollapsed``, because ``rangeCount`` can return 1 even though nothing is selected.


oyeanuj:
@skogsmaskin this also might fix #597 and #598 as well. thank you!

ianstormtaylor:
Awesome, thank you @skogsmaskin!!

PR #618, opened 2017-02-23
Fix behaviour where deleteBackward/Forward on empty non-void block deleted the void block above/below

skogsmaskin:
The expected behaviour should be to delete the current empty non-void block instead.

### DIFF
--- a/src/transforms/at-range.js
+++ b/src/transforms/at-range.js
@@ -204,9 +204,14 @@ export function deleteBackwardAtRange(transform, range, n = 1, options = {}) {
     return
   }
 
-  // If the closest block is void, delete it.
   const block = document.getClosestBlock(startKey)
+  // If the closest block is void, delete it.
   if (block && block.isVoid) {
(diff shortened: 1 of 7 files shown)


src/transforms/at-range.js
```diff
@@ -204,12 +204,17 @@ export function deleteBackwardAtRange(transform, range, n = 1, options = {}) {
...
+  // If the closest block is void, delete it.
   if (block && block.isVoid) {
     transform.removeNodeByKey(block.key, { normalize })
     return
   }
+  // If the closest is not void, but empty, remove it
+  if (block && !block.isVoid && block.isEmpty) {
+    transform.removeNodeByKey(block.key, { normalize })
+    return
+  }
```
ianstormtaylor: Can you add this same (but reversed, for "next") logic to the `deleteForward` transform as well with a test? Thanks!!

skogsmaskin: Done!

ianstormtaylor:
Thank you @skogsmaskin this is awesome!

PR #621, opened 2017-02-23
Allows children of Editor

Mangatt:
Minimal changes to code that allows custom rendering of Editor contents that are currently impossible.

- full backwards compatibility
- supports both elements and functions

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -698,9 +698,9 @@ class Content extends React.Component {
 
   render = () => {
     const { props } = this
-    const { className, readOnly, state, tabIndex, role } = props
+    const { className, readOnly, state, tabIndex, role} = props
     const { document } = state
-    const children = document.nodes
+    const children = props.children || document.nodes
       .map(node => this.renderNode(node))
       .toArray()
 
@@ -755,7 +755,7 @@ class Content extends React.Component {
         role={readOnly ? null : (role || 'textbox')}
         tabIndex={tabIndex}
       >
-        {children}
+        {typeof children == 'function' ? children(state) : children}
       </div>
     )
   }
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -698,7 +698,7 @@ class Content extends React.Component {
 
   render = () => {
     const { props } = this
-    const { className, readOnly, state, tabIndex, role} = props
+    const { className, readOnly, state, tabIndex, role } = props
     const { document } = state
     const children = props.children || document.nodes
       .map(node => this.renderNode(node))
--- a/src/components/editor.js
+++ b/src/components/editor.js
@@ -260,6 +260,7 @@ class Editor extends React.Component {
         style={props.style}
         tabIndex={props.tabIndex}
         role={props.role}
+        children={props.children}     
       />
     )
   }


ianstormtaylor:
Hey @Mangatt can you give an example for what exactly this would be used for? I like that it's minimal, just don't want to accidentally add too much API surface to support in the future. Thanks!

Mangatt:
I need to work with multiple editors. This will allow me mainly three things:

- have single state for multiple editors
- use Slate history instead writing my own
- have custom editor contents structure

Of course this means that some edge cases will have to be managed with custom plugin, but it will be still much more flexible.

Event if that should be undocumented feature, please merge that with master branch, I can't think of case where that could be a problem.

ianstormtaylor:
Hey @Mangatt I'm not going to just merge something into master and keep it undocumented, because that's how things get hard to maintain and end up with poor APIs. I'd either want to support the use case well, or decide not to support it at all.

Since this is has the potential to be a fairly big change to how Slate works (at least as far as I can tell from what you've said so far), I'd want more information before I make a decision. I'd especially like to see working code (probably even a real example in `./examples`) that shows exactly how this works and what benefits it provides. I don't have time myself to put together a working example showing all of that, since I don't even really understand the feature, and I think it's reasonable to expect the onus to be on you to provide the compelling argument for this.

Hope that makes sense.

Mangatt:
Most basic example what that allows is that:

```javascript
<Editor {...props} readOnly>
   {state => <div>
       <div {...editableProps}><RenderNode node={state.document.node[0]} /></div>
       <h1>Not editable content</h1>
       <p>Still not editable</p>
       <div>
            <h2>Neither that is editable</h2>
            <div {...editableProps}><RenderNode node={state.document.node[1]} /></div>
       </div>
   </div>}
</Editor>
```

But working example might be a problem, there has to be plugin in place that handles behavior at the both ends of document nodes since Slates default might not be optimal.

It might not look like a big deal to work with two editors, but in our case there will be a bit more of them a they will be created dynamically.

We are working with Slate 0.16 modified in that way for about a month (actually with my other PR that uses context) and even without rewriting history is our codebase that handles editors much slimmer, more performant and less prone to errors.

Mangatt:
If you are not comfortable with that in core, make that plugin option ```plugin/render(state)```. It could be same as other events:

```If no other plugin handles this event, it will be handled by the Core plugin.``` - e.g. it will work exacly as now. Slate has first class plugins, so this approach is perfectly in line with Slate's philosophy.

Mangatt:
@ianstormtaylor What are your thoughts about that?

ianstormtaylor:
Hey @Mangatt, I've actually recently added something which sounds like what you're asking for (or maybe similar), there's a `plugin.render` method now which allows you to override what is rendered by the editor. Can you check that out and see if it solves your use case?

ianstormtaylor:
Hey @Mangatt I think the added `plugin.render` functionality solves your case, so I'm going to close this. Let me know if not though and we can discuss. Thanks!

Mangatt:
Hi, sorry for long pause, I've been busy with other projects. I'm going back to working with Slate now, so I will try that in couple of days.

Although with just quick peak this looks like one of solutions that I proposed, thank you!

Issue #505, opened 2016-12-07
documents should be allowed to have no nodes


tyler-johnson:
This is an extreme edge case, but if a Slate document only contains a single void block node, the editor locks up and no changes can be made. This can be seen in the "Images" example on slatejs.org.

I attempted to fix this with a schema rule that added blocks before and after the void node, however this lead other issues with the node key. I also don't like that solution as it adds additional nodes that would be otherwise unnecessary.

I'm not sure the best way to fix this and I'm not totally sure this is Slate's responsibility. Just looking for some direction here.

tyler-johnson:
Quick update, I wrote a schema rule to add a block node after the void node which satisfies me for the time being.

```js
{
  match(node) {
    return (node.kind === "block" || node.kind === "document") && !node.isVoid;
  },
  validate(node) {
    return node.nodes.size && node.nodes.last().type === "embed";
  },
  normalize(transform, node) {
    const block = Block.create({ type: "line" });
    transform.insertNodeByKey(node.key, node.nodes.size, block);
  }
}
```

I still think that attempting to delete the void node should not throw an error and that I do believe is a Slate issue.

ianstormtaylor:
Damn, yeah I agree this is a bug. It's tricky because Slate can't get into adding a new block when the image is deleted, since it doesn't know what `type` it should add. But I'd agree that allowing the `document` to be in a state with zero nodes should be allowed.

And then it would be simpler to add a schema validation rule that inserts and empty default node whenever the document is totally empty.

I'd be open to a pull request that makes this possible!

oyeanuj:
@skogsmaskin @ianstormtaylor With the PR #622, I am seeing the following error when I have an empty object (or an object with empty node array) as the initial state: 

```
error given was: TypeError: Cannot read property 'key' of undefined
[1]     at Object.deserializeDocument (node_modules/slate/lib/serializers/raw.js:101:18)
```

The error is thrown here:
https://github.com/ianstormtaylor/slate/blob/master/src/serializers/raw.js#L64

skogsmaskin:
@oyeanuj I don't think that is related to this pull request. To me it sounds like something goes wrong when the deserializer tries to deserialize your object into a Slate document and that is before it is even put into Slate i guess. The #622  only deals with a situation where you have a non-empty document and it allows you to delete all it's nodes (which was not possible before).

PR #622, opened 2017-02-23
Allow deleting a single void block

skogsmaskin:
For a description of the problem see https://github.com/ianstormtaylor/slate/issues/505

This PR will fix the following:

* When there is only a single void block in the document. Allow it to be deleted, leaving a empty document without errors.

* When deleting a single non-void block, it will work as before (leaving the empty block in the document)

When dealing with the the empty document, after the void node is deleted, one should (as for now) make a schema rule that inserts a chosen block into the document after it is emptied:

```
      // Rule to insert a specified block when document is empty
      {
        match: node => {
          return node.kind == 'document'
        },
        validate: document => {
          return document.nodes.size ? null : true
        },
        normalize: (transform, document) => {
          const block = Block.create('paragraph')
          transform
            .insertNodeByKey(document.key, 0, block)
            .focus()
        }
      }
```

I am thinking of if it would be a good idea if the schema defined a ``defaultBlock`` which was automatically inserted when the document was emptied (using rules slows up every change). This could also be used to insert a default block after keypress enter on a void block (which is why I made https://github.com/skogsmaskin/slate-insert-block-on-enter). But this is a larger task at hand. However, if you think it is a good idea, I could look into it.

At least with this, we can now delete void nodes without errors and restore a working state via schema rules.


src/transforms/normalize.js
```diff
@@ -60,16 +60,24 @@ export function normalizeNodeByKey(transform, key, schema) {
...
-  // pointing to nodes that no longer exist, warn and reset the selection.
+  // pointing to nodes that no longer exist, warn (if not unset) and reset the selection.
   if (
     selection.isUnset ||
     !document.hasDescendant(selection.anchorKey) ||
     !document.hasDescendant(selection.focusKey)
   ) {
-    warn('The selection was invalid and was reset to start of the document. The selection in question was:', selection)
+    if (!selection.isUnset) {
+      warn('The selection was invalid and was reset to start of the document. The selection in question was:', selection)
```
ianstormtaylor: What's the need here to make this only warn when it wasn't unset? Is that because if the document can be empty as a legit case, the warning would be impossible to turn off?

skogsmaskin: When the document is emptied the selection is unset.  And when you add a new node this code is activated and it makes a new default selection. I don't think it should warn then at least, but I could not figure out a way to determine that this was the specific scenario right there. So I figured that a unset selection elsewhere maybe not was such a big deal as it made again. However the  ``!document.hasDescendant(selection.anchorKey)`` seems like something that really should be warned about

PR #623, opened 2017-02-24
Fix issue with scrollTo

kennethtruong:
I realized I made a mistake in the scrollWindow function I was calculating the scrollBy position incorrectly. I decided to just switch to scrollTo instead of scrollBy

Also I was getting an issue where the `getBoundingClientRect().bottom` of one of my wrapper elements was incorrect when the height changed via newline or deletion. I'm not entirely sure why but I noticed that if I used wrapperRect.top + wrapper.offsetHeight it gave the correct bottom so I swapped to use that instead. Let me know if there's any issue with these changes.

Also I'm wondering is there anyway to add tests for these changes? I'm not too familiar with testing for scroll.

### DIFF
--- a/src/utils/scroll-to.js
+++ b/src/utils/scroll-to.js
@@ -6,19 +6,19 @@
 import getWindow from 'get-window'
 
 function scrollWindow(window, cursorTop, cursorLeft, cursorHeight) {
-  let deltaX = 0
-  let deltaY = 0
+  let scrollX = window.scrollX
+  let scrollY = window.scrollY
   let cursorBottom = cursorTop + cursorHeight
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @kennethtruong!

There aren't tests currently for lots of the in-browser interactions since they are so hard to mock and test, scroll being one of them. The hardest thing is that `contenteditable` is so complex that's it's not really "mockable" in an easy way (unless someone know of one, in which case please tell me!)

So for now all of the in-browser stuff is tested manually, and for stuff like this scrolling code we might find bugs, but overtime it should settle into something pretty solid.

PR #626, opened 2017-02-24
Fix bug where first/last char of inline node couldn't be selected with keyboard

skogsmaskin:
To show the bug:

* Load the link example. Delete the exclamation sign in the end of the first paragraph, leaving the inline link node at the end of the paragraph.

* Try to select the last character of the link with shift-left-arrow. It will not be selected.

This PR fixes that. I would love to write a test for it, but not sure how to do that.

### DIFF
--- a/src/plugins/core.js
+++ b/src/plugins/core.js
@@ -491,11 +491,12 @@ function Plugin(options = {}) {
       const previousBlock = document.getClosestBlock(previous.key)
       const previousInline = document.getClosestInline(previous.key)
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Nice, @skogsmaskin this looks good to me. Can you apply the same logic to the right arrow key handler too? Thanks!

skogsmaskin:
Done!

ianstormtaylor:
Thanks!

PR #628, opened 2017-02-24
Image example: fix delete bug. Fixes #625

### LINKED ISSUE #625: Bug with delete backward on image example
There is a bug with backspace in the image example. I'm not sure what is the exact reason.

### Steps to reproduce:

1. Open http://slatejs.org/#/images?_k=f2krgi
2. Select the whole paragraph at the end (`This example ... in the editor!`)
3. Press <kbd>Backspace</kbd> once (the paragraph is now empty)
4. Press <kbd>Backspace</kbd> again
5. The cursor is now before the image (weird?)
6. Pressing character, insert it twice on different line (very weird)

### Tested on

- Chrome
- Safari

skogsmaskin:
As void-nodes now can be deleted with backspace, use a schema rule to normalize the document, and insert a paragraph when the document is empty. Delete old "onDocumentChange" handler.

Fixes #625

### DIFF
--- a/examples/images/index.js
+++ b/examples/images/index.js
@@ -1,9 +1,23 @@
 
-import { Editor, Raw } from '../..'
+import { Editor, Block, Raw } from '../..'
 import React from 'react'
 import initialState from './state.json'
 import isImage from 'is-image'
 import isUrl from 'is-url'
+
+
+/**
+ * Default block to be inserted when the document is empty,
+ * and after an image is the last node in the document.
+ *
+ * @type {Object}
+ */
+
+const defaultBlock = {
+  type: 'paragraph',
+  isVoid: false,
+  data: {}
+}
 
 /**
  * Define a schema.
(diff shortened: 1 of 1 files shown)


SamyPesse:
@skogsmaskin I don't understand how this is related to #625. In my case, the document is never empty:

### Before pressing backspace

screen shot 2017-02-25 at 00 23 54

### After pressing backspace

screen shot 2017-02-25 at 00 24 06

### After entering **one** character

screen shot 2017-02-25 at 00 24 16

skogsmaskin:
@SamyPesse: Because of this, which is written when you could not delete void-nodes.

```		
  onDocumentChange = (document, state) => {		
    const blocks = document.getBlocks()		
    const last = blocks.last()		
    if (last.type != 'image') return		
		
    const normalized = state		
      .transform()		
      .collapseToEndOf(last)		
      .splitBlock()		
      .setBlock({		
        type: 'paragraph',		
        isVoid: false,		
        data: {}		
      })
```

SamyPesse:
@skogsmaskin Ok 👍 Thank you for the explanations!

ianstormtaylor:
Thanks for this @skogsmaskin. Can you introduce another schema rule in this PR so that we maintain the "always end in a paragraph" logic as well? It's a useful one to show people since it comes up a lot in questions. Thank you!

skogsmaskin:
I have squashed in such a rule now @ianstormtaylor.

PR #642, opened 2017-03-02
docs: Delete dangling }

rgbkrk:
Running through the walkthroughs and the clean slate for the Applying Custom Formatting section was wrong.


ianstormtaylor:
Thanks @rgbkrk!

PR #645, opened 2017-03-03
Methods missed in upgrade 0.17

oyeanuj:
Hey @ianstormtaylor, while upgrading to 0.17, I noticed that some of the methods that were mentioned in the History file were not in the reference docs. I added those!

Let me know if the descriptions are accurate!


ianstormtaylor:
Thank you @oyeanuj!

PR #648, opened 2017-03-04
Fix getting index for inline node in unwrapInlineAtRange, fixes #620

### LINKED ISSUE #620: Unwrap Inline items in a selection moves spaces in the end of selection
I have some issue, when try to unwrap more than one link in selection, which are separated between each other with space.

### Steps to reproduce

* Create `link` node in schema for Slate editor
* Put 2 links with space between each other in Editor
* Select the links
* Call `unwrapInline('link')`.

### Expected result

The wrapper should be removed and spaces should be placed in the same place

### Actual result

Spaces is moved in the end of selection.

Please, look at the [jsfiddle example](https://jsfiddle.net/x4e692ve/3/). For reproducing, you should just click on the `Unwrap links` button.

Rebulus:
Hi.
This pull request fix issue #620, when inline nodes with text between are unwrapped in incorrect way.

### DIFF
--- a/src/transforms/at-range.js
+++ b/src/transforms/at-range.js
@@ -1032,7 +1032,7 @@ Transforms.unwrapInlineAtRange = (transform, range, properties, options = {}) =>
     .toList()
 
   inlines.forEach((inline) => {
-    const parent = document.getParent(inline.key)
+    const parent = transform.state.document.getParent(inline.key)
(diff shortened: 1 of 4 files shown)


ianstormtaylor:
@Rebulus this is amazing, thank you so much!

PR #649, opened 2017-03-05
Clarify usage of plugin.render

oyeanuj:
Hey @ianstormtaylor, I am adding some clarification to the docs so that others don't run into the issue that I did. However, I am not sure if that is the clearest way to explain the point - so please double-check!

Also, since React doesn't allow you to return an array of nodes, this necessarily means that all the plugins will render nested components, i.e., if you have two toolbars, one would be the parent of another as well as the parent of the editor itself.

Let me know if that should be added, and what do you think is the best way do phrase that note. 

Finally, do you think there should be a way to allow plugins to render as siblings rather than parents of each other?

### DIFF
--- a/docs/reference/plugins/plugin.md
+++ b/docs/reference/plugins/plugin.md
@@ -220,7 +220,7 @@ Like `onChange`, `onBeforeChange` is cummulative.
 ### `render`
 `Function render(props: Object, state: State, editor: Editor) => Object || Void`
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @oyeanuj thanks, good call on the docs update. I think for now it's okay that everything has to render as parents, unless we come to a case where that really makes something not possible.

PR #655, opened 2017-03-07
Fixed autocorrect causing dupe text

lxcid:
For #540

Here's an explanation of what the following codes try to achieve.

> During autocorrection (in iOS’s Safari), `onSelect` event is triggered after `onBeforeInput` event.
> The plugins/core updates the state during `onBeforeInput` event, thus causing selection triggered by autocorrect to be lost/overridden.
> This behaviour caused dupe text bug during autocorrection.
>
> To overcome this issue, we try to query the selection and conditionally fix out of sync cases with an additional transform before inserting the text.

The con of this fix is that it adds an additional query to native selection per text input.

I choose this solution because it is less destructive to existing execution flow.

Feel free to reject or rewrite if this isn't the right way to fix this.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -2,7 +2,7 @@
 import Base64 from '../serializers/base-64'
 import Debug from 'debug'
 import Node from './node'
-import OffsetKey from '../utils/offset-key'
+import getPoint from '../utils/get-point'
 import React from 'react'
 import ReactDOM from 'react-dom'
 import Selection from '../models/selection'
@@ -148,24 +148,7 @@ class Content extends React.Component {
 
   getPoint(element, offset) {
     const { state, editor } = this.props
-    const { document } = state
-    const schema = editor.getSchema()
-
-    // If we can't find an offset key, we can't get a point.
-    const offsetKey = OffsetKey.findKey(element, offset)
-    if (!offsetKey) return null
-
-    // COMPAT: If someone is clicking from one Slate editor into another, the
-    // select event fires two, once for the old editor's `element` first, and
-    // then afterwards for the correct `element`. (2017/03/03)
-    const { key } = offsetKey
-    const node = document.getDescendant(key)
-    if (!node) return null
-
-    const decorators = document.getDescendantDecorators(key, schema)
-    const ranges = node.getRanges(decorators)
-    const point = OffsetKey.findPoint(offsetKey, ranges)
-    return point
+    return getPoint(element, offset, state, editor)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/components/content.js
```diff
@@ -148,24 +148,7 @@ class Content extends React.Component {
 
   getPoint(element, offset) {
```
ianstormtaylor: Can we remove this `#getPoint` method entirely now, and just use the new utility function that you created inline in each of the places?

src/plugins/core.js
```diff
@@ -97,9 +98,35 @@ function Plugin(options = {}) {
 
     const chars = initialChars.insert(startOffset, char)
 
+    let nextTransform = state.transform()
```
ianstormtaylor: Can you change this variable to just be called `transform` instead of `nextTransform`?

src/plugins/core.js
```diff
@@ -97,9 +98,35 @@ function Plugin(options = {}) {
...
+
+    // Determine if selection is out of sync, which can happen during autocorrect,
+    // and update accordingly.
```
ianstormtaylor: Can you change this comment to:

```js
// COMPAT: In iOS, when choosing from the predictive text suggestions, the
// native selection will be changed to span the existing word, so that the word
// is replaced. But the `select` event for this change doesn't fire until after
// the `beforeInput` event, even though the native selection is updated. So we
// need to manually adjust the selection to be in sync. (03/18/2017)
```

lxcid:
Hey, thanks for reviewing! I've made the necessary changes!

> Question for you: if we were to only trigger this extra processing when the native selection was expanded, would that have any negative effects? It seems like that could get around having to trigger it for 99% of the cases where people are just typing individual characters. And might be good for performance.

I think that's a great improvement. It do sounds like a reasonable hypothesis, I couldn't think of a scenario where autocorrect would perform without expanding the selection.

So we just had to check for [`native.isCollapsed`](https://developer.mozilla.org/en-US/docs/Web/API/Selection/isCollapsed) and trigger the logic when it is `false`. I made this change at f4aa680a3181870879c813b3515f2bea29b22ed7.

I think that is all of it! 😃

lxcid:
Unfortunately, f4aa680 caused Android to regress and the autocorrect to work intermittently. 😞 Without f4aa680, Android is usable (autocorrect works), although it has it quirks (Some repeats may happen after the cursor at times, which will disappears as you continue typing at times as well, weird). I decides to remove f4aa680 from this PR.

ianstormtaylor:
Ah damn, alright well we'll just have to keep an eye in the performance and if it becomes a problem we can figure something out!

Thanks @lxcid!

Issue #654, opened 2017-03-06
Changing Block.types by Text/Input is hard.


bunterWolf:
## I would love to make this possible:

![markdown](https://cloud.githubusercontent.com/assets/5108263/23621963/09a32f6e-029d-11e7-8600-aa16b2878fd9.gif)
A slate markdown plugin that works like (gif) [StackEdit](https://stackedit.io).
(Yes I know the [auto-markdown example](http://slatejs.org/#/auto-markdown?_k=2cfkq2) It is not what I want.)

I think the perfekt way to would look like that:

1. onDocumentChange
2. get blocks part of the selection
3. Try some regExs on block.text
4. Set block type. 
5. Schema, Nodes, Render a h1, h2 or p....

- I like to keep # in the text
- Any manipulation of the text should be able to update its block.type. For example: delete one of the #
- The headline is just an example, there are more similar markdown types

### Schema, validate & normalize
Would be grate but there are two problems.

**1. The validation runs not on Typing.**
If you add a # Charakter, nothing would happen.
Validation runs only if you delete something or hit enter.
Try it: https://jsfiddle.net/bunterWolf/4gLsretr/1/

**2. Validation checks the whole document.**
There will be some regExs, which have to check every line, all the time.
The validation function gets only the document as a Parameter, therefore it is not possible to check the selection for the active block.

### onKeyDown
A lot of control! But the state is not updated jet. That means the block.text is without the action the user is doing right now. To merge that on my own (Input/Delete + Selection) is to insane :)

### onChange, onBeforeInput, onDocumentChange:
Could be nice, if there wouldn't be #402  (Which is not marked as bug, isn't it one?)
You can see what happens here: https://jsfiddle.net/bunterWolf/ze6nd2sb/7/ 
if I change the block.type in onChange, onBeforeInput, or onDocumentChange.

## Thats it, for now.
Maybe I get there with marks (Not tried jet), but I would like to use the block.type and keep the marks separate for markdown inline stuff.

I would love to get some ideas or comments?:)

ianstormtaylor:
Hey @bunterWolf, good question.

I think the best way to do this might be to use the `schema` and a custom `onBeforeInput`.

Like you mention the schema isn't validated for all typing changes. This is because in the core plugin's `onBeforeInput` it tries to avoid the "slow" way of changing the content, and prefer native DOM insertion if possible (in the case of simply typing).

[...]

PR #656, opened 2017-03-08
Improve code highlighting example

bunterWolf:
Put the solution from https://github.com/ianstormtaylor/slate/issues/654 
in the code highlighting example.

### DIFF
--- a/examples/Readme.md
+++ b/examples/Readme.md
@@ -45,6 +45,7 @@ If you want to edit the source while running the examples and have those changes
 ```
 npm run watch
 ```
+And open `http://localhost:8080/dev.html`
 
 
 ## Development Examples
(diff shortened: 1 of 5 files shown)


bunterWolf:
I think this is right now the best place this way of markdown handler.
If I would create a new markdown-decorations example that would end in two very similar prism integrations.
Maybe I come up with a more advanced markdown editor for an extra example soon ;)

ianstormtaylor:
Hey @bunterWolf thanks for taking the time to do this! I opted to add it as a separate example, because I'd like to keep code highlighting simple, since it too is a common use case. Now we've got both handled, in the simplest way that they can be, so hopefully they're not too hard to maintain. Thanks!

PR #662, opened 2017-03-13
Avoid unneeded re-rendering of `Leaf` components.

AlbertHilb:
After any change in Slate `Selection`, `Leaf` nodes containing an edge of the selection are updated.
However this is not always necessary. It suffices to update `Leaf` nodes when the native selection is not already inside them.
These unneeded re-renderings not only are a waste of resources but worsen the editing user experience.
Moving the caret up and down in a document using arrow keys, should keep its horizontal position unchanged unless the actual text row is too short, in that case the position should be restored as soon as the caret arrives on a row long enough.
In Chrome, this requirement is not fulfilled by Slate. After any caret movement the DOM selection is reset and the old horizontal position lost.

This PR wants to fix this issue.

It adds a new property, `selectionNeedsRedraw`, to Slate `State`.
After a native selection change, the `onSelect` event is fired, the new Slate `Selection` is determined and `selectionNeedsRedraw` is set to `false`.
When any transform is applied to a Slate `State`, unless `selectionNeedsRedraw` option is specified, the state `selectionNeedsRedraw` property is set to `true` to indicate native selection must be updated to reflect the changes in the editor state.
Finally, on editor re-rendering, an unchanged `Leaf` node containing a selection edge is updated only if `selectionNeedsRedraw` is true.

As always I'm open to any change to make this PR able to be merged. :blush:

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -708,8 +708,8 @@ class Content extends React.Component {
     // Store the document and selection after the `onSelect` event.
     // At rendering time, if they are unchanged we'll avoid to render
     // the native selection again.
-    this.onSelectDocument = document
-    this.onSelectSelection = data.selection
+    this.tmp.document = document
+    this.tmp.selection = data.selection
 
     debug('onSelect', { event, data })
     this.props.onSelect(event, data)
@@ -726,25 +726,24 @@ class Content extends React.Component {
     const { className, readOnly, state, tabIndex, role } = props
     const { document, selection } = state
 
-    let selectionNeedsRedraw = true
+    // If selection is blurred or document and selection are the same we have
+    // stored after `onSelect` event, selection does not need to be rendered.
+    let redrawSelection = selection.isFocused
 
-    // If document and selection are the same we have stored after `onSelect`
-    // event, native selection does not need to be rendered again.
-    if (this.onSelectSelection) {
-      selectionNeedsRedraw = (
-        this.onSelectSelection.anchorKey !== selection.anchorKey ||
-        this.onSelectSelection.anchorOffset !== selection.anchorOffset ||
-        this.onSelectSelection.focusKey !== selection.focusKey ||
-        this.onSelectSelection.focusOffset !== selection.focusOffset ||
-        this.onSelectDocument !== document
+    if (this.tmp.selection) {
+      redrawSelection = (
+        redrawSelection && (
+          !selection.equals(this.tmp.selection) ||
+          document !== this.tmp.document
+        )
       )
 
-      this.onSelectDocument = null
-      this.onSelectSelection = null
+      this.tmp.document = null
+      this.tmp.selection = null
     }
 
     const children = document.nodes
-      .map(node => this.renderNode(node, selectionNeedsRedraw))
+      .map(node => this.renderNode(node, { redrawSelection }))
       .toArray()
 
     const style = {
@@ -811,7 +810,8 @@ class Content extends React.Component {
    * @return {Element}
    */
 
-  renderNode = (node, selectionNeedsRedraw) => {
+  renderNode = (node, options) => {
+    const { redrawSelection } = options
     const { editor, readOnly, schema, state } = this.props
 
     return (
@@ -823,7 +823,7 @@ class Content extends React.Component {
         state={state}
         editor={editor}
         readOnly={readOnly}
-        selectionNeedsRedraw={selectionNeedsRedraw}
+        redrawSelection={redrawSelection}
       />
     )
   }
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -703,13 +703,18 @@ class Content extends React.Component {
       data.selection = selection
         .merge(properties)
         .normalize(document)
-    }
 
-    // Store the document and selection after the `onSelect` event.
-    // At rendering time, if they are unchanged we'll avoid to render
-    // the native selection again.
-    this.tmp.document = document
-    this.tmp.selection = data.selection
+      // If selection edges aren't inside void inline nodes, store document and
+      // selection. At rendering time, if they are unchanged there will be no
+      // need to render the native selection again.
+      if (
+        (!anchorInline || !anchorInline.isVoid) &&
+        (!focusInline || !focusInline.isVoid)
+      ) {
+        this.tmp.document = document
+        this.tmp.selection = data.selection
+      }
+    }
 
     debug('onSelect', { event, data })
     this.props.onSelect(event, data)
@@ -726,16 +731,14 @@ class Content extends React.Component {
     const { className, readOnly, state, tabIndex, role } = props
     const { document, selection } = state
 
-    // If selection is blurred or document and selection are the same we have
-    // stored after `onSelect` event, selection does not need to be rendered.
-    let redrawSelection = selection.isFocused
+    // If document and selection are the same we have stored after `onSelect`
+    // event, native selection does not need to be re-rendered.
+    let redrawSelection = true
 
     if (this.tmp.selection) {
       redrawSelection = (
-        redrawSelection && (
-          !selection.equals(this.tmp.selection) ||
-          document !== this.tmp.document
-        )
+        !selection.equals(this.tmp.selection) ||
+        document !== this.tmp.document
       )
 
       this.tmp.document = null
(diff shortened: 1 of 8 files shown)


AlbertHilb:
I rewrite the code to avoid adding `selectionNeedsRedraw` to `State`.
Now, `Document` and `Selection` after `onSelect` event `Content` handler are cached.
At rendering time, if they result unchanged we'll avoid to render native selection again.
To nofity `Leaf`components if they have to update the native selection, a new property `selectionNeedsRedraw` is added to `Node`.

src/components/content.js
```diff
@@ -705,6 +705,12 @@ class Content extends React.Component {
         .normalize(document)
     }
 
+    // Store the document and selection after the `onSelect` event.
+    // At rendering time, if they are unchanged we'll avoid to render
+    // the native selection again.
+    this.onSelectDocument = document
+    this.onSelectSelection = data.selection
```
ianstormtaylor: Can we put this inside `this.tmp`, just for consistency?

src/components/content.js
```diff
@@ -718,9 +724,27 @@ class Content extends React.Component {
...
+    let selectionNeedsRedraw = true
+
+    // If document and selection are the same we have stored after `onSelect`
+    // event, native selection does not need to be rendered again.
+    if (this.onSelectSelection) {
+      selectionNeedsRedraw = (
+        this.onSelectSelection.anchorKey !== selection.anchorKey ||
+        this.onSelectSelection.anchorOffset !== selection.anchorOffset ||
+        this.onSelectSelection.focusKey !== selection.focusKey ||
+        this.onSelectSelection.focusOffset !== selection.focusOffset ||
```
ianstormtaylor: Why the need here to compare all of the properties individually, instead of just comparing `onSelectSelection !== selection`?

If we really do need the deep comparison, we could use `Immutable.is` maybe?

src/components/content.js
```diff
@@ -718,9 +724,27 @@ class Content extends React.Component {
...
+        this.onSelectDocument !== document
+      )
+
+      this.onSelectDocument = null
+      this.onSelectSelection = null
+    }
+
     const children = document.nodes
-      .map(node => this.renderNode(node))
+      .map(node => this.renderNode(node, selectionNeedsRedraw))
```
ianstormtaylor: Can you change this to be `{ redrawSelection }` as an options object instead?

src/components/leaf.js
```diff
@@ -88,23 +88,16 @@ class Leaf extends React.Component {
...
-    // one or both of the selection's edges.
-    if (props.state.isFocused) {
+    // If the selection will be focused and native selection is not already
+    // in syncro with it, update native selection.
+    if (props.state.isFocused && props.selectionNeedsRedraw) {
       const { index, node, ranges, state } = props
       const { start, end } = OffsetKey.findBounds(index, ranges)
-      if (state.selection.hasEdgeBetween(node, start, end)) return true
+      if (state.selection.hasEdgeBetween(node, start, end)) {
+        this.updateSelection(props)
```
ianstormtaylor: It feels weird to be running this function inside `shouldComponentUpdate`, instead of returning `true/false` and doing this logic elsewhere. Is there any alternative?

src/components/node.js
```diff
@@ -37,6 +37,7 @@ class Node extends React.Component {
     parent: React.PropTypes.object.isRequired,
     readOnly: React.PropTypes.bool.isRequired,
     schema: React.PropTypes.object.isRequired,
+    selectionNeedsRedraw: React.PropTypes.bool.isRequired,
```
ianstormtaylor: Similar to above, can we rename this to `redrawSelection`?

src/components/content.js
AlbertHilb: Yes, of course!

src/components/content.js
AlbertHilb: We can't just compare identity of `selection` and `onSelectSelection` because the former is got by merging the ladder into the previous state selection.
So they are different objects even when they represent the same selection.
But, of course, we can use `Immutable.is`.

src/components/content.js
AlbertHilb: Yes, of course!

src/components/leaf.js
AlbertHilb: I agree with you, it's a bit weird, but we can't really do otherwise at moment.
To avoid this we should rise the selection update code up in the component hierarchy and this would make the `Node` component source a mess unless we move the `Text` node related content in a separated component.

src/components/node.js
AlbertHilb: Yes, of course!

src/components/leaf.js
ianstormtaylor: But wouldn't returning `true` here cause the `updateSelection` to be called from inside `componentDidUpdate` next? Which feels more like the "right" way. I might be missing something though.

AlbertHilb: Yes, but returning true triggers (unneeded) re-rendering of the Leaf, which is just what I want to avoid.

ianstormtaylor: Hmm yeah good point, I forgot about the main goal haha. Okay hmm, that still feels weird so I wonder if there's another way to do it.

Would it work for us to set a `data.isNative` property on the `data` inside `onSelect`, and then use that to trigger the existing native behavior logic? It seems like there's a mention of `data.isNative` inside `onSelect` already, but that it's not actually being used by the core plugin right now. I wonder if fixing that would solve this too? But without needing to mess with calling updating from inside `shouldComponentUpdate`?

src/components/node.js
AlbertHilb: Unfortunately we can't use `data.isNative`.
First of all, `updateSelection` is triggered by selection changes that are generated programmatically and not by native selection changes. Indeed when the changes are consequences of native selection transformations the two selections are already in sync and there is no need to call 'updateSelection'.
Moreover setting `data.isNative` to `true` stops the updating of the whole editor content, but when selection changes, some components need re-rendering: for example `Void` `Inline` nodes which want to show an highlight border if selected.
The solution I adopted is to split in two the `Leaf` component.
The *higher* component, still named `Leaf`, is responsible for updating the native selection, the *lower*, named `LeafContent`, instead is used to render the leaf text content.
In this way we can update the `Leaf`, and therefore the native selection, without touching its content.

src/components/leaf.js
AlbertHilb: Unfortunately we can't use `data.isNative`. :disappointed: 
First of all, `updateSelection` is triggered by selection changes that are generated programmatically and not by native selection changes. Indeed when the changes are consequences of native selection transformations the two selections are already in sync and there is no need to call 'updateSelection'.
Moreover setting `data.isNative` to `true` stops the updating of the whole editor content, but when selection changes, some components need re-rendering: for example `Void` `Inline` nodes which want to show an highlight border if selected.
The solution I adopted is to split in two the `Leaf` component.
The *higher* component, still named `Leaf`, is responsible for updating the native selection, the *lower*, named `LeafContent`, instead is used to render the leaf text content.
In this way we can update the `Leaf`, and therefore the native selection, without touching its content.

ianstormtaylor:
Hey @AlbertHilb, I've just been thinking about this more after looking into some of the selection-related behaviors for https://github.com/ianstormtaylor/slate/issues/688.

I think it might actually be best if we change the way we handle selection updating, to decouple it from leaf rendering altogether. Right now we're in this weird place where to get the selection to update we have to invoke the entire rendering stack. In this PR we're making it slightly better by ensuring that the leaves don't actually re-render, but doing it with the weirdness of invoking logic inside `shouldComponentUpdate`. (Which is still much better than what I originally wrote, and I'm really grateful for your PR!) 

One of the problems I encountered from https://github.com/ianstormtaylor/slate/issues/688 is that in some cases a "duplicate" selection is made as far as Slate is concerned, which causes `Content#onSelect` to return early. But the native selection has already been updated at this point, so it then gets out of sync, and Slate doesn't realize it. 

This results in things like this...



...where the selection is able to get into "invalid" positions, because Slate thinks it's actually just a duplicate selection.

So...

I think because of cases like that, we should move to handling all of the selection syncing logic in `Content#componentDid{Mount,Update}`. That way we'd over deep rendering of leaves when selection changes, and we'd also be able to "sync" the selection in the case that it's out of date for other reasons.

What do you think?

ianstormtaylor:
@AlbertHilb I've updated the selection updating implementation to happen at the top-level now to avoid the leaf re-rendering!

Issue #663, opened 2017-03-14
Internet Explorer 11 support


tyler-johnson:
I have been working on getting our platform compatible in IE11 and Slate is really close. In fact with just a couple polyfills, it mostly works. I know there isn't official support for IE with Slate, but I figured I would document my findings here seeing as I don't believe it would take too much work to get it working in IE.

- No support for `Element.closest()`. This is used through out the Slate codebase. Easy to polyfill, I used the one here: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill. This might be good to replace with a built in closest utility instead of relying on the browser seeing as Edge also lacks support for it.
- No support for `Selection.extend()`. I know this is used a few times in [components/leaf.js](https://github.com/ianstormtaylor/slate/blob/6a92f9384d5bc0437d17f178fdc22ab101c928b4/src/components/leaf.js#L179), possibly elsewhere as well. Unfortunately I couldn't find a solid polyfill for this. If someone finds one I'd love to get my hands on it. Instead I wrote my own: https://gist.github.com/tyler-johnson/0a3e8818de3f115b2a2dc47468ac0099. It works for Slate, but I'm not sure how compatible it is with the native version.
- `DataTransfer.setData()` [only supports a type value of 'text'](http://stackoverflow.com/a/18051912). This affects all of the drag-n-drop support, but more specifically these lines: [utils/get-transfer-data.js#L21-L25](https://github.com/ianstormtaylor/slate/blob/6a92f9384d5bc0437d17f178fdc22ab101c928b4/src/utils/get-transfer-data.js#L21-L25). Judging by that code, I think its possible to fix this by making an exception for IE where only text types are allowed and everything else is ignored or interpreted as text. Not the most elegant, but at least some forms of drag-n-drop would be possible.
- No `window.scrollX` or `window.scrollY`. This is used in [utils/scroll-to-selection.js](https://github.com/ianstormtaylor/slate/blob/master/src/utils/scroll-to-selection.js). [MDN says to use `pageXOffset` and `pageYOffset`](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollX) and my local tests show that these replacements do work.

Let me know if I can help to get Slate working in IE; I can definitely open pull-requests. Thanks.

ianstormtaylor:
Hey @tyler-johnson thanks for posting this!

[...]

PR #664, opened 2017-03-14
use pageXOffset instead of scrollX for IE11 support

tyler-johnson:
Since I already had the code loaded up for this one I figured a pull-request was easy. This will add support for scrolling in IE (even though other things are broken) which prior was scrolling to the top frequently. See #663 for my other IE11 support items.

For more info on IE compatibility with window scrolling, see the [MDN article on scrollX](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollX).


ianstormtaylor:
Thanks @tyler-johnson!

PR #665, opened 2017-03-16
Allow inserting successive empty non-void blocks

skogsmaskin:
Slate will not allow inserting successive empty non-void blocks through ``transform.insertBlock``. Not sure about the rationale behind this logic, but I suppose there is a reason for this (please enlighten me).

This PR is an experiment with allowing ``insertBlock`` to allow take an option to allow inserting successive empty non-void blocks (``replaceEmpty: false``).

What do you think?

### DIFF
--- a/src/transforms/at-current-range.js
+++ b/src/transforms/at-current-range.js
@@ -163,11 +163,11 @@ Transforms.deleteWordForward = (transform) => {
  * @param {String|Object|Block} block
  */
 
-Transforms.insertBlock = (transform, block) => {
+Transforms.insertBlock = (transform, block, options) => {
   block = Normalize.block(block)
   const { state } = transform
   const { selection } = state
-  transform.insertBlockAtRange(selection, block)
+  transform.insertBlockAtRange(selection, block, options)
 
   // If the node was successfully inserted, update the selection.
   const node = transform.state.document.getNode(block.key)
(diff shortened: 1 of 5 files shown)


ianstormtaylor:
@skogsmaskin good point. Originally I had added it with the current behavior for cases where you want to "insert an image block", which often when trying to do that inside an empty block you mean to actually replace it.

But I think you're right that it's a bit non-expected, and could be improved.

I wonder if it would actually be better to remove the replacing from `insertBlock` altogether, and add a separate method called `replaceBlock` instead, and then have that custom "insert or replace" logic be pushed into userland for the specific use cases.

Would you be down to make that change?

skogsmaskin:
@ianstormtaylor yes, I am down. This is what should be the expected behaviour, and it is much better that it is up to the user to choose to insert or replace. I will have a go at it very soon. The new function ``replaceBlock``would of course replace the existing block empty or not.

skogsmaskin:
Sorry for taking so long, I am still down, just haven't had the time :/

oyeanuj:
@skogsmaskin Curious if this is still on your radar?

skogsmaskin:
Hey! Again, sorry for the delay here. I had parental leave a couple of months until this august.

I've made a new PR #1117 for this, as the code has changed a lot since the time I created this PR.

PR #666, opened 2017-03-16
added onFocus, and docs. fixes #650

### LINKED ISSUE #650: first backspace after gaining window focus does not behave as expected.
**Expected:** a character to be deleted on pressing backspace after focusing back to window with the editor.

**Unexpected:** First backspace after gaining window focus seems to be ignored for some reason. Backspaces after the first works as expected.

**Reproduce:**
1) Type something in the editor.
2) Focus on any other window.
3) Focus back on slate editor window.
4) Hit backspace key (last character left of caret does not get deleted).

I'm on MacOS. Tested on Google Chrome and Chromium.

Tried it in a couple of the Slate Example pages as well.

yalu:
I have not ran into any problems running this for a while and it seems useful to have generally.


ianstormtaylor:
Awesome, thank you @yalu!

Issue #660, opened 2017-03-11
Unable to copy or cut void nodes


oyeanuj:
And hey again! Another issue/question on the void nodes. It seems from the examples that one can't cut or copy the void nodes by default - that there is no native editor behavior. I was wondering if I could understand how and why it behaves the way it does. 

Does it imply that we have to implement custom `onCut`, `onCopy` handlers everytime we have void nodes? 

Or should the default ones fire with the default behavior - like lets say, in the case of copying a void node from one part of the editor to another?

Thanks!

oyeanuj:
Also, FWIW, the `document.queryCommandEnabled('copy')` returns false inside the void node on Chrome.

ianstormtaylor:
Hey @oyeanuj good question. It should handle this by default, but because of the way that copying currently works I think it aborts if the selection is collapsed, so it doesn't void nodes. But I'd definitely accept a PR that fixes that!

oyeanuj:
Okay, I think I can tackle that. So, just to be sure, would the desired solution be to abort only if selection is collapsed AND NOT in a void node? 

I did find [this line that was responsible here](https://github.com/ianstormtaylor/slate/blob/b8543b537c95b87d984c3bb368a9c51cc3d29ad1/src/plugins/core.js#L225), so I'll play around and test it -

skogsmaskin:
Any progress on this @oyeanuj ? I need a solution to the same problem.

oyeanuj:
@skogsmaskin Apologies, I haven't had the chance to look into it, but will in the next day or two. Will post an update soon, assuming my comment above is the desired solution.

ianstormtaylor:
That solution sounds good to me. Anything that gets it to function for all types of nodes for internal copies sounds good to me, mimicking the native behavior as much as possible.

oyeanuj:
A quick update and question for @ianstormtaylor:

So, it has turned out to be more complicated than I thought (as always). But tracing thru functions, it seems like the root cause seems to be that `state.fragment.nodes` is empty when copying from a void node. Because of this, even if I comment out the line mentioned in the previous comment, there is nothing to copy.

`state.fragment.nodes` being empty seems to happen because `document.getFragmentAtNode` [returns early if the selection is collapsed](https://github.com/ianstormtaylor/slate/blob/master/src/models/node.js#L504).

[...]

PR #676, opened 2017-03-22
Allow cut/copy for void nodes

oyeanuj:
I'm starting this PR to take care of #660. So far, this PR does the following - 

- [x] Adds a flag in the `data` object to detect if the selection is in the void node or not. If it is in void node, then it does not return early even if selection is collapsed.

- [x] Ensures that the void node is returned without error via `data.fragment`, and is encoded as an attribute `data-slate-fragment` to a span.

 - [x] To copy, need to add that span incase of void nodes as well. 

 - [ ] The above should have ensured that either the HTML or the fragment is copied to the clipboard, so that it can be pasted elsewhere in Slate. But that doesn't happen yet.. 
I'd appreciate any ideas!

The relevant line is [here in the original code](https://github.com/ianstormtaylor/slate/blob/master/src/plugins/core.js#L267) and [here in the diff](https://github.com/ianstormtaylor/slate/pull/676/files#diff-f519e645b4d20e4ca833307b39827cabR280).


src/components/content.js
```diff
@@ -342,6 +343,7 @@ class Content extends React.Component {
     const data = {}
     data.type = 'fragment'
     data.fragment = state.fragment
+    data.inVoidNode = state.fragment.nodes.some(node => node.isVoid)
```
ianstormtaylor: Hey @oyeanuj why do we need to add this here, instead of just performing this `some` check inside the `onCutOrCopy` handler? I think it would be cleaner to do it that way if possible.

oyeanuj: Updated.

oyeanuj:
Question for @ianstormtaylor or anyone else familiar with DOM selection and DocumentFragment:

I am a little stuck at the third task. I am able to append an empty span with the fragment in an attribute to the DOM but that doesn't seem to do the trick, as it does for text. What might be the best way to ensure that HTML fragment is copied in non-text nodes?

ianstormtaylor:
Closing in favor of https://github.com/ianstormtaylor/slate/pull/676/files

Thank you @oyeanuj!

Issue #514, opened 2016-12-09
insert inline void in the first character of a line will insert an additional line break before it.


chilijung:
demo emoji have the exact problem.

![emoji](https://cloud.githubusercontent.com/assets/1216029/21045056/a77ed542-be39-11e6-8f49-aa6f0ae15031.gif)

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/682

PR #682, opened 2017-03-23
Fix https://github.com/ianstormtaylor/slate/issues/514

kib357:
It's enough hard to re-implement and compare with previous state all logic of leaf/renderText, but i think it's not so expensive to re-render text node if it was last and now it is not.

### DIFF
--- a/src/components/node.js
+++ b/src/components/node.js
@@ -145,23 +145,35 @@ class Node extends React.Component {
       return true
     }
 
-    // For text nodes, which can have custom decorations, we need to check to
-    // see if the block has changed, which has caused the decorations to change.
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @kib357 sorry for the delay on this, and thank you for digging into this!

I just recently refactored a the `shouldComponentUpdate` logic, and it was causing conflicts with this PR, so I copied it over manually and am releasing a new patch version with the fix now. Thank you!!

kib357:
Thank you for fixing issue! And also thank you for all your work on Slate – we are currently using editor in ManyChat in production, and it's great. This was last issue that we have fixed in fork, and now we can totally move to official npm package)

ianstormtaylor:
@kib357 awesome, I'm glad to hear that :D

PR #684, opened 2017-03-24
Documentation: use correct localStorage method for setting content

conorcussell:
the section on saving and loading content was using `set` instead of `setItem`


ianstormtaylor:
Ah shoot, thanks @conorcussell!

PR #694, opened 2017-03-30
Fix mismatched event parameter.

AlbertHilb:
Fix mismatched event parameter.


lxcid:
Just found this. Thanks for fixing! Sorry for this!

PR #705, opened 2017-04-01
a handful of performance improvements

ianstormtaylor:


ianstormtaylor:
Some decent improvements after lots of these little quick fixes. The benchmarks aren't the most accurate probably, but at least in the general area. This compares pre/post this PR:

image

PR #717, opened 2017-04-10
Keep selection while applying Transform.unwrapNodeByKey

### LINKED ISSUE #670: Node is lost selection after unwrapNodeByKey
Hi.

I found interesting behavior when using `unwrapNodeByKey`. When I tried to unwrap list, which had been created by plugin [slate-edit-list](https://github.com/GitbookIO/slate-edit-list), it lost its cursor.

I think, it happens, when we trying to unwrap children from its parent with using `unwrapNodeByKey`.

I prepared sandbox and steps to reproduce. Please, could you fix this issue or help me to understand, how can I fix it myself?

**Steps to reproduce**
* Open slate sandbox https://jsfiddle.net/9prvpfvu/2/
* Put cursor on the text
* Click button "Wrap in quote"
* Clock button "Unwrap in quote"

**Actual result**
Button is lost its cursor

**Expected result**
Button should be not lost its cursor.

vpontis:
Transforms should be applied in an order so that the selection is
preserved. This replaces "remove --> insert" with "move --> remove".

Close #670

### DIFF
--- a/src/transforms/by-key.js
+++ b/src/transforms/by-key.js
@@ -361,9 +361,8 @@ Transforms.unwrapNodeByKey = (transform, key, options = {}) => {
 
 
   if (parent.nodes.size === 1) {
-    // Remove the parent and replace it by the node itself.
-    transform.removeNodeByKey(parent.key, { normalize: false })
-    transform.insertNodeByKey(parentParent.key, parentIndex, node, options)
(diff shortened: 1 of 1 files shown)


vpontis:
@ianstormtaylor let me know if you think this should also have a test case.

ianstormtaylor:
Boom, thanks man!

PR #728, opened 2017-04-18
feat: Add findClosestNode utility

Craga89:
* A lightweight utility that wraps the functionality of the native `Node.closest` method
* IE11 and Edge don't support this native method
* For these browsers that don't support it, we mock it with a [shim](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill) taken from MDN
* All supporting browsers use the native method

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -3,6 +3,7 @@ import Base64 from '../serializers/base-64'
 import Debug from 'debug'
 import Node from './node'
 import getPoint from '../utils/get-point'
+import findClosestNode from '../utils/find-closest-node';
 import React from 'react'
 import Selection from '../models/selection'
 import getTransferData from '../utils/get-transfer-data'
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Thanks @Craga89!

bendulum:
FYI this issue is popping up again in `slate-react@0.10.19`. A string search for `closest(` in the package shows several use cases:

`return el.isContentEditable && (el === element || el.closest('[data-slate-editor]') === element);`

`var targetEditorNode = target.closest('[data-slate-editor]');`

`var closest = element.closest('[data-key]');`

`var rangeNode = parentNode.closest(RANGE_SELECTOR);`

`var textNode = rangeNode.closest(TEXT_SELECTOR);`

`var voidNode = parentNode.closest(VOID_SELECTOR);`

Tried in IE11 on Windows 10.
I might create a PR for this, but someone else may be faster.
Please let me know if I should create a separate issue for this specific bug.

ianstormtaylor:
Hey @bendulum, we reverted this and a few other like it, and now you must bring your own polyfills for old browsers like IE11.

PR #729, opened 2017-04-18
feat: Add ES6 polyfill to examples page

Craga89:
* Fixes issue with IE11 simply not displaying anything on load due to missing `Symbol` shim


ianstormtaylor:
Thank you @Craga89, good call!

Issue #734, opened 2017-04-19
Internal copy paste is broken


skogsmaskin:
Version 0.19.16 was the last version where copy and pasting internally in the document worked as expected (keeping the formatting intact). The ``.fragment`` document is now ``null``, and there is only a ``.html`` part, thus it falls back to ``onPasteText`` by the default core plugin. I suspect it has something to do with #716.

If you try copying and pasting the contents of http://slatejs.org/#/rich-text?_k=952y37, you see what I mean.

skogsmaskin:
Confirmed it happened with 3e573453a8c5d92963d3f290fb7266f66e2a6b78.

skogsmaskin:
What is happening is that the span with the encoded data is no longer on the clipboard, thus [getTransferData](https://github.com/ianstormtaylor/slate/blob/master/src/utils/get-transfer-data.js#L30) don´t recognize it as a fragment.

skogsmaskin:
Temporarily fixed by #736 (revert patch 3e573453a8c5d92963d3f290fb7266f66e2a6b78).

skogsmaskin:
Regretted closing it. Should be some tests in place.

ianstormtaylor:
Seems like this is fixed now.

SebAshton:
@ianstormtaylor I think this may have regressed. Confirmed on http://slatejs.org/#/rich-text.

![bug_report](https://cloud.githubusercontent.com/assets/881759/26447281/bedd04e6-413f-11e7-99e9-0133ea4ff750.gif)

Neither the marks, or blocks are correctly copied/pasted.

**Browser** Chrome - Version 58.0.3029.110 (64-bit)

illesg:
I can reproduce it too with the official example the same way as @SebAshton 

Browser: Chrome, Version 59.0.3071.109 (Official Build) (64-bit)

danburzo:
There are a hosts of problems relating to copy/paste in Chrome and Safari and they stem from a single problem:

```js
    const r = window.document.createRange()
    r.selectNodeContents(div)
    native.removeAllRanges()
    native.addRange(r)
```

[...]

PR #716, opened 2017-04-09
Remove space after first char when copying from Slate

vpontis:
Right now if you copy "This thing" from Slate and paste
into a plain text app you will get "T his thing" because
Slate is making the first character a data holding span.

This diff makes the data containing span empty to eliminate
this extra space.

Closes #562

PR #736, opened 2017-04-20
Fix copy paste

skogsmaskin:
Internal copy/paste broke with PR #716.
This reverts 3e573453a8c5d92963d3f290fb7266f66e2a6b78

Se issue https://github.com/ianstormtaylor/slate/issues/734 for more details.

EDIT: I thought I had a solution to support the intention of 3e57345 with this PR, but it was still containing a whitespace on the clipboard. Removed that commit from this PR. As it is just now a revert, feel free to close without merge. However, this should be taken care of ASAP as it breaks core functionality.


ianstormtaylor:
Thanks @skogsmaskin that's super unfortunate, we'll have to figure out a different solution there.

PR #738, opened 2017-04-21
Keep node style when deleting empty line above node

### LINKED ISSUE #548: preserve block type when joining with empty adjacent node
This issue is about discussing the general problem of what should happen when joining or splitting lines resp. adjacent block nodes where the types of the block nodes differ. Specifically, empty lines seem to make the problem more complicated.

The default behavior is as follows. Removing a line break joins the previous and current node such that "the current node gets the style of the previous node (e.g. header gets paragraph style)". Conversely, creating a new line applies the style of the previous line. In some sense it's a nice symmetric behavior but not conventional in all situations.

For example (https://streamable.com/z0v6c), suppose you have a header block node and a paragraph block node and the situation is the following:

```
Some paragraph text

<caret>Header
```

When you press backspace the `Header` block is joined with the previous empty line and the header styling is lost. It appears that the previous block is dominant by default. However, it would be more conventional if pressing backspace in the above situation would keep the header and its styling because the line above the header is empty. If you'd press backspace again then losing the header styling is justified (because there is no empty line between the paragraph and header anymore).

------------

I was thinking about adding special cases to the `onDocumentChange` handler. The idea for handling joins is to get the last transformation type, check if it's 'join_node', then check if the new node is empty and if yes "apply the style" of the second node. In code:

```
  getTopmostParent = (key: string) => {
    const parent = this.state.document.getParent(key)
    const depth = this.state.document.getDepth(parent.key)
    if (depth > 1) return this.getTopmostParent(parent.key)
    return parent
  }

  getLastTransform = (state) => {
    if (state.history.undos && state.history.undos._head && state.history.undos._head.value && state.history.undos._head.value.length) {
      return state.history.undos._head.value[state.history.undos._head.value.length - 1]
    }
    return null
  }

  onDocumentChange = (state) => {
    const lastTransform = this.getLastTransform(state)
    if (lastTransform != null && lastTransform.type === 'join_node') {
      const newNode = this.getTopmostParent(state.selection.startKey)
      const oldNode = this.getTopmostParent(this.prevState.selection.startKey)
      if (newNode.isEmpty && oldNode.type === 'header') {
        state = state.transform().setBlock('header').apply({ save: false })
      }
    }
    // ...
  }
```

What do you think. Is this a good approach?

Regarding splitting. The use case I want to implement is that you cannot have a header that contains a new line. If you press enter while the selection is somewhere in-between the characters of the header, it should be split into two headers. This works by default. However, when pressing at the end of a header a new paragraph should be inserted. This is how I implemented the latter behavior:

```
  onKeyDown = (event, data, state) => {
    if (event.key === 'Enter') {
      const node = this.getTopmostParent(state.selection.startKey)
      if (node.type === 'header' && state.selection.isAtEndOf(node)) {
        return state.transform().insertBlock('paragraph').apply()
      }
    }
  }
```

It works but I wonder if there is a better approach.

vpontis:
Right now when you merge a block with the previous block, the previous
block's style always wins, even if the previous block's content is
empty. This does not make sense since deleting empty lines should not
affect style.

This updates the `deleteAtRange` transform to keep the style of the
bottom block in the case that the previous block is empty.

Closes #548

### DIFF
--- a/src/transforms/at-range.js
+++ b/src/transforms/at-range.js
@@ -117,7 +117,7 @@ Transforms.deleteAtRange = (transform, range, options = {}) => {
     let blockToKeep = startBlock
 
     // We keep the style of the blockToKeep. endBlock should keep style if deleting empty line.
-    if (startBlock.isEmpty && !endBlock.isEmpty) {
+    if (startBlock.isEmpty) {
       blockToRemove = startBlock
       blockToKeep = endBlock
     }
--- a/src/transforms/at-range.js
+++ b/src/transforms/at-range.js
@@ -113,14 +113,23 @@ Transforms.deleteAtRange = (transform, range, options = {}) => {
   const endBlock = document.getClosestBlock(document.getNextText(endKey).key)
 
   if (startBlock.key !== endBlock.key) {
-    endBlock.nodes.forEach((child, i) => {
-      const newKey = startBlock.key
-      const newIndex = startBlock.nodes.size + i
+    let blockToRemove = endBlock
+    let blockToKeep = startBlock
+
+    // We keep the style of the blockToKeep. endBlock should keep style if deleting empty line.
+    if (startBlock.isEmpty && !endBlock.isEmpty) {
+      blockToRemove = startBlock
+      blockToKeep = endBlock
+    }
+
+    blockToRemove.nodes.forEach((child, i) => {
+      const newKey = blockToKeep.key
+      const newIndex = blockToKeep.nodes.size + i
       transform.moveNodeByKey(child.key, newKey, newIndex, OPTS)
     })
 
-    // Remove parents of endBlock as long as they have a single child
-    const lonely = document.getFurthestOnlyChildAncestor(endBlock.key) || endBlock
+    // Remove parents of blockToRemove as long as they have a single child
+    const lonely = document.getFurthestOnlyChildAncestor(blockToRemove.key) || blockToRemove
     transform.removeNodeByKey(lonely.key, OPTS)
   }
 


skogsmaskin:
Ah, this is great @vpontis, this has been a real annoyance for me too.

src/transforms/at-range.js
```diff
@@ -113,14 +113,23 @@ Transforms.deleteAtRange = (transform, range, options = {}) => {
...
 
   if (startBlock.key !== endBlock.key) {
-    endBlock.nodes.forEach((child, i) => {
-      const newKey = startBlock.key
-      const newIndex = startBlock.nodes.size + i
+    let blockToRemove = endBlock
+    let blockToKeep = startBlock
+
+    // We keep the style of the blockToKeep. endBlock should keep style if deleting empty line.
+    if (startBlock.isEmpty && !endBlock.isEmpty) {
```
vpontis: I actually think this should probably be just `if (startBlock.isEmpty)`.

When you have:

```
<blank p>
<blank h1>
```

And you delete on the `h1` should the new node be `h1` or `p`? I'm thinking `h1`...

src/transforms/at-range.js
```diff
@@ -113,14 +113,23 @@ Transforms.deleteAtRange = (transform, range, options = {}) => {
...
+    blockToRemove.nodes.forEach((child, i) => {
+      const newKey = blockToKeep.key
+      const newIndex = blockToKeep.nodes.size + i
       transform.moveNodeByKey(child.key, newKey, newIndex, OPTS)
     })
 
-    // Remove parents of endBlock as long as they have a single child
-    const lonely = document.getFurthestOnlyChildAncestor(endBlock.key) || endBlock
+    // Remove parents of blockToRemove as long as they have a single child
+    const lonely = document.getFurthestOnlyChildAncestor(blockToRemove.key) || blockToRemove
```
vpontis: @ianstormtaylor what does this do?

src/transforms/at-range.js
ianstormtaylor: I'm down with that!

src/transforms/at-range.js
ianstormtaylor: It retrieves the highest ancestor of `blockToRemove` that has only a single child, (in this case probably an empty text node, not sure).

vpontis: Ah got it. That makes sense. You are moving all of the children so you want to delete the new part of the tree that is empty.

src/transforms/at-range.js
vpontis: Ah so apparently when you press backspace on an empty node, it doesn't call thru to `deleteAtRange`. It must go thru another transform.

vpontis: So that could be filed as another issue. Let me know if you want to do that. Not sure the level of granularity you are going for.

vpontis:
@skogsmaskin, I'm not caught up on the testing framework yet so I'll need to spend time to do that first.

ianstormtaylor:
This is fixed in the latest release, thanks @vpontis!

PR #740, opened 2017-04-21
Clarify instructions for plugin preventing default behavior

oyeanuj:
A small clarification to ensure no one else misunderstands how to prevent default behavior.


ianstormtaylor:
Thanks @oyeanuj!

PR #748, opened 2017-04-24
Add a badge for gzip size

conorcussell:
Couldn't work out how to mix github markdown + html so just copied the rendered version of this `gzip size`

Also wasn't sure what (if anything) it should be linked to? So linked to the file we are weighing. Let me know if you have any other ideas.


ianstormtaylor:
Linking to the file sounds smart to me! Thanks man :)

Issue #746, opened 2017-04-24
inline node at top level in document?


bodyflex:
do the top level nodes in a document always have to be block nodes?

when I initialise a state with:
```js
Raw.deserialize({
  nodes: [
    {
      kind: 'inline',
      type: 'input',
      nodes: [{
        kind: 'text',
        text: ''
      }]
    }
  ]
}, {terse: true})
```
I get:
```
Cannot read property 'nodes' of undefined
TypeError: Cannot read property 'nodes' of undefined
    at Object.validate (eval at 1565 (http://localhost:9001/0.5e3e7a34a61aac7d5516.hot-update.js:6:2), <anonymous>:89:41)
    at eval (eval at <anonymous> (http://localhost:9001/static/preview.bundle.js:3342:2), <anonymous>:123:22)
    at Array.find (eval at <anonymous> (http://localhost:9001/static/preview.bundle.js:4854:2), <anonymous>:1181:20)
    at Schema.__validate (eval at <anonymous> (http://localhost:9001/static/preview.bundle.js:3342:2), <anonymous>:119:30)
    at Document.validate (eval at <anonymous> (http://localhost:9001/static/preview.bundle.js:5990:2), <anonymous>:1650:19)
    at Document.object.(anonymous function) [as validate] (eval at <anonymous> (http://localhost:9001/static/preview.bundle.js:6002:2), <anonymous>:109:30)
    at iterate (eval at <anonymous> (http://localhost:9001/static/preview.bundle.js:16767:2), <anonymous>:215:21)
    at normalizeNode (eval at <anonymous> (http://localhost:9001/static/preview.bundle.js:16767:2), <anonymous>:243:3)
    at normalizeNodeAndChildren (eval at <anonymous> (http://localhost:9001/static/preview.bundle.js:16767:2), <anonymous>:182:5)
    at Object.Transforms.normalizeNodeByKey (eval at <anonymous> (http://localhost:9001/static/preview.bundle.js:16767:2), <anonymous>:77:3)
```

however, when I change the kind from inline to block, it works.

ianstormtaylor:
Hey @bodyflex, yup that's one of the current limitations of the schema—the top-level nodes have to be `Block` nodes. I'm open to loosening this down the road, but for now it would be a lot of work to change.

PR #755, opened 2017-04-25
add note about only allowing block nodes as direct descendants of document

bodyflex:
can perhaps spare some time from someone else wondering about #746


ianstormtaylor:
Thanks @bodyflex!

PR #759, opened 2017-04-26
Add "data-key" to root div for the whole document

SamyPesse:
The goal is to be able to use `findDOMNode` with the `document` itself.

The tests were failing since https://github.com/ianstormtaylor/slate/commit/a98b687aa538138be427f419ac5489cdbf72d51c. Since PR also fixes them.


ianstormtaylor:
Beautiful, thanks @SamyPesse!

Issue #612, opened 2017-02-19
Pressing return between text and inline block duplicates the inline block


michaelpietrykowski:
This seems to be 'the opposite' of #466 
## Unexpected Behavior
When the cursor is between some text and an inline block and a newline is inserted, the inline block is duplicated. This can be seen in the emoji example:
![possible_bug](https://cloud.githubusercontent.com/assets/8776388/23105924/a07f1bb8-f6e6-11e6-8949-6710f0949228.gif)
## Expected Behavior
When the inline block stands by itself, behavior is as expected.
![expected](https://cloud.githubusercontent.com/assets/8776388/23105975/1fcd3256-f6e7-11e6-8386-88bb530fd864.gif)

ianstormtaylor:
This issue hasn't had a response in a while, and it's unclear if it's still present in the latest versions of Slate. To keep the issues easier to manage, since there are so many being opened every day, I'm going to close this old one out. 

But if you notice that the bug is still happening, please feel free to comment and it can be revisited. It might be an edge case that not many people run into, in which case the fastest way to get it fixed would probably be to write a pull request for it.

Thanks for understanding!

tobiasandersen:
I can confirm that this issue is still present (just tried on demo site).

ianstormtaylor:
This is fixed now!

PR #761, opened 2017-04-27
Dont reapply void inline blocks

kellyjosephprice:
This is a workaround for a bug in upstream slate.

https://github.com/ianstormtaylor/slate/issues/612


kellyjosephprice:
I didn't mean to make this PR against `ianstormtaylor:master`

PR #764, opened 2017-04-27
Fix #763: Add prop "props" to pass custom props to nodes

### LINKED ISSUE #763: Passing custom props to nodes
In our application, some of our node components requires some props from the parent.

The solution we tried was to create a schema depending on these props:

```js
function makeSchema(commonProps) {
    return {
        nodes: {
            paragraph: (props) => (
                <Paragraph {...props} {...commonProps}>
                    {props.children}
                </Paragraph>
            )
        }
    };
}
```

But Slate doesn't re-render when the `schema` changed.

So I'm proposing 3 solutions:

1. Adapt the `shouldComponentUpdate` to re-render when the `schema` changed
    - **Pro:** it seems coherent
    - **Cons:** How to compare `prevProps.schema` with `props.schema` ? If the schema is inlined in the `render()` method, and so unique to each `render` call
2. Pass all other props of `Editor` to the nodes
    - **Cons:** we have to avoid conflicts between prop's names
3. Add a prop `props` to `Editor`: `<Editor props={{ customPropToPassToNode: 'test' }} ... />`

@ianstormtaylor @Soreine what do you guys think ?

SamyPesse:
This PR fixes #763 by adding a prop `props` that can be used to pass a set of props to all nodes (inline, marks and blocks).

The nodes are re-rendered if `props` changes. It uses `fbjs/lib/shallowEqual` to compare the props.

- [x] Add prop `props`
- [x] Re-render if the prop changed
- [x] Documentation
- [x] Unit tests

### DIFF
--- a/docs/reference/components/editor.md
+++ b/docs/reference/components/editor.md
@@ -16,6 +16,9 @@ The top-level React component that renders the Slate editor itself.
   - [`readOnly`](#readonly)
   - [`state`](#state)
   - [`style`](#style)
+  - [`tabIndex`](#tabindex)
+  - [`role`](#role)
+  - [`props`](#props)
 - [Placeholder Properties](#placeholder-properties)
   - [`placeholder`](#placeholder)
   - [`placeholderClassName`](#placeholderclassname)
@@ -99,6 +102,10 @@ Indicates if it should participate to [sequential keyboard navigation](https://d
 ### `role`
 
 ARIA property to define the role of the editor, it defaults to `textbox` when editable.
+
+### `props`
+
+Custom properties to pass to all node components. An update of these props re-renders all the nodes.
 
 ## Placeholder Properties
 
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -52,6 +52,7 @@ class Content extends React.Component {
     onKeyDown: React.PropTypes.func.isRequired,
     onPaste: React.PropTypes.func.isRequired,
     onSelect: React.PropTypes.func.isRequired,
+    props: React.PropTypes.object,
     readOnly: React.PropTypes.bool.isRequired,
     role: React.PropTypes.string,
     schema: React.PropTypes.object,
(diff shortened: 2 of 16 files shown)


SamyPesse:
One remaining question is: should we pass the props to `Component` as `props={props}` or {...props}` ?

oyeanuj:
@SamyPesse Thanks for starting this, I ran into something similar as well. 

Two suggestions around the current solution:
1. Could we name it `editorProps` or something equally descriptive? `prop.props` feels a bit weird to type 😄 

2. With the current proposed solution, any change in prop would re-render all the nodes. What do you think about having props be a nested object, where the keys are the name of the nodes? 
In that case, in `shouldComponentUpdate` only if `editorProps.bold` changes, then only the `bold` node is updated. This can also make it an efficient way to communicate from the editor to specific type of nodes.

SamyPesse:
@oyeanuj 

1. Yes, we can probbaly change the name, but I'm not a big fan of `editorProp`, maybe `nodeProps` ? I'll wait for @ianstormtaylor opinion on this PR before potentially changing the prop name.

2. I disagree. From my point of view, these props should be used as a "global context". For GitBook, we want to have this PR merged to fix https://github.com/ReactTraining/react-router/issues/5037 because the React context can't be used for this (https://github.com/facebook/react/issues/2517).
  What is your use case ?

oyeanuj:
@SamyPesse, on the second point, it is good to understand your usecase. In my usecase, while fetching some metadata for a url (and couple of other similar usecases), it is good to know when those operations are happening via plugin without connecting individual node components to a store. 

So, it seems like we are maybe talking about two different functionalities - 

1. Communicating from the editor to **all** nodes ('global context', your use-case).
2. Communicating from the editor to **specific** nodes (my use-case).

And I guess this PR seemed very close to achieving the second functionality as well. Do you think it should be part of this PR?

@ianstormtaylor, thoughts?

Issue #737, opened 2017-04-20
Clarify Placeholder props


oyeanuj:
Hi @ianstormtaylor! In the [`Placeholder` docs](https://docs.slatejs.org/reference/components/placeholder.html), we have two props: `node` and `parent`.

My understanding based on the docs is that the former is used to determine what to render and position on the top of, and the latter to determine the node whose content triggers the `Placeholder`. I imagine they are usually the same. I ask because with that assumption, the `Placeholder` doesn’t render.

```jsx
//In the node being rendered
<div {...attributes}>
  <Placeholder node = {node} parent = {node} state = {state} >
    {placeholderText}
  </Placeholder>
  {children}
</div>
```

It fails in the `isVisible` function here - https://github.com/ianstormtaylor/slate/blob/master/src/components/placeholder.js#L55

So, I am not sure what to pass as the `node` prop. The text node of the node for which it is a placeholder?

Alternatively, I feel this is a place where the `firstOnly` prop that [you referred to](https://github.com/ianstormtaylor/slate/issues/66#issue-164915571) can be used. I'm happy to make a PR that checks for the line linked above in `isVisible` only if `firstOnly` is set to true. 

Thoughts?

ianstormtaylor:
Hey @oyeanuj sorry for the delay on this.

You've got the `parent` and `node` understanding correct. I think `parent` should actually be renamed to `target`, since that might be more helpful in understanding it.

I'm not sure exactly what you mean though about what's not working. What is your use case? The default case is that `node` might be something like a `Block(type:paragraph)` and the `parent` would be `Document`.

oyeanuj:
Thanks for the clarification! I think it helps me see the root issue. I think this part of the doc confused me:

**`parent`**: The node to check for non-empty content...
**`node`**: The node to render the placeholder element on top of...

Let's say we have a structure like the one below and we want a placeholder for each block if they are empty.
```yaml
Document
-- FirstBlock
-- SecondBlock
-- ThirdBlock
```
If `parent` is supposed to refer to the Document, then I am not sure if `target` helps.. it might be more confusing since in the above case, Document is not the `target`.

It seems like there are two distinct cases -
1. `Placeholder` if the document is empty.
2. `Placeholder` if a particular node is empty.

Maybe I can modify the PR #765 to use the `parent` prop only if `firstOnly` is true. Does that work for you?

ianstormtaylor:
[...]

Issue #66, opened 2016-07-11
add <Placeholder> component


ianstormtaylor:
I think this could be block-independent, and take in a `node` and `state`. Then people can use it and add styles as they please.

Also, if it also took a `firstOnly={false}` property, you could use it for cases like image caption nodes, where you want the placeholder to render even though it's not the first node in the tree.

oyeanuj:
@ianstormtaylor I didn't see the `firstOnly` property you mentioned above, in the source code. Is there an alternative way that you'd suggest implementing that functionality?

ianstormtaylor:
I think as currently implemented you'd pass in a `parent` node that isn't the `document` to get the placeholding to trigger on its contents. Although that name isn't great

PR #765, opened 2017-04-27
Add firstOnly prop to Placeholder

oyeanuj:
Hi @ianstormtaylor, adding the `firstOnly` prop to Placeholder as you originally envisioned in #66. 

This also addresses part of the problem I raised in #737, although it still needs clarification from you for the intended values of `node` and `parent`.

As always, happy to make any edits needed! 

PS. Not sure how to get rid of those extra merge commits that come in with every PR - sorry about that!


src/components/placeholder.js
```diff
@@ -37,7 +48,8 @@ class Placeholder extends React.Component {
     return (
       props.children != this.props.children ||
       props.className != this.props.className ||
-      props.parent != this.props.parent ||
+      props.firstOnly != this.props.firstOnly ||
+      (props.firstOnly && (props.parent != this.props.parent)) ||
```
ianstormtaylor: Why the need to only update when `parent` changes if `firstOnly == true`?

src/components/placeholder.js
```diff
@@ -50,11 +62,18 @@ class Placeholder extends React.Component {
...
-    return false
+
+    if (firstOnly) {
+      if (parent.nodes.size > 1) return false
+      if (parent.nodes.first() === node) return true
+      return false
+
+    } else {
+      return true
+
```
ianstormtaylor: Can you remove the extra line breaks here? Thanks!

src/components/placeholder.js
oyeanuj: This is because `parent` prop is used in `Placeholder` only when `firstOnly` is set to true. So this would prevent a needless re-render (super edgecase where someone has `firstOnly` set to false and is passing parent accidentally and that changes).

src/components/placeholder.js
oyeanuj: Done!

src/components/placeholder.js
ianstormtaylor: Let's remove it for now, since it's such an edge case, just to keep it simple.

oyeanuj: done!

PR #768, opened 2017-04-28
Splitting Before a Void Inline Block Adds a Duplicate

kellyjosephprice:
Related Issue: #612

### DIFF
--- a/src/models/node.js
+++ b/src/models/node.js
@@ -1399,7 +1399,10 @@ const Node = {
         // Try to preserve the nodes list to preserve reference of one == node to avoid re-render
         // When spliting at the end of a text node, the first node is preserved
         let oneNodes = nodes.takeUntil(n => n.key == one.key)
-        oneNodes = (oneNodes.size == (nodes.size - 1) && one == nodes.last()) ? nodes : oneNodes.push(one)
+
+        if(!one.isVoid) {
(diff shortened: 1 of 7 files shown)


kellyjosephprice:
I'm still unfamiliar with the code base, so I wanted to start with small changes to confirm I'm moving in the right direction.

ianstormtaylor:
Hey @kellyjosephprice this is awesome!

Could you add a test for splitting block nodes as well? Since it looks like this will apply to both cases. Thank you!

tobiasandersen:
Hey, I just tried this out. It does indeed fix the issue with duplicated inline nodes, but it also introduces some new problems:

1. When splitting on an inline node, the inline node moves down, but the cursor stays.
2. Performing `undo` in this state breaks the editor.

The problem with undo arrises when joining the two nodes, specifically on [this line](https://github.com/ianstormtaylor/slate/blob/master/src/models/node.js#L1398). The reason it breaks is that the first node this time is a `Text` node (which doesn't have the `nodes` property). (The bug we're trying to solve here is actually masking this issue, since the first node in that case will be of kind `Inline`.) What's interesting to me is that joining the two blocks using backspace works just fine. 

I'm also completely unfamiliar with the code base (although I've been using Draft extensively for some time), but I'll see if I can track it down even further.

ianstormtaylor:
This is fixed now!

PR #782, opened 2017-05-04
`moveNode` operation fix

AlbertHilb:
`moveNode` moves a `node` from its old `parent` to a new parent `target` with path `newPath`.

When `parent` is an ancestor of `target` and the index of `node` is less than the index of the ancestor of `target` with the same depth of `node`, removing `node` changes the path to `target`.
As consequence of that, `newPath` needs to be adjusted.

### DIFF
--- a/src/transforms/apply-operation.js
+++ b/src/transforms/apply-operation.js
@@ -188,21 +188,39 @@ function moveNode(state, operation) {
   let { document } = state
   const node = document.assertPath(path)
   const index = path.pop()
+  const parentDepth = path.length
 
   // Remove the node from its current parent
   let parent = document.getParent(node.key)
   parent = parent.removeNode(index)
   document = parent.kind === 'document' ? parent : document.updateDescendant(parent)
 
-  // When `parent` is an ancestor of `target` and `node` index is less than the
-  // index of the `target` ancestor with the same depth of `node`, then removing
-  // `node` changes the path to `target`. So we have to adjust `newPath`.
-  const d = path.length
-  const isAncestor = d === 0 ? true : path.every((x, i) => x === newPath[i])
-  if (isAncestor && index < newPath[d]) newPath[d]--
+  // Check if `parent` is an anchestor of `target`
+  const isAncestor = path.every((x, i) => x === newPath[i])
+
+  let target
+
+  // If `parent` ia an ancestor of `target` and they have the same depth,
+  // then `parent` and `target` are the same node.
+  if (isAncestor && parentDepth === newPath.length) {
+    target = parent
+  }
+
+  // Else if `parent` is an ancestor of `target` and `node` index is less than
+  // the index of the `target` ancestor with the same depth of `node`,
+  // then removing `node` changes the path to `target`.
+  // So we have to adjust `newPath` before picking `target`.
+  else if (isAncestor && index < newPath[parentDepth]) {
+    newPath[parentDepth]--
+    target = document.assertPath(newPath)
+  }
+
+  // Else pick `target`
+  else {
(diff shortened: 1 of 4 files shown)


src/transforms/apply-operation.js
```diff
@@ -187,19 +187,24 @@ function moveNode(state, operation) {
...
-  const index = parent.nodes.indexOf(node)
   parent = parent.removeNode(index)
-  document = isParent ? parent : document.updateDescendant(parent)
+  document = parent.kind === 'document' ? parent : document.updateDescendant(parent)
+
+  // When `parent` is an ancestor of `target` and `node` index is less than the
+  // index of the `target` ancestor with the same depth of `node`, then removing
+  // `node` changes the path to `target`. So we have to adjust `newPath`.
+  const d = path.length
+  const isAncestor = d === 0 ? true : path.every((x, i) => x === newPath[i])
```
ianstormtaylor: We can actually remove the ternary here since `.every()` will return true with arrays of zero length.

src/transforms/apply-operation.js
```diff
@@ -187,19 +187,24 @@ function moveNode(state, operation) {
...
+  if (isAncestor && index < newPath[d]) newPath[d]--
```
ianstormtaylor: Hey @AlbertHilb I think I might just be out of touch with this part of the codebase, but can't `path` and `newPath` be the same length? How do we ensure that `d` always refers to an existing point in the `newPath`?

src/transforms/apply-operation.js
AlbertHilb: Ok

src/transforms/apply-operation.js
AlbertHilb: If `path` and `newPath` have same length then the second test in the `if` condition becomes
```
index < undefined
```
and it is always `false`, so `newPath`remains untouched.
Anyway, I rewrote the code to more explicitly deal with this case.

ianstormtaylor:
Thanks @AlbertHilb this is seriously awesome!

KohheePeace:
Hi @AlbertHilb and @ianstormtaylor .I really appreciate your great work. Really really thank you.

I think the bug related to the "Undo/Redo" will appear after this pull request. 
 

link for full-size gif: https://giphy.com/gifs/gqu1g7L8usVVe/fullscreen

Error console message: 
```
build.prod.js:18309 Uncaught TypeError: Cannot read property 'removeNode' of null
    at moveNode (http://slatejs.org/build.prod.js:18309:18)
    at Object.Transforms.applyOperation (http://slatejs.org/build.prod.js:18143:21)
    at Transform.(anonymous function) [as applyOperation] (http://slatejs.org/build.prod.js:15178:32)
    at http://slatejs.org/build.prod.js:21501:15
    at Array.forEach (native)
    at Object.Transforms.redo (http://slatejs.org/build.prod.js:21500:8)
    at Transform.(anonymous function) [as redo] (http://slatejs.org/build.prod.js:15178:32)
    at onKeyDownZ (http://slatejs.org/build.prod.js:16064:61)
    at Object.onKeyDown (http://slatejs.org/build.prod.js:15748:16)
    at Stack.(anonymous function) [as onKeyDown] (http://slatejs.org/build.prod.js:13661:37)
```

When I change the `function moveNode(state, operation){...}`  in `transforms/apply-operation.js` to the previous commit one, this error has disappeared. 

Im beginner about coding so..., If you have the time, I would like you to check the code. Thanks.

AlbertHilb:
Sorry, my fault.
But a fix is on the way! :blush:

KohheePeace:
Really thank you so much! 😄

Issue #775, opened 2017-05-03
node.insertText is not a function


oyeanuj:
Hey @ianstormtaylor, it seems like when trying to insert text inside a node using `insertTextByKey`, I am getting this error. On searching through the repo, I can't seem to find `insertText` as a function on the `node` model either.

Here is the line where it is failing: https://github.com/ianstormtaylor/slate/blob/master/src/transforms/apply-operation.js#L119

And here is a simple JSFiddle illustrating it (insert text in the default node): https://jsfiddle.net/c77z6ax4/2/

ianstormtaylor:
I believe this is happening because `insertTextByKey` expects a `key` that refers to a Text node, but it's probably getting passed a Block or Inline.

I'd be open to a pull request that allows it to handle either though.

oyeanuj:
@ianstormtaylor you were right, passing a text node works. I've issued a PR that clarifies the doc as well. 

As for allowing other node, to be honest, I don't feel confident enough to mess within the core but I will try to come back to it. But On a quick investigation, it seems that the node model would need to implement an `insertText` method which would select the first text of the node and then call `Text.insertText` on that text node. 

Does that sound right?

ianstormtaylor:
@oyeanuj I think instead if implementing `insertText` on all nodes, we'd instead want to use the key and offset to fetch the leaf text node with updated offset, and then run the insert text operation on that new leaf text node.

oyeanuj:
Thanks @ianstormtaylor! Implemented in #784.

PR #784, opened 2017-05-04
Accept non-text nodes in insertTextByKey

oyeanuj:
Hi @ianstormtaylor, as per your comments in #775, this function now accepts a non-text node and calculates the leaf text node at the offset, and then proceeds to insert text.


ianstormtaylor:
Yay! Great work @oyeanuj. Can you add a test case that covers it and then I'm super down to merge?

oyeanuj:
@ianstormtaylor Thanks! Added a test, let me know if it looks good!

ianstormtaylor:
Hey @oyeanuj looks like a good start. I think we need to edit the test though to handle the more complex case. 

Right now, since the text node being inserted into is the first and only node, the `offset` won't change when the new `key` is computed. But we should change the test case to be a block with `text, inline, text` and inserting into the last text node. Then i think we'll end up with a failing test case, since the `offset` will need to be recomputed to be relative to the new `key`. (Which you should be able to do adjust using the return value from `block.getOffset(textKey)` I think.

Let me know how that goes!

oyeanuj:
@ianstormtaylor I added a set up for a[ test case like you suggested](https://github.com/ianstormtaylor/slate/pull/784/files?diff=unified#diff-a2f06c47a9b53b7d1cd3b0c66a9f2b50). Can you take a look to make sure it reflects what you were thinking?  If the test case looks good, I'll do the rest of testing it and fixing the failing case.

ianstormtaylor:
@oyeanuj Looks good!

I think we should also change the tests to not use `insertBlock`. Instead, they can use `document.getBlocks().first()` or similar to get the block they need, and insert text into it. That way they are doing less.

oyeanuj:
Hey Ian, so I added the test case that we discussed and it didn't fail. So clearly I misunderstood your comment above - can you take a look and tell me what were you expecting instead?

ianstormtaylor:
Hey @oyeanuj I'm not sure what I was talking about.

I'm sorry for the slowness on this. Since this has been opened, it seems like we've had other issues where using offsets isn't the most robust, so I think I'd like to wait to merge this until we get those figured out. Right now we'd been trying to remove all offset-use-across-depths from the codebase I believe, so I'd rather not reintroduce one yet.

oyeanuj:
@ianstormtaylor That makes sense, happy to wait!

ianstormtaylor:

PR #788, opened 2017-05-04
Copy void

ianstormtaylor:
Hey @oyeanuj thank you so much for your work in #676! 

I realized that it was going to require a lot more digging and rethinking how copy/cut worked, so I pulled in your changes and went from there.


oyeanuj:
This is awesome, and far more complicated than I had imagined! Thanks!

oyeanuj:
@ianstormtaylor Btw, I tried it on the Images example page and getting the following error after trying to copy-paste the image within the editor.
```
Uncaught TypeError: Cannot read property 'split' of null
    at onPasteText (http://slatejs.org/build.prod.js:15967:14)
    at Object.onPaste (http://slatejs.org/build.prod.js:15934:16)
```

ianstormtaylor:
@oyeanuj shit you're right. Weirdly if you click the image again and then copy it it seems to work. I'll investigate!

ianstormtaylor:
@oyeanuj fix incoming!

PR #792, opened 2017-05-05
Add tests for plugins

mxstbr:
I set out to provide a failing test for some weird behaviour I'm seeing with a custom plugin I'm writing where the `rules` of a plugin schema are not ran. This is JSFiddle reproducing the issue: https://jsfiddle.net/8y30hx90/ (note how the `alert()` is never called)

For some reason the tests passes, so I guess I'm doing something wrong? Time to quadruplecheck... Anyways, at least you now have a first test for plugins!

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -11,6 +11,7 @@ import 'babel-polyfill'
  * Tests.
  */
 
+import './plugins'
 import './rendering'
 import './schema'
 import './serializers'
--- a/test/plugins/fixtures/custom-rules/index.js
+++ b/test/plugins/fixtures/custom-rules/index.js
@@ -16,12 +16,13 @@ function decorate(text, block) {
 }
 
 export const plugins = [{
-  match: () => true,
-  decorate,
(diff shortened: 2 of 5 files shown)


ianstormtaylor:
Awesome! Thank you @mxstbr 😄

mxstbr:
I still don't understand why the JSFiddle doesn't work though: https://jsfiddle.net/8y30hx90/ (note how the alert() is never called)

Any ideas?

ianstormtaylor:
@mxstbr if I update it to have a decorator function that matches anything then it seems to work: https://jsfiddle.net/8y30hx90/1/ — I think just matches without other parts of the rule will never be called because they don't do anything?

mxstbr:
I'm so bad, it turns out it didn't work because I accidentally typo'ed `rule.decorator` rather than `rule.decorate` and didn't notice until copy and pasting your working code. Ugh, thanks for helping me resolve that!

PR #797, opened 2017-05-06
Make rules apply to everything if no match property is provided

mxstbr:
Wanted to kick off a discussion, I don't see any reasons to throw an error if a rule doesn't define `match`. (which is what currently happens)

This commit changes that behaviour by making rules match everything by default. Maybe there's a reason for the current behaviour, but I figured it'd be worth discussing!

### DIFF
--- a/docs/reference/models/schema.md
+++ b/docs/reference/models/schema.md
@@ -100,7 +100,7 @@ Internally, the `marks` and `nodes` properties of a schema are simply converted 
 }
 ```
 
-Slate schemas are built up of a set of rules. Each of the properties will add certain functionality to the schema, based on the properties it defines. 
+Slate schemas are built up of a set of rules. Each of the properties will add certain functionality to the schema, based on the properties it defines.
 
 ### `match`
 `Function match(object: Node || Mark)`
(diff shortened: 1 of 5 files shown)


ianstormtaylor:
Hey @mxstbr! What's your use case for matching everything? On my phone now but I'd need to look deeper. 

I think since rules can match the document itself, block and inline nodes, and text nodes, (and in the future maybe even the selection and the state) it's probably going to be a weird case where you'd want to match all three?

mxstbr:
> What's your use case for matching everything?

Literally the official markdown example: https://github.com/withspectrum/slate-markdown

Maybe this is less common than I thought/doesn't make sense, to me it seemed unnecessary to _have_ to specify `match: () => true`.

ianstormtaylor:
Ah shoot yeah that example should be changed, not future proof.

ianstormtaylor:
I'm going to close this one, because I'd like to extend the schema matches to work with selections, and potentially marks as well. Which will end up making `match: () => true` a very weird behavior I think. Better to have it be explicit if people really do want that.

PR #798, opened 2017-05-06
Fix state.json in examples/links

oh3vci:
Because of comma, this json file occured an syntax error. I removed it.
Below shows the error message.
```
Error: Parse error on line 18:
...hyperlinks"					}, ]				},				{					
----------------------^
Expecting 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '[', got ']'
```

### DIFF
--- a/examples/links/state.json
+++ b/examples/links/state.json
@@ -18,7 +18,7 @@
             {
               "kind": "text",
               "text": "hyperlinks"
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thank you @oh3vci!

PR #801, opened 2017-05-06
Docs improvements suggestions

asfktz:
Hi!
Thanks for taking the time to build such an awesome project!

I'm reading the documentation (which is great by the way) and thought to take the time to send a few PRs for some typos and things that wasn't immediately clear for me as a first time reader.

This is the first one.

Cheers,
Asaf K

### DIFF
--- a/docs/walkthroughs/adding-event-handlers.md
+++ b/docs/walkthroughs/adding-event-handlers.md
@@ -7,7 +7,7 @@
 
 Okay, so you've got Slate installed and rendered on the page, and when you type in it, you can see the changes reflected. But you want to do more than just type a plaintext string.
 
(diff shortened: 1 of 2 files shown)


asfktz:
In `Using Plugins`, is the 'schema' part of the plugin is left outside intentionally?
As I read it for the first time I felt a bit confused because the schema seem like integral part of the plugin, and seeing it left outside seemed less encapsulated.

Will it be useful if I add a paragraph about moving the `schema` intro the plugin?

ianstormtaylor:
Hey @asfktz I'm really sorry for the delay here. This PR is great! And your note about the schema in Using Plugins is also a really good one. I think that part of the walkthrough might have been written before the schema existed, but I agree with you! I'd be super glad to accept a PR that made that section better

Issue #731, opened 2017-04-19
Add sample code for inserting inline


thepian:
There seems to be no examples of inserting inline. I looked at examples and tests for insertInline. I tried the insertInline call, and nothing seemed to work. Then I tried to insert with a serialiser and insertFragment, which didn't work either.

I could only get it to work when inserting a block.

In general the tests are very limited and seem like a warning sign for the project for people evaluating. I think adding tests that cover the API well would be very helpful to show how to use Slate.

ianstormtaylor:
Hey @thepian, happy to accept PRs that add tests. There are quite a few for transforms, but testing the browser interaction with `contenteditable` is a bit harder to mock, so I haven't dug into it yet.

ianstormtaylor:
Closing this since it's old. There are examples of inserting inlines I believe (emoji, link, etc.) and if anyone wants to help with the docs or tests they are welcome!

PR #804, opened 2017-05-07
Failing test for insertInline

oyeanuj:
Hey @ianstormtaylor, I ran into the #731 as well. I wasn't immediately sure about what the problem might be, so here is a failing test.

cc: @thepian @mxstbr


ianstormtaylor:
Hey @oyeanuj I'm going to close this since no one has had time to look into it it seems. If you're down to open it up again with a full fix I'd love that though! Thank you for your help for for understanding!

PR #808, opened 2017-05-09
update the examples/rich-text, be able to active ol&ul button

coocooooo:
In the examples/rich-text, the ol&ul button should be active, when user navigate to correspond DOM


ianstormtaylor:
Hey @coocooooo sorry for the delay. I think this change! But can you move the logic to be inside the `hasBlock` method instead of inside the `renderBlock` one? Thank you!

davidchang:
Sorry, I didn't see this PR and made a fix that was merged here: https://github.com/ianstormtaylor/slate/pull/1803

This PR should be closed.

PR #814, opened 2017-05-14
Hacky: Remove cheerio

natew:
Ok, just wanted to test out if this was doable.

Parse5 has a totally different DOM output from cheerio.

Luckily! Cheerio already uses the much cooler looking https://github.com/fb55/htmlparser2/

So I started by testing a hacky version of it, and it seems to be close but not exact. Will need to explore what options cheerio uses for it. But this should e a big win!

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,12 +6,12 @@
   "repository": "git://github.com/ianstormtaylor/slate.git",
   "main": "./lib/index.js",
   "dependencies": {
-    "cheerio": "^0.22.0",
     "debug": "^2.3.2",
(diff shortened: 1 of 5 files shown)


natew:
What I'm getting is that htmlparser adds newline elements where cheerio seemed not to.

lxcid:
Want to note that this will affect me, why are we avoiding cheerio?

natew:
Actually looking at this more, cheerio is a little heavy but a lot of it actually comes from htmlparser2. So hard to get around that.

PR #822, opened 2017-05-17
Local examples url update

AverageZ:
When attempting to mess around locally I could never get anything to update. Sure enough I was looking in the wrong place and had missed the `/dev.html` on the end of my url.

Here's to hoping no one else struggles with this again.


ianstormtaylor:
Ah, thank you @AverageZ! Good call.

PR #828, opened 2017-05-22
Avoid unnecessary renders and normalization

Soreine:
I noticed in my app that, even if every props I passed to `Slate.Editor` had not changed, there was a schema normalization call.

I thought we had fixed these problems long ago. I could not find again when, in Slate, the normalization is called. I could only find the `Content.shouldComponentUpdate`, which looked all right. So I simply wrapped the Editor in a `React.PureComponent` and now it updates as needed.

### DIFF
--- a/src/components/editor.js
+++ b/src/components/editor.js
@@ -55,7 +55,7 @@ const PLUGINS_PROPS = [
  * @type {Component}
  */
 
-class Editor extends React.Component {
+class Editor extends React.PureComponent {
 
   /**
    * Property types.


ianstormtaylor:
Hey @Soreine I'm curious to know why you're seeing these extra re-renders, and potentially fix the root cause instead of covering it up. Although I'm open to using `PureComponent` if we need to after we've figured that out.

ianstormtaylor:
Hey @Soreine I think with a lot of the changes in core this will be solved differently now, and might be solved in a lower-level way from other bugs. So I'm going to close this out for now.

Issue #709, opened 2017-04-03
Let plugins extend `State` schema


AlbertHilb:
Plugins add new functionalities to the editor. Editor `State`, therefore, should be able to keep also data related to these new functionalities.
To this aim, maybe, it suffices to add a new field to `State`, an Immutable `Map`, which can be used by plugins to store their own data.

What do you think?

Rebulus:
My additional, it would be nice, if this state were personalized by plugin for excluding intersections of their state (like personal name spacing) 😊

ianstormtaylor:
What kinds of use cases are you thinking about?

Rebulus:
For example, we have plugin like [hovering menu](http://slatejs.org/#/hovering-menu) and we want to add a link functionality. It could be a button in hovering menu. When an user click on the button, then a link popover will be opened.
Currently, for this behavior, we should create a closure in the plugin and place there a state of opening the link popover. Of course, we should find a way, how to clear this state, when it is not needed. It would be cool, if we have some functionality to do that with using plugin system.
Sorry, if an example is not so good, but it is first, what I invented 😊

Rebulus:
The main goal of this, I think, to find way to have communication between events and `render`.

ianstormtaylor:
Gotcha, I think that makes sense to me! I'd be open for a pull request that adds a `state.data` property that mimics the existing `.data` properties of `Nodes`.

AlbertHilb:
Perfect! :blush:
I'm going to create it in the next days.

oyeanuj:
This would be great! Another usecase that I was thinking of would be using it as a central list/repository of things (like links, mentions, languages, font, etc) that are like metadata for the document. (I'm assuming it will be part of the raw JSON that is serialized/deserialized).

Thanks @AlbertHilb for suggesting this!

soulwire:
Any progress on this @AlbertHilb? This kind of behavior would be very useful for something I'm working on too...

oyeanuj:
@AlbertHilb I'm not sure your current direction of how this state would exist but I wanted to throw out an idea for you to consider. We could extend the state's data to be an object like:
```js
stateData: {
  document: { ... },
  paragraphNode: { ... },
  headingNode: { ... }
}
```

and any change in `document` props would re-render the editor, whereas any change in any other key would only re-render the nodes of that relevant type.

PR #830, opened 2017-05-22
Add `data` field to editor `State`.

AlbertHilb:
This PR is related to issue #709.

It adds a `data` field to editor `State`, an `immutable` `Map` that can be used by plugins to store their own internal state.

`Plugin`s gain three new optional properties: `data`, `serialize` and `deserialize`:

- The first is a data structure of the type that can be passed to the `immutable` `Map` constructor.
If present, it represents the default values of the plugin state.

- The second is a function, if present it is called by `Raw` to serialize `data`.
It receives the `State` `data` as parameter and should return a JSON object representation of the plugin internal parameters that should be preserved in the process.

- Also the third is a function. It is the inverse of `serialize`, it should return a `Map` representing the plugin internal state associated to the JSON object passed as parameter.

Other changes involve the `State` `create` method and `serialize` and `deserialize` methods of the `Raw` module.

- `create`, now, recognizes a `plugins` property in the `options` object, an array of `Plugin`s scanned in search for `data` parameters to build `State` `data`.

- The optional object passed to `serialize` can contain a boolean parameter `preserveData` and an array of plugins. If the former is `true` then the `serialize` functions, from the plugins in the array, are used to serialize the `State` data.

- Finally, the `deserialize` functions, from the plugins array property of the `options`object passed to `deserialize`, are used to reconstruct the editor `State` `data`.

As always I'm open to any change you think it's needed. :blush:


oyeanuj:
@AlbertHilb This is great, thank you for the PR. I wanted to bring up the idea that I had thrown in the [original issue here](https://github.com/ianstormtaylor/slate/issues/709#issuecomment-299327883):

> We could extend the state's data to be an object like:
```jsx
stateData: {
  document: { ... },
  paragraphNode: { ... },
  headingNode: { ... }
}
```
>and any change in document props would re-render the editor, whereas any change in any other key would only re-render the nodes of that relevant type.

So your current `data` that you added in the PR, by default, would live in the `data.document` (subject to better naming), but a user would have ability to add 'global/common' data for different nodes here as well.

AlbertHilb:
@oyeanuj Thank you for your comment. :blush:
I want to stress some things.

In my view the `State` `data` is a place where plugins can store informations **not** related to the document structure and content.
To that purpose there are already the namesake fields of `Node`, for data associated to a specified node, and of `Document`, for data related to a whole class of nodes.

So, `State` `data` should be used essentially to keep infos about plugin UIs.
For example, I use a HUD (Head Up Display) to edit certain `isVoid` classes of nodes, and `State` `data` serves me to store if it is open, what node is being edited and so on.

Moreover, the use of a Map to store data, without further structures, gives the developers great freedom with no particular downside.
For example, one can use `Symbol`s as Map keys for the data he wants to keep _private_ and  string for the data he wants to share with other plugins.

oyeanuj:
@AlbertHilb Thanks for the context, appreciate it.  

I believe the two ideas (your original intent and my suggestion) are actually not in conflict. I was also looking to use this object for plugin operations. Except that I ran into cases where plugins were saving information which were at time, node-type specific and not just for the general document. (Btw, there actually isn't a way to store information by node-type).

Also ,assuming a change in `data` would re-render the editor (?), the other advantage of providing that level of fidelity (by node-type) that change in `data` of a node-type would re-render only those node types rather than the whole editor.

In summary, I think my suggestion is to allow for nesting and the default place where a developer would add `data` would be `data.document`. That object can again be just a `Map` and continue the flexibility you intended. The only disadvantage would be that it is a little different from `data` objects that we use in state, or nodes today - but one that I think is worth it.

ianstormtaylor:
Hey @AlbertHilb thanks for this!

I'd like to remove the `serialize` and `deserialize` plugin methods for now, and only allow JSON-serializable data to be stored in the `state.data` property, like we do in the other places it's used. I don't want it to introduce too much complexity to other parts of the system, in terms of serializing.

If you make that change I'm happy to merge!

---

@oyeanuj I think the nested structure you're suggesting is trying to solve a different problem of node re-rendering, but doesn't really belong in the `state.data` object because it's very opinionated. You should be able to solve this by passing emitters around and using `setState` instead.

AlbertHilb:
@ianstormtaylor I removed `serialize` and `deserialize` plugin methods. :blush:

ianstormtaylor:
@AlbertHilb I just realized two things...

I think we'll need to also add a transform function for people to use to update the `state.data` property, since I don't think we should encourage doing `state.data.set('key', value)` directly as everything else is handled by low-level operations. 

So we'd need to add a `set_state` operation (to mirror `set_selection` and `set_node`. And then we'd need to add a `setData` transform (unfortunately calling it `setState` would be to confusing I think?) that will generate that operation.

I also realized that this will be slightly odd in the Operational Transform future. Since `state.data` isn't part of the document, and might reflect local state on the user's machine. I think this is find, since `state.selection` is the same way. So I don't think it's something that needs to block this inclusion. Just a thing to keep in mind, I wouldn't recommend people use `state.data` to store data that will be collaborated on I don't think.

Anywayyyyss... how does that operation+transform piece sound? Are you down to add that? With a test for the transform that matches the other.

AlbertHilb:
HI @ianstormtaylor.

I hadn't add before an `operation` to update `state.data` because `state.data` should not store informations about the document therefore it should remain  unmodified by `undo`, whereas `operation`s are strictly bound to the history system.

Anyway, it's here now, together with a couple of tests. :blush: 

`setDataOperation` accepts the data parameter in the form of an object as usual, but also as an array of `[key, value]` tuples. In this way we can use objects different from strings as keys. For example we can use `Symbol`s as keys to keep some data private.

src/transforms/operations.js
```diff
@@ -286,6 +286,43 @@ Transforms.removeTextOperation = (transform, path, offset, length) => {
 }
 
 /**
+ * Merge `properties` into state `data`.
+ * `properties` can be an object or an array of `[key, value]` pairs.
+ * The array form can be used to store data with keys different from strings.
```
ianstormtaylor: Hey @AlbertHilb can we also change this. Since we're trying to keep the `state.data` JSON serializable, we can't allow keys to be anything other than strings. This will keep it consistent with the other places that `Data` is used now. 

I think the idea of using `Symbol`s for private keys is interesting, but I'd like to keep the scope of this addition smaller for now. If we really decide to allow for non-string keys, we can do it in a separate pass and do it everywhere. Thank you!

test/transforms/index.js
```diff
@@ -177,4 +177,25 @@ describe('transforms', async () => {
...
+
+      it(test, async () => {
+        const testDir = resolve(dir, test)
+        const fn = require(testDir).default
+        const input = await readYaml(resolve(testDir, 'input.yaml'))
+        const expected = await readYaml(resolve(testDir, 'output.yaml'))
+
+        let state = Raw.deserialize(input, { terse: true })
+        state = fn(state)
+        const json = Array.from(state.data.entries())
```
ianstormtaylor: For these tests, let's stick with the `Raw.serialize(state, { terse: true })` like the other tests for now, so that they don't deviate and get confusing.

AlbertHilb: To use `Raw.serialize` we should have an option to serialize `state.data` too, at moment it is discarded. Should I add that?

AlbertHilb:
Hi @ianstormtaylor, I updated the PR following your suggestions. :blush:

ianstormtaylor:
Thanks for bearing with me @AlbertHilb, looks great!

PR #831, opened 2017-05-23
Minor update to "Saving to Database" doc

mike-hearn:
I was going through the documentation, and ran into an error when reproducing the code in the "Saving to Database" section:

*TypeError: Cannot read property 'key' of undefined*

```
Object.deserializeDocument
node_modules/slate/lib/serializers/raw.js:101
   98 | 
   99 | deserializeDocument: function deserializeDocument(object, options) {
  100 |   return _document2.default.create({
> 101 |     key: object.key,
  102 |     data: object.data,
  103 |     nodes: _block2.default.createList(object.nodes.map(function (node) {
  104 |       return Raw.deserializeNode(node, options);

Object.deserializeState
node_modules/slate/lib/serializers/raw.js:234
  231 | 
  232 | if (options.terse) object = Raw.untersifyState(object);
  233 | 
> 234 | var document = Raw.deserializeDocument(object.document, options);
  235 | var selection = void 0;
  236 | 
  237 | if (object.selection != null) {
```

Indicating that the `initialContent` is `terse: true` seems to fix it.

### DIFF
--- a/docs/walkthroughs/saving-to-a-database.md
+++ b/docs/walkthroughs/saving-to-a-database.md
@@ -173,7 +173,7 @@ const initialContent = (
 class App extends React.Component {
 
   state = {
-    state: Raw.deserialize(initialContent)
+    state: Raw.deserialize(initialContent, {terse: true})
   }
 
   onChange = (state) => {


ianstormtaylor:
Ah shoot, thank you @mike-hearn!

Issue #259, opened 2016-08-19
Collaborative editing


davolokh:
@yatskevich and me did a research and tried to design an appropriate architecture for supporting collaborative editing. There are some key points with explanation below.

The main idea is to implement a kind of Redux-like behavior within Slate.
`<Editor store={Store} … />`
The goal of having Store is ability to have clear separation and understandable flow.

UseCase: 
- User press key (types symbol)
- Action is generated
- Action is dispatched to Store
- Store updates State
- setState called
- Editor re-rendered

This way allows us to have ability to manage operations in any suitable way (CRDT, OT, etc.) within Store.

Let’s see how to handle text paste operation in more details:
1. Action `pasteText` dispatched to Store (`data: {position, text}`)
   No need to get current state, call transform() and return new state. Just dispatch an Action.
2. Store has a Reducer for handling such an Action
   If we’re talking about ColloborativeStore - one type, if about SimpleStore - another one. We can even have few implementation of CollaborativeStores (CRDT and OT).

CollaborativeStore:
- Splits `pasteText` operation into a set of atomic `insertCharacter` operations
- Applies operations to State
- Translates this operations set to server
- Listens for updates from server (i.e. smb else is editing this doc and generates the same type of operations set)

SimpleStore:
- Applies operations to State (no need to split to atomic operations)

In our mind such architecture allows Slate to be more flexible and extensible. But this is not a completely analyzed proposal - just an idea. 

And a few pictures for clarification.

Workflow Chart:
![fullsizerender](https://cloud.githubusercontent.com/assets/2715874/17812946/f88bff2e-6631-11e6-8055-1723fe784848.jpg)

Implementation example
![fullsizerender 5](https://cloud.githubusercontent.com/assets/2715874/17812955/01e4e914-6632-11e6-8b70-b2b299e9a017.jpg)

`pastText` in depth:
![fullsizerender 3](https://cloud.githubusercontent.com/assets/2715874/17812958/040e3a4c-6632-11e6-8788-cc9cbcb1f531.jpg)

P.S. Possibly, Store should send Transfroms (not Actions) to Server and get Transforms back.

@ianstormtaylor: Could you please take a look at this and share your opinion?

ianstormtaylor:
Hey @davolokh thank you so much for writing this up!

[...]

PR #836, opened 2017-05-24
Expose transforms

### LINKED ISSUE #683: Incorrect preparing a fragment by range
Hi.

I found issue, when I was using `getFragmentAtRange` method.

Please, imagine, you have 3 blocks. For example, as here:
```html
<p>First line</p>
<ul>
  <li>Second line{selection starts here}</li>
  <li>Third line{selection ends here}</li>
</ul>
```

When slate tries to prepare a fragment by selection, it incorrect splits nodes' tree by selection. End of `Second line` tree is stayed on part of tree, which isn't included in fragment. I think, it happens, because selection moving on the `Third line` text node. But, in this behavior, calculation for end of selection is gone with error:

```bash
VM136 slate.js:37184 Uncaught TypeError: Cannot read property 'key' of undefined
    at Document.insertNode (https://npmcdn.com/slate@0.19.7/dist/slate.js:37184:27)
    at Document.splitNode (https://npmcdn.com/slate@0.19.7/dist/slate.js:37461:21)
    at Document.splitBlockAtRange (https://npmcdn.com/slate@0.19.7/dist/slate.js:37401:17)
    at Document.getFragmentAtRange (https://npmcdn.com/slate@0.19.7/dist/slate.js:36543:17)
    at Document.object.(anonymous function) [as getFragmentAtRange] (https://npmcdn.com/slate@0.19.7/dist/slate.js:48238:30)
    at State.get (https://npmcdn.com/slate@0.19.7/dist/slate.js:40100:83)
    at MyEditor.getSelectedText (eval at transform.run (https://fiddle.jshell.net/js/babel/babel.js:3192:10), <anonymous>:124:39)
    at onClick (eval at transform.run (https://fiddle.jshell.net/js/babel/babel.js:3192:10), <anonymous>:151:28)
    at Object.ReactErrorUtils.invokeGuardedCallback (https://npmcdn.com/react-dom@15.4.2/dist/react-dom.js:9017:16)
    at executeDispatch (https://npmcdn.com/react-dom@15.4.2/dist/react-dom.js:3006:21)
```

Please, try the example [here](https://jsfiddle.net/5nsbem8u/) and just click the button. When you do it, you will see the error above in console.

Has slate some normalizer for selection?
Because, as I understand, user can create an selection with offset, which is more than node's text length.

### LINKED ISSUE #767: Toggling mark on a selection that only partly had it then undoing produces wrong results
1. Open the example at http://slatejs.org
2. Select the “editable **rich** text” passage
3. Press the Bold button. The “rich” word loses its bold formatting
4. Press Ctrl/Cmd-Z

Expected: “rich” should get its formatting back, the other words should remain unchanged.

Actual: the entire selection becomes bold.

### LINKED ISSUE #787: rename `transform.apply()` to something else
It's confusing to have both a `.call()` and `.apply()` method which are unrelated.

Might be good to change it to `commit()` or something similar. Better ideas welcome!

### LINKED ISSUE #813: Bug: Undo/Redo behaviour seemed to be bugged with Inline nodes

Bug

Ive run into this problem on my project at work (quite a severe problem unfortunately :( ) which I can replicate on the demo as well. Only tested my Window machine so far will check if it happens on Mac when i get the chance. (Windows 10, Chrome Version 58.0.3029.96 (64-bit)). 

1. Go to link demo[http://slatejs.org/#/links](http://slatejs.org/#/links)
2. Add a link (with or without an expanded selection)
3. Press CTRL+Z to undo (works)
4. Press CTRL+SHIFT+Z to redo (ERROR)
```
Uncaught Error: Could not find a descendant at path "1,1".
    at Document.assertPath (http://slatejs.org/build.prod.js:10074:13)
    at moveNode (http://slatejs.org/build.prod.js:18313:27)
    at Object.Transforms.applyOperation (http://slatejs.org/build.prod.js:18120:21)
    at Transform.(anonymous function) [as applyOperation] (http://slatejs.org/build.prod.js:15160:32)
    at http://slatejs.org/build.prod.js:21471:15
    at Array.forEach (native)
    at Object.Transforms.redo (http://slatejs.org/build.prod.js:21470:8)
    at Transform.(anonymous function) [as redo] (http://slatejs.org/build.prod.js:15160:32)
    at onKeyDownZ (http://slatejs.org/build.prod.js:16046:61)
    at Object.onKeyDown (http://slatejs.org/build.prod.js:15730:16)
```

Able to undo and redo Inline node insertions/removals

Thanks

### LINKED ISSUE #833: Undoing inline marks doesn't work as expected

Feel free to close this report if it's a duplicate – I searched "undo" but nothing seemed to quite fit. I am not currently developing with Slate.js but discovered a reproducible bug in the example editor.

Tested in Chrome 58 on Mac

1. Visit http://slatejs.org/#/rich-text
2. On the top paragraph of the editor, select "**ch** text, _mu_"
3. Hit delete.
4. Hit `cmd-z`.

*Current behavior:* The re-done text is "_mu_ **ch**"

The editor should restore text to the original state: "**ch** text, _mu_"

### LINKED ISSUE #841: change `updateDescendant` to `updateNode` and handle document case
Right now there's lots of things like...

```
document = isParent ? node : document.updateDescendant(node)
```

...which are tedious. Would be nicer if it compared the key to the top-most node and just returned it directly, so we could do...

```
document = document.updateNode(node)
```

...and the case would be handled.

### LINKED ISSUE #912: Error when trying to delete selection of list item to paragraph

Report a bug.

1. Make a selection from or to the end of a list item (except the last list item) to a paragraph
1. Hit delete/backspace 
1. An error is thrown (`Could not find a node with key X.`) and the selection is not removed

Observed on the [Rich Text Slate Example](http://slatejs.org/#/rich-text):

bo7tysbrpi

Tested with Slate 0.20.3 in latest Chrome (59.0.3071.115) on Windows 7 and 10 (64-bit).

The selected text should be deleted.

### LINKED ISSUE #970: consider changing "join" to "merge"
I don't know why, but it feels more parallel, and seems slightly more obvious of a word to choose to convey the meaning. Also seems to make it clearer why you can only combine adjacent nodes.

### LINKED ISSUE #973: change inverse operations to be computable
Right now when we create an operation we create the inverse at the same time. But this shouldn't technically be necessary if the operation contained all of the contextual information it needed.

- One requirement is that `add_mark` needs to not be called for ranges where the mark is already present, otherwise inverting it will cause issues. (This is already a problem.)

- The other is that removing text or removing nodes should include the removed text/node, so that when trying to invert it we can include that content in the inverted `insert_*` operation. This will make removal operations slightly larger in size, but I think that's alright.

This would allow us to omit the `operation.inverse` and instead just compute `inverse(operation)` when we needed to use it, which for now is only in the local undo scenarios.

### LINKED ISSUE #1003: node.insertText should not default marks
Right now `node.insertText(index, text, marks)` accepts a `marks` argument, but defaults it to the marks currently at the `index` if none is provided. This doesn't work with operations, since the operation will have `marks: null`, but marks will actually end up being applied.

Instead, the marks logic needs to be handled in the transform layer, so the operation correctly gets the marks that are added.

### LINKED ISSUE #1030: add deprecation util to make deprecations more helpful
Right now we're using `./utils/warn` with non-standardized warning messages. It would be nice to add a little layer on top that requires passing in a version number, so that we always log something a bit more helpful like:

```
Deprecation (v0.22.0): The actual deprecation warning here...
```

ianstormtaylor:
Fixes #903 
Fixes #973 
Fixes #970 
Fixes #912
Fixes #1003 
Fixes #841 
Fixes #833 
Fixes #813 
Fixes #767 
Fixes #683 
Fixes #787 
Fixes #1030 

**This is a breaking change to the API.**

This PR adds the last piece of functionality to core that I think is required for #259 to have everything needed for Operational Transform–style collaborative editing.

(Still in progress, but if you have thoughts please comment!)

A summary of the breaking changes so far...

---

**`onChange` now receives a `Change` object (previously named `Transform`) instead of a `State`.** Plugins and users will need to now use:

```js
onChange(change) {
  this.setState({ state: change.state })
}
```

Or more tersely:

```js
onChange({ state }) {
  this.setState({ state })
}
```

Which achieves the same behavior as before. This is needed because it enforces that all changes are represented by a single set of operations. Otherwise right now it's possible to do things like `state.transform()....apply({ save: false }).transform()....apply()` and result in losing the operation information in the history. With OT, we need all transforms that may happen to be exposed and emitted by the editor.

---

**Similarly, handlers now receive `e, data, change` instead of `e, data, state`.** Instead of doing `return state.transform()....apply()` the plugins can now act on the change object directly.

```js
function onKeyDown(e, data, change) {
  if (data.key == 'enter') {
    return change.splitBlock()
  }
}
```

Plugins can still `return change...` if they want to break the stack from continuing on to other plugins. (Any `!= null` value will break out.) But they can also now not return anything, and the stack will apply their changes and continue onwards. This was previously impossible.

---

**The `onChange` and `on[Before]Change` handlers now receive `Change` objects.** Previously they would also receive a `state` object, but now they receive `change` objects like the rest of the plugin API.

---

**The `transform.apply()` method is deprecated.** Previously this is where the saving into the history would happen, but it created an awkward convention that wasn't necessary. Now operations are saved into the history as they are created with change methods, instead of waiting until the end. You can access the new `State` of a change at any time via `change.state`.

---

**The `.apply({ save })` option is now `state.change({ save })` instead.** This is the easiest way to use it, but requires that you know whether to save or not up front. If you want to use it inline after already saving some changes, you can use the `change.setSave(save)` flag instead. This shouldn't be necessary for 99% of use cases though.

---

**The `.undo()` and `.redo()` transforms don't save by default.** Previously you had to specifically tell these transforms not to save into the history, which was awkward. Now they won't save the operations they're undoing/redoing by default.

---

**`onBeforeChange` is no longer called from `componentWillReceiveProps`,** when a new `state` is passed in as props to the `<Editor>` component. This caused lots of state-management issues and was weird in the first place because passing in props would result in changes firing. **It is now the parent component's responsibility to not pass in improperly formatted `State` objects**.

---

**The `splitNodeByKey` change method has changed to be shallow.** Previously, it would deeply split to an offset. But now it is shallow and another `splitDescendantsByKey` change method has been added (with a different signature) for the deep splitting behavior. This is needed because splitting and joining operations have been changed to all be shallow, which is required so that operational transforms can be written against them.

---

**The shape of many operations has changed.** This was needed to make operations completely invertible without any extra context. The operations were never really exposed in a consumable way, so I won't detail all of the changes here, but feel free to look at the source to see the details.

---

**All references to "joining" nodes is now called "merging".** This is to be slightly clearer, since merging can only happen with adjacent nodes already, and to have a nicer parallel with "splitting", as in cells. The operation is now called `merge_node`, and the transforms are now `merge*`.

---

**The `.length` property of nodes has been removed.** This property caused issues with code like in Lodash that checked for "array-likeness" by simply looking for a `.length` property that was a number.


oyeanuj:
@ianstormtaylor One of the things that I had been hoping would be possible when we move to collaborative editing features, was the ability to save 'diff' changes. The most common case would be to send the diff to backend rather than the whole document.

Now to do that, a simple version would be to know which block changed but a more advanced version would be to get a translation of the transform operations. Do you see any of these being possible with the new `onChange` that you suggested above?

ianstormtaylor:
@oyeanuj with this updated API, the `onChange` handler receives a `Transform` object, and those transform objects have an `operations` property that contains the individual operations that have been applied from the change—essentially a diff. So you could do:

```js
onChange = (transform) => {
  const { state, operations } = transform
  this.setState({ state }) // for rendering the update
  this.sendOperations(operations) // to send the operations to the server
}
```

Not sure if that's what you mean. There isn't currently a way to take two `State` objects and diff them to get the operations required to move from a to b, if that's what you're asking.

oyeanuj:
@ianstormtaylor Got it. I did mean the resulting diff between state A and state B. I guess the operations are the steps to get from one state to another, but harder to deal with for saving updates. But then that might be a separate issue to explore than this one.

thomsbg:
I wonder, since this is already a breaking change to the API, if now would be a good time to do any renaming of the `Transform` class, as you mentioned here: https://github.com/ianstormtaylor/slate/issues/259#issuecomment-242504922

If so, may I suggest the word `Delta`? I like that it doesn't overload the word `change`, which is used as a verb as often as it is used as a noun.

eamontaaffe:
This looks great! This is exactly the functionality that I have been hoping for from slate. 

What is left to do before this PR can be merged @ianstormtaylor? Im pretty keen to use this functionality on a project, so if there is anything I can do to help I am happy to.

I have some ideas around a slightly different API which might not require any breaks. Ill try to code up an example to show what I mean.

eamontaaffe:
# Making slate collaboration ready

To get to collaborative editing, we need a separation of state updates and render.

We want to remain agnostic to the choice of OT vs CRDT and the choice of state store like Redux or similar. If possible, it would be great if we did not need to change the public facing API.

To do this we can look at other examples that separate state updates and the underlying logic. A great example is Facebook's [Apollo](http://dev.apollodata.com/) project.

## Apollo

Apollo manages a server side connection with a data API and makes it easy to connect this data to your front end UI.

When using Apollo, there is a separation between the client which is responsible for retrieving the data from the server and provider which funnels this data into your UI components.

At an extremely high level, a basic setup looks like this:

``` javascript
import ApolloClient from 'apollo-client'
import { ApolloProvider } from 'react-apollo'

const client = ApolloClient();

ReactDOM.render(
  <ApolloProvider client={client}>
    <MyRootComponent />
  </ApolloProvider>,
  rootEl,
);
```

This example shows a clear separation from the core logic of fetching data in the `apollo-client` package and the presentation of data in the `react-apollo` package.

Apollo happens to be built on top of Redux, however in this example, there is nothing to suggest that Redux is utilised. If you want to [use an external store](http://dev.apollodata.com/react/redux.html), you also have the ability to do that:

``` javascript
import { createStore, combineReducers, applyMiddleware } from 'redux';
import { ApolloClient, ApolloProvider } from 'react-apollo';
import { todoReducer, userReducer } from './reducers';

const client = new ApolloClient();

const store = createStore(
  combineReducers({
    todos: todoReducer,
    users: userReducer,
    apollo: client.reducer(),
  }),
  {}, // initial state
  applyMiddleware(client.middleware()),
);

ReactDOM.render(
  <ApolloProvider store={store} client={client}>
    <MyRootComponent />
  </ApolloProvider>,
  rootEl
)
```

In this example, the Client is instantiated, then the reducer and middleware are unpacked and slotted into an existing Redux setup. The actions that Apollo sends to the reducer are exposed to any other reducer in the setup and can easily be streamed to a centralised server or similar.

## Slate

We can now take some of these lessons and apply them to Slate.

### Client

The slate client is responsible for all transformations of the current state. You can think of it like a Redux reducer or what is currently the Transforms module of slate. It is instantiated and fed into a resolver which listens for actions sent by the editor itself (which is entirely UI).

### Provider

The slate provider is responsible for collecting user inputs and rendering the current state of the document. The user inputs are collected from DOM events and formatted into a boiled down format like we currently see in the Transformations module. These represent actions taken by the user, ie. `set_selection` and `insert_text` and contain any arguments required, ie. `offset: 4` or `text: \"d \"d`.

### Editor

So that we do not break the current API, we are able to bundle the Client and Provider into a single Editor component that behaves identically to what a user currently expects. It is only when someone wants to expose more advanced functionality that they need to unpack the client and provider.

``` javascript
class Editor extends React.Component {
    constructor(props) {
        super(props);
        this.client = new Client({
            // The client transforms events that it receives and notifies
            // the user of an updated state. The user can then update the state
            // prop as normal.
            onChange: (state) => { this.props.onChange(state); },
        });
    }

    render() {
        // When the provider receives an event it is dispatched to the client.
        return (
            <Provider client={this.client} state={state} />
        );
    }
}
```

### Simple Example

In this example, the easier to use Editor component is imported and implemented exactly the same as you would be able to do previously.

``` javascript
import { Editor, Plain } from 'slate';

class Plain extends React.Component {
    constructor(props) {
        super(props);
        this.state = {
            state: Plain.deserialize("Hello world"),
        }
    }

    render() {
        return (
            <Editor
                state={ state }
                onChange={(state) => { this.setState({ state }); }}
            />
        )
    }
}

ReactDOM.render(
    <Plain />,
    rootEl
)
```

### Advanced Example

In a more advanced use case, the client is attached directly the redux store and given the ability to manage it's own state. If required, the client can be configured with a middleware or something to assist in OT of CRDT or alternatively a Redux middleware can be used to handle all events in a consistent fashion.

``` javascript
import { createStore, combineReducers, applyMiddleware } from 'redux';
import { Editor, Plain } from 'slate';
import { todoReducer, userReducer } from './reducers';

const client = new Client({
    reduxRootSelector: (state) => state.slate
});

const store = createStore(
  combineReducers({
    todos: todoReducer,
    users: userReducer,
    slate: client.reducer(),
  }),
  {}, // initial state
  applyMiddleware(client.middleware())
);

ReactDOM.render(
  <Provider state={store} client={client}>,
  rootEl
)
```

## Conclusion

The slate project is already very close to being able to implement a structure like this. The way the the `Transforms` module is setup lends itself extremely well for OT in particular (much better than draftjs does). If we rearrange some of the components, then we can separate the state update from the render which would allow for a lot more possibilities to be opened up.

ianstormtaylor:
@eamontaaffe thanks for the writeup! I like the goal. Question for you: what behavior is currently impossible given that the editor isn't split into two parts like that?

eamontaaffe:
If we are moving to a collaborative editing design, then we need to have assurances around the events (I use 'events' interchangeably with 'transformations') that are being recorded. We need to be sure that every event that is altering the document is being captured, so that it can be relayed to the other people editing the same document.

#### Current system

The current system looks like this:

current slate

As the user edits the document, the state is updated and the server is notified of the new state of the document accompanied by the operations that were taken to get there. There is a level of trust placed in the editor (which we can think of a as a black box) that it is giving us all of the events we need to describe the transformation. However, there is always a chance that there are some internal events that  might be dropped due to an error or transformations to the state that are forgotten to be logged.

This can mean that a user on another version of the document will be editing a document that has missed some events. We could occasionally remedy this with a full state swap but it defeats the purpose of OT if we do.


#### Split system

One way that we get assurances that all of the events are being exposed is by moving the state transformation outside of the system itself:

split slate

In this system, the only way for the UI to update its state is by passing its events to the external reducer. Therefore, every transformation that is required to get to the current state is visible.

In the collaborative editing scenario, the events would be relayed to the server as they are being passed to the reducer. Any events that fail to make it to the reducer would not be rendered.

Presumably, you would use optimistic UI and update the current state of the editor before the server has responded that it has received the events. In the case that the server times-out or responds with an error status the user can be notified accordingly and the specific events effected can be made known.

#### TL;DR

To answer your question, currently it is impossible to give assurances that all transformations are being captured. This means that we cannot give any other user assurances that they are editing the exact same document collaboratively.

By splitting the system into parts we are able to create an external dependency which exposes all transformations inherently.

ianstormtaylor:
@eamontaaffe can you point out the specific places where events can be lost/failed in the current codebase? I'm curious about what you're thinking there.

emanjavacas:
Hi everybody, I am looking for a way to apply a number of edit operations coming from a client A to another client B. Client B would have a read-only version of the editor, so, in principle none of the OT-related issues would apply to my setup. I am considering using this branch for my purposes, since the new refactoring would make the implementation trivial. But I am unsure how I could apply the incoming operations from A to the current state of the editor of client B.

lxcid:
I agree with @eamontaaffe view on splitting up the state and render. Currently, I had to duplicate the state of Editor in my redux.

The best argument for this is when a change to the state coming from a remote location arrive at the editor. It would have to make changes to the state. Because the state is currently stored inside the component, you have to navigate through react lifecycle to apply the changes, which can complicate the issue.

ianstormtaylor:
@lxcid can you give an example with code snippets if what you're thinking? I'm open to it, I'm just not sure I understand fully what you two are talking about.

robertf224:
Any ETA on merging this?  Just implemented a line-based (as opposed to character-based, going to do soft line-locking like Quip) CRDT backend for a collaborative text editor and this seems perfect for frontend (evaluated DraftJS but didn't seem feasible and then found issue on that repo about collaborative editing that led me here).

ianstormtaylor:
Hey @robertf224, good question. I'm currently working through this. The recent change was to fix splitting/joining nodes so that they could be expressed as simple operations properly. I think there is at least one thing to fix with marks, specifically how mark operations are constructed, which should fix a few of the existing bugs. And then we need to ensure that nothing regressed.

My guess is within the month, but the territory is a bit unchartered for me, so I'm not totally sure, it could even be sooner. But feel free to use this pull request branch in the meantime!

Is your implementation open-source by any chance?

robertf224:
Cool, I'll pull down the branch and give it a shot.  Taking a look at Quill right now but it looks like their Delta format is a bit biased towards OT.  Their data model is just a string of characters so I would probably have to write a lot of custom logic to translate their changes into line-based CRDT operations like adding a line (whereas from what I can gather from Slate's Transform documentation, I should get a lot of that for free with operations like insertBlock/splitBlock).

And unfortunately no, it's a work-related side-project and is pretty intertwined with our products so it's not really something I can (or would make sense to) open-source but I'll write up a post with architecture and general code snippets on this thread for posterity ( / maybe help inform API decisions?)

robertf224:
So client/server both have implementations of a Replicated Growable Array (RGA) similar to that described here (https://hal.inria.fr/inria-00555588/PDF/techreport.pdf) in section 3.5, but the translation from an academic paper to an actual code implementation and the alternate use of lines as atoms instead of characters introduced some twists.  Would recommend skimming section 3 of the paper for most of this to make sense.

Our "Line" object is represented as follows in an interface definition language (from which Java/Typescript objects/interfaces are generated):
```yaml
Line:
  fields:
    created: datetime
    updated: datetime
    removed: boolean
    value: string
```

The `created` field supports the `addRight` operation explained in the paper (we create a line with a timestamp so adds into our mutable list of lines are commutative across clients).  The `updated` field actually serves two purposes.  First, as a timestamp for marking a line with a tombstone (or removing the tombstone!).  This effectively turns our underlying set of lines in the RGA into a Last-Writer-Wins (LWW) Set as opposed to a 2P Set as described in Specification 19 of the paper.  But it also serves to effectively turn our line into a LWW Register, so when clients submit concurrent `update` operations on the content of a particular line, those are also commutative.  If you created composable versions of all of these CRDT data structures and then combined them to create an RGA, these two uses would actually involve two separate timestamps, but I implemented the RGA with all of these assumptions in mind to simplify the code and the data model (which I'll get into in a bit).  We layer this with some "soft" line-locking separate from the RGA implementation (grant a lock to client when they start updating the line, and let them release it when they navigate away from that line or close their connection) so that clients can only accidentally step on each other's toes during initial concurrent edits to the same line.

Our document is serialized/deserialized by a simple object describing the RGA:
```yaml
RgaState:
  fields:
    lines: map<string, Line>
    edges: map<string, string>
```
Where the string keys in each of the above maps are unique identifiers for lines (for which we'll just use UUIDs, which clients will generate in practice).  This serializes into a simple json object we can commit to permanent storage periodically while a document is live, and on close of the last session connected to a document.  We can also send this object over the wire to initialize a new client session with the current state of the document (and then only broadcast/receive operations to/from them from there on).

The RGA implementation looks like this (https://gist.github.com/robertf224/73ce69b944e198dade6944b117d0a3df) in Java, and almost identical in Typescript.  Because operations are commutative, clients can just immediately patch their RGA with their own operations and then submit them to the server, and then when they receive operations (update/addRight from other clients) patch the same object without worrying about ordering.

Went with a line-based operation-based CRDT to avoid the magnitude of garbage that could be created by a character-based CRDT (and we can also just remove any garbage on close of last session before we commit to storage), and because I think line-locking (which meshes nicely with this data structure) actually makes a lot of sense from a user-perspective (though obviously this is subjective).  Obviously we're broadcasting a whole line across the wire for each update operation, but the actual patching of client state is very simple (set your line with this key to this content) and lines shouldn't be *that* long in practice.  And we hopefully get rendering optimizations for free with React (our unique line keys map quite nicely to React element keys).  And most importantly, the implementation is actually rather simple (basically just the data structures above + submitting json objects describing update/addRight operations back-and-forth via WebSockets) and only took me a few days to get right (whereas OT would probably take a lot lot longer).  Obviously there are tradeoffs, but CRDTs seemed like a good fit overall from an implementation perspective, and particularly for our workflows which won't see any sort of large-scale concurrency.

robertf224:
Playing around with the branch, on first pass seems like I'll be able to do what I need to do 😄 (that is, translate Slate's operations to and from my CRDT operations to keep my data structure and the editor in sync).  Having some trouble figuring out exactly the semantics of what's happening though, could you explain in this example why there are 2 split_node operations and how to figure out which node keys the operations occur in (assuming there's some functions I can use to get the node from the path/position)
screen shot 2017-08-09 at 2 23 22 am

robertf224:
Also, setKeyGenerator is totally stellar for a lined-based CRDT implementation.  Noticed undoing deletions restores original key which is huge (makes tombstoned lines usable again).

ianstormtaylor:
@robertf224 good question! The multiple `split_node` operations there are because splits and join were changed to only refer to one node in the tree at a time, and to be recursive. So if you "deeply" split the tree, a `split_node` operation will be created for each level down the tree you do. (Actually it goes from bottom up, as the paths in your screenshot show.) Hope that makes sense.

Thanks for sharing all of that! That's super helpful to hear. Let me know if you run into anything else, there are still some things on this branch that works in progress.

robertf224:
Ran through an example on paper, I understand now haha.  So just to confirm, I should basically be able to keep track of the state that was emitted in a previous cycle, and on a new cycle, for each operation:
- traverse the document tree in the prior state using the .nodes() function on each node to follow the paths attached to the operation
- potentially translate the operation into a CRDT operation
- apply that operation to the prior state and move on to the next operation in the list

Does that sound right?

ianstormtaylor:
@robertf224 not totally sure I understand what those steps mean. For (1) check out the `./models/node.js` file for the methods available, you can use `getNodeAtPath` I think to retrieve a node by path (or something similar, don't have the code in front of me right now).

For applying, there's an apply operation bit of code. On this branch I think it's more nicely abstracted into a util, and I'd be down to expose it publicly too.

If you clarify your steps/plan for me I can comment more!

lxcid:
@ianstormtaylor

Sorry for the late reply. Please ignore my previous comment. I think I understand it wrongly, since `state` is the source of truth, we can already apply the changes separately from the render. We just had to store it outside of the React lifecycle.

Apology for the confusion.

robertf224:
probably just clearer to explain w/ js-ish pseudocode.  here's what I would like to be able to do:
```javascript
onChange = ({ state, operations }) => {
    let oldState = this.state.state;
    for (let operation of operations) {
      switch(operation.type) {
        case "split_node":
            // perform a CRDT update operation on the original node key with the first part of the split text based on operation.position
            // perform a CRDT addRight operation with the key of the original node and the key of the new node and the second part of the split text
        default:
          break;
      }
      oldState = oldState.transform().apply(operation);
    }
}

```

robertf224:
Obviously will have to do some collapsing of split_node / join_node operations, and I think I've figured out roughly the mechanisms through which I can get at the values I need.

My remaining concern is with applying transforms on each iteration of the loop to the old state so that it heads towards the current state.  There doesn't seem to be anything in the operation objects specifying what the key of the new node will be, so I probably can't just iteratively apply the transforms to get each intermediate state, because I won't end up with the correct keys on my nodes if I'm randomly generating them?

robertf224:
Ok I'm now realizing there might be a conflict between my use-case and the API here in general for the same reason.  I'll also need to be able to insert nodes with particular keys, but there's not any way to do that explicitly.  I can imagine some workaround that hacks setKeyGenerator to keep track of keys that were generated during a set of operations so I can play them back consistently during onChange, and to make sure that the node keys I want synced from remote clients into the editor state are used on new nodes.  But that feels wrong.

robertf224:
Oh but I guess we do explicitly set the new state, so I actually can just play back the operations from the old state and ensure that the keygen function uses the keys that I want  by updating the value of some shared reference before applying a split_node operation.  And then finally setting the editor state as my newly created state instead of the one passed into onChange.  Still requires hacking setKeyGenerator but much more mildly.

ianstormtaylor:
@lxcid awesome!

@robertf224 I think I don't know enough about CRDT's requirements to understand why you want to do the things you want to do 😄  so I'm not sure I'll be much help.

robertf224:
Fair enough.  I think I'll be able to do what I need (and will explain more carefully what that actually is in the future) but the one thing I can't figure out in the API is how to apply each of the operation objects in onChange to get to the new state (which I need to do rather than just using the state object passed in).  Is there a convenience function available so that on each iteration of the loop through the operations I can just do something like `state = state.transform(operation)`?

robertf224:
Jk looks like that's what applyOperation is for.  Sorry for all the spam :|

ianstormtaylor:
@robertf224 ah yup, `transform.applyOperation(operation)` should do the trick inside of a loop!

robertf224:
Messing around with the plain-text example on the 79ec9bc commit of this branch (in newest commit things stopped working for some reason) and this almost works:
```javascript
onChange = ({ state, operations }) => {
    let transform = this.state.state.transform();
    for (let operation of operations) {
      transform.applyOperation(operation);
    }
    this.setState({ state: transform.apply() });
}
```
But there's a weird issue with text entry where I'm seeing double characters and the below stack trace:
```
editor.js:230 Uncaught Error: As of slate@0.21.0 the `editor.onChange` method must be passed a `Transform` not a `State`.
    at Object.onChange (editor.js:230)
    at Content.onChange (content.js:332)
    at onInput (content.js:612)
```
And undo only works once before throwing this:
```
on-history.js:124 Uncaught TypeError: Cannot read property 'forEach' of undefined
    at http://localhost:8080/build.dev.js?1502347683081:21190:15
    at Array.forEach (<anonymous>)
    at Object.Transforms.undo (http://localhost:8080/build.dev.js?1502347683081:21189:30)
    at Transform.(anonymous function) [as undo] (http://localhost:8080/build.dev.js?1502347683081:15272:32)
    at onKeyDownZ (http://localhost:8080/build.dev.js?1502347683081:16195:46)
    at Object.onKeyDown (http://localhost:8080/build.dev.js?1502347683081:15884:16)
    at Stack.(anonymous function) [as onKeyDown] (http://localhost:8080/build.dev.js?1502347683081:13813:37)
    at Object._this.(anonymous function) [as onKeyDown] (http://localhost:8080/build.dev.js?1502347683081:7245:23)
    at onKeyDown (http://localhost:8080/build.dev.js?1502347683081:6884:18)
    at Object.ReactErrorUtils.invokeGuardedCallback (http://localhost:8080/build.dev.js?1502347683081:123600:16)
```

Any idea what's up / what I'm doing wrong?

ianstormtaylor:
@robertf224 I've added a new commit that should fix the `editor.onChange` error. 

I'm not sure why the history one is being thrown, but it sounds like the operation being passed in has no `inverse` property. Right now `operation.inverse` is an array that represents the operations required to undo the operation. This will change in the future to be computed on the fly, but right now it's still required. (Hopefully the future being on this branch before it is merged.)

And then what are the bugs you're seeing on the latest HEAD of this branch?

robertf224:
Here's me typing "hello" into the plain text example with the above onChange function:
screen shot 2017-08-10 at 4 43 41 pm

End up with "heelllloo".  Looks like the whole line is getting removed and then all operations are getting totally reapplied every other `onChange`.  Also undo still only works once before throwing that error.

Also getting this on page load in the console:
```
Warning: Failed prop type: The prop `onChange` is marked as required in `Content`, but its value is `undefined`.
    in Content (created by Editor)
    in Editor (created by PlainText)
    in PlainText (created by Route)
    in Route (created by App)
    in Switch (created by App)
    in div (created by App)
    in div (created by App)
    in App
    in Router (created by HashRouter)
    in HashRouter
```

robertf224:
Also don't think it's causing the issue but something strange going on in `apply()` in transforms.js.  `options` never gets used, nor do `_merge` nor `_save` (think that line should probably be `let { merge, save } = options`)

ianstormtaylor:
@robertf224 I just pushed a mid-sized refactor and changes. The operations now no longer include the `.inverse` property, and instead there is an `inverse()` function exposed as `Slate.Operations.inverse` that inverts them in the history logic right before applying them. This required their shape has changed slightly, so beware for any operations sensitive logic you're writing.

I fixed the prop type warning there. But I'm not sure what you mean about typing "hello". Typing in the `rich-text` example locally for me results in what I'd expect.

robertf224:
So I used the plain-text example but replaced the onChange function with the one I posted above (where I apply each operation manually in a loop to get to the final state) and it results in the behavior I described

robertf224:
So I haven't been able to track down the cause of the behavior myself, but here's some more careful description.  Replacing the onChange function in the plain-text example with this one:
```javascript
onChange = ({state, operations}) => {
    console.log("onchange...");
    const transform = this.state.state.transform();
    for (let operation of operations) {
      console.log(operation);
      transform.applyOperation(operation);
    }
    this.setState({ state: transform.apply() });
}
```
I get this behavior: https://www.useloom.com/share/2dcab9077ea24b538a588da3b98ddedf

If I use that same function but pass it as onTransform or onBeforeTransform to the Editor component, I get this behavior: https://www.useloom.com/share/f46c492ccdf7493aacec058e3b407b7e

Which is better but still not quite right.  And in both cases, undo goes backward correctly one step, but then jumps back to the newer state on the second undo (and just cycles between them if you keep undoing).

The state that is created from manually applying the operations to the previous state in onChange/onBeforeTransform/onTransform obviously isn't equivalent to the state that's passed into the function that already has all the operations applied (and apparently also some secret sauce that doesn't break the editor).

robertf224:
so it seems that with the above onChange, applying a set of operations works, but then it ends up triggering an onInput on the content component, which ends up bubbling up from here (https://github.com/ianstormtaylor/slate/blob/expose-transforms/src/components/content.js#L587-L598) to trigger another onChange that we don't want.

the reason I'm going down this path in the first place is that I need a way to figure out what actually happened in a set of operations in terms of keys (not paths, since I've seen those can be ephemeral during a set of operations).  one way of doing that is actually controlling the application of the operations so I could effectively assign keys myself along the way.  if the editor also emitted keys for each operation (in addition to paths) that would probably do the trick, because then I could effectively do a diff by looking at what keys were created/destroyed during a set of operations, and which keys had text updated.  so split_node and merge_node would emit the keys of both nodes, insert_text would emit the key in which text was inserted, etc.

ianstormtaylor:
Hey @robertf224, I don't totally follow you. But the duplicate character typing you're seeing is because the `isNative` flag isn't being preserved when you're essentially "copying" the transforms. Take a look at the new `isNative` transform for how to set it. (And you can search issues for other people that ran into this.)

isubastiCadmus:
@ianstormtaylor copy is broken on rich text as getFragmentAtRange calls splitNode with the old path and offset instead of index and position

src/models/node.js
```diff
@@ -1540,94 +1534,42 @@ const Node = {
...
 
-  splitNode(path, offset) {
-    let base = this
-    const node = base.assertPath(path)
-    let parent = base.getParent(node.key)
-    const isParent = base == parent
-    const index = parent.nodes.indexOf(node)
-
-    let child = node
+  splitNode(index, position) {
```
ashutoshrishi: A few places still call `splitNode` with a path (array) instead of an index. This throws an exception when you tried to do `Cmd-C` (Copy command) on the editor (observable in the rich text example on this branch).

Like here:
https://github.com/ianstormtaylor/slate/pull/836/files#diff-82007cf8e57fb38e037a7553af96e41fR1533

ashutoshrishi:
I was using the `expose-transforms` branch to store the transforms somewhere. There are a few bugs in it that I found:

1.  Pressing `Cmd+C/Meta+C` (Copy command) an exception is thrown:

```
Uncaught TypeError: Cannot read property 'kind' of undefined
    at Document.splitNode (http://localhost:3000/static/js/bundle.js:205038:14)
    at Document.splitBlockAtRange (http://localhost:3000/static/js/bundle.js:205018:17)
    at Document.getFragmentAtRange...
```

1.  Populate a paragraph, add a list block below it and start pressing Undo, an exception is thrown:

```
Uncaught Error: Tried to merge two nodes of different kinds: "text" and "block".
    at Block.mergeNode (http://localhost:3000/static/js/bundle.js:204867:13)
    at merge_node (http://localhost:3000/static/js/bundle.js:208796:21)
    at applyOperation...
```

Both of these are observable in the rich text example in the `expose-transforms` branch. I have tried digging deeper into the issues, and probably found something useful to fixing the bug

ianstormtaylor:
Thanks @ashutoshrishi those two should be fixed now!

isubastiCadmus:
@ianstormtaylor the issues that @ashutoshrishi pointed out are indeed fixed but cut is broken
`Uncaught TypeError: fn is not a function
    at Editor.transform (build.prod.js:7870)
    at build.prod.js:16914`

src/plugins/core.js
```diff
@@ -170,74 +163,64 @@ function Plugin(options = {}) {
...
     // delete the content in the current selection.
     window.requestAnimationFrame(() => {
-      const next = editor
-        .getState()
-        .transform()
-        .delete()
-        .apply()
-
-      editor.onChange(next)
+      const t = editor.transform().delete()
```
isubastiCadmus: editor.transform need to have a function callback and editor.applyTransform has been removed on the newest changes

ianstormtaylor:
Thanks @isubastiCadmus, `onCut` should be fixed now!

ianstormtaylor:
Okay, this is good to go! 

A significant change to how things work, but I think it makes things actually quite a bit cleaner, and closes a bunch of different issues.

I'm gone next week, but I'll be releasing this the Monday the week after!

PR #845, opened 2017-05-26
Documenting node's getFurther* functions

oyeanuj:
I remembered coming across these functions and then didn't find them in the docs. So, here they are now!


ianstormtaylor:
Thanks @oyeanuj this is awesome!

PR #847, opened 2017-05-26
Wrap scrolling logic in setTimeout

pa-nam:
For the problem come here: https://github.com/ianstormtaylor/slate/issues/279#issuecomment-304215238

Detail fix:
The native selection will be updated after componentDidMount or componentDidUpdate. We use setTimeout to queue scrolling to the last, when the native selection has been updated to the correct value.


ianstormtaylor:
@pa-nam this is awesome, thank you for digging in and finding out what the cause was!

Issue #750, opened 2017-04-24
Characters are typed twice with both onBeforeChange and onChange


jatins:
Here's a fiddle illustrating the behaviour: https://jsfiddle.net/4890mgej/4/

Code:
```js
class MyEditor extends React.Component {

  constructor(props) {
    super(props)
    this.state = {
      state: Plain.deserialize('')
    }
  }

  onChange(state) {
    this.setState({ state })
  }
  
  onBeforeChange(state) {
  	return state.transform().setBlock('paragraph').apply()
  }

  render() {
    return (
      <Editor
        placeholder="Enter some text..."
        onChange={state => this.onChange(state)}
        onBeforeChange={state => this.onBeforeChange(state)}
        state={this.state.state}
      />
    )
  }

}
```

ianstormtaylor:
Hey @jatins, this is actually a UX issue, I'm curious to get your thoughts on. (And anyone else!)

Basically what's going on here is that for certain operations, (like inserting single text characters), Slate uses the `isNative` flag to avoid having to re-render, and instead just accept what is in the DOM, since these are cases where we know `contenteditable` will get the insertion right. This greatly improves performance since it avoids more re-renders.

But in this case, when using `onBeforeChange` and performing another transform, that flag is removed. This results in the original DOM event not having been `preventDefault()`'d, and also that the rendering logic adds the character too.

What are you looking to use `onBeforeChange` for here?

I'm considering removing it entirely since it leads to weird issues.

jatins:
My use case is essentially a series of these steps:

1. Get _text_ content from editor
2. Based on text content set block type, or do some formatting. However, at this point I want the text _after_ edit. (So, I can't use something like `onKeyDown` which essentially gives me the state _before_ edit is applied)
3. Render the new state.

But also, I would want the typed character to be displayed as soon as user types (which I think won't happen if I `preventDefault`).

jatins:
Also, that double character typed behaviour doesn't have much to do with `onBeforeChange`. If I do this inside my `onChange` I get the same behaviour.

```js
 onChange(state) {
    this.setState({ state: state.transform().setBlock('paragraph').apply() })
  }
```

https://jsfiddle.net/4890mgej/5/

soulwire:
[...]

PR #850, opened 2017-05-26
Avoid duplicate content with `onChange` handlers

AlbertHilb:
This PR intends to fix issue #750.

It changes a bit the way Slate elaborates a user input.

At the end of the `onBeforeInput` handlers stage, if the `onBeforeInput` event has not been canceled, the editor does not call any more `onBeforeChange` and `onChange` handlers, but stores temporarily the state.

After the changes are applied to the DOM an `input` event is fired, then the editor takes the stored state, calls `onBeforeChange` and `onChange` handlers and finally sets the outcoming editor state.

In this way, if a `isNative = true` state returned by the `onBeforeInput` stage is mutated by `onBeforeChange` or `onChange` handlers, it is rendered after DOM has already been mutated and the changes don't cumulate.

As always any change you think needed is welcome. :blush:

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -551,7 +551,11 @@ class Content extends React.Component {
     debug('onInput', { event })
 
     const window = getWindow(event.target)
-    const { state, editor } = this.props
+    const { editor } = this.props
+    const { pendingOnBeforeInputState } = editor.tmp
+    const state = pendingOnBeforeInputState ?
+      pendingOnBeforeInputState :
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Hey @AlbertHilb I think this is no longer needed now that `isNative` was removed, so I'm going to close it out. Let me know if that sounds wrong!

AlbertHilb:
@ianstormtaylor. It's ok!

PR #852, opened 2017-05-28
Check to not break if node does not have getRanges

oyeanuj:
I ran into a weird bug with slate-trailing-block where a `Block` was making it way here via `onSelect` on down arrow key. Anyway, that was leading to a Javascript error in the console as only `Text` has getRanges defined. This check gracefully returns if the node does not have a `getRanges` method.

### DIFF
--- a/src/utils/get-point.js
+++ b/src/utils/get-point.js
@@ -24,7 +24,7 @@ function getPoint(element, offset, state, editor) {
   // then afterwards for the correct `element`. (2017/03/03)
   const { key } = offsetKey
   const node = document.getDescendant(key)
-  if (!node) return null
+  if (!node || !node.getRanges) return null
 
   const decorators = document.getDescendantDecorators(key, schema)
   const ranges = node.getRanges(decorators)


ianstormtaylor:
Hey @oyeanuj what was causing the `Block` to get to that part of the code? I'd like to figure out if there's a root cause we can fix instead.

oyeanuj:
@ianstormtaylor To be honest, I wasn't easily able to reproduce it in a JSFiddle, so used a different workaround. But here is what was happening since it might feel familiar to you - 

---

On click of a button, I created a new custom node, `Step`, with a `StepHeading`, and `StepBody` containers. The `StepBody` is just a shell, which can contain paragraphs or other nodes. So, I initialize it with a paragraph node (tried by creating text node inside as well as without).

Here is what it looks like before, and after:
```yaml
Editor
 - Paragraph
```

```yaml
Editor
 - Paragraph
 - Step
   - StepHeading (H2)
   - StepBody
     - Paragraph
```
Now before the changes are complete, `slate-trailing-block`, comes in and adds another paragraph to result into this:

```yaml
Editor
 - Paragraph
 - Step
   - StepHeading (H2)
   - StepBody
     - Paragraph
 - Paragraph
```

Now, going from the paragraph in`StepBody` to the last Paragraph node in the editor by pressing the down arrow key was throwing the error. The node in question, in the `getPoint` function (called via `onSelect`) was the parent `Step`, hence the error.

---

I do intend to come back to it and look it up more, to understand why the issue was happening. But, in the meantime, I figured it should atleast return from the function gracefully. Alternatively, it could try and find the first text within the node, or throw some error?

And any thoughts on what might be going on, in the case I described above?

ianstormtaylor:
Hmm interesting, I don't think I've had that case locally with triple-nested, like that so that could be why I haven't seen it. I think we should figure out what the cause is, and change the `getPoint` (or deeper) logic to handle that case gracefully, instead of swallowing the error basically. Because it sounds like a case that should be fixed.

oyeanuj:
Agree. Do you think picking the first text node incase a `Block`, or `Inline` is passed is a graceful fallback? Or rather throw a warning? Or do both, as it lets the developer know and picks the first text node?

ianstormtaylor:
I think the logic should go in `OffsetKey.findKey()`, and it depends on what the actual target DOM element of the initial selection event is I think. But selecting the appropriate (either start, end, or specific element) inside the block sounds right.

ianstormtaylor:
Hey @oyeanuj I'm going to close this one for sanity, but if you figure out what's causing the error in the first place let me know. Thank you!

PR #855, opened 2017-05-30
Fix missing isAltKey

koole:
The first MarkHotkey example threw an error about isAltKey being undefined.

In the second MarkHotkey example copying using `cmd/ctrl+c` did not work anymore, because there was nothing checking if the alt key was pressed.


ianstormtaylor:
Ah great catch, thanks @koole!

PR #856, opened 2017-05-30
Ignore comments in Html deserializer

kellyjosephprice:
If you paste in HTML with comments, the comments get pasted in as text. I noticed this when using Google Sheets with cell borders.


ianstormtaylor:
Ah awesome, thanks @kellyjosephprice, looks perfect!

PR #857, opened 2017-05-31
When the parameter passed to `isInEditor` is a text node use its parent instead

AlbertHilb:
To test if `target` belongs to the editor, `isInEditor` `Content` method uses the code:

```
target.isContentEditable &&
(target == element || findClosestNode(target, '[data-slate-editor]') == element)
```

but `isContentEditable` is a property of `HTMLElement` only.
Therefore when `target` is a text node `isInEditor` returns false even when it should not.
To fix this, use `target` parent in the check when `target` is a text node.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -247,6 +247,8 @@ class Content extends React.Component {
 
   isInEditor = (target) => {
     const { element } = this
+    // If `target` is a text node use the parent in the check.
+    if (target.nodeType === 3) target = target.parentElement
     return (
       (target.isContentEditable) &&
       (target == element || findClosestNode(target, '[data-slate-editor]') == element)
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @AlbertHilb good catch. Can you update the change to instead use another variable and not override target for the further checks, since I think it might not be clear in the future. (And also change the comment mention the `isContentEditable` need.) Something like...

```js
// COMPAT: Text nodes don't...
const el = target.nodeType == 3 ? target.parentNode : target

return (
  (el.isContentEditable) &&
  (target == element || ...
```

Just want to make sure we don't accidentally override it later, since that check function has received lots of little tweaks as pieces don't work. Thank you for this!

AlbertHilb:
Hi @ianstormtaylor, I applied the changes you suggested. :blush:

ianstormtaylor:
Thanks @AlbertHilb!

Issue #127, opened 2016-07-19
IME forced rendering could be optimized


ianstormtaylor:
In `<Content>` right now any IME input triggers a re-render with the `this.forces` property, but really (I believe) it could trigger only when the leaf was empty and now isn't (or something similarly less frequent)!

PR #858, opened 2017-05-31
Fix editor should update on composition end.

XuHaoJun:
In my usecase, editor use fullscreen style on focused and disable fullscreen style on mobile soft-keyboard close(detect by height change), but oncompositionend and onclosesoftkeyboard trigger same time, editor not update when forces key is changed, may be have more better way to fix it.


XuHaoJun:
#127, put forces on tmp prop for less frequent.
May be find another way to trigger update by myself.

PR #867, opened 2017-06-05
Fix: findDOMNode was called on an unmounted component

pa-nam:
This bug occurs after my previous fix at https://github.com/ianstormtaylor/slate/pull/847
It happens because some nodes are unmount right after they did update.

It's supposed to wrap only the get native scroll, not the whole findDOMNode included. That's my fault when I didn't think about this problem.

### DIFF
--- a/src/components/node.js
+++ b/src/components/node.js
@@ -162,6 +162,17 @@ class Node extends React.Component {
   }
 
   /**
+   * There is a corner case, that some nodes are unmounted right after they update
+   * Then, when the timer execute, it will throw the error
+   * `findDOMNode was called on an unmounted component`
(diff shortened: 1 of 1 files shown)


pa-nam:
I was thinking that we can clear the timeout when component unmount, may be it's a better solution, like this:

```js
constructor() {
  ...
  this.scrollTimer = 0
}

componentWillUnmount() {
  clearTimeout(this.scrollTimer)
}

updateScroll = () => {
  ...
  this.scrollTimer = setTimeout(...)
}
```

ianstormtaylor:
Good call @pa-nam thank you!

PR #872, opened 2017-06-08
Fix selection after transform.wrapInline

SamyPesse:
When using `transform.wrapInline` with a selection containing the entire content of a block, the selection is reset to the beginning of the block. Instead it should preserve around the whole block.

It impacts user for example if wrapping all the text in a paragraph in a link.

I'm adding a failing test, but I'm not sure how to fix correctly (I'll investigate).

- [x] Add failing test
- [ ] Fix it


ianstormtaylor:
Thanks @SamyPesse and @Soreine this is great!

Issue #875, opened 2017-06-09
Chinese input method has issue in firstLine


liSong5713:
platform: windows 10
version ：chrome 58
when i swith to Chinese input method in first line,Slate will show 'given range isn't in document'  and change  to show text slowly

ianstormtaylor:
Closing in favor of #854

liSong5713:
thanks

Issue #844, opened 2017-05-26
[bug] Error when using Chinese Input Method


terryoy:
Hi, I'm using Chrome on OSX. 

I'm trying to use slatejs in a Chinese environment. I find out the Pinyin method will cause exceptions sometimes.

I made a test on the Rich Text demo.
http://slatejs.org/#/rich-text

After typing a few letters, usually there will be a Chinese Character/Word assist that you can choose to put in the line. If I press space, the charater/word should be chosen and show in the editor. 

But when I press space, the editor throws exceptions instead. I'm trying to figure out why and log an issue here first.

```
build.prod.js:22435 Uncaught TypeError: Cannot read property 'firstChild' of null
    at findDeepestNode (build.prod.js:22435)
    at Content.updateSelection (build.prod.js:6439)
    at Content.componentDidUpdate (build.prod.js:6371)
    at CallbackQueue.notifyAll (build.prod.js:116950)
    at ReactReconcileTransaction.close (build.prod.js:127356)
    at ReactReconcileTransaction.closeAll (build.prod.js:130323)
    at ReactReconcileTransaction.perform (build.prod.js:130270)
    at ReactUpdatesFlushTransaction.perform (build.prod.js:130257)
    at ReactUpdatesFlushTransaction.perform (build.prod.js:128377)
    at Object.flushBatchedUpdates (build.prod.js:128460)
findDeepestNode	@	build.prod.js:22435
updateSelection	@	build.prod.js:6439
componentDidUpdate	@	build.prod.js:6371
notifyAll	@	build.prod.js:116950
close	@	build.prod.js:127356
closeAll	@	build.prod.js:130323
perform	@	build.prod.js:130270
perform	@	build.prod.js:130257
perform	@	build.prod.js:128377
flushBatchedUpdates	@	build.prod.js:128460
closeAll	@	build.prod.js:130323
perform	@	build.prod.js:130270
batchedUpdates	@	build.prod.js:124764
batchedUpdates	@	build.prod.js:128385
dispatchEvent	@	build.prod.js:125162
```

Screen Recording:
http://recordit.co/H6YToYThiX

ianstormtaylor:
Closing in favor of https://github.com/ianstormtaylor/slate/issues/854 which I believe is the same issue?

liuningnb:
[1209](https://github.com/ianstormtaylor/slate/issues/1209)

PR #880, opened 2017-06-12
Fix IME input not work

### LINKED ISSUE #871: Chinese input methods breaks. " Cannot read property 'firstChild' of null". (with official demo)

bug.

#### How to produce the bug?

1. Using the official demo here: https://jsfiddle.net/2zokvrvt/7/
2. Download a Chinese input method, like [https://www.google.com/intl/zh-CN/ime/pinyin/](https://www.google.com/intl/zh-CN/ime/pinyin/) or
or [http://pinyin.sogou.com/mac/](http://pinyin.sogou.com/mac/)
3. Enter the demo, press **ENTER** to start a new line(must be an empty new line), 
4. Switch to Chinese input method(usually **ctrl + space** to switch input method), 
5. Type something and the help tool will show, then press **space**(important) to choose the default first option. A bug will show as "Cannot read property 'firstChild' of null", and only a space will show on the line you type in. 
6. (Then press **ENTER** or **backspace**, the actual Chinese characters will show this time.)

Chinese character properly shown when you press space and no error pops. like [https://www.google.com/intl/zh-CN/inputtools/try/](https://www.google.com/intl/zh-CN/inputtools/try/)

btw, I tried draft.js, and they don't have the same issue. I switch to draft.js only temporarily as I love the api of slate.js so much more. Hope this one gets solved soon :D

### LINKED ISSUE #879: [IME] Should not be re-rendered when use input method
The case:

slate-re-render

When the editor has video, iframe, or some complex component like MathJax, it will very clear to see the html DOM re-render.

I found [`onCompositionEnd`](https://github.com/ianstormtaylor/slate/blob/master/src/components/content.js#L357) event  has some problems with use setTimeout to listen another composition event, and increment the `forces` key, and I will try to fixed it.

pizn:
At first I just wanted to closes #879 , but I found is the wrong way. 

I found when use IME input, while the state is updated but react did not render the current DOM. And it really need to re-render when onCompositionEnd to update the forceKey. But not enough, I have to use forceUpdate() to update the DOM.

It can fixed #871 #875 #844

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -83,7 +83,6 @@ class Content extends React.Component {
     super(props)
     this.tmp = {}
     this.tmp.compositions = 0
-    this.tmp.forces = 0
   }
 
   /**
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @pizn I don't know enough about IME to get it to work reliably. I'm open to a pull request to completely fix IME, but I'd rather have it done all in one go, and I'm not sure if this does it? If you're up for making that kind of PR, that would be awesome, I'd need you to add a lot more information so that I can be more sure of merging it though.

Issue #854, opened 2017-05-30
IME input does not work, removes extra characters


pizn:
As follows:

![input-bug](https://cloud.githubusercontent.com/assets/986480/26564088/1c7199aa-450f-11e7-88b9-65294af04a63.gif)

I use Chrome on OSX.

yuanyong:
I've got the same issue too.

yuanyong:
The issue is just happing when the input source is simple Chinese.

pizn:
@ianstormtaylor I try to solve the IME question I have got in my RP #880 .

It's a difficult problem for IME, I don't want it always re-render when user input, but it have to re-render when react state is update and the DOM is old. Be exact, the DOM is empty(null) and should forceUpdate() to re-render.

Composition Event and the unreconcilable state is the problem，maybe I only solved the surface of the problem.

blackcater:
It doesn't work. 

![2017-07-06 14 17 38](https://user-images.githubusercontent.com/9465551/27897954-ea310462-6255-11e7-9be4-0b3cf54d6fea.gif)

PR #881, opened 2017-06-12
Check platform before adjusting selection

### LINKED ISSUE #810: [IME] Wrong character deletion after inputting Japanese character

bug

I found that the character next to the cursor is wrongly deleted when I input a Japanese character.
(maybe it also occurs in some other multibyte languages...?)

I looked into the code and figured out [this part](https://github.com/ianstormtaylor/slate/blob/master/src/plugins/core.js#L104-L130) of the core plugin causes it.
But the fundamental problem is `anchorOffset` and `focusOffset` returned by native `window.getSelection()` is different in each browsers.

I tested the behavior by inputting Japanese character `あ` into the sentence below. (`|` is cursor)

```
A line of text in a para|graph.
```

The result is so strange:

| Browser | `anchorOffset` | `focusOffset` |
|----|----|----|
| **Expected** | **24** | **24** |
| Safari (iOS/Mac) | 24 | 24 |
| Chrome 58 (Mac/Win) | 24 | 25 |
| Firefox 53 (Mac/Win) | 25 | 25 |
| IE 11 (Win) | 25 | 25 |

I think it's hard to fix, but it's a fact that the fix only for iOS affects Japanese input unexpectedly.
So my suggestion is checking platform before updating the selection.

I can submit PR if my suggestion is accepted.

questbeat:
This PR closes #810.

There is a selection adjustment in `onBeforeInput()` in core.js, to fix a problem with iOS text suggestion (#655).
But it's always applied regardless of the current platform.
This causes #810.

I've confirmed this does not break iOS text suggestion.

### DIFF
--- a/src/constants/environment.js
+++ b/src/constants/environment.js
@@ -75,5 +75,6 @@ export const IS_CHROME = BROWSER === 'chrome'
 export const IS_FIREFOX = BROWSER === 'firefox'
 export const IS_SAFARI = BROWSER === 'safari'
 
+export const IS_IOS = OS === 'ios'
 export const IS_MAC = OS === 'macos'
 export const IS_WINDOWS = OS === 'windows'
--- a/src/plugins/core.js
+++ b/src/plugins/core.js
@@ -8,7 +8,7 @@ import Placeholder from '../components/placeholder'
 import React from 'react'
 import getWindow from 'get-window'
 import findDOMNode from '../utils/find-dom-node'
-import { IS_CHROME, IS_MAC, IS_SAFARI } from '../constants/environment'
+import { IS_CHROME, IS_IOS, IS_MAC, IS_SAFARI } from '../constants/environment'
 
 /**
  * Debug.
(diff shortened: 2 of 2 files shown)


ianstormtaylor:
Hey @questbeat @yujianok @pizn it looks like there are two different solutions to what might be the same issue? So I wanted to see if you could give me some info on which is the "right" one, or whether they're not the same.

- #881 (Fixes #810) (This PR)
- #885 (Fixes #854)

pizn:
Hey @ianstormtaylor , This RP is the right way to fix #810 #854.

I add PR info for #880, the setTimeout and forceKey make editor DOM reRender, it will blur and lost ArchorKey when writing use IME. So I try to fixed it.

And IME has a lot of same issues like #871 #875, it's another question when use IME at the empty line.

yujianok:
hi @ianstormtaylor  @questbeat @pizn , i have test the both solutions, they all works fine. But in my opinion, the problem is not directly concerned by platforms, it is directly caused by the input suggestion and the input method. Beware,  input suggestion and the input method may be not the same thing. i have tested that the input suggestion is not trigger composition event. If we fix this problem just on ios, one day if windows or andriod also has a input suggestion feature, then this issue will occur again. So we had better to find a solution to differentiate the input suggestion and the input method.

By the way, when i test, i found another issue: On ios, after you select one suggestion word then press backspace and blankspace key, the last letter will be replace by blank space.

questbeat:
Hi @ianstormtaylor @pizn @yujianok, thanks for discussion and testing.

I pretty much agree with @yujianok.
I think my PR is not an essential fix for this problem.

But I don't think that

>  If we fix this problem just on ios, one day if windows or andriod also has a input suggestion feature, then this issue will occur again.

Because a input suggestion feature on different platforms could behave differently.
And leaving workaround code that is only for iOS to all paltforms may cause another problem.
So we had better follow YAGNI.

In this case I think it's reasonable to check platform before adjustment.

yphc:
Hi,  I found a issue that  it will re-render when move the cursor back or enter new line.

ianstormtaylor:
Hey all, I'm not fluent enough in IME to be able to make good decisions on this on my own. Can someone provide a detailed summary of each of the issues that the PR's are trying to address? And also can someone evaluate whether they all three need to be merged or whether just some of them do if they overlap? Thank you.

danburzo:
@ianstormtaylor I've added some of my own research here: https://github.com/ianstormtaylor/slate/issues/1012

danburzo:
@questbeat according to environment.js:

```js

const OS_RULES = [
  ['macos', /mac os x/i],
  ['ios', /os ([\.\_\d]+) like mac os/i],
  ['android', /android/i],
  ['firefoxos', /mozilla\/[a-z\.\_\d]+ \((?:mobile)|(?:tablet)/i],
  ['windows', /windows\s*(?:nt)?\s*([\.\_\d]+)/i],
]
```
would mean that IS_IOS will always return false (since `macos` will match the userAgent first).

doodlewind:
@ianstormtaylor I found this PR can also be closed by #1236 , whose approach is essentially the same. Sorry for omitting this, should have found this PR if it has `ime` label 😢

ianstormtaylor:
Hey @doodlewind, no worries because things had changed a bit since then. Thanks!

@questbeat, thank you for this! This should be working in the latest release!

PR #884, opened 2017-06-14
Fix dragging text in the editor

SamyPesse:
Currently drag and drop of text in the same editor doesn't work (You can test it at http://slatejs.org/#/rich-text).

The `data.type` is set to `node` only when dragging a void node. I've also moved to the core plugin, the `onDrop` handling when `data.type == 'node'`.

I don't understand why the type `node` has been introduced instead of just using the type `fragment` when dragging a void node; but anyway with this fix, dragging text works fine, and dragging void nodes works fine as well.

### DIFF
--- a/examples/images/index.js
+++ b/examples/images/index.js
@@ -178,27 +178,7 @@ class Images extends React.Component {
   onDrop = (e, data, state, editor) => {
     switch (data.type) {
       case 'files': return this.onDropOrPasteFiles(e, data, state, editor)
-      case 'node': return this.onDropNode(e, data, state)
     }
-  }
-
-  /**
-   * On drop node, insert the node wherever it is dropped.
-   *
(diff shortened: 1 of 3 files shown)


dmitrizzle:
Oh I thought draggable text is something that we need to build ourselves...

These might be helpful: #302 and #628

ianstormtaylor:
@SamyPesse ah interesting, good call. I think we actually want the `type` to be set to `fragment` instead, and use the `insertFragment` case there. Otherwise it's not going to preserve formatting or other things when dragged.

SamyPesse:
@ianstormtaylor I've updated the PR, and it now fixes "all" the issues with dragging text or void nodes.

ianstormtaylor:
Thanks @SamyPesse, looks great!

PR #885, opened 2017-06-15
fix IME input will replace characters after cursor

yujianok:
When using Chinese input method, the characters after cursor will be replaced by the committed words.

Reason:
The IME will delete preview characters in the editor when user commits input words, so need not to update selection during composing, or the words after will be replaced.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -266,7 +266,7 @@ class Content extends React.Component {
     if (this.props.readOnly) return
     if (!this.isInEditor(event.target)) return
 
-    const data = {}
+    const data = { isComposing: this.tmp.isComposing }
 
     debug('onBeforeInput', { event, data })
     this.props.onBeforeInput(event, data)
--- a/src/plugins/core.js
+++ b/src/plugins/core.js
@@ -111,7 +111,11 @@ function Plugin(options = {}) {
     const { anchorNode, anchorOffset, focusNode, focusOffset } = native
     const anchorPoint = getPoint(anchorNode, anchorOffset, state, editor)
     const focusPoint = getPoint(focusNode, focusOffset, state, editor)
-    if (anchorPoint && focusPoint) {
+
+    // The IME will delete preview characters in the editor when user commits input words,
+    // so need not to update selection during composing, or the words after will be replaced.
+    const isComposing = data.isComposing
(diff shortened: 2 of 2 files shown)


yujianok:
hey @ianstormtaylor, first of all, thank you for opening the great editor framework. It performs wonderfully in our application. But this issue is troubling our user a little. The Chinese IME is important to our user when using slate editor. Maybe there is a better solution, but just now, this PR seems a right one at least. And we are waiting for this issue to be closed hardly.  Can you please merge this PR first. If there is any new issues cascade by the PR, we can adjust it with a new sight.

dzmitry-stramavus:
I see the same issue for French IME.

doodlewind:
This PR indeed solves the bug that IME cursor 'eats' text behind. However it also introduces new issue: When user selects across multi blocks or marks, replacing this selection with IME throws error like below:

```
DOMLazyTree.js:69 Uncaught TypeError: Cannot read property 'replaceChild' of null
    at Function.replaceChildWithTree (DOMLazyTree.js:69)
    at Object.dangerouslyReplaceNodeWithMarkup (Danger.js:41)
    at Object.dangerouslyReplaceNodeWithMarkup [as replaceNodeWithMarkup] (DOMChildrenOperations.js:124)
    at ReactCompositeComponentWrapper._replaceNodeWithMarkup (ReactCompositeComponent.js:784)
    at ReactCompositeComponentWrapper._updateRenderedComponent (ReactCompositeComponent.js:774)
    at ReactCompositeComponentWrapper._performComponentUpdate (ReactCompositeComponent.js:724)
    at ReactCompositeComponentWrapper.updateComponent (ReactCompositeComponent.js:645)
    at ReactCompositeComponentWrapper.receiveComponent (ReactCompositeComponent.js:547)
    at Object.receiveComponent (ReactReconciler.js:125)
    at Object.updateChildren (ReactChildReconciler.js:109)
```
Original slate without this PR won't ran into this issue. Any ideas troubleshooting @yujianok ?

yujianok:
I have a long time not trace this issue, can not remember much clearly about the issue`s detail. About the new introduced issue, my guess is that maybe the input status “isComposing”  is not very accurate.  The anchor and the focus is not adjusted correctly when you select some blocks. Then react can not find the child need to be replaced.

PR #886, opened 2017-06-15
Update core plugin to use new transform 

cosmophile:
onCut method updated to use the new transform object on the expose-transforms branch. Tiny change but debugging it helped me understand slate better, and if it's correct then hopefully it saves you a couple seconds.

Thanks for a fantastic lib!

### DIFF
--- a/src/plugins/core.js
+++ b/src/plugins/core.js
@@ -216,8 +216,8 @@ function Plugin(options = {}) {
     // Once the fake cut content has successfully been added to the clipboard,
     // delete the content in the current selection.
     window.requestAnimationFrame(() => {
-      const t = editor.transform().delete()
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @cosmophile good catch, that existing code is broken. I think with newest code on that branch the change should actually end up as:

```js
editor.transform(t => t.delete())
```

Thanks for writing this up! And glad it was helpful for understanding 😄

isubastiCadmus:
this has been fixed on https://github.com/ianstormtaylor/slate/pull/836/commits/e40820b5e6e66e9c55a4ea02d73e4b0496f0e19f

ianstormtaylor:
Ah nice, thanks @isubastiCadmus!

PR #890, opened 2017-06-18
Add example of using plugin.render to /examples/plugins

conorcussell:
There was a question in Slack about how this property worked, so thought it would be worth adding a simple example. Let me know if you think any thing should be changed?


ianstormtaylor:
Thanks @conorcussell, cool idea for it!

PR #892, opened 2017-06-19
Updated findDOMNode function

Dtphelan1:
Newer implementations of findDOMNode take the node to find as the argument, not the key of the node


ianstormtaylor:
Thanks @Dtphelan1!

PR #894, opened 2017-06-21
Don't mutate `path` array, use instead a copy.

AlbertHilb:
This fixes issue reported in a comment to PR #782.


ianstormtaylor:
Thank you @AlbertHilb, good catch.

Issue #812, opened 2017-05-12
delete the content of the ol or ul, it's tag keep in dom


coocooooo:
#### Do you want to request a *feature* or report a *bug*?
*bug*

#### What's the current behavior?
when insert a `<ol>` or `<ul>`, then delete it's content, the `<ol>` or `<ul>` tag won't be deleted 

#### What's the expected behavior?
when tags content is empty, the `<ol>` or `<ul>` tag should be removed

coocooooo:
### the screenshot
![gif](https://cloud.githubusercontent.com/assets/7315454/25998932/cd94fb5e-3755-11e7-9194-a933246f554c.gif)

komcal:
@ianstormtaylor Do you have any suggestion about this bug? I also have problem. Maybe I can fix this issue.

ianstormtaylor:
@komcal I don't have a suggestion since I haven't run into it or investigated. I'd welcome a PR though! Thanks

ianstormtaylor:
This issue hasn't had a response in a while, and it's unclear if it's still present in the latest versions of Slate. To keep the issues easier to manage, since there are so many being opened every day, I'm going to close this old one out. 

But if you notice that the bug is still happening, please feel free to comment and it can be revisited. It might be an edge case that not many people run into, in which case the fastest way to get it fixed would probably be to write a pull request for it.

Thanks for understanding!

drenther:
This issue still seems to exist.

![2019-05-06 23 06 04](https://user-images.githubusercontent.com/12991390/57243377-9a9ba980-7053-11e9-9f7c-cb75d345fe4a.gif)

Just wanted to get the discussion going again around this issue. I haven't dug too much into it yet. But, I have been facing this issue myself. I am currently trying to handle it by adding a condition in `normalizeNode` method. I think there might be a better way to do it.

PR #900, opened 2017-06-26
Remove parent of tag when it has empty parent

komcal:
from this [issue](https://github.com/ianstormtaylor/slate/issues/812#issuecomment-310693286)
Is this Ok to do this?


ianstormtaylor:
Hey @komcal, I'm not sure. This should have a test case to ensure that whatever behavior it was trying to fix is fixed and stays fixed in the future. Thanks!

ianstormtaylor:
Hey @komcal I'm going to close this since it's an old pull request. But if this is still something that you need feel free to open a pull request again against the latest codebase.

Issue #772, opened 2017-05-02
Q: Is it possible to drag and drop blocks with a sort of dragging preview block?


deep-c:
Hi,

I am looking to create drag and drop blocks and while i know that selections can be dragged etc my use case is a bit more complicated. Say i have a custom block defined similar to your embeds example where by the YouTube player is embedded within some wrapper component with a drop zone/handle area at the top. Would it be possible to make the block draggable as well as show a sort of 'drag preview' while this block is being dragged within the document? Here is a gif of what i would like to accomplish.

![Alt Text](https://media.giphy.com/media/3o7bu2WlIvBzEEPgU8/giphy.gif)

Also not too certain if it would be possible to use react-dnd inside the editor?

Any insights would be greatly appreciated!

SamyPesse:
We've implemented in a beta version of GitBook using [`react-dnd`](https://github.com/react-dnd/react-dnd).

A placeholder to indicate the drop position is easy, but a "live" dropping when hover is not working because of slate performance (like on this example: http://react-dnd.github.io/react-dnd/examples-sortable-simple.html).

lxcid:
react-dnd is also what I end up using for drag and drop.

oyeanuj:
@SamyPesse @lxcid Would you have a gist/plugin available for integrating `react-dnd` with Slate?

@SamyPesse Also curious about the performance issue in "live" dropping! Could you share more?

deep-c:
@SamyPesse @lxcid While i havent really spent a lot of time yet (just started using it over the last couple of days so im still getting familiar with it) however it would be great to get a bit of an idea of how you approached using react-dnd with slate. For example did u define slate block component(s) a wrapper component which is designated as a drag-source with the editor itself a target? How did you go about reordering the slate state on drop?

thepian:
Very interesting

hmaurer:
@SamyPesse Could you elaborate on this as per @deep-c please? I would be very interesting to hear your approach too!

lxcid:
My use case is probably out of the norm. My drag source and drop target is component outside of react but they wrapped a slate component. But I think I understand slate enough to give some info'ed guess.

You should be able to create drag source as part of your slate components. Best way to do that by defining them in your schema. If you wan to drag block like paragraph, you probably want to override the default rendering in schema to include the drag source and/or drop target there.

The snapshot mechanism is a browser level function but there's way to control that manually. Browser snapshot only work for html based component if I'm not wrong.

[...]

PR #901, opened 2017-06-27
Make onDrop in more cases available for other plugins to use

ms88privat:
At our company we have multiple plugins which are using the `onDrop` handler. We have the issue, that there a some cases where the onDrop event is returned early and not exposed to the plugins.

To get our drop-indicator plugin to work, I need these changes to happen. The Plugin will get open sourced after we decoupled it from our internal code basis.

Could somebody please explain me the reasons behind these if statements?
@ianstormtaylor 

What do you think about this? With this PR and [the other one](https://github.com/ianstormtaylor/slate/pull/903), the drop-indicator plugin is able to do something like this mentioned here: https://github.com/ianstormtaylor/slate/issues/772

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -492,7 +492,6 @@ class Content extends React.Component {
 
   onDrop = (event) => {
     if (this.props.readOnly) return
-    if (!this.isInEditor(event.target)) return
 
     event.preventDefault()
 
@@ -525,8 +524,10 @@ class Content extends React.Component {
       isFocused: true
     })
 
-    // If the target is inside a void node, abort.
-    if (state.document.hasVoidParent(point.key)) return
+    // If the target is inside a void node.
+    if (state.document.hasVoidParent(point.key)) {
+      data.isInsideVoid = true
+    }
 
     // Add drop-specific information to the data.
     data.target = target


ianstormtaylor:
Hey @ms88privat, thanks for the PR!

The case where the drop happens inside a void node makes sense to not swallow to me, since that should be allowed to be handled by plugins. For the other case, where the element isn't inside the editor, what's your use case for that? When would that one fire and be swallowed incorrectly? Thanks!

ms88privat:
@ianstormtaylorI think it was related to the iframe of storybook or social media embeds

ianstormtaylor:
Hey @ms88privat the first case has been handled by merging https://github.com/ianstormtaylor/slate/pull/921 now. I'm not sure about the second case, because it seems like something that goes against the handling logic we've been using elsewhere for Slate. But I'm open to it if you can explain why it's necessary and we should change elsewhere.

ms88privat:
@ianstormtaylor Currently I'm sick and not at work. I will come back early next weak to resolve your questions.

fyi, I'm not using the `data.target` attribute because it was also undefined in this case. But let me recreate this behaviour in an example again.

ianstormtaylor:
Hey @ms88privat this should be solved in newer versions of Slate now! Thanks for bringing this up originally though.

ms88privat:
Just upgraded to the newest version (v30). Looks like both cases are fixed 👍

PR #903, opened 2017-06-28
Enable calling event methods of plugins in sequence

ms88privat:
At the moment every event method of the plugins will get called in `parallel` (plugins can not build up on each others state transformation). The alternative was to return the transformation, but this prevented every other plugin from handling the same kind of event. With this commit I added the ability to choose a third way of doing it. It is fully backwards compatible. 

If you specify `inSequence: true` the next plugin in line will receive the updated state of this transformation in the current plugin.

Example Usage: 
```js
onDrop(event, data, state) {
  return {
      inSequence: true,
      state: state.transform()....apply(),
  }
}
```

What do you think about this? 

This PR and [the other one](https://github.com/ianstormtaylor/slate/pull/901) are needed to get my drop-indicator plugin to work. I will share this code in the next couple of days.

rte-dropindicator

### DIFF
--- a/src/models/stack.js
+++ b/src/models/stack.js
@@ -135,6 +135,34 @@ class Stack extends new Record(DEFAULTS) {
 
 }
 
+function callEventMethodsOfPlugins(index, method, plugins, state, editor, ...args) {
+  const plugin = plugins[index]
+
+  if (!plugin) return state
+
+  const nextIndex = index + 1
+
+  if (!plugin[method]) {
+    return callEventMethodsOfPlugins(nextIndex, method, plugins, state, editor, ...args)
+  }
+
+  const nextState = plugin[method](...args, state, editor)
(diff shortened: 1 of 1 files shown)


ms88privat:
@ianstormtaylor Can we please have any comment on this one :)?

ianstormtaylor:
Hey @ms88privat, thanks for opening this. I think your use case of allowing multiple plugins to act on a single event passing through the stack makes sense. This is actually solved by https://github.com/ianstormtaylor/slate/pull/836 already with the slightly tweaked way of handling transformations. In that case, I'd rather close this and get that PR merged in instead, since it takes us further in the direction the API should go. I hope that makes sense, feel free to respond with more thoughts though!

ms88privat:
Just tested it, works 👍

Issue #508, opened 2016-12-07
[bug] Void node onClick behavior


gabrielrtakeda:
My use case is better described with the screenshot below:
**Use case screenshot**
<img width="163" alt="screen shot 2016-12-07 at 20 09 56" src="https://cloud.githubusercontent.com/assets/5435389/20988812/75fb4dba-bcb9-11e6-8b18-77d5f70e7114.png">
When I click in the text ("Hello" or "World"), the link works but, when I click on the image (void inline image), the link don't works.

I have tested with `readOnly` prop set to `true` in the Editor component and, when I click on the image, it seems that the click was dispatching some action that enables `state.selection.hasEdgeIn(node)` to true, avoiding the default behaviour of the `<a />` tag.

**When I click on the image, even when `readOnly={true}`**
<img width="157" alt="screen shot 2016-12-07 at 20 23 25" src="https://cloud.githubusercontent.com/assets/5435389/20989241/265f4e26-bcbb-11e6-9390-7721b533396e.png">

The code that I working on currently is below:

**ImageNode**
```js
const ImageNode = ({ node, state, attributes }) => {

  const isFocused = state.selection.hasEdgeIn(node)
  const src = node.data.get('src')
  const className = isFocused ? 'active' : null

  return (
    <img
      {...attributes}
      role="presentation"
      className={className}
      src={src}
    />
  )
}
```

**ImageTransform**
```js
export const insertImage = state => {
  const src = 'https://orig06.deviantart.net/7869/f/2009/131/9/e/zubat_icon_by_the_fry_bat.gif'
  return state
    .transform()
    .insertInline({
      type: 'image',
      isVoid: true,
      data: { src }
    })
    .focus()
    .apply()
}
```

**LinkNode**
```js
const LinkNode = ({ node: { data }, attributes, children }) => (
  <a
    {...attributes}
    href={data.get('href')}
  >
    {children}
  </a>
)
```

**LinkTransform**
```js
export const hasLinks = state => state.inlines.some(inline => inline.type === 'link')

export const insertLinkStrategy = state => {
  if (hasLinks(state)) {
    return state
      .transform()
      .unwrapInline('link')
      .apply()
  } else if (state.isExpanded) {
    const href = window.prompt('Enter the URL of the link:')
    const newState = state
      .transform()
      .wrapInline({
        type: 'link',
        data: { href }
      })
      .collapseToEnd()
      .apply()
    return newState
  } else {
[...]

PR #908, opened 2017-07-03
Ignore Void nodes onClick in read-only

Soreine:
As pointed here, in read-only mode, Void nodes onClick still triggers.

https://github.com/ianstormtaylor/slate/issues/508#issuecomment-265607814


ianstormtaylor:
Good call @Soreine, thanks!

Issue #910, opened 2017-07-04
Get selected text of first inline node in a block [Bug]


rholliger:
#### Do you want to request a *feature* or report a *bug*?

A Bug.

#### What's the current behavior?

Hello folks.
I am currently adding a possibility to edit a link and his text, after it was added to the editor.
For this I need to get the text I just selected.

#551 told me, that I can get the selected text with `state.fragment.text`

Now I ran into a problem when selecting a link. The bug only occurs if the link is the first element in a new block and the selection starts from the first character (Shift + Right-Arrow).
Here's a recording of the bug:

![](http://g.recordit.co/BhqvTAMnlN.gif)

I assume that it has something to do with the zero-width character, which is wrapped around an inline node and the splitting of this node and the subsequent node.

![storybook](https://user-images.githubusercontent.com/6562926/27833145-c8360eb4-60d1-11e7-9d81-b1cb53640291.jpg)

Is this a known bug? Is there another way of selecting a text of an inline node?

ianstormtaylor:
This issue hasn't had a response in a while, and it's unclear if it's still present in the latest versions of Slate. To keep the issues easier to manage, since there are so many being opened every day, I'm going to close this old one out. 

But if you notice that the bug is still happening, please feel free to comment and it can be revisited. It might be an edge case that not many people run into, in which case the fastest way to get it fixed would probably be to write a pull request for it.

Thanks for understanding!

PR #913, opened 2017-07-06
Handle edge case in getFragmentAtRange

ms88privat:
fixes https://github.com/ianstormtaylor/slate/issues/910

At the moment `getFragmentAtRange` assumes, that the Text to split will be part of the first split (`start`). Thats why the `end` will be at the `nextText` Node.

In the case of an `inline` Node at the first position of a new line, it will split the fake empty text before it. The second split will then raise an error. (That's what I think is happening)

### DIFF
--- a/src/models/node.js
+++ b/src/models/node.js
@@ -585,18 +585,35 @@ const Node = {
     node.assertDescendant(startKey)
     node.assertDescendant(endKey)
 
+    const originalFurthestAncestor = node.getFurthestAncestor(startKey)
+
     // Split at the start and end.
     const start = range.collapseToStart()
     node = node.splitBlockAtRange(start, Infinity)
 
-    const next = node.getNextText(startKey)
+    const furthestAncestor = node.getFurthestAncestor(startKey)
+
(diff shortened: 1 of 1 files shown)


ms88privat:
Thanks for the feedback, will do!

ianstormtaylor:
Hey @ms88privat I've since refactor `getFragmentAtRange`, so I'm going to close this because I think it may have already been fixed. Thank you for the PR!

ms88privat:
@ianstormtaylor Thanks for the new slate version. It's awesome! But this issue was not resolved, but it does behave differently now. It will not select the **first or the last char** of an **inline node** at the **first position of a new line**. In the earlier version it did throw an error instead, so this is better ^^

You can see it in this gif:
- Selecting normal text with (shift + arrow keys)
- Transforming the text to a Link (inline node)
- Selecting Link doesn't work correctly

http://recordit.co/CoNRHVA00w


Or try it urself in the example: http://slatejs.org/#/links

It does work if you select it with your mouse tough!

PR #914, opened 2017-07-07
Issue with the Overriding Defaults example.

eamontaaffe:
I tried using the exact code from the example to be met by errors.

The order of arguments in the function example is wrong, it should be use the API described in the [docs](https://docs.slatejs.org/reference/plugins/plugin.html#onbeforeinput).

```
Function onBeforeInput(event: Event, data: Object, state: State, editor: Editor) => State || Void
```


ianstormtaylor:
Thanks @eamontaaffe!

Issue #539, opened 2016-12-24
HtmlSerializer cannot deserialize simple string


LeeKevin:
`HtmlSerializer.deserialize('test')` will raise an error:

> TypeError: Cannot read property 'kind' of undefined

```
	getLeafText	@	get-leaf-text.js:15
        collapseToStartOf	@	selection.js:367
        create	@	state.js:537
        deserializeState	@	raw.js:238
        deserialize	@	raw.js:62
        deserialize	@	html.js:206
```

It should create a text block (or an equivalent) for the string. Currently the nodes passed to the Raw serializer is an empty array.

ianstormtaylor:
Thanks for reporting this @LeeKevin. I agree with you, and I'd be happy to accept a pull request that allows for simple string deserializing! Let me know if you need any guidance on putting one together. Thanks!

vpontis:
I was looking at this as a bug to fix but I didn't find the `HtmlSerializer` in the codebase... 

Ah, I think this is actually called `Html` in `/src/serializers/html.js`. Maybe it got updated...

PR #915, opened 2017-07-07
Allow HTMLSerializer to deserialize an empty doc

conorcussell:
Thanks to @vpontis for doing the actual work, I noticed that all we need to close this PR https://github.com/ianstormtaylor/slate/pull/742  and this issue https://github.com/ianstormtaylor/slate/issues/539 was a test case to cover deserialising an empty string. Let me know if this looks ok?

### DIFF
--- a/src/serializers/html.js
+++ b/src/serializers/html.js
@@ -96,6 +96,11 @@ class Html {
     const children = $.children().toArray()
     let nodes = this.deserializeElements(children)
 
+    const { defaultBlockType } = this
+    const defaults = typeof defaultBlockType == 'string'
+      ? { type: defaultBlockType }
+      : defaultBlockType
+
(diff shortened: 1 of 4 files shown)


ianstormtaylor:
Hey @conorcussell thanks! Is there any reason we can use the existing test fixtures setup? (Instead of creating another test case with `it(` with custom logic?)

conorcussell:
I'm sure i thought there was a reason i couldnt use them at the time, but looking again now it will work fine with the normal fixtures setup. Will fix! Thanks

conorcussell:
Hey @ianstormtaylor changed it to use the fixtures now! Thanks

ianstormtaylor:
Thanks @conorcussell!

PR #916, opened 2017-07-07
enable source maps during development

erquhart:
Enables inline sourcemaps for the lib watch task


ianstormtaylor:
Thanks @erquhart!

Issue #556, opened 2017-01-17
Nested blocks issue


Mangatt:
There is problem with transformations of doubly nested blocks (document - documentNode - nestedNode - doubleNestedNode). There is problem in following cases:

- caret at end of the doubleNestedNode, press delete
- caret at beggining of (not first) node, press backspace
- paste selection of more than one block

```javascript
Uncaught Error: Could not find a node with key "47".
    at Document.assertNode
    at Object.normalizeNodeByKey
    at Transform.(anonymous function) [as normalizeNodeByKey] 
    at Object.deleteAtRange
    at Transform.(anonymous function) [as deleteAtRange]
    at Object.deleteBackwardAtRange
    at Transform.(anonymous function) [as deleteBackwardAtRange] 
    at Object.deleteCharBackwardAtRange
    at Transform.(anonymous function) [as deleteCharBackwardAtRange] 
    at Object.deleteCharBackward
```

Double nested blocks have actually multiple issues and strange behaviors that regular nested blocks does not have. For example backspace with caret at first double nested block deletes without producing error, but deletes whole content of nestedBlock keeping only first doubleNestedBlock.

I am aware that this could be solved with custom plugin, but since Slate supports unlimited nesting of blocks, this should be probably solved in core.

Mangatt:
To be precise, error is produced everytime when editor tries to merge multiple doubleNestedBlocks.

ianstormtaylor:
Hey @Mangatt thanks for the report, can you create a smallest-possible JSFiddle to reproduce this so that it's easier for people to see what causes the bug? That will help in terms of it being resolved. Check out the [`Contributing.md`](https://github.com/ianstormtaylor/slate/blob/master/Contributing.md) file for a template to use. Thanks!

Mangatt:
Here it is, @ianstormtaylor :

https://jsfiddle.net/3mherqht/1/

Place cursor at beginning of line 2, right before text DoubleNested2. Press backspace. Or try any of bullets above.

Mangatt:
I guess that this is connected, but transform.moveNodeByKey has this issue as well. It behaves correctly until I try to move node into doubleNestedBlock.

When I removeNodeByKey and subsequently insertNodeByKey, everything works just fine, although I guess that moveNodeByKey could internaly be just combination of these two.

Mangatt:
When I was testing behavior of my PR #563, I came across strange behavior in nested blocks:

https://jsfiddle.net/3mherqht/2/

[...]

PR #917, opened 2017-07-07
fix insertion of multiple deeply nested blocks

erquhart:
This fix attempts to address an issue where grandparents of an inserted
fragment's first leaf block are inadvertently removed from the fragment,
producing an error when Slate tries to access the missing nodes.

### DIFF
--- a/src/transforms/at-range.js
+++ b/src/transforms/at-range.js
@@ -596,21 +596,28 @@ Transforms.insertFragmentAtRange = (transform, range, fragment, options = {}) =>
   const blocks = fragment.getBlocks()
   const firstBlock = blocks.first()
   const lastBlock = blocks.last()
+  const multipleLeafBlocks = firstBlock != lastBlock
+  const firstParent = fragment.getFurthest(firstBlock.key, p => p.nodes.size === 1)
+  const lastParent = fragment.getFurthest(lastBlock.key, p => p.nodes.size === 1)
+  const commonParent = firstParent && lastParent && firstParent == lastParent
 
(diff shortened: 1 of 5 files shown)


erquhart:
Related to #556.

Minimal JSFiddle: https://jsfiddle.net/erquhart/L05uz8qp/1/

A research excerpt from my comment on that issue:

* Insert a fragment where two sibling leaf blocks are nested more than one layer deep.
* Example: A list with a single item, and that item contains two paragraphs with text.
* `Transform.insertFragmentAtRange` checks the first and last leaf blocks, and if they are not the same, it removes the parent of the first leaf block pending a potential node split. The intention is to add the first leaf node parent back in after the split.
* In this case, the first and last leaf node share the same parent, so both nodes are inadvertently removed. I believe this is due to [this line](https://github.com/erquhart/slate/blob/5d8a9b99ccd5840b3a5c53822988b3e10186849a/src/transforms/at-range.js#L609), where the iterator for finding `lonelyParent` simply requires the parent to contain only one node. **This doesn't account for that one node having multiple children.**
* We then attempt to move the last leaf node into the document, but it's been removed from the fragment, hence the error associated with `moveNodeByKey`.

erquhart:
Adding WIP, found at least one regression caused by this.

ianstormtaylor:
Hey @erquhart thanks for this. Let me know when you've figured out the regressions!

bkniffler:
Hey, this one is very important, I've been bitten by these errors (copy/paste a bunch of nodes) since the beginning of slate! Thanks for your work :)

PR #918, opened 2017-07-09
Fix Windows dev install

mjadobson:
There were a few errors thrown when trying to set up Slate dev on Windows 10.

This pull request alters the npm scripts so that they will work cross platform. Additionally, it adds a note that `windows-build-tools` may be required before attempting to install.


erquhart:
@mjadobson CI build is failing - you'll need to call the `mkdirp` binary directly. Example:

```
"build": "./node_modules/mkdirp/bin/cmd.js ./tmp && ..."
```

Otherwise it seems to work great - awesome improvement!

mjadobson:
That's interesting. Do you know why that's the case? I thought it would have installed the devDeps and been able to run them like it's doing with eslint/browserify/etc

erquhart:
Yeah that does make sense. Might be yarn, checking.

erquhart:
Yep, it's yarn. Very strange. Known issue: https://github.com/yarnpkg/yarn/issues/760.

It's only affecting `mkdirp`. Might be yarn mishandling `bin` when defined as a path instead of an object (`npm-run-all` binaries are mapped out in an object, for example). Does installing with yarn work for you locally?

brendancarney:
Closing, as this looks pretty out of date now. If there are still Windows issues, please feel free to update the PR or create an issue. Thanks!

PR #919, opened 2017-07-09
Fix image drop in example

mjadobson:
Dropped images were being inserted at the previous selection rather than the drop target.


ianstormtaylor:
Good call, thank you @mjadobson!

PR #920, opened 2017-07-09
Allow drop on void nodes

mjadobson:
As described here - https://github.com/ianstormtaylor/slate-drop-or-paste-images/issues/9, it isn't possible to fire drop events when dragging items onto void nodes.

This pull request adds drag handlers to the Void component, to temporarily switch off the `contenteditable=false` div, and adds new behaviour, so dropped text is inserted into the nearest non-void text node.


mjadobson:
Closed as mixed commits in branch.

PR #921, opened 2017-07-09
Allow drop events on void nodes

mjadobson:
As described here - https://github.com/ianstormtaylor/slate-drop-or-paste-images/issues/9, it isn't possible to fire drop events when dragging items onto void nodes.

This pull request adds drag handlers to the Void component, to temporarily switch off the `contenteditable=false` div, and adds new behaviour, so dropped text is inserted into the nearest non-void text node.

### DIFF
--- a/src/plugins/core.js
+++ b/src/plugins/core.js
@@ -399,18 +399,16 @@ function Plugin(options = {}) {
       .transform()
       .select(target)
 
-    let hasVoidParent = document.hasVoidParent(target.get('anchorKey'))
+    let hasVoidParent = document.hasVoidParent(target.anchorKey)
 
-    // For void targets, we want to find the nearest non-void text node
+    // Insert text into nearest text node
     if (hasVoidParent) {
-      let node = document.getNode(target.get('anchorKey'))
+      let node = document.getNode(target.anchorKey)
 
       while (hasVoidParent) {
-        node = document.getNextText(node.get('key'))
-
+        node = document.getNextText(node.key)
         if (!node) break
-
-        hasVoidParent = document.hasVoidParent(node.get('key'))
+        hasVoidParent = document.hasVoidParent(node.key)
       }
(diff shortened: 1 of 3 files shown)


src/plugins/core.js
```diff
@@ -394,10 +394,28 @@ function Plugin(options = {}) {
...
+    const { document } = state
     const transform = state
       .transform()
       .select(target)
 
+    let hasVoidParent = document.hasVoidParent(target.get('anchorKey'))
+
+    // For void targets, we want to find the nearest non-void text node
+    if (hasVoidParent) {
+      let node = document.getNode(target.get('anchorKey'))
```
ianstormtaylor: Can you change all of these `.get(` calls to just use property notation, like `target.anchorKey` instead? Thanks!

src/plugins/core.js
```diff
@@ -394,10 +394,28 @@ function Plugin(options = {}) {
...
+
+      while (hasVoidParent) {
+        node = document.getNextText(node.get('key'))
+
```
ianstormtaylor: Can you remove the extra line breaks here?

src/plugins/core.js
```diff
@@ -394,10 +394,28 @@ function Plugin(options = {}) {
...
 
     const { text, target } = data
+    const { document } = state
     const transform = state
       .transform()
       .select(target)
 
+    let hasVoidParent = document.hasVoidParent(target.get('anchorKey'))
+
+    // For void targets, we want to find the nearest non-void text node
```
ianstormtaylor: Can you clarify this comment to just mention that we'll be inserting the text into the next nearest text node? Right now it's not very clear as written.

mjadobson:
Those comments should all be addressed now; thanks!

ianstormtaylor:
Thanks @mjadobson!

Issue #922, opened 2017-07-11
Drag and Drop bugs with Edge


mjadobson:
#### Do you want to request a *feature* or report a *bug*?

Bug

#### What's the current behavior?

![](http://g.recordit.co/QLDUccWBBy.gif)

- "stop" sign when dragging over most of editor contents, except narrow area in between blocks.
- Unable to drag external text into editor content
- The rich text example's current code restricts dragging and dropping editor contents (described here - https://github.com/ianstormtaylor/slate/pull/884), however in Edge, if you carefully drop in between blocks it will insert it.

#### What's the expected behavior?

- No "stop sign" when dragging text over editor contents
- Restrict dragging and dropping editor contents (https://github.com/ianstormtaylor/slate/pull/884)
- Fire drop events

mjadobson:
So it turns out there's a few hurdles with this:

- `ondragover` DragEvents don't have permission to `getData()`. This is true for all browsers I think [1].
- `ondragover` needs to have `e.preventDefault()` called to prevent the "stop" sign preventing drops, but this isn't consistently done due to the issue above.
- `dataTransfer` properties `dropEffect` and `effectAllowed` are "Permission Denied" in Edge, which throws errors when trying to access them [2].
- In Edge, errors are thrown if you try to set a custom data type in `setData()`. [3][4]
- Also, while `dataTransfer` methods `getData()` and `setData()` work currently, they are marked as deprecated in Edge [5]

Will have a go at working around them.

[1] http://www.w3.org/TR/html51/editing.html#protected-mode
[2] ![drag-event-denied-permission](https://user-images.githubusercontent.com/787996/28066447-dc351fa0-6634-11e7-939d-549803dea49b.png)
[3] https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8007622/
[4] https://codepen.io/anon/pen/ZOdwAZ
[5] https://msdn.microsoft.com/en-us/library/ms535861(v=vs.85).aspx

mjadobson:
Sorted by #923

PR #923, opened 2017-07-11
Fix Edge Drag & Drop Issues

mjadobson:
As described here: https://github.com/ianstormtaylor/slate/issues/922

A few weird things to work around; the main problem was the custom node and fragment types causing errors to be thrown with `dataTransfer.setData()`. Now when this happens, all custom data is stored in JSON object under `text/plain`.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -527,7 +527,8 @@ class Content extends React.Component {
     // Add drop-specific information to the data.
     data.target = target
 
-    // Edge throws "Permission denied" errors when accessing `dropEffect` or `effectAllowed`
+    // COMPAT: Edge throws "Permission denied" errors when
+    // accessing `dropEffect` or `effectAllowed` (2017/7/12)
     try {
       data.effect = dataTransfer.dropEffect
     } catch (err) {
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -14,6 +14,7 @@ import findClosestNode from '../utils/find-closest-node'
 import findDeepestNode from '../utils/find-deepest-node'
 import getPoint from '../utils/get-point'
 import getTransferData from '../utils/get-transfer-data'
+import setTransferData from '../utils/set-transfer-data'
 import { IS_FIREFOX, IS_MAC } from '../constants/environment'
 
 /**
@@ -473,7 +474,8 @@ class Content extends React.Component {
     const { state } = this.props
     const { fragment } = state
     const encoded = Base64.serializeNode(fragment)
-    dataTransfer.setData(TYPES.FRAGMENT, encoded)
+
+    setTransferData(dataTransfer, TYPES.FRAGMENT, encoded)
 
     debug('onDragStart', { event })
   }
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -444,13 +444,7 @@ class Content extends React.Component {
   onDragOver = (event) => {
     if (!this.isInEditor(event.target)) return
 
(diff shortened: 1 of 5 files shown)


src/components/content.js
```diff
@@ -444,13 +445,7 @@ class Content extends React.Component {
...
     if (!this.isInEditor(event.target)) return
 
-    const { dataTransfer } = event.nativeEvent
-    const data = getTransferData(dataTransfer)
-
-    // Prevent default when nodes are dragged to allow dropping.
-    if (data.type == 'node') {
-      event.preventDefault()
-    }
+    event.preventDefault()
```
ianstormtaylor: Why don't we need to check for `type == 'node'` anymore?

src/components/content.js
```diff
@@ -530,7 +526,13 @@ class Content extends React.Component {
 
     // Add drop-specific information to the data.
     data.target = target
-    data.effect = dataTransfer.dropEffect
+
+    // Edge throws "Permission denied" errors when accessing `dropEffect` or `effectAllowed`
```
ianstormtaylor: Can you check out the other `COMPAT:` comments in the repo and make this one match those? Thanks

src/utils/get-transfer-data.js
```diff
@@ -57,6 +73,30 @@ function getTransferData(transfer) {
...
+
+  if (text.substring(0, prefix.length) !== prefix) {
+    return { 'text/plain': text }
+  }
+
+  // Attempt to parse, if fails then just standard text/plain
+  // Otherwise, already had data embedded
+  try {
+    return JSON.parse(text.substring(prefix.length))
+  } catch (err2) {
```
ianstormtaylor: Why the need to name this `err2`?

src/utils/get-transfer-data.js
```diff
@@ -37,17 +37,33 @@ function getTransferData(transfer) {
...
   if (node) node = Base64.deserializeNode(node)
 
-  // Get and normalize files if they exist.
-  if (transfer.items && transfer.items.length) {
-    files = Array.from(transfer.items)
-      .map(item => item.kind == 'file' ? item.getAsFile() : null)
-      .filter(exists => exists)
-  } else if (transfer.files && transfer.files.length) {
-    files = Array.from(transfer.files)
+  // Edge sometimes throws 'NotSupportedError' whena accessing `transfer.items`
```
ianstormtaylor: Can you clean up this `COMPAT:` comment too please?

src/utils/get-transfer-data.js
```diff
@@ -37,17 +37,33 @@ function getTransferData(transfer) {
     if (encoded) fragment = encoded
   }
 
+  // Handle embedded types in text (Edge doesn't handle custom data types)
```
ianstormtaylor: Can you change this to a `COMPAT:` comment?

src/components/content.js
mjadobson: It wasn't being checked anyway; it's not possible to access this information within ondragover. The DragEvent is in "protected mode" (one reason for this mode is to prevent data sniffing when you drag past other windows to get to the desired drop destination.)

src/utils/get-transfer-data.js
mjadobson: Copy paste job haha

mjadobson:
Those comments should all be addressed now; thanks!

ianstormtaylor:
Thanks @mjadobson!

PR #925, opened 2017-07-12
Use `block.getClosest` instead of `document.getClosest`

hubgit:
`getClosest` is [a method of Slate's `Node` model](https://docs.slatejs.org/reference/models/node.html#getclosest)


YurkaninRyan:
@hubgit I'm confused by this PR.  isn't Document is the top level node of the Slate Document Model? Also I think that you could call getClosest on any block but it will only work for the tree below it

hubgit:
Sorry, yes, my mistake - I lost track of `document` during refactoring and thought it was referring to `window.document`.

Issue #706, opened 2017-04-01
optimize some of the hot code paths


ianstormtaylor:
This isn't a specific actionable item, but I figured I'd just throw it here so others can have the opportunity to help if they're interested...

Right now, there are a handful of "hot" methods that are invoked for lots of actions. Specifically I'm concerned with a few of the most common actions:

- rendering
- inserting single characters
- backspacing single characters
- splitting nodes

Some of the methods that are often invoked are:

- `Node.getAncestors`
- `Node.getChild`
- `Node.getClosestBlock`
- `Node.getDescendant`
- `Node.getKeys`
- `Node.getNextText`
- `Node.getParent`
- `Node.getPath`
- `Node.getTexts`
- `Node.updateDescendant`
- `Selection.hasAnchorIn`
- `Selection.hasFocusIn`
- `Text.getDecorations`
- `Text.getRanges`

Which is to say, any (even small) performance improvements to how those methods work can result in potentially large gains when multiplied across how many times they are invoked. So if you're looking for a place to help improve performance, those might be a good start! For a few examples of previous simple changes that are improvements:

- https://github.com/ianstormtaylor/slate/pull/705/commits/124c0eea9aad3be8396320f69b3dbcb3e3a5f7b3
- https://github.com/ianstormtaylor/slate/pull/705/commits/ac4bd8487790d14345eaae89d38424e42edfed37
- https://github.com/ianstormtaylor/slate/pull/705/commits/f8a7fae890a3eef67926d03b5901b999c654e2c7

And for an example of how to look into the performance and test that it's improved:

- https://github.com/ianstormtaylor/slate/pull/794

Thanks! 😄

ianstormtaylor:
Closing because it's old and current hot code paths have probably changed some, but there are still lots of things to help optimize!

PR #926, opened 2017-07-12
improve mark handling performance for large selections

erquhart:
Related to #706.

When highlighting a 40k character document (not an uncommon blog post size for larger web publications) and applying marks, scripting time compounds with each additional mark, mostly due to the performance of `Node.getMarksAtRangeAsArray`. This was only tested in Chrome, but it seems that concatenating arrays at scale is less performant than looping and pushing.

**Using `concat`:**
* Add first mark: ~9s
* Add second mark: ~21s
* Add third mark: ~35s
* Total time: ~65s
screen shot 2017-07-12 at 12 39 11 pm

**Using `forEach` + `push`:**
* Add first mark: ~6s
* Add second mark: ~6s
* Add third mark: ~7s
* Total time: ~19s
screen shot 2017-07-12 at 12 48 54 pm

### DIFF
--- a/src/models/node.js
+++ b/src/models/node.js
@@ -904,7 +904,8 @@ const Node = {
     return this
       .getCharactersAtRange(range)
       .reduce((memo, char) => {
-        return memo.concat(char.marks.toArray())
+        char.marks.toArray().forEach(c => memo.push(c))
+        return memo
       }, [])
   },
 


ianstormtaylor:
Interesting, thanks @erquhart!

Issue #927, opened 2017-07-12
DnD text sometimes inserted in wrong location - also no drop cursor


mjadobson:
#### Do you want to request a *feature* or report a *bug*?

Bug/feature

#### What's the current behavior?

![](http://g.recordit.co/Yd9Uw3fs2v.gif)

- Text sometimes inserted at wrong location.
- No visual cursor visible to show insertion location while dragging text.

#### What's the expected behaviour?

- Text should be inserted at correct drop location.
- Visual cursor should be visible to show insertion location while dragging text.

---

- I'm not sure why the insertion point sometimes isn't correct, will investigate.
- In Prosemirror there is a plugin that adds a fake dropcursor (1px border-left `<span>`) [1]

[1] https://github.com/ProseMirror/prosemirror-dropcursor

mjadobson:
Ok, so the drop cursor can be brought back quite simply by only calling `e.preventDefault()` for external drags (to prevent some browsers loading dropped external files/links).

However, this change breaks Edge support again. I've looked into it some more and bizarrely the main issue is the `<span>` elements used in Slate. For some reason Edge doesn't let you drop onto inline elements such as `<span>` or `<a>`. The fix seems to be set the style of these elements to `display: inline-block`.

https://jsfiddle.net/69z2wepo/82817/
![](http://g.recordit.co/ovpnVh32Hi.gif)

mjadobson:
Looked into this some more, and Trix fixes this by calling preventDefault in ondragover, and then brings back the cursor by using the mouse pointer location to manually set the selection during the drag. This approach works in Edge & Chrome

ianstormtaylor:
I believe this is fixed in the newest version of Slate. Let me know if you're still seeing it though!

mjadobson:
The cursor is now visible, but as discussed above, it has brought back the issue in Edge where it is impossible to drag+drop text.  Without `e.preventDefault()`, dragging text in Slate displays a stop/restricted sign blocking drops. The only fix for this would be to preventDefault, then manually set the cursor position, as in Trix, or add a fake cursor, as in Prosemirror.

ichbinedgar:
Hi, this issue should be open. The cursor is visible now, but it's still inserting in the wrong location sometimes.

PR #928, opened 2017-07-12
Fix drop cursor

mjadobson:
This change brings back the drop cursor, which became invisible after the commits for Edge support.

This pull request will break Edge support for DnD, but we can get that back with the strategies discussed here: https://github.com/ianstormtaylor/slate/issues/927

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -445,7 +445,9 @@ class Content extends React.Component {
   onDragOver = (event) => {
     if (!this.isInEditor(event.target)) return
 
-    event.preventDefault()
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @mjadobson thank you very much for this and I'm sorry for the delay. I fixed this with https://github.com/ianstormtaylor/slate/pull/1040

mjadobson:
It's a while since I worked on this, but I believe it was required to allow dropping links/images/etc from outside the editor

https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Drag_operations#droptargets

PR #929, opened 2017-07-13
Fix drop contents insertion

mjadobson:
Now drop fragments insert at correct location. As described here: https://github.com/ianstormtaylor/slate/issues/927


ianstormtaylor:
Good call @mjadobson, thank you!

Issue #924, opened 2017-07-11
Warning: Failed prop type: Invalid prop `state` of type `State` supplied to `Editor`, expected instance of `State`.


jatins:
#### Do you want to request a *feature* or report a *bug*?
bug

#### What's the current behavior?
<img width="1016" alt="screen shot 2017-07-12 at 2 19 53 am" src="https://user-images.githubusercontent.com/1556536/28090052-a79fb952-66a8-11e7-80b7-9c4f551ec20d.png">

#### What's the expected behavior?
no error

jatins:
Seems to be failing some type check.

<img width="908" alt="screen shot 2017-07-12 at 6 50 18 pm" src="https://user-images.githubusercontent.com/1556536/28119606-151edb0e-6733-11e7-8441-7187b2d7fcad.png">

Although, looking at `Watch` tab, it should not happen.

jatins:
Ah, I think I figured out the issue. I am using [lerna](https://github.com/lerna/lerna) to manage packages in my code. 

* The component I was passing props _from_ and the editor component which is using those props are two different (npm) packages.
* Both of them have `slate` as their dependency. 

That's why the `instanceof` check is failing. Because the instance is created using one instance of slate, and being checked against other.

YurkaninRyan:
I've encountered similar problems using other Plugins with instanceof

ianstormtaylor:
Open to ideas on how to get around this, based on what other libraries are doing.

YurkaninRyan:
@ianstormtaylor I'm gonna do some research and see if I can get a pull request rolling.  There must be a standard way seems like a common problem

YurkaninRyan:
Looks like moment sets a "private" property on the prototype.  Should we just set a flag somewhere to compare against?

ianstormtaylor:
Is there any prior art from other projects that use Immutable.js that use a different approach?

YurkaninRyan:
Hmm, well look at immutable itself they use this https://github.com/facebook/immutable-js/blob/master/src/Predicates.js#L10

Which seems to be a combination of private flag checking and internal instance checking?

ianstormtaylor:
Doing that way with pseudo-Symbols sounds good to me.

PR #930, opened 2017-07-13
Remove instanceOf checks to allow Slate objects to be identifiable across module instances

YurkaninRyan:
IN PROGRESS

related to #924 

Currently, if there are two Slate module instances, instanceOf will fail, even if the object we are passing between the two of them are of the correct type.

To get around this I am in favor of the pseudo-symbol approach.  Basically every data model we want to check against will gain a @@__SLATE-${MODEL_NAME}__@@ key.  This is allows for "instance checking" across module instances.

Just looking for some opinions and for anyone to help me hunt for potential code smell from what I'm writing.  Currently i'm trying to get all the tests passing, but it's a bit difficult because there are some side-effects.  Now the models are getting these pseudo-symbols, they also are coming out in the output of Raw.

Opinions?


ianstormtaylor:
Hey @YurkaninRyan thanks for working on this. A few opinions for you...

I think we should change the type checking to be `Block.isBlock()`, mimicking how Immutable.js exposes its checking methods. That way we don't need to expose the constants themselves, and users can just us the static methods on the models.

Second, I think we should add the instance-checking key to the prototype, instead of adding it into the `Record`s default, this will prevent it from showing up in the JSON output I believe.

For the prop types, I think we should expose a set of helper functions, via a `PropTypes` import from Slate. Which could later be a separate `slate-prop-types` module that people can import and that gets used internally.

YurkaninRyan:
Awesome! I agree with you on all points and will aim for that accordingly

YurkaninRyan:
Hey @ianstormtaylor I updated the code to be closer to what you were suggesting I still need to

1. Create the Prop-Types for custom slate objects
2. Get a couple of the tests passing

Is this update more in line with what you were thinking?

YurkaninRyan:
Alright so, travis will confirm in a second but it looks like on my end all tests are passing, and i've updated the documentation.

Should we add tests for these?  Are we okay with where I put the prop types?

src/models/slate-prop-types.js
```diff
@@ -0,0 +1,172 @@
+import Block from './block'
```
ianstormtaylor: Can you move this file to `./utils/prop-types`?

src/models/slate-prop-types.js
```diff
@@ -0,0 +1,172 @@
...
+  }
+)
+
+/**
+ * Exported Slate proptypes
+ *
+ * @type {Object}
+ */
+
+export default {
```
ianstormtaylor: Can you change these exports to be just `block`, `character`, etc. instead of `slate*` prefixed?

src/models/types.js
```diff
@@ -0,0 +1,26 @@
+/**
```
ianstormtaylor: Can you move this file to `./constants/model-types`? And then also change the existing `./constants/types` to be called `./constants/transfer-types`? Thanks!

src/models/types.js
```diff
@@ -0,0 +1,26 @@
...
+ * Slate-specific item types.
+ *
+ * @type {Object}
+ */
+
+const TYPES = {
+  IS_SLATE_STATE: '@@__SLATE_STATE__@@',
```
ianstormtaylor: Can you change these to remove the `IS_SLATE` prefix? Since we're only using them internally, we don't need the extra verbosity.

src/models/inline.js
```diff
@@ -14,6 +13,7 @@ import Block from './block'
 import Data from './data'
 import Node from './node'
 import Text from './text'
+import TYPES from './types'
```
ianstormtaylor: Given the file name change (from another comment), can we change all of these `TYPES` variable names to `MODEL_TYPES`? Thanks!

YurkaninRyan:
@ianstormtaylor Added all the requested changes and fixed the merge conflict ^.^

src/components/content.js
```diff
@@ -476,7 +476,7 @@ class Content extends React.Component {
     const { fragment } = state
     const encoded = Base64.serializeNode(fragment)
 
-    setTransferData(dataTransfer, TYPES.FRAGMENT, encoded)
+    setTransferData(dataTransfer, MODEL_TYPES.FRAGMENT, encoded)
```
ianstormtaylor: I think this should be `TRANSFER_TYPES`?

src/utils/prop-types.js
```diff
@@ -0,0 +1,172 @@
...
+)
+
+/**
+ * Exported Slate proptypes
+ *
+ * @type {Object}
+ */
+
+export default {
+  slateBlock,
```
ianstormtaylor: Can you change all of these to remove the `slate*` prefix? So that they are just plain `block`, `character`, etc.

ianstormtaylor:
Thanks @YurkaninRyan!

PR #934, opened 2017-07-15
Replace cheerio with parse5

### LINKED ISSUE #539: HtmlSerializer cannot deserialize simple string

### LINKED ISSUE #771: Remove Cheerio
Proposing this, not sure of how hard it will be, but cheerio is designed for the server and comes with a ton of node core dependencies it wants. It was causing me some headaches in terms of build size (Buffer and other files coming in are huge).

Separately, recently tried to use webpackbin to show some code on Slate, and it failed. See this issue: https://github.com/cerebral/webpackbin/issues/17#issuecomment-298134310

For a browser-focused lib, seems we could do better.

schneidmaster:
Hi there!

Thanks for making such a great library. I wanted to dig into the code a bit so I decided to tackle an open issue.

This PR is mostly a drop-in replacement of cheerio with parse5. I left cheerio in devDependencies for now since there are a few usages in the tests that seem a little more in-depth and since the main point is to avoid schlepping around extra deps in production. I also added a function to be run on each element to convert parse5's output and make it match cheerio's structure, thus minimizing necessary code changes (and potential API breakages).

There is currently 1 breaking test but I wanted to discuss before fixing. The test is on 

```html
<p><hr /></p>
<div><hr /></div>
```

and can be trivially reproduced as `<p><hr /></p>`. Cheerio parses it as a `p` node with a child `hr` node (which makes logical sense) but parse5 parses it as three empty nodes: `p`, `hr`, and another `p`. Oddly enough, the native `DOMParser` in latest Chrome agrees with parse5's interpretation, which made me wonder if this is really a Cheerio bug and perhaps I should update the test expectation rather than changing the behavior so the existing test passes.

I saw your note in #771 and considered using the native `DOMParser` with `parse5` as a polyfill/backup, but it isn't clear to me what the benefit of that is. It creates potential inconsistencies and we have to ship out parse5 anyway (since we don't know til runtime whether the browser supports `DOMParser`) so there's no bundle size savings to be had. Let me know if you do want me to implement that though.

Finally, I was hoping this might address #539 as well but it doesn't since slate still assumes the presence of some blocks in post-parse operations. #915 should still work fine in conjunction with this change though (slight modification might be needed).

Cheers,
Zach

Fixes #771


schneidmaster:
Actually, I guess upon further reading, any browser that Slate supports has `DOMParser` available. So we could use `DOMParser` or `parse5` based on the environment (server or browser), and allow an uglifier to strip out parse5 as unused for users with a typical webpack-type browser build.

However, as far as I know, the only way to really implement that is to use an environment variable. For example, [here](https://github.com/facebook/react/blob/master/packages/react/index.js) is how React does it. We could have a `SLATE_CONTEXT` variable which users could define as `browser` in their webpack config or something like that. Thoughts?

schneidmaster:
Update (sorry for the multiple comments) I decided to go ahead and remove cheerio from the tests as well since it wasn't doing anything irreplaceable. So it's totally removed now and I squashed back down to one commit.

ianstormtaylor:
@schneidmaster thank you! This is a seriously high quality PR, I really appreciate it. Here's a few thoughts...

I think for the `DOMParser` compat, we can actually just make the `Html` serializer's constructor take an `options.DOMParser` argument, that we default to the global (if it exists, otherwise it'll be undefined). That way people can pass in `parse5` if they need to, but not increase the bundle side if it's all browser-side in modern browsers.

And then, I'm actually down to make the change a breaking change, and get rid of the Cheerio-compat layer (although thank you for writing that, that was well considered). That way we can just say that it matches the spec'd `DOMParser` API exactly. Seems like it'll be nicely future-friendly in that case. (And during the Slate beta we've been releasing breaking changes fairly regularly, to not build up too much debt.)

Also thanks for converting all of the tests and removing the rest of Cheerio, and no worries about multiple comments.

Let me know how that sounds!

schneidmaster:
@ianstormtaylor No problem, thanks for the feedback! That all sounds good to me, I can implement those changes. I'm guessing we should then keep `parse5` in `devDependencies` and pass it to `options.DOMParser` for the tests? (Since mocha runs in node and, I assume, doesn't have the native `DOMParser` available)

schneidmaster:
Updated per your recommendations, let me know how that looks. (Also let me know if I missed any documentation updates -- I think I got 'em all but I'm not as familiar with everything.)

schneidmaster:
Rebased against master to ensure my changes are compatible with #915, still looks good to me 👍

ianstormtaylor:
Thanks @schneidmaster that looks great! I'll hopefully merge tomorrow, but might be a slight delay since it's breaking. (But I do like that it's breaking, since we're removing that debt!)

schneidmaster:
No worries! Just let me know if you need anything else from me. (PS: added documentation to `docs/reference/serializers/html.md` for the `domParser` option, and also added documentation there for the `defaultBlockType` option since it seemed to be missing)

ianstormtaylor:
Hey @schneidmaster one last thing. I originally thought we should call the option `options.DOMParser`, and just accept any object that matches the `DOMParser` API exactly, but it looks like Parse5 doesn't actually match it. So instead, I think we should call the option `options.parseHtml` and accept just a plain function that takes `html` and returns the parsed tree. (Exactly like you've done, just changing the option name to `parseHtml` instead.) Thanks!

schneidmaster:
Hey no worries, made that update 👍

ianstormtaylor:
Released as `0.21.0`, thanks again @schneidmaster!

PR #947, opened 2017-07-21
Omit version on gzip size badge

conorcussell:
I noticed that the file size badge didn't change when Cheerio was removed in 0.21.0. Because I stupidly pinned it to version 0.19.0 😳  - this will now always show the gzip size of the latest release


ianstormtaylor:
Thanks @conorcussell!

PR #949, opened 2017-07-21
true sourcemaps for example dev build

erquhart:
The babelify transform in the `watch:examples` call isn't putting out sourcemaps, so source mapping only goes back to Babel output. This change provides inline sourcemaps in the Babel output so that dev tools can map back to the original source. Console error messages still reference the Babel output.


erquhart:
Breakpoints aren't working for some reason, troubleshooting.

Issue #752, opened 2017-04-24
triple-click to select block, selection extends to next block


AlexeiAndreev:
http://slatejs.org/#/rich-text?
Triple click to select the first paragraph. Click header 1 button. Note that the second paragraph became H1 as well.

ianstormtaylor:
Hey @AlexeiAndreev thanks for reporting this.

I agree it's something that should be fixed. It looks like [Draft handles this](https://github.com/facebook/draft-js/blob/213cd764f61cc64552bddd672ae1748529d55333/src/model/modifier/RichTextEditorUtil.js#L262-L265) by fixing these types of selections before using `setBlock`. If anyone is interested in helping, I'd be open to a pull request that does similar!

ianstormtaylor:
See https://github.com/ianstormtaylor/slate/pull/863#issuecomment-307659171

pauliusef:
Could we re-open this issue, as it's still happening with applying list styles (ordered, unordered) in the same rich text editor example.

http://slatejs.org/#/rich-text
Triple click to select the first paragraph. Click ordered/unordered list button. See, that the previous block gets added inside the list element.

prichodko:
@ianstormtaylor I can work on this if @ianstormtaylor could point in the right direction.

jaredtmartin:
Like prichodko suggests, this is still an issue.  When you triple-click, the selection in slate is only on the word, not the whole leaf.  My theory is that if you then edit, slate's state gets out o sync with reality causing it to crash. I'm gonna see if I can find a way to patch this.

rohankeskar19:
Is there a fix for this now?

smnbbrv:
For the affected ones, here is quick-and-dirty solution.

On every selection change, run the following:

```
const selection = this.editor.selection;

if (selection && selection.anchor.path[0] === selection.focus.path[0] - 1 && selection.focus.path[1] === 0) {
  Transforms.move(this.editor, { distance: 1, unit: 'character', edge: 'focus', reverse: true });
}
```

This actually checks if the end of the selection lands on a start of a new line while the selection itself originates at the previous line and moves the cursor 1 symbol back.

Of course I would appreciate a real fix for this instead

norahsana:
For the affected ones, the above solution did not work for me but this did:

```
const { anchor, focus } = editor.selection

  if (
    anchor.path[1] !== undefined &&
    focus.path[1] !== undefined &&
    anchor.path[1] < focus.path[1] &&
[...]

PR #950, opened 2017-07-21
Fix triple select bug

sethmcleod:
This relates to issue #752 and is still a work in progress. Just wanted to go ahead and open a PR as I debug this issue to keep discussion in one place.

The idea is to update the selection conditionally if it has been _triple selected_ (similar to how Draft.js [does it](https://github.com/facebook/draft-js/blob/213cd764f61cc64552bddd672ae1748529d55333/src/model/modifier/RichTextEditorUtil.js#L262-L265)) in order to avoid the next block from being selected.

Here is the info I've found so far:

image

It seems when the next block is selected, it becomes the focus. I wonder what the best way to update the focus/selection is...

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -175,16 +175,6 @@ class Content extends React.Component {
     let anchorOff
     let focusOff
 
-    console.log('=============')
-    if (focusKey !== anchorKey && anchorOffset === 0 && focusOffset === 0) {
-      console.log('next block selected', selection)
-    } else {
-      console.log('selection', selection)
-    }
-
-    console.log('anchorRanges', anchorRanges)
-    console.log('focusRanges', focusRanges)
-
     anchorRanges.forEach((range, i, ranges) => {
       const { length } = range.text
       a += length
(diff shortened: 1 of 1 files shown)


sethmcleod:
@ianstormtaylor are you aware of any other cases where `focusKey !== anchorKey && anchorOffset === 0 && focusOffset === 0` would return true? Also, would you mind adding an _in progress_ label to this?

sethmcleod:
I've added some code to **moveToRange** of the **startBlock**, and that seems to work pretty well. The only problem I can find is that if the user _undos_ their changes afterward, the block contents (anchorRanges) get reversed. Here's what it looks like when I triple-click to select the first row (then the selection is adjusted) then I click backspace, then CMD+Z

image

ianstormtaylor:
Hey @sethmcleod thanks for this! Looking at this, and how Draft.js does it, I'm remebering a previous conversation in https://github.com/ianstormtaylor/slate/pull/863#issuecomment-307659171 where I realized that I don't think this is the best way for us to solve it.

Instead, since it seems like the thing that really trips people up is "which blocks are affected when I select then do `x`..." it seems like we might just want to edit the `getBlocksAtRange()` methods logic to omit blocks that happen to have the trailing/leading hanging selection in them. Does that make sense? I'd love a PR that implements that. I think it should solve the same issue, but in a way that has less problems going forward. Thanks!

sethmcleod:
Ohh that makes sense @ianstormtaylor, thanks for the reply. I'm going to close this PR and I'll work on updating the `getBlocksAtRange()` logic in another PR.

PR #952, opened 2017-07-21
Reckon with inconsistencies between parse5 and native DOMParser

### LINKED ISSUE #948: Problems with the new Html deserializer

A bug.

The new Html deserializer doesn't work with the default `parseHtml` function. You can see the errors trying to paste some html in the [paste html example](http://slatejs.org/#/paste-html).

First you have this error:
`Uncaught TypeError: _this.parseHtml is not a function`

That could be resolved by passing explicitily `null` to `Html` constructor like this, but I think it should be fixed to also use the default parser if the options is `undefined`:

```js
const serializer = new Html({
    rules: RULES,
    parseHtml: null
});
```
Then I created this fiddle to show the next error: [https://jsfiddle.net/6u1e543z/](https://jsfiddle.net/6u1e543z/)

`Uncaught TypeError: elements.filter is not a function`

This happen because `elements` is a `NodeList`, not an `Array` so `filter` is not defined there.

I don't know if there are more, couldn't get past this one.

The Html should work as before after updating the deserialize methods.

schneidmaster:
There's a few inconsistencies between parse5's output and the native DOMParser that were causing problems:

* The native DOMParser outputs tag names in CAPS. It's not clear to me if this is a consistent behavior across browsers, so I just updated the HTML example to force to lower case.

* The native DOMParser uses a `NodeList` for `childNodes`. Updated to coerce into an array if that's the case so `filter` exists.

* The native DOMParser will parse error with `application/xml` and something that is not valid HTML. I changed to `text/html` which will always work but does wrap the output in `html` and `body` tags, and then added a line to dig back to the actual output

* The native DOMParser will render a parsing of a plain text string as a `span` instead of a `#text` node

Also changed from `if (options.parseHtml !== null)` to `if (typeof options.parseHtml === 'function')` to account for `parseHtml` being undefined (or some other invalid value)

Fixes #948

### DIFF
--- a/examples/paste-html/index.js
+++ b/examples/paste-html/index.js
@@ -80,7 +80,7 @@ const MARK_TAGS = {
 const RULES = [
   {
     deserialize(el, next) {
-      const block = BLOCK_TAGS[el.tagName]
+      const block = BLOCK_TAGS[el.tagName.toLowerCase()]
       if (!block) return
       return {
         kind: 'block',
@@ -91,7 +91,7 @@ const RULES = [
   },
   {
     deserialize(el, next) {
-      const mark = MARK_TAGS[el.tagName]
+      const mark = MARK_TAGS[el.tagName.toLowerCase()]
       if (!mark) return
       return {
         kind: 'mark',
@@ -103,9 +103,9 @@ const RULES = [
   {
     // Special case for code blocks, which need to grab the nested childNodes.
     deserialize(el, next) {
-      if (el.tagName != 'pre') return
+      if (el.tagName.toLowerCase() != 'pre') return
       const code = el.childNodes[0]
-      const childNodes = code && code.tagName == 'code'
+      const childNodes = code && code.tagName.toLowerCase() == 'code'
         ? code.childNodes
         : el.childNodes
 
@@ -119,7 +119,7 @@ const RULES = [
   {
     // Special case for links, to grab their href.
     deserialize(el, next) {
-      if (el.tagName != 'a') return
+      if (el.tagName.toLowerCase() != 'a') return
       return {
         kind: 'inline',
         type: 'link',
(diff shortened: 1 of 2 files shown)


guicapanema:
This works fine for me!

ianstormtaylor:
Thanks @schneidmaster sorry for the delay.

Question for you, but merging to save time: even if the native DOM parser rendered text strings as spans, wouldn't the current behavior pass through the spans and get to the leaf text nodes anyways? I'm confused as to why we need the extra `<span>`-catching behavior. Thanks!

schneidmaster:
No problem, thanks for merging. The issue is that parse5's leaf node has a `nodeName` of `#text` (and no `tagName`) while DOMParser's leaf node has a `tagName` of `span` (and no `nodeName`), at least as best as I could tell from testing. So that's why we need to handle both. There's a handful of those sorts of discrepancies that I had to clean up (such as `value` vs. `nodeValue`).

I'm also curious to see how the `DOMParser` implementation generally works for folks. I was talking to @erquhart and he mentioned seeing some browser discrepancies in the `DOMParser` implementation. There might be other bugs that crop up, and/or we may end up needing to ship parse5 just to have a stable/consistent implementation (if this ends up being the case, we could also consider moving the serializer to a package to keep the core lean). I don't think we need to take that step yet but something to keep in mind.

ianstormtaylor:
Ah interesting. But wouldn't the `DOMParser`'s `span` have a child text node inside it? (eg. `nodeType == 3`?) I don't understand how it would be a `span` without any child nodes yet still contain text content?

And agreed, if the DOMParser is very buggy or not standardized that's very annoying and we'll need to ship `parse5` probably.

schneidmaster:
Interesting. I could've sworn the `span` had no children (because I remember being confused by exactly the same thing when working on this PR) but I just tried deleting the `el.tagName == 'span'` part locally and it still seems to work fine. I probably just goofed, want me to send another PR to remove that?

ianstormtaylor:
@schneidmaster no worries! I know that feeling 😄 another PR to fix it would be awesome, thanks!

PR #954, opened 2017-07-23
Add a Forced Layout Example

YurkaninRyan:
Often on the Slack channel people ask questions that I think could be answered by showcasing how to effectively use Schema Rules.  This example is my best attempt at illustrating a very common use case.

1. Always have a title at the top
2. Only ever have one title
3. Always have at least one paragraph underneath it.


YurkaninRyan:
On the slack channel so far gavinhewitt has made some pretty good suggestions that I'm going to add here for posterity

1. Rename to Fixed Layout Example
2. Add placeholders to make it more obvious empty titles and an empty content section are actually there

hubgit:
This is really helpful, thanks @YurkaninRyan!

If it's not too much trouble, would it be possible to add an example of how to restrict the inline marks allowed in the title block (e.g. allow italic, but not bold)?

ianstormtaylor:
Great idea, thanks @YurkaninRyan!

PR #961, opened 2017-08-01
Remove unneeded check.

schneidmaster:
Even if the `el.tagName` is a span, it will fall through to the child text node so the check introduced in #952 is unneeded.


ianstormtaylor:
Thanks @schneidmaster!

PR #962, opened 2017-08-01
Upgrade react-frame-component.

schneidmaster:
Gets rid of warning in dev about directly accessing `React.PropTypes`

screenshot 2017-07-31 22 57 24


ianstormtaylor:
Thanks @schneidmaster!

Issue #963, opened 2017-08-01
remove isNative?


ianstormtaylor:
I haven't looked into this deeply, but I remember having a conversation where it seemed like the performance benefits we were getting from `isNative` might not be as much as we thought, given that we do some other performance-enhancing things that Draft.js does not (and `isNative` was borrowed from Draft).

So we might be able to get rid of it, which would simplify a lot of other things.

If someone wanted to investigate this with a PR and check the performance after removing it, that would be amazing!

Issue #938, opened 2017-07-17
Press space in the wrapped second line eats the first line


terryoy:
#### Do you want to request a *feature* or report a *bug*?

bug

#### What's the current behavior?

![image](http://g.recordit.co/bl4qrkSQ6b.gif)

#### What's the expected behavior?

Just insert plain space...

hubgit:
I'm able to reproduce this on http://slatejs.org/#/plain-text by entering a long line of text and trying to type a space at the start of the wrapped line.

J22Melody:
Same here, on Chrome it will eat the first line while on Safari it won't get any response

tommyokeefe:
Any eta on a fix for this? We're using slate @ NPR for the editor in our cms and I just ran into this bug when an editor lost a chunk of text in a story.

danburzo:
I can confirm the simplified use-case: just causing the line to wrap and pressing Space at the beginning of the second line will eat the first lien of text in Chrome 60/macOS. Safari/FF seem to be okay.

danburzo:
Okay, found it! The problem is generated by the CSS rule `-webkit-user-modify: read-write-plaintext-only;`. It causes Chrome to wrongly report window.getSelection().anchorNode as being only the _second_ line of text, and the offset = 0.

danburzo:
To mitigate the problem we can prevent the default behavior in Chrome and replace it with a synthetic insertion of the Space character: 

```js
  // this is in plugins/core.js
  function onKeyDownSpace(e, data, state) {
    if (IS_CHROME) {
      e.preventDefault();
      return state.transform().insertText(' ').apply();
    }
  }
```

Seems to get rid of the problem. You can put this into a plugin, or — if it's something we can have as part of the core — directly in the core plugin.

tommyokeefe:
@danburzo I played with that approach to mitigating the problem locally but ended up deciding against it. The extra space characters all get pushed to the end of the previous line (before the break) so a user could theoretically add a bunch of spaces unintentionally and not see them until text is published. Just something to consider.

danburzo:
@tommyokeefe you're right, but logically that's where it's supposed to be, and by patching Chrome all browsers behave in a fairly consistent manner. You'll see the same behavior in Google Docs and various desktop apps, as well.

tommyokeefe:
Ahh. True. Good point.

danburzo:
[...]

PR #964, opened 2017-08-02
Remove isNative

conorcussell:
@ianstormtaylor I've [removed isNative](https://github.com/ianstormtaylor/slate/issues/963)

I guess the library benchmarks don't give us much info on the performance changes here, any tips on checking the performance of this in the browser?

```
benchmarks
    models
      get-blocks
        295.26 → 283.60 iterations/sec
      get-blocks-at-range
        2.73 → 2.80 iterations/sec
      get-characters
        86.72 → 100.63 iterations/sec
      get-characters-at-range
        7.47 → 8.48 iterations/sec
      get-inlines
        294.12 → 294.65 iterations/sec
      get-inlines-at-range
        2.64 → 2.68 iterations/sec
      get-marks
        17.04 → 17.32 iterations/sec
      get-marks-at-range
        5.04 → 4.42 iterations/sec
      get-path
        37.46 → 38.18 iterations/sec
      get-ranges
        1098.85 → 1056.25 iterations/sec
      get-texts
        259.95 → 244.42 iterations/sec
      get-texts-at-range
        95.68 → 97.23 iterations/sec
      update-descendant
        72.57 → 77.09 iterations/sec
    rendering
      normal
        1.23 → 1.25 iterations/sec
    serializers
      raw-deserialize
        5.64 → 5.63 iterations/sec
      raw-serialize
        62.00 → 59.60 iterations/sec
    transforms
      delete-backward
        246.63 → 285.42 iterations/sec
      insert-text
        285.92 → 289.88 iterations/sec
      normalize
        87.54 → 105.78 iterations/sec (21% faster)
      split-block
        69.83 → 83.92 iterations/sec (20% faster)
```

Also this appears to change the nature of at least one bug https://github.com/ianstormtaylor/slate/issues/938, instead of deleting the previous line, it does not allow you to add spaces on a wrapped line. I was working on a solution to this separately but still had some things to iron out.

### DIFF
--- a/src/plugins/core.js
+++ b/src/plugins/core.js
@@ -1,7 +1,6 @@
 
 import Base64 from '../serializers/base-64'
 import Content from '../components/content'
-import Character from '../models/character'
 import Debug from 'debug'
 import getPoint from '../utils/get-point'
 import Placeholder from '../components/placeholder'
@@ -60,8 +59,7 @@ function Plugin(options = {}) {
   }
 
   /**
-   * On before input, see if we can let the browser continue with it's native
-   * input behavior, to avoid a re-render for performance.
+   * On before input.
    *
    * @param {Event} e
    * @param {Object} data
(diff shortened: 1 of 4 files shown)


kgdev:
Why we can remove isNative? Does draftjs do so too?

src/plugins/core.js
```diff
@@ -137,44 +134,9 @@ function Plugin(options = {}) {
     const nextText = next.startText
     const nextChars = nextText.getDecorations(decorators)
```
ianstormtaylor: I think `nextChars`, `nextText` (and probably others) are now unused so we don't need to compute them too? There might be a handful more lines of code we can remove here which will improve performance when checking further.

ianstormtaylor:
@kgdev I'm not sure we can remove it, but it's a hunch I wanted to investigate. One thing we do which Draft doesn't is that we memoize the Immutable.js objects, so many of their methods are very low cost to call for the second time if nothing has changed. I have no idea if it's something we can avoid doing, but it was worth an investigation since `isNative` adds lots of complexity.

conorcussell:
Thanks @ianstormtaylor I removed the rest of the code we don't need from that method. Will try and take a look at the perf impact in devtools and reply with my findings when I get a chance.

ianstormtaylor:
Hey @conorcussell thanks again for kicking this off! I took your changes and added some more on #1088, so I'm going to close this now. Cheers!

PR #966, opened 2017-08-02
Avoid duplicate content with `onChange` handlers

kgdev:
This is according to a similar fix in draftjs: facebook/draft-js#667


kgdev:
#850

ianstormtaylor:
Hey @kgdev I think now that we've removed `isNative` this is no longer needed, so I'm going to close this out. But thanks for putting in the time to try to solve this!

PR #969, opened 2017-08-02
Refactor render arrow functions

hanswilw:
Having debugged a problem while using server rendering for two days I figured that the render functions in slate does not work nicely with [react-prepare](https://github.com/elierotenberg/react-prepare/blob/master/src/utils/isReactCompositeComponent.js#L8)'s component check. The reason is that when using the `render() {}` method it transpiles to
```
  _createClass(Test, [{
    key: "render",
    value: function render() {}
  }]);
```
and is therefore possible to determine the `render` property on the prototype. While `render = () =>` transpiles to `this.render = function () {};` and results in undefined render on the prototype. This makes `_callClassCheck` raise `TypeError("Cannot call a class as a function");`, since `this` is undefined due to no [preparation](https://github.com/elierotenberg/react-prepare/blob/master/src/prepare.js#L51) of the component in react-prepare.

Babel transpile examples are provided [here](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=es2015%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&debug=false&code_lz=MYGwhgzhAEAqCmEAu0DeAoa1PQE7wDsATeXAQV1wHsB3aAXmgAoBKBgPjQF8cd9jSAZSRhiYXEVbd0PdMCoEIVEPAB0IKgHMmAcgrUaALh0AaOIiSqADtSRUkATytr-JcpVos5CpSvVbdYVEicSJjMwRka1t7JxdCNyCxCRYgA)

I've done some research and haven't really found much usage of the render as arrow functions, is this done by intention? My understanding was that arrow functions usually are used for helper methods and event handlers in components for implicit .bind(), but not for the `render` method itself.

One could argue that this is something that should be accepted and improved in react-prepare, but I think that slate should use the most common practice.

### DIFF
--- a/docs/walkthroughs/adding-event-handlers.md
+++ b/docs/walkthroughs/adding-event-handlers.md
@@ -24,7 +24,7 @@ class App extends React.Component {
     this.setState({ state })
   }
 
-  render = () => {
+  render() {
     return (
       <Editor
         state={this.state.state}
@@ -54,7 +54,7 @@ class App extends React.Component {
     console.log(event.which)
   }
 
-  render = () => {
+  render() {
     return (
       <Editor
         state={this.state.state}
@@ -101,7 +101,7 @@ class App extends React.Component {
     return newState
   }
 
-  render = () => {
+  render() {
     return (
       <Editor
         state={this.state.state}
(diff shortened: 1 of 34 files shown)


ianstormtaylor:
Fair enough @hanswilw, thanks! I only did the arrow function version because it was easier to not have to distinguish between when to use arrows and when not. But this does seem more "correct", so I'm happy to change.

PR #974, opened 2017-08-05
When the `isInEditor` parameter is a text node use its parent only if the parent exists

davegw:
Fix for a bug encountered in IE 10/11 where `isInEditor` can throw due to `el` not having an `isContentEditable` property (found while using the `Raw` serializer).

The issue appears to be that `target` can have a `nodeType` of 3 and no `parentElement`, which sets `el` to `undefined`. Likely related to https://github.com/ianstormtaylor/slate/pull/857.

To reproduce in IE 10/11: Click into the editor, make no edits or delete some text, then click out of the editor.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -251,6 +251,8 @@ class Content extends React.Component {
     const { element } = this
     // COMPAT: Text nodes don't have `isContentEditable` property. So, when
     // `target` is a text node use its parent element for check.
+    // COMPAT: `parentElement` is not defined on text nodes in certain browsers:
+    // https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement#Browser_compatibility
     const el = (target.nodeType === 3 && target.parentElement) ? target.parentElement : target
     return (
       (el.isContentEditable) &&
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -251,7 +251,7 @@ class Content extends React.Component {
     const { element } = this
     // COMPAT: Text nodes don't have `isContentEditable` property. So, when
     // `target` is a text node use its parent element for check.
-    const el = target.nodeType === 3 ? target.parentElement : target
+    const el = (target.nodeType === 3 && target.parentElement) ? target.parentElement : target
     return (
       (el.isContentEditable) &&
       (el === element || findClosestNode(el, '[data-slate-editor]') === element)


ianstormtaylor:
Hey @davegw when would there be an event fired on a text node that doesn't have a parent node? (aka wasn't in the current DOM?) I might be misunderstanding.

davegw:
Thanks for the quick response @ianstormtaylor. 

I'm new to the slate library and used the walkthrough to quickly get up and running, so I could test against IE since that's a browser I need to support. In doing so I encountered the case described above where a parent node was not defined. I should note that I only could produce the error in IE 10/11, I also tested against Chrome and even Edge and the parent node existed in both browsers. My code is identical to the walkthrough code used [here](https://docs.slatejs.org/walkthroughs/using-plugins.html).

As to your question, it appears this is an IE-specific issue. Unfortunately, I'm not able to perform a more thorough investigation right now and it sounds like IE is not a browser you regularly test against. Happy to open an issue until a more comprehensive investigation can be done if you prefer.

davegw:
Did a little more digging. Looks like in [certain browsers](https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement#Browser_compatibility) `parentElement` is only defined on nodes that are themselves an `Element` and is not defined on text nodes. 

Makes sense as I'm seeing the behavior occur only on text nodes. Let me know if the fix looks reasonable and I can include a comment to document this edge case.

ianstormtaylor:
Ah interesting, okay yeah sounds good. If you could add a comment with `COMPAT:` like the similar ones in the codebase so we can remember it that would be great!

davegw:
Thanks @ianstormtaylor. Added a `COMPAT:` comment.

ianstormtaylor:
Thanks @davegw!

AlbertHilb:
Hi @davegw, @ianstormtaylor.
I think the right way to solve this issue is to use `parentNode` instead of `parentElement`.
At moment, in browsers not supporting `parentElement` on text nodes, as before PR  #857, the code checks the property `isContentEditable` on text nodes but that property is only defined for `HTMLElement` nodes.

ianstormtaylor:
Ah @AlbertHilb thank you for clarifying that, I'd love a PR if someone is down to fix this!

davegw:
Good call @AlbertHilb. My fix was to prevent errors in unsupported browsers and ignore the `isContentEditable` check in such cases. I wasn't aware of the differences between `parentNode` and `parentElement`. After reading up on `parentNode` and testing, it looks like that would provide a complete solution to the issue. I'll go ahead and add a PR with the revision. Thanks for the suggestion!

davegw:
https://github.com/ianstormtaylor/slate/pull/1004

PR #975, opened 2017-08-08
Speed up getting blocks at a range

jacobbloom:
Old implementation was far too slow if a selection included many nodes.
The performance hit occurred when converting each and every text node
into a block.  This is now avoided by using the same logic as
getTextsAtRangeAsArray for getBlocksAtRangeAsArray, but using blocks
directly.

### DIFF
--- a/src/models/node.js
+++ b/src/models/node.js
@@ -254,9 +254,19 @@ const Node = {
    */
 
   getBlocksAtRangeAsArray(range) {
-    return this
-      .getTextsAtRangeAsArray(range)
-      .map(text => this.getClosestBlock(text.key))
+    range = range.normalize(this)
+    const { startKey, endKey } = range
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Aha, brilliant idea @jacobbloom, thank you! Marking this as performance related so that I can hopefully benchmark it tomorrow and merge. Thanks!

ianstormtaylor:
image

Looks like it was a great improvement, thanks @jacobbloom! I've merge the branch with `master`, although via a potentially not totally correct way, so GitHub isn't auto-merging this PR. But if you check out the commits you'll see yours in there, so I'm going to close this now, thanks!

jacobbloom:
Thanks for the quick response.  Glad to see those results were as good as
they were!

PR #978, opened 2017-08-12
Custom wrapper component

craigblagg:
If attempting to mount a SlateJS instance into markup on a pre-rendered page - to allow for inline editing of that content, the use of a `<div>` element as the container for the SlateJS instance - causes a DOM violation. An error thrown in React dev builds.

To alleviate this, this PR looks to allow for passing any component, to act as the wrapper element for the instance - while defaulting to the current behavior of a `<div>` wrapper.


ianstormtaylor:
Hey @craigblagg, very interesting! Does that mean that all of your nodes inside the editor are also all inlines too? How do you handle blocks nodes?

Just curious, I'm definitely open to this, thanks for submitting the PR!

craigblagg:
Hey, 

> Does that mean that all of your nodes inside the editor are also all inlines too

Yeah, exactly that. We have a HOC around the Slate instance, and our edit bar (which displays the editing options for the user). And pass a prop of whether it is being rendered into an inline element.

If so, we then show a smaller subset of options to the user in that toolbar - as they will only be able to italicize, bold, underline, anchors etc (and potentially other meta like colors).

ianstormtaylor:
@craigblagg nice that sounds like a great use case! I'm super down, can you just make a few changes:

- Can you change the `props.container` option to be called `props.tagName`? Just to match the `className` a bit, and I think other React libs (although they're all over the place).

- And then sadly, could you remove the extra example? I seriously love and appreciated that you added an example (really!), but I think since it's such a small use case we can omit it to save on maintenance costs over time.

Other than that I'm good to merge. Thank you!

davidosomething:
I think there is one more `<div>` created that also needs to be replaced with a generic or a `<span>` here: https://github.com/craigblagg/slate-1/blob/e386f1f2802121b572a29a7831fab319c5036ff8/src/plugins/core.js#L972-L994

E.g. render with a span customContainer into a `<p>` and you'll get `<p><span data-slate-editor><div style="positon:relative;
>`

craigblagg:
@ianstormtaylor - all sounds great. Will update in the next day or so.

@davidosomething - I'm not at a machine right now. But *think* the lines you point to are just defaults - and can already be changed and handled with a custom schema in your editor impl. Nevertheless I will check and confirm - when making the above changes.

craigblagg:
@ianstormtaylor I've just made the changes to the PR (removing the example, and changing prop name `container -> tagName`. 

`tagName` - felt to me like it is specifying a string, rather than a component e.g. `<Editor tagName={'span'} />` So updated the propType to a `string` to reflect this (I previously had it as `any`). 

As a string it seems easier to reason about accepted types, while a user could then still wrap Slate in a HOC, if they wanted it within a component. 

Let me know your thoughts.

craigblagg:
@davidosomething as thought, you can already configure what you pointed out by providing rules in your schema - as per the docs at: https://docs.slatejs.org/reference/models/schema.html#rules

In your impl. providing a schema something like:

```
const schema = {
  nodes: { ... },
  marks: { ... },
  rules: [
    {
      match: (node) => {
        return node.kind == 'block'
      },
      render: (props) => {
        return (
          <span {...props.attributes} style={{ position: 'relative', display: 'block' }}>
            {props.children}
          </span>
        )
      }
    }
  ]
}
```

ianstormtaylor:
@craigblagg nice that sounds good to me! Thank you!

PR #979, opened 2017-08-12
Improve get-text behavior for nested blocks

oyeanuj:
Hi @ianstormtaylor, a small PR that improves on the behavior in the case of nested blocks. Currently, the blocks get concatenated joining the last word from the previous block with the first word from the next block.

This PR follows plain serializer tradition behavior by adding a newline when concatenating nested blocks.


src/models/node.js
```diff
@@ -1179,7 +1179,7 @@ const Node = {
 
   getText() {
     return this.nodes.reduce((string, node) => {
-      return string + node.text
+      return node.kind == 'block' ? `${string}\n${node.text}` : string + node.text
```
YurkaninRyan: Even if you have only one block, it will always start with a link break.  I don't think we want that right?

oyeanuj: Just about to push in a fix for this (also, why the tests are failing) :)

YurkaninRyan: Nice :D good idea btw i'm all for this pr

oyeanuj:
So, the other two errors are stemming from a mismatch between `block.text.length` and `getOffset` for a block. It shows up on `deleteAtRange` transform when the [`offset` is created](https://github.com/ianstormtaylor/slate/blob/master/src/transforms/at-range.js#L87), and (then via `splitNodeByKey` method) fails at [this line of code](https://github.com/ianstormtaylor/slate/blob/master/src/models/node.js#L1196). Earlier both would be the same here but now `this.length` accounts for the newline and `offset` doesn't.

@ianstormtaylor Question for you since you know this code the best, should `getOffset` be modified [here](https://github.com/ianstormtaylor/slate/blob/master/src/models/node.js#L1045) as well to account for the newline change in this PR above?

oyeanuj:
I am now thinking that a less messy option might be to modify [the plain serializer code here](https://github.com/ianstormtaylor/slate/blob/master/src/serializers/plain.js#L50) to compute text within a block with newlines instead of modifying `getText` (like above). So, I'd probably close this PR and issue a new one with changes in the plain serializer.

The only downside seems to be maybe performance impact(?). Any thoughts @YurkaninRyan @ianstormtaylor?

ianstormtaylor:
@oyeanuj I think your intuition is right about making this change in the `Plain` serializer instead of in the `node.text`, because that text is more for internal offset calculation type things. (Although I do understand that it's a bit confusing in the `document.text` case.)

oyeanuj:
Great, then I'll close this PR and create a new one with that change!

Issue #976, opened 2017-08-09
Update HTML Example in Docs


YurkaninRyan:
Currently, the tutorial makes assumptions that are no longer true.  (You don't get an object anymore for example, you can get an array or a raw dom node while deserializing.)

We should update it to match the switch.

towfiqi:
Just followed the guide and implemented it in my project. But the issue is, the editor doesnt render the `<p>` tags. It saves it correctly with html. But when rendering inside the editor, it strips all the html and shows it inside in a span. like this:

`<div data-key="12" style="position: relative;"><span data-key="10"><span data-offset-key="10-0">Firebase queries allow you to order your data by any child key on the fly. However, if you know in advance what your indexes will be, you can define them via the .indexOn rule in your Security Rules for better performance. </span></span></div>`

Instead of:

`<div data-key="12" style="position: relative;"><span data-key="10"><span data-offset-key="10-0"><p>Firebase queries allow you to order your data by any child key on the fly.</p><p> However, if you know in advance what your indexes will be, you can define them via the .indexOn rule in your Security Rules for better performance.</p></span></span></div>`

The schema, marks, rules are exactly as in the documentation:
https://docs.slatejs.org/walkthroughs/saving-and-loading-html-content.html

YurkaninRyan:
@towfiqi el.tagName on a native dom element isn't going to be "p" anymore instead it will be "P", that's why it isn't rendering a paragraph.

towfiqi:
Worked! Thanks. Looks like the documentation needs updating.

schneidmaster:
@YurkaninRyan I think this is fixed via your #985 and my #1001 -- are there any other HTML documentation issues you're aware of?

YurkaninRyan:
This looks good to me!

PR #985, opened 2017-08-14
Updating HTML Walkthrough Documentation

YurkaninRyan:
With the changes to how we handle HTML Parsing, you no longer get a lower case `tagName`.  This updates the documentation to account for that change.   This should reduce some confusion.

### DIFF
--- a/docs/walkthroughs/saving-and-loading-html-content.md
+++ b/docs/walkthroughs/saving-and-loading-html-content.md
@@ -54,7 +54,7 @@ const rules = [
   // Add our first rule with a deserializing function.
   {
     deserialize(el, next) {
-      if (el.tagName == 'p') {
+      if (el.tagName == 'P') {
         return {
           kind: 'block',
           type: 'paragraph',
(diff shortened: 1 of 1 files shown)


YurkaninRyan:
https://github.com/ianstormtaylor/slate/issues/976

YurkaninRyan:
@ianstormtaylor Something that confuses me, should we be recommending that everyone passes in parse5 as their html parser?  All of our tests seem to use it, and it also returns lowercase tag names

echenley:
I believe this was dealt with in https://github.com/ianstormtaylor/slate/pull/952? It was released as 0.21.1. Perhaps you are still on 0.21.0?

YurkaninRyan:
@echenley From what I can tell that was only updated in the `paste-html` example.  while this touches the actual html walkthrough in the docs (which from what I can tell don't use `.toLowerCase`

But now i'm wondering if that's what we should be suggesting... to always force lower case.  It makes sense to me.

ianstormtaylor:
Thanks @YurkaninRyan!

Issue #984, opened 2017-08-14
HTML deserializeMark function no longer allows for Data on mark creation


YurkaninRyan:
Assume when you are deserializing HTML you hit a span, that you want to represent text color.  on this span, it has a `style` attribute, that you want to parse to get a base color in your data.

Previously you could return this from the deserializer
```js
  return {
    type:  "text-color",
    kind: "mark",
    data: {
      color: getColorFromElement(el),
    },
  }
```

and then in your schema access that in your render as `props.mark.data.get("color").  However do to this line change

https://github.com/ianstormtaylor/slate/commit/4bbf7487ea827bffa5a253fd0649190944dfb4ee#diff-223f3c7786c59981c20d377fdf6ff99fL240

Was this name change intentional?  What is the `value` of a `mark` now?

YurkaninRyan:
@schneidmaster I figure you would probably be able to add some insight to this, is this more so a documentation issue that I can contribute to?  Can we switch it back to data?

schneidmaster:
Hmmm. I had thought this was necessary when switching from cheerio to parse5, a lot of the node syntax changed because they're now DOM elements instead of cheerio objects. However your PR seems to work fine for me and the tests pass so I must have been mistaken 👍

PR #986, opened 2017-08-14
Allow Data to be passed to Marks when using HTML Deserializer

YurkaninRyan:
This pull requests reverts a previous change that I couldn't find any information on.  I'm not sure what `value` means to a mark, and I couldn't find anything relevant in `models/mark.js`.

I wrote a test to make sure in the future we don't break this functionality.  In master this test is failing but in my branch this test passes.

### DIFF
--- a/src/serializers/html.js
+++ b/src/serializers/html.js
@@ -246,7 +246,7 @@ class Html {
    */
 
   deserializeMark = (mark) => {
-    const { type, value } = mark
+    const { type, data } = mark
 
     const applyMark = (node) => {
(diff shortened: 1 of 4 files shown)


YurkaninRyan:
https://github.com/ianstormtaylor/slate/issues/984

ianstormtaylor:
Thanks @YurkaninRyan this is great!

abobwhite:
I thought I was going crazy because my data was always `undefined`. Glad this was fixed just before I needed it!

Issue #873, opened 2017-06-08
Maybe toggleMark shouldn't expect perfect equality?


alexsegura:
#### Do you want to request a *feature* or report a *bug*?

Not sure this is a bug

#### What's the current behavior?

Currently, when you use `toggleMark`, [you need to pass the _exact_ same object](https://github.com/ianstormtaylor/slate/blob/94e0e6254ad6d05f9d345c1d51bb56df5c5a20b9/src/transforms/at-current-range.js#L362) to make it work.

For example:

```javascript
// Working
transform.toggleMark({ type: 'bold', data: { foo: 'bar' } })
transform.toggleMark({ type: 'bold', data: { foo: 'bar' } })

// Not working
transform.toggleMark({ type: 'bold', data: { foo: 'bar' } })
transform.toggleMark('bold')
```

#### What's the expected behavior?

Maybe when a String is passed, we should remove all the marks of that type, no matter the other properties?
Just asking, not sure if this could have side effects. 
If you agree, I think I can make a pull request.

YurkaninRyan:
@ianstormtaylor @alexsegura So I recently encountered this problem when trying to apply a "text-color" mark.  I think it makes more sense that remove/setting a mark works purely on for that specific use case, and any use case where you can only have one of a type of mark, but it can have multiple different values in it's data.

If we made this change it would certainly break anyones marks that can be applied more then once, with different data.  I personally don't know if that's currently possible, or what use case would require that

ianstormtaylor:
I think there's an issue in that marks are compared based on the combination of their `type` and `data`. So if you apply two different marks to a span of text that have the same `type` but different `data`, then you would not be able to specify which to remove.

For example:

```js
{ type: 'comment', { data: { id: 1 }}}
{ type: 'comment', { data: { id: 2 }}}
```

I'm not sure how to nicely resolve this issue. I agree that depending on the use case a decision in either direction would be annoying. But since the marks have to be part of the operations, at some point it needs to be decided how they are compared. It feels like the current solution allows the most flexibility.

You can always retrieve the marks in the range, and then iterate through them, removing any mark that has the `type` you want to remove.

YurkaninRyan:
Would it make sense to remove all of the type if they just supply the type to `removeMark`, but if they supply data then it must be completely equal?

I'm not sure if that really captures the essence of what `removeMark` should do because then it's possible to remove more then 1 mark at a time...

ianstormtaylor:
[...]

PR #990, opened 2017-08-15
Introduces state.activeMarks and has add/remove/toggleMark use it instead of state.marks

### LINKED ISSUE #988: toggleMark should behave as per RTE conventions

A bug regarding the behavior of `toggleMark`.

Currently, `toggleMark` removes the mark if it already exists in some part of the selection, and adds it otherwise.

The convention (followed by most if not all text editing apps, and on the web by `document.execCommand`) is to:
* remove the mark if _the entire selection_ contains it
* add the mark otherwise

danburzo:
Fixes #988


danburzo:
## Tests to add

* [ ] toggleMark collapsed selection at start 0
* [ ] toggleMark collapsed selection at start N
* [ ] toggleMark non-collapsed selection causes addMark in mixed scenario
* [ ] toggleMark non-collapsed selection causes addMark in non-marked scenario
* [ ] toggleMark non-collapsed selection causes removeMark in all-marked scenario

src/models/node.js
```diff
@@ -900,21 +901,25 @@ const Node = {
       const previous = this.getPreviousText(startKey)
       if (!previous || !previous.length) return []
       const char = previous.characters.get(previous.length - 1)
-      return char.marks.toArray()
+      return options.perCharacter ? [char.marks.toArray()] : char.marks.toArray()
```
YurkaninRyan: Would this warrant it's own subset of functions?  `getMarksPerCharacter` and `getMarksPerCharacterAsArray`?

YurkaninRyan:
Great PR!  This is going to be very useful :D

src/models/node.js
danburzo: Unfortunately I don't know enough of the innards to make a call on this yet... but whichever you decide I can conform :)

examples/rich-text/index.js
```diff
@@ -70,7 +70,9 @@ class RichText extends React.Component {
 
   hasMark = (type) => {
     const { state } = this.state
-    return state.marks.some(mark => mark.type == type)
+    const characterMarks = state.marksPerCharacter
+    return characterMarks.length &&
+      characterMarks.every(charMarks => charMarks.some(mark => mark.type == type))
```
ianstormtaylor: I now see what you mean about why `state.marks` isn't enough by itself. But it does look like this kind of computation is going to get expensive, since people will always be using anonymous functions which are hard to memoize.

I wonder if we should instead introduce `state.hasMark(mark)` and `state.hasMarkByType(type)` methods that we can memoize ourselves?

src/models/node.js
ianstormtaylor: Have it as two separate methods sounds good to me.

examples/rich-text/index.js
ianstormtaylor: If we do do that, I'm not sure if it's worth it to have the `marksPerCharacter` anymore, since people can just do `state.characters.every(c => c.marks.some` for their own custom cases, which I think would be just as expensive?

ianstormtaylor:
@danburzo great questions.

Thinking about it more, I think we may be able to come up with a slightly improved abstraction. Considering that both "give me any marks in a range" and "give me the 'currently active' (as specified by rich text editor rules) marks in a range" seem like valid use cases for people...

I think it might be best to instead distinguish between the two and treat them as first class options. So that we'd end up with:

```
state.marks
state.activeMarks
```

Which would have underlying implementations of:

```
node.getMarksAtRange(range)
node.getMarksAtRangeAsArray(range)

node.getActiveMarksAtRange(range)
node.getActiveMarksAtRangeAsArray(range)
```

I think this will help performance in general, since the "active" state will be a lot cheaper to compute. It should just grab the first character and keep intersecting the set of marks it has with following characters. And if at any point the set has no entries, we can abort early. (Which will often be at the very first character, for the shortest possible iteration.)

Let me know how that sounds, and if you're down to make those changes.

---

For context, in the future...

- We could opt to have Slate offer similar "active" helpers for blocks/inlines based on classic rich text editor behavior. I think this might already be useful in my codebase, but I can't quite remember.

- We could switch `state.marks`, `state.blocks`, etc. to be prefixed by `selected`, to get `state.selectedMarks`, etc. This might further help the distinction between "selected marks" and "active marks", and might make blocks/inlines/texts methods clearer too.

danburzo:
Perfect! I'm on it.

danburzo:
Okay, done. Thank you for the feedback, @ianstormtaylor, this version turned out much cleaner, faster, and with better terminology :)

danburzo:
In regards to having `activeBlocks` and `activeInlines`, that would definitely help with toggling buttons in RTE toolbar. I wonder, however, how the _equality_ between blocks/inline would work, and whether `activeBlockTypes` and `activeInlineTypes` (with equality on the `type` attribute) would be a clearer (and still sufficient) abstraction — although to draw a parallel with `activeMarks`, we'd need to also account for the `data` object.

src/models/node.js
```diff
@@ -919,6 +931,37 @@ const Node = {
...
+    if (range.isCollapsed) {
+      const text = this.getDescendant(startKey)
+      const char = text.characters.get(range.startOffset - 1)
+      return char.marks.toArray()
+    }
+
+    // Otherwise, get a set of the marks for each character in the range.
+    const chars = this.getCharactersAtRange(range)
+    const first = chars.first()
+    if (!first) return []
```
ianstormtaylor: I think if since we already return early for `isCollapsed` above, then this `if (!first)` will never fire?

src/models/node.js
```diff
@@ -919,6 +931,37 @@ const Node = {
...
+    let memo = first.marks
+    chars.slice(1).forEach((char) => {
```
ianstormtaylor: I think we might want an extra early return here where if the first character has no marks (`marks.size == 0`), we can `return []` right away, since that seems like it will be a very common case.

Ideally we'd even actually be able return early out of the `forEach` loop whenever the `memo.size == 0`, which would take the shortest looping for all cases.

src/models/node.js
danburzo: True, I just added it here as a safety precaution

src/models/node.js
danburzo: Note taken on first point, on second point it actually returns early based on `memo.size`. Excuse the brevity, I'm ony phone :)

ianstormtaylor: Ah shoot I missed that. Could you make it `return memo.size != 0` for clarity? 

In that case ignore my first thing about checking the first character explicitly, handling it all inside the iterator like you've done is nicer.

src/models/node.js
ianstormtaylor: Gotcha, can you take it out for now? I'd like to just keep it terse, and if it becomes an issue we can solve the underlying problem. Thanks!

danburzo:
Morning! I've cleaned up `getActiveMarksByRange` and made `addMark`/`removeMark` use it instead.

ianstormtaylor:
@danburzo thank you!

I'm having an internal debate right now...

As I just commented in the `activeInlines/activeBlocks` issue, I think your conclusion makes sense that it won't nicely handle the common use cases, and will end up just having to be circumvented by people who want that behavior.

Given that conclusion, the `activeMarks` vs. `marks` distinction isn't ever going to have a nice parallel structure, and will continue to just be a thing you have to look up in the docs to understand what it means. Which makes me feel slightly weird about it. So... bare with me while I present two potential other solutions for us...

#### Solution B

Instead of relying on `activeMarks`, which adds a slight amount of opinion (slight because the opinion is pretty standardized, so it's not horrible), we could have people do the more low-level option, which is still concise:

```js
state.characters.every(char => char.marks.some(mark => mark.type == 'bold'))
```

This is theoretically slightly less performant, since the looping of the characters won't be memoized. But my guess is that 99% of cases will exit early on the first character anyways, since marks are sparse. And for the cases that end up not exiting on the first, that will be only for one button in the toolbar out of `n`, and the chance that multiple buttons don't exit early will be _incredibly_ small.

So in other words for performance, if I'm understanding correctly, I think we could make an argument that this kind of optimization isn't worth putting the extra opinion/code into core.

#### Solution C

Instead of adding an additional `activeMarks` property, we could change the behavior of the current `marks` to actually be `activeMarks`. There is precedent for this already because both `blocks` and `inlines` have a bit of opinion baked into them, based on RTE conventions—`blocks` happens to return the leaf blocks, and `inlines` happens to return the root inlines, because that's what is most commonly used by RTE editors.

Similarly, we can have `marks` just be the active intersection of marks, which is what RTE editors often are referring to. This would be a breaking change, but we're fine making breaking changes while in beta, and make small ones like this often. (In fact, if people don't do anything to "fix" the change, their editors will just inherit the proper RTE semantics from this change I think.)

I can't think of a time when I've actually wanted to get the non-intersected marks while building with Slate. And even in the rare cases that people want that, they can loop the characters and create the union themselves.

---

What do you think? I'm leaning towards **C**, but open to **B** too.

I realize this is invalidating some of your time spent on this pull request, but hopefully it gets us to a nicer solution in the end. Thanks for taking the time to discuss!

danburzo:
I'm all about the nicest solution here, and having spent some time with the codebase gives me a better grasp of how it's architected, regardless of the outcome of this pull request — Slate is the most promising of RTE frameworks out there, and I want to help with anything that will get it closer to production-quality. (Not that I'm entirely selfless with this — the grand scheme is to replace Scribe in the Moqups codebase :P) 

Let me add some further thoughts on the matter:

I think `marks` is already atypical in relation with `blocks` and `inlines` in at least two aspects: on the one hand it accrues values from `selection.marks`, and on the other hand it looks outside the range in case the selection is collapsed.  So, right off the bat we can conclude that `state.characters` will be insufficient to infer the equivalent of `activeMarks`. Additionally, we need it internally in a few transforms (where activeMarks is currently referenced).

As for having `marks` always work as `activeMarks` (intersection of Marks), that would be perfectly fine for simple marks, without any attached data. However, I was playing earlier with a Mark that had data attached, and of course I have been hit by what seems [to be a rite of passage] :-) figuring out how to toggle it correctly. 

The concrete example is a `font color` mark, which will have something akin to:

```js
{
  type: 'color',
  data: { color: '#ccc' }
}
```

(P.S. I believe that having dynamic Marks is something that sets Slate apart from Draft, where you need to awkward hacks, such as serializing the color's value inside the mark's type to circumvent its lack of metadata)

So, assuming we want to change the color of the selection, with the current API, we'd need to:

1. know which chars are marked with the `color` Mark to remove the previous color
2. set the new color

(these are just assumptions, haven't had the chance to explore how to make best use of the APIs)

To that end, the intersection of Marks (`activeMarks`) would return a null set, and I would not know that there exist marks with the same type but different `data` objects. Having `marks` (the set of unique type/data combination) allows me to iterate and remove as necessary.

So I guess in the end, the main concern is finding a set of APIs (`addMark`, `removeMark`, `toggleMark`) that can solve the common, RTE-convention-compliant, scenario for both _simple marks_ and _marks with data attached to them_....

Hope this makes sense!

danburzo:
(I'll also get back to a previous idea of somehow augmenting `marks` with a qualifier — a sort of Solution C2 — but I can't figure out how that might be accomplished in, ideally, a non-breaking manner.)

ianstormtaylor:
Damn, yes that color mark use case is a good counter example.

---

One potential variable to consider: we could always choose to make `removeMark` flexible, and calling `removeMark('color')` could actually be data-inspecific. (As @YurkaninRyan suggested in #873.) At the comparison level, and operation level we're always going to be data-specific. But the `removeMark` transform could be smart enough to be data-inspecific, and do the data-specific lookup work itself. 

Just a variable to consider in thinking about the ideal API. Because `removeMark('color')` does feel very intuitive to remove all marks of a given color.

The question of whether that asymmetry has too many confusing negative impacts on `toggleMark` is another problem.

Also, even with this change to `removeMark`, it would still necessitate having the non-intersected set of marks available to do the proper removals. But... I think you could make a case that for that one-off case, it could be fine to compute it manually by union'ing each character's marks?

---

Also, while we think about this, definitely don't limit yourself to non-breaking changes. I don't want us to be sacrificing the optimal API for existing implementations right now while Slate is in beta, seeing as it's already a complex enough problem if we get the API perfect 😄

danburzo:
I went further down this rabbit hole, and one important aspect is, I think, to decide on:

#### Can a Character have two Marks of the same `type` but with different data?

Intuitively, it makes no sense for me to allow multiple instances of the same mark type. Question is, do we enforce this, and if yes, at what level. 

* This restriction may be too opinionated and I'm missing a potential scenario that might be hindered by it — no enforcement
* Enforcing it at the `character.marks` Set — which currently does identity on type + data, so we can blindly push into it; checking the existence of a `type` in the set is both inelegant and computationally more expensive
* Enforcing it at the core schema level — haven't looked closely at how schemas work internally, but I guess you could normalize `character.marks` here
* Enforcing it at the userland schema level, as needed

Having some sort of enforcing of this rule then allows us to `transform.toggleMark` without handling all sorts of scenarios, by simply:

1. deciding whether this is an `add` or `remove`-type operation
2. blindly adding or removing the mark by calling `addMark` / `removeMark`

__`addMark(mark: Mark | Object | String)`__
* if a character already contains a Mark of the same type, update the mark with the new properties
* if a character does not contain a Mark for the same type, add the mark to `character.marks`

__`removeMark(mark: Mark | Object | String)`__ — similar to addMark, is there any scenario where you wouldn't want to remove any Mark that matches the mark's type? is there a specific need to target specific type + data combinations?

(Updating this as I think it through)

danburzo:
To summarize my rambling:

__Both `state.marks` and `state.activeMarks` can be derived from the State object.__ Whether this is provided by core Slate (either as methods on the State object, or as utilities) remains an open question.

__`toggleMarks` can be ignored.__ Not necessarily remove it from the API, but you can achieve whichever variation of toggling is best for you by using `state.marks` and `addMark`/`removeMark` in their current incarnation.

__Making `removeMarks` more flexible will not get rid of the need for `state.marks`_, as it will simply not cover all the plausible needs.

danburzo:
(I'm also starting to feel that `activeMarks` as currently implemented still makes too many assumptions :D — that demanding intersection on strict equality might be too restrictive for it to be generic)

ianstormtaylor:
@danburzo thank you for all your ideas here!

I think basically there's still a doubt in our minds whether it is "correct" to have `activeMarks` be part of core, but seeing as it is useful to us now, and "properly" implementing RTE's relies on it heavily, I think it's okay for us to merge this in now and use the wait and see approach, because we may never have a clear answer.

So we can use `activeMarks`, and if it ends up feeling weird, we can consider removing it at a later point and people can go back to the slightly more verbose way.

But we definitely shouldn't remove the "unioned marks" being exposed, since as you point out it's useful for more complex workflows.

Sound good?

danburzo:
Sounds excellent! I agree on both points.

ianstormtaylor:
Okay awesome. I also just realized, this is ever-so-slightly a breaking change, in that `toggleMark` and selection marks will now use the `activeMarks`. So I'm going to merge this, but I want to wait until I can release the existing master first so I can bump this one.

danburzo:
Cool, if I have any other ideas in the meantime I'll report back here.

Issue #817, opened 2017-05-16
Pasting any string containing " and no spaces throws an error


darkyshiny:
#### Do you want to request a *feature* or report a *bug*?
Bug

#### What's the current behavior?
Pasting " anywhere in slate throws the following error.
![pasteerror](https://cloud.githubusercontent.com/assets/16773900/26141303/7abad5b4-3adb-11e7-897c-20edc8fde665.JPG)
Here is a very basic jsfiddle where the error can be reproduced: https://jsfiddle.net/4wzp6hxa/3/

Tested on Windows in Firefox with the latest version of Slate

#### What's the expected behavior?
" should be pasted into the editor.

darkyshiny:
Just noticed that if the string contains at least one space the error is not thrown. Ex. 'asd asd"'

malloryerik:
Couldn't reproduce the bug on Mac in Chrome, Safari, or Firefox.
" pasted as expected.

darkyshiny:
Looked at it some more today and can only reproduce in Firefox think the error I had in Chrome was entirely different (Sorry yesterday was a looong day). Updated the original comment with a new error stack

danburzo:
Clarification: this seems to be the case in Firefox when copying a `"` character from within Slate, and then trying to paste it. Clipboard reveals:

```
[...]

PR #992, opened 2017-08-15
Fixes fragment matcher regex for pasting Slate fragments

danburzo:
Re: #817


ianstormtaylor:
Ah good catch @danburzo, thank you!

PR #994, opened 2017-08-16
Don't break macOS emoji insert nor composition when fixing #938

danburzo:
Re: https://github.com/ianstormtaylor/slate/issues/938 this makes synthetic Space insertion in Chrome less obtrusive, now allowing insertion of emoji chars and possibly fixing IME (although I don't have one around to test)


ianstormtaylor:
Ah good call @danburzo. Thanks!

Issue #983, opened 2017-08-14
Slate fails in IE11. Objects are not valid as a React child (found: List)


kvolkovich-sc:
Bug in IE11.

Slate fails in IE because it uses `Symbol` data type. 

I have add this to my webpack config to polyfill `Symbol`:

```
config.entries = [
  "core-js/es6/array.js", // because Slate uses Array.prototype.find
  "core-js/es6/symbol.js",
  "core-js/fn/symbol/iterator.js",
  ...
]
```

But still get an error in IE11.

`Objects are not valid as a React child (found: List [ [object Object] ]). If you meant to render a collection of children, use an array instead or wrap the object using createFragment(object) from the React add-ons. Check the render method of Node.`

I found, it can be fixed by modify Slate's **Node** code. We should add `.toJS()` call

```
renderText() {
   ...
   return (
     <span data-key={node.key}>
        {leaves.toJS()}
     </span>
    );
   ...
}
```

Will you approve pull request with this change or easier just make a fork?

Or maybe someone have another idea how to fix it in better way?

kvolkovich-sc:
Have made changes [here](https://github.com/OpusCapita/slate/compare/master...OpusCapita:ie11-compat). `leaves.toJS()` were redundant.

ianstormtaylor:
For the `for of` loops, wouldn't it just be a question of tweaking that Babel config that Slate uses for its exported files? I'm assuming that `for of` is supported everywhere else, so we haven't been transpiling it before now, but I'm open to adding support for that.

kvolkovich-sc:
I found old [discussion](https://github.com/babel/babel/issues/1534) at babel repo.
@kittens (main contributor of babel) claims that:

> This is expected behaviour. It's impossible to replicate for...of without a Symbol and iterator polyfill. Babel ships with a version of core-js that you can use, see the docs for more info.

kvolkovich-sc:
@ianstormtaylor can i make a pull request?

ianstormtaylor:
@kvolkovich-sc that would be great, thank you! For the `find` util, can we use `lodash/find` instead? Thanks!

kvolkovich-sc:
@ianstormtaylor see:
https://github.com/ianstormtaylor/slate/pull/996
https://github.com/ianstormtaylor/slate/pull/997

kvolkovich-sc:
Pull requests merged in master.

danburzo:
[...]

PR #996, opened 2017-08-16
IE11 compat

kvolkovich-sc:
# Fix for this issue: https://github.com/ianstormtaylor/slate/issues/983

## After babel transformations, processed `for ... of` became a code structure which uses ES6 `Symbol` data type

IE11 don't supports `Symbol`s.

**Solution:**

Instead of polyfill `Symbol`s, replace `for ... of` to simple `for(;;)`. 

## IE11 don't supports `Array.prototype.find`

**Solution:**

Instead of polyfill `Array.prototype.find`, use `lodash/find`


ianstormtaylor:
Thanks @kvolkovich-sc!

PR #997, opened 2017-08-16
Add .editorconfig and .gitattributes

kvolkovich-sc:
* [.editorconfig](http://editorconfig.org/) for indent consistency
* .gitattributes for ends of lines consistency (problem is relevant for Microsoft Windows users)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+
+# Change these settings to your own preference
+indent_style = space
+indent_size = 2
+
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Can these be solved with `.eslintrc` instead? We're already using that and I don't super wanna have tons of different ways of doing things.

kvolkovich-sc:
Ends of lines can be solved using by eslint. So `.gitattributes` maybe not necessary.

But indenting - no. When you open a new project, the least you want to go into the settings of the editor and setup `tab / 2 space / 4 space` and different indent for different kinds of files.

What if you have 3-5 opened different project simultaneously?)

ianstormtaylor:
Fair enough, thank you @kvolkovich-sc!

PR #1001, opened 2017-08-17
Consistently use toLowerCase in tag name comparisons.

schneidmaster:
Currently, we use the native `DOMParser` for HTML deserialization in browser environments, while recommending `parse5` for server-side rendering or node environments. The `DOMParser` implements the DOM spec which returns uppercase tag names, while `parse5` implements the more low-level HTML spec and ends up returning lowercase tag names (it's a little more nuanced but that's the TL;DR, see [this issue](https://github.com/inikulin/parse5/issues/210) for the full discussion). Because of this discrepancy, for the time being we should consistently use & recommend case-insensitive comparisons of tag names to prevent potential environment-specific bugs and confusion. (Currently we use `DOMParser`-compatible CAPS tags in some documentation/examples while using `toLowerCase` elsewhere.)

### DIFF
--- a/docs/walkthroughs/saving-and-loading-html-content.md
+++ b/docs/walkthroughs/saving-and-loading-html-content.md
@@ -54,7 +54,7 @@ const rules = [
   // Add our first rule with a deserializing function.
   {
     deserialize(el, next) {
-      if (el.tagName == 'P') {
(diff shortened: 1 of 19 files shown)


ianstormtaylor:
Thanks @schneidmaster, this is great!

PR #1004, opened 2017-08-17
Use parent node to find text node parent

davegw:
Provides a more complete solution to https://github.com/ianstormtaylor/slate/pull/974 by using `parentNode` in place of `parentElement` to find the parent of a text node and perform the `isContentEditable` check.

`parentNode` has full browser compatibility as opposed to `parentElement`, which [is not defined for text nodes](https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement#Browser_compatibility) in IE and Opera.


ianstormtaylor:
Awesome, thank you @davegw and @AlbertHilb!

PR #1006, opened 2017-08-18
Add State.characters to docs

danburzo:
Realized while discussing https://github.com/ianstormtaylor/slate/pull/990 that the docs were missing this.


ianstormtaylor:
Ah good call!

PR #1017, opened 2017-08-21
Add a missing dependency which was causing errors

laurenkt:
Because `src/serializers/html.js` depends on `ReactDOMServer.renderToStaticMarkup`, which is no longer bundled with ReactDOM, ReactDOMServer is needed to avoid errors when the HTML serializer is used.


ianstormtaylor:
Thanks @laurenkt!

Issue #1019, opened 2017-08-21
importing slate failed because `.babelrc` appears in dist npm module


doodlewind:
#### What's the current behavior?
Following error happens when trying to import slate:

```
ERROR in ./~/slate/lib/index.js
Module build failed: Error: Couldn't find preset "es2015" relative to directory "/Users/ewind/code/mtdp/wiki/citadel-web/node_modules/slate"
    at /Users/ewind/code/mtdp/wiki/citadel-web/node_modules/babel-core/lib/transformation/file/options/option-manager.js:293:19
    at Array.map (<anonymous>)
    at OptionManager.resolvePresets (/Users/ewind/code/mtdp/wiki/citadel-web/node_modules/babel-core/lib/transformation/file/options/option-manager.js:275:20)
    at OptionManager.mergePresets (/Users/ewind/code/mtdp/wiki/citadel-web/node_modules/babel-core/lib/transformation/file/options/option-manager.js:264:10)
    at OptionManager.mergeOptions (/Users/ewind/code/mtdp/wiki/citadel-web/node_modules/babel-core/lib/transformation/file/options/option-manager.js:249:14)
    at OptionManager.init (/Users/ewind/code/mtdp/wiki/citadel-web/node_modules/babel-core/lib/transformation/file/options/option-manager.js:368:12)
    at File.initOptions (/Users/ewind/code/mtdp/wiki/citadel-web/node_modules/babel-core/lib/transformation/file/index.js:212:65)
    at new File (/Users/ewind/code/mtdp/wiki/citadel-web/node_modules/babel-core/lib/transformation/file/index.js:135:24)
    at Pipeline.transform (/Users/ewind/code/mtdp/wiki/citadel-web/node_modules/babel-core/lib/transformation/pipeline.js:46:16)
    at transpile (/Users/ewind/code/mtdp/wiki/citadel-web/node_modules/babel-loader/lib/index.js:49:20)
 @ ./client/src/editor/EditorApp.js 10:0-40
 @ ./client/src/editor/index.js
 @ multi (webpack)-dev-server/client?http://0.0.0.0:3000 webpack/hot/dev-server react-hot-loader/patch ./client/src/editor
```

Since `.babelrc` is reserved in slate's dist npm module files, which causes `babel-loader` transpiling slate's dist code with its `es2015` preset by default, but since slate is installed as a dependency without devDependency installed, this behavior results to error above.

User may have to add following webpack config as a workaround: 

``` js
      {
[...]

PR #1020, opened 2017-08-21
Add .babelrc into .npmignore

doodlewind:
See Issue #1019 
This PR solves babel transform issue when user imports slate in webpack project with default `babel-loader` config.


ianstormtaylor:
Thanks @doodlewind!

PR #1023, opened 2017-08-24
[WIP] Sort out cut/copy/paste issues

danburzo:
Re: #734 (Work in progress)

References: 
* https://bugs.chromium.org/p/chromium/issues/detail?id=758485
* https://github.com/w3c/editing/issues/168

Basically I'm swapping the "move into contenteditable div, select it, and let the browser copy the contents" technique with manually setting the clipboard data for `text/plain` and `text/html`.

There's an opportunity here to plugin in a Slate HTMLSerializer and PlainSerializer if the author wants to make the clipboard data conform to a certain schema. (There's a discussion about this, I think?) — DONE ✓

What still needs to be done (any help appreciated):

1. See if anything obvious regresses
2. Assess the browser support for `setData()` and whether we can be sure the data has been set. (Otherwise, browsers silently ignoring `setData()` will not be able to copy anything to the clipboard) — and fall back to the div technique if necessary

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -392,8 +392,6 @@ class Content extends React.Component {
 
     const { state } = this.props
     const data = {}
-    data.type = 'fragment'
-    data.fragment = state.fragment
 
     debug('onCopy', { event, data })
     this.props.onCopy(event, data)
@@ -417,8 +415,6 @@ class Content extends React.Component {
 
     const { state } = this.props
     const data = {}
-    data.type = 'fragment'
-    data.fragment = state.fragment
 
     debug('onCut', { event, data })
     this.props.onCut(event, data)
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -390,7 +390,6 @@ class Content extends React.Component {
       this.tmp.isCopying = false
     })
 
-    const { state } = this.props
     const data = {}
 
     debug('onCopy', { event, data })
@@ -413,7 +412,6 @@ class Content extends React.Component {
       this.tmp.isCopying = false
     })
 
-    const { state } = this.props
     const data = {}
 
     debug('onCut', { event, data })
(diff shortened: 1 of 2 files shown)


danburzo:
Well, can't test with Browserstack since it messes with the clipboard, but I will [take this](https://www.lucidchart.com/techblog/2014/12/02/definitive-guide-copying-pasting-javascript/) as a sign the APIs are there since at least 2014, so that's that :)

We just need to do a regression test and I think we're good to go.

danburzo:
It seems the [chromium issue](https://bugs.chromium.org/p/chromium/issues/detail?id=758485) was marked as WontFix, in which case `setData()` seems the way to go. One concern I have is whether Webkit, when "canonizing" the text/html content, will wrap things like `<td>content</td>` in `<table><tr></tr></table>` and other sorts of normalization prior to placing it in the clipboard — needs to be tested.

ianstormtaylor:
Hey @danburzo I just commented in the original issue. It feels like we're dealing with multiple separate things here in this PR that we should potential detangle from each other. 

1. One is the ability to override the internal pasting with custom serializers. It seems like this is already solvable (although core doesn't give you help with it) by simply adding your own custom `onCut/onCopy` middleware with a plugin? Using `setData` here seems like exactly how those plugins would want to do it.

2. Another is the ability to tweak the `data.fragment` so that plugins can change the internal pasting behavior without having to re-implement the internal pasting logic themselves.

3. Another is to fix the behavior of copying HTML content when it is inside a single node in Chrome/Safari? This one seems like the most in need of fixing, and ideally we could find a way that doesn't involve changing much in core?

For (3) I haven't investigated as much as you, so I'm not as sure where the limitations are, but is there no way for us to fix that as currently architected?

Let me know what you think.

doodlewind:
@danburzo Does `setData` only works when event fired from `onCopy` or `onCut`? I'm working on an editor with copy/cut buttons in toolbar, is there a recommended way to programmatically triggering copy and cut events from button click? I found that `execCommand` also works, but this seems to be too tricky if slate has introduced a way of doing this, any idea in this case? 🙏

ianstormtaylor:

PR #1026, opened 2017-08-25
expose onKeyUp

isubastiCadmus:
This PR exposed onKeyUp to editor so that plugins can use it


ianstormtaylor:
@isubastiCadmus awesome, thank you very much!

Issue #980, opened 2017-08-12
Unable to drop files on void nodes


oyeanuj:
#### Do you want to request a *feature* or report a *bug*?
Bug

#### What's the current behavior?
Currently, when you drop a file on a void node, nothing happens. This is because the code [here](https://github.com/ianstormtaylor/slate/blob/master/src/components/content.js#L492) returns early in the `onDrop` function in `content`. It seems like the `isInEditor` function was designed to return `false` on void nodes or their content.

#### What's the expected behavior?
We should be able to drop a file on the void node, and have the event continue as on a non-void node.

#### Next Steps
I'm not sure though if: 
1. We want to add a function testing if the drop is on a void node to modify this condition OR modify `isInEditor` to NOT return false on void nodes or their content.
2. And if so what is the most reliable way to test if a DOM element is a child of a void node.

oyeanuj:
@ianstormtaylor Realized that the issue above didn't make the questions clearer. So, I've updated the issue above with clearer API questions for you. I could then take a stab at resolving this.

ianstormtaylor:
@oyeanuj good find. I think an easy way to solve this might be to make it so that we have two different methods to use:

- `isInEditor` which is the current one without the editable check.
- `isEditableInEditor` which calls to `isInEditor` and then adds the editable check too.

And change all the current use cases to use the new one, with the `onDrop` one not caring about whether the node is editable.

Are you down to make that change? Thanks!

oyeanuj:
Sounds good, I'll take a shot at it! This would also remove the need for checking for void nodes specifically.

AlbertHilb:
Hi @oyeanuj, @ianstormtaylor.
Let me add my opinion.
I think the principle void nodes are based on, is their content should be outside the control of Slate.
In particular they should by themselves implement user interaction logic.
According to that, in this case, maybe, a better solution is to add `onDrop` handler on the component used to render the void node.

tobiasandersen:
I think @AlbertHilb has a very good point. We've actually given the control of `onDrop` to our void nodes without even thinking about it. Until now, I haven't even thought about whose responsibility that should be. It has just felt natural to let the void nodes control its content.

Not saying this is always the case. Other people might use, or view, it differently. Just thought I'd share my thoughts since we're doing quite a lot of dnd inside the editor.

oyeanuj:
[...]

PR #1027, opened 2017-08-25
Separate out isInEditor from isEditableInEditor

oyeanuj:
Hey @ianstormtaylor, [as per your recommendation](https://github.com/ianstormtaylor/slate/issues/980#issuecomment-322546282), here is the fix for #980.

The only deviation from that is that you had suggested replacing current usage of `isInEditor` to `isEditableInEditor` everywhere except `onDrop`. Looking at the code, it seems like the following methods should still check for only `isInEditor` since these seem like valid operations on a `void` node.

1. `onCopy`
2. `onCut`
3. `onDragEnd`
4. `onDragOver`
5. `onDragStart`
6. `onPaste`

 Let me know if you think otherwise.


src/components/content.js
```diff
@@ -249,15 +249,28 @@ class Content extends React.Component {
...
+   *
+   * @param {Element} target
+   * @return {Boolean}
+   */
+
   isInEditor = (target) => {
     const { element } = this
     // COMPAT: Text nodes don't have `isContentEditable` property. So, when
     // `target` is a text node use its parent node for check.
     const el = target.nodeType === 3 ? target.parentNode : target
```
ianstormtaylor: I think we no longer need this line (and comment) in the `isInEditor` function, judging by the comments text since it was only to allow for using `el.isContentEditable`.

oyeanuj: fixed!

oyeanuj:
@ianstormtaylor So, the only issue that is coming up in testing so far is the Embed example. Pasting into an input in the void node breaks unless I update the `onPaste` call to use `isEditableInEditor` as well. Right now, as expected, it pastes it where the selection currently. I can see us making the void node the selection on paste, and pasting right after.. but then that breaks the Embed example.

On the other hand, we can check for `isEditableInEditor`, but that feels a bit wrong since one should be able to paste on a void node? Thoughts?

I'll keep testing a little bit more to ensure no breakage.

ianstormtaylor:

PR #1028, opened 2017-08-25
Fix how plain text serializer serializes nested block

oyeanuj:
Hi @ianstormtaylor, [as per our discussion](https://github.com/ianstormtaylor/slate/pull/979#issuecomment-322542454) in #979, this PR replaces that and fixes the functionality (with a test to go with it).

Let me know if this looks good!


src/serializers/plain.js
```diff
@@ -38,6 +38,18 @@ function deserialize(string, options = {}) {
   return options.toRaw ? raw : Raw.deserialize(raw)
 }
 
+function getSerializedTextForNode(node) {
```
ianstormtaylor: Can you rename this to `serializeNode`, and can you move it below `serialize` and give it a DocBlock-style comment to match the others?

oyeanuj: Done! (And sorry about that missing the comment block!)

ianstormtaylor:
Thanks @oyeanuj! Can you add a test case that includes nested blocks that also contain inline nodes. With a quick read of the code I feel like we're missing that case, but I'm not sure.

oyeanuj:
Added, and passes!

ianstormtaylor:
@oyeanuj nice I like that test. I should have been more clear though, I think the case that might be failing is this one:

```
- document
  - block
    - text: ""
    - inline
      - text: one
    - text: ""
    - inline
      - text: two
    - text: ""
```

Which I _think_ will result in:

```
one
two
```

Even thought there is only a single block? Would you mind adding that as a test? If it doesn't fail, I'm probably just reading the code incorrectly, although it won't hurt to have it in there in the tests. Thanks for bearing with me!

oyeanuj:
@ianstormtaylor No worries, happy to make more changes to get it where it feels solid. And you were right about that case, fixed it now. Thanks for bringing it up, and let me know if anything else feels suspect to you!

ianstormtaylor:
Hey @oyeanuj it looks like there's still a case missing (the one I mention above) where a block has multiple top-level inlines inside of it:

```
- document
  - block
    - text: ""
    - inline
      - text: one
    - text: ""
    - inline
      - text: two
    - text: ""
```

Can you add that one? Thanks!

Issue #536, opened 2016-12-20
shouldComponentUpdate in isVoid nodes


Mangatt:
Hi,
I'm currently not able to determine, which nodes are inside selection, not with help of docs anyway.

It can be demonstrated at emoji demo (http://slatejs.org/#/emojis). When I click on emoji, it has blue border around it. But when I make selection with text around emoji, it doesn't have selected style.

Is there a way how to achieve that? To indicate that node is within selection?

![emoji](https://cloud.githubusercontent.com/assets/2706990/21364428/20f9f7e8-c6f1-11e6-8479-a2f86730ea95.png)

Mangatt:
Ok, I just found that state.blocks contain this information that I need. But there's another problem - selection change does not trigger isVoid node within selection re-render. What now?

thesunny:
Hmm... yeah, I can confirm this on blocks as well. Feels like this is a bug.

Mangatt:
Is there a way how to trigger re-render? Or do I have to change all void blocks into normal ones?

Mangatt:
I guess that it is components/node.js shouldComponentUpdate that is to blame. It is not taking void nodes into account at all.

Since I am not familiar with slate updating logic, is it possible to look into this? I found that @ianstormtaylor , you're author of this code, right?

Mangatt:
I am currently bypassing that with redux, but that is far from ideal.

michaelpietrykowski:
How did you end up handling this? I also need to force a re-render on my void nodes but haven't been able to do it.

Mangatt:
@michaelpietrykowski I'm using Redux to do that.

PR #1029, opened 2017-08-25
Make sure nodes always know if they're selected

tobiasandersen:
As described in #536, it's currently impossible (or at least very hard) for `block` and `inline` nodes to know whether or not they're part of a selection. The reason being that they only re-render if their focus changes, which I don't think is enough.

What's worked for us is to instead re-render if:

1. A node became part of a selection

**or**

2. A node was part of a selection, but isn't any longer.

### DIFF
--- a/examples/embeds/video.js
+++ b/examples/embeds/video.js
@@ -17,7 +17,7 @@ class Video extends React.Component {
 
   isSelected = () => {
     const { node, state } = this.props
-    const isSelected = state.selection.hasEdgeIn(node)
+    const isSelected = state.isFocused && state.blocks.includes(node)
     return isSelected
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Hey @tobiasandersen this sounds good to me!

tobiasandersen:
Okay cool, I'll go through it tomorrow to see if anything else needs to be updated.

tobiasandersen:
I couldn't see any problems with this. But I still updated the examples since I think this way is a bit more intuitive (happy to revert if you don't agree).

ianstormtaylor:
Nice, thanks @tobiasandersen I like this!

PR #1035, opened 2017-08-28
Reset drag counter properly in void nodes

tobiasandersen:
I stumbled upon this while debugging some weird behavior regarding DnD. Currently, once a void node has had something dragged over it, it will always stay editable (since the drag counter is never decremented).


TimYi:
I also find this issue. Comments and code are inconsistent.

ianstormtaylor:
Nice thanks @tobiasandersen!

PR #1037, opened 2017-08-29
Fix bug in insertBlock when focused node is void

skogsmaskin:
Fix bug where insertBlock and insertBlockAtRange always inserted the new block below the target, when focused node was a void block, even though the selection was collapsed to start of this block.

Also some tests for this.


ianstormtaylor:
Thanks @skogsmaskin!

PR #1038, opened 2017-08-29
Add tests for collapseToStart/End when selected node is void

skogsmaskin:
There were no tests for this so I added some.


ianstormtaylor:
Ah thank you @skogsmaskin, I really appreciate PR's like these!

PR #1041, opened 2017-08-30
Fix bugs related to void nodes within deleteAtRange

skogsmaskin:
If there were a void node in the beginning of the document and you selected everything, the void node would not get deleted.

There was also a bug where if you select a void node (and it's anchor and focus offset was 1) the block below would be deleted.

Also a bug where if you selected a text block and it was followed by a void block, the void block below would be deleted.

This PR fixes that.

Also several tests added.

EDIT: this was WIP, but I think it is ready now.


ianstormtaylor:
Thank you @skogsmaskin this is amazing!

PR #1042, opened 2017-09-01
Fix bug with up and down arrow keys near void blocks

skogsmaskin:
When pressing up-arrow key or down-arrow-key near a void block, it would not position the cursor properly, making shift-key-up-down-arrow-selections very difficult.

This will fix it.


ianstormtaylor:
Thanks @skogsmaskin!

PR #1043, opened 2017-09-02
Always call `preventDefault()` on `onDrop` events

AlbertHilb:
As stated in #9, when a drop happens inside Slate, on a position not allowed by the editor, then the default browser action takes place: for example, if an URL is dropped on a void node, the browser loads that resource in the current tab.

PR #921 should have solved this issue, but it didn't, at least in Firefox.
Drop a string on an emojii in the Slate example using Firefox to see that.

This PR fixes definitively the issue calling always `preventDefault` on  drop events.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -491,10 +491,10 @@ class Content extends React.Component {
    */
 
   onDrop = (event) => {
+    event.preventDefault()
+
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Sounds good to me, thanks @AlbertHilb!

PR #1044, opened 2017-09-02
Don't call `preventDefault` on `onDragOver` event.

AlbertHilb:
What is supposed to do the [`preventDefault` call](https://github.com/ianstormtaylor/slate/blob/master/src/components/content.js#L452) within the `onDragOver` `Content` handler?

In a non-editable context, default browser behavior is to not allow drop action, so in those circumstances calling `preventDefault` on `onDragOver` is necessary if you want give user a visual feedback.

On the contrary, when the target is `contentEditable`, like the Slate `Content` component, calling `preventDefault` on `onDragOver` stops the browser from showing with the text caret where the drop will take place: an extremely useful feedback!

If you drag around a string inside a Slate editor, you'll see the caret jumping onto the insert point only when drag enters a new element, for example when it enters a new paragraph. This happens because `preventDefault` is not called `onDragEnter` event.

This PR intends to fix this.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -449,8 +449,6 @@ class Content extends React.Component {
   onDragOver = (event) => {
     if (!this.isInEditor(event.target)) return
 
-    event.preventDefault()
-
     if (this.tmp.isDragging) return
     this.tmp.isDragging = true
     this.tmp.isInternalDrag = false


skogsmaskin:
I had a go at this as well https://github.com/ianstormtaylor/slate/pull/1040, but then I discovered there was a PR fixing this already. I hope we can merge some PRs soon, as it is starting to get difficult to track what is fixed and not.

ianstormtaylor:
Hey @AlbertHilb thank you for this. I fixed this with https://github.com/ianstormtaylor/slate/pull/1040 just because it removed one extra line break, but thank you for helping!

Issue #1039, opened 2017-08-29
Leaf Block Discussion


YurkaninRyan:
Recently, I've noticed bugs in `slate-edit-list` and `slate-edit-table`, where upon pasting, the inner most child would end up as the first block, and then after it the container would be pasted.  After digging into the code I think I found a couple of pain points.

`insertFragmentAtRange` does logic involving `firstBlock` and `lastBlock` being the same block.  I think this is where these two plugins start to fall apart.  `getBlocks` only returns leaf blocks, so you won't get your full table, or your bulleted list.  You'll get the inner most bullet, or a table cell.

Now, I know I've heard before that a lot of operations refer to Leaf Blocks, however, i'm wondering if that means we should rethink things like `wrapBlock`.  What do we need to fix here?  `insertFragment`'s behavior?

ianstormtaylor:
Hey @YurkaninRyan, search issues for `insertFragmentByKey`, I think there's a discussion of it. Let me know if you think that would solve the issue. I might be misunderstanding though.

YurkaninRyan:
@ianstormtaylor I think it just might, i'll take a crack at it

Issue #989, opened 2017-08-14
add `insertFragmentByKey` transform


ianstormtaylor:
I think lots of the confusion around `insertFragment` stems from it being used to insert fragments at specific ranges—the way that copy/paste works—instead of being used to insert a group of nodes all at the same time in a specific location.

If we added an `insertFragmentByKey(key, index, fragment)` transform I think that would give people the other option they're often looking for. The implementation would be pretty simple, it would just loop the `fragment.nodes` and call into `insertNodeByKey(key, index + i, node)`.

linonetwo:
Should I provide a unique key to each node in the Fragment?

I'm doing so on click a button, loads `section` from server, to solve https://github.com/ianstormtaylor/slate/issues/790 :

```js
props.editor.insertFragmentByKey(props.node.key, 0, new Document({ nodes: section }));
```

But it will warn a lot about `A node with a duplicate key of "139" was found! Duplicate keys are not allowed, you should use `node.regenerateKey` before inserting if you are reusing an existing node.`.

PR #1045, opened 2017-09-05
insertFragmentByKey

YurkaninRyan:
PR for #989 

I believe this will fix the issues with copy and paste that #1039 covers.


ianstormtaylor:
Beautiful, thank you @YurkaninRyan!

PR #1048, opened 2017-09-05
Update schema model documentation

travisperson:
When talking about `rules` and old interface was being used.


ianstormtaylor:
Awesome, thanks @travisperson!

PR #1033, opened 2017-08-25
Fixing onDrop for Inline Nodes, Harder type checking in normalize fun…

YurkaninRyan:
Addresses the issues found here https://github.com/ianstormtaylor/slate/issues/1025

PR #1049, opened 2017-09-05
fixing a breaking change in node/mark properties

YurkaninRyan:
@ianstormtaylor in #1033 I had a sloppy moment in `nodeProperties` and `markProperties` where I returned nothing if there was a `@@__SLATE` key in the node/mark.  This causes `properties` to be undefined for any transform and does some pretty bad stuff.

This new fix just makes it so it skips over those keys, which was the original intended effect.


ianstormtaylor:
Thanks @YurkaninRyan!

PR #1051, opened 2017-09-06
add setSave on Change

YurkaninRyan:
Let me know if i'm missing something but it doesn't look like `change.setSave()` actually made it in.  This PR fixes that!


ianstormtaylor:
Hey @YurkaninRyan thank you for catching this!

I've actually just added a commit/release that fixes this slightly differently (and updates the changelog to reflect it). Instead of `.setSave(true)`, it's now:

```js
change.setOperationFlag('save', true)
```

Which is slightly more verbose, but will handle other flags like `'merge'`, and others in the future without needing to add methods for each!

PR #1052, opened 2017-09-06
Add Linking to Change in docs

YurkaninRyan:
Currently you can't click on `Change` in the navbar on the documentation page :scream:


ianstormtaylor:
Ah good catch, thank you!

PR #1053, opened 2017-09-06
Fix hasUndos and hasRedos

YurkaninRyan:
since `state.history.undos` and `state.history.redos` are Immutable `Stack`'s we should look to see if their `size` is greater then 0 as opposed to their `length`

### DIFF
--- a/src/models/history.js
+++ b/src/models/history.js
@@ -118,11 +118,9 @@ class History extends new Record(DEFAULTS) {
       undos = undos.take(100)
     }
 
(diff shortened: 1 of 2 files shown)


YurkaninRyan:
Also digging into this some more, I believe redos used to be cleared if there was an undo placed on the stack?  Was that an intentional change or is that a bug?

ianstormtaylor:
Good catch, thanks @YurkaninRyan!

PR #1056, opened 2017-09-06
fixes errors for basic app with installation

peterklipfel:
The old onChange function caused a bunch of `Uncaught TypeError: _this.state.state.change is not a function` errors. Adding the object destructure to the arguments fixes this.


ianstormtaylor:
Thanks @peterklipfel!

PR #1062, opened 2017-09-06
Update walkthrough docs to use `change` object instead of `state`

davegw:


davegw:
@ianstormtaylor from the examples it looks like the best practice is to return `true` as the successful value from a plugin rather than the `change` object. Let me know if that's not the case and I can update accordingly.

YurkaninRyan:
https://github.com/ianstormtaylor/slate/pull/836

Per the comments in that pull it looks like it doesn't really matter what you return as long as you return something that isn't `null` or something at all.  I would guess that it's just personal choice

davegw:
Thanks for the reference @YurkaninRyan, was pretty sure it was just a matter of style preference, but figured we should aim to be consistent in usage across examples.

ianstormtaylor:
Thank you @davegw this is super super appreciated! I totally forgot about the walkthroughs when I was making the updates.

I think I agree that returning `true` should be the standard, since it feels like the least amount of magic considering that it feels similar to native event handlers. But returning `change` is also an option for terse-ness when people want to do that too.

Issue #1058, opened 2017-09-06
shouldComponentUpdate and Selection


YurkaninRyan:
While upgrading `slate-sticky-inlines` to `slate@0.22.4` I noticed some strangeness with #1029 and inlines.  For `blocks` it works perfectly, but `inlines` aren't guaranteed to re render even if their selection has changed, because a parent node could choose to not update, which disrupts the entire chain.

Should it be that the only nodes the implement shouldComponentUpdate are leaf nodes/text nodes?  It feels like every node should have a chance to say whether or not it should update.

@tobiasandersen

ianstormtaylor:
Actually, looking at that PR again, I think it will also fail for nested blocks/inlines, since `state.blocks` and `state.inlines` only ever contains the leaf blocks and root inlines. So if it's nested, it will never show up as selected?

YurkaninRyan:
So I wonder what the possible solutions are 🤔  is Selection Changes the only real use case we have where parents wouldn't easily know if their children would need to update or not?

tobiasandersen:
You're right, I didn't take those cases into consideration. Not sure what the best solution is though. We could always take back the old logic as well, but we still wouldn't cover all use cases, right?

PR #1064, opened 2017-09-07
sCU Changes for #1058

YurkaninRyan:
In #1058 we figured out that the selection logic checks in sCU aren't aggressive enough to cover every case.  I believe this change would cover all of our bases.  Basically, if you have nodes that aren't text nodes, then you must let your child nodes determine wether or not they can re-render.  We can change this to be more performant if we find that selection is the only case where this is true, but I kept it open ended.

I was having trouble figuring out how to run the performance tests, is there any documentation on that?

Thoughts?

### DIFF
--- a/src/components/node.js
+++ b/src/components/node.js
@@ -111,6 +111,10 @@ class Node extends React.Component {
     // for simplicity we just let them through.
     if (nextProps.node != props.node) return true
 
+    // If the Node has more Nodes inside of it that aren't Texts, allow
+    // them to decide if they should render or not.
+    if (nextProps.node.nodes.size === nextProps.node.getTexts().size) return true
+
     // If the node is a block or inline, which can have custom renderers, we
     // include an extra check to re-render if the node either becomes part of,
     // or leaves, a selection. This is to make it simple for users to show a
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @YurkaninRyan thanks, I like this idea. I feel like it probably won't cause huge issues for performance because nested nodes are going to be rarer—like less than 20% at least. But I'm not totally sure. Either way, I'm down to try it until it becomes an issue and we have to figure something else out.

I think the current way of checking is either not always guaranteed to be right, or else just a bit of a mind bender. I _think_ with the new `Text.isTextList(value)` method in the latest release we can change the check to be:

```js
if (nextProps.node.kind != 'text' && Text.isTextList(nextProps.node.nodes) == false) return true
```

And that should cover all cases?

YurkaninRyan:
@ianstormtaylor So looking at `isTextList`, it seems that it only checks if the first value in the list is a `Text` in order to return `true`.  Would that be a problem if a block contained an `inline` that isn't at the start?

ianstormtaylor:
@YurkaninRyan actually just recently changed it to check every value in the list!

YurkaninRyan:
You got it!  Then that makes perfect sense!

PR #1067, opened 2017-09-07
remove all deprecated methods, closes #1034

### LINKED ISSUE #1034: remove old deprecations completely to save filesize
For versions that are no longer prevalent, since they're just taking up space. Not a huge priority, but good to do at some point.

ianstormtaylor:
Not going to merge this right now, since it's another breaking change. But whenever the next minor release happens I think it would be good to reset the technical debt from having these around.


YurkaninRyan:
Not sure if it was added in here, but `onDocumentChange` and `onSelectionChange` was also deprecated recently

ianstormtaylor:
Good point. There's many more added recently, I'll just re-write this right before we want to remove the deprecations. Hopefully soon.

PR #1071, opened 2017-09-07
Add missing import

skogsmaskin:
Text was not imported into the node component, breaking things (happened in https://github.com/ianstormtaylor/slate/pull/1064).

``TypeError: Text.isTextList is not a function``


YurkaninRyan:
Strange the linter didn't catch that for me 🤔

AlbertHilb:
> Strange the linter didn't catch that

It's because `Text` is among browser global variables.

YurkaninRyan:
Brutal!

ianstormtaylor:
Thanks y'all. Browser global overlap is pretty frustrating, but I think it's still worth keeping.

PR #1082, opened 2017-09-07
refactor delete at range, fixing edge cases

### LINKED ISSUE #548: preserve block type when joining with empty adjacent node

ianstormtaylor:
This cleans things up a lot. It handles:

- All cases (that I can think of) of deleting a range that starts of ends with a void node. 
- The "hanging" selections being deleted, (the triple-click issue), where you don't actually want the final node to be deleted.
- Not merging the child nodes of a block if the entire start block was deleted, so that you don't accidentally keep a block quote style if you're deleting the entire quote.

Fixes #548

cc @skogsmaskin

### DIFF
--- a/src/changes/at-range.js
+++ b/src/changes/at-range.js
@@ -16,16 +16,6 @@ import { List } from 'immutable'
 const Changes = {}
 
 /**
- * An options object with normalize set to `false`.
- *
- * @type {Object}
- */
-
-const OPTS = {
-  normalize: false
-}
-
-/**
  * Add a new `mark` to the characters at `range`.
  *
  * @param {Change} change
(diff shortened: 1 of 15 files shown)


skogsmaskin:
Nice refactor! Seems like all the edgecases are dealt with.

PR #1085, opened 2017-09-07
Remove an incorrect selection manipulation from Content

skogsmaskin:
This was a little too eagerly placed here while trying to deal with the deleteRange edge cases. It must be removed.

### DIFF
--- a/src/components/content.js
+++ b/src/components/content.js
@@ -803,12 +803,6 @@ class Content extends React.Component {
         properties.focusOffset = 0
       }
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Nice thanks @skogsmaskin! I think the check above might also be removable?

skogsmaskin:
If I remove the one above the "delete the whole selected first non-void-block, but not the following void-node"-issue occurs and the void-node gets deleted.

ianstormtaylor:
@skogsmaskin okay cool I'll check that one out locally and see what we can do!

PR #1088, opened 2017-09-08
Remove is native

### LINKED ISSUE #963: remove isNative?

ianstormtaylor:
Thanks to @conorcussell for kicking this off in #964!

**If anyone can test this branch to double-check my work, that would be super appreciated!**

This removes the `isNative` flag from Slate, with seemingly zero performance reprecutions! 🎉  Instead of spending a significant amount of time calculating things in `onBeforeInput`, it now spends an equivalent-ish amount of time in `render` and `updateSelection`. Previously these two were being skipped for "native" updates.

As far as I can tell the performance benefits from `isNative` are negligible with all of the other performance handling logic we have (memoizing method, `shouldComponentUpdate` shortcuts, etc.) so we can remove it. Which seems nice, because it is a cause for lots of headaches UX-wise.

I won't merge this for a little bit, because I want to give the recent `0.22` a chance to settle and for plugins to get updated first before doing another breaking change. (Even though it is way less breaking.)

Fixes #963


YurkaninRyan:
Looked over it and nothing is glaring,  this is going to do amazing things for the complexity of this part of the codebase though 🥂

Issue #1089, opened 2017-09-08
Documentation bug, onKeyDown(), State -> Change


stefanheine:
https://docs.slatejs.org/reference/plugins/plugin.html#onkeydown
documents:
Function onKeyDown(event: Event, data: Object, **state: State,** editor: Editor) => State || Void

but in fact, probably since v0.22.0, the third parameter is a a **Change** instead of the State.

PR #1093, opened 2017-09-08
Update plugin reference to use Change instead of State

tobiasandersen:
Fixes https://github.com/ianstormtaylor/slate/issues/1089 along with the rest of the event handlers.


ianstormtaylor:
Thanks @tobiasandersen!

PR #1095, opened 2017-09-08
start converting to use lerna

### LINKED ISSUE #848: Split slate into multiple NPM modules
At GitBook, we are using more and more Slate for its data modeling in our backend. But one disturbing thing is to always have `react` required as a peer dependencies, while most of our modules only requires the models from Slate.

I'm suggesting to use [lerna](https://github.com/lerna/lerna) and publish slate as at least 3 modules:

- `slate-core`: exports `Block`, `Text`, `Document`, `State`, `Inline`, `Transform`, `Selection`, `Character`, `Range`, `resetKeyGenerator`, `setKeyGenerator `
- `slate-renderer`: exports the react component `Editor`
- `slate`: exports everything by having the two other modules as dependencies

It'll make it easier to also move the serializers to specific modules (`slate-serializer-html`, `slate-serializer-plain`), and maybe these serializers should be optional dependencies.

I can also see more and more stuffs arriving as dependencies, for example the collaborative/OT (https://github.com/ianstormtaylor/slate/issues/259) should only requires the `slate-core`.

Doing this will **not** be a breaking change, since `slate` will keep exporting everything.

ianstormtaylor:
Still in progress. I think I'm noticing a few things I'd like to do before pulling the trigger on this:

- [ ] Change the `Raw` serializer to be part of each model.
  - [ ] Remove `terse: true` from raw serialization.

Fixes #848 

---

This splits up the current `slate` package into multiple packages:

- `slate` (changes, models, operations, core schema, utils)
- `slate-react` (components, core plugin, utils)
- `slate-plain-serializer`
- `slate-html-serializer`
- `slate-prop-types`

And...

- `slate-base64-serializer` (maybe)

All of the new packages have a peer dependency on `slate`.

---

I'm thinking this is going to be a breaking change, because the `slate` package will now contain only the core pieces of slate—things that plugins, server-side, etc. would all depend on. And the other packages will be broken out into their own.


ianstormtaylor:
Closing in favor of #1106.

PR #1098, opened 2017-09-09
start removing raw serializer

### LINKED ISSUE #585: remove "terse" representation in Raw serializer
I feel like it adds extra complexity to dealing with "raw" that isn't really needed for most cases, and if it really is needed for someone, they're probably going to want to go with an even more minified representation that saves more bytes with keys being minified and other things too.

Right now I think it's really only useful to make typing JSON representations slightly easier, which isn't enough gain for the uncertainty it introduces.

@SamyPesse @Soreine curious if you can think of any other "legit" uses of terse mode that I haven't thought of, or whether you've also avoided it.

### LINKED ISSUE #602: change Html serializer's `defaultBlockType` option to `defaultBlock`
Since it no longer accepts only a `type`, as of https://github.com/ianstormtaylor/slate/pull/595

### LINKED ISSUE #1092: consider using slate-sugar for writing tests
Right now we use YAML for creating documents for tests, which is not bad honestly. But one thing I'd love to have is the entire test defined in a single file, so you wouldn't have to switch between `index.js`, `input.yaml` and `output.yaml` all the time.

With YAML this seems less nice, because defining YAML in JS means using big strings without highlighting. But with [`slate-sugar`](https://github.com/GitbookIO/slate-sugar) and JSX this might be nicer?

I could imagine a test file like:

```js
export const input = (
  <state>
    <document> 
      <paragraph>one</paragraph>
    </document>
    <selection 
        anchorKey="1" 
        anchorOffset={1} 
        focusKey="1" 
        focusKey={2}
    />
  </state>
)

export const output = (
  <state>
    <document> 
      <paragraph>oe</paragraph>
    </document>
    <selection 
        anchorKey="1" 
        anchorOffset={1} 
        focusKey="1" 
        focusKey={1}
    />
  </state>
)

export default function (change) {
  change.delete()
}
```

Or even defining a selection using `<anchor/>`, `<focus/>` and `<cursor/>`, like:

```js
export const input = (
  <state>
    <document> 
      <paragraph>
        o<anchor/>n<focus/>e
      </paragraph>
    </document>
  </state>
)

export const output = (
  <state>
    <document> 
      <paragraph>
        o<cursor/>e
      </paragraph>
    </document>
  </state>
)

export default function (change) {
  change.delete()
}
```

I think it could make tests a lot terser, and way easier to reason about.

cc @Soreine @SamyPesse

### LINKED ISSUE #1096: change `Raw` serializer to be part of models as `to/fromRaw`
I noticed this issue while working on #1095. Basically these things are in conflict:

- Slate serializers should be split apart from "core", such that no one serializer is given more "priority" in the abstraction that others.
- Slate serializers should only need a peer dependency on `slate` to work. (And of course other peer dependencies if they need for the specific serializer, like `react` for the `Html` serializer.)

Because of how `Raw` is currently handled, it feels like it should be separated from "core" since it's one of the many serializers. But at the same time it is so central to everything, and the other serializers would then all want to depend on it too.

I had been thinking this before, but running into that issue made me more aware of it.

I think instead, "raw" should be part of core and added to the Slate models, so that you could do:

```js
Block.fromRaw({
  type: 'paragraph',
  nodes: [{
    kind: 'text',
    text: 'A string of text.',
  }],
})
```

Or...

```js
const raw = Block.toRaw()
```

The `fromRaw` method sounds similar to `Block.create`, but `create` would actually use `fromRaw` under the covers. Keeping `create` as the flexible one, with `fromRaw` expecting a properly formed raw JSON object.

---

Alternatively, we could override the `toJSON` methods of Immutable.js records. But I'm not sure if we'll regret that or not. It seems like it would be cleaner mentally if possible.

ianstormtaylor:
This pull request gets rid of the complexity of the "raw" serialization concept of Slate, and instead introduces the same thing as the canonical JSON representation for Slate objects. This means...

###### NEW

- **Slate models now have `Model.fromJSON(object)` and `model.toJSON()` methods.** These methods operate with the canonical JSON form (which used to be called "raw"). This way you don't need to `import` a serializer to retrieve JSON, if you have the model you can serialize/deserialize.

- **They also have `toJS` and `fromJS` aliases.** This is just to match Immutable.js objects, which have both methods. For Slate though, the methods are equivalent.

###### DEPRECATED

- **The `Raw` serializer is now deprecated.** Instead, use the `fromJSON` and `toJSON` models directly.

- **The `toRaw` options for the `Plain` and `Html` serializers are now called `toJSON`.** To stay symmetrical with the removal of the "raw" concept everywhere.

- **The `terse` option for JSON serialization has been deprecated!** This option causes lots of abstraction leakiness because it means there is no one canonical JSON representation of objects. You had to work with either terse or not terse data.

- **The `Html` serializer no longer uses the `terse` representation.** This shouldn't actually be an issue for anyone because the main manifestation of this has a deprecation notice with a patch in place for now.

- **The `defaultBlockType` of the `Html` serializer is now called `defaultBlock`.** This is just to make it more clear that it supports not only setting the default `type` but also `data` and `isVoid`.

Fixes #585 
Fixes #602 
Fixes #1092 
Fixes #1096


eromoe:
Hi, I have a piece of code using `"slate": "^0.21.2",` 

there are code like:
```
state.set('editor', Raw.deserialize(next_editor, { terse: true }))

state.get('editor').transform().toggleMark(action.mark).apply()
```

What is the proper way to translate those to latest version?

doodlewind:
@eromoe Since `next_editor` is a plain JSON, `State.fromJSON(next_editor)` should work.

eromoe:
Hi @doodlewind 
I tried both
```
var z = State.fromJS(next_editor)
```
and 
```
var z = State.fromJSON(next_editor)
```
z is not correct
image

-----------------------
I work it out by 

```
      var z = State.fromJS({
        document: Document.fromJS(next_editor)
      })

      return state.set('document', action.doc)
        .set('editor', z)
```

Looks like the old `Raw.deserialize`  do a lot of things.

PR #1099, opened 2017-09-09
Fix state.isEmpty

HaNdTriX:

`state.isEmpty()` returns true even though the selection state is not empty.

`state.isEmpty()` should return false when there is text in the selection state.

#### Example

Check out http://slatejs.org/#/hovering-menu and select an entire block.

hover-menu-bug

#### Solution 

~~Use `state.fragment.text.size` instead of `state.fragment.text.length`~~

```js
if (startOffset < endOffset) {
    return false
}
```

### DIFF
--- a/src/models/state.js
+++ b/src/models/state.js
@@ -502,7 +502,7 @@ class State extends Record(DEFAULTS) {
       return true
     }
 
-    if (endOffset != 0 && startOffset != 0) {
+    if (startOffset < endOffset) {
       return false
     }
 


ianstormtaylor:
Hey @HaNdTriX thanks for finding this! Im fairy sure that `fragment.text` is a string, so using `.size` doesn't feel like the right solution. Maybe there's a separate issue with the `isEmpty` method? Or do you know why this is the fix?

HaNdTriX:
My fault! :grimacing: I have updated the [PR](https://github.com/ianstormtaylor/slate/pull/1099/files)

ianstormtaylor:
No worries @HaNdTriX! 

Can you explain your change? I can't tell what situations it covers from just looking at it. Thank you!

ianstormtaylor:
@HaNdTriX I think this was solved in one of the latest fixes to `getFragmentAtRange`. Thanks for taking the time to submit this pull request though!

PR #1100, opened 2017-09-09
Avoid finding native selection focus position when selection is collapsed

AlbertHilb:
When updating selection, if it is collapsed, there is no need to look for focus position: just reuse already found anchor position.
Move code to get caret position from a selection point to a standalone function in `utils`.


ianstormtaylor:
Super good call @AlbertHilb, and beautiful PR. Thank you!

Issue #1103, opened 2017-09-10
Caret navigation issue on Firefox when void blocks are selected


AlbertHilb:
On Firefox, when void block are selected, pressing down arrow has no effect.
Select the image on "Images" example, then press down arrow to experience this problem.
This is longstanding bug.

AlbertHilb:
Fixed by #1104.

PR #1104, opened 2017-09-10
Reshape void component structure.

AlbertHilb:
This PR modifies the structure of the component used to render void blocks, mainly, to get rid of the ugly off-screen spacer.
As side effect it fixes issue #1103.
Maybe it also allows us to throw away some [hacks](https://github.com/ianstormtaylor/slate/blob/master/src/components/content.js#L767). :thinking:
On the other side this introduce an issue in the "Embeds" example: video does not respect the supplied size.
I'm going to fix this issue in a coming PR.

### DIFF
--- a/examples/embeds/video.js
+++ b/examples/embeds/video.js
@@ -59,30 +59,22 @@ class Video extends React.Component {
 
     const wrapperStyle = {
       position: 'relative',
-      paddingBottom: '66.66%',
-      paddingTop: '25px',
-      height: '0',
       outline: isSelected ? '2px solid blue' : 'none',
     }
 
     const maskStyle = {
       display: isSelected ? 'none' : 'block',
       position: 'absolute',
-      top: '0px',
-      left: '0px',
-      right: '0px',
-      bottom: '0px',
+      top: '0',
+      left: '0',
       height: '100%',
+      width: '100%',
       cursor: 'cell',
       zIndex: 1,
     }
 
     const iframeStyle = {
-      position: 'absolute',
-      top: '0px',
-      left: '0px',
-      width: '100%',
-      height: '100%',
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
@AlbertHilb this is great! I would love to be able to simplify this thing, and from my testing it seems like this still preserves all the cases. (And fixes the Firefox bug like you mentioned!)

Do you know a solution for the embed not respecting the size problem?

Further, is it problematic for any block children to live inside an `inline-block`? Just wondering because we don't have control over what people put in voids, so inside block void nodes they should be able to put block-level nodes without issue.

Also, I think if you add `width: '0px'` to the spacer for block nodes, it will remove the little bit of extra width it creates on the left of the image (which pushes the image in ever so slightly).

I like this where is going!

AlbertHilb:
> Do you know a solution for the embed not respecting the size problem?

Yes, I added a commit to fix that

> is it problematic for any block children to live inside an inline-block?

No. As stated in the [specification](https://www.w3.org/TR/CSS2/visuren.html#display-prop), "the inside of an inline-block is formatted as a block box".
Anyway, `inline-block` is a legacy value. There are more modern values I could have used, `inline flow-root`, that is strictly equivalent to `inline-block`, or `flex`, but I want to be conservative: I can't test changes on browsers other than Firefox and Chrome, and, more importantly, I saw there are people who want to use Slate on old browsers (Internet Explorer) and I fear new values are not well supported by them.

> I think if you add width: '0px' to the spacer for block nodes, it will remove the little bit of extra width it creates on the left of the image (which pushes the image in ever so slightly).

Added!

ianstormtaylor:
Thanks @AlbertHilb, this looks great!

Issue #1012, opened 2017-08-19
A couple of problems with core plugin onBeforeInput


danburzo:
While investigating IME bugs, I came across a few problems with `onBeforeInput` in the core plugin:

1. Is it assumed here that `e.data` will contain a single character?

https://github.com/ianstormtaylor/slate/blob/239d97fcad6d097d564674846a08924e78d2a503/src/plugins/core.js#L89-L98

In the case of a composition event, `onBeforeInput` will receive the entire composed string in `e.data` just before calling the `compositionend` event, so instantiating a Character with a multiple-char string may break somewhere. (At fist impression, seems the list insertion works either way)

2. I understand this line should mimic what a native insertion would be?

https://github.com/ianstormtaylor/slate/blob/239d97fcad6d097d564674846a08924e78d2a503/src/plugins/core.js#L100

If that's the case, it should take the selection into consideration, since you may be actually replacing some characters in the process.

This is all I have for now in regards to inserting composed words, I'll keep investigating — is there any other place that assumes e.data is a single char but then ends up pushing a multi-character string into the state? That may be the source of some of the problems we currently have with IMEs.

Right now I reckon these are causing:  

1. Unnecessarily using a non-native insertion when it could have been native, due to mismatch between computing "native" outcome vs. "synthetic" outcome
2. Noticing that when overwriting text with a composed word, the outcome of the "synthetic" action is not accurate (selection is out of sync in the meantime), but the "native" one is, allowing a native insertion might fix some bugs.

danburzo:
A big chunk of problems with IME is caused [by the iOS autocomplete fix](https://github.com/ianstormtaylor/slate/pull/655), as correctly identified in these pull requests:

* https://github.com/ianstormtaylor/slate/pull/881 — it proposes making that transform _only_ on IOS.
* https://github.com/ianstormtaylor/slate/pull/885 — it proposes making that transform regardless of platform, but not when composing, since that introduces regressions.

If I were to choose between the two, I'd go with the one that limits it to IOS, since leaving it around may have other implications other than composing. I'm looking to see if the IOS fix can be achieved in a way that doesn't cause regressions with composing.

danburzo:
A tangential update on IMEs / iOS autocomplete fix.
* I ended up building this http://danburzo.github.io/input-methods to help debug the events triggered for various composition actions
[...]

PR #1107, opened 2017-09-11
Handle IME composing state in `onBeforeInput`

doodlewind:
This PR suggests a moderate solution for issue #854 based on PR #885 and #880, which can also be a prerequisite to solve issue #1012.

Desktop IME has native behaviors which is different from mobile. Briefly, there are three types of events related to IME:

- `onCompositionStart`
- `onCompositionUpdate`
- `onCompositionEnd`

On mobile the `onCompositionStart` and `onCompositionUpdate` event does not ever fire since user can choose words inside IME UI and insert the result all at once with `onCompositionEnd` fires. On desktop the default behavior generally works as below:

1. When user inputs first character with IME, `onCompositionStart` fires.
2. When user inputs each character with IME, `onCompositionUpdate` fires on each key down, but during which `onKeyDown` does not fire.
3. When user choose the resulting word to input via <kbd>space</kbd> or mouse, `onBeforeInput` fires.
4. After `onBeforeInput` fires, `onCompositionEnd` fires.

One essential key to IME related issues is that `onBeforeInput` fires **before** `onCompositionEnd`, so it's necessary for `onBeforeInput` to know whether current input state is composing. In this PR we mimic the [onKeyDown](https://docs.slatejs.org/reference/plugins/plugin.html#onkeydown) way that pass in a `isComposing` state inside `data` arg to plugin's `onBeforeInput`, which enables both core and custom plugins to add extra logic fixing IME state.

As as result, `onBeforeInput` inside core plugin can have more control: The reason causes #854 is that during desktop IME inputs, the iOS COMPAT code which manually changes selection inside  core plugin's `onBeforeInput` should not have been worked. Here is an example: 

1. If a Chinese IME user begins input with init selection collapsed, after 5 keys `haode` stroke, IME will interpret the keys as `hao'de` with 6 letters, which appears on screen underlined (`haode` means `好(hao)的(de)` in Chinese).
2. When user ends input, `onBeforeInput` fires.
3. Before the word `好的` applys to state, the iOS COMPAT code inside `onBeforeInput` will consider it a mismatch, selecting a new selection with a width of 6 letters after current cursor, then replace them with `好的`, but since the temp letters `hao'de` is natively removed by browser, this is not necessary, resulting 'eating' letters after current cursor.

In this case, with `isComposing` provided to plugin, simply ignoring iOS COMPAT code on desktop solves this issue.

Hope it helps :-)


ianstormtaylor:
@doodlewind this is an extremely written up PR! Thank you!

I'll be back in a few hours and I can check this out locally and hope to get it merged. Thanks for taking the time to work on this!

src/utils/find-deepest-node.js
```diff
@@ -7,7 +7,7 @@
  */
 
 function findDeepestNode(element) {
-  return element.firstChild
+  return element && element.firstChild
```
ianstormtaylor: @doodlewind do you know where this was causing an issue?

danburzo:
@doodlewind thank you for the thorough write-up! 

I'd looked at the IME problem when I opened #1012 and I understand your solution echoes PR #885, that is go around the problem introduced by the iOS autocomplete fix for _composition_ specifically. I have a feeling that composition may not be the only area affected by that, so merging this patch will potentially obscure other issues.

src/utils/find-deepest-node.js
doodlewind: This should have been done with another PR. When element itself is null, IMO returning null is somehow better then throwing an error. This happens when removing nodes in a loose schema. Sorry for not providing reproduction here, while if this is not suitable to be merged here, it's okay to close this with another pull request.😅

doodlewind:
@danburzo thanks for your insight!
You are suggesting Input Events on iOS 10.1+ on mobile, this is a great step on mobile and since this PR focus on desktop IME, our solutions can be in coexistence. In fact composition state is not only related in issue #854 , I submitted issue #1114 showing another case needs to be handled with desktop composition events, which is also related to native desktop IME behaviors. To solve this, IMO it also requires core plugin's `onBeforeInput` to know composition state to handle IME's native behavior. Very welcomed for your suggestion. 🙏

doodlewind:
Closing this since my forked branch contains some unrelated changes.

PR #1108, opened 2017-09-11
Fix an invalid require statement in base-64 serializer

skogsmaskin:
It's puzzles me that it has to be required like this, but I see things break if I do a ``import Node from '../models/node'`` at top.

Anyway, it must get the ``.default`` export if required.


ianstormtaylor:
Ah shit, sorry. I need to add a few tests for the Base64 serializer.

Yeah there was an annoying circular dependency issue and I didn't want to solve it while refactoring so I just added that hack for now 😄

PR #1117, opened 2017-09-12
Add replaceNodeByKey change

skogsmaskin:
This follows up https://github.com/ianstormtaylor/slate/pull/665.

Add ``change.replaceBlock(oldBlock, newBlock)`` method to replace a block with another block.

Remove code that replaced an empty block in ``.insertBlock``

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -601,8 +601,7 @@ Changes.insertBlockAtRange = (change, range, block, options = {}) => {
   }
 
   else if (startBlock.isEmpty) {
-    change.removeNodeByKey(startBlock.key)
-    change.insertNodeByKey(parent.key, index, block, { normalize })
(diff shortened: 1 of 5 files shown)


ianstormtaylor:
Hey @skogsmaskin this is great!

I think we should call this method `replaceNodeByKey` to match the other ones that take a specific node as input, and also not to mention block because it doesn't appear block-specific?

Thanks!

skogsmaskin:
@ianstormtaylor, good points!  I have updated the PR. Not sure about the options.normalization though, is it necessary?

ianstormtaylor:
Nice, looks perfect @skogsmaskin, thank you!

I just realized this is a breaking change for people currently relying on `insertBlock`, so I'm going to label it as such and wait for another breaking release to merge this in, to try to lessen the number of breaking releases a bit. Hope that's okay!

skogsmaskin:
@ianstormtaylor yes there is no rush :+1:

PR #1124, opened 2017-09-14
sCU - Let only nodeless Nodes determine if they need to update or not.

YurkaninRyan:
This change adds a line that doesn't allow a node to return false in `shouldComponentUpdate` unless it has no children nodes.  Here is the problem this is trying to solve.  Say you have something that looks like this.

```
- Paragraph
  - Text
  - Link
  - Text
- Paragraph
  - Text
```

If at any point `Paragraph` decides that it shouldn't update, every node underneath it won't.  This would be a safe assumption but since `Node` is a recursive component that is basically walking a tree.  It's difficult to make calculations at the top of the tree that determine if the rest of the tree is useless or not.

### DIFF
--- a/packages/slate-react/src/components/node.js
+++ b/packages/slate-react/src/components/node.js
@@ -3,6 +3,7 @@ import Base64 from 'slate-base64-serializer'
 import Debug from 'debug'
 import React from 'react'
 import ReactDOM from 'react-dom'
+import Slate from 'slate'
 import SlateTypes from 'slate-prop-types'
 import Types from 'prop-types'
 import getWindow from 'get-window'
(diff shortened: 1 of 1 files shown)


YurkaninRyan:
@ianstormtaylor I updated this to use the `isTextList` check as well.  However, unless i'm misunderstanding, I don't believe the other optimization you suggested would work all the time.

Say your selection is inside a child inline.  You click on another line entirely.  now:
  1. p.node.isSelected is `true`
  2. n.node.isSelected is `false`
  3. The child still has the `isSelected` prop of `true`

In this scenario, you are also no longer selecting the child, but since we would return `false` because the parent isn't staying selected, the child never gets a chance to update it's props.

The reverse is true as well.  What if you click from a line, into a line that wasn't previously focused, into a child inline.  Now we have:
  1. p.node.isSelected is `false`
  2. n.node.isSelected is `true`
  3. the child still has `isSelected` prop of `false`

This also wouldn't update the inline to have the proper `isSelected`.  The parent somehow needs to know if it has a child that needs to be allowed to update

YurkaninRyan:
Closing in favor of #1125 !

PR #1125, opened 2017-09-14
Fix SCU `isSelected` logic.

### LINKED ISSUE #1102: Issue selecting inline void nodes.
Open "Emojiis" example, place the caret on the left of an emojii ((few chars away), then move the caret to the right, the emojii does not appear selected (no blue outline).
If you select the emojji with a mouse click, the blue outline appears but remains there even if you move away the caret pressing left or right arrows.

AlbertHilb:
This fixes #1102.

The sCU `isSelected` logic could be made more effective if `isSelected`was a three-state variable:

* `-1` if the node does not intersect selection
* `0` if the node contains one of the selection edges
* `1` if the node is all inside the selection

In this case we can avoid to update a node that is, and was already, entirely selected.

But I don't know if the gain is worth the code we should add.

### DIFF
--- a/packages/slate-react/src/components/node.js
+++ b/packages/slate-react/src/components/node.js
@@ -114,9 +114,12 @@ class Node extends React.Component {
     // for simplicity we just let them through.
     if (n.node != p.node) return true
 
-    // If the node's selection state has changed, re-render in case there is any
-    // user-land logic depends on it to render.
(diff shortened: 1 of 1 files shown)


YurkaninRyan:
I think this makes sense to me.  As long as selected is true or was true the render will happen?

ianstormtaylor:
Thanks @AlbertHilb!

PR #1131, opened 2017-09-15
Fixing Drag and Drop, and getFragmentAtRange

### LINKED ISSUE #1123: Drag&Drop is broken
Open an Example, select a word, drag it around, then drop it.
The word disappears in the original position but nothing is inserted in the drop position.
That happens because the drag&drop `fragment` is empty.
This regression has been introduced by PR #836.

YurkaninRyan:
fixes #1123 

This one is weird, and I think there may be another underlying problem?

Basically, if they start and end keys are the same, we `getNextText`, but that always returns an empty text node next to our startNode.  This causes splice to fire on the same index, and returns no  nodes.

Is there a better way to do this?  I don't think `getNextText` should be modified.  Is it okay to assume that we just need to do it twice?

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -762,7 +762,7 @@ class Node {
     }
 
     // Get the start and end nodes.
-    const next = node.getNextText(startKey)
+    const next = node.getNextText(node.getNextText(startKey).key)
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Nice thanks @YurkaninRyan I think doing it twice is exactly what we needed there.

Issue #1133, opened 2017-09-18
Scroll position update issue


AlbertHilb:
When the caret is moved programmatically inside a hidden portion of the editor without changing the document, the scroll position isn't updated in order to make it visible.

See [this example](https://jsfiddle.net/Ls6wy7jz/28/).

AlbertHilb:
Fixed by #1134.

PR #1134, opened 2017-09-18
Call `scrollToSelection` on selection update.

AlbertHilb:
Why all [this complexity][1] to update the editor scroll position?
Isn't it sufficient to call `scrollToSelection` on selection update?

This PR implements that idea and it fixes issue #1133 too.

[1]: https://github.com/ianstormtaylor/slate/blob/f52967c9e6cc821856bbd5537e72ae0f52eac6c3/packages/slate-react/src/components/node.js#L183

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -17,6 +17,7 @@ import getHtmlFromNativePaste from '../utils/get-html-from-native-paste'
 import getPoint from '../utils/get-point'
 import getTransferData from '../utils/get-transfer-data'
 import setTransferData from '../utils/set-transfer-data'
+import scrollToSelection from '../utils/scroll-to-selection'
 import { IS_FIREFOX, IS_MAC, IS_IE } from '../constants/environment'
 
 /**
@@ -164,6 +165,8 @@ class Content extends React.Component {
     range.setStart(anchor.node, anchor.offset)
     native.addRange(range)
     if (!isCollapsed) extendSelection(native, focus.node, focus.offset)
+
+    scrollToSelection(native)
 
     // Then unset the `isSelecting` flag after a delay.
     setTimeout(() => {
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Haha great question @AlbertHilb! My guess is that the `<Node>` code was written back when we had `<Leaf>` nodes in charge of updating the selection instead of `<Content>`. But this looks a lot cleaner to me. Thanks!

---

_Side note: I've just realized, one issue we have (with both solutions I believe) is that if you `applyOperations` in the case of a collaborative editor, we need to have the scrolling distinguish between whether the change has happened on your local machine or whether it's a remote one being applied. But that can be a future problem to solve._

AlbertHilb:
> Side note: I've just realized, one issue we have (with both solutions I believe) is that if you `applyOperations` in the case of a collaborative editor, we need to have the scrolling distinguish between whether the change has happened on your local machine or whether it's a remote one being applied. But that can be a future problem to solve.

I don't think this will be a hard problem to solve. :blush:
Of course, the exact solution depends on how the collaborative editing will be implemented.
Anyway, you can always apply remote changes without altering the (local) selection, in that case `updateScroll` is not fired.

See [this example](https://jsfiddle.net/qtoku90g/4/)

ianstormtaylor:
@AlbertHilb, I think the harder part will be for operations that have an effect on the selection. For example an `insert_text` in the start node but before the current selection. It'll need to increment the start offset. Or for split/merge node.

AlbertHilb:
Hi @ianstormtaylor.

>  I think the harder part will be for operations that have an effect on the selection.

I guess the difficulty is in calculating the new selection, not with `updateScroll`.

The only problem, I see, with the current `updateScroll` code arises when the change happens without modifying the selection and above it; like in [this example](https://jsfiddle.net/wb0mwbac/2/).
But this should be easy to fix.

PR #1140, opened 2017-09-19
Add a way to customize shouldComponentUpdate logic

### LINKED ISSUE #959: Support for dynamic decorator

feature

Currently, decorators doesn't get re-calculated when the node doesn't change. This is great for #hashtag and @mention but bad for search highlight where the decorator is driven by external changes instead of internal changes.

The problem is that we are not able to communicate to `shouldComponentUpdate` of `Node` to return `true` when such external changes occurs. This end up skipping rendering its child text `Node` that will actually redecorate the decorator.

I notice there's a way to `suppressShouldComponentUpdate` but it seems fairly obscure to use. I'm looking for ways we can pass maybe a method in from `Editor` that can control this behaviour, kinda like how `suppressShouldComponentUpdate` is meant to do.

e.g.

```jsx
shouldRedecorate = (nextProps) => {
  return this.props.searchTerms != nextProps.searchTerms; // searchTerms can be a set of search terms
}
render = () => {
  return <Editor needsUpdate={this.shouldRedecorate} />
}
```

Note that `needsUpdate` should only be called once (at editor level) and pass down the decision to all child nodes recursively.

Zhouzi:
We face issues with `shouldComponentUpdate` from times to times. For example, we're currently implementing a way for the user to sort top-level blocks in a document, with drag and drop. The node components need to be passed their index (within their parent, the `document`). While they're properly sorted, the `index` prop is not updated because the component rendering is blocked by `shouldComponentUpdate`.

For example, here's what would do the job for us:

```js
MyBlock.shouldNodeComponentUpdate = (props, nextProps) => {
    if (nextProps.parent !== nextProps.state.document) {
      return false;
    }

    const prevIndex = props.parent.findIndex(
        child => child.key === props.node.key
    );
    const newIndex = nextProps.parent.findIndex(
      child => child.key === nextProps.node.key
    );

    return prevIndex !== newIndex;
};
```

Basically, we'd just want to allow a re-render when the index changes and still benefit from Slate's `shouldComponentUpdate` performance gain. `suppressShouldComponentUpdate` is though as it requires us to duplicate Slate's logic.

This PR introduces a `shouldNodeComponentUpdate` that can be specified just like `suppressShouldComponentUpdate`. The function receives the previous and next properties and can return true to trigger a re-render. If it returns false, the rest of the logic remains.

Also, I added a deprecation warning to `suppressShouldComponentUpdate` although I'm not sure about the version to specify in the `logger.deprecate` call.

Probably fix #959

### DIFF
--- a/docs/reference/slate-react/custom-nodes.md
+++ b/docs/reference/slate-react/custom-nodes.md
@@ -12,6 +12,7 @@ Slate will render custom nodes for [`Block`](../slate/block.md) and [`Inline`](.
   - [`parent`](#parent)
   - [`readOnly`](#readonly)
   - [`state`](#state)
+- [`shouldNodeComponentUpdate`](#shouldnodecomponentupdate)
 
 ## Properties
 
@@ -97,3 +98,17 @@ Whether the editor is in "read-only" mode, where all of the rendering is the sam
 `State`
 
 A reference to the current [`State`](../slate/state.md) of the editor.
+
+## `shouldNodeComponentUpdate`
+
+By default, Slate implements a `shouldComponentUpdate` preventing useless re-renders for node components. While the default implementation covers most use cases, you can customize the logic to fit your needs. For example:
+
+```js
+class CustomNode extends React.Component {
+  static shouldNodeComponentUpdate(previousProps, nextProps) {
+    // return true here to trigger a re-render
+  }
+}
+```
+
+If `shouldNodeComponentUpdate` returns false, Slate will still figure out whether a re-render is needed or not.
--- a/packages/slate-react/src/components/node.js
+++ b/packages/slate-react/src/components/node.js
@@ -4,6 +4,7 @@ import Debug from 'debug'
 import React from 'react'
 import ReactDOM from 'react-dom'
 import SlateTypes from 'slate-prop-types'
+import logger from 'slate-dev-logger'
 import Types from 'prop-types'
 import getWindow from 'get-window'
 
@@ -102,7 +103,16 @@ class Node extends React.Component {
 
     // If the `Component` has enabled suppression of update checking, always
     // return true so that it can deal with update checking itself.
-    if (Component && Component.suppressShouldComponentUpdate) return true
+    if (Component && Component.suppressShouldComponentUpdate) {
+      logger.deprecate('2.2.0', 'The `suppressShouldComponentUpdate` property is deprecated because it led to an important performance loss, use `shouldNodeComponentUpdate` instead.')
+      return true
+    }
+
+    // If the `Component` has a custom logic to determine whether the component
+    // needs to be updated or not, return true if it returns true.
+    // If it returns false, we still want to benefit from the
+    // performance gain of the rest of the logic.
+    if (Component && Component.shouldNodeComponentUpdate && Component.shouldNodeComponentUpdate(p, n)) return true
 
     // If the `readOnly` status has changed, re-render in case there is any
     // user-land logic that depends on it, like nested editable contents.
--- a/packages/slate-react/src/components/node.js
+++ b/packages/slate-react/src/components/node.js
@@ -112,7 +112,13 @@ class Node extends React.Component {
     // needs to be updated or not, return true if it returns true.
     // If it returns false, we still want to benefit from the
     // performance gain of the rest of the logic.
-    if (Component && Component.shouldNodeComponentUpdate && Component.shouldNodeComponentUpdate(p, n)) return true
+    if (
+      Component &&
+      Component.shouldNodeComponentUpdate &&
+      Component.shouldNodeComponentUpdate(p, n)
+    ) {
+      return true
+    }
 
     // If the `readOnly` status has changed, re-render in case there is any
     // user-land logic that depends on it, like nested editable contents.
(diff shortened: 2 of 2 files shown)


AlbertHilb:
> The function receives the previous and next properties and can return true to trigger a re-render. If it returns false, the rest of the logic remains.

Maybe it's even better to call `shouldComponentUpdate` iff `shouldNodeComponentUpdate` returns `undefined`.

SamyPesse:
@AlbertHilb This is the case, the rest of the logic remains if it return a falsy value (`null`, `undefined`, `false`, etc).

Soreine:
Maybe we want the following contract instead?
This gives the most control to the user.

```js
case:
  shouldNodeComponentUpdate -> true
    // The component should be re-rendered, disregarding Slate's shouldComponentUpdate
  shouldNodeComponentUpdate -> false
    // The component should not be re-rendered, __disregarding Slate's shouldComponentUpdate__
  shouldNodeComponentUpdate -> falsy
    // The component should be re-rerendered according to Slate's shouldComponentUpdate
```

AlbertHilb:
I meant exactly what @Soreine stated. :blush:

SamyPesse:
@AlbertHilb @Soreine Interesting, but I can't see any situation where we want to discard Slate's shouldComponentUpdate, do you have something specific in mind ?

Also making a distinction between fasly and `false` may lead to confusions.

Zhouzi:
Completely disabling Slate's `shouldComponentUpdate` seems like a low effort hack to get around something that could more precisely be tackled. If you're having trouble with the default `shouldComponentUpdate`, the correct fix would be to figure out why and then add a specific condition.

But I might miss certain use cases 😊 

As to `false`/`falsy`, I agree with @SamyPesse that it would lead to confusion.

Soreine:
It would cover cases where:

- either you know for sure that your component should not update (no need to call Slate's check)
- or you know that Slate will want to update it needlessly and you want to avoid this.

But I could not think of real world examples. So I'm good with the simpler, initial proposal

Soreine:
I changed my mind. @SamyPesse is right. If we really want to not re-render our component, we still have React's `shouldComponentUpdate`. For the rest, we should let Slate do its internal work and update `<Node>` as needed. Right ?

AlbertHilb:
I don't have any particular use case in mind.
Maybe, an example could be a situation in which you know the descendants of a node don't need to re-render if only their `isSelected` state is changed.
Anyway, also the @Zhouzi issue, perhaps, can be fixed adding an `index` entry to the node `data` without modifying `shouldComponentUpdate` code at all.
If it were up to me, I would leave the code as it is now, but if we decide to give component authors more power I think we should allow them to force update as well as to prevent it.

SamyPesse:
@AlbertHilb 

- Adding an `index` to `data` is dangerous: it requires duplicating an information and keeping it in sync (everytime the node moves, we need to ensure that the `index` data is updated, etc).
- Forcing an update is not dangerous, the only impact can be performance, not the behaviour of Slate.
**But preventing Slate from updating one of its component is dangerous**, it can lead to side effect in Slate (ex: some sync between the dom selection and the selection in state in leafs)

Zhouzi:
There's currently no way to trigger a re-render if Slate has decided not to. On the other hand, as @Soreine said, you can still specify a `shouldComponentUpdate` to prevent unnecessary updates.

Regarding the issues with `index`, it was just one use case where `shouldComponentUpdate` is problematic. But here's another example: we have a block that reads data from its parent to display a label. When the parent's data changes, the child is not updated and thus the label.

Now, if we really want to provide full control over Slate's `shouldComponentUpdate`, we can still provide a way to overwrite it completely. Slate would export its default `shouldComponentUpdate` that we could use without duplicating the code. In the end, it'd look like this:

```js
import { shouldComponentUpdate } from 'slate-react';

class CustomNode extends React.Component {
  static shouldNodeComponentUpdate(prevProps, nextProps) {
    // custom logic

    // and use Slate's shouldComponentUpdate if you want to
    return shouldComponentUpdate(prevProps, nextProps);
  }
}
```

AlbertHilb:
@SamyPesse, @Soreine, @Zhouzi, you convinced me! :blush:

Anyway, I want to add:

1. Preventing update on the Component is not exactly the same as preventing update on the Node: in the first case, for example, the child `Node` elements are constructed, in the second they aren't.

2. Reactiveness is not a minor feature in an editor.

3. I assume whoever plays with `shouldNodeComponentUpdate`, knows what he is doing.

ianstormtaylor:
Hey @Zhouzi (and others) this looks good! I like how much more flexible it is than the `suppressShouldComponentUpdate` hack.

My one concern is that because of our decision to not allow `return false`, which seems like it is a sound decision, our `shouldNodeComponentUpdate` now no longer mirrors React's native one. If someone reads the code they'll expect that `return prevIndex !== newIndex` will either update or not, based only on that logic.

Instead, I think it might be good to throw a warning if `shouldNodeComponentUpdate` return `false`, explaining that you can't disable Slate's internal logic, and to use the regular `shouldComponentUpdate` to prevent extra updates instead. Can you make that change?

Zhouzi:
@ianstormtaylor I just added a warning when returning false in `shouldNodeComponentUpdate` 👍

ianstormtaylor:
Beautiful, thanks @Zhouzi!

AlbertHilb:
Maybe, to avoid confusion, we could rename `shouldNodeComponentUpdate` to something different, like `needsNodeUpdate` so the test becomes:

```
if (Component && Component.needsNodeUpdate && Component.needsNodeUpdate(p, n)) ... 
```

PR #1141, opened 2017-09-19
Mention change in core block normalization

Soreine:
The following change surprised me in some of my tests concerning normalization:

https://github.com/ianstormtaylor/slate/blame/master/packages/slate/src/schemas/core.js#L55-L60

I thought it was worth mentioning in the Changelog.


ianstormtaylor:
Ah good call, I had assumed this already so I totally didn't realize it would be breaking, but definitely is. Thank you @Soreine!

Issue #1129, opened 2017-09-15
Pasting external HTML does not work with Slate example


dmitrizzle:
### Current behaviour

Right now pasting HTML into `PasteHTML` example works **only if I was to copy HTML from Slate editor**:

![slate-to-slate-paste](https://user-images.githubusercontent.com/8587882/30461863-af746de6-99eb-11e7-8cb4-b97dfffe6c8e.gif)

**However, if I am to paste HTML from any other source Chrome 60 throws an error:**

![html-to-slate-paste](https://user-images.githubusercontent.com/8587882/30461888-d26da68c-99eb-11e7-8cf9-f7c936797a6f.gif)

```
build.prod.js:4902 Uncaught TypeError: Cannot read property 'find' of undefined
    at Object.deserialize (build.prod.js:4902)
    at Html.deserializeElement (build.prod.js:98382)
    at build.prod.js:98341
    at Array.forEach (<anonymous>)
    at Html.deserializeElements (build.prod.js:98340)
    at next (build.prod.js:98368)
    at Html.deserializeElement (build.prod.js:98402)
    at build.prod.js:98341
    at Array.forEach (<anonymous>)
    at Html.deserializeElements (build.prod.js:98340)
```

### Expected behaviour

Pasting HTML from any source should work.

### Context
I am using Chrome 60 on macOS 10.12.6, presumably Slate v0.24.1 - since that's the latest version posted (but I can't guarantee that's the same version as what slatejs.org uses in the above examples; above gifs were created shortly after v0.24.1 release was posted).

There's an issue #734 which addresses a similar problem but _I don't think it's relevant anymore_, since it described actions with much older versions of Slate. It was opened mentioning v0.19.16.

### More context
v0.20.7 > v0.21.0 `parse5` replaced `cheerio` but pasting caused JavaScript errors (see #948) which were fixed with #985. I’m not sure if pasting HTML has been tested for with the above PR, since the functionality never returned for pasting form the external sources.

### My understanding
After DOMParser has replaced Cherio in Slate pasting HTML from external sources became broken. Most likely it's because there's no test to check for this case and a bug has slipped in.

ianstormtaylor:
Hey @dmitrizzle thanks for opening this. Can you try with copying a more regularly structured piece of content? Like maybe just a section of a GitHub readme. It could be just a problem with certain bits of HTML in the copy.

travisperson:
A simple example is pasting a div with an inline link.

--------------
[...]

PR #1146, opened 2017-09-20
Pasting external HTML does not work with Slate example. (#1129)

jf423:
*Due to the upgrade of slate, adjust the way of el.attrs.find to el.href.
Since the upgrade of slate, el is now become an NamedNodeMap and not support .find anymore,
so I adjust it with el.href to explicit get the href value.


ianstormtaylor:
Thanks for submitting this @jf423!

PR #1148, opened 2017-09-20
Fix warning with latest react due to autocorrect being bool

bkniffler:
`Received 'true' for non-boolean attribute 'autoCorrect'. If this is expected, cast the value to a string.`

Latest react won't autocast the bool to on/off


ianstormtaylor:
Ah, thanks @bkniffler!

PR #1150, opened 2017-09-20
Move code used to render text nodes into a separate React component

AlbertHilb:
The code used to render text nodes shares almost nothing with the code used to render other nodes.
So move it into a separate React component.
This improves the code readability a lot, many  `if (node.kind === 'text')` are eliminated, and we can make more fine-grained rendering optimization without adding others.
For example, we can safely avoid to re-render a text node if only its `isSelect` state has changed.


ianstormtaylor:
Nice @AlbertHilb, I dig this! Thank you.

PR #1153, opened 2017-09-21
[WIP] Fix copy paste issues

skogsmaskin:
The way it was made it impossible to copy a text beginning with whitespace in Chrome, preserving the leading whitespace.

image

### DIFF
--- a/packages/slate-react/src/plugins/core.js
+++ b/packages/slate-react/src/plugins/core.js
@@ -199,14 +199,17 @@ function Plugin(options = {}) {
       els.forEach(el => el.removeAttribute('contenteditable'))
     }
 
+    let plainText
+
     // Set a `data-slate-fragment` attribute on a non-empty node, so it shows up
     // in the HTML, and can be used for intra-Slate pasting. If it's a text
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
@skogsmaskin I'm getting a Javascript error when trying to run your fix:

image

skogsmaskin:
Yes, there is something wrong there, sorry! Looking into it.

skogsmaskin:
@ianstormtaylor: Like that. It could be prettier, maybe the whole thing should be refactored.

skogsmaskin:
There is more here. Copy pasting a single word don't maintain the marks:


alt text

skogsmaskin:
See also https://github.com/ianstormtaylor/slate/issues/734

ianstormtaylor:
Hey @skogsmaskin thanks!

With the updated code which GIF is this supposed to solve? or both of them?

skogsmaskin:
Just the first for now. I have updated the commit some.

I hope we could fix the select single word as richtext issue also.
It doesn't copy the marks if a single word is selected and copied.

PR #1164, opened 2017-09-22
Fix copy single marked word bug (Chrome/Safari)

skogsmaskin:
If selecting a single marked word and copying it, it will not preserve the marks when pasting. This works fine in Firefox.

Before:

alt text

After:

alt text

### DIFF
--- a/packages/slate-react/src/plugins/core.js
+++ b/packages/slate-react/src/plugins/core.js
@@ -162,7 +162,7 @@ function Plugin(options = {}) {
     const window = getWindow(e.target)
     const native = window.getSelection()
     const { state } = change
-    const { endBlock, endInline } = state
+    const { startKey, endKey, startText, endBlock, endInline } = state
     const isVoidBlock = endBlock && endBlock.isVoid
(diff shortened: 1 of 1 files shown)


danburzo:
I'm getting a bit of open source guilt since I've not been able to make good on my promise to pick up my previous work-in-progress regarding the copy-paste issue in Chrome/Safari.

[The underlying problem](https://github.com/ianstormtaylor/slate/issues/734#issuecomment-324493319) is with how WebKit (Safari/Chrome) does a sort of "collapse" of the selection when you copy things to your clipboard — [I had opened an issue with them](https://bugs.chromium.org/p/chromium/issues/detail?id=758485). I'm curious if this approach circumvents the issue, I'll take a look!

skogsmaskin:
Yeah, that's the problem. I think it is more reasonable the way Firefox does it. Great you have opened issues in the proper places. Hope you survived btw :)

ianstormtaylor:
Sorry for the delay here @skogsmaskin thank you! I'm down to merge this for now and we can see if we can find another workaround or if it gets fixed.

PR #1166, opened 2017-09-22
fixing empty leaft nodes in text getting desynced

### LINKED ISSUE #1159: shouldComponentUpdate strangeness with <br> text nodes
You can find it in the links example, just create an inline in an empty line and observe how the break tag doesn't disappear.  I think this has to do with the parent node not updating right as the inline is inserted, which never allows the text node to see the moment that the inline gets inserted.

YurkaninRyan:
fixes #1159 

Now if the parent node is changed, leaves will make sure that they are rendered correctly, instead of not updating at all.


ianstormtaylor:
Nice, thanks @YurkaninRyan!

PR #1167, opened 2017-09-22
Move `onDragStart` handler from `Node` to `Void`.

AlbertHilb:
`onDragStart` `Node` handler is used only by void nodes, so move it to `Void`.


ianstormtaylor:
Hey @AlbertHilb is this one still relevant? It looks like it might be, if so can you rebase and I can merge it in?

AlbertHilb:
Hi @ianstormtaylor.
Yes, this is still relevant. I rebased.

ianstormtaylor:
Thanks @AlbertHilb!

PR #1172, opened 2017-09-23
[Doc] Remove unused code in Plugins Walkthroughs

renchap:
The `BoldMark` function has been inlined below, it is no longer used here


ianstormtaylor:
Thanks @renchap! Sorry for the delay.

PR #1174, opened 2017-09-23
[Doc] Update installation guide and import declaration in walkthrough.

doodlewind:
From v0.24 slate is maintained as a monorepo with multi packages. This changes the way how user imports slate. Demos are updated while this isn't reflected in walkthrough doc. This PR updates doc contents to keep it in sync. Changes include:

- Installation of `slate` and `slate-react` instead of `slate`.
- Installation of `immutable` as peer dependency.
- Import declaration of `Editor` and `State`.
- Import declaration of `Html` and `Plain`.

### DIFF
--- a/docs/walkthroughs/installing-slate.md
+++ b/docs/walkthroughs/installing-slate.md
@@ -1,16 +1,16 @@
 
 # Installing Slate
 
-Slate is an npm module, so to install it you do:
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Ah thank you @doodlewind!

PR #1180, opened 2017-09-25
Remove depecration for Text.fromJSON({ test })

Soreine:
We would like to revert the deprecation of:

```js
Text.fromJSON({
  text: '...'
})
```

We have a **lot** of tests, and older documents in our database, that are written using this shortcut:

```yaml
document:
  nodes:
    - kind: block
      type: paragraph
      nodes:
        - kind: text
          text: "..."
```

It was a real pain to update the tests. And now it is complicated for us to avoid that warning when we load old documents from our database.

As a result, I am questioning the will to deprecate this, in the first place.

Thoughts?

### DIFF
--- a/packages/slate/src/models/text.js
+++ b/packages/slate/src/models/text.js
@@ -91,7 +91,6 @@ class Text extends Record(DEFAULTS) {
     } = object
 
     if (object.text) {
-      logger.deprecate('0.23.0', 'Passing `object.text` to `Text.fromJSON` has been deprecated, please use `object.ranges` instead.')
       ranges = [{ text: object.text }]
     }
 


ianstormtaylor:
Hey @Soreine, good question.

I'd like to keep `fromJSON` to use the "canonical" representation of nodes, to try to eliminate a bit of the confusion we had earlier around many different JSON representations being serialized/deserialized (using `terse`).

That said, the `.create()` methods are my current solution to getting the more permissive shortcuts that are useful for plugins and testing and such. So if you do `Text.create({ text })` it will actually work just fine.

But I believe there's an issue right now where if you do `Block.create` for instance, the child text nodes will end up using `fromJSON` still, showing the deprecation warning. I'd be open to a PR that fixes this such that `create` calls with nested nodes ensure that all nested objects use `create` too instead of `fromJSON`. Thanks!

Soreine:
:+1: Roger

Merci :)

Le 25 sept. 2017 6:10 PM, "Ian Storm Taylor" <notifications@github.com> a
écrit :

PR #1188, opened 2017-09-27
Add `props.attributes` to the element used to render paragraphs.

AlbertHilb:
In the "Emojis" example, `props.attributes` are not added to the `p` elements used to render paragraphs.
This fixes that.


ianstormtaylor:
Thanks @AlbertHilb!

Issue #1192, opened 2017-09-28
Issue dropping content between two inline void nodes


AlbertHilb:
Open "Emojis" example, insert two emoticons one after the other, drag and drop a word in the space between the two emoticons.

In Firefox, a caret is drawn in that space but the drop does not happen.

In Chrome, the drop does happen but sometimes incorrectly: the dropped word is inserted between the emoji but, right after that word, a duplicate of the first emoticon is inserted too.

Issue #1170, opened 2017-09-23
Drag&Drop issue


AlbertHilb:
Open "Images" example (on Chrome or Firefox), select a word and drag it over the image, the feedback shows a caret on the left side of the image but on drop the word is removed from the original position and nothing is inserted at the image side.

ianstormtaylor:
Hey @AlbertHilb I think this issue is no longer there now, although it seems like dropping on images doesn't work entirely? Either way, I'm going to close this one, but feel free to open more!

PR #1193, opened 2017-09-28
Rewrite drag&drop code related to void nodes.

AlbertHilb:
The state of drag&drop in Slate is far from perfect especially when void nodes are involved (see #1170 and #1192 for examples).

This PR aims to solve some of those problems.

First of all, `onDragEnter`, `onDragLeave` and `onDrop` `Void` handlers are removed . That code is a bit tricky (among other things it temporally changes the editable state of the node content) and we can live without it.

In place of the removed code, a new function, `getDropPoint`, is added to `utils`.
It returns the document point (key and offset) for a drop event.
The point is chosen following the rules below:

1. if the drop event DOM target is inside an inline void node, the point is the first position of the next sibling text node if the event happens closer to the right side of the void node, the last position of the previous sibling text node otherwise.

1. If the drop event DOM target is inside a block void node, the point is the first position of the next sibling block node if the event happens closer to the lower side of the void node, the last position of the previous sibling block node otherwise.

1. If the drop event DOM target isn't inside a void node, the point is the caret position where the drop occurred as in the current code.

`getDropPoint` is used by `onDrop` `Content` handler to get the Slate drop action target.

Since void nodes are now valid DOM drop targets, `isInEditor` checks within `Content` drag&drop handlers are removed.
However, `isInEditor` was used also to avoid managing drag&drop events coming from other editor instances embedded inside the first. To keep that behavior, the new code prevents those events from bubbling up calling `stopPropagation`.

A problem this PR leaves open is the drag feedback effect not in sync with drop action: dragging over a void node, you miss a caret showing where the insert point is, for example.
To fix this, I guess, the only solution is to not let the browser draw the feedback, doing it by ourselves.
On drag over events we could call `getDropPoint` to determine where is the insertion point and then draw a fake caret there.
If you want I can try to add this in a future PR.

Let me know your opinion, I'm open to all changes you believe are needed. :blush:

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -15,6 +15,7 @@ import findClosestNode from '../utils/find-closest-node'
 import getCaretPosition from '../utils/get-caret-position'
 import getHtmlFromNativePaste from '../utils/get-html-from-native-paste'
 import getPoint from '../utils/get-point'
+import getDropPoint from '../utils/get-drop-point'
 import getTransferData from '../utils/get-transfer-data'
 import setTransferData from '../utils/set-transfer-data'
 import scrollToSelection from '../utils/scroll-to-selection'
@@ -369,7 +370,7 @@ class Content extends React.Component {
    */
 
   onDragEnd = (event) => {
-    if (!this.isInEditor(event.target)) return
+    event.stopPropagation()
 
     this.tmp.isDragging = false
     this.tmp.isInternalDrag = null
@@ -384,7 +385,8 @@ class Content extends React.Component {
    */
 
   onDragOver = (event) => {
-    if (!this.isInEditor(event.target)) return
+    event.stopPropagation()
+
     if (this.tmp.isDragging) return
     this.tmp.isDragging = true
     this.tmp.isInternalDrag = false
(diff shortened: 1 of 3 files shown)


oyeanuj:
@AlbertHilb This is great, as it also solves for #980. Just wanted to point out that for `isInEditor`, @ianstormtaylor had earlier suggested breaking it into two different functions: `isInEditor` and `isEditableInEditor`. I started a PR  #1027 for that but I think this PR should subsume that PR and those changes can probably be incorporated in this one - let me know!

> A problem this PR leaves open is the drag feedback effect not in sync with drop action: dragging over a void node, you miss a caret showing where the insert point is, for example.
To fix this, I guess, the only solution is to not let the browser draw the feedback, doing it by ourselves.
On drag over events we could call getDropPoint to determine where is the insertion point and then draw a fake caret there.
If you want I can try to add this in a future PR.

Maybe we could add this to the Image or Embed example?

ianstormtaylor:
@AlbertHilb thank you for this! Sorry for the delay. This PR looks really well done.

I like the idea for the `getDropPoint` util. I'm sure we can tweak little pieces of logic in the future if we realize we need to. Looks good to me!

Question for you that doesn't block the merge: Will the `onDragOver` and `onDragEnter` events in `<Void>` not propagate up to the `<Content>` component? Is that why we need to retain the `e.preventDefault` in them? Otherwise it would be cool to move it all into the `core.js` plugin.

AlbertHilb:
Hi @ianstormtaylor.

Of course, `onDragOver` and `onDragEnter` events bubble up to the `Content` component and they could be handled there. But the default action should be prevented only when the target is inside a `Void` node and deciding, from within `Content`, if a DOM element is inside a void node requires a bit of extra logic and slightly more resources, so for the moment I think it is better to leave those handlers into `Void` component.

About the second question, I think that after the refactoring in #1231, moving `onDrop` logic to `core.js` does not make sense anymore, since `core.js` should not contain code (like `getBoundingClientRect()`) directly interacting with browsers.

ianstormtaylor:
@AlbertHilb actually I think `plugins/core.js` is okay to have browser code. I've been going that way more and more over time. It gets passed an instantiated React component anyways, so those handlers will only ever fire when the components are mounted in the browser.

AlbertHilb:
@ianstormtaylor.

I understand your position, but I think we should distinguish between lower and higher level browser code.
The logic which, using `getBoundingClientRect()`, determines if the target of a drop event is on the left or on the right of an inline void node, for example, is so fundamental that an editor based on Slate can hardly work without it.
However, if that piece of code is moved to `core.js`, then anyone wanting to personalize drop behavior would have to rewrite it, because it would be overridden by his custom drop handler.

ianstormtaylor:
@AlbertHilb I'm not sure. If someone overrides it, there's a decent chance they're doing something completely different, and thus don't need it. (Although I think it would be nice to package up as an exposed utility function.)

As long as there are a few utils exposed, and as long as people are returning early when they don't want to override, I feel like it will be okay.

AlbertHilb:
@ianstormtaylor 

> I'm not sure. If someone overrides it, there's a decent chance they're doing something completely different

For example, one can want to keep the "default" drop target but to insert a rendered math formula if the (external) dropped text is TeX code.

echenley:
Seems like using stopPropagation here leads to some undesirable behavior: https://github.com/ianstormtaylor/slate/issues/1277

PR #1200, opened 2017-10-02
Document `replaceNodeByKey`

conorcussell:
I noticed this method recommended in the changelog for 0.25.0 so this just adds it to the docs. It does take an [optional Options argument](https://github.com/ianstormtaylor/slate/blob/f9894ed443706174085e3317fc5ca3d4e6b69ec8/packages/slate/src/changes/by-key.js#L379) to manage normalization, but I presume thats an internal implementation detail is its been omitted from other methods that feature it too?


ianstormtaylor:
Awesome, yup, thanks @conorcussell!

PR #1202, opened 2017-10-04
fix backward delete of nested inline void node

erquhart:
If you backspace into an inline node, we currently assume that the last child of that node will be text, and do not account for void inline nodes. This leads to an error where Slate is looking for an inline void node by key amongst the text node children of the parent inline.

More simply, here's a fiddle repro: https://jsfiddle.net/erquhart/2zokvrvt/184/

This fix could be further improved by checking arbitrarily nested inlines for voids rather than only checking down one level, but checking one level should cover it since there are no logged issues about this, and it's existed since before 0.22. Settled on this solution for lack of time.

Did not add tests because there are no `at-range` tests, and getting hyperscript working with selections seemed non-trivial. Ensured no tests were broken, and that this solution works in practice.

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -370,10 +370,37 @@ Changes.deleteBackwardAtRange = (change, range, n = 1, options = {}) => {
       return
     }
 
-    // If the previous inline is void, remove it.
-    if (prevInline && prevInline.isVoid) {
-      change.removeNodeByKey(prevInline.key, { normalize })
-      return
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @erquhart, sorry for the delay here. Can you add a test case for this in the `at-current-range` folder? Those will end up testing the `*AtRange` behavior because they all delegate into them.

erquhart:
Hey @ianstormtaylor - sounds good. I'm swamped at the moment, but I will come back to this.

ianstormtaylor:

PR #1205, opened 2017-10-04
Revert the removal of isNative

### LINKED ISSUE #1176: iOS Safari: Predictive text suggestions don't seem to recognize word breaks

bug

### LINKED ISSUE #1177: iOS Safari: Double-tapping the space bar doesn't insert a period

bug

Even when iOS's '"." Shortcut' keyboard setting is enabled, double-tapping the spacebar in Slate doesn't insert a period.

When iOS's '"." Shortcut' keyboard setting is enabled, double-tapping the spacebar should insert a period followed by a space, as it does in a `<textarea>` or a native text field.

rgrove:
Preventing native text input events completely breaks autocorrect, autocomplete, auto-capitalization, predictive text, and double-space period insertion on iOS. In order to allow iOS to perform these kinds of operations, we have to avoid preventing the default event behavior and allow native input handling whenever possible.

This is mostly a reversion of the changes introduced in #1088, but I've tried to maintain the cleanup that was done in `packages/slate-react/src/plugins/core.js`'s `onBeforeInput` method since then.

That method also includes one additional difference, which is the use of `change.deleteCharBackward()` to avoid double input when `isNative` is `true`. Restoring the previously removed `isNative` code throughout the codebase no longer seems to be enough to prevent rendering double input since focus-related state changes cause the `isNative` state to get reset to `false` before rendering happens. Deleting the inserted char seems like a bit of a hacky workaround to me, but it does work, and I wasn't able to come up with a better solution with my limited knowledge of the codebase. I'm open to suggestions.

Fixes #1176
Fixes #1177

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -93,6 +93,32 @@ class Content extends React.Component {
   }
 
   /**
+   * Should the component update?
+   *
+   * @param {Object} props
+   * @return {Boolean}
+   */
+
+  shouldComponentUpdate = (props) => {
+    // If the readOnly state has changed, we need to re-render so that
+    // the cursor will be added or removed again.
+    if (props.readOnly != this.props.readOnly) return true
+
+    // If the state has been changed natively, never re-render, or else we'll
+    // end up duplicating content.
+    if (props.state.isNative) return false
+
+    return (
+      props.className != this.props.className ||
+      props.schema != this.props.schema ||
+      props.autoCorrect != this.props.autoCorrect ||
+      props.spellCheck != this.props.spellCheck ||
+      props.state != this.props.state ||
+      props.style != this.props.style
+    )
+  }
+
+  /**
    * When the editor first mounts in the DOM we need to:
    *
    *   - Update the selection, in case it starts focused.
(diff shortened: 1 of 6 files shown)


packages/slate-react/src/components/content.js
```diff
@@ -93,6 +93,32 @@ class Content extends React.Component {
...
+   */
+
+  shouldComponentUpdate = (props) => {
+    // If the readOnly state has changed, we need to re-render so that
+    // the cursor will be added or removed again.
+    if (props.readOnly != this.props.readOnly) return true
+
+    // If the state has been changed natively, never re-render, or else we'll
+    // end up duplicating content.
+    if (props.state.isNative) return false
```
YurkaninRyan: This never appears to be true.  From what I can tell the change.state.isNative gets set to true, but this doesn't get updated in the props before it checks here.

packages/slate/src/models/state.js
```diff
@@ -21,6 +21,7 @@ const DEFAULTS = {
   selection: Selection.create(),
   history: History.create(),
   data: new Map(),
+  isNative: false,
```
YurkaninRyan: Curious if this should go as a flag on the change object instead along with merge and save

rgrove:
Thanks for the feedback, @YurkaninRyan!

I'm going to close this PR in favor of an alternative approach (discussed in #1177). I'll keep your suggestions in mind as I work on that.

PR #1206, opened 2017-10-05
Fixing getSelectionIndexes

### LINKED ISSUE #1204: Selecting Multiple Blocks results in all nodes receiving an isSelected prop of true.
You can see this behavior on the Links example,

Here's a video showcasing the bug:

http://recordit.co/43nB54MK4T

YurkaninRyan:
Fixes #1204, and perhaps more?

The current behavior of the changed line is opposite of the comment above said line.  Return `false` from a Immutable.List forEach loop causes the forEach to exit early.  I'm pretty sure we only want to return early if both start and end aren't `null`.

Let me know if you see any problems @ianstormtaylor

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1448,7 +1448,7 @@ class Node {
       }
 
       // PERF: exit early if both start and end have been found.
-      return start != null && end != null
+      return start == null || end == null
     })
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
@YurkaninRyan thank you for this! I totally thought it was `return true` that we needed 😄

Issue #1198, opened 2017-09-30
Slate base64 serializer and SSR


bkniffler:
Hey, 
I tried using the base64 serializer with my SSR enabled nodejs app and an error was thrown since `window` is used. Also, `atob` and `btoa` isn't available in node.

I could do a PR, but how would you like this solved? We could polyfill these two methods directly
```
function atob(str) {
  return new Buffer(str, 'base64').toString('binary');
}
function btoa(str) {
  let buffer;

  if (str instanceof Buffer) {
    buffer = str;
  } else {
    buffer = new Buffer(str.toString(), 'binary');
  }

  return buffer.toString('base64');
}
```
or make it a requirement that a global polyfill for these is available.  Either way, we should get rid of the window references.

ianstormtaylor:
@bkniffler good call. I think we should just have a `if` statement that checks if the window methods are defined, and tries to use them, or falls back to `Buffer.from()`, or errors out if there is no implementation to use. I'd love a PR on this!

isubastiCadmus:
@ianstormtaylor atob/btoa conversion doesn't really work out of the box for non ASCII character such as emoticons.

ianstormtaylor:
@isubastiCadmus what happens with it? what bug does it create?

isubastiCadmus:
@ianstormtaylor if the unicode is bigger than U+00FF, it will errors out see [btoa](https://html.spec.whatwg.org/multipage/webappapis.html#dom-windowbase64-btoa)

ianstormtaylor:
@isubastiCadmus do you have a situation that causes that error?

isubastiCadmus:
@ianstormtaylor looking at the code for base-64-serializer, you're already doing window.encodeURIComponent and window.decodeURIComponent which solves the atob/btoa error on emoji so all good. should have checked the source code before making comments, my bad. I can make a PR on this.

PR #1208, opened 2017-10-06
Slate base64 serializer and SSR

isubastiCadmus:
This provides a polyfill for btoa and atob on base64 serializer for SSR
fixes:  https://github.com/ianstormtaylor/slate/issues/1198

### DIFF
--- a/packages/slate-base64-serializer/src/index.js
+++ b/packages/slate-base64-serializer/src/index.js
@@ -1,13 +1,11 @@
 
 import { State } from 'slate'
-import { Buffer } from 'buffer'
 
-function newBuffer(data, encoding, len) {
+function bufferFrom(data, encoding, len) {
+  if (Buffer.from) {
+    return Buffer.from(data, encoding, len)
+  }
   return new Buffer(data, encoding, len)
-}
-
-if (!Buffer.from) {
-  Buffer.from = newBuffer
 }
 
 /**
@@ -19,14 +17,14 @@ if (!Buffer.from) {
 
 function btoa(str) {
   let buffer
-  if (window && window.btoa) {
+  if (typeof window !== 'undefined' && window.btoa) {
     return window.btoa(str)
   }
 
   if (str instanceof Buffer) {
     buffer = str
   } else {
-    buffer = Buffer.from(str)
+    buffer = bufferFrom(str)
   }
   return buffer.toString('base64')
 }
@@ -39,10 +37,11 @@ function btoa(str) {
  */
 
 function atob(str) {
-  if (window && window.atob) {
+  let buffer
+  if (typeof window !== 'undefined'  && window.atob) {
     return window.atob(str)
   }
-  return Buffer.from(str, 'base64').toString('binary')
+  return bufferFrom(str, 'base64').toString('binary')
 }
 
 /**
--- a/packages/slate-base64-serializer/src/index.js
+++ b/packages/slate-base64-serializer/src/index.js
@@ -2,6 +2,14 @@
 import { State } from 'slate'
 import { Buffer } from 'buffer';
 
+function newBuffer(data, encoding, len) {
+  return new Buffer(data, encoding, len);
+}
+
+if (!Buffer.from) {
+  Buffer.from = newBuffer;
+}
+
 /**
  * encode string `str` to base64
  * @param  {String} str
--- a/packages/slate-base64-serializer/src/index.js
+++ b/packages/slate-base64-serializer/src/index.js
@@ -1,22 +1,24 @@
 
 import { State } from 'slate'
-import { Buffer } from 'buffer';
+import { Buffer } from 'buffer'
 
 function newBuffer(data, encoding, len) {
-  return new Buffer(data, encoding, len);
+  return new Buffer(data, encoding, len)
 }
 
 if (!Buffer.from) {
-  Buffer.from = newBuffer;
+  Buffer.from = newBuffer
 }
 
 /**
  * encode string `str` to base64
+ *
  * @param  {String} str
  * @return {String}
  */
+
 function btoa(str) {
-  let buffer;
+  let buffer
   if (window && window.btoa) {
     return window.btoa(str)
   }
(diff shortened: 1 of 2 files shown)


packages/slate-base64-serializer/src/index.js
```diff
@@ -1,5 +1,49 @@
 
 import { State } from 'slate'
+import { Buffer } from 'buffer'
```
bkniffler: where does this package come from?

packages/slate-base64-serializer/src/index.js
```diff
@@ -1,5 +1,49 @@
...
+
+function newBuffer(data, encoding, len) {
+  return new Buffer(data, encoding, len)
+}
+
+if (!Buffer.from) {
+  Buffer.from = newBuffer
```
bkniffler: Monkey-patching anything could potentially be bad

packages/slate-base64-serializer/src/index.js
```diff
@@ -1,5 +1,49 @@
...
+/**
+ * encode string `str` to base64
+ *
+ * @param  {String} str
+ * @return {String}
+ */
+
+function btoa(str) {
+  let buffer
+  if (window && window.btoa) {
```
bkniffler: will throw on server since window is undefined. better ignore window and use `typeof btoa !== 'undefined'`

packages/slate-base64-serializer/src/index.js
```diff
@@ -10,7 +54,7 @@ import { State } from 'slate'
 
 function encode(object) {
   const string = JSON.stringify(object)
-  const encoded = window.btoa(window.encodeURIComponent(string))
+  const encoded = btoa(window.encodeURIComponent(string))
```
bkniffler: should remove window reference and use the globally defined `encodeURIComponent` directly, should work on node and browser

packages/slate-base64-serializer/src/index.js
```diff
@@ -22,7 +66,7 @@ function encode(object) {
  */
 
 function decode(string) {
-  const decoded = window.decodeURIComponent(window.atob(string))
+  const decoded = window.decodeURIComponent(atob(string))
```
bkniffler: same as above

packages/slate-base64-serializer/src/index.js
```diff
@@ -1,5 +1,49 @@
...
+
+/**
+ * decode back base64-encoded string `str`
+ *
+ * @param  {String} str
+ * @return {String}
+ */
+
+function atob(str) {
+  if (window && window.atob) {
```
bkniffler: same as above

bkniffler:
fyi, here is a lib trying to do similar: https://github.com/dchest/tweetnacl-util-js/blob/master/nacl-util.js (outputs  Uint8Array instead of strings)

packages/slate-base64-serializer/src/index.js
isubastiCadmus: this is nodejs's built in Buffer

packages/slate-base64-serializer/src/index.js
isubastiCadmus: this is only necessary for node before v4 as Buffer.from is only available from v4 onwards

packages/slate-base64-serializer/src/index.js
isubastiCadmus: it won't throw an error since it checks if window is there or not as well. this way window.btoa(client side btoa) will still work but it it doesn't exist then use Buffer conversion

packages/slate-base64-serializer/src/index.js
isubastiCadmus: yeah totally missed that window call

packages/slate-base64-serializer/src/index.js
isubastiCadmus: I've used this import for both client and server side and it works.
Is there any scenario that you've found where this import doesn't work?

bkniffler: I just can't imagine it is a good idea to import node specific modules to your client bundle, especially if the Browser already defines the imported stuff. Did you check what happens with your client bundle in terms of size?

We could easily have a check for typeof Buffer === 'undefined' and only require the node buffer module in that case.

packages/slate-base64-serializer/src/index.js
bkniffler: Yes, but you are extending the Buffer object and people might run into edge cases. Imagine people are checking for Buffer.from somewhere else to see if they are running an old version of node.

packages/slate-base64-serializer/src/index.js
bkniffler: If you check for window reference on nodejs, you will run into a ReferenceError since window is not defined. Just try for yourself and run this through node:
```js
if (!window) { console.log('hi') } // => will throw
if (typeof window === 'undefined') { console.log('hi') } // => will work
```

packages/slate-base64-serializer/src/index.js
isubastiCadmus: fairPoint, technically speaking this import can be removed since nodejs have Buffer defined globally

packages/slate-base64-serializer/src/index.js
isubastiCadmus: fair point, I'll remove the monkey patching and use it without monkey patching

packages/slate-base64-serializer/src/index.js
bkniffler: Whoops, thats true 👍

packages/slate-base64-serializer/src/index.js
isubastiCadmus: good pickup, didn't know that window check throw an error on server side

ianstormtaylor:
Hey @isubastiCadmus thanks for looking into this! Could you change this to use the [`isomorphic-base64`](https://github.com/ksheedlo/isomorphic-base64) package instead actually? That way we don't have to maintain this code into the future. Thank you!

isubastiCadmus:
@ianstormtaylor I'll have a look tomorrow or on the weekends

PR #1210, opened 2017-10-08
Remove `display: inline-block` from void blocks

dmitrizzle:
As implied by the name itself, "blocks" shouldn't be expected to be displayed as inline elements.

Aside from that, styled `<div/>` wrappers can change the behaviour of nested elements without giving any API controls to the developer. For example, if the editor contains a `<figure/>` element that is expected to float left of the text, `display:inline-block` property will overwrite that behaviour, which can not be fixed with CSS:
***
You can see from the code below that the `<figure/>` element is prescribed to float left, but that doesn't work:
Broken behaviour, text shifted down
***
Expected behaviour is to have the image float left and the text wrap around it on the right, which can only work if the wrapping `<div/>` isn't displayed as `inline-block`:
Expected behaviour, text wraps around image
***

**This tiny PR simply removes the CSS rule and restores expected behaviour.**

### DIFF
--- a/packages/slate-react/src/components/void.js
+++ b/packages/slate-react/src/components/void.js
@@ -143,7 +143,6 @@ class Void extends React.Component {
     if (node.kind === 'block') {
       Tag = 'div'
       style = {
-        display: 'inline-block',
         verticalAlign: 'top',
         width: '100%'
       }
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Hey @dmitrizzle thanks for opening this, that makes total sense. I fixed it slightly differently just now on `master`. Thanks!

dmitrizzle:
Thank you very much @ianstormtaylor - I think your commit will also fix https://github.com/dmitrizzle/Analog.Cafe/issues/621 - an issue that showed unexpected single-line spacer in FireFox only. 🍻

PR #1212, opened 2017-10-09
Sanitize tagName check

davidosomething:
`In XML (and XML-based languages such as XHTML), tagName preserves case. On HTML elements in DOM trees flagged as HTML documents, tagName returns the element name in the uppercase form.`


ianstormtaylor:
Thanks @davidosomething, good catch!

PR #1214, opened 2017-10-10
Fix broken link to setKeyGenerator

Zhouzi:
Just noticed a 404 when clicking "setKeyGenerator" in the "key" property description here: https://docs.slatejs.org/slate-core/node

This PR fixes the path to the utils.md file.


ianstormtaylor:
Thanks @Zhouzi!

Issue #893, opened 2017-06-20
Decoration updates breaks selection


Soreine:
In the case where the decorations of a text node depend on something external to this text node, the decorations may be updated as the selection enters this node. This breaks the selection update logic.

# Reproducing

Let's take this live example: https://soreine.github.io/slate-prism/

Steps to reproduce:

1. We will edit the last code block:
   ```html
   <!-- Some HTML -->
   <b>Hello World</b>
   ```
2. Delete the end of the comment tag
   ```html
   <!-- Some HTML
   <b>Hello World</b>
   ```
3. Point the cursor at the end of the second line `<b>Hello World</b>|`
4. The error is thrown
  ```
  bundle.js:77822 Uncaught TypeError: Cannot read property 'text' of undefined
    at findBounds (https://soreine.github.io/slate-prism/bundle.js:77822:23)
    at Object.findPoint (https://soreine.github.io/slate-prism/bundle.js:77882:21)
    at getPoint (https://soreine.github.io/slate-prism/bundle.js:76917:35)
    at onSelect (https://soreine.github.io/slate-prism/bundle.js:61544:45)
    at Object.ReactErrorUtils.invokeGuardedCallback (https://soreine.github.io/slate-prism/bundle.js:47800:16)
    at executeDispatch (https://soreine.github.io/slate-prism/bundle.js:41728:21)
    at Object.executeDispatchesInOrder (https://soreine.github.io/slate-prism/bundle.js:41751:5)
    at executeDispatchesAndRelease (https://soreine.github.io/slate-prism/bundle.js:41155:22)
    at executeDispatchesAndReleaseTopLevel (https://soreine.github.io/slate-prism/bundle.js:41166:10)
    at Array.forEach (native)
  ```
5. If you start typing, you will probably encounter weird bugs because of an out of sync selection

# Explanation

This example showcases a decorator for code highlighting. Each code block is made of several code lines (for performance in case of large code blocks).
The decorator is applied to every code line, but takes into account the whole text of the parent code block (to support multi-line comments for example).
When you mess with the HTML comment on the first line, it will affect the result of the decoration on the second line. However, the decoration of the second line is not yet updated.
[...]

PR #1221, opened 2017-10-12
refactor decorations to use selections

ianstormtaylor:
This changes how decorations are applied in Slate.

Previously, you would use a `decorate(text, block)` function in the schema. It would be called for each `text` node in the block, and you would add the marks to the characters yourself. But this led to issues because there are cases (like multi-line comments in syntax highlighting) where the decoration applies across multiple text nodes, and this system wasn't really built to handle that well.

Now, you use a `decorate(block)` function in the schema. It is no longer called for each of the text nodes. Instead of applying the new marks to the characters yourself, you return an array of selection objects, with associated marks. For example:

```js
{
  match: (object) => {
    return object.type == 'code_block',
  },
  decorate: (block) => {
    const text = block.getFirstText()

    return [{
      anchorKey: text.key,
      anchorOffset: 0,
      focusKey: text.key,
      focusOffset: 1,
      marks: [{ type: 'bold' }]
    }]
  }
}
```

This would turn add a `bold` mark to the first character of the matched blocks.

This would also make implementing things like multiple-cursor rendering easier, since the "decorations" now map exactly to the existing selections/ranges logic.

cc @Soreine, how does this sound? I think it should solve https://github.com/ianstormtaylor/slate/issues/893 and https://github.com/GitbookIO/slate-prism/pull/5.


lxcid:
This is awesome!

I do have question about does it improve dynamic decorator? When force rerender based on state not directly related to the editor (e.g. search result highlighting)

ianstormtaylor:
Hey @lxcid good question!

This PR itself doesn't support that, because the `decorate` logic in the schema is still based on node in the document themselves. 

But! It does get us closer to being able to do things like that. I was thinking of adding a `state.decorations` property and a `change.setDecorations(decorations)` method that would allow you do to that. That way you could change the decorations in the editor as the outside information changes (eg. someone performs a search) and then things will re-render.

I haven't tested the performance of doing that for large editors, since you'll be re-decorating the entire document at once, but I'm sure there are things we can do to improve it.

packages/slate/src/models/node.js
```diff
@@ -165,19 +165,12 @@ class Node {
     first = normalizeKey(first)
     second = normalizeKey(second)
 
-    let sorted
+    const keys = this.getKeysAsArray()
```
Soreine: Oh, here is the optimization :D

packages/slate/src/models/schema.js
```diff
@@ -126,24 +127,33 @@ class Schema extends Record(DEFAULTS) {
...
-          return rule.decorate(text, object)
-        }
+  __getDecorations(object) {
+    const array = []
+
+    this.rules.forEach((rule) => {
+      if (!rule.decorate) return
+      if (!rule.match(object)) return
+
+      const decorations = rule.decorate(object)
```
Soreine: So the `decorate` calls themselves now live under a memoized function, which is nice :)

packages/slate-react/src/components/text.js
```diff
@@ -95,10 +90,19 @@ class Text extends React.Component {
...
+    const { decorations, node, state } = props
     const { document } = state
-    const decorators = schema.hasDecorators ? document.getDescendantDecorators(node.key, schema) : []
-    const ranges = node.getRanges(decorators)
+    const { key } = node
+
+    const decs = decorations.filter((d) => {
+      const { startKey, endKey } = d
+      if (startKey == key || endKey == key) return true
+      const startsBefore = document.areDescendantsSorted(startKey, key)
```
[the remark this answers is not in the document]
ianstormtaylor: I agree though, this can be improved.

One improvement I think would be a method that checks `(before, key, after)` or something about checking whether a node "overlaps" a range. Because here we have to do two separate checks which each iterate keys, etc. when really it could be handled in one go.

packages/slate-react/src/plugins/core.js
```diff
@@ -82,8 +81,8 @@ function Plugin(options = {}) {
     // the selection has gotten out of sync, and adjust it if so. (03/18/2017)
     const window = getWindow(e.target)
     const native = window.getSelection()
-    const a = getPoint(native.anchorNode, native.anchorOffset, state, editor)
-    const f = getPoint(native.focusNode, native.focusOffset, state, editor)
+    const a = findPoint(native.anchorNode, native.anchorOffset, state)
+    const f = findPoint(native.focusNode, native.focusOffset, state)
```
ianstormtaylor: Actually any place where we can doing `findPoint` for both anchor and focus can be slightly improved by checking if `collapsed` and if so only doing it once. But I'm not sure this is a bottleneck, so leaving for later.

packages/slate/src/models/node.js
```diff
@@ -958,18 +925,29 @@ class Node {
...
   /**
-   * Return a set of all keys in the node.
+   * Return a set of all keys in the node as an array.
    *
-   * @return {Set<String>}
+   * @return {Array<String>}
    */
 
-  getKeys() {
+  getKeysAsArray() {
```
ianstormtaylor: I think this method can also be improved in performance if it was recursive, such that each nested child's keys would be cached, instead of having to iterate the entire tree each time the document changes.

ianstormtaylor:
Actually @lxcid I've just implemented that behavior that you're asking about! I added `state.decorations`, which can be set based on any outside information. Along with a simple `search-highlighting` example to show you how it works.

lxcid:
Wow! Thank for considering my request @ianstormtaylor and getting it in so soon! I can't wait to use it soon! The demo looks great!

PR #1225, opened 2017-10-13
Some fixes

davidosomething:
Ensure el has a tagName (e.g. textnode will not have a tagName) - found in slack channel

Strict equality checks for strings


ianstormtaylor:
Thanks @davidosomething!

PR #1230, opened 2017-10-14
Fix hovering menu

winfield:
Try to fix hovering menu example on slatejs.org. It uses the new portal api, so it only works with React v16+. I am not sure whether it's a problem with examples.


ianstormtaylor:
Thank you @winfield this is awesome! No problem at all that the examples depend on React 16. I love being able to get rid of `react-portal`. Thanks!

PR #1232, opened 2017-10-15
Use native `beforeinput` events to handle text insertion when possible

### LINKED ISSUE #1176: iOS Safari: Predictive text suggestions don't seem to recognize word breaks

### LINKED ISSUE #1177: iOS Safari: Double-tapping the space bar doesn't insert a period

rgrove:
In browsers that support it (currently only Safari has full support), the native `beforeinput` DOM event provides much more useful information about text insertion than React's synthetic `onBeforeInput` event.

By handling text insertion with the native event instead of the synthetic event when possible, we can fully support autocorrect, spellcheck replacements, and related functionality on iOS without resorting to hacks.

See the discussion in #1177 for more background on this change.

Note: Chrome has partial support for `beforeinput`, but doesn't allow the event to be canceled and doesn't reliably provide information about the affected range. Possibly because of this, Chrome doesn't report the existence of the `beforeinput` event, so it fails the feature test for that event and will still use the synthetic `onBeforeInput` code path.

I've tested this change extensively in macOS Safari, iOS Safari, Chrome, and Firefox. Don't have access to Edge at the moment, but I plan to test it next week (I expect it to fail the event feature test and use the synthetic `onBeforeInput` code path). More testing would be appreciated! I'm especially curious how this works on Android Chrome, but I don't have a device to test with right now.

Fixes #1176
Fixes #1177

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -19,7 +19,7 @@ import getHtmlFromNativePaste from '../utils/get-html-from-native-paste'
 import getTransferData from '../utils/get-transfer-data'
 import scrollToSelection from '../utils/scroll-to-selection'
 import setTransferData from '../utils/set-transfer-data'
-import { IS_FIREFOX, IS_MAC, IS_IE } from '../constants/environment'
+import { IS_FIREFOX, IS_MAC, IS_IE, SUPPORTED_EVENTS } from '../constants/environment'
 
 /**
  * Debug.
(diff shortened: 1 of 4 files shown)


renchap:
@rgrove you can easily test for Edge on BrowserStack, Microsoft sponsored them to offer it for free: https://www.browserstack.com/test-on-microsoft-edge-browser#live-cloud

They also offer the service to OSS projects, it might be a good idea to contact them for this (I can do it if you want). Selenium testing might not be useful for now (but it would be great) but their live mode allows you to launch pretty much every browser/OS combinaison in 1 click which would help a lot to ensure Slate is compatible with the majors browsers

rgrove:
Fixed a lint error and resolved recently introduced conflicts with `master`. Also tested Edge and confirmed that it continues to use the synthetic `onBeforeInput` code path.

ianstormtaylor:
Thank you @rgrove this is a really great PR, and awesome research/testing. Thank you very much! Hopefully more browsers are quick to implement this and it will begin solving more mobile issues.

Issue #1229, opened 2017-10-14
Chinese IME throws out errors


winfield:
#### What's the current behavior?

Using Chinese IME on slatejs.org throws out errors

````
Error: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at removeChild (http://slatejs.org/build.prod.js:4828:7974)
    at a (http://slatejs.org/build.prod.js:4825:25233)
    at commitDeletion (http://slatejs.org/build.prod.js:4825:26981)
    at n (http://slatejs.org/build.prod.js:4825:31488)
    at u (http://slatejs.org/build.prod.js:4826:1356)
    at c (http://slatejs.org/build.prod.js:4826:1793)
    at batchedUpdates (http://slatejs.org/build.prod.js:4826:4655)
    at qb (http://slatejs.org/build.prod.js:4825:3879)
    at ob (http://slatejs.org/build.prod.js:4825:3813)
    at Object.batchedUpdates (http://slatejs.org/build.prod.js:4826:19682)
`````
![1](https://user-images.githubusercontent.com/6850/31575362-e442a440-b0aa-11e7-82b8-362a87154d84.gif)

````
TypeError: Cannot read property 'node' of null
    at t.n.updateSelection (http://slatejs.org/build.prod.js:5014:4115)
    at t.n.componentDidUpdate (http://slatejs.org/build.prod.js:5014:3374)
    at commitLifeCycles (http://slatejs.org/build.prod.js:4825:27709)
    at n (http://slatejs.org/build.prod.js:4825:31692)
    at u (http://slatejs.org/build.prod.js:4826:1356)
    at c (http://slatejs.org/build.prod.js:4826:1793)
    at batchedUpdates (http://slatejs.org/build.prod.js:4826:4655)
    at qb (http://slatejs.org/build.prod.js:4825:3879)
    at ob (http://slatejs.org/build.prod.js:4825:3813)
    at Object.batchedUpdates (http://slatejs.org/build.prod.js:4826:19682)
````
![2](https://user-images.githubusercontent.com/6850/31575363-e7f71210-b0aa-11e7-80cd-5d8cd01408f4.gif)

OS: macOS High Sierra
Browser: Chrome version 61
Slate version: slatejs.org

doodlewind:
This error is only triggered with macOS native pinyin IME, third party pinyin IME (Sougou, Baidu, etc.) does not throw this error, instead they 'eat' characters afterwards.

[...]

Issue #1209, opened 2017-10-07
 Error when using Chinese Input Method  解决办法


liuningnb:
```
onChange = ({ state }) => {
	if (state.document !== this.state.state.document) {
		const string = html.serialize(state);
		localStorage.setItem('content', string)
	}
	
	const input = new Promise( (resolve,reject) => {
		setTimeout(()=>{
			if(!this.state.isPinyin){
				resolve();
			}
		},10);
	});
	
	input
	.then(()=>{
		this.setState({ state })
	});
	
};
```
添加监听判断中文输入:
```
onCompositionStart={()=>{
	this.setState({
		isPinyin:true
	});
}}
onCompositionEnd={()=>{
	this.setState({
		isPinyin:false
	});
}}
```

doodlewind:
Does `isPinyin` indicator accurate enough? onCompositionStart does not only works with Chinese pinyin IME, and on OSX, Sougou and native pinyin IME behaves differently...

ThomasChan:
for anyone still have this problem

in your render
```javascript
render() {
  // ...
  {this.state.forceRerender ? null : <Editor {...} />}
  // ...
}
```

```javascript
onChange = next => {
  if (!this.__isPinYin && this.__dirtyCheckEnd < 1e3) {
    const { value } = next;
    this.setState({
      value,
      forceRerender: false,
    });
    this.__dirtyCheckEnd = 0;
  } else {
    clearTimeout(this.__pinYinDelay);
    this.__pinYinDelay = setTimeout(() => {
      this.__dirtyCheckEnd += 1;
      clearTimeout(this.__pinYinDelay);
      this.setState({ forceRerender: true }, () => {
        this.onChange(next);
      });
    });
  }
}

onCompositionStart = () => {
  this.__isPinYin = true;
  this.__dirtyCheckEnd = 0;
}

onCompositionEnd = () => {
  this.__isPinYin = false;
}
```

PR #1236, opened 2017-10-16
[IME] Disable iOS COMPAT behaviour on desktop

doodlewind:
This PR closes a series of desktop IME issues: #1229 #1209 #854 #810 .
It also closes PR #885 #880 .

All issues mentions above is caused by a workaround introduced in #655 , which adds a piece of iOS COMPAT code fixing autocorrect behaviour. In this workaround, `onBeforeInput ` will adjust slate's selection **if selection is expanded**, resolve the mismatch.

However, during desktop IME composition, the selection is also expanded. This leads to deletion of unrelated text, in other word, 'eating' characters afterwards:

### Current Behaviour
before

Previous solution adds composition-state-related `data` for `onBeforeInput`, this works but may introduce more potential issues. In fact, by simply limiting the iOS COMPAT code according to environment, this issue can be solved in a much simpler way. Moreover, it also benefits performance by reducing unnecessary DOM ops like `window.getSelection()` on other platforms.

This patch not only solves Chinese Pinyin IME issues, for other languages it also works:

### After: Native Pinyin IME
after-native-pinyin

### After: Thiry Party Pinyin IME
after-third-party-pinyin

### After: Native Japanese IME
after-native-japanese

### After: Native Korean IME
after-native-korean

BTW, another IME related bug can be solved based on this PR with React 16's error boundary, I may submit another PR if this one can get merged 💪

### DIFF
--- a/packages/slate-react/src/constants/environment.js
+++ b/packages/slate-react/src/constants/environment.js
@@ -76,5 +76,6 @@ export const IS_FIREFOX = BROWSER === 'firefox'
 export const IS_SAFARI = BROWSER === 'safari'
 export const IS_IE = BROWSER === 'ie'
 
+export const IS_IOS = OS === 'ios'
 export const IS_MAC = OS === 'macos'
 export const IS_WINDOWS = OS === 'windows'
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thank you @doodlewind, this is awesome. Thanks for the very detailed PR!

lxcid:
Hey @doodlewind! Thanks for the fix! Sorry for introducing the bug!

Issue #1224, opened 2017-10-13
Chinese IME fails on Microsoft Edge


J22Melody:
I know some known IME issues such as #854 and I can solve them by patch PRs like #1107 and #880, but in Microsoft Edge, Chinese IMEs like Microsoft Pinyin still behave wierd.

Here is my example in Slate 0.25.0 Microsoft Edge 20 Windows 10 with Microsoft Pinyin:

![untitled](https://user-images.githubusercontent.com/2316987/31532200-631df800-afb1-11e7-8e22-60fec999ed6f.gif)

I'm not sure if it is related or it's another issue only appears in Microsoft Edge. Does anyone encounter the same problem. Any workaround?

By the way, slatejs.org does not work in Microsoft Edge, which is already mentioned in #1186.

Issue #967, opened 2017-08-02
[IME] Failed to execute 'removeChild' on 'Node'


pizn:
The current behavior?

![removechild](https://user-images.githubusercontent.com/986480/28872520-26ea7598-77bc-11e7-81f6-0ecf9e448a20.gif)

---

macOS + Chrome 60 

1. Select two lines
2. Use IME input some word
3. Enter and the console will throw error

--- 

with #854

chemzqm:
#885 should fix this, instead of #880 , I hope it could get merged.

Soreine:
Can reproduce using dead keys, on Chrome (not Firefox).

1. Switch to a French keyboard layout
2. Type twice the ` character at the start of the document (under the `\` key on US layouts).
3. Slate crashes

Soreine:
Possibly the same issue: https://github.com/ianstormtaylor/slate/issues/1229

ianstormtaylor:
Closing in favor of https://github.com/ianstormtaylor/slate/issues/2368.

Issue #1237, opened 2017-10-16
Uncaught DOMException: Failed to execute 'getRangeAt' on 'Selection': 0 is not a valid index.


dmitrizzle:
#### Bug report/question
It's a bug, or I might have made an error in my implementation during upgrade to slate@0.27.0 and slate-react@0.4.0

Tested on Chrome 61 macOS 10.12.6 with React16.

#### Current behaviour:

In my implementation everything works fine in the editor (I'm ignoring depreciation warnings for now) until I add an image to the document.
![err](https://user-images.githubusercontent.com/8587882/31597981-6f4dc4c2-b275-11e7-82ad-4c304899b94b.gif)

This is the code snippet that handles inserting images (done when clicking on one of the images in the grid) to the document:
```javascript
handleImageSuggestion = src => {
    const { state, editor, node } = this.props
    const resolvedState = state
      .change()
      .insertBlock({
        type: "image",
        isVoid: true,
        data: { src }
      })
      // remove docket
      .state.change()
      .removeNodeByKey(node.key)
    this.suggestionHandlerTimeout = setTimeout(() => {
      editor.onChange(resolvedState)
    }, 10)
  }
```

#### Error stack printout:
```
content.js:167 Uncaught DOMException: Failed to execute 'getRangeAt' on 'Selection': 0 is not a valid index.
    at Content._this.updateSelection (http://localhost:3000/static/js/12.chunk.js:21531:28)
    at Content._this.componentDidUpdate (http://localhost:3000/static/js/12.chunk.js:21504:13)
    at commitLifeCycles (http://localhost:3000/static/js/bundle.js:27565:24)
    at commitAllLifeCycles (http://localhost:3000/static/js/bundle.js:28342:9)
    at HTMLUnknownElement.callCallback (http://localhost:3000/static/js/bundle.js:17347:14)
    at Object.invokeGuardedCallbackDev (http://localhost:3000/static/js/bundle.js:17386:16)
    at invokeGuardedCallback (http://localhost:3000/static/js/bundle.js:17243:27)
    at commitAllWork (http://localhost:3000/static/js/bundle.js:28463:9)
    at workLoop (http://localhost:3000/static/js/bundle.js:28735:13)
    at HTMLUnknownElement.callCallback (http://localhost:3000/static/js/bundle.js:17347:14)
    at Object.invokeGuardedCallbackDev (http://localhost:3000/static/js/bundle.js:17386:16)
[...]

PR #1238, opened 2017-10-16
Recover from fatal error with React 16's error boundary

### LINKED ISSUE #1114: Error replacing multi blocks with IME composition inputs
Bug.

When replacing multi blocks text with IME composition inputs, an `fail to removeChild` error raised, and the IME result is also lost, only remains temp keys during composition. Reproduction: 

1. Select multi blocks.
2. Use IME to input some texts.
3. Choose a word.

compositon-bug

In this case, `啊啊啊` is the expected result instead of `a'a'a`.

Text input from IME inserted correctly without error.

doodlewind:
This PR closes #1114 #967 and maybe #1224 .

Currently there are certain cases which causes editor to break. It happens when environment break React's vDOM against DOM. Draft.js also [suffers from this](https://draftjs.org/docs/advanced-topics-issues-and-pitfalls.html#browser-plugins-extensions):

> As with any React application, browser plugins and extensions that modify the DOM can cause Draft editors to break.
> 
> Grammar checkers, for instance, may modify the DOM within contentEditable elements, adding styles like underlines and backgrounds. Since React cannot reconcile the DOM if the browser does not match its expectations, the editor state may fail to remain in sync with the DOM.
> 
> Certain old ad blockers are also known to break the native DOM Selection API -- a bad idea no matter what! -- and since Draft depends on this API to maintain controlled selection state, this can cause trouble for editor interaction.

This is somewhat inevitable with Slate's architecture, but the good news is that we have React 16's error boundary as a fallback now. For example, when native selection selects across multi blocks, texting with IME will remove selected DOM, which can't be cancelled. In this case a fatal error is raised, user can't even continue editing:

### Current Behaviour
error-handle-before

As a solution, this PR utilizes the `componentDidCatch` hook to recover from current state. Although error is still raised, editing can now seamlessly continued:

### Patched with Error Boundary
error-handle-after

Besides, this hook does not affect React 15's behaviour, so it won't break backward compatibility.

### DIFF
--- a/packages/slate-react/src/components/editor.js
+++ b/packages/slate-react/src/components/editor.js
@@ -193,6 +193,16 @@ class Editor extends React.Component {
   }
 
   /**
+   * When vDOM and DOM are out of sync, recover from current state.
+   */
+
+  componentDidCatch = () => {
+    requestAnimationFrame(() => {
+      this.componentWillReceiveProps(this.props)
+    })
+  }
+
+  /**
    * Cache a `state` object to be able to compare against it later.
    *
    * @param {State} state


ianstormtaylor:
Hey @doodlewind this is interesting, I didn't even know that was possible! Do you have any links to React folks talking about this as an okay solution?

I'm a bit worried that it would be swallowing all of the errors, meaning that people couldn't choose to handle them how they'd like. Especially in cases where the error shouldn't be swallowed. 

For #1237 it's actually a bug in the way native DOM selections are handled by `<Content>`. So we can fix that one like that. For the IME case, would there be a way to actually interpret the delete first, so that the editor's state isn't out of sync?

renchap:
Error boundaries are new in React 16, here is the introduction blog post: https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html

ianstormtaylor:
Thanks @renchap, to be more clear: I'm wondering if any React people advocate for doing this kind of thing where you basically ignore the error. It seems like it defeats the purpose of error boundaries, so I'm worried about it.

doodlewind:
@ianstormtaylor Error boundary is newly introduced and its best practice is somewhat unclear indeed. Even Draft.js does not use this feature for now.

Still do I believe that error handling makes sense in a framework sophisticated like slate, since throwing all errors to user may be intimidating. Another approach I can think about is to introduce a "default" error handling behaviour in core plugin as a fallback, making it possible for users to customize it. How about this idea? Currently all editor content is dynamically rendered as children of `<Editor>`, so here can be the bottom-most component to add unified error handling logic.

BTW, I'm still looking for other approaches to fix the IME bug, may PR it later if a more decent workaround can be found 🤖

doodlewind:
I found that `componentDidCatch` can also be handled in user's component, so this one seems to be a bit overkill. May open another one when React has an official best practice for error handling 😶

PR #1256, opened 2017-10-18
Mention schema rules as a valid place to make changes

alanctkc:
Not sure if this was an intentional omission in the `Change` guide for simplicity (understandable), but thought I'd add a block mentioning schema rules as a valid place to make changes.

Also, happy to have tiny nitpicks on individual words/tone/whatever, just roughed it in to see if it was worth adding. :)


ianstormtaylor:
Ah, thanks @alanctkc! I totally forgot about schema rules 😄

PR #1262, opened 2017-10-19
remove rendering from schema & make it expressive

### LINKED ISSUE #735: define `isVoid` behavior in the schema itself
Hey @ianstormtaylor! While working on embeds, I noticed a particular pattern emerging within my code. That led me to a question/proposal around how blocks are declared in the schema. It is still fairly high level but wanted to put it out there to start discussion (and make sure I am not missing anything in my understanding). 

---

**1. Declaring isVoid in the schema instead of block properties**

I noticed that for certain blocks, especially for `void` ones, I kept adding that wherever I needed to insert or toggle a block.
```js
const embedContent = Block.create({
  type:	embedBlock.data.get('embedType'),
  isVoid: true
});
```

Now I know, that the above problem can be mitigated by having one function creating the block and calling that function everywhere, or by having a normalization rule that ensures that it is always `void`.

But it seems to me that `isVoid` information is something that is not likely to change in a schema, and hence should be defined there (alongside `render` or `decorate` in a `node`).

---

**2. Nested block structure in the schema**

Another pattern that has emerged in different cases (lists, embed) is that a block might need to **always** have a particular structure as its `children`. 
```yaml
Embed
  - EmbedContent
  - EmbedInfo
    -- EmbedTitle
    -- EmbedAuthor
```
Now it feels a little weird to enforce or insert this structure everytime one inserts or toggles to an embed block. It takes a few `insertBlock` and `wrapBlock` operations to do so. To be fair, this can again be abstracted into a function and reused, or be enforced through schema validation and normalization rules as you have in some examples. 

But again, it made me wonder if the more appropriate place might be the schema declaration (especially, when the structure shouldn't change). 

This one is more complicated than the previous one (especially around flexibility with nested structures), and I am aware that the schema declaration is also translated into rules, so this might be more about it being a more readable and shorthand for writing schema rules.

Thanks for your time, would love to hear your thoughts!

### LINKED ISSUE #799: add schema rule naming validation
To allevoate problems like #796 where it's easy to get the name wrong and have nothing work.

### LINKED ISSUE #1111: consider adding `schema` into `state` objects
I can't find the other place this was discussed right now, but wanted to open this for discussion.

---

Right now there are two schemas: the "core" one, and the user-land one. 

The core one is used internally to normalize the document after any change method are run, and never exposed to the user. It ensures that the core rules of the document model are preserved and the user never touches a document that doesn't have them in effect. (Things like block nodes can't have inlines, etc.)

Then there is another schema, that is passed into the `<Editor>`, and can be added to by plugins. This schema is normalized using the `onBeforeChange` handled of the "core" plugin.

---

A few issues from this current setup:

- The `onBeforeChange` handler is kind of awkward, in the if the `schema` was part of `state` it wouldn't even need to exist. And it also is leaky in that it only normalizes after the other `on*` handlers are run, meaning that it doesn't guarantee that a developer never sees an invalid document (according to their schema).

- Since the user-land schema isn't part of `state`, we can't normalize against it automatically in the change methods, because we don't have access to the `editor` instance there.

- If you are running changes outside the editor, and just passing them back in via `setState`, then you actually aren't normalizing against the user-land schema for these changes. This is unexpected, and probably is causing issues that people aren't aware of?

So those would be solved.

But it would introduce a few new problems to find solutions for:

- If it the `schema` was part of `state`, then it would be more complex for plugins to augment it. Right now you can deal with the `state` in server-side environments, or in places where you don't have the editor with all of its plugins. If `schema` needed to be inside `state`, then anywhere you were working with a document you'd need access to all of the plugins that the editor uses when working with that type of document? This seems very problematic.

- If the `schema` is part  of `state`, and a normalization happens automatically when you create the state, how do you ensure that any changes to a state are always observable by the parent? This is required for collaborative situations, where any operations that occur must be logged.

---

I would love thoughts from others on this, either for or against, or with ideas for how to solve some of these problems.

### LINKED ISSUE #1258: expressive schemas
One of the trickiest parts about using Slate in more complex situations, especially once nested blocks are concerned, is ensuring that your document is in a "normalized" state with respect to your schema.

Since the schema is completely customizable, Slate doesn't help you by default. You have to add logic to ensure that states you don't want are normalized away.

A few examples of the kind of things you might want to enforce...

#### Quotes
- `quote` blocks always contain block nodes.
- `quote` blocks always contain at least on `paragraph`.

#### Lists
- `list` blocks contain only `list_item` blocks.
- `list_item` blocks contain only `paragraph` blocks.

#### Images
- `image` blocks are always wrapped in a `figure` block.
- `image` blocks always have a `data.src` attribute.
- `figure` blocks contain one `image` and an optional `figure_caption` block.
- `figure_caption` blocks only ever appear inside `figure` blocks.

#### Code
- `code` blocks contain only `code_line` blocks.
- `code` blocks always have a `data.language` attribute.
- `code_line` blocks contain only text nodes.
- Text nodes in `code_line` blocks have no marks.

#### Links
- `link` inlines always have a `data.href` attribute.

#### Comments
- `comment` marks always have a `data.author` attribute.

---

Right now you can do this via schema rules, by defining a set of `match, validate, normalize` functions that will correct any invalid states, like so:

```js
{
  match(obj) {
    return obj.kind == 'block' && obj.type == 'quote',
  },
  validate(quote) {
    const invalidChildren = quote.nodes.filter(n => n.kind != 'block')
    if (!invalidChildren.size) return
    return invalidChildren
  },
  normalize(change, quote, invalidChildren) {
    invalidChildren.forEach((node) => {
      change.removeNodeByKey(node.key)
    })
  },
}
```

But while this gives you the maximum flexibility, it's hard to read and write. It's hard to tell if you're leaving certain cases unhandled.

There's [prior art from Prosemirror](https://prosemirror.net/docs/guide/#schema) of having schemas that are more expressive.

I'd like to see if we can find something that Slate core could provide that could make managing schemas easier for people. Ideally it would solve the 90% cases well, and leave the underlying function-based approach for the other 10%.

---

There are a few hard challenges...

One issue is just of the complexity in allowing for all of the "common" schema-based restrictions. You should be able to... restrict the `kind` of child nodes, restrict the number and types of child nodes, restrict marks, restrict data properties, etc.

That said, although it will be more to learn, I do think that 90% of cases can be handled by _not too many_ keywords/syntaxes. So although it'll take research to figure them out, this should be solveable.

Another problem is that the `normalize` function of schema validations can often be context-specific, to be the least destructive possible. For example, in the case of...

> `figure_caption` blocks only ever appear inside `figure` blocks.

...if you see a rogue `figure_caption` at the top-level of the document, you could remove it. But instead, it you converted it to a `paragraph`, you'd avoid accidentally deleting content. Like so:

```js
{
  match: (object) => {
    return object.kind == 'block' && object.type != 'figure'
  },
  validate: (block) => {
    const invalids = block.nodes.filter(n => n.type == 'figure_caption')
      return invalids.size ? invalids : null
    },
  normalize: (change, block, invalids) => {
    invalids.forEach(n => change.setNodeByKey(n.key, 'paragraph'))
  }
},
```

But how do you get the expressive schema definitions to understand these kinds of complexities and perform the least destructive normalizations? I'm not exactly sure here. One idea could be exposing the `reason` that a schema rule failed, and maybe the normalizations are always left up the user. Or maybe the user can choose to opt-in to more fine-grained normalizing logic if the default doesn't suit them.

---

I'd love to hear people's ideas about what a nice, expressive schema API might look like!

### LINKED ISSUE #1260: splitting the "behavior/rendering" logic out of schemas
This was brought up by @Soreine and @SamyPesse in #1111. I want to open a new issue for it so that the discussions can not get derailed in each place though.

I feel like there's a sort of Catch-22  scenario occurring between

- **This issue** — splitting the "behavior/rendering" logic out of schemas
- **#1111** — consider adding `schema` into `state` objects
- **#1258** — expressive schemas

We'd like to put the schema into `state`, but to do that it would be best to separate it into two different pieces so that only the structural logic is in `state`. 

But to split it into structural pieces, we sort of want schemas to be more expressive otherwise it will be hard to convince plugins like `edit-list/code/etc` to stop defining schemas inside the plugin itself. 

But to make schemas more expressive, we kinda of need to see past the behavioral/rendering logic that is clumped in there.

---

So here, I just want to present a way to split out the behavioral/rendering logic from schemas, so that others can have this in mind when discussing those other issues. And maybe we can unblock this piece while figuring out the other two.

The parts we need to account for removing at:

- `render` for rendering nodes.
- `placeholder` for rendering placeholders.
- `decorate` for adding decorations

We actually already have some "rendering" logic in the plugins in the form of:

- `plugin.render` for rendering around the editor, adding toolbars, etc.
- `plugin.renderPortal` for rendering extra portals.

And we also used to have `plugin.renderNode` and `plugin.renderMark` methods instead of the current "schema" concept. When the schema was created, we ported them there to have a single place to manage them.

But I think we did that too soon, without realizing that having the schema be structure-only would be valuable. So I think we could go back to that system (with a slight tweak, for people who remember that)...

---

What we'd end up with is a few new plugin properties:

- `renderNode(props: Object) => ReactElements|Void`
- `renderMark(props: Object) => ReactElements|Void`
- `renderPlaceholder(props: Object) => ReactElements|Void`

These are actually very similar to `react-router`'s `Route` `render` property. They take in a `props` object, and return react elements to be rendered. It would look like this:

```js
plugin.renderNode = (props) => {
  if (props.node.type == 'paragraph') {
    return <MyParagraphComponent {...props} />
  }
}
```

Such that "converting" the existing schema would look like:

```js
plugin.renderNode = (props) => {
  switch (props.node.type) {
    case 'paragraph': return <Paragraph {...props} />
    case 'quote': return <Quote {...props} />
    case 'image': return <Image {....props} />
    ...
  }
}
```

Pretty simple, but powerful too. And I think it would feel more React-ish. And it would allow people to pass in extra props to node components, just like you can with `react-router`. (As long as you are aware of when node's rerender and when they don't.)

Rendering marks and rendering placeholders would work in the same way. For example, here's what the "current" default placeholder could look like...

```js
core.renderPlaceholder = (props) => {
  const { node, state } = props
  if (node.kind != 'block') return
  if (!Text.isTextList(node.nodes)) return
  if (node.text != '') return
  if (state.document.getBlocks().size > 1) return

  const style = {
    pointerEvents: 'none',
    display: 'inline-block',
    width: '0',
    maxWidth: '100%',
    whiteSpace: 'nowrap',
    opacity: '0.333',
  }

  return (
    <span contentEditable={false} style={style}>
      {editor.props.placeholder}
    </span>
  )
}
```

I think it's actually easier to understand how/why this works than the current implementation.

And then we can solve `decorate` similarly, by adding a `plugin.decorateNode` function that gets called with `(node)` as the current one does. This should be fairly straightforward.

---

The only thing this approach leaves up in the air, is how the `NodeComponent.shouldNodeComponentUpdate` functionality would be handled. Since Slate no longer has a reference to your custom node component, it can't just read the static property from it. But I think we might be able to solve this with another plugin method:

```js
plugin.shouldNodeComponentUpdate(props, nextProps) {
  ...
}
```

Which would be called at the same time as currently.

---

This would completely pull all of the behavioral/rendering logic out of the schema, so that schemas can be purely structural. And in doing so, I feel like it kind of becomes even more React-ish as a result.

Would love to hear people's thoughts on this!

ianstormtaylor:
This is the first stab at #1260, just to get the idea going. It's working with the `rich-text`, `plain-text` and `code-highlighting` examples.

You can see the [diff for the `rich-text`](https://github.com/ianstormtaylor/slate/pull/1262/files#diff-9638d80b1bc63ebe2eca582da542b7e4) example to get an idea of how the changes work.

---

###### BREAKING

- **The `Schema` objects in Slate have changed!** Previously, they used to be where you could define normalization rules, define rendering rules, and define decoration rules. This was overloaded, and made other improvements hard. Now, rendering and decorating is done via the newly added plugin functions (`renderNode`, `renderMark`, `decorateNode`). And validation is done either via the lower-level `validateNode` plugin function, or via the new `schema` objects.

- **The `plugin.onBeforeChange` function was removed.** Previously there was both an `onBeforeChange` handler and an `onChange` handler. Now there is just an `onChange` handler, and the core plugin adds it's own logic before others.

- **The `normalize*` change methods no longer take a `schema` argument.** Previously you had to maintain a reference to your schema, and pass it into the normalize methods when you called them. Since `State` objects now have an embedded `state.schema` property, this is no longer needed.

###### NEW

- **`State` objects now have an embedded `state.schema` property.** This new schema property is used to automatically normalize the state as it changes, according to the editor's current schema. This makes normalization much easier.

- **A new `renderNode` plugin function was added.** This is the new way to render nodes, instead of using the schema. Any plugin can define a `renderNode(props)` function which is passed the props to render the custom node component with. This is similar to `react-router`'s `render={...}` prop if you are familiar with that.

- **A new `renderPlaceholder` plugin function was added.** This is similar to the `renderNode` helper, except for rendering placeholders.

- **A new `decorateNode` plugin function was added.** This is similar to the old `rule.decorate` function from schemas. Any plugin can define a `decorateNode(node)` function and that can return extra decoration ranges of marks to apply to the document.

- **A new `validateNode` plugin function was added.** This is the new way to do specific, custom validations. (There's also the new schema, which is the easier way to do most common validations.) Any plugin can define a `validateNode(node)` function that will be called to ensure nodes are valid. If they are valid, the function should return nothing. Otherwise, it should return a change function that normalizes the node to make it valid again.

---

Fixes #735 
Fixes #799 
Fixes #1111 
Fixes #1258 
Fixes #1260


ianstormtaylor:
@SamyPesse @Soreine @Zhouzi just added a few commits, I've got the first stab at an expressive schema here. You can check out the [diff for the `forced-layout`](https://github.com/ianstormtaylor/slate/pull/1262/files#diff-633ca03f9e39cd074d00835ef5d53065) example to see it in action. (Also, commented out there is the equivalent `validateNode` approach.) Would love to hear your thoughts!

ianstormtaylor:
One of my biggest worries is just whether the schema will be flexible enough, yet still know enough to provide useful normalizations. If we decide it can't, then maybe the normalizing itself would all be left to the user, but with a much more targeted `reason` supplied to make it easier to decide what went wrong. Not sure.

packages/slate/src/models/schema.js
```diff
@@ -88,225 +98,444 @@ class Schema extends Record(DEFAULTS) {
...
+      const children = Node.createList(nodes || [])
+      node.nodes.forEach(child => change.removeNodeByKey(child.key, { normalize: false }))
+      children.forEach((child, i) => change.insertNodeByKey(node.key, i, child))
+      logger.debug('Normalized by replacing with default children.', { node, children })
+    }
+
+    function insertDefaultChild(index) {
+      const { nodes } = rule.defaults
+      if (!nodes) return removeNode()
+      const def = nodes[index]
```
Soreine: I'm a bit lost here. `index` is too generic, I don't know if it refers to the index of a child node, or some index in a rule's array. It is used to retrieve the definition for the default node, and for inserting a new child as well.

Soreine:
Are you sure you want to make expressive schemas in this PR too ? I understand that life is too short not to do it now, but this seems risky 😄 

It's looking good though. The rendering separation is neat, the `render` functions feel simpler. However, same as you, I'm worried about the expressive schema, `normalization` especially.

---

Concerning `reasons`, you were able to cover all the cases of normalization allowed by the expressive schema. So we can be confident it's viable.

---

Concerning the expressiveness, maybe we are missing something. At this point, the expressive schema does not allow to define proper rules for a node's parent.

I would like to say that:


- `quotes` are container blocks, they can contain any content block, including `lists` containers defined by another plugin
   => I specify that `quotes` can only contain `block` kind.

- `list-items` must live inside a `list` container. If I find a `list-item` in a container block, such as a document or a `quote`, but not a `heading`, I want to be able to wrap it in a `list`.
  => I can't express that in the expressive schema, or I need to write a custom global-aware normalization for every container block.

The `list-item` case is widespread. In our application, this has become a performance bottleneck. We have several complex blocks, that are composed of specific sub-blocks, that can only live in the appropriate container. Lists have such a structure with a depth of 1 (list -> item). Tables have a depth of 2 (table -> row -> cell). And we have a complex block to represent an API endpoint, whose depth is around 5.

Maybe we can't generalize `parent node` rules to be just `children nodes` rules on the containers. When I encounter a `list-item` in a `document`, the behavior should be defined by rules on`list-item`, not global rules on `document`.

Here is an idea of how it could look:

```js
const schema = {
    document: {
        // No specific rules for children here
        defaults: {
            nodes: ['paragraph']
        }
    },

    blocks: {
        quote: {
            nodes: [
                // Quotes are generic containers, just like document
                { kind: 'block' }
            ]
        },

        list: {
            nodes: [
                { type: 'list-item' }
            ]
        },

        "list-item": {
            nodes: [
                // Items are generic containers
                { kind: 'block' }
            ],
            parent: [
                { type: 'list' }
            ],

            normalize: (change, node, reason, ...args) => {
                // Specific handling for invalid parent
                if (reason == 'node_invalid_parent') {
                    const [ parent ] = args
                    if (parent.kind == 'block') {
                        // Wrap the item in a list container.
                        return change.wrapNodeByKey(...)
                    }
                }
                // Otherwise, default handling
                return null
            }
        }
    }
}
```

Soreine:
Am I right that the "split rendering from schema" is a breaking change, but not the addition of the "expressive schema" ? If yes, you should really do it in two steps.

ianstormtaylor:
@Soreine awesome, thank you for that feedback!

> Am I right that the "split rendering from schema" is a breaking change, but not the addition of the "expressive schema" ? If yes, you should really do it in two steps.

It could be done as a non-breaking change, but I'm particularly interested in how changing up the ways the schema is defined can simplify some other parts of the system, and potentially even get us to add `state.schema`. So because they're so inter-related, I think I'm going to do it all in one go. Usually I'd agree with you and try to make things deprecation-only, then breaking. But for this it's already hard enough to arrive at a good API, I want to have room to break as needed when thinking it through. Not the best excuse, but... 😄  Appreciate you checking though.

> Concerning the expressiveness, maybe we are missing something. At this point, the expressive schema does not allow to define proper rules for a node's parent.

Good call, I now understand why those are separate cases. In the latest commits I've added two new validation abilities:

```js
block: {
  parent: ['list'],
}
```

For validating the parent node as you rightly point out. 

```js
block: {
  text: /^[^\n]$/m,
}
```

For validating the text content of a node. This can be useful for example in the case of code line blocks being required not to contain newlines, because they should instead be split into multiple lines.

---

I'm struggling with another piece of "magic"...

Right now there's `rule.nodes`, `rule.parent`, `rule.text`, etc. But often the provide the "best" normalization automatically, which is the goal ideally, we need to know what the "default" the user expects the text to be, so we can change it to that. 

For example, for nodes:

```js
quote: {
  nodes: [{ kind: ['block'] }],
}
```

We can't know for example what to wrap any inline nodes we encounter with. The user would prefer that inlines are automatically wrapped in a `paragraph` block. So we have to ask them to write instead:

```js
quote: {
  nodes: [{ kind: ['block'] }],
  default: {
    nodes: [{ type: 'paragraph', kind: 'block' }],
  }
}
```

It's just starting to feel a bit complex when you have to do all that. There are some cases where the default could be "smartly inferred", by taking the first `kind[0]` and `type[0]` from the nodes definition. But that feels like it starts to get really magic-y and hard to reason about what normalizations will actually happen...

If we instead go with the approach that a "validation" and a "default" must just always be declared together, to reduce the magic, while maintaining the ability to provide the most useful normalizations by default, we might end up with something like:

```js
const schema = {
  blocks: {
    list: {
      nodes: {
        validate: [{ type: ['item'] }],
        default: [{ kind: 'block', type: 'item' }],
      }
    },
    item: {
      nodes: {
        validate: [{ type: ['paragraph'] }],
        default: [{ kind: 'block', type: 'paragraph' }],
      },
      parent: {
        validate: { type: ['list'] },
        default: { kind: 'block', type: 'list' },
      }
    }
  }
```

Where every validation takes a `validate` and a `default` for normalizing. 

We could make slightly more convenient by treating "types" as shorthands...

```js
const schema = {
  blocks: {
    list: {
      nodes: {
        validate: ['item'],
        default: ['item'],
      }
    },
    item: {
      nodes: {
        validate: ['paragraph'],
        default: ['paragraph'],
      },
      parent: {
        validate: 'list',
        default: 'list',
      }
    }
  }
```

But then we're back to the, "why not just infer it?", starting point.

Any thoughtS?

ianstormtaylor:
Okay, I've added more commits...

I opted to solve all of the magic-y, defaults-based issues by removing the concept of defaults completely, and only have the built-in rules be the most basic, usually more-destructive-than-you-want normalizations.

For example, consider the `list > item` case...

By default, if you only define the validation for `item.parent`, if an item is encountered outside of the parent, the default normalization logic will remove the item, since it is invalid. Like so:

```js
const schema = {
  blocks: {
    list: {},
    item: {
      parent: { types: ['list'] },
    }
  }
}

const input = (
  <state>
    <document>
      <paragraph>
        <item />
      </paragraph>
    </document>
  </state>
)

const output = (
  <state>
    <document>
      <paragraph />
    </document>
  </state>
)
```

Notice how in the `output` the `<item>` is removed entirely.

But if you want to actually try to "save" the information in the document by providing a less destructive normalization behavior, you have to supply a `normalize` function. Like so:

```js
const schema = {
  blocks: {
    list: {},
    item: {
      parent: { types: ['list'] },
      normalize: (change, reason, { node }) => {
        if (reason == 'parent_type_invalid') {
          change.wrapBlockByKey(node.key, 'list')
        }
      }
    }
  }
}

const input = (
  <state>
    <document>
      <paragraph>
        <item />
      </paragraph>
    </document>
  </state>
)

const output = (
  <state>
    <document>
      <paragraph>
        <list>
          <item />
        </list>
      </paragraph>
    </document>
  </state>
)
```

Now in the `output` the lone item has been wrapped in a `<list>` instead.

---

I think this provides a nice balance of expressiveness, while still maintaining sanity when try to understand how things will be normalized.

It still provides expressiveness for the validations, which is the hardest (and most performance-sensitive) pieces of code to write over and over. And it removes the "shorthands" that get confusing to parse, so you can quickly read exactly which kinds of `reasons` you'll even have to encounter while validating.

In most cases, you're probably going to want to define your own normalization functions, so that you can have full control over the how the replacements are made. But I think this is actually a good thing in the end, because it makes the normalizations more understandable, since you can read them directly. Instead of them being a more complex mix of core, defaults and overrides.

ianstormtaylor:
Okay, now this PR also includes embedding the newly simplified, and structure-only `Schema` objects into `State` objects under `state.schema`.

Because of this, the normalization happens automatically, the same way that normalizing with the core schema used to happen. And now the `normalize*` change methods use the internal schema instead of accepting an external one.

I've also removed `onBeforeChange`, and handled it all via `onChange` instead, to simplify the two, since we have the before plugin to do distinguish between before/after instead.

Soreine:
It looks great. I haven't checked the code since last time. Will the custom normalize functions be able to handle some reasons, and return void for the rest to let the default normalization happen ?

ianstormtaylor:
@Soreine yup! The schema logic will check to see if any new `operations` have been created on the `change` object, and if none have then the default logic will kick in, so you only have to define normalizations for what you want, and the defaults will handle the rest!

PR #1264, opened 2017-10-19
Fix IE11 error

istarkov:
`createNodeIterator` does not work at IE11 if you provide less than 4 arguments.

Also to make slate work with IE11 currently I use next hack (otherwise you cant focus editor)

```javascript
<Editor
  ref={ref => { ref_ =  ref; }}
  onFocus={() => { ref_.focus() }}
  ...
```

Without additional focus call as like as above, for unknown reason in IE11 this line is always called 
https://github.com/ianstormtaylor/slate/blob/6c42f6c9c30f6874ff1f20a916df36e922ef2f24/packages/slate-react/src/components/content.js#L137

Everything other looks like works fine.

### DIFF
--- a/packages/slate-react/src/utils/find-dom-point.js
+++ b/packages/slate-react/src/utils/find-dom-point.js
@@ -15,14 +15,17 @@ import findDOMNode from './find-dom-node'
 function findDOMPoint(key, offset) {
   const el = findDOMNode(key)
   const window = getWindow(el)
+  let start = 0
+  let n
+  
+  // COMPAT: In IE, this method's arguments are not optional, so we have to 
+  // pass in all four even though the last two are defaults. (2017/10/25)
   const iterator = window.document.createNodeIterator(
     el,
     NodeFilter.SHOW_TEXT,
     () => NodeFilter.FILTER_ACCEPT,
     false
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @istarkov sorry for the delay here. Do you have a picture of the error that IE11 generates? I just want to see what we're dealing with for future reference before I merge. Thanks!

istarkov:
Hey, here is ie 11 debugger error
image

[here](https://developer.mozilla.org/en-US/docs/Web/API/Document/createNodeIterator) is also information about 4th parameter (which is also not optional in ie 11)

```
Note: Prior to Gecko 12.0 (Firefox 12.0 / Thunderbird 12.0 / SeaMonkey 2.9), this method accepted an optional fourth parameter (entityReferenceExpansion) that is not part of the DOM4 specification, and has therefore been removed. This parameter indicated whether or not the children of entity reference nodes were visible to the iterator. Since such nodes were never created in browsers, this paramater had no effect.
```

ianstormtaylor:
Beautiful, thanks @istarkov!

istarkov:
Thank you, amazing library!!!

Issue #891, opened 2017-06-18
Checkbox Example Inconsistent Across Browsers


Slapbox:
http://slatejs.org/#/check-lists

Firefox works properly, but IE and Chrome do not.

## On pressing enter to create a new checkbox line item (with cursor at the end of a line):

**Chrome:** Creates the checkbox. Cursor disappears and focus is not on the new checkbox

**Internet Explorer:** Creates the checkbox. Cursor moves to the new line, but focus does not actually move.

**Firefox:** Creates the checkbox. Cursor moves to the new line and focus moves as well
____
- There are similar issues on IE and Chrome on backspacing. Here too, Firefox works fine.
- These issues are not present if you press enter in the middle of text in a checkbox, such that the new checkbox is immediately populated with some text.

AlbertHilb:
This is related to the `onBlur`issue described in #853.

danburzo:
I just want to add that Firefox/macOS also seems to have some problems with clicking on the text for the check items — it randomly selects text or makes the cursor jump to the beginning of the content area.

PR #1265, opened 2017-10-19
Reshape `Void` component.

### LINKED ISSUE #1223: can't escape void blocks in Firefox on down arrow key
Because of https://github.com/ianstormtaylor/slate/commit/c36362a4253fdb5fe0a39137b8bf9c88b77fbba2 (which I think was the right call) the old behavior which has a bug in Firefox is back. If you press the down arrow key while selecting a void block, nothing will happen.

AlbertHilb:
Reshape `Void` component and prevent editor blur in some circumstances.

This fixes #1223 and #891 (at least on Chrome, Edge needs some extra work).

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -228,9 +228,16 @@ class Content extends React.Component {
       this.tmp.key++
     }
 
-    // If the `onSelect` handler fires while the `isUpdatingSelection` flag is
-    // set it's a result of updating the selection manually, so skip it.
-    if (handler == 'onSelect' && this.tmp.isUpdatingSelection) {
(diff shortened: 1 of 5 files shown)


ianstormtaylor:
Thank you @AlbertHilb, this is amazing as usual.

And also thanks for submitting all of these different void fixes over time. It's kind of insane how many weird ways there are to get this to work in varying combinations of uses cases! 😄 

This looks great in my testing in Firefox and Chrome. Seems like we have an existing issue with Safari, which I just tracked down to the `draggable` attribute being the culprit. I think I'll solve this by moving the `draggable` to the inner `<Tag>` here.

Thanks!

Slapbox:
@AlbertHilb... I think I love you... Thank you so much for this.

PR #1268, opened 2017-10-20
Making the default for Text.create to have leaves

YurkaninRyan:


ianstormtaylor:
Thanks @YurkaninRyan!

PR #1272, opened 2017-10-20
Save marks when creating native ranges in onNativeBeforeInput

### LINKED ISSUE #1270: active marks aren't working in safari?

Bug.

Pressing <kbd>cmd-b</kbd> in the `rich-text` example doesn't set active marks on the next character in Safari:

It should! Just like Chrome, etc.

justinweiss:
When we used native ranges in onNativeBeforeInput, we weren't carrying
over the marks from Slate's selection. This made it impossible to know
that the next character typed should have a set of marks.

Fixes #1270.

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -304,14 +304,15 @@ class Content extends React.Component {
     event.preventDefault()
 
     const { editor, state } = this.props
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Good call, thank you @justinweiss!

PR #1273, opened 2017-10-21
Keep marks after splitting a block

### LINKED ISSUE #1269: support carrying over marks between paragraphs

Bug.

If you split a block, the "active" marks aren't carried over:

When you press enter, any active marks should stay active.

justinweiss:
If you have, for example, a bold mark, and you hit "Enter", the new
block won't carry over the formatting. Instead, it should (at least if
you start typing right away).

Fixes #1269.

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -474,9 +474,15 @@ function AfterPlugin(options = {}) {
     const { state } = change
 
     if (HOTKEYS.SPLIT_BLOCK(event)) {
-      return state.isInVoid
-        ? change.collapseToStartOfNextText()
-        : change.splitBlock()
+      if (state.isInVoid) {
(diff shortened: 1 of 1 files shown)


justinweiss:
I like the flexibility of doing this at the plugin level, but it means anyone who overrides the Enter keybinding will have to reimplement it if they want the same behavior. 

There's also still some things that seem off to me with this approach:

* If you move the cursor and bring it back, you lose the marks
* If you delete to the beginning of the line, you lose the marks
* If you select all the text in a block and delete it, you lose the marks

I'm sure there are others, but that's what I've found so far!

ianstormtaylor:
Thanks @justinweiss! This looks good.

I hear you on the issues, although I think this is a good improvement for now, so I'm going to merge it and we can choose to potentially handle those things as another issue. Thank you!

PR #1274, opened 2017-10-23
Removed `data` argument from simulator event handlers

ashutoshrishi:
In `slate-react@0.6.0`: The data argument to event handlers has been removed...

The event handlers generated in the slate-simulator `Simalator` object was
still passing a data argument. This data argument got passed as the `change` argument to the plugin handlers because of:

```javascript
  handle(handler, change, editor, ...args) {
    debug(handler)

    for (let k = 0; k < this.plugins.length; k++) {
      const plugin = this.plugins[k]
      if (!plugin[handler]) continue
      const next = plugin[handler](...args, change, editor)
      if (next != null) break
    }
  }
```

which was called as `handle(handler, change, editor, event, data)`.

### DIFF
--- a/packages/slate-simulator/src/index.js
+++ b/packages/slate-simulator/src/index.js
@@ -1,4 +1,3 @@
-
 import { Stack } from 'slate'
 
 /**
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Good call @ashutoshrishi, thank you for catching that!

Issue #1277, opened 2017-10-23
Drag events are swallowed by Slate


echenley:
#### Do you want to request a *feature* or report a *bug*?

Bug.

#### What's the current behavior?

https://jsfiddle.net/6n439eh9/1/

Drag events do not propagate outside of Slate. This was introduced in https://github.com/ianstormtaylor/slate/pull/1193 (cc @AlbertHilb). While the onDragEnter/Leave/Over hooks work on the Editor component itself, any drag hooks outside of the editor will not be called. This makes using something like ReactDropzone alongside Slate difficult if not impossible.

@ianstormtaylor mentioned in slack that it should be possible to do this without using stopPropagation: https://slate-js.slack.com/archives/C1RH7AXSS/p1508433948000130

#### What's the expected behavior?

Drag events should propagate normally.

Tested using slate-react version 0.7.0 (introduced in 0.6.0).

ianstormtaylor:
Fixed now!

PR #1278, opened 2017-10-24
Don't stop propagation of drag events.

AlbertHilb:
Don't stop propagation of drag events.
Fix issue #1277.

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -252,6 +252,20 @@ class Content extends React.Component {
       }
     }
 
+    // Don't handle drag events coming from embedded editors.
+    if (
+      handler == 'onDragEnd' ||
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Beautiful, thank you @AlbertHilb! This is great.

lxcid:
Hey @AlbertHilb, sorry for the sudden ping, is there a reason for not removing `stopPropagation()` in `onDrop()` plugin?

https://github.com/ianstormtaylor/slate/blob/6678245fba1965d971b32408a5800b892b8854a7/packages/slate-react/src/plugins/before.js#L304-L305

I faced similar issue as described in #1277 but for drop event. I'm using slate with react-dnd.

Looking at the current code here, it seems that `stopPropagation()` might be something I can remove? Which will allow me to drop on slate with react-dnd.

PR #1286, opened 2017-10-25
Fix Edge pasting exception

kellyjosephprice:
Pasting HTML was failing on Edge. It looks like `event.clipboardData.transfer.types` doesn't respond to `indexOf`. Alternatively, it appears to have `contains`.

Tested against Edge 38


ianstormtaylor:
Thanks @kellyjosephprice, great PR!

PR #1288, opened 2017-10-25
remove some polyfills

### LINKED ISSUE #1252: switch to BYO polyfills
It seems unfortunate to have to bundle `es6-map`, `lodash/find`, etc. and to have to maintain utils like `find-closest-node` when the browser support for lots of the things we're polyfilling is good. It makes the bundle sizes needlessly bigger. 

So I think it makes sense to stop polyfilling these things and to instead have people BYO polyfill to the environment using something like [polyfill.io](https://polyfill.io/v2/docs/).

Would be good to know what other commonly used libraries are doing on this front.

ianstormtaylor:
Work in progress to fix #1252


istarkov:
Just few hours all was fine with IE 11 ;-) this PR kills it support again 

There is no method closest in IE `import findClosestNode from '../utils/find-closest-node'`

It's really weird but IE 11 is still used in a lot of big companies as the only allowed browser. (Security blabla reasons) 
I'll try to use polyfill for this, will write tomorrow

istarkov:
All is fine, this polyfill fixes it https://www.npmjs.com/package/element-closest

ianstormtaylor:
@istarkov also you might want to check out [polyfill.io](https://polyfill.io/v2/docs/), which we [use in the Slate examples](https://github.com/ianstormtaylor/slate/blob/master/examples/index.html#L12) to add support for older browsers.

istarkov:
Cool will try. Thank you.

PR #1289, opened 2017-10-26
Revert "Keep marks after splitting a block (#1273)"

### LINKED ISSUE #1269: support carrying over marks between paragraphs

justinweiss:
This reverts commit 6f676d67712b15216cf611e01dead990ee0dd8c1.

This behavior should be in userland, because there isn't agreement
among text editors on how it should behave.

Resolves #1269.


ianstormtaylor:
Awesome, thank you @justinweiss!

PR #1292, opened 2017-10-26
Fix table example

doodlewind:
Handlers in table example are passed in `state` instead of `change`, leading to bug when press delete or backspace key in table. this PR simply fixes it:

table-bug


ianstormtaylor:
Ah good catch, thanks @doodlewind!

PR #1296, opened 2017-10-26
Fix wrap-unwrap Block-Inline docs

istarkov:
current doc for `{wrap/unwrap}{Block/Inline}` is incorrect.


ianstormtaylor:
Awesome! Thanks @istarkov!

PR #1297, opened 2017-10-26
Include Array.prototype.includes polyfill

bjohn465:
In 070a700, the `default` feature set was added and the `Array.prototype.includes` feature removed from the polyfill.io request. I think it's great to include the `default` feature set, but `Array.prototype.includes` is not included in the `default` set, so we have to explicitly request it. IE 11 needs an `Array.prototype.includes` polyfill in order for the examples to work. Without it, IE 11 gets this error:
TypeError: Object doesn't support property or method 'includes'

### DIFF
--- a/examples/dev.html
+++ b/examples/dev.html
@@ -8,7 +8,7 @@
   </head>
   <body>
     <main></main>
-    <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=default,es6"></script>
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Ah shoot, I'm sorry about that @bjohn465 I just quickly assumed that it was in `es6`, but it looks like it's actually in `es7` instead. I'm going to change this to add `es7`, since we're relying on things from there. Thank you!

bjohn465:
No problem. Thanks!

PR #1299, opened 2017-10-26
Fix property access of undefined object in getActiveMarksAtRangeAsArray

istarkov:
This situation is possible in following situation:
Create Inline as last element in a block, move cursor somewhere in other block, then set it at the end of previous block. If at every selection change you are trying to access `activeMarks` you'll got an error  `Uncaught TypeError: Cannot read property 'marks' of undefined at Document.getActiveMarksAtRangeAsArray`

Here is a gif:
slate-inline

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1091,7 +1091,7 @@ class Node {
     // If the range is collapsed, check the character before the start.
     if (range.isCollapsed) {
       const text = this.getDescendant(startKey)
-      if (text.characters.size === 0) return [];
+      if (text.characters.size === 0) return []
       const char = text.characters.get(range.startOffset - 1)
       return char.marks.toArray()
     }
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @istarkov I believe this was fixed by https://github.com/ianstormtaylor/slate/pull/1298, but let me know if you're still seeing this!

istarkov:
@ianstormtaylor no it's not fixed there, and I see it with latest slate and slate-react

istarkov:
Oups sorry, life is so fast. new Latest available ;-)

istarkov:
All is fine, thank you ;-)
BTW strange effect - small cursor jump in situation described above. 
(Not a problem for me)

istarkov:
jump video.
jump

mb it was better to have an empty text nodes around standalone inlines?

ianstormtaylor:
@istarkov Damn, yeah you're right, that is annoying. If you open another issue for it I can take a look. Thanks!

istarkov:
I solved this with css
```
  & [data-slate-zero-width] {
    width: 0;
    display: inline-block;
  }
```
seems like work fine ;-) so not an issue for me.

istarkov:
(As a fast solution for library the same can be done via style tag)

Issue #1283, opened 2017-10-25
add an `addMarks(marks)` change


ianstormtaylor:
Just as a convenience, instead of having to loop a marks set.

conorcussell:
This was fixed in https://github.com/ianstormtaylor/slate/pull/1306

PR #1306, opened 2017-10-27
Feature/add marks

ch1ll0ut1:
add an `addMarks(marks)` change #1283


ianstormtaylor:
Hey @ch1ll0ut1 thanks for this, it looks great!

Do you think the `addMarks()` change could also just loop `marks` and call into `addMark()` instead? Or did you try that already and there was an issue? Thanks!

ch1ll0ut1:
Hi @ianstormtaylor,
there was no real issue but my though was to create a performant loop but after looking at addMarks i wasn't sure that would be so easy. For the moment i let it in so that someone could easy replace that one line with a more performant version and does not have to change any other code as long as it is using addMarksAtRange. Also i think it makes the api more consistent by providing the same kind of method throughout the api. 

But anyway I also wanted to know what you think. I don't mind if you want me to take it out.

.yarnrc
```diff
@@ -0,0 +1,2 @@
+
+workspaces-experimental true
```
ianstormtaylor: Could you remove this file? I think with `yarn@1.x` this is no longer required.

packages/slate/src/changes/at-current-range.js
```diff
@@ -73,6 +73,40 @@ Changes.addMark = (change, mark) => {
 }
 
 /**
+ * Add a list of `marks` to the characters in the current selection.
+ *
+ * @param {Change} change
+ * @param {Mark} mark
+ */
+
+Changes.addMarks = (change, marks) => {
```
ianstormtaylor: Could this method's logic be implemented like the `addMarksAtRange` one you've added where it simply loops `marks.forEach( ... )` and calls into the old `addMark`? That way we don't end up with two almost-identical implementations to keep in sync?

.yarnrc
ch1ll0ut1: removed

packages/slate/src/changes/at-current-range.js
ch1ll0ut1: done.

ch1ll0ut1: sry for misunderstanding. Didn't read proper.

ch1ll0ut1: now locally i have no failing tests but circleci has the new tests all failing with:  
Error: No hyperscript creator found for tag: "state"
Do you have a idea what is the problem?

ianstormtaylor: @ch1ll0ut1 Thank you! If you rebase against master you should probably see the error. In the newest versions of Slate, the `state` has been renamed to `value`.

ch1ll0ut1: fixed

ianstormtaylor:
Awesome, thanks @ch1ll0ut1!

Issue #1215, opened 2017-10-10
add method: Change.removeAllMarksByKey


ch1ll0ut1:
#### Do you want to request a *feature* or report a *bug*?
feature

#### What's the expected behavior?

add a connivence method to simply remove all marks from a node. Right now doing that simple task seems to be very complicated and easy to make an mistake.

doodlewind:
@ianstormtaylor guess this one can be closed?

PR #1307, opened 2017-10-27
add change method to remove all marks from node by key

ch1ll0ut1:
add method: Change.removeAllMarksByKey #1215

### DIFF
--- a/packages/slate/src/changes/by-key.js
+++ b/packages/slate/src/changes/by-key.js
@@ -293,7 +293,7 @@ Changes.removeAllMarksByKey = (change, key, options = {}) => {
   const { state } = change
   const { document } = state
   const node = document.getNode(key)
-  const texts = node.getTextsAsArray()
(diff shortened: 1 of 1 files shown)


packages/slate/src/changes/by-key.js
```diff
@@ -281,6 +281,28 @@ Changes.removeMarkByKey = (change, key, offset, length, mark, options = {}) => {
...
+ * @param {String} key
+ * @param {Object} options
+ *   @property {Boolean} normalize
+ */
+
+Changes.removeAllMarksByKey = (change, key, options = {}) => {
+  const { state } = change
+  const { document } = state
+  const node = document.getNode(key)
+  const texts = node.getTextsAsArray()
```
ianstormtaylor: I think this will fail if `node` is a text node, so we need to handle that case too

ch1ll0ut1: fixed

ianstormtaylor:
Thanks @ch1ll0ut1!

Issue #1173, opened 2017-09-23
Selection sync issue on drop


AlbertHilb:
1. Open an example, "Rich Text" for instance, 
2. select a word,
3. drag&drop the word into another place in the document, let X be the position of the caret after the drop,
4. move the cursor to another position, Y, with the arrow keys,
5. press backspace,
6. a char is deleted at position X.

tobiasandersen:
This sounds very related to https://github.com/ianstormtaylor/slate/issues/907.

ianstormtaylor:
Here's a GIF:

![](http://g.recordit.co/9rcSW1rdDI.gif)

Interestingly, it seems to only happen in certain scenarios?

tobiasandersen:
I can confirm that this is the same issue as #907, and that it's a problem with React and not Slate.

tobiasandersen:
Should be ok to close this one now with #1309

PR #1309, opened 2017-10-27
fix broken onSelect event

### LINKED ISSUE #907: onDrop breaks onSelect event
Bug

`onDrop` events result in an editor with a broken selection. More specifically, [the `onSelect` handler of the main contentEditable `div`](https://github.com/ianstormtaylor/slate/blob/master/src/components/content.js#L736) never gets called after the drop event has fired.

As you can see in this gif below, the `onSelect` event gets called each time I move the cursor around (using the arrow keys). But once the `onDrop` event has fired, `onSelect` doesn't fire when moving the cursor around (also using the keys):

slate-gif

[Here's the fiddle](https://jsfiddle.net/c3nscb3c/1/) I used to record the demo above.

#### Additional info:
- It doesn't matter what we do in the `onDrop` handler. Inserting nodes (of any kind), or simply giving the editor focus, results in the same behavior. 
- If we blur the editor in `onDrop`, and let the user re-focus (by clicking), everything works again. 
- I remember having a similar problem when using Draft a few months back.

tobiasandersen:
This is a workaround for a [bug in React](https://github.com/facebook/react/issues/11379), and should fix #907 and #1173.

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -373,15 +373,15 @@ function AfterPlugin() {
     // has fired in a node: https://github.com/facebook/react/issues/11379.
     // Until this is fixed in React, we dispatch a mouseup event on that
     // DOM node, since that will make it go back to normal.
-    const focusNode = document.getNode(target.focusKey)
-    const domNode = findDOMNode(focusNode)
-    if (!domNode) return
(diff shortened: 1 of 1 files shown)


tobiasandersen:
cc @AlbertHilb

ianstormtaylor:
Thanks @tobiasandersen!

PR #1312, opened 2017-10-27
Fix focus after redo/undo triggered by hotkey

### LINKED ISSUE #1305: Unaccessible history state
bug

https://jsfiddle.net/doodlewind/wo348Lc0/

1. Using the <kbd>delete</kbd> button to delete some selected text.
2. Focus on text, press `cmd-z`.

Previous state can't be replayed with `cmd-z` since the editor is again blurred.

Behavior with `cmd-z` that loses focus:

bug-cmd-z

`cmd-z` shouldn't blur the editor. Expected Behavior with <kbd>undo</kbd> and <kbd>redo</kbd> button:

normal-undo-redo

#### Some investigations
This behaviour is regularly encountered in my editor project and it's frustrating, since user will think their changes are 'lost'. But in fact the history stack is preserved.

In Slate's default examples, toolbar buttons doesn't trigger this issue since changes are fired from `onMouseDown` and prevented default, editor is still focused afterwards. But when building advanced toolbar with color pickers or table pickers, mouse events can still blur the editor. In this case `cmd-z` is not usable.

IMHO it makes sense to **always** focus after redo or undo, the question is the place for the logic. Calling `redo().focus()` and `undo().focus()` as a workaround will break history stack and throw errors. Another workaround that calls `delete().focus()` is also awkward.

doodlewind:
Closes #1305 

User may continuously press `cmd-z`, during which if any state loses focus, states before can't be found by `cmd-z` then. this PR fixes it.

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -479,11 +479,15 @@ function AfterPlugin() {
     }
 
     if (HOTKEYS.REDO(event)) {
-      return change.redo()
+      // Keeps focus after redo triggered by hotkeys, such that user can always
+      // perform redo without being disturbed if any change blurred the editor.
+      return change.redo().focus()
     }
 
     if (HOTKEYS.UNDO(event)) {
-      return change.undo()
+      // Keeps focus after undo triggered by hotkeys, such that user can always
+      // perform undo without being disturbed if any change blurred the editor.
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Closing because I think we want the more "correct" way to solve it discussed in #1305

PR #1313, opened 2017-10-27
Rename "state" to "value" everywhere

### LINKED ISSUE #277: remove props.state from node components
That `state` prop passed into a Block instance is often stale or behind the real state, and `props.editor.getState()` is the preferred method for transforming state from within a block. Having  `state` as its own prop seems overkill and a trap for the unsuspecting, I think the best way to fix this is just remove the prop!

Is there a use case for having `state` as it's own prop?

**Steps to reproduce:**
1. Go to http://slatejs.org/#/embeds
2. Add a new paragraph at the very end, type ‘Testing’ into it
3. Click into the YouTube node <input>, type ‘c’
4. ‘Testing’ paragraph is gone

<div align="center"></div>

### LINKED ISSUE #834: change the word "state" to "value" everywhere
Open to discussion on this. I initially took the word `state` from Draft.js, but I think it ends up conflicting with React in confusing ways—both in terms of users understanding what "state" we're referring to, and in terms of naming conflicts for variables in the codebase.

I think `value` might be nicer, since it mimics native controlled `<input>` and `<textarea>` and we could even make `editor.value` a getter for super easy access. ~~And then `setValue` is available (right now we can't make our own `setState` function)~~.

Would do it in a backwards compatible way with deprecations hopefully, so to have the least negative impact when upgrading.

ianstormtaylor:
I wanted to get this one out of the way before we get more and more code depending on it—I should have done it sooner really. This cleans up a few things, helps us mimic the native DOM API even more, and helps us pass less props around.

---

###### BREAKING

- **Custom components no longer receive `props.state` or `props.schema`.** These are now exposed directly on the `props.editor` instance itself as `editor.value` and `editor.schema`. This helps eliminate a common issue where because of `shouldComponentUpdate` returning `false`, the `props.state` value was actually outdated, and transforming from it would cause incorrect behaviors.

- **The `plugin.renderEditor` function's signature has changed.** Previously it received `(props, state, editor)` but it now receives just `(props, editor)`. If you need access to the editor's current value, use the new `editor.value` property. This is simply to clean up the API, since the value is already accessible on `editor`.

- **The `set_state` operation has been renamed `set_value`**. This shouldn't affect almost anyone, but in the event that you were relying on the low-level operation types you'll need to update this.

###### DEPRECATED

- **The "state" has been renamed to "value" everywhere.** All of the current references are maintained as deprecations, so you should be able to upgrade and see warnings logged instead of being greeted with a broken editor. This is to reduce the confusion between React's "state" and Slate's editor value, and in an effort to further mimic the native DOM APIs.

- **The editor `getSchema()`, `getStack()` and `getState()` methods are deprecated.** These have been replaced by property getters on the editor instance itself—`editor.schema`, `editor.stack` and `editor.value`, respectively. This is to reduce confusion with React's own `setState`, and to make accessing these commonly used properties more convenient.

###### NEW

- **Added a new `editor.value` getter property.** This now mimics the DOM for things like `input.value` and `textarea.value`, and is the new way to access the editor's current value.

- **Added new `editor.schema` and `editor.stack` getters.** Similarly to the new `value` getter, these two new getters give you access to the editor's current schema and stack.

---

Fixes #277 
Fixes #834


Soreine:
Good work on the API cleanup.

I did not see your suggestion to rename `state` to `value` everywhere. `state` was a really poor name indeed. It was appropriately named `state`, as "how things are at a given time". But the word `state` alone is rarely used in English. The state of what? And there is no real alternative, everyone is using `state` (React in particular). It's all fine until we start having states for several things. Throughout our code base, we name it `EditorState`, for disambiguation. It's simple, maybe a bit longer, but dead clear. You could have renamed `State` to `EditState` for example. Then in discussion, people would no longer say "state" but "edit state", and there would be no confusion anymore.

There are not much words meaning "state" in English. We could have chosen "snapshot", to convey the idea that this is the top-level representation of the state of an editor, at a given time (contains history + selection + content). "snapshot" is less often used than "state", so it would reduce confusion as well.

"value" however... is as ambiguous as "state" (we will probably name our variable 'editorValue' to avoid confusion), but also further from the correct meaning. Text areas have a value, which is a simple string. No selection, no history. In Slate, one could expect the value to be the document.

But I see the point of mimicking the DOM API, so that you can write

```jsx
<input value={...} />
<Slate.Editor value={...} />
```

That way, both are _kind of_ the same things.

It just feels like moving the city to build a house.

---

That was just my late-to-the-party opinion. I do encourage going fast and breaking things as long as Slate is in beta. So let's see how it goes in the long-term :)

Cheers

ianstormtaylor:
Hey @Soreine, thanks for writing up your thoughts, I really appreciate it!

I agree with you that both `state` and `value` are ambiguous, because they are overloaded terms in programming in general. `state` is obviously the bigger issue, because React has already co-opted that term for their use, and it results in very confusing code when juxtaposed.

I agree that `value` isn't the most differentiated name. The way I see it, we could either go for very differentiated, in which case "snapshot" is actually a nice term—I like it. Or we could go for very mimicking of existing terminology/use, in which case I prefer `value`. It gives us things like:

```
<Editor value={...} />
<Editor defaultValue={...} />
editor.value
```

These all feel natural, such that anyone who reads them likely immediately understands what the concepts are achieving, without having to learn what a "snapshot" is. And I like that it mimics the native APIs. (This mimicking wasn't actually a principle I had earlier on when designing Slate, but I'm noticing more and more how powerful it can be.)

I think for now if people want to be super explicit, they can definitely continue to use something like `editorValue` when using it as a variable name, if it helps disambiguate. But at the same time, one would argue that when storing the native `input.value` it would need to be prefixed in the same way to become `inputValue` to disambiguate it too. (In a way it's kind of like the `users.user_id` vs. `users.id` column naming debate in SQL land.)

That said, if we do discover that `value` is too problematic for more reasons over time, I'm very open to consider renaming it to "snapshot" down the road.

My goal while Slate is in beta is to not let us get too freaked out to make breaking changes that improve the UX and long-term maintenance—I feel like a lot of codebases get stuck in the no-more-core-changes mentality too soon. Of course, balanced by using helpful deprecations whenever possible to make migration much easier.

PR #1320, opened 2017-10-28
Adjust drop code to the new structure of void nodes.

AlbertHilb:
After the change in the structure of void nodes (#1265), `getEventRange` does not work correctly anymore.
This PR fixes that issue.

### DIFF
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -243,7 +243,8 @@ function BeforePlugin() {
   function onDragOver(event, change, editor) {
     // If the target is inside a void node, and only in this case,
     // call `preventDefault` to signal that drops are allowed.
-    // When the target is editable, drop is allowed by default.
+    // When the target is editable, dropping is already allowed by 
+    // default, and calling `preventDefault` hides the cursor.
     const node = findNode(event.target, editor.value)
     if (node.isVoid) event.preventDefault()
 
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -241,14 +241,17 @@ function BeforePlugin() {
    */
 
   function onDragOver(event, change, editor) {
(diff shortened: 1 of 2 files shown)


packages/slate-react/src/plugins/before.js
```diff
@@ -241,15 +241,18 @@ function BeforePlugin() {
    */
 
   function onDragOver(event, change, editor) {
+    // If the target is inside a void node, and only in this case,
```
ianstormtaylor: Hey @AlbertHilb, what's your thinking here? Or even more generally, should we be calling `preventDefault` at all in any cases? 

I would have thought that either we'd call it in all cases, which allows for text dragging/dropping? (But maybe text dragging/dropping is automatic and doesn't require this?) Or we'd call it in no cases, and have the user always opt-in?

I might be thinking about this incorrectly though. Let me know!

AlbertHilb: Preventing default action on `dragover` event, has different effects if the target is in an editable context or in a non-editable one.

When the target is in a non-editable context, calling `preventDefault` is necessary to allow dropping.
Conversely, in an editable context, dropping is allowed by default, and it still remains allowed even if `preventDefault` is called, but in that case the browser does not draw the caret showing where the insertion point is.

Therefore, if we call `preventDefault` in any case, we get dropping allowed everywhere but we lose the browser caret feedback.
If we never call `preventDefault, instead, we get dropping allowed only inside text, not on void nodes; this breaks the metaphor of void nodes behaving like single characters and makes difficult dropping contents on their sides.

For these reasons, I chose to prevent default action only when the target is inside void nodes (non-editable context). 
So we can preserve the single character metaphor of void nodes - with dropping taking place on the side closest to the drop point, when the drop point is inside a void node - and, at same time, the caret feedback drawn by the browser.

ianstormtaylor: @AlbertHilb thank you that is an amazing explanation. This sounds great.

PR #1329, opened 2017-10-29
Prevent history mutation during initialization

### LINKED ISSUE #1327: Losing redo state after popping last undo item
Bug

https://jsfiddle.net/doodlewind/6ofmL42m/

1. Select some texts and click `delete`, repeat several times to fill history stack.
2. Click `undo`, begin traveling in history states.
3. When `undos.size == 1`, click `undo`, **redo count is cleared!**

During traveling in history, `undos.size` + `redos.size` is an constant. But **when user wants to pop up last undo item, redo stack is cleared**. In this case user can't travel back with `redo`, bringing disturbing UE.

bug-lose-redo

When `undos.size == 1`, popping up last undo item shouldn't clear the redo stack.

I've fixed this in my local branch, a PR is landing soon!

### LINKED ISSUE #1328: Wrong undo count after initialization
Bug

https://jsfiddle.net/doodlewind/6ofmL42m/

Open this page and do nothing, `undos.size` is 1 instead of 0. Neither can we popped up this undo item. As a result, it leads to inaccurate computing of `state.hasUndos`.

1. `undos.size` should be 0 as an init state.
2. Last undo item should be able to popped up.

I've fixed this in my local branch, a PR is landing soon!

doodlewind:
Fixes #1327 
Fixes #1328 

Currently there are two weird history stack behaviors:

1. When popping up last undo item, redo stack is cleared.
2. After initialization, `undos.size` is 1 instead of 0.

According to existing comments, during initialization, `editor.schema` will change once, then `onChange` in before-plugin triggers `change.setValue`, this saves to history by default, making undo stack's size 1 instead of 0.

Then when popping up this undo item, a redundant `setValue` operation is performed as inverse, this operation contains `{ save: true }`, so it fills undo stack again, which is makes undo stack behaves as alway 'remaining' one sole item.

This screenshot demonstrates the process described above when trying to pop up last undo item:

redo-clear-marked

As as solution, this PR detects history stack size when we tries to update `editor.schema`, setting the `save` args to false for initialization, in this way both issue can be solved.

A side effect I can think of is that we'll now keep the first passed in schema as 'default', this seems to be reasonable. If schema changes afterward, it'll still be recorded into history stack and still undoable.

### DIFF
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -102,15 +102,10 @@ function BeforePlugin() {
 
     // If the value's schema isn't the editor's schema, update it. This can
     // happen on the initialization of the editor, or if the schema changes.
+    // Since only schema is updated, don't save this change into history.
     if (value.schema != editor.schema) {
-      // When change on initialization, don't save change into history, making
-      // this schema as default and history clean after initialization ends.
-      const isInit = (
-        value.history.undos.size === 0 &&
-        value.history.redos.size === 0
-      )
       change
-        .setValue({ schema: editor.schema }, { save: !isInit })
+        .setValue({ schema: editor.schema }, { save: false })
         .normalize()
     }
 
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -102,7 +102,7 @@ function BeforePlugin() {
 
     // If the value's schema isn't the editor's schema, update it. This can
     // happen on the initialization of the editor, or if the schema changes.
-    // Since only schema is updated, don't save this change into history.
+    // This change isn't save into history since only schema is updated.
     if (value.schema != editor.schema) {
       change
         .setValue({ schema: editor.schema }, { save: false })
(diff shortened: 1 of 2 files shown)


packages/slate-react/src/plugins/before.js
```diff
@@ -103,8 +103,14 @@ function BeforePlugin() {
...
     if (value.schema != editor.schema) {
+      // When change on initialization, don't save change into history, making
+      // this schema as default and history clean after initialization ends.
+      const isInit = (
+        value.history.undos.size === 0 &&
+        value.history.redos.size === 0
+      )
       change
-        .setValue({ schema: editor.schema })
+        .setValue({ schema: editor.schema }, { save: !isInit })
```
ianstormtaylor: Actually, I think we could just always not save for this one? Since no matter what it's going to continue to get set even if the user does undo it?

doodlewind: Fixed!

ianstormtaylor:
Thanks @doodlewind!

PR #1330, opened 2017-10-29
Omit focus change when traveling in history

### LINKED ISSUE #1305: Unaccessible history state

doodlewind:
Closes #1305 

We've discussed an approach fixing the unaccessible history state issue:

> Instead of discarding only blurs, we actually change `undo/redo` to ignore all changes to isFocused, so that the focused state will never change when undoing, it will remain whatever it currently is.

This is a cool idea and this PR implements this by checking operation type during `undo/redo`. Now no matter `undo/redo` is triggered from toolbar button click or keyboard hotkey press, editor focus remains intact, making successive history traveling unhindered.

### DIFF
--- a/packages/slate/src/changes/on-history.js
+++ b/packages/slate/src/changes/on-history.js
@@ -30,6 +30,12 @@ Changes.redo = (change) => {
 
   // Replay the next operations.
   next.forEach((op) => {
+    // When the operation mutates selection, omit its `isFocused` props to
+    // prevent editor focus changing during continuously redoing.
+    const { type, properties } = op
+    if (type === 'set_selection') {
+      delete properties.isFocused
+    }
     change.applyOperation(op, { save: false })
   })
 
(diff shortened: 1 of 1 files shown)


packages/slate/src/changes/on-history.js
```diff
@@ -30,6 +30,12 @@ Changes.redo = (change) => {
 
   // Replay the next operations.
   next.forEach((op) => {
+    // When the operation mutates selection, omit its `isFocused` props to
+    // prevent editor focus changing during continuously redoing.
+    const { type, properties } = op
+    if (type === 'set_selection') {
+      delete properties.isFocused
```
ianstormtaylor: @doodlewind can you change the logic for these to use `lodash/omit` instead, so we aren't accidentally mutating the `op` object. Thanks!

doodlewind: Fixed!

ianstormtaylor:
Thanks @doodlewind!

PR #1331, opened 2017-10-29
Add history example

doodlewind:
Slate's examples are awesome for quick prototyping and bug reporting. For history-stack-related issues, it'll be more efficient tracking and reporting them with a live example. The more user playing with this, the easier to expose related bugs.

Another major point for adding this example is that to my knowledge, each example shipped with Slate demonstrates one major feature.  For custom components and changes, one great advantage of Slate is that their history operations are handled automatically, without the need implementing their own undo/redo method. So from this point of view, adding it as a standalone example also makes sense.

Currently this example looks like below:

history-demo

Willing to update whatever needed 😀

### DIFF
--- a/examples/history/index.js
+++ b/examples/history/index.js
@@ -46,18 +46,6 @@ class History extends React.Component {
   }
 
   /**
-   * On delete selected text.
-   *
-   */
-
-  onClickDelete = (event) => {
-    event.preventDefault()
-    const { value } = this.state
-    const change = value.change().delete()
-    this.onChange(change)
-  }
-
-  /**
    * On redo in history.
    *
    */
@@ -106,7 +94,6 @@ class History extends React.Component {
     const { value } = this.state
     return (
       <div className="menu toolbar-menu">
-        <ToolbarButton icon="delete" onMouseDown={this.onClickDelete} />
         <ToolbarButton icon="undo" onMouseDown={this.onClickUndo} />
         <ToolbarButton icon="redo" onMouseDown={this.onClickRedo} />
         <span className="button">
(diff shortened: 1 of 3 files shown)


examples/history/index.js
```diff
@@ -0,0 +1,146 @@
...
+   * Render the toolbar.
+   *
+   * @return {Element}
+   */
+
+  renderToolbar = () => {
+    const { value } = this.state
+    return (
+      <div className="menu toolbar-menu">
+        <ToolbarButton icon="delete" onMouseDown={this.onClickDelete} />
```
ianstormtaylor: I don't think we need this button because a user can just use <kbd>backspace</kbd> or <kbd>delete</kbd> to achieve the same behavior? I'd rather keep it super simple.

examples/history/value.json
```diff
@@ -0,0 +1,34 @@
...
+    "nodes": [
+      {
+        "kind": "block",
+        "type": "paragraph",
+        "nodes": [
+          {
+            "kind": "text",
+            "leaves": [
+              {
+                "text": "Editor maintains its own history stack. For custom components and changes, history operations are automatically handled, no need implementing their undo/redo methods by yourself!"
```
ianstormtaylor: For the example text, can we go with:

> Slate editors save all changes to an internal "history" automatically, so you don't need to implement undo/redo yourself. And the editor automatically binds to the browser's default undo/redo keyboard shortcuts.

> Try it out for yourself! Make any changes you'd like then press "cmd+z".

Thanks!

examples/history/index.js
doodlewind: Fixed!

examples/history/value.json
doodlewind: Fixed!

ianstormtaylor:
Looks great, thank you @doodlewind.

PR #1333, opened 2017-10-29
Add default transferred text

### LINKED ISSUE #1332: Error pasting single image
Bug

1. In [Slate Image Example](http://slatejs.org/#/images), choose image and copy it.
2. Open https://jsfiddle.net/doodlewind/86dw9gno/, focus the editor and paste.
3. Error thrown.

```
Uncaught TypeError: Cannot read property 'split' of null
    at Object.deserialize (https://npmcdn.com/slate-react@0.10.1/dist/slate-react.js:7136:21)
    at Object.onPaste (https://npmcdn.com/slate-react@0.10.1/dist/slate-react.js:11005:49)
    at Stack.run (https://npmcdn.com/slate@0.30.1/dist/slate.js:15496:38)
    at https://npmcdn.com/slate-react@0.10.1/dist/slate-react.js:8780:20
    at Change.call (https://npmcdn.com/slate@0.30.1/dist/slate.js:9611:10)
    at Editor.change (https://npmcdn.com/slate-react@0.10.1/dist/slate-react.js:8762:63)
    at Editor.onEvent (https://npmcdn.com/slate-react@0.10.1/dist/slate-react.js:8779:12)
    at Object._this.(anonymous function) [as onPaste] (https://npmcdn.com/slate-react@0.10.1/dist/slate-react.js:8529:23)
    at Content.onEvent (https://npmcdn.com/slate-react@0.10.1/dist/slate-react.js:8251:26)
    at _this.(anonymous function) (https://npmcdn.com/slate-react@0.10.1/dist/slate-react.js:8135:15)
```

Copying from hotkeys or browsers menus results the same.

See GIF:

bug-paste-img

Image inserted as a void block since its deserializing rule is added.

I've fixed this in my local branch, a PR is landing soon!

doodlewind:
Closes #1332 

In current implementation, transferred text may be `null` instead of empty string, which may breaks deserializer. This PR fixes it.

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -581,7 +581,7 @@ function AfterPlugin() {
       change.insertFragment(fragment)
     }
 
-    if (type == 'text' || type == 'html') {
+    if ((type == 'text' || type == 'html') && text) {
       const { value } = change
       const { document, selection, startBlock } = value
       if (startBlock.isVoid) return
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -581,8 +581,8 @@ function AfterPlugin() {
       change.insertFragment(fragment)
     }
 
-    // Only allows pasting plain text or html, with valid text
-    if ((type == 'text' || type == 'html') && text) {
+    if (type == 'text' || type == 'html') {
+      if (!text) return
       const { value } = change
       const { document, selection, startBlock } = value
       if (startBlock.isVoid) return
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
@doodlewind I think this should actually be fixed by changing the line in `./plugins/after.js@onPaste` from:

```js
if (type == 'text' || type == 'html')
```

To something like:

```js
if ((type == 'text' || type == 'html') && text)
```

Right? I think this seems safer, because it means we're not adding a default when it maybe shouldn't exist?

doodlewind:
@ianstormtaylor Fixed!

PR #1334, opened 2017-10-29
Handle pasting image in `Paste HTML` example

doodlewind:
Currently when testing paste feature, the `Paste HTML` example mainly supports:

* Block nodes with text, like `h1` and `paragraph`.
* Block nodes with text but is 'flat', like `code`.
* Void Inline nodes, like `link`.

So how about **void block nodes** like `image`? IMHO supporting image makes sense, since we can copy image from the `Images` example and paste it then, making it easier for smoke tests.

This PR simply adds this functionality.

### DIFF
--- a/examples/paste-html/index.js
+++ b/examples/paste-html/index.js
@@ -87,7 +87,7 @@ const RULES = [
     }
   },
   {
-    // Special case for images, to grab their srs.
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Awesome, thanks @doodlewind!

PR #1336, opened 2017-10-29
Delete single void node when cutting

### LINKED ISSUE #1335: Unable to delete single void node after cutting
bug

* Open [Emojis](http://slatejs.org/#/emojis) example, click on an emoji, `cmd-x` doesn't remove it.
* Open [Images](http://slatejs.org/#/images) example, click on an image, `cmd-x` doesn't remove it either.

* `cmd-x` should also removes the selected node.

Fixed in my local branch, PR landing soon!

doodlewind:
Closes #1335 

Currently when user click on a void block/inline node, this node appears in `value.blocks` or `value.inlines`, but selection is collapsed in this case. Then if user cuts it, we can't simple calling `change().delete()` to remove it, instead we manually remove it.

Since we still perform deletion through changes, undo/redo behaves just as what it is.

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -125,9 +125,9 @@ function AfterPlugin() {
       // If user cuts a void block or a void inline,
       // manually removes them since selection is collapsed in this case.
       const { value } = change
-      const { endBlock, endInline } = value
-      const isVoidBlock = endBlock && endBlock.isVoid
-      const isVoidInline = endInline && endInline.isVoid
+      const { endBlock, endInline, isCollapsed } = value
+      const isVoidBlock = endBlock && endBlock.isVoid && isCollapsed
+      const isVoidInline = endInline && endInline.isVoid && isCollapsed
 
       if (isVoidBlock) {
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/plugins/after.js
```diff
@@ -122,7 +122,20 @@ function AfterPlugin() {
...
     window.requestAnimationFrame(() => {
-      editor.change(c => c.delete())
+      // If user cuts a void block or a void inline,
+      // manually removes them since selection is collapsed in this case.
+      const { value } = change
+      const { endBlock, endInline } = value
+      const isVoidBlock = endBlock && endBlock.isVoid
+      const isVoidInline = endInline && endInline.isVoid
+
+      if (isVoidBlock) {
```
ianstormtaylor: It seems like it would be better to use something like `if (isCollapsed && isVoid)` as the condition? Because as currently written, won't this fail to delete the full range if it happens to end in a void?

I might be missing something thought, let me know!

doodlewind: Fixed!

ianstormtaylor:
Awesome, thanks @doodlewind.

PR #1337, opened 2017-10-29
Fix lint

doodlewind:
All currently open PR builds are failed since #1320 introduced a minor lint issue. This PR simply fixes it.


ianstormtaylor:
Ah that's my fault, not sure why the GitHub editor is introducing those. Thanks  @doodlewind!

doodlewind:
@ianstormtaylor seems that #1336 #1333 #1331 #1330 #1329 are failed because of this, while I believe these patches won't fail the test though. Would you please help re-running the Travis test or just getting them merged if possible? Thanks.

PR #1342, opened 2017-10-30
Trigger `onBlur` regardless of `relatedTarget`

### LINKED ISSUE #1341: onBlur not firing
onBlur on the Editor component doesn't fire anymore.

```
    "slate": "^0.30.1",
    "slate-html-serializer": "^0.4.1",
    "slate-react": "^0.10.1",
```

doodlewind:
Closes #1341

`onBlur` events are not firing because the PERF code in before-plugin returns `true` when `event.relatedTarget` is null, preventing custom callbacks from being called.

This is somewhat overkill, since according to [this post](https://stackoverflow.com/questions/42764494/blur-event-relatedtarget-returns-null), `event.relatedTarget` returns elements **only if this element can gain focus**, so clicking on plain `<div>` outside editor won't trigger `onBlur`, which is not very reasonable.

This patch removes this limitation and `onBlur` intuitively works. As for the side effect I can think of, this change may allow some `set_selection` pushed into history stack, causing focus change on undoing, but we already have #1330 handle this case.

### DIFF
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -71,27 +71,26 @@ function BeforePlugin() {
     const { value } = change
     const focusTarget = event.relatedTarget
 
-    // If focusTarget is null, the blur event is due to the window itself being
-    // blurred (eg. when changing tabs) so we should ignore the event, since we
-    // want to maintain focus when returning.
-    if (!focusTarget) return true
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @doodlewind, that's interesting. I guess we'll have to figure out another way to handle the window blurring case.

PR #1343, opened 2017-10-30
Update walkthrough documents' initial values

rwoody:
This appears to have been missed recently when `ranges` was renamed.

Strange because it seemed `ranges` had initially worked for me, and then stopped working later on in the day... :man_shrugging:


ianstormtaylor:
Thanks for catching this @rwoody!

Check the changelog, `0.30.0` removed all previously deprecated changes, which might be what you're referring to about no longer working.

PR #1350, opened 2017-10-30
[onCutOrCopy] Creating the phony copy div at the same level as the real div

### LINKED ISSUE #1339: Copy on Firefox causes a window scroll jump (MacOS)
*bug*

In Firefox: Just select some text and fire a copy command (shortcut or right-click menu) on the official slatejs examples: http://slatejs.org/#/huge-document (document should be larger than a window frame). The selection will jump to the bottom.

xvqg75pk3e

No jumps

#### Possible Fix

Comparing the current code in `onCutOrCopy`, where a fake `div` is created with the slate fragment: https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/after.js#L217-L224

with a similar solution in [`cliboard.js`](https://github.com/zenorocha/clipboard.js): https://github.com/zenorocha/clipboard.js/blob/master/src/clipboard-action.js#L56-L76

`clipboard.js` adds an extra Y-offset to the fake div, matching the current window scroll offset. This keeps the dummy div (which is offscreen) at the same vertical offset as the selection. So when the selection is added to the cliboard, there is no jump in firefox.

I added this to the current `onCutOrCopy`:
```
    const yPosition = window.pageYOffset ||
                      window.document.documentElement.scrollTop
    div.style.top = `${yPosition}px`
```

This did stop the jumping, but I am not sure if this is the right solution.

ashutoshrishi:
The function `onCutOrCopy` in `after.js` plugin stack creates a phony div with encoded slate fragment in an attribute and selects it (to make the browser copy those contents instead of the actual selection). 

In Firefox this causes the viewport to jump to the bottom (if the window can scroll), to find the div (presumably). `clipboard.js` fixes this by creating the phony div at the same scroll offset as the current viewport.

(See attached issue for gif)

# Changelog
###### Fixed
- Creating phony div at the scroll offset as the window scroll offset 

Resolves #1339

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -1,4 +1,3 @@
-
 import Base64 from 'slate-base64-serializer'
 import Debug from 'debug'
 import Plain from 'slate-plain-serializer'
@@ -245,6 +244,20 @@ function AfterPlugin(options = {}) {
     div.setAttribute('contenteditable', true)
     div.style.position = 'absolute'
     div.style.left = '-9999px'
+
+    // The snippet below is from `clipboard.js`
+    // Prevent zooming on iOS
+    div.style.fontSize = '12pt'
+    // Reset box model
+    div.style.border = '0'
+    div.style.padding = '0'
+    div.style.margin = '0'
+    // Creating the phony div at the same level as the real div so that
+    // the viewport doesn't jump (in Firefox)
+    const yPosition = window.pageYOffset ||
+                      window.document.documentElement.scrollTop
+    div.style.top = `${yPosition}px`
+
     div.appendChild(contents)
     body.appendChild(div)
 
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -245,15 +245,13 @@ function AfterPlugin(options = {}) {
     div.style.position = 'absolute'
     div.style.left = '-9999px'
 
-    // The snippet below is from `clipboard.js`
-    // Prevent zooming on iOS
-    div.style.fontSize = '12pt'
-    // Reset box model
+    // COMPAT: In Firefox, the viewport jumps to find the phony div. Hence it
+    // should be created at the current window scroll offset (setting
+    // 'style.top'). The box model attributes which can interact with 'top' are
+    // also reset.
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/plugins/after.js
```diff
@@ -245,6 +244,20 @@ function AfterPlugin(options = {}) {
...
     div.style.position = 'absolute'
     div.style.left = '-9999px'
+
+    // The snippet below is from `clipboard.js`
+    // Prevent zooming on iOS
+    div.style.fontSize = '12pt'
+    // Reset box model
+    div.style.border = '0'
+    div.style.padding = '0'
+    div.style.margin = '0'
```
ianstormtaylor: Do we need all of these tangential things that don't seem related to the scroll issue?

packages/slate-react/src/plugins/after.js
```diff
@@ -245,6 +244,20 @@ function AfterPlugin(options = {}) {
...
+    // Creating the phony div at the same level as the real div so that
+    // the viewport doesn't jump (in Firefox)
```
ianstormtaylor: Can you rewrite this comment to be in the `COMPAT:` form found elsewhere in the codebase?

packages/slate-react/src/plugins/after.js
ashutoshrishi: The box dimensions interact with the 'top' attribute. It won't be a problem unless someone has all divs default to have margin near to the current window height. 

What do you think?

packages/slate-react/src/plugins/after.js
ashutoshrishi: cool

packages/slate-react/src/plugins/after.js
ianstormtaylor: In that case, for any that are necessary, let's put them all under the same `COMPAT:` comment so that it's clear. And any that aren't we remove

ashutoshrishi: I will remove:

```
    // Prevent zooming on iOS
    div.style.fontSize = '12pt'
```

and add the rest under COMPAT.

ianstormtaylor:
Thanks @ashutoshrishi!

Issue #1340, opened 2017-10-30
[IME] Crashing when inputting more than a single character


blakeembrey:
#### What's the current behavior?

![](https://d26dzxoao6i3hh.cloudfront.net/items/0D072i3v3q3q3o3G0F2L/Screen%20Recording%202017-10-29%20at%2010.10%20pm.gif)

#### What's the expected behavior?

It shouldn't crash. I've been trying to debug and figure out where to start, but unfortunately the codebase is still a bit beyond me. It appears to be related to https://github.com/ianstormtaylor/slate/issues/1229 and meant to be fixed with https://github.com/ianstormtaylor/slate/pull/1322, but it seems to still be broken running from `master`.

doodlewind:
I believe this is a common pitfall in editors based on React tech stack, since this also happens in [Draft.js](https://draftjs.org/docs/advanced-topics-issues-and-pitfalls.html#browser-plugins-extensions). The direct reason is that **vDOM and DOM are out of sync**.

We've ran into several kinds of IME bugs in Slate for now:

1. Selection collapsed, cursor inside blocks, errors thrown during/after IME composition. See #1314.
2. Selection collapsed, cursor in an empty new block, errors thrown after IME composition. See #1229.
3. Selection expanded, error thrown after IME composition. See #1114.

On desktop, for 1 and 2 we've generally fixed through COMPAT code in `slate-react`'s plugin (only desktop Safari still facing case 2), but case 3 is the most difficult one, since browser **natively removes all selected blocks on IME updates, and such behavior happens before IME composition event triggers, so when Slate tries updating these nodes, React crashed**.

Seems that this issue you raised belongs to case 3. To reproduce just do as below:

1. Select multi blocks.
2. Use IME to input some texts.
3. Choose a word.

Just do this in [demo](https://draftjs.org) of Draft.js and you'll also crash them 😈 I've yet not thinking about any 'elegant' approaches handling this for now, but hope this writeup somewhat helps.

blakeembrey:
Thanks @doodlewind, I was looking through the other PRs and issues you were opened earlier and trying to use that as a starting point for what the issue may be. Interestingly, I can't crash the Draft.js demo - can you share your replication here? When I do the same input with Draft.js, it correctly responds immediately by hiding the placeholder text and doesn't crash when inputting more than a single character from IME input.

[...]

PR #1322, opened 2017-10-28
Replace <br /> with line break

doodlewind:
Closes #1229 

When inserting new line in Slate with IME, browser automatically removes `<br />` generated for rendering empty block, then the mapping from Slate Value => React vDOM => DOM is broken, React then throws error, page is stopped.

#1316 introduces a workaround but is reverted, this PR is a leaner version as a replacement, with related assumptions updated.

BTW, currently I can see many other cases that IME's native behavior breaks the sync from vDOM to DOM, resulting in serious error thrown by React. Any ideas for a universal workaround?

PR #1353, opened 2017-10-31
Fix native IME input at the beginning of a line

blakeembrey:
Closes https://github.com/ianstormtaylor/slate/issues/1340.
Related to https://github.com/ianstormtaylor/slate/pull/1322.

/cc @doodlewind, did you want to test your non-native pinyin input first? This works for the native macOS input.

### DIFF
--- a/packages/slate-react/src/components/leaf.js
+++ b/packages/slate-react/src/components/leaf.js
@@ -120,13 +120,7 @@ class Leaf extends React.Component {
    */
 
   renderText() {
-    const { block, node, parent, text, index, leaves } = this.props
-
-    // COMPAT: If the text is empty and it's the only child, we need to render a
-    // line break to get the block to have the proper height.
-    if (text == '' && parent.kind == 'block' && parent.text == '') {
-      return <span data-slate-empty-block>{'\n'}</span>
-    }
+    const { block, node, text, index, leaves } = this.props
 
     // COMPAT: If the text is empty otherwise, it's because it's on the edge of
     // an inline void node, so we render a zero-width space so that the
--- a/packages/slate-react/src/utils/find-dom-point.js
+++ b/packages/slate-react/src/utils/find-dom-point.js
@@ -39,16 +39,6 @@ function findDOMPoint(key, offset) {
     start = end
   }
 
-  // COMPAT: For empty blocks with only a single empty text node, we will have
-  // rendered a `<span>` with `\n` inside, instead of a text node.
-  if (
-    el.childNodes.length == 1 &&
-    el.childNodes[0].childNodes.length == 1 &&
-    el.childNodes[0].childNodes[0].hasAttributes('data-slate-empty-block')
-  ) {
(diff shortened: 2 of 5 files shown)


packages/slate-react/src/components/leaf.js
```diff
@@ -125,7 +125,7 @@ class Leaf extends React.Component {
     // COMPAT: If the text is empty and it's the only child, we need to render a
     // line break to get the block to have the proper height.
     if (text == '' && parent.kind == 'block' && parent.text == '') {
-      return <span data-slate-empty-block>{'\n'}</span>
+      return <span data-slate-empty-block>{'\u200B'}</span>
```
doodlewind: This works! For the remaining two failing test case, just use

```
<span data-slate-empty-block="true"></span>
```

instead of

```
<span data-slate-empty-block="true">\n</span>
```

and I believe that'll be adequate.

doodlewind:
Third party pinyin IME works fine still, good job @blakeembrey !

packages/slate-react/src/components/leaf.js
blakeembrey: Good catch! Forgot to run the tests 😄

blakeembrey:
Just a note, but I'm still working on this. I think I ran into some edge cases that can happen and found a cleaner way around all this.

blakeembrey:
Ok, managed to kill a bunch of code and fix the functionality all around. Appreciate if you test it again @doodlewind using your custom IME input since I only have the macOS default included.

I killed the `data-slate-empty-block` span because it made things more complicated - it actually had been missed about when copying and pasting. There's special code that removed `data-slate-zero-width` but not the new empty element. Combining them back together makes it easier to avoid that mistake.

I also killed the code moving the offset in `findDOMPoint` since it gets corrected in `findPoint` anyway. All that was required to handle IME input was instead subtracting 1 from the size of the zero width element instead of resetting it back to 0. This allows the browser to type in the empty span while using IME and then fix it up later.

ianstormtaylor:
This is awesome @blakeembrey! Love the code removal. I'll wait for the testing on other IMEs before merging. Looks good for regular stuff from my own testing though.

ianstormtaylor:
@doodlewind no idea! 😄 Looks good to me though.

doodlewind:
@ianstormtaylor Done!

blakeembrey:
@doodlewind What does "Replace Multi Blocks" mean and does the cross mean it's failing today (and was it before)?

doodlewind:
@blakeembrey This is not a regression, but a known issue in #967 I think.

ianstormtaylor:
@blakeembrey @doodlewind so this is good to merge?

blakeembrey:
@ianstormtaylor Looks like it! Exciting 😄

doodlewind:
@ianstormtaylor I've tested and feeling it's good. 😁 To my knowledge no regression introduced.

ianstormtaylor:
Thanks you two!

Issue #1308, opened 2017-10-27
State sometimes change twice after one change


chemzqm:
#### Do you want to request a *feature* or report a *bug*?

**BUG**

#### What's the current behavior?

https://jsfiddle.net/31tb8gyx/30/

Select some text and click bold button, and you will see the `onChange` event fired twice, and as a result, the history state of editor is undesierable.

This issue happens only when call `e.preventDefault` in `onMouseDown` handler,  but without `e.preventDefault`, the editor would blur and user could not continue editing, which is also too bad.

* OS: MacOS
* Browser: Chrome
* Slate: 0.28.2

#### What's the expected behavior?

No twice state change, no unnecessary rerender, the history state should be clean.

doodlewind:
IMHO this is the same issue with #1069 since slate's history stack records selection change.

ianstormtaylor:
When I click <kbd>bold</kbd> in that JSFiddle I only see a single change firing. If I click <kbd>undo</kbd> I see two changes firing though, because it's using `onClick` instead of `onMouseDown`.

Can you clarify what you mean here?

chemzqm:
@ianstormtaylor one change without text selection, but two changes with text selection.
Same issue exists on Safari and Firefox.

isubastiCadmus:
I can confirm this. but it only happens on drag selection, not double click/triple click selection. What's the difference in drag/double click/triple click that can possibly cause a set_selection operation needed after toggling mark @ianstormtaylor ?

isubastiCadmus:
@ianstormtaylor actually there is a more specific case when this occurs, @chemzqm can you clarify this as well. So with the drag selection, if the selection is forward selection it seems to be running onChange only once. It doesn't seem to be the case on backward selection, on backward selection it triggers onChange twice.

chemzqm:
@isubastiCadmus yes, the problems exists for backward selectin only, not forward selection.

isubastiCadmus:
@chemzqm yeah, for some reason it run's a selection change that flips the selection from backward to forward if the selection is backward. still trying to chase the problem down.

ianstormtaylor:
@chemzqm I'm still not clear on what you are talking about. If you change the `undo` handler to be `onMouseDown`, then there's no issue with undoing.

isubastiCadmus:
[...]

PR #1354, opened 2017-10-31
fix backward selection always changed to forward

isubastiCadmus:
previously performing a change to backward selection will trigger two changes, the change itself and selection change from backward to forward selection. This PR fixes the updateSelection logic to make sure that it preserve the backward selection on changes.

fixes: https://github.com/ianstormtaylor/slate/issues/1308
fixes: https://github.com/ianstormtaylor/slate/issues/1355

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -131,6 +131,7 @@ class Content extends React.Component {
     const { editor } = this.props
     const { value } = editor
     const { selection } = value
+    const { isBackward } = selection
     const window = getWindow(this.element)
     const native = window.getSelection()
     const { rangeCount } = native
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/components/content.js
```diff
@@ -161,20 +162,33 @@ class Content extends React.Component {
...
+        return
+      }
+      if (
+        range.startContainer == current.endContainer &&
+        range.startOffset == current.endOffset &&
+        range.endContainer == current.startContainer &&
+        range.endOffset == current.startOffset
+      ) {
+        return
+      }
```
ianstormtaylor: It seems weird to be doing both of these checks here.

Is the problem that the `range` returned from `findDOMRange` is properly set forward/backward depending on the Slate selection? If so wouldn't the fix be to fix the `findDOMRange` util instead?

packages/slate-react/src/components/content.js
```diff
@@ -161,20 +162,33 @@ class Content extends React.Component {
...
 
     // Otherwise, set the `isUpdatingSelection` flag and update the selection.
     this.tmp.isUpdatingSelection = true
     native.removeAllRanges()
-    native.addRange(range)
+    if (isBackward) {
+      native.setBaseAndExtent(range.endContainer, range.endOffset, range.startContainer, range.startOffset)
+    } else {
+      native.setBaseAndExtent(range.startContainer, range.startOffset, range.endContainer, range.endOffset)
+    }
```
ianstormtaylor: And then if the util was fixed, wouldn't this code be unnecessary?

packages/slate-react/src/components/content.js
isubastiCadmus: @ianstormtaylor I tried those first but then it actually get the start and end node correctly but ```range.setStart()``` and ```range.setEnd()``` resulting in the same range object whether it's forward/backward. that's why this check is needed

packages/slate-react/src/components/content.js
isubastiCadmus: same explanation, the range object is always a forward range. seems to be Range API implementation issue

ianstormtaylor:
Thanks @isubastiCadmus!

We discovered (in Slack) that the native `Range` objects can never be "backwards/forwards", they just have a start/end that is always before/after in the document. So we have to always check both orientations.

PR #1356, opened 2017-10-31
Avoid rendering an empty placeholder by default

blakeembrey:
For some reason the rendering the empty placeholder `<span />` ends up making the caret render out of the element.

Before: 

image

After:

image


ianstormtaylor:
Thanks @blakeembrey!

PR #1357, opened 2017-10-31
Avoid executing hotkeys while composing

blakeembrey:
This PR avoids executing _any_ hotkey shortcuts while in `isComposing` mode. This is needed because, in `master`, when hitting something like `mod+z` while composing will result in _both_ an undo event and `z` being inserted. To repro, enter some text, then type with IME, hit `mod+z` and `mod+shift+z`. You should see the data before the IME input changing and the characters being appended to the current IME ASCII text.

/cc @doodlewind I want to verify the behaviour of your custom IME input, because I can't tell why `HOTKEYS.COMPOSING` should exist otherwise. If it's to handle that, I'm happy to restore it and tidy the code up differently.

### DIFF
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -371,11 +371,11 @@ function BeforePlugin() {
   function onKeyDown(event, change, editor) {
     if (editor.props.readOnly) return true
 
-    // When composing, these characters commit the composition but also move the
-    // selection before we're able to handle it, so prevent their default,
-    // selection-moving behavior.
-    if (isComposing && HOTKEYS.COMPOSING(event)) {
-      event.preventDefault()
+    // When composing, we need to prevent all hotkeys from executing while
+    // typing. However, certain characters also move the selection before
+    // we're able to handle it, so prevent their default behavior.
+    if (isComposing) {
+      if (HOTKEYS.COMPOSING(event)) event.preventDefault()
       return true
     }
 


ianstormtaylor:
I think the arrow key logic was there because sometimes when using an IME you'd arrow left/right to choose one of the options? And without it it would end up moving the selection left/right at the same time? But I don't use IME, so I might be ignorant here.

blakeembrey:
@ianstormtaylor It is used, but the event never leaves the selection by default. Preventing default doesn't actually do anything on any of the code, using the built-in macOS IME input. Possibly it's different with another OS and IME input combination, that's why I pinged @doodlewind since he knows more about it than I do.

Here's any example in this comment field: 



It does appear Draft.js does similar handling of arrow keys (noted to handle Safari: https://github.com/facebook/draft-js/blob/cd20b992d4b054d70f756c258e5e75e518995e12/src/component/handlers/composition/DraftEditorCompositionHandler.js#L99), but unfortunately it's been there ever since they open sourced the code so no way to tell why it was introduced. 

I'm going to test across the browsers I have on my mac now to see if it matches the comment. If it does or doesn't, I don't mind adding that code back anyway in addition to the blocking of undo/redo events.

blakeembrey:
Ok, definitely verified. I shouldn't have removed much of that 😄 It's actually the backspace and enter too, missed it last night.

blakeembrey:
@ianstormtaylor Good catch, fixed the PR!

ianstormtaylor:
Thanks @blakeembrey!

Issue #909, opened 2017-07-03
Spellcheck removes entire line when used with bold


Soletiq:
Go here http://slatejs.org/#/rich-text
Type a line in the editor, intentionally misspelling something "Paraggraphs are cool"
Make that line bold
Right click the misspelled word, correct it, and the entire line is removed except that one word.

It shouldn't be removing the entire line..

ianstormtaylor:
Hey @Soletiq can you add a video showing the issue? Thanks!

SebAshton:
@ianstormtaylor I've experience this too, so have created a gif of the issue @Soletiq describes:

![error](https://user-images.githubusercontent.com/881759/29023057-dd191964-7b63-11e7-8381-3f801c06dcd6.gif)

While trying to reproduce the error I also experienced an issue whereby the cursor was writing to two places in the document after a spell check:

![error2](https://user-images.githubusercontent.com/881759/29023113-36f13016-7b64-11e7-8075-9b9b1119e03d.gif)

danburzo:
More generally, in Chrome/macOS, correcting the spelling of a marked word:

1. Makes the corrected word lose its marks
2. Introduces a state inconsistency that breaks in all sorts of ways.

Some steps to repro:

1. Open the editor http://slatejs.org/#/rich-text
2. Misspell a word, then __bold it__
3. Right-click on the word and correct it — notice how it's introduced as plain text
4. Press Cmd + Z to undo

When correcting the first word in the paragraph:

<img width="1279" alt="screen shot 2017-08-17 at 16 31 35" src="https://user-images.githubusercontent.com/205375/29414804-927bb278-8369-11e7-89ae-123cd5c76ec0.png"> 

When correcting a word in the middle of the paragraph:

<img width="1279" alt="screen shot 2017-08-17 at 16 32 50" src="https://user-images.githubusercontent.com/205375/29414854-bcb30924-8369-11e7-96bf-ab2fad132255.png">

bengotow:
Hmm I'm seeing this too... In Chrome, the browser inserts the correction by running `document.execCommand('insertText', false, correction)` so we might be able to identify / test the issue using the behavior of `insertText`.

I'm fixing this in Mailspring (an Electron app) by implementing my own `contextmenu` event handler on the Editor component and then displaying a custom menu that performs Slate changes rather than native DOM changes, but I'm not sure if there's a general purpose fix.

ianstormtaylor:
I think this is fixed in the latest versions of Slate.

PR #1359, opened 2017-11-01
Fallback to last leaf node when inserting change

blakeembrey:
I swear I found the related issue before I discovered this issue and started working on fixing it, but I can't find it again now.

Before: 

After:

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -428,25 +428,25 @@ function AfterPlugin() {
     // Get the text node and leaf in question.
     const { document, selection } = value
     const node = document.getDescendant(point.key)
+    const block = document.getClosestBlock(node.key)
     const leaves = node.getLeaves()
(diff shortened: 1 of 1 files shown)


doodlewind:
Spellcheck/autocorrect is another pain point for now. Does this works for #909 ?

It seems to be hard since `onInput` fired **after** DOM is mutated, and `onBeforeInput` does not fire on spellcheck/autocorrect.

blakeembrey:
Unfortunately, no. But I'll look into it slowly as I get a grasp on how the codebase works. This only affects the changes that occur at the end of the line that would today crash the editor (because DOM offset is larger than leaf).

ianstormtaylor:
If https://github.com/facebook/react/issues/11211 was implemented, we'd actually have `onBeforeInput` fire in the case of spellcheck/autocorrect in modern browsers which would be nice!

@blakeembrey this looks great. Thank you!

PR #1361, opened 2017-11-01
Avoid changing selection when deleting old content

blakeembrey:
Previously, `.select(entire)` would end up changing the selection to the entire line. When undoing, it results in an odd experience of the entire line selected instead of just the change (or wherever you were before in the case of Firefox). This avoids moving the selection when changing the content.

Before:

After:


ianstormtaylor:
This is awesome, thanks @blakeembrey!

PR #1365, opened 2017-11-01
Fix broken iOS UA sniff that identified iOS as macOS

### LINKED ISSUE #1176: iOS Safari: Predictive text suggestions don't seem to recognize word breaks

rgrove:
UA sniffing stops at the first match when trying to determine the client's OS. Since the macOS sniff always ran before the iOS sniff and matched the string "mac os x", which is also present in iOS Safari's user agent string, `IS_MAC` was always `true` for iOS Safari and `IS_IOS` was always `false`.

The iOS sniff now runs before the macOS sniff, which prevents false positives. Existing uses of `IS_MAC` that should also apply to iOS have been updated to check for `IS_MAC || IS_IOS`.

This also re-fixes #1176 and #1177, which regressed when a recent change added `IS_IOS` checks that inadvertently prevented code from running on iOS.

### DIFF
--- a/packages/slate-react/src/constants/environment.js
+++ b/packages/slate-react/src/constants/environment.js
@@ -37,8 +37,8 @@ const EVENT_RULES = [
  */
 
 const OS_RULES = [
+  ['ios', /os ([\.\_\d]+) like mac os/i], // must be before the macos rule
   ['macos', /mac os x/i],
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Good catch, thanks @rgrove!

Issue #639, opened 2017-03-01
Deserialized HTML list is not correctly inserted with insertFragment


pavelvolek:
Deserializing a HTML snippet `<ul><li>b</li><li>c</li></ul>` and pasting it using `insertFragment` doesn't work as expected.

```
const { document } = serializer.deserialize('<ul><li>b</li><li>c</li></ul>');
state
  .transform()
  .insertFragment(document)
  .apply()
```

Seems that the HTML is correctly deserialized, but after inserting the fragment the first list item is pulled out of the list:

```
<li class="ListItem">
  <span data-key="9">
    <span data-offset-key="9-0">b</span>
  </span>
</li>
<ul class="List isUnordered">
  <li class="ListItem">
    <span data-key="11">
      <span data-offset-key="11-0">c</span>
    </span>
  </li>
</ul>
```

If the list is preceded for example by a paragraph (`<p>a</p><ul><li>b</li><li>c</li></ul>`), the fragment is inserted correctly.

Can be replicated in http://slatejs.org/#/paste-html.

kellyjosephprice:
I spent some time looking at this, and I'm not sure how to proceed. `insertFragment` tries really hard to insert the first block as inline. That makes a lot of sense for text, but it's the wrong direction for actual blocks, ie `ul`'s. So as to not break functionality, it seems it meet be worth it to add another transform like, `insertFragmentAsBlock`.

ianstormtaylor:
The `insertFragment` behavior of melding the inlines is by design, but I think in this case you should be able to use `insertBlock` instead.

PR #1366, opened 2017-11-01
Limit melding behavior in `insertFragmentAtRange`

### LINKED ISSUE #1364: Nested table broken on pasting
Bug

https://jsfiddle.net/doodlewind/qzvmpd9a/2/

1. Open Slate's [official table example](http://slatejs.org/#/tables), select **from first cell to last cell**.
2. Open JSFiddle page, paste it in.
3. First table cell is extracted out, table shape is broken.

See GIF, notice that `123` should have remained in first cell, instead of joining with text:

bug-paste-table

Meanwhile, if selection includes the paragraph before, pasting works fine:

good-paste-table

When user's selection merely includes one table, this table should remain its structure.

I've located this, PR on the way!

doodlewind:
Closes #917
Closes #1364 

According to #639, for now when inserting fragment, Slate tries to merge first block node into the block at cursor position. This is great if the fragment contains multi 'flat' blocks, while with a case that we insert one root block with multi child blocks (e.g. table), it will change the nested structure of the fragment.

This can be optimized with the detection of *single nested block*, and fallback to `insertBlockAtRange` in this case, which can be pretty straightforward.

For the table case, `firstBlock` and `lastBlock` are different table cells, so they are naturally different nodes. If the fragment has only one child but its `firstBlock` and `lastBlock` are different, we can confirm that the fragment is consist of one nested block.

As for test cases, I found that the fragment created by `slate-hyperscript` automatically injects a text node at the end of `fragment.nodes`, while with `slate-html-deserializer` the fragment generated doesn't follow this. So it's a bit confusing for me to add test cases if needed.

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -732,6 +732,13 @@ Changes.insertFragmentAtRange = (change, range, fragment, options = {}) => {
     return
   }
 
+  // If the fragment contains a single nested block, (e.g., table),
+  // simply insert this block to prevent the melding behavior
+  if (firstBlock != lastBlock && fragment.nodes.size == 1) {
+    change.insertBlockAtRange(range, fragment.nodes.first(), options)
+    return
+  }
+
   // If the first and last block aren't the same, we need to insert all of the
   // nodes after the fragment's first block at the index.
   if (firstBlock != lastBlock) {
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -732,10 +732,17 @@ Changes.insertFragmentAtRange = (change, range, fragment, options = {}) => {
     return
   }
 
-  // If the fragment contains a single nested block, (e.g., table),
-  // simply insert this block to prevent the melding behavior
-  if (firstBlock != lastBlock && fragment.nodes.size == 1) {
-    change.insertBlockAtRange(range, fragment.nodes.first(), options)
+  // If the fragment starts or ends with single nested block, (e.g., table),
+  // do not merge this fragment with existing blocks.
+  if (
+    firstBlock != lastBlock &&
+    Block.isBlock(fragment.nodes.first()) &&
+    Block.isBlock(fragment.nodes.last()) &&
+    (firstBlock != fragment.nodes.first() || lastBlock != fragment.nodes.last())
+  ) {
+    fragment.nodes.reverse().forEach((node) => {
+      change.insertBlockAtRange(range, node, options)
+    })
     return
   }
 


ianstormtaylor:
Hey @doodlewind, thank you for this, this is a great explanation.

I think I like this PR's solution potentially over the solution I described in https://github.com/ianstormtaylor/slate/issues/1367. Do you agree? 

I'm trying to think of cases where you want the Google Docs joining behavior for single-top-level-nested-block situations (as you describe in this PR) and I can't think of any. If anything maybe the Google Docs behavior of merging the lists only when you paste inside a paragraph is a bit unintuitive and we shouldn't copy it.

I think there might be one case that this current PR doesn't solve which we should add though if we're going to do this. Right now it solves the UX for fragments of...

```jsx
<paragraph />
```
```jsx
<paragraph />
<paragraph />
```
```jsx
<table>...</table>
```
```jsx
<paragraph />
<table>...</table>
<paragraph />
```

But it doesn't solve these cases:

```jsx
<paragraph />
<table>...</table>
```
```jsx
<table>...</table>
<paragraph />
```

...since the table in either of those cases will be broken by having its leading/trailing node merged into the destination.

It seems like we should tweak the logic to avoid merging the edge if the node at that edge is a nested-block node. Instead of only checking for the single-child case?

ianstormtaylor:
Actually @doodlewind I think I just thought of a counter-case. Consider the nested structure of quote blocks, where a fragment would be...

```jsx
<quote>
  <paragraph />
  <paragraph />
</quote>
```

This will be picked up as a non-merge situation, but you probably still expect this to behave the same way as the equivalent fragment without the `<quote>` wrapper. This might be evidence that https://github.com/ianstormtaylor/slate/issues/1367 is a more expected solution.

doodlewind:
@ianstormtaylor Good catch! I tweak a bit and it works! Use this [table in comment](https://github.com/ianstormtaylor/slate/pull/1353#issuecomment-340811532) as an example and now it works in these following cases:

``` html
<table>...<table/>
<p />
```

``` html
<p />
<table>...<table/>
```

``` html
<p />
<table>...<table/>
<table>...<table/>
```

``` html
<table>...<table/>
<table>...<table/>
<p />
```

``` html
<p />
<table>...<table/>
<table>...<table/>
<p />
```

ianstormtaylor:
Hey @doodlewind, I'm sorry for the delay here.

I think for these two cases:



``` html
<p />
<table>...<table/>
<table>...<table/>
```

``` html
<table>...<table/>
<table>...<table/>
<p />
```

We need to handle it by merging the `<p />` but not the `<table>...</table>`. But as far as I can tell, the way we're writing it right now it will not merge either. I think this is inconsistent, since the paragraph merging behavior is pretty standard. Can you make that fix?

Soreine:
> As for test cases, I found that the fragment created by slate-hyperscript automatically injects a text node at the end of fragment.nodes, while with slate-html-deserializer the fragment generated doesn't follow this. So it's a bit confusing for me to add test cases if needed.

That's what I'm trying to fix in https://github.com/ianstormtaylor/slate/pull/1379

zhujinxuan:
Hi, the question maybe a bit stupid, but I am just wondering why not
```
if (!fragment.nodes.size) {
  return
}
if (fragment.nodes.size === 1) {
  change.insertFragmentByKey(..., {normalize: false})
  return
}
const firstNode = fragment.nodes.first();
change.insertFragmentByKey(..., {normalize: false, regenerateKey: false});

const nextRange = Range.create().collapseToStartOf(document.getNextText(firstNode.getLastText().key))
const nextFragment = fragment.set("nodes", fragment.nodes.unshfit())
change.insertFragmentAtRange(nextRange, nextFragment, {normalize: false, regenerateKey: false})
```

zhujinxuan:
@Soreine @ianstormtaylor @doodlewind   Hi, I rewrite the insertFragmentAtRange at https://github.com/ianstormtaylor/slate/pull/1553;  I have added tested there and am sure that it can deal with table-nested block and undo behavior

However, there have to be a slight change about behaviors in `test/changes/at-current-range/insert-fragment/fragment-nested-blocks.js`

```
      <quote>
        <quote>
          one
        </quote>
        <quote>
          two
        </quote>
      </quote>
```
because slate cannot tell whether this quote is table.

ianstormtaylor:
@doodlewind any update on this one? I'd love to merge it if that last condition was fixed.

ianstormtaylor:
Going to merge this now and we can figure out the edge cases next! Thank you again @doodlewind, this is great.

PR #1369, opened 2017-11-02
fix scrolling on slate inside scrollable container that is not a window

isubastiCadmus:
previously findScrollerContainer always return window as scroller variable is assigned to window.
this fix the intended fix from https://github.com/ianstormtaylor/slate/pull/1165


ianstormtaylor:
Thanks @isubastiCadmus!

PR #1370, opened 2017-11-02
scroll to the correct position

isubastiCadmus:
previous scrolling fix works fine if we're typing at the end of the document. if the edited text is at the middle or beginning of the document, it's unusable.


ianstormtaylor:
@isubastiCadmus thank you very much!

PR #1371, opened 2017-11-02
Replace IME `setTimeout` with `requestAnimationFrame`

doodlewind:
Currently we use a plain `setTimeout` to delay `onCompositionEnd`, while this timeout may be leaky for some IMEs.

In [workaround of Draft.js](https://github.com/facebook/draft-js/blob/master/src/component/handlers/composition/DraftEditorCompositionHandler.js#L74) there's an 'magic' constant timeout, while it'll be more semantic and COMPAT-proof with the `requestAnimationFrame` approach.

Smoke tested on Chinese / Japanese / Korean / French and no regression found.


ianstormtaylor:
Thanks @doodlewind!

PR #1372, opened 2017-11-02
Fixed simulator handler argument order

ashutoshrishi:
The Simulator was calling the handlers with the `change` argument as the first argument instead of the `event`.


ianstormtaylor:
Thank you @ashutoshrishi!

PR #1373, opened 2017-11-02
Fix find-point.js when void node is input (in readOnly mode)

joehua87:
slate-find-node-error

Error occurs when render `<input />` in readOnly mode

```js
function Input(props: any) {
  return (
    <input
      onFocus={(e) => {
        e.preventDefault()
        e.stopPropagation()
        // console.log(e)
      }}
    />
  )
}
```

```js
<Editor
  placeholder="Enter a title..."
  renderNode={this.renderNode}
  readOnly
/>
```

```js
renderNode = (props: any) => {
  const { node } = props
  const Component = {
    input: Input,
  }[node.type]

  if (!Component) return undefined

  return <Component {...props} />
}
```


ianstormtaylor:
Thanks @joehua87!

PR #1375, opened 2017-11-03
Fix image example's onPaste

winfield:
> This example shows images in action. It features two ways to add images. You can either add an image via the toolbar icon above, or if you want in on a little secret, copy an image URL to your keyboard and paste it anywhere in the editor!

Paste is broken on slatejs.org. on Paste event, getEventRange(event) always returns null because there is no clientX and clientY from the Clipboard event.

### DIFF
--- a/examples/images/index.js
+++ b/examples/images/index.js
@@ -1,5 +1,5 @@
 
-import { Editor, getEventRange, getEventTransfer } from 'slate-react'
+import { Editor, getEventTransfer } from 'slate-react'
 import { Block, Value } from 'slate'
 
 import React from 'react'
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @winfield thanks for this! Won't the current change make the drop `target` not work anymore? Such that it will be dropping into the incorrect spot?

winfield:
Hey @ianstormtaylor It actually works (accidentally) because the ``FileReader``'s load listener happens after core plugin's ``onDrop``, so the target has been selected there.

I've changed the code so that ``onDropOrPaste`` only returns when it's a drop event & target is null

ianstormtaylor:
Awesome, thank you @winfield!

PR #1376, opened 2017-11-04
Disable dropping in inline void node example

### LINKED ISSUE #1192: Issue dropping content between two inline void nodes

doodlewind:
Fixes #1192.

For the emojis example, default dropping behavior may raise unexpected issue on corner cases. Simply `preventDefault` on drop fixes this.

This tweak does not affect current word DnD behavior in this example. When dropping on emojis, still can word inserted into text in correct order.

### DIFF
--- a/examples/emojis/index.js
+++ b/examples/emojis/index.js
@@ -18,6 +18,14 @@ const EMOJIS = [
 ]
 
 /**
+ * No ops.
+ *
+ * @type {Function}
+ */
+
+const noop = e => e.preventDefault()
+
+/**
  * The links example.
  *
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @doodlewind!

PR #1377, opened 2017-11-04
Fix `deleteLineBackward` when first node is void inline

### LINKED ISSUE #1199: Changes.deleteBackwardAtRange fails when first node is a void node
This problem is reproducible on http://slatejs.org/#/emojis.

Screencast: http://recordit.co/RUJliirLNR

Platform: macOS Sierra 10.12.6
Browser: Chrome 61.0.3163.100（Official Build)

Failure happens when a line begins with a void node + some texts, go to end of line, cmd + backspace. 

It fails at https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/changes/at-range.js#L428 when `node = document.getNextText(parent.key)` returns `undefined` on #L423.

Also, correct me if I'm wrong, but if I'm deleting backwards shouldn't the final focus be the one before the void node (if any)? So maybe `getPreviousSubling(parent.key)` instead?

doodlewind:
Fixes #1199 

`Change.deleteLineBackward` is generally called by `cmd-del`, which deletes the whole line before cursor. This method fails if first node in the line is an void inline node, e.g. `<emoji />`.

Reason for this issue is that when we calculate the offset to delete, the void inline node has an empty text child node, which disturbs the offset.

Say we'd like to delete such a line:

```
<p>
  <emoji />
  hello world
</p>
```

The actual structure is somewhat like following:

``` html
<p>
  <text key="1" />
  <emoji key="2">
    <text key="3"/>
  </emoji>
  <text key="4">hello world</text>
</p>
```

Currently when we attempt to find the beginning text node, we loop from end of the line, and eventually try finding previous text node of "hello world", which is `3` in this example. But the "actual" previous text node is `1`. This results to an error.

In this case we detect *whether first node in the line is an void inline node*, omit it in `deleteLineBackwardAtRange`, then perform a simple `deleteBackward` as a workaround. This approach does not mutate any underlying delete changes, so neither will it introduce too much complexity.

Test cases for `deleteLineBackward` was missing, this PR also complements them.

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -296,8 +296,26 @@ Changes.deleteLineBackwardAtRange = (change, range, options) => {
   const { startKey, startOffset } = range
   const startBlock = document.getClosestBlock(startKey)
   const offset = startBlock.getOffset(startKey)
-  const o = offset + startOffset
+  const startWithVoidInline = (
+    startBlock.nodes.size > 1 &&
+    startBlock.nodes.get(0).text == '' &&
+    startBlock.nodes.get(1).kind == 'inline'
+  )
+
+  let o = offset + startOffset
+
+  // If line starts with an void inline node, the text node inside this inline
+  // node disturbs the offset. Ignore this inline node and delete it afterwards.
+  if (startWithVoidInline) {
+    o -= 1
+  }
+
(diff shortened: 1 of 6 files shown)


ianstormtaylor:
Great PR @doodlewind, thank you very much!

PR #1379, opened 2017-11-06
slate-hyperscript: Fix extra empty text at the end of children

### LINKED ISSUE #1154: hyperscript: Extra text nodes when not using <state> wrapper
I think this is a bug.

If you declare a document like this:

```js
const h = createHyperscript({
  blocks: {
    paragraph: 'paragraph',
  },
})

const document = (
  <document>
    <paragraph>Some text</paragraph>
  </document>
);
```

You get an extra empty text after the `paragraph`:

```js
    {
      "data": {},
      "kind": "document",
      "nodes": [
        {
          "data": {},
          "kind": "block",
          "isVoid": false,
          "type": "paragraph",
          "nodes": [
            {
              "kind": "text",
              "ranges": [
                {
                  "kind": "range",
                  "marks": [],
                  "text": "Some text"
                }
              ]
            }
          ]
        },
        {
          "kind": "text",
          "ranges": [
            {
              "kind": "range",
              "marks": [],
              "text": ""
            }
          ]
        }
      ]
    }
```

There's no problem if you create a whole `<state>...</state>` thanks to state normalization.

Soreine:
This fixes #1154 

Extra empty texts were added in blocks. They are not visible if created using `<value>...</value>` because of normalization.

But sometimes I want to just build a document, and avoid normalization on purpose.

### DIFF
--- a/packages/slate-hyperscript/src/index.js
+++ b/packages/slate-hyperscript/src/index.js
@@ -184,13 +184,15 @@ function createChildren(children, options = {}) {
     node = next
   }
 
-  children.forEach((child) => {
+  children.forEach((child, index) => {
+    const isLast = index === children.length - 1
+
     // If the child is a non-text node, push the current node and the new child
     // onto the array, then creating a new node for future selection tracking.
     if (Node.isNode(child) && !Text.isText(child)) {
       if (node.text.length || node.__anchor != null || node.__focus != null) array.push(node)
       array.push(child)
-      node = Text.create()
+      node = isLast ? null : Text.create()
       length = 0
     }
 
@@ -230,7 +232,9 @@ function createChildren(children, options = {}) {
   })
 
   // Make sure the most recent node is added.
-  array.push(node)
+  if (node != null) {
+    array.push(node)
+  }
 
   return array
 }
--- a/packages/slate-hyperscript/test/index.js
+++ b/packages/slate-hyperscript/test/index.js
@@ -1,0 +1,30 @@
+/** @jsx h */
+
+import h from '../'
+import assert from 'assert'
+import { Document, Block, Text } from 'slate'
+
+
+describe('slate-hyperscript', () => {
+  it('should create a document with a single block', () => {
+    const output = (
+        <document>
+            <block type="paragraph">
+                Single block
+            </block>
+        </document>
+    )
+    const expected = Document.create({
+      nodes: [
+        Block.create({
+          type: 'paragraph',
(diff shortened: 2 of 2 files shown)


Soreine:
It seems to have fixed weird results in tests with selection (insert-fragment)
I don't know why this test still fails though.
https://github.com/ianstormtaylor/slate/blob/master/packages/slate/test/changes/at-current-range/insert-fragment/start-inline.js#L29

image

ianstormtaylor:
@Soreine were you able to figure out why that other test diverged?

Soreine:
I'm investigating right now. But as you can see, the selection after `insertFragment` is invalid. The text offset is 8 but the text node is 4 characters long.

Soreine:
Honestly, the test itself looks odd to me.

```jsx
export const input = (
  <value>
    <document>
      <paragraph>
        <link><cursor />word</link>
      </paragraph>
    </document>
  </value>
)

export const output = (
  <value>
    <document>
      <paragraph>
        fragment<link><cursor />word</link>
      </paragraph>
    </document>
  </value>
)
```

Where did the `<quote>`  block go ? And why the fragment text is inserted outside the current `<link>` ?

doodlewind:
@Soreine #1366 shows that slate tries to merge the fragment into cursor position, so first block in the fragment may be merged into `value.startBlock`. Does this behavior relate to the issue?

Soreine:
I understand that there are attempt to merge the fragment at the cursor position. I'm not versed into this part of Slate's logic, so I still haven't looked up your discussion in #1366 thoroughly.

In the case of our failing test, this is the code path taken:
https://github.com/ianstormtaylor/slate/blob/fix-slate-hyperscript-extra-text/packages/slate/src/changes/at-current-range.js#L159

Slate thinks that the text was inserted in the current text node, and tries to advance the selection. But it was not completely merged at cursor (start of block, not start of inline).

Soreine:
It's a shame that this can't be merge, simply because it reveals an erroneous test. Should we remove the test and open an issue for it?

ianstormtaylor:
@Soreine sounds good to me!

Soreine:
Most of the tests for insert-fragment-at-range make more sense now. I marked the `start-inline` test to be skipped, and documented the reason it fails. I have opened an issue for it https://github.com/ianstormtaylor/slate/issues/1754

ianstormtaylor:
Thanks @Soreine, this looks perfect now!

Issue #1032, opened 2017-08-25
Scroll-to-cursor is broken inside scrollable containers


jennaplusplus:
#### Do you want to request a *feature* or report a *bug*?
bug

#### What's the current behavior?
Set the editor to a fixed height and `overflow-y: auto` (standard procedure for scrolling content). Press enter several times to exceed this height and the and cursor disappears; a scrollbar appears but the editor does not scroll down. However, if you type a few characters (instead of just creating new lines with 0-1 characters each), the editor suddenly snaps down as expected. 

Demo (direct from example page):
![slate scroll gif](https://user-images.githubusercontent.com/13112590/29724731-741e86aa-897e-11e7-8473-5739ad3b3b8f.gif)

Slate JSFiddle with same issue: 
https://jsfiddle.net/f4150bbq/

This is not reproduceable in a simple JSFiddle with a contenteditable div:
https://jsfiddle.net/gz2wrkxz/1/

#### What's the expected behavior?

When content exceeds available height, scroll editor down to cursor (i.e., the behavior that happens when you type several characters in the demo above).

I've tried my best to figure out the cause of this, but I'm coming up short- could there be a problem with `scrollToSelection`? Seeing the expected behavior when typing several characters is also interesting; are we accidentally overriding some default browser behavior?

Mac OS 10.12.3, Chrome 60.0.3112.101, Slate current

YurkaninRyan:
@jennaplusplus I also struggled with this in my editor, have you found a solution?

The solution I came up with was using `componentWillReceiveProps` to compare the last focusBlock with the current focusBlock in a variety of ways.  If they were found different I would `scrollIntoView` the dom node

jennaplusplus:
@YurkaninRyan I haven't :( Do you mind sharing your `componentWillReceiveProps` method?

oyeanuj:
@jennaplusplus @YurkaninRyan I used @conorcussell's solution [here](https://github.com/ianstormtaylor/slate/issues/689#issuecomment-290417680). See if that works for you?

YurkaninRyan:
I wonder if something like this should go into core, or if there is another underlying issue here.  I tried debugging through this, but it seems in some cases it works, and in others it doesn't.  Potentially a browser compat issue.

conorcussell:
[...]

PR #1383, opened 2017-11-06
corrects overscroll issue inside scrollable containers (#1032)

RobertDaleSmith:
After recent fix to scrolling within scrollable containers, the editor now scrolls the scroller container to the end on every key press. Seems to be doing this because it is not subtracting the distance between the top of the window and the scrollable container. (I have mine within an absolute position modal.)

I fixed the issue for myself by subtracting that top distance from `top` var before calculating the final `y` to scroll to within `scroll-to-selection.js`.

### DIFF
--- a/packages/slate-react/src/utils/scroll-to-selection.js
+++ b/packages/slate-react/src/utils/scroll-to-selection.js
@@ -62,8 +62,6 @@ function scrollToSelection(selection) {
   let height
   let yOffset
   let xOffset
-  let scrollerYOffset;
-  let scrollerXOffset;
 
   if (isWindow) {
     const { innerWidth, innerHeight, pageYOffset, pageXOffset } = scroller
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -80,6 +82,9 @@ function scrollToSelection(selection) {
   const top = (backward ? rect.top : rect.bottom) + yOffset
   const left = (backward ? rect.left : rect.right) + xOffset
 
+  top -= scrollerYOffset;
```
isubastiCadmus: @RobertDaleSmith scrollerYOffset and scrollerXOffset never got assigned

ianstormtaylor:
Hey @RobertDaleSmith thanks for this. Can you fix the build failures?

isubastiCadmus:
just making sure. what's the effect of `-=` to `undefined` value since `scrollerXOffset` and `scrollerYOffset` will be undefined if the scroller is a window in this case

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -75,11 +77,16 @@ function scrollToSelection(selection) {
     height = offsetHeight
     yOffset = scrollTop
```
isubastiCadmus: might want to assign do it as
```
scollerRect = scroller.getBoundingClientRect()
yOffset = scrollTop - scrollerRect.top
xOffset = scrollLeft - scrollerRect.left
```
since `scrollerYOffset` and `scrollerXOffset` is only being used when `scroller != window`

ianstormtaylor:
Looks good, thanks @RobertDaleSmith!

PR #1386, opened 2017-11-08
IE 11 compatibility, use Selection.collapse/extend instead of Selection.setBaseAndExtent

### LINKED ISSUE #1385: IE 11 compatibility
bug

If you try to use the editor with IE 11, the editor crashes 

```
TypeError: Object doesn't support property or method 'setBaseAndExtent'
   at n.updateSelection (http://slatejs.org/build.prod.js:4994:3878)
   at n.componentDidUpdate (http://slatejs.org/build.prod.js:4994:3021)
   at commitLifeCycles (http://slatejs.org/build.prod.js:4807:27707)
   at n (http://slatejs.org/build.prod.js:4807:31678)
   at u (http://slatejs.org/build.prod.js:4808:1303)
   at c (http://slatejs.org/build.prod.js:4808:1793)
   at batchedUpdates (http://slatejs.org/build.prod.js:4808:4655)
   at qb (http://slatejs.org/build.prod.js:4807:3872)
   at ob (http://slatejs.org/build.prod.js:4807:3806)
   at sb.batchedUpdates (http://slatejs.org/build.prod.js:4808:19675)
```

The editor should not crash.

this compatibility issue have been introduced by #1354

IE11 does not support the Selection.setBaseAndExtent function
pull request is on the way

illesg:
fixes #1385  
by using Selection.collapse/extend, we can preserve the backward/forward selection direction, and they are supported on IE 11

EDIT: Selection.extend is not available in IE11...

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -134,7 +134,7 @@ class Content extends React.Component {
     const { isBackward } = selection
     const window = getWindow(this.element)
     const native = window.getSelection()
-    const { rangeCount } = native
+    const { rangeCount, anchorNode, focusNode } = native
 
     // If both selections are blurred, do nothing.
     if (!rangeCount && selection.isBlurred) return
@@ -142,7 +142,7 @@ class Content extends React.Component {
     // If the selection has been blurred, but is still inside the editor in the
     // DOM, blur it manually.
     if (selection.isBlurred) {
-      if (!this.isInEditor(native.anchorNode)) return
+      if (!this.isInEditor(anchorNode)) return
       native.removeAllRanges()
       this.element.blur()
       debug('updateSelection', { selection, native })
@@ -175,11 +175,16 @@ class Content extends React.Component {
     if (current) {
       if (
         (
+          !isBackward &&
+          focusNode == current.startContainer &&
           startContainer == current.startContainer &&
           startOffset == current.startOffset &&
           endContainer == current.endContainer &&
           endOffset == current.endOffset
-        ) || (
+        ) ||
+        (
+          isBackward &&
+          focusNode == current.endContainer &&
           startContainer == current.endContainer &&
           startOffset == current.endOffset &&
           endContainer == current.startContainer &&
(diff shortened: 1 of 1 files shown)


lxcid:
This fix is awesome, though I was wondering if we should conditional switch the path based on browser?

illesg:
Thank you :), 
I don't think we should, it works in the other browsers too, so I think it's simpler this way

illesg:
i just noticed that the backward selection through spans is buggy again, but in a different way than before: 
selection-bug

i will look into it today

isubastiCadmus:
@illesg might want to try putting the addRange inside the else statement of isBackward and see how it goes?

illesg:
@isubastiCadmus yes, you're right, addRange is the problem
i went with the Selection.collapse/extend for both condition solution, it seems it works with all browsers, thanks

isubastiCadmus:
@illesg no problem. It was kinda my bad as well putting a fix with setBaseAndExtend as I forgot that it's not available in all browser

illesg:
ok it's weird, it looks like IE does not need the selection restoration at all, (and doesn't support the Selection.extend either) 
and by the way there was another bug in the current implementation with Chrome and Safari:
selection-bug-2

i fixed it by checking the selection direction, and the focusNode
let me know if there is a problem with this implementation, i'm pretty new here

packages/slate-react/src/components/content.js
```diff
@@ -192,14 +197,23 @@ class Content extends React.Component {
...
+      // COMPAT: Since the DOM range has no concept of backwards/forwards
+      // we need to check and do the right thing here.
+      if (isBackward) {
+        native.collapse(range.endContainer, range.endOffset)
+        native.extend(range.startContainer, range.startOffset)
+      } else {
+        native.collapse(range.startContainer, range.startOffset)
+        native.extend(range.endContainer, range.endOffset)
+      }
+    } else if (selection.isCollapsed) {
```
ianstormtaylor: I'm not sure about this check, because won't this mean that in IE11 when the selection is expanded we won't be updating it at all? This seems like it won't handle programmatic selection changes that result in an expanded selection.

packages/slate-react/src/components/content.js
```diff
@@ -175,11 +175,16 @@ class Content extends React.Component {
     if (current) {
       if (
         (
+          !isBackward &&
+          focusNode == current.startContainer &&
```
ianstormtaylor: What was the need for these updates to these conditionals? The added `isBackward` check makes sense to me, and seems like a smart improvement. The other one I'm not sure I understand.

illesg: Sorry, i messed up that part, i thought it's a bug, but it isn't, the native textarea works the same way:



select text backward by clicking on a text, and select with shift+click, and continue the selection with shift + right arrow

I will clean it up today

packages/slate-react/src/components/content.js
illesg: it's because I introduced a weird bug in IE with the previous change (by checking the direction and the focus node, it's not necessary anymore) 
change is on the way

ianstormtaylor:
Thanks @illesg looks good!

lxcid:
@illesg, this broke Firefox on my side though. This is an old bug. The previous implementation actually fixed it.

In my implementation, I had multiple `<Editor />`, when user press the up arrow, I'll unselect the current one and select the previous one. This cause firefox to just crash. The crash log shows the following:

screenshot 2017-11-17 15 01 43

I'll try to replicate it in jsfilddle and open a new issue regarding this.

PR #1388, opened 2017-11-08
Render text inside void nodes with a zero-width space.

AlbertHilb:
Render text inside void nodes with a zero-width space.
So the node can contain selection but the text is not visible.


ianstormtaylor:
Thank you @AlbertHilb, good call.

PR #1396, opened 2017-11-10
Add support to iframe

### LINKED ISSUE #1390: Slate vanishes with error: Unable to find DOM node...
bug

Implementing the Slate Text Editor inside an **iframe**, when calling `this.setState({ value })` like on the following example the editor vanishes with the error message:
`Uncaught Error: Unable to find a DOM node for "3". This is often because of forgetting to add 'props.attributes' to a custom component.` But I don't have any custom component as you can see in the example:

```js
class App extends React.Component {

  state = {
    value: initialValue
  }

  onChange = ({ value }) => {
    this.setState({ value })
  }

  render() {
    return (
      <Editor
        value={this.state.value}
        onChange={this.onChange}
      />
    )
  }

}
```

I was expecting the editor to call the function and deal with the result properly :)

Thank you! :)

joaomlap:
In findDOMNode util function the global window was being used, by passing down window from Content we assure it works inside an iframe.
Fixes #1390

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -154,7 +154,7 @@ class Content extends React.Component {
 
     // Otherwise, figure out which DOM nodes should be selected...
     const current = !!rangeCount && native.getRangeAt(0)
-    const range = findDOMRange(selection)
+    const range = findDOMRange(selection, window)
 
     if (!range) {
       logger.error('Unable to find a native DOM range from the current selection.', { selection })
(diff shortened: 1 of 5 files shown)


sethmcleod:
Just tested this in my codebase and I can confirm it fixes the #1390 issue. The window context is now correctly passed down while inside an iframe and this doesn't seem to break any of the other examples.

ianstormtaylor:
Hey @joaomlap I think this might be missing a usage of `findDOMNode` in `./plugins/after.js`? Can you double check?

joaomlap:
Hey @ianstormtaylor yes it was missing, thanks :)

ianstormtaylor:
Thanks @joaomlap!

J22Melody:
What about adding a default parameter value that is the global `window` so that we don't need to pass it most of the time we call `findDOMNode`?

ianstormtaylor:
@J22Melody sounds great to me, can you PR it?

Issue #1189, opened 2017-09-27
Error inserting text with hanging selection (slatejs.org)


echenley:
#### Do you want to request a *feature* or report a *bug*?

Bug.

#### What's the current behavior?

![2513bc427fe3f3226b94fbffcd2bd27e](https://user-images.githubusercontent.com/6344219/30940285-2a591b24-a3a6-11e7-97fa-8a7832d6e810.gif)

1. Triple click to select entire block
3. Type or paste some text.

It seems to have something to do with the selection wrapping to the beginning of the next block.

echenley:
Did a little bit of digging today. It seems to have to do with the isHanging special case in deleteAtRange. The target block for the paste is being deleted because of this.

I can think of two non-ideal options to fix it:

1. Use the same behavior regardless of isHanging, i.e. merging the next block into the first. This is not ideal because if the second block has a different type, it will be converted to the first type, even though the selection doesn't *appear* to cover the second block. However, this may be technically "correct".

2. Pretend that the selection is not hanging during a paste, i.e. change endKey and endOffset to cover the previous block so it is replaced instead of deleted.

I have a feeling this is problematic for all rich text editors (not just Slate). Here is similar buggy behavior in Apple Mail:

![41a250fa6a9b2933c966b49fac2f7343](https://user-images.githubusercontent.com/6344219/31256226-c8e7a142-a9f6-11e7-8d0c-8d54a4cdcc63.gif)

echenley:
A hacky but effective fix:

```js
function handleKeyDown(e, data, change) {
  // ...
  if (state.startKey !== state.endKey && state.endOffset === 0) {
    const endNode = state.document.getFurthestBlock(state.endKey)
    const prevNode = state.document.getPreviousSibling(endNode.key)
    change.extendToEndOf(prevNode)
  }
  // ...
}
```

ianstormtaylor:
I think this is due to [these lines of code](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/changes/at-range.js#L228-L234) where the start block will actually be removed in the hanging case, so when it goes to insert text into it it's no longer there.

czechdave:
Managed to figure out a possible fix to this issue. Let me know if it's a reasonable PR or if it needs any additional work or different approach @ianstormtaylor 

https://github.com/ianstormtaylor/slate/pull/1432

rubyonrailsStar:
Hello @czechdave 

I have some questions.
[...]

PR #1398, opened 2017-11-13
Fix "Could not find a node with key ${key}"

pizn:
This PR is to fix the case like this:

kapture 2017-11-13 at 11 34 50
---

Hi, @ianstormtaylor , when I insertText, it will throw the error "Could not find a node with key ${key}". Because it have selection when insertText (should remove selection content and insert new text), but the `deleteAtRange` function remove the startBlock, so it lost the current node and throw error when `insertText`.  

I think `merge` is better then `remove`, but I'm not sure if there are any other scenarios.

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -239,13 +239,7 @@ Changes.deleteAtRange = (change, range, options = {}) => {
         change.moveNodeByKey(endBlock.key, startParent.key, startParentIndex + 1)
       }
 
-      // If the selection is hanging, just remove the start block, otherwise
-      // merge the end block into it.
-      if (isHanging) {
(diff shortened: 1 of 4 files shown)


ianstormtaylor:
Hey @pizn, thanks!

I don't think we should be changing the test cases though, those were there for a reason to preserve the hanging block, instead of merging it into the previous. I'd be open to a solution that doesn't need to modify those tests.

czechdave:
This is related to the issue #1189 that I just submitted a PR fix for https://github.com/ianstormtaylor/slate/pull/1432

ianstormtaylor:
Fixed by #1432

rubyonrailsStar:
Hello @pizn 

I have some problems.
Using SlateJS,  when delete one character such as "u", can I delete the character all in the content?

PR #1399, opened 2017-11-13
fix calling onChange handler after Editor component unmount

### LINKED ISSUE #1395: Editor onChange is called after unmount

Bug?

After mounting the `Editor` component from slate-react, if it unmounts very quickly, onChange is  called after unmount. Because I use `setState` in the change handler, it results in this warning:

> Warning: Can only update a mounted or mounting component. This usually means you called setState, replaceState, or forceUpdate on an unmounted component. This is a no-op.

https://jsfiddle.net/g0ghks7q/

I know it's not ideal to unmount the editor almost immediately, but this happens in our app because of how it's built, and it is somewhat difficult to change.

Could the unmount be handled internally in the Editor component, so that it would not call the onChange handler after unmount?

danielsarin:
This should fix #1395.

This is related to #1324. The `flushChange()` function in the Editor component sets a timeout for calling onChange, which means it can be called after the component has unmounted.

I added a `componentWillUnmount()` handler which simply clears the timeout if it exists.

I did some local testing by modifying one of the examples, and it seemed to work as expected. But I'm not sure how to write a test case for this, if it is needed...

### DIFF
--- a/packages/slate-react/src/components/editor.js
+++ b/packages/slate-react/src/components/editor.js
@@ -158,6 +158,17 @@ class Editor extends React.Component {
   }
 
   /**
+   * When the component unmounts, clear flushTimeout if it has been set
+   * to avoid calling onChange after unmount.
+   */
+
+  componentWillUnmount = () => {
+    if (this.tmp.flushTimeout) {
+      clearTimeout(this.tmp.flushTimeout)
+    }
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @danielsarin, looks good!

Issue #1285, opened 2017-10-25
Uncaught TypeError: Cannot read property 'marks' of undefined


chemzqm:
#### Do you want to request a *feature* or report a *bug*?

Bug of slate 0.27.4 with react 16

#### What's the current behavior?

Previous text node could be empty, so it will throws and the entire editor get crashed.

#### What's the expected behavior?

Should never crash.

The solution is:
``` js
diff --git a/packages/slate/src/models/node.js b/packages/slate/src/models/node.js
index 43d93d8a..549b37a0 100644
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1057,14 +1057,14 @@ class Node {
       const previous = this.getPreviousText(startKey)
       if (!previous || previous.text.length == 0) return []
       const char = previous.characters.get(previous.text.length - 1)
-      return char.marks.toArray()
+      return char ? char.marks.toArray() : []
     }
 
     // If the range is collapsed, check the character before the start.
     if (range.isCollapsed) {
       const text = this.getDescendant(startKey)
       const char = text.characters.get(range.startOffset - 1)
-      return char.marks.toArray()
+      return char ? char.marks.toArray() : []
     }
 
     // Otherwise, get a set of the marks for each character in the range.
@@ -1094,14 +1094,14 @@ class Node {
       const previous = this.getPreviousText(startKey)
       if (!previous || !previous.length) return []
       const char = previous.characters.get(previous.length - 1)
-      return char.marks.toArray()
+      return char ? char.marks.toArray() : []
     }
 
     // If the range is collapsed, check the character before the start.
     if (range.isCollapsed) {
       const text = this.getDescendant(startKey)
       const char = text.characters.get(range.startOffset - 1)
-      return char.marks.toArray()
+      return char ? char.marks.toArray() : []
     }
 
     // Otherwise, get a set of the marks for each character in the range.

```

YurkaninRyan:
I bet this was fixed in 0.27.5 by #1268 

Can you try that out @chemzqm ?

chemzqm:
@YurkaninRyan It's not, I tried with latest master branch, same problem exists

![](https://gfycat.com/LateClutteredGentoopenguin)

This error happens when I move the cursor to the end of an inline tag which positioned at the end of a block tag.

PR #1400, opened 2017-11-13
inline bug fixes related to #1285

RobertDaleSmith:
After updating to master I am still having this error happen. (#1285) But now only happens to me when I have an inline void image node inside of an inline link node, and I attempt to use arrow keys to move back and forth over the edges, then the uncaught errors fire.

I applied @chemzqm's suggested fix and it resolved that problem but then I started getting another uncaught error within `find-dom-range.js` because the offsets were coming back as `-1`. I fixed this one by just checking if the offset is greater than or equal to zero.

> Not sure if best fix for issue I am hitting but this got me up and running without crashing the browser when using inline void image nodes within inline link nodes.

### DIFF
--- a/packages/slate-react/src/utils/find-dom-range.js
+++ b/packages/slate-react/src/utils/find-dom-range.js
@@ -20,8 +20,8 @@ function findDOMRange(range) {
   const r = window.document.createRange()
   const start = isBackward ? focus : anchor
   const end = isBackward ? anchor : focus
-  r.setStart(start.node, start.offset)
-  r.setEnd(end.node, end.offset)
+  r.setStart(start.node, start.offset >= 0 ? start.offset : 0)
+  r.setEnd(end.node, end.offset >= 0 ? end.offset : 0)
   return r
 }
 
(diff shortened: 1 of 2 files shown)


packages/slate-react/src/utils/find-dom-range.js
```diff
@@ -20,8 +20,8 @@ function findDOMRange(range) {
   const r = window.document.createRange()
   const start = isBackward ? focus : anchor
   const end = isBackward ? anchor : focus
-  r.setStart(start.node, start.offset)
-  r.setEnd(end.node, end.offset)
+  r.setStart(start.node, start.offset >= 0 ? start.offset : 0)
+  r.setEnd(end.node, end.offset >= 0 ? end.offset : 0)
```
ianstormtaylor: Hey @RobertDaleSmith can you fix these inside `findDOMPoint` instead so that it doesn't return `-1` in these cases, instead of at this layer above? I think that will be the most bulletproof for the future. Thanks!

ianstormtaylor:
I think this PR is taken care of now from the two others that do similar things. Thank you @RobertDaleSmith!

PR #1412, opened 2017-11-16
fix some typos in docs/guides

tobiasandersen:
I just got around to read the new Guides section, and wow was I'm impressed! I fixed some typos while reading (and my editor apparently stripped some hanging whitespace as well, let me know if that's annoying and I'll add them back).

### DIFF
--- a/docs/guides/rendering.md
+++ b/docs/guides/rendering.md
@@ -3,14 +3,14 @@
 
 One of the best parts of Slate is that it's built with React, so it fits right into your existing application. It doesn't re-invent its own view layer that you have to learn. It tries to keep everything as React-y as possible.
 
-To that end, Slate gives you control over the rendering behavior of every node and mark in your document, any placeholders you want to render, and even the top-level editor itself. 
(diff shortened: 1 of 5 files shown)


docs/guides/rendering.md
```diff
@@ -29,9 +29,9 @@ function renderNode(props) {
 }
 ```
 
-> 🤖 Be sure to mix in `props.attributes` and render `props.children` in your node components! The attributes required for utilities like Slate's `findDOMNode`, and the children are the actual text content of your nodes.
+> 🤖 Be sure to mix in `props.attributes` and render `props.children` in your node components! The attributes are required for utilities like Slate's `findDOMNode`, and the children are the actual text content of your nodes.
 
-Or course, you don't have to use simple HTML elements, you can use your own custom React components too:
```
tobiasandersen: I assume it should say "Of course", but I opted to remove it instead of fixing it, since I generally don't like phrases like "of course", "obviously" etc. in guides like this. It easily makes you feel a bit stupid as a beginner. Other than that, I think the general tone and language was absolutely on point, great work!

ianstormtaylor:
Thank you @tobiasandersen this is awesome.

PR #1421, opened 2017-11-19
Managing History limits

dmitrizzle:
Undos are measured in `.size` rather than `.length`; previously Slate would allow for unlimited undos _(see screenshot below)_. This patch fixes that.

I am looking for help to implement an API that would allow the user to control the max history step size. It's very easy for me to replace `100` with a variable, but I don't understand how the `options:{}` work - where the values come from and whether the user has any control over it.

Should it be like `value.history.undos.set({max: 20})`?
Your insights are appreciated :)

***
Unlimited history stack bug

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "lint": "eslint packages/*/src packages/*/test examples/*/*.js examples/dev/*/*.js",
     "open": "open http://localhost:8080/dev.html",
     "release": "yarn run test && yarn run lint && lerna publish && yarn run gh-pages",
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Hey @dmitrizzle, good catch.

For changing the `100` value, it should be an argument in the `object` passed to `History.fromJSON` static method. And then when people create a Slate value, they should also add in a history with it, or call `setValue({ history })` I believe. But that should be in a different pull request I think.

Thank you!

PR #1424, opened 2017-11-20
Improve getFurthestAncestor

### LINKED ISSUE #1136: refactor `getFurthestAncestor` to use `getAncestors`
I think these might be performance improvements once you factor in memoization, it would need to be checked...

Right now `getFurthestAncestor` uses `this.nodes.find` directly instead of using `this.getAncestors` under the covers.

Soreine:
Fix #1136 
I added a benchmark. The performances should be the same, or better once we take memoization in account.

I wanted to use `getAncestors` to implement `getFurthestAncestor`. But `getFurthestAncestor` is misleading in its implementation, because it returns the descendant itself if it has no ancestors, unlike the other `getFurthest*` methods. This is because it was called `getHighestChild` before.

So we can't use `getAncestors` directly because, right now:

```jsx
- a
  - b

a.getAncestors('b')  -> List []
a.getFurthestAncestor('b') -> b
```

We can't change all ancestor methods to include the node itself. This would break a lot of things and the naming would be misleading.

So I think we should fix `getFurthestAncestor` to be consistent...

# Fix `getFurthestAncestor`

We fix `getFurthestAncestor` to not include the descendant itself:

```jsx
- a
  - b
    - c

a.getFurthestAncestor('b') -> null
a.getFurthestAncestor('c') -> b
```

In this case we should also implement `getDescendants` which is the same as `getAncestors` but returns the descendant node itself, so that we can reimplement the old `getHighestChild` as `getHighestDescendant`. 

```jsx
- a
  - b
    - c

a.getAncestors('c') -> List [a, b]
a.getDescendants('c') -> List [b, c]
```

All ancestors methods would then rely on `getDescendants` implementation. Only `getDescendants` would be memoized, not `getAncestors`. Because `getAncestors` is simply written as:

```jsx
getAncestors(key) {
  const descendants = this.getDescendants(key)
  return descendants
    ? null
    : descendants.butLast().unshift(this)
}
```

### DIFF
--- a/packages/slate/benchmark/models/get-furthest-ancestor.js
+++ b/packages/slate/benchmark/models/get-furthest-ancestor.js
@@ -1,0 +1,31 @@
+/** @jsx h */
+/* eslint-disable react/jsx-key */
+
+import h from '../../test/helpers/h'
+import { __clear } from '../../lib/utils/memoize'
+
+export default function ({ value, text }) {
+  value.document.getFurthestAncestor(text.key)
+}
+
+export function before(value) {
+  const text = value.document.getLastText()
+  __clear()
+  return { value, text }
+}
+
+export const input = (
+  <value>
+    <document>
+      {Array.from(Array(10)).map(() => (
+        <quote>
+          <paragraph>
+            <paragraph>
+              This is editable <b>rich</b> text, <i>much</i> better than a textarea!
+            </paragraph>
+          </paragraph>
+        </quote>
+      ))}
+    </document>
+  </value>
+)
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -788,12 +788,7 @@ class Node {
    */
 
   getFurthestAncestor(key) {
-    key = assertKey(key)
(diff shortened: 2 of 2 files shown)


Soreine:
There is a lot of methods on `Node` already. I would not want to add complexity with new methods... @ianstormtaylor if you have any better idea...

ianstormtaylor:
@Soreine this sounds very smart to me!

Instead of `getHighestDescendant` I think we should keep it as `getFurthestDescendant` to match the other cases?

Soreine:
`getFurthestDescendant` sounds okay, except that we should make clear that "furthest" still means "highest".

Soreine:
Coming back to it, it confuses me because when I read "document.getFurthestDescendant('A')", I understand "get the furthest descendant of A in this document"...

ianstormtaylor:
Hey @Soreine, I feel you on that. I'd like to keep `furthest` as the names for now, to match the others, since it's the antonym to `closest`.

Soreine:
Closing this PR, since it will be outperformed by more recent PRs about performance.

PR #1427, opened 2017-11-22
Fixing Editor validation/normalize after schema prop change.

### LINKED ISSUE #1426: Editor doesn't immediately normalize/validate after changing schema.
bug

Upon receiving a new schema, the editor does not normalize/validate according to the new schema's validation rules. Instead, it relies on the previous schema.

JSFiddle: https://jsfiddle.net/fj9dvhom/14/

GIF: http://g.recordit.co/GfBFoB3OKh.gif

If i'm not mistaken, the editor should normalize/validate **upon** getting a new schema rather than **after**.

potatopotaro:
Upon reassigning the schema prop for an Editor, the Editor would use the previous schema's validation rules for the single current render. Thus the Editor's schema validation/normalize was always one render behind.

The solution was to create a copy of the state in componentWillReceiveProps and mutate that copy directly, as setState's delay was the problem. However, I believe my solution could be improved upon as I'm still an amateur with React and Slate.

Edit: Fixes #1426

### DIFF
--- a/packages/slate-react/src/components/editor.js
+++ b/packages/slate-react/src/components/editor.js
@@ -112,6 +112,11 @@ class Editor extends React.Component {
   componentWillReceiveProps = (props) => {
     let { schema, stack } = this
 
+    // this.state is not immediately changed after this.setState so we will make
+    // a copy of "this" in order to mutate its state directly and pass it in to
+    // stack.run('onChange', change, thisCopy) below.
+    const thisCopy = this
+
     // Increment the updates counter as a baseline.
     this.tmp.updates++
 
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/components/editor.js
```diff
@@ -121,6 +126,11 @@ class Editor extends React.Component {
       const plugins = this.resolvePlugins(props.plugins, props.schema)
       stack = Stack.create({ plugins })
       schema = Schema.create({ plugins })
+
+      // There may be a better alternative to the 2 lines below.
+      thisCopy.state.stack = stack
+      thisCopy.state.schema = schema
```
ianstormtaylor: I believe since everything is by reference that these are the same as acting directly on `this.` instead. So I think the `thisCopy` is redundant. But I'm not sure if it's a good idea or not, or if there's another solution.

zhujinxuan:
Shall we close this PR?  As the problem is solved in #1975

PR #1417, opened 2017-11-17
Revert #1383 "corrects overscroll issue inside scrollable containers"

erquhart:
Related to #1032, #1382.

This reverts commit b973580c54be4cfddd3e7b61e6c7f4222d6ee5b2 due to regression of scroll behavior outlined in #1416.

Closes #1416.

PR #1428, opened 2017-11-22
Refactor scrollIntoView logic

### LINKED ISSUE #1416: Editor scrolls on every keypress when inside a scrollable container

bug

When the editor is within a scrollable container (that isn't the window), scroll position jumps on every keypress.

Regression introduced in #1383.

https://jsfiddle.net/fj9dvhom/6/

slate-scroll-issue

No scroll jumping.

echenley:
Tested on OS X 10.12 in Safari 11, Firefox 57, and Chrome 62, with both horizontal and vertical overflow, for both window and a scrollable wrapper div, with and without borders.

Fixes #1416 and #1032.

Also sort of fixes an issue where `Range.getBoundingClientRect()` returns 0s in Safari, causing the scroll behavior to fail.

Known caveat:

1. Will not work (does nothing) if `document.body` is actually the scrollable element, in which case `window.scrollTo()` is incorrect. I believe this is only the case if the height is explicitly set via css:

```
body {
  height: 100vh;
}
```

Unfortunately `document.body.scrollTop` does not work correctly in Chrome (always returns 0), so you have to use `document.documentElement` instead. Trying to handle this case was somewhat agonizing, but if this scenario needs to be supported, I can revisit it.

### DIFF
--- a/packages/slate-react/src/utils/scroll-to-selection.js
+++ b/packages/slate-react/src/utils/scroll-to-selection.js
@@ -20,13 +20,13 @@ const OVERFLOWS = [
  * @param {el} Element
  */
 
-function findScrollContainer(el) {
-  const window = getWindow(el)
+function findScrollContainer(el, window) {
   let parent = el.parentNode
   let scroller
 
   while (!scroller) {
     if (!parent.parentNode) break
+
     const style = window.getComputedStyle(parent)
     const { overflowY } = style
 
@@ -38,7 +38,14 @@ function findScrollContainer(el) {
     parent = parent.parentNode
   }
 
-  if (!scroller) return window
+  // COMPAT: Because Chrome does not allow doucment.body.scrollTop, we're
+  // assuming that window.scrollTo() should be used if the scrollable element
+  // turns out to be document.body or document.documentElement. This will work
+  // unless body is intentionally set to scrollable by restricting its height
+  // (e.g. height: 100vh).
+  if (!scroller) {
+    return window.document.body
+  }
 
   return scroller
 }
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -38,11 +37,64 @@ function findScrollContainer(el) {
     parent = parent.parentNode
   }
 
-  if (!scroller) return window
```
echenley: You have to explicitly set html and body to `overflow: visible /* or none */` in order for this to be hit. In the browsers I tested, the default value for both is `auto`.

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -53,48 +105,34 @@ function scrollToSelection(selection) {
...
-  let height
-  let yOffset
-  let xOffset
+  let selectionRect = range.getBoundingClientRect()
+
+  // range.getBoundingClientRect() returns 0s in Safari when range is collapsed.
+  // It appears to be a long-standing WebKit issue that was eventually fixed in Chrome.
+  // In this case, expanding the range by 1 is an effective workaround.
+  // https://bugs.webkit.org/show_bug.cgi?id=138949
+  // https://bugs.chromium.org/p/chromium/issues/detail?id=435438
```
echenley: You can see that this is currently broken in Safari on slatejs.org.

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -53,47 +55,108 @@ function scrollToSelection(selection) {
...
   const range = selection.getRangeAt(0)
-  const rect = range.getBoundingClientRect()
+  let selectionRect = range.getBoundingClientRect()
+
+  // range.getBoundingClientRect() returns 0s in Safari when range is collapsed.
+  // It appears to be a long-standing WebKit issue that was eventually fixed in Chrome.
+  // Expanding the range by 1 is a relatively effective workaround for vertical
+  // scroll, although horizontal may be off by 1 character.
+  // https://bugs.webkit.org/show_bug.cgi?id=138949
+  // https://bugs.chromium.org/p/chromium/issues/detail?id=435438
```
echenley: https://bugs.webkit.org/show_bug.cgi?id=138949
https://bugs.chromium.org/p/chromium/issues/detail?id=435438

isubastiCadmus:
@ianstormtaylor this PR looks more complete but at minimum https://github.com/ianstormtaylor/slate/pull/1417 is needed to fix overscroll issue

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -38,7 +38,9 @@ function findScrollContainer(el) {
     parent = parent.parentNode
   }
 
-  if (!scroller) return window
+  if (!scroller) {
```
ianstormtaylor: Can you add a `COMPAT:` comment here for future reference?

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -53,41 +55,101 @@ function scrollToSelection(selection) {
   if (!selection.anchorNode) return
 
   const window = getWindow(selection.anchorNode)
-  const scroller = findScrollContainer(selection.anchorNode)
-  const isWindow = scroller == window
+  const scroller = findScrollContainer(selection.anchorNode, window)
+  const isWindow =
+    scroller == window.document.body ||
+    scroller == window.document.documentElement
```
ianstormtaylor: Can you put this on a single line?

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -53,41 +55,101 @@ function scrollToSelection(selection) {
...
 
-  const x = left < yOffset || (width + xOffset) < left
-    ? left - width / 2
-    : xOffset
+  if (selectionLeft < xOffset) {
+    // selection to the left of viewport
+    x = selectionLeft
+  } else if (
+    selectionLeft + selectionRect.width + scrollerBordersX >
+    xOffset + width
```
ianstormtaylor: Can you put this on a single line?

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -53,41 +55,101 @@ function scrollToSelection(selection) {
...
+    // selection to the right of viewport
+    x = selectionLeft + scrollerBordersX - width
+  }
+
+  if (selectionTop < yOffset) {
+    // selection above viewport
+    y = selectionTop
+  } else if (
+    selectionTop + selectionRect.height + scrollerBordersY >
+    yOffset + height
```
ianstormtaylor: Can you put this on a single line?

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -53,41 +55,101 @@ function scrollToSelection(selection) {
...
+    scroller == window.document.documentElement
   const backward = isBackward(selection)
+
   const range = selection.getRangeAt(0)
-  const rect = range.getBoundingClientRect()
+  let selectionRect = range.getBoundingClientRect()
+
+  // range.getBoundingClientRect() returns 0s in Safari when range is collapsed.
+  // Expanding the range by 1 is a relatively effective workaround for vertical
+  // scroll, although horizontal may be off by 1 character.
```
ianstormtaylor: Can you add `COMPAT:` in front of this comment?

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -53,41 +55,101 @@ function scrollToSelection(selection) {
...
+      borderTopWidth,
+      borderBottomWidth,
+      borderLeftWidth,
+      borderRightWidth,
+    } = window.getComputedStyle(scroller)
+
+    const scrollerBorderTop = parseInt(borderTopWidth, 10)
+    const scrollerBorderBottom = parseInt(borderBottomWidth, 10)
+    const scrollerBorderLeft = parseInt(borderLeftWidth, 10)
+    const scrollerBorderRight = parseInt(borderRightWidth, 10)
```
ianstormtaylor: If these aren't used anywhere else, can we just inline them in the variables below instead?

maddrag0n:
Please merge this asap… this is a blocking issue for us right now. 👍

@echenley Also let me know if you need a hand to add the changes to get this merged.

packages/slate-react/src/utils/scroll-to-selection.js
echenley: Added. Does that seem appropriate to you?

packages/slate-react/src/utils/scroll-to-selection.js
echenley: Fixed.

packages/slate-react/src/utils/scroll-to-selection.js
echenley: Sure, updated.

packages/slate-react/src/utils/scroll-to-selection.js
echenley: Fixed.

packages/slate-react/src/utils/scroll-to-selection.js
echenley: Fixed.

echenley:
@ianstormtaylor Updated, thanks!

Issue #1411, opened 2017-11-15
Allow overriding of delete behavior in `onCut`


justinweiss:
#### Do you want to request a *feature* or report a *bug*?

Request a feature

#### What's the current behavior?

Say I have a table with some selected table cells. When a user cuts the cells, I would like the cells to be cleared instead of deleted. In order to make this work, I need to override onCut, and reimplement all of the clipboard and browser compatibility pieces of onCutOrCopy.

#### What's the expected behavior?

In my plugin, I should be able to write a function that specifies the behavior onCut should use to delete the selected content. It should run somewhere around here : https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/after.js#L124
, and it should be able to defer to the existing behavior in the same way other plugins do.

We can default to the current behavior in `after.js.`

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

Issue #1410, opened 2017-11-15
Allow overriding `fragment` in onCutOrCopy


justinweiss:
#### Do you want to request a *feature* or report a *bug*?

Request a feature.

#### What's the current behavior?

If you want to change the fragment that gets copied in onCopy, you have to reimplement the entire function, including all of the encoding and browser compatibility quirks. 

For example, if I have a table, and I want to programmatically copy a non-continuous range of cells onto the clipboard, I have no way to do that.

#### What's the expected behavior?

My plugin should be able to define a method that onCutOrCopy will call to get the fragment here: https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/after.js#L164. 

`after.js` can hold the default behavior of returning `value.fragment`. This will allow a plugin author to customize just the part of copying that will most likely need to be changed.

ianstormtaylor:
Is there a way we might instead package up the `onCutOrCopy` logic into a helper similar to `getEventRange` or `getEventTransfer` that could be used. That way there wouldn't need to be interdependency between plugins, but just a util that gets used that contains the majority of the complex logic?

justinweiss:
Yep, that would more-or-less solve #1411, I think. It'd also solve my use case here, if there were a way to pass a fragment into `onCutOrCopy`.

tobiasandersen:
I've been wanting to do similar things as @justinweiss, and the idea proposed by @ianstormtaylor sounds perfect to me.

justinweiss:
@ianstormtaylor I took a stab at this in https://github.com/ianstormtaylor/slate/pull/1429.

ms88privat:
That's a step in the right direction. I'm upgrading slate to the new version since 0.20 and I see that the "beforePlugin` can also not be overwritten? Is this statement correct?

majelbstoat:
It has since become very hard to do this again. Possibly during 0.50 rearchitecture? The only approach appears once more to reimplement the full logic of setFragmentData from slate-react because it isn't exported.

I'm putting together a proposed PR to make this flexible again.

PR #1429, opened 2017-11-22
Add a `copyFragment` helper for plugin onCut/onCopy functions

justinweiss:
I took a stab at fixing #1410 and #1411 by creating an extra exposed helper function. Most of the new function was a direct move from `onCutOrCopy`, I just changed `fragment` to be a parameter.

### DIFF
--- a/packages/slate-react/src/index.js
+++ b/packages/slate-react/src/index.js
@@ -1,5 +1,6 @@
 
 import Editor from './components/editor'
+import copyFragment from './utils/copy-fragment'
 import findDOMNode from './utils/find-dom-node'
 import findDOMRange from './utils/find-dom-range'
 import findNode from './utils/find-node'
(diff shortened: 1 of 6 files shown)


ianstormtaylor:
Hey @justinweiss this looks great to me!

Would you be down to rename the helper to `cloneFragment`? Since it's not actually performing the copy logic, just to be slightly more correct. Thanks!

justinweiss:
Sure thing!

packages/slate-react/src/index.js
```diff
@@ -1,5 +1,6 @@
 
 import Editor from './components/editor'
+import cloneFragment from './utils/copy-fragment'
```
ianstormtaylor: This one got missed I think and is causing the tests to tail. Once those are passing I'm down to merge! Thanks @justinweiss!

ianstormtaylor:
Thanks @justinweiss!

PR #1432, opened 2017-11-28
Fix inserting text with hanging selection, issue #1189

czechdave:
Added 2 new tests to exploit the bug from #1189 :
changes/at-current-range/insert-text/hanging-selection-multiple-blocks.js
changes/at-current-range/insert-text/hanging-selection-single-block.js

removed test:
changes/at-current-range/delete/whole-block-with-other-block-type-below.js
which was just a duplicate of:
changes/at-current-range/delete/hanging-selection-single-block.js

implemented a fix for the 2 new failing test cases by updating `startKey` and `startOffset` after `deleteAtRange()` call.


ianstormtaylor:
Great call @czechdave, thank you!

PR #1434, opened 2017-11-28
Fix activeMarks at the beginning of a line

justinweiss:
If you move to the beginning of a text node, and the previous text
node has marks, `marks` is set but `activeMarks` is not. This was
caused by `getActiveMarksAtRangeAsArray` looking for a function that
didn't exist. With this change, `activeMarks` is brought in line with
how `marks` works for collapsed selections.


ianstormtaylor:
Thank you @justinweiss, good catch.

PR #1435, opened 2017-11-30
Fix small error in docs

conorcussell:
- make the placeholder have an opacity
- correct `contentEditable` jsx attribute


ianstormtaylor:
Thanks @conorcussell!

PR #1438, opened 2017-12-01
fix editor auto focus

raffy2010:
current implement of auto focus is buggy, HTMLElement.focus method
have no effect on selection state of editor, another re-render will
cause editor blur unexpectedly.


ianstormtaylor:
Thanks @raffy2010!

PR #1440, opened 2017-12-02
Document the `position` field of `merge_node`

justinweiss:
`merge_node` has a `position` field, which is exactly what I needed in order to transform some operations. It should be documented.


ianstormtaylor:
Thanks @justinweiss!

Issue #1418, opened 2017-11-18
Ctrl+Z to undo on empty editor crashes the editor


Slapbox:
Creating an empty editor with `html.deserialize('<p></p>')`, and then pressing `Ctrl+Z` crashes the editor.

Instead initialize with a raw value:
```
Value.fromJSON({
  document: {
    nodes: [
      {
        kind: 'block',
        type: 'paragraph',
        nodes: [
          {
            kind: 'text',
            leaves: [{ text: '' }]
          }
        ]
      }
    ]
  }
})
```

Slapbox:
Okay nevermind that. It was working for about 2 minutes it seemed, but I'm seeing the issue again. In any event, this is an edge case.

The error is: `Uncaught Error: Invalid `key` argument! It must be a key string, but you passed: null` in `slate/lib/models/node.js on line 2229`

Using Slate 0.31.

julienp:
This also happens when you paste text into an empty editor and then Ctrl-Z

In the fiddle https://jsfiddle.net/mh0an9ss/ paste some text, ctrl-z and it throws Error: Invalid `key` argument! It must be a key string, but you passed: null

anewusr:
The underlying problem has been an issue since https://github.com/ianstormtaylor/slate/issues/1161.

You can't do any transformations in the SlateEditor until it has been focused, for some reason.

conorcussell:
I may just be repeating things people already know, but here are my findings from digging into this today. 

When you create an empty editor the selection is unset, and an invalid selection is created because text is `undefined`.

https://github.com/ianstormtaylor/slate/blob/199c32a2b5f776c586e7202cfa3183e551153fe4/packages/slate/src/models/value.js#L117-L120

This also happens with copy and paste because the onPaste plugin invokes [`Plain.deserialize`](https://github.com/ianstormtaylor/slate/blob/199c32a2b5f776c586e7202cfa3183e551153fe4/packages/slate-react/src/plugins/after.js#L613) which in turn invokes [`Value.fromJSON`](https://github.com/ianstormtaylor/slate/blob/199c32a2b5f776c586e7202cfa3183e551153fe4/packages/slate-plain-serializer/src/index.js#L58)

[...]

Issue #1161, opened 2017-09-21
Error when calling `focus()` on an empty editor


rgrove:
#### Do you want to request a *feature* or report a *bug*?

bug

#### What's the current behavior?

Calling `focus()` on an empty editor after it's mounted causes an exception:

```
Uncaught Error: Invalid `key` argument! It must be either a block, an inline, a text, or a string. You passed: null
    at normalizeKey (https://npmcdn.com/slate@0.25.1/dist/slate.js:14384:9)
    at Document.getFurthestAncestor (https://npmcdn.com/slate@0.25.1/dist/slate.js:12810:13)
    at Document.object.(anonymous function) [as getFurthestAncestor] (https://npmcdn.com/slate@0.25.1/dist/slate.js:19296:28)
    at Document.getSelectionIndexes (https://npmcdn.com/slate@0.25.1/dist/slate.js:13493:26)
    at Content.render (https://npmcdn.com/slate-react@0.1.8/dist/slate-react.js:8409:30)
    [ ... ]
```

This was working as recently as 0.24.5, so the breakage must be pretty new.

https://jsfiddle.net/rgrove/sjcc7p0e/2/

![kapture 2017-09-21 at 14 47 08](https://user-images.githubusercontent.com/1465/30720409-31fac374-9edc-11e7-9173-1999608762e4.gif)

#### What's the expected behavior?

The editor should be focused.

ianstormtaylor:
Hey @rgrove, in this case there isn't anything to "focus" because there are no nodes in the editor's `document`. If you change the initial state to have at least one node this should go away.

ianstormtaylor:
Actually, maybe you're suggesting that it should not throw an error, which seems very reasonable to me 😄  if you want to PR!

rgrove:
> Actually, maybe you're suggesting that it should not throw an error, which seems very reasonable to me 😄 if you want to PR!

Thanks for the guidance!

I took a stab at this, but it looks like `focus()` was a red herring. It turns out even trying to type in an editor that's initially empty leads to errors at various levels since the initial range's `startKey` and `endKey` are always `null`. I'm not sure if that's expected or not, so I didn't dive in further. Seems easier just to ensure the editor always has at least one node.

skogsmaskin:
It could warn and suggest to write a schema ormalization rule.

anewusr:
I don't think this is a red herring. 

[...]

PR #1443, opened 2017-12-03
Deal with invalid range passed to getSelectionIndexes

conorcussell:


packages/slate/src/models/node.js
```diff
@@ -1405,6 +1405,11 @@ class Node {
       return null
     }
 
+    // if we've been given an invalid selection we can exit early.
+    if (!startKey && !endKey) {
```
ianstormtaylor: There's actually a `range.isUnset` property for checking this I believe, we should use that instead.

conorcussell:
Fixed that! I've deliberately left it as a separate statement so it's clear its not just a performance issue, but it could easily just be part of the statement above with a different comment. I'll try and test it later today to check it doesn't break anything else.

conorcussell:
This appears to fix both the cases from https://github.com/ianstormtaylor/slate/issues/1418

It makes the error from https://github.com/ianstormtaylor/slate/issues/1161 happen slightly later - it fails to focus, then when you click and start writing you get the error from assertKey, so its probably slightly worse in that case. I'll make a PR to make that error clearer anyway

ianstormtaylor:
Thanks @conorcussell!

PR #1444, opened 2017-12-04
Fix nodes validation with e.g. an optional first child

tpreusse:
When moving to the next definition with `continue` `nextChild` gets called twice and skips over a child before it's proven valid.

This rewinds the index and offset before `continue`. Ensuring the current child gets validated with the next definition instead of being skipped.


ianstormtaylor:
Good call @tpreusse, thank you!

PR #1446, opened 2017-12-05
fix(scroll-to-selection): use getClientRects when selectionRect.top/height is still 0 in Safari

### LINKED ISSUE #1445: Editor scrolls on every newline when inside a scrollable container in Safari

bug

when you press enter and there's no text following up (aka newline), the scroll jumps around in Safari.

related with #1416 #1428 

@erquhart will use your fiddle since it still serves it's purpose, hope you don't mind.
https://jsfiddle.net/fj9dvhom/6/

GIF

Sorry for the GIF quality.

I shall push a PR after this.

no jumping.

zGrav:
Tested on Safari 11.0.1 on OS X 10.13.1 with a scrollable container.

Fixes #1445 .

### DIFF
--- a/packages/slate-react/src/utils/scroll-to-selection.js
+++ b/packages/slate-react/src/utils/scroll-to-selection.js
@@ -1,6 +1,7 @@
 
 import getWindow from 'get-window'
 import isBackward from 'selection-is-backward'
+import { IS_SAFARI } from '../constants/environment'
 
 /**
  * CSS overflow values that would cause scrolling.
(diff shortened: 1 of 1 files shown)


zGrav:
Apologies for so many commits, I finally figured out how to do the eslint test with yarn, took me a bit longer than expected 😢

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -80,6 +80,10 @@ function scrollToSelection(selection) {
     }
 
     selectionRect = range.getBoundingClientRect()
+
+    if (selectionRect.top == 0 && selectionRect.height == 0) {
```
maddrag0n: `if (selectionRect.top === 0 && selectionRect.height === 0)`

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -80,6 +80,10 @@ function scrollToSelection(selection) {
...
+      selectionRect = range.getClientRects()[0]
```
maddrag0n:
```
if (range && range.getClientRects && range.getClientRects().length) {
    selectionRect = range.getClientRects()[0]
}
```

zGrav: merci <3

packages/slate-react/src/utils/scroll-to-selection.js
maddrag0n: as slate apparently uses improper `==` checks  throughout the code base... ignore this maybe?

lxcid:
In my own codebase I use this check though, is this alternative suggestion also solve the issue?

```js
    const selection = window.getSelection();
    let range = selection.getRangeAt(0);
    if (IS_SAFARI) {
      // `getBoundingClientRect` is buggy on Safari when there is no selected text active.
      // A workaround is available at https://stackoverflow.com/a/44261188
      range = range.cloneRange();
      range.setStart(range.startContainer, 0);
    }
```

lxcid:
`IS_SAFARI` is available in `src/contants/environment.js`

zGrav:
@lxcid , I believe I tried the cloneRange and it wasn’t working as intended, my implementation does work for empty text and keeps the original implementation in place when there is text.

I did not know about IS_SAFARI thanks!

lxcid:
I see! Good to know of alternative solution!

ianstormtaylor:
Thanks @zGrav!

PR #1450, opened 2017-12-05
fix(at-range): if no getDescendant, insert blocks

### LINKED ISSUE #1447: Select all + paste results in an error when there is a trailing empty block

Bug

If there is a trailing empty block, and we select all with cmd+a and then paste anything, an error occurs:

```
Uncaught TypeError: Cannot read property 'key' of undefined
    at Changes.insertFragmentAtRange (https://npmcdn.com/slate@0.31.5/dist/slate.js:7158:55)
    at Change.call (https://npmcdn.com/slate@0.31.5/dist/slate.js:10521:10)
    at Change.(anonymous function) [as insertFragmentAtRange] (https://npmcdn.com/slate@0.31.5/dist/slate.js:10580:15)
    at Changes.insertFragment (https://npmcdn.com/slate@0.31.5/dist/slate.js:6193:10)
    at Change.call (https://npmcdn.com/slate@0.31.5/dist/slate.js:10521:10)
    at Change.(anonymous function) [as insertFragment] (https://npmcdn.com/slate@0.31.5/dist/slate.js:10580:15)
    at Object.onPaste (https://npmcdn.com/slate-react@0.10.20/dist/slate-react.js:11554:14)
    at Stack.run (https://npmcdn.com/slate@0.31.5/dist/slate.js:16879:38)
    at https://npmcdn.com/slate-react@0.10.20/dist/slate-react.js:9290:20
    at Change.call (https://npmcdn.com/slate@0.31.5/dist/slate.js:10521:10)
```

Issue can be reproduced on the default fiddle: [https://jsfiddle.net/fj9dvhom/1/](https://jsfiddle.net/fj9dvhom/1/)
Here is a gif: [https://imgur.com/a/1KENt](https://imgur.com/a/1KENt)

Chrome version 62, Ubuntu 17, Slate version 0.31.5

After selecting all with cmd+a, paste should work correctly no matter if there is a trailing empty block or not.

zGrav:
Should fix #1447

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -734,6 +734,13 @@ Changes.insertFragmentAtRange = (change, range, fragment, options = {}) => {
   const { startKey, startOffset } = range
   const { value } = change
   let { document } = value
+
+  // if we can't find getDescendant, we assume it's an empty block.
+  if (document.getDescendant(startKey) === undefined) {
+    fragment.nodes.forEach(n => change.insertBlock(n))
+    return
+  }
+
   let startText = document.getDescendant(startKey)
   let startBlock = document.getClosestBlock(startText.key)
   let startChild = startBlock.getFurthestAncestor(startText.key)


ianstormtaylor:
@zGrav I don't think there should be a case where an `*AtRange` method is used where the range's key corresponds to a block that doesn't exist. It seems like whatever the issue is there's going to be some other solution, or the root cause isn't what we think it is.

zGrav:
Hello @ianstormtaylor , possibly.

I only did this fix because in my case it happens due to the range being expanded so, it goes right through the check, deletes the range and suddenly I'm thrown with errors about not being able to find Descendants...

Regards.

ianstormtaylor:
@zhujinxuan I think this one should use the same logic you added for other cases where an expanded collection is removed? is that right?

zhujinxuan:
I think https://github.com/ianstormtaylor/slate/issues/1447 is fixed in https://github.com/ianstormtaylor/slate/pull/1742.  I cannot reproduce #1447 in current demo.

ianstormtaylor:
Ah okay, sweet. So sounds like this has been fixed by other means. @zgrav could you confirm?

zGrav:
@ianstormtaylor seems to be, closing :D

PR #1451, opened 2017-12-06
Remove `value` from `operation.fromJSON`

### LINKED ISSUE #1439: Operations don't round-trip through JSON

Report a bug.

https://jsfiddle.net/gm13srfv/1/

Open the console and trigger a change (set_selection by clicking in the editor should be enough). You'll see the following error: 

> `Operation.fromJSON` was passed a “set_selection” operation without the required “value” attribute.

I would expect to get an Operation back that could be used like the original Operation.

It looks like the `value` attribute gets stripped in `toJSON`, but Slate tries to parse it back in in `fromJSON`.

I _think_ the value is only used for local undo / redo, it which case I think it should be stripped in `toJSON` and not expected in `fromJSON`. But I wanted to confirm before I send a PR. Either way, it should be consistent.

justinweiss:
Value is only used for local undo / redo, so it shouldn't be serialized along with the rest of the data. This fixes #1439.

### DIFF
--- a/packages/slate/src/models/operation.js
+++ b/packages/slate/src/models/operation.js
@@ -125,6 +125,10 @@ class Operation extends Record(DEFAULTS) {
         v = Range.create(v)
       }
 
+      if (key == 'value') {
+        v = Value.create(v)
+      }
+
       if (key == 'properties' && type == 'set_mark') {
         v = Mark.createProperties(v)
       }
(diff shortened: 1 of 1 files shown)


packages/slate/src/models/operation.js
```diff
@@ -124,10 +125,6 @@ class Operation extends Record(DEFAULTS) {
         v = Range.create(v)
       }
 
-      if (key == 'value') {
-        v = Value.create(v)
-      }
-
```
ianstormtaylor: Is removing this part required? It looks like the other skipped properties are still present in this section.

justinweiss: Ah, I read that as dead code. But I guess it would still run if you specifically passed something for the `value` key.

ianstormtaylor:
Thanks @justinweiss!

PR #1453, opened 2017-12-06
fix(at-range): if cant assertDescendant, assume empty block

### LINKED ISSUE #1452: Select all + enter results in an error when there is a trailing empty block

bug

Refer to #1447 

Issue can be reproduced on the default fiddle: https://jsfiddle.net/fj9dvhom/1/

should be able to type.

zGrav:
Should fix #1452


ianstormtaylor:
Hey @zGrav thanks for this, but Slate core should never actually be inserting blocks of a specific `type` on its own, that's up to the schema. It sounds like there's probably a different solution this. Happy to discuss more!

PR #1459, opened 2017-12-07
fix(at-range): forward deletion now stays with next block.

### LINKED ISSUE #1457: Delete key behaves more like backspace when selected block is empty
Bug

When the selection is on an empty block, the delete key behavior is more like backspace.

This causes an odd behavior where the cursor moves to the previous sibling when delete key is pressed instead of staying at the same block. Additionally, delete does not promote the content next sibling block to the current selected block as expected.

http://recordit.co/cBMAOQVAr8

Expected behavior is that delete key always maintains current block selection.

zGrav:
Should fix #1457 .

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -564,7 +564,7 @@ Changes.deleteForwardAtRange = (change, range, n = 1, options = {}) => {
         focusKey: nextBlock.key,
         focusOffset: 0,
       })
-      change.moveToRangeOf(nextBlock)
+      change.select(range)
     }
     return
   }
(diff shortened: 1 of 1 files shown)


packages/slate/src/changes/at-range.js
```diff
@@ -557,7 +557,12 @@ Changes.deleteForwardAtRange = (change, range, n = 1, options = {}) => {
 
   // If the closest is not void, but empty, remove it
   if (block && !block.isVoid && block.isEmpty && document.nodes.size !== 1) {
+    const nextBlock = document.getNextBlock(block.key)
     change.removeNodeByKey(block.key, { normalize })
+    if (nextBlock && nextBlock.key) {
+      change.moveToRangeOf(nextBlock)
+      change.moveOffsetsTo(0, 0)
```
ianstormtaylor: I think this can be summarized as `moveToStartOf(nextBlock)`

zGrav: Yup. Pushed, thanks!

CameronAckermanSEL:
@zGrav thanks so much :)

PR #1460, opened 2017-12-08
Add a missing test case for wrapping an inline inside an inline

czechdave:
Just stumbled into this unexpected behavior with `wrapInline()` method inside a single inline parent.
The current behavior splits the parent inline into 3 inlines and wraps the middle into the one supplied to the method. Is that really the expected result?

@ianstormtaylor please let me know if this is indeed intentional, otherwise I can put some effort into changing the behavior to pass this test.

### DIFF
--- a/packages/slate/test/changes/at-current-range/wrap-inline/inside-inline.js
+++ b/packages/slate/test/changes/at-current-range/wrap-inline/inside-inline.js
@@ -1,0 +1,27 @@
+/** @jsx h */
+
+import h from '../../../helpers/h'
+
+export default function (change) {
+  change.wrapInline('hashtag')
+}
+
+export const input = (
+  <value>
+    <document>
+      <paragraph>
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
@czechdave yeah the current behavior is the expected one. If you want to open an issue where it can be discussed, to have an alternative, I'd be open to that. The thing is, "wrap" generally means to go outside whatever is currently there, so it's strange to have "wrap" act inside the inline. But I could be convinced if someone writes up a well-argued issue. Thanks!

czechdave:
OK, I've opened an issue https://github.com/ianstormtaylor/slate/issues/1463
This test case seems super obvious use case to me, is there a different way to achieve this that I'm missing?

ianstormtaylor:
Closing in favor of https://github.com/ianstormtaylor/slate/pull/1876, thank you @czechdave!

PR #1462, opened 2017-12-08
fix: account for container padding in scrollToSelection

### LINKED ISSUE #1461: scroll-to-selection: doesn't scroll all the way down
Bug

It doesn't scroll all the way down:
Before

It should scroll all the way down:
After

The problem seems to be here:

https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/scroll-to-selection.js#L120

```
const selectionFocusTop = backward ? selectionRect.top : selectionRect.bottom
```

Should  be the other away:

```
const selectionFocusTop = backward ? selectionRect.bottom : selectionRect.top
```

EDIT: [updated proposal to fix](https://github.com/ianstormtaylor/slate/issues/1461#issuecomment-350358506)

echenley:
Fixes #1461.

8f64f3bff0e009587b8e8976352996d0

ed136e9474c737e07877830e0216a784

Also updated to use only the cursor position (rather than the entire selection) to calculate scroll position. Consider this selection:

screen shot 2017-12-08 at 3 24 34 pm

If this is a forward selection, there is no real way to find the top of the cursor, so the calculation is currently off when you're trying to scroll up to the cursor position. Same issue if it's backwards and scrolling down.

Since we only care about cursor position, you can just collapse the range and using that rect instead.

### DIFF
--- a/packages/slate-react/src/utils/scroll-to-selection.js
+++ b/packages/slate-react/src/utils/scroll-to-selection.js
@@ -65,8 +65,9 @@ function scrollToSelection(selection) {
   const isWindow = scroller == window.document.body || scroller == window.document.documentElement
   const backward = isBackward(selection)
 
-  const range = selection.getRangeAt(0)
-  let selectionRect = range.getBoundingClientRect()
+  const range = selection.getRangeAt(0).cloneRange()
+  range.collapse(backward)
+  let cursorRect = range.getBoundingClientRect()
 
   // COMPAT: range.getBoundingClientRect() returns 0s in Safari when range is
   // collapsed. Expanding the range by 1 is a relatively effective workaround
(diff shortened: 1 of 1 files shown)


echenley:
@ianstormtaylor This technically does two things: account for padding, and fix some positioning logic. The first fix does seem to diverge from the normal browser behavior (at least on OS X). But it may be worth it if that's what users expect.

Example: in a native textarea, padding is ignored when scrolling to the cursor location:

cb31a8ff20cfc2e787b3ce630f024234

The second fix is actually necessary. If we want to get that in separately, let me know and I can put up another PR.

skogsmaskin:
Great work @echenley!

We are discussing what is the expected behaviour when going to position 0 in the top and position [text.length] in the end. Should it scroll all the way up / all the way down?



vs



Text editors does this a little differently. Word does it like the first image, and LibreOffice does it like the second image.

I think the second one it preferable and it clearly shows that "nothing more here to see, you are at the end". And you don't have to do that little scroll yourself to see the actual end / beginning of the document.

Anyway, this is way much better than it was. And 👍 for merging this in.

isubastiCadmus:
do we need to incorporate your changes on#1446  @zGrav or will this fix #1445 as well?

zGrav:
@isubastiCadmus my changes is in a different area, so they will still be needed.

ianstormtaylor:
Thanks @echenley, this sounds good. Can you fix the conflict and then I'm good to merge?

echenley:
@ianstormtaylor Should be good to go.

ianstormtaylor:
Thanks @echenley!

Issue #853, opened 2017-05-29
Issue related to native `onBlur` event handling


AlbertHilb:
Let me start introducing a notation. By _edit island_ I mean a DOM subtree with a structure similar to

```
<div contentEditable={false}>
  <div>Not Editable Content</div>
  <EditableElement />
</div>
```

where `EditableElement` is an element which contains an editable text. It can be a native element, like `input`or `textarea`, a javascript based editor different from Slate or a collection of Slate nodes.

The code responsible to handle browser `onBlur` and `onFocus` events has been rewritten a lot of times.
Every time, one issue is fixed and two or more are introduced. :wink:

In the past, I made two changes to `onBlur` and `onFocus` (PR #693 and #692) to fix keyboard navigation issues in presence of edit islands with editable elements consisting of Slate block collections.
After commit 619870808699ed853ca153419e616b764c639762 an issue (#749) concerning embedded `input` has been opened, the handler code has been modified to fix that bug, but this has raised again the problems I managed to solve with my PRs.

Here, I'll try to show what is the root of these issues.

There are three main different (from the Slate point of view) cases in which browsers fire `blur` events.

1. The focus switches from within the Slate `Editor` to an element not contained into the editor DOM subtree.

2. The Slate editor DOM subtree contains an edit island and the cursor goes from an element within the editor subtree but outside the edit island
to an element inside the edit island (or vice versa).

3. The `document` `visibilityState` changes to `hidden`.

The case gives us headaches is the second.

The task of the handlers is to decide if the focus state of the Slate editor must change or not in response to `blur` and `focus` events fired by the browser.

All the versions of the `onBlur` event handling code take this decision based on the relative inclusion relations among the element losing focus, the one that gains focus and the editor container.
This is a reasonable choice but it can't always give the right answer: there are cases in which the relations are the same but the editor must have different behaviors.

For example, let's consider two Slate editor instances containing custom blocks with edit islands. Let the island editable elements be:
- an `input` element in the first case
- a collection of Slate blocks in the second.

When the caret entering the edit island triggers the `onBlur` event, the relative inclusion relations among the elements losing and gaining the focus and the editor container are the same but in the first case the editor must be blurred because it is no longer responsible of the editing process while in the second case it has to retain its `focus` state since it is still Slate that must take care of the editing.

[...]

PR #1465, opened 2017-12-12
Move contentEditable attr to content element

delijah:
# Software versions
OS: macOS Sierra v10.12.3
Browser: Google Chrome 62.0.3202.94
slate: 0.31.5
slate-react: 0.10.20

# Problem
The newest slate version renders void nodes differently than older versions (in particular where the contenteditable attribute is applied has been changed). This causes the browser to fire blur and focus events, each time you move the caret from a void node to text or the other way around. This did not happen before. We believe the new behaviour is wrong. The user is not really leaving the editor, he is only moving the caret trough the editor (why should a blur event been fired?).

A short example how void nodes were rendered before and how they are rendered now.

Before:
```
<div data-slate-void="true" ...>
    <span ...>
        <!-- Zero width space -->
    </span>

    <div contenteditable="false">
        <!-- My custom void node contents --->
    </div>
</div>
```

Now:
```
<div data-slate-void="true" contenteditable="false" ...>
    <div data-slate-spacer="true" contenteditable="true" ...>
        <!-- Zero width space -->
    </div>

    <div draggable="true">
        <!-- My custom void node contents --->
    </div>
</div>
```

Since in the new version a `contenteditable="true"` is inside a `contenteditable="false"`, the browser kind of thinks the zero width space is a "new" editor and fires a blur event for the outer editor (the actual slate editor) and a focus event for the inner editor (zero width space) when moving the caret by using the keyboards arrow keys.

# Actual behaviour
When moving the caret through the editor and passing by void nodes, blur and focus events are fired.

# Desired behaviour
Blur and focus events should not be fired when moving the caret through the editor since we are not really leaving nor entering the editor. The focus stays within the editor during the whole process.

We have created a small test project to easily see the different behaviours. Just switch slate-react between the above mentioned version and this PR by using `npm link`:
https://github.com/DND-IT/slate-void-node-blur-showcase

### DIFF
--- a/packages/slate-react/src/components/void.js
+++ b/packages/slate-react/src/components/void.js
@@ -69,7 +69,6 @@ class Void extends React.Component {
 
     const spacer = (
       <Tag
-        contentEditable
         data-slate-spacer
         suppressContentEditableWarning
         style={style}
@@ -79,7 +78,7 @@ class Void extends React.Component {
     )
 
     const content = (
-      <Tag draggable={readOnly ? null : true}>
+      <Tag draggable={readOnly ? null : true} contentEditable={readOnly ? null : false}>
         {children}
       </Tag>
     )
@@ -90,7 +89,6 @@ class Void extends React.Component {
       <Tag
         data-slate-void
         data-key={node.key}
-        contentEditable={readOnly ? null : false}
       >
         {readOnly ? null : spacer}
         {content}


ianstormtaylor:
@delijah this was done for reasons, please look at the older PRs and see what those reasons were, and then make sure that your pull request resolves them. I'm happy to get this fixed so that the blurring doesn't happen, but I don't want to do it at the expense of work that has already happened.

cc @AlbertHilb

ms88privat:
I think this is the one you are talking about @ianstormtaylor https://github.com/ianstormtaylor/slate/pull/1265

delijah:
Looks like it is also related to #853.

I had a look at those issues you've mentioned and ended up having a look at how other editors are solving problems like that (draft-js, ckeditor, ory etc.). A lot of them decided to not use contenteditable because of issues like that.

Interesting readings:
https://medium.com/content-uneditable/contenteditable-the-good-the-bad-and-the-ugly-261a38555e9c
https://medium.com/content-uneditable/fixing-contenteditable-1a9a5073c35d

Unfortunately i do not have the time to dig any deeper into this issue since it seems to be huge.

I will open up an issue and hope this will be considered in future fixes/changes.

delijah:
Another question which came up in our team: Is it the right behaviour to select a void node when moving the caret by using the arrow keys? This is not really consequent, since with text the editor is behaving differently. If you move the caret through text, a selection of text never happens (you would have to press shift AND an arrow key to make a selection happen).

Current behaviour:
```
TEXT
EMBED
T|EXT
```

Pressing arrow left

```
TEXT
EMBED
|TEXT
```

Pressing arrow left

```
TEXT
|EMBED| // Embed is selected
TEXT
```

Pressing arrow left

```
TEXT|
EMBED
TEXT
```

Probably better behaviour?

```
TEXT
EMBED
T|EXT
```

Pressing arrow left

```
TEXT
EMBED
|TEXT
```

Pressing arrow left

```
TEXT|
EMBED
TEXT
```

To be more consequent, a selection of the embed might only occur, when the user presses the shift key together with arrow keys.

skogsmaskin:
I agree with you on (last post regarding void node selections)  that this should be the expected behavior. There is another problem with how it currently works. You never know if you are before or after the void node. This is a problem when you want to press enter and make a new paragraph below or above the void node. When several void nodes are in a row it becomes a real problem. I had to make e a plugin to deal with this. I think this deserves an separate Github issue.

ianstormtaylor:
Feel free to open separate issues to discuss this. (When doing so, please research them well and take into account that the current behavior probably wasn't decided randomly, so come up with good solutions to solve any issues that the changes will present.)

Issue #1470, opened 2017-12-14
Regression with text replacement


blakeembrey:
#### What's the current behavior?

Introduced with https://github.com/ianstormtaylor/slate/pull/1432. Here's a GIF of the introduced issue (text gets jumbled up):

![](https://cl.ly/0H232W2t1k0F/Screen%20Recording%202017-12-13%20at%2011.11%20pm.gif)

#### What's the expected behavior?

The text should be replaced in place as expected.

PR #1473, opened 2017-12-14
Fix selection replacement and marks

blakeembrey:
Fixes https://github.com/ianstormtaylor/slate/issues/1470. Also encountered and fixed an issue with marks being incorrectly applied (basically anything that was in the target text becomes an active mark). Behaviour is based on native document editing behaviour of marks (e.g. Word, Pages).

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1009,6 +1009,18 @@ class Node {
    * Get a set of the marks in a `range`.
    *
    * @param {Range} range
+   * @return {Set<Mark>}
+   */
+
+  getInsertMarksAtRange(range) {
+    const array = this.getInsertMarksAtRangeAsArray(range)
+    return new Set(array)
+  }
+
+  /**
+   * Get a set of the marks in a `range`.
+   *
+   * @param {Range} range
    * @return {OrderedSet<Mark>}
    */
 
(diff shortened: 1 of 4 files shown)


ianstormtaylor:
Hey @blakeembrey thank you for this!

I think the change to first vs. all characters for marks need to be fixed in a different way though. As far as I can tell, this fix would change the `activeMarks` behavior for all cases of active marks, which is going to break the logic where people want it to be based on all of the characters. It sounds like this insertion handling needs a special case, instead of changing the default case.

I might be misunderstanding though.

blakeembrey:
@ianstormtaylor Maybe I'm misunderstanding (did the PR a bit ago now), but the intention was certainly to change the mark behaviour to match how rich text editors actually work (which is insertion position, not all marks - e.g. Word, Pages). I believe `activeMarks` vs `marks` is unaffected though, it's a different method and is used for the toggle behaviour right?

packages/slate/src/models/node.js
```diff
@@ -1053,17 +1053,12 @@ class Node {
...
 
-    // Otherwise, get a set of the marks for each character in the range.
-    return this
-      .getCharactersAtRange(range)
-      .reduce((memo, char) => {
-        char.marks.toArray().forEach(c => memo.push(c))
-        return memo
-      }, [])
+    // Otherwise, return the marks of the first character in the range.
+    return this.getCharactersAtRange(range).first().marks.toArray()
```
ianstormtaylor: @blakeembrey this change here sounds to me like it changes how the `getMarksAtRange` method works, so it will affect any place that is using it, [including `value.marks`](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/models/value.js#L521-L525) (which is often used for toolbar active states, etc.)

Am I misunderstanding that? It seems like this is fixing the "expanded inserting" case by changing the behavior for all cases?

blakeembrey: So should I add a new method for this behaviour? This seems like the ideal behaviour as it matched other rich text editors.

ianstormtaylor: I think we might not need a new method. Instead it sounds like the `insertTextAtRange` should be changed to not use this method without first changing the range to only include one character? Or something similar. If that ends up requiring a method to control complexity it might be warranted. I think the fix is more related to the text insertion behavior than the mark reading behavior.

blakeembrey: IIRC, I tried this, but the `.marks` getter uses this method and that's re-used everywhere - should I create a new getter and changed the `.marks` getter behaviour?

blakeembrey: Now I recall more, it's an issue with pasting also because it uses `document.getMarksAtRange(selection.collapseToStart())` which results in the wrong marks. When you paste with a selection, the selection is being collapsed which results in using the `isCollapsed` logic to select the previous character. In Pages and Word, however, the behaviour is to mirror the selection of the first character - this would need to be a new method just for handling the complexity and I'd need to move pasting, character insertion and other methods over to it. Happy to see how that looks now though, updating the PR ASAP.

blakeembrey:
@ianstormtaylor Tidied it up and re-using a method for getting the insertion marks.

ianstormtaylor:
Thanks @blakeembrey!

PR #1475, opened 2017-12-14
Fix iOS autocorrect issues with Backspace and Return

### LINKED ISSUE #1471: *bug* iOS backspace and enter keystrokes not registering correctly with autocorrect.
**What is the current behavior?**
1) When pressing backspace, the iOS autocorrect bar does not register the change. For example if you type "The", and then delete all the characters to "", the iOS autocorrect bar still thinks "The" is the current state. If you then type "none", it thinks you've typed "Thenone".
2) When pressing enter, the iOS autcorrect bar does not register the change. For example, if you type "none<enter>pere" it thinks you typed "nonepere" (one word, no space).

slate_ios_autocorrect

**What is the expected behavior?**
1) When pressing backspace, the iOS autocorrect bar should update appropriately.
2) When pressing enter, the iOS autocorrect bar should update appropriately.

**Which versions of slate and which browser / OS are affected by this issue?** 
Tested on iOS safari with:
"react": "16.2.0",
"react-dom": "16.2.0",
"slate": "0.31.7",
"slate-react": "0.10.22"

Note that draft.js has a very similar issue (cf. https://github.com/facebook/draft-js/issues/1574)

rgrove:
Handling Backspace and Return keystrokes in `onNativeBeforeInput` ensures that iOS recognizes word and paragraph breaks and generates appropriate autocorrect suggestions.

Fixes #1471 

**Caveat:** I don't currently have the ability to test on an Android device. Slate already doesn't exactly work well on Android and I don't _think_ this will make things any worse there, but I'd really appreciate it if someone with an Android device would give this a spin.


davidlibland:
This PR resolves a number of the issues on iOS, but backspace and enter still seem to cause problems when working with `slate-edit-list`; specifically when creating a new list item (enter) or deleting that list item (I've attached a video). I'll try to investigate it a little more and see what I find.

edit-list-ios-auto

vshia:
I'm working on getting slate usable on Android (https://github.com/humandx/slate/tree/fix/android). I've tested this branch and it doesn't look like it affects Android.

rgrove:
@davidlibland I haven't looked at slate-edit-list, but it sounds like it may be preventing the default behavior of `keydown` events, which would prevent the `beforeinput` event from firing and would bypass the iOS fix.

@vshia That's great to hear! Thanks for testing.

packages/slate-react/src/constants/hotkeys.js
```diff
@@ -22,19 +22,19 @@ const ITALIC = isKeyHotkey('mod+i')
...
+const SPLIT_BLOCK = e => !IS_IOS && (ENTER(e) || SHIFT_ENTER(e))
 
 const BACKSPACE = isKeyHotkey('backspace')
 const SHIFT_BACKSPACE = isKeyHotkey('shift+backspace')
 const DELETE = isKeyHotkey('delete')
 const SHIFT_DELETE = isKeyHotkey('shift+delete')
-const DELETE_BACKWARD = e => BACKSPACE(e) || SHIFT_BACKSPACE(e)
-const DELETE_FORWARD = e => DELETE(e) || SHIFT_DELETE(e)
+const DELETE_BACKWARD = e => !IS_IOS && (BACKSPACE(e) || SHIFT_BACKSPACE(e))
+const DELETE_FORWARD = e => !IS_IOS && (DELETE(e) || SHIFT_DELETE(e))
```
ianstormtaylor: What's the need for all these `!IS_IOS` additions here?

ianstormtaylor: Gotcha! That makes sense. If you're down to clean this one up, I'd like that. I think currently #1491 adds a few other unrelated style things that I'd rather not have combined with this bug fix.

Slapbox:
@vshia can you let me know how Slate with React-Native is going for you? I'm embarking on a similar project now.

ianstormtaylor:
Ah I totally missed that you cleaned it up already @rgrove, thank you!

PR #1477, opened 2017-12-15
Properly invert `merge_node` operations

### LINKED ISSUE #1469: The `move_node` operation doesn't invert properly

Report a bug.

Fiddle is here: https://jsfiddle.net/fj9dvhom/21/

Wait one second, and you'll see an error in the console: `Error: Could not find a descendant at path "2,1".`

There should be no visible change, since we performed an operation and immediately ran the inverse.

This happens because moving a node changes the structure of the document, but `move_node` ignores that structure change when it inverts the operation. For example, the inverse of

`{type: "move_node", path: [1], newPath: [2, 1]}`

should be

`{type: "move_node", path: [1, 1], newPath: [1]}`

I can try to fix this.

justinweiss:
When inverting a merge operation, we need to take into account the structure change caused by the merge operation. I can come up with two places where this happens:

* If you're moving from a left sibling of an ancestor of the new position, we need to adjust part of the new position by -1.
* If you're moving to an ancestor of the original position, or a left sibling of an ancestor of the original position, we need to adjust part of the path by 1.

I added tests for each of these, and have some other examples that could make good low-level test cases. Is there a good place to put those? I didn't see any tests on operations.

This fixes #1469.

### DIFF
--- a/packages/slate/src/operations/invert.js
+++ b/packages/slate/src/operations/invert.js
@@ -48,7 +48,34 @@ function invertOperation(op) {
 
   if (type == 'move_node') {
     const { newPath, path } = op
-    const inverse = op.set('path', newPath).set('newPath', path)
+    let inversePath = newPath
+    let inverseNewPath = path
+
+    const pathLast = path.length - 1
+    const newPathLast = newPath.length - 1
+
+    // If the node's old position was a left sibling of an ancestor of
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Great pull request @justinweiss, thank you very much.

There are no low-level tests on the operations themselves for now, although I'd definitely be open to having some added if you wanted to take a stab!

PR #1478, opened 2017-12-15
Using the wrong property to access length of `change.operations`

lxcid:
As operations is now a `Immutable.List`, `length` will always be `undefined`, thus changes through `<Editor />` never get flushed.

I am quite reluctant to fix this bug but ensuring the correctness of the code first should be the highest priority.

I'll file a separate issue, but the problem I have with fixing this is that it will reintroduce an extra `onChange` event scheduled with `setTimeout()` when `<Editor />` first get created and mounted. (It is trying to queue/flush the changes.)

This actually hurt a muilti-editors environment.

### DIFF
--- a/packages/slate-react/src/components/editor.js
+++ b/packages/slate-react/src/components/editor.js
@@ -178,7 +178,7 @@ class Editor extends React.Component {
    */
 
   queueChange = (change) => {
-    if (change.operations.length) {
+    if (change.operations.size) {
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Ah shoot, thank you @lxcid. I'll respond about the race condition in the other thread, that's going to be interesting to solve. Thanks!

Issue #1479, opened 2017-12-15
Race condition caused by Editor queuing/flushing changes


lxcid:
#### Do you want to request a *feature* or report a *bug*?

bug

#### What's the current behavior?

I have contributed to the discussion around here about how the behaviour of Editor queuing/flushing changes is detrimental for a multi-editor environment, performance wise (mainly due to the need to use `setTimeout()`). See #1401 & https://github.com/ianstormtaylor/slate/issues/1318#issuecomment-341956697.

But recently, I realise that this behaviour also introduce race condition if you want to use it in a collaborative multi-editor environment.

I'll try to walkthrough what's going on and will also accompany by [an example](https://jsfiddle.net/lxcid/q91uydtz/) that will require #1478 to be merged to be able to reproduce.

First of all when you instantiate an `<Editor />`, it will queue a change (the change have operation) immediately.

https://github.com/ianstormtaylor/slate/blob/22576e521fc956563996e91de23551ca03cce312/packages/slate-react/src/components/editor.js#L90-L95

That is because `value.schema` and `editor.schema` are different. During `Editor.constructor()`, `value.schema` likely contains only a default `Schema` while `editor.schema` is assigned a newly created `Schema` with plugins resolved from the `Editor.props`.

https://github.com/ianstormtaylor/slate/blob/22576e521fc956563996e91de23551ca03cce312/packages/slate-react/src/plugins/before.js#L113-L126

Thus, this explain that `Editor` will always generate and queue a change on `Editor.constructor()`.

The issue lie in that the `Editor.constructor()` decide to queue the change. So during `Editor.componentDidMount()`, when `Editor` finally decide to flush the change, instead of processing it immediately, it delay it slightly with `setTimeout()`.

https://github.com/ianstormtaylor/slate/blob/0afe15fa7694e53bad1d9843a15486a56d0ad91e/packages/slate-react/src/components/editor.js#L181-L192

`setTimeout()` is unpredictable and if multiple `setTimeout()` is called (happen during multi-editor environment), you are likely see significant delay and cause performance issue.

At this point, I would like to note a few points:

- this is a regression, old version of slate was capable of large number of editors.
- this behaviour break the predictability of React model

Now I move on to the part where I explain why this also introduce a _race condition_, which I believe is the more serious of problem here.

Most of us will store the `Value`s somewhere and might manipulate it outside the component. For my use case, remote changes can arrive in real time.

If the remote changes come in the following order:

1. `Create an Editor at Location A`
[...]

PR #1480, opened 2017-12-15
[WIP] Flush change immediately (Proposal fix for #1479)

lxcid:
Propose to flush change immediately instead of delaying it using `setTimeout()`.

- Flushing change is only use for `props.onChange()` callback and invoked at the end of render lifecycle (didMount/didUpdate), can't find the rationale for delaying it.
- `setTimeout()` delay are not dependable for a predictive behaviour
- `onChange()` is an important callback for a controlled component, delaying it can cause race condition. e.g. any changes outside of the component lifecycle that happen between the delay of flushing change will be overridden

@ianstormtaylor I like to check with you on this proposal to see what you think of it

Ref #1479

### DIFF
--- a/packages/slate-react/src/components/editor.js
+++ b/packages/slate-react/src/components/editor.js
@@ -158,17 +158,6 @@ class Editor extends React.Component {
   }
 
   /**
-   * When the component unmounts, clear flushTimeout if it has been set
-   * to avoid calling onChange after unmount.
-   */
-
-  componentWillUnmount = () => {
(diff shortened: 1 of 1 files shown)


lxcid:
I also notice that `Editor.queueChange()` and `Editor.change()` are unaware of each other.

But I have not notice scenario where this is an issue.

ianstormtaylor:
@lxcid very good point, I'm not sure why I thought that a `setTimeout` was needed there. Thank you! I think this should work well for all cases.

PR #1483, opened 2017-12-19
Fix schema marks validation

julienp:
<s>`NODE_MARK_INVALID` should only happen if none of the node's mark are valid.</s>

### DIFF
--- a/packages/slate/src/models/schema.js
+++ b/packages/slate/src/models/schema.js
@@ -287,10 +287,8 @@ class Schema extends Record(DEFAULTS) {
       const marks = node.getMarks().toArray()
 
       for (const mark of marks) {
-        for (const def of rule.marks) {
-          if (def.type != mark.type) {
-            return this.fail(NODE_MARK_INVALID, { ...ctx, mark })
-          }
+        if (!rule.marks.some(def => def.type === mark.type)) {
+          return this.fail(NODE_MARK_INVALID, { ...ctx, mark })
         }
       }
     }
--- a/packages/slate/test/schema/custom/node-mark-invalid-default.js
+++ b/packages/slate/test/schema/custom/node-mark-invalid-default.js
@@ -5,7 +5,7 @@ import h from '../../helpers/h'
 export const schema = {
   blocks: {
     paragraph: {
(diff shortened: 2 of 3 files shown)


CameronAckermanSEL:
So if I apply an invalid mark to a node that has several valid marks on it, there's no way to normalize that?

packages/slate/test/schema/custom/node-mark-invalid-default.js
```diff
@@ -5,7 +5,7 @@ import h from '../../helpers/h'
 export const schema = {
   blocks: {
     paragraph: {
-      marks: ['bold'],
+      marks: [{ type: 'bold' }, { type: 'underline' }],
```
CameronAckermanSEL: could we avoid breaking the interface here by continuing to pass a string?

packages/slate/src/models/schema.js
```diff
@@ -287,10 +287,8 @@ class Schema extends Record(DEFAULTS) {
       const marks = node.getMarks().toArray()
 
       for (const mark of marks) {
-        for (const def of rule.marks) {
-          if (def.type != mark.type) {
-            return this.fail(NODE_MARK_INVALID, { ...ctx, mark })
-          }
+        if (!rule.marks.some(def => def.type === mark.type)) {
```
CameronAckermanSEL: we can avoid breaking the interface by just checking against `mark` instead of `mark.type`

julienp:
> So if I apply an invalid mark to a node that has several valid marks on it, there's no way to normalize that?
No, my comment above is wrong, the code should be correct though. We check every mark of the node and verify that it is present in the schema's marks.

packages/slate/test/schema/custom/node-mark-invalid-default.js
julienp: It seemed to me that the tests were using the wrong interface, as the existing code was comparing to `mark.type`

CameronAckermanSEL: Oh yeah you're right! haha

ianstormtaylor:
Thanks @julienp!

PR #1488, opened 2017-12-23
fix for #1487

### LINKED ISSUE #1487: Bug: Undoing an inline wrap change across multiple blocks throws an error
JSFiddle: https://jsfiddle.net/fj9dvhom/26/
If you call `Change.wrapInlineAtRange` with a range that extends into multiple blocks, undoing that change will throw the error `Tried to merge two nodes of different kinds: "text" and "inline".` from `Block.mergeNode`.

screen shot 2017-12-21 at 11 40 09 pm

The error is not thrown

brainkim:
Fixes #1487 

The problem was that `startIndex - 1` usually equalled -1, which worked with `Immutable.List.insert` to insert the new inline at the end of the startNode, but this messed up inverting the operation so that inserted inline node remained.

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -1351,7 +1351,7 @@ Changes.wrapInlineAtRange = (change, range, inline, options = {}) => {
     const startNode = inline.regenerateKey()
     const endNode = inline.regenerateKey()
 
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Wow, good catch @brainkim, thank you! Funny how that ended up working.

PR #1489, opened 2017-12-23
`this.element` might be `null` when the callback is executed

lxcid:
We need to check that `this.element` is not `null` before calling `focus()`.


ianstormtaylor:
Thanks @lxcid!

Issue #1481, opened 2017-12-15
Selection issue in Safari


kib357:
#### Do you want to request a *feature* or report a *bug*?

bug

#### What's the current behavior?

When trying to implement one-line editor, i'm getting this error in Safari:
https://jsfiddle.net/tesvg2qw/
Please try to put the cursor in the end of input and then press backspace.
Error only throws when editor (or it's descendent) has `white-space: nowrap` style applied.
![image](https://user-images.githubusercontent.com/347668/34057847-8222cb60-e1fa-11e7-9eef-1e7f6cbfa6c0.png)

#### What's the expected behavior?

I know it's looks like a Safari bug, but maybe we can found some workaround for this behaviour.

zGrav:
After a bit of testing, it seems that it loses anchorNode/focusNode due to the wrapping... it's one of those quirks with Safari I guess...

I did find a workaround, but not sure if... well, it's the right approach.

`      if (_environment.IS_SAFARI) {
        if (native.anchorNode.textContent.endsWith(' ')) {
            native.anchorNode.textContent = native.anchorNode.textContent.replace(' ', '\u00a0')
        }
      }`

If @ianstormtaylor doesn't see a problem with it, I can build a PR for it.

ianstormtaylor:
Hey @zGrav thank you for investigating.

For that PR, I think editing the underlying text itself should be our last resort, since it's going to cause issues with copy-pasting, where we end up with artifacts.

Is there a way to fix this in the selection-handling logic instead? or other behavioral logic?

zGrav:
@ianstormtaylor if there is, I haven't discovered it as of now, it just feels like a very very quirky issue with Safari and how it handles wrapping the whitespace itself since the issue is not present in Chrome (have not tested Firefox).

ianstormtaylor:
Yeah true. Is there an open browser issue for this? 

I'm worried about just patching it without having some knowledge of why it's happening, but I'd be open to the PR. I think it would need to be reworked though to be more foolproof. It should probably be at render time, instead of editing the node at selection update time. And it should be done in a way that doesn't affect the copy-paste content (using a wrapping span that can be adjusted I think, like we do with zero-width spaces).

zGrav:
@ianstormtaylor picking this one back up as well.

I am not quite sure there is a browser issue for it, at least I cannot find one.

kinhbang89:
Is there any update or fixes in this Safari issue?

[...]

PR #1490, opened 2017-12-24
fix(slate-react/components/content): do not use range.extend on Safari Desktop/iOS

zGrav:
…Content ends in whitespace, replace it with unicode no-break space

Fix for #1481

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -12,6 +12,7 @@ import findDOMRange from '../utils/find-dom-range'
 import findRange from '../utils/find-range'
 import scrollToSelection from '../utils/scroll-to-selection'
 import {
+  IS_SAFARI,
   IS_FIREFOX,
   IS_IOS,
   IS_ANDROID,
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/components/content.js
```diff
@@ -206,7 +207,7 @@ class Content extends React.Component {
     native.removeAllRanges()
 
     // COMPAT: IE 11 does not support Selection.extend
-    if (native.extend) {
+    if ((native.extend && !IS_IOS) && (native.extend && !IS_SAFARI)) {
```
skogsmaskin: Maybe ``if (native.extend && !(IS_SAFARI || IS_IOS))`` for more clarity?

packages/slate-react/src/components/content.js
```diff
@@ -206,7 +207,7 @@ class Content extends React.Component {
     native.removeAllRanges()
 
     // COMPAT: IE 11 does not support Selection.extend
```
skogsmaskin: Comment should be updated.

packages/slate-react/src/components/content.js
zGrav: you are indeed correct! ;)

packages/slate-react/src/components/content.js
zGrav: Done!

skogsmaskin: Great, but also keep the IE-comment about not supporting range.extend please.

skogsmaskin:
@zGrav Can you please utdate the title and description of this PR to make a little more sense?
Also I have a feeling it is solving something the wrong way. I will have a closer look tomorrow when I have the Safari stack around.

kinhbang89:
ANy update on this PR? I got the same error in Safari

ianstormtaylor:
Can anyone verify this? Looking at the support tables it says that `selection.extend` is supported in Safari? Is it something specific we're doing wrong?

ianstormtaylor:

PR #1491, opened 2017-12-26
Ios autocorrect buffer

### LINKED ISSUE #1471: *bug* iOS backspace and enter keystrokes not registering correctly with autocorrect.

davidlibland:
This builds of the work of @rgrove to make sure that the iOS autocomplete buffer is correctly updated on Backspace and Return keystrokes.

Fixes #1471

We handle backspace and return keystrokes as part of the `KeyDown` events rather than in `onNativeBeforeInput`. This makes slate more extensible - plugins generally override slate's default behavior for those keystrokes from within `KeyDown` events. Placing the functionality in `onNativeBeforeInput` means those plugins might need to override it as well.
In particular, the changes in this PR make `slate-edit-list` compatible with the iOS autocomplete buffer (once a small update has been made to that plugin); i.e. it helps resolve the issue noted [here](https://github.com/ianstormtaylor/slate/pull/1475).

I would love to get @rgrove 's comments on this PR.


rgrove:
@davidlibland It looks like this incorporates my changes from PR #1475, differing only in how the `keydown` `preventDefault()` call is avoided. Seems okay to me, but I'm on vacation at the moment and haven't tested it.

Given that there are extraneous commits in this PR that aren't really relevant to Slate, it seems like the cleanest approach might be to add this change to my PR, no? Or if @ianstormtaylor doesn't mind the extra commits, he could just merge this one and close mine. In any event, only one PR or the other can be merged, since they conflict.

rgrove:
This PR has been superseded by #1475 and #1520, so it can be closed.

ianstormtaylor:
Thanks @rgrove!

ianstormtaylor:
And thanks @davidlibland for your work on this!

PR #1495, opened 2017-12-28
Don't attempt to remove event listeners from non-existent window

cmmartin:
```
<IFrameComponent>
  <SlateEditor {...props} />
</IFrameComponent>
```

Since react unmounts from top down, `IFrameComponent` will unmount first (destroying its window), then `SlateEditor` will unmount and attempt to access the iframe window (which no longer exists) throwing `Uncaught TypeError: Cannot read property 'document' of undefined`.

It should be safe to skip removing event listeners from windows that no longer exist since they will be garbage collected upon destruction of the window anyway

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -117,7 +117,9 @@ class Content extends React.Component {
   componentWillUnmount() {
     const window = getWindow(this.element)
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Ah, didn't realize that, thanks @cmmartin!

PR #1503, opened 2018-01-04
slate-html-serializer: consistent element.value lookups

jtadmor:
`el.value` vs. `el.value || el.nodeValue` were inconsistently used when deserializing HTML.

This caused a specific bug in which html looked like:

```
<ul>
<li>Some text</li>
</ul>
```

The HTML parser was finding `'#text'` nodes with no `value` but with a `nodeValue`, which was causing them to not be filtered out, resulting in improper nodes (text nodes as direct children of the `<ul>`).
  
To reproduce, see this fiddle: https://jsfiddle.net/usf40krv/

Go to line 320 and change `el.value` -> `getValue(el)`.

### DIFF
--- a/packages/slate-html-serializer/src/index.js
+++ b/packages/slate-html-serializer/src/index.js
@@ -37,13 +37,14 @@ const TEXT_RULE = {
     }
 
     if (el.nodeName == '#text') {
-      if (el.value && el.value.match(/<!--.*?-->/)) return
+      const value = getValue(el)
+      if (value && value.match(/<!--.*?-->/)) return
 
       return {
         kind: 'text',
         leaves: [{
           kind: 'leaf',
-          text: el.value || el.nodeValue
+          text: value
         }]
       }
     }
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @jtadmor thank you for this!

I actually can't find any reference to `element.value` in the DOM API reference, I only see `element.nodeValue`. I'm not sure if that was something that we missed when converting from Cheerio or what, but I think we'd be good to only use `nodeValue` everywhere from what I can tell.

Would you be down to make that change? Thanks!

jtadmor:
Absolutely (I was curious about that myself and didn't know if it related to something the HTML parsing was doing in other env or some such)

ianstormtaylor:
Beautiful, thanks @jtadmor!

ianstormtaylor:
(Also, although this is the right move, I think this will be a breaking change for people using `parse5`—which although technically should never have worked, was recommended before we knew what we were doing, so I'm going to include this in a breaking release right now, instead of a patch.)

PR #1506, opened 2018-01-05
Adding normalize false to the moveNodeByKey call in deleteAtRange

YurkaninRyan:
Currently the `moveNodeByKey` call in `change.deleteAtRange` normalizes the output.  This is problematic because this allows normalizers to modify the change object, potentially removing the endBlock node.

If that happens, we don't update our endKey, and that causes the `mergeNodeByKey` call afterwards to fail.  I think it's better if we just don't normalize any of the operations, and then normalize all at once at the end.

@ianstormtaylor This sounds like pretty common behavior to not want to normalize on a string of changes.  I wonder if this warrants some kind of flag?

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -236,7 +236,7 @@ Changes.deleteAtRange = (change, range, options = {}) => {
 
       // Move the end block to be right after the start block.
       if (endParentIndex != startParentIndex + 1) {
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Good call @YurkaninRyan, thanks!

You're right about this being a common behavior. I'm not sure what API would make it nicer, but I'm open to ideas if you think of something. Definitely would be cool if there's a way to pave over the need to care about normalization as much.

PR #1514, opened 2018-01-09
Move selection to the offset of remove_text if it's in its bounds

justinweiss:
If selection is inside a block of text that's removed, the part of the selection that's inside the bounds should clamp to the offset of the removed text. 

For example, if your cursor is at offset 5, and you're deleting from 2-10, your cursor should end up at offset 2.

Here's a fiddle showing the broken behavior: https://jsfiddle.net/w1dxg25n/2/ Wait 5 seconds, and you should see an error in the console.

### DIFF
--- a/packages/slate/src/operations/apply.js
+++ b/packages/slate/src/operations/apply.js
@@ -287,12 +287,20 @@ const APPLIERS = {
     const { anchorKey, focusKey, anchorOffset, focusOffset } = selection
     let node = document.assertPath(path)
 
-    if (anchorKey == node.key && anchorOffset >= rangeOffset) {
-      selection = selection.moveAnchor(-length)
+    if (anchorKey == node.key) {
+      if (anchorOffset >= rangeOffset) {
+        selection = selection.moveAnchor(-length)
+      } else if (anchorOffset > offset) {
+        selection = selection.moveAnchorTo(anchorKey, offset)
+      }
     }
 
-    if (focusKey == node.key && focusOffset >= rangeOffset) {
-      selection = selection.moveFocus(-length)
+    if (focusKey == node.key) {
+      if (focusOffset >= rangeOffset) {
+        selection = selection.moveFocus(-length)
+      } else if (focusOffset > offset) {
+        selection = selection.moveFocusTo(focusKey, offset)
+      }
     }
 
     node = node.removeText(offset, length)
--- a/packages/slate/test/index.js
+++ b/packages/slate/test/index.js
@@ -20,6 +20,7 @@ describe('slate', () => {
   require('./schema')
   require('./changes')
   require('./history')
+  require('./operations')
 })
 
 /**
--- a/packages/slate/test/operations/apply/remove_text/cursor-inside-removed-text.js
+++ b/packages/slate/test/operations/apply/remove_text/cursor-inside-removed-text.js
@@ -2,15 +2,13 @@
 
 import h from '../../../helpers/h'
 
-export default function (change) {
-  return change.applyOperation({
-    type: 'remove_text',
-    path: [0, 0],
-    offset: 2,
-    text: 'is is some text inside ',
-    marks: []
-  })
-}
+export default [{
+  type: 'remove_text',
+  path: [0, 0],
+  offset: 2,
+  text: 'is is some text inside ',
+  marks: []
+}]
 
 export const input = (
   <value>
--- a/packages/slate/test/operations/apply/remove_text/cursor-inside-removed-text.js
+++ b/packages/slate/test/operations/apply/remove_text/cursor-inside-removed-text.js
@@ -1,0 +1,33 @@
+/** @jsx h */
+
+import h from '../../../helpers/h'
+
+export default function (change) {
+  return change.applyOperation({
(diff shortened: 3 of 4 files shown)


skogsmaskin:
Nice catch! This looks good to me!

skogsmaskin:
I think I was a bit eager saying it was looking good to me, sorry. I have now looked a bit closer into this.

First, please fix the fiddle - for others to inspect the the bug as well.

I am wondering about the tests. I see there are no tests for any operation before. What is up with this @ianstormtaylor - are the tests in the wrong place or are there simply no tests added for operations yet?

justinweiss:
@skogsmaskin You're totally right, the fiddle is wrong. I simplified the case here: https://jsfiddle.net/w1dxg25n/5/

If you put your cursor at before the 't' in text, your cursor will move to the wrong place when the text is deleted. If you put it before the 'a', you'll see an error: `Unable to find a native DOM range from the current selection.` in the console when the text is deleted.

skogsmaskin:
Thank you @justinweiss now we can all clearly see the bug.

I think this looks good for me if the tests are confirmed to be in the right place.

skogsmaskin:
@justinweiss Just checked out your PR, and it seems like the tests doesn't actually test your new code.

justinweiss:
Thanks for checking, I'll double-check -- I think I wrote the fix against v0.31.7 (before I did the upgrade locally), and the specific error I was seeing might be recovered from now. I'm still seeing some selection transformation weirdness on v0.32, so I'll see if I can make the test reflect that.

justinweiss:
When I comment out the `else if` blocks with `yarn watch` running, I get this when I run the test:

```slate operations apply removeText cursor-inside-removed-text:

      AssertionError: { object: 'value',
  document: { object: 'document', data: {}, nodes: [ [Object] ] },
  data: {},
  selection:
   { object: 'ra deepEqual { object: 'value',
  document: { object: 'document', data: {}, nodes: [ [Object] ] },
  data: {},
  selection:
   { object: 'ra
      + expected - actual

           "object": "document"
         }
         "object": "value"
         "selection": {
      -    "anchorOffset": 5
      +    "anchorOffset": 2
           "anchorPath": [
             0
             0
           ]
      -    "focusOffset": 5
      +    "focusOffset": 2
           "focusPath": [
             0
             0
           ]

      at _callee$ (packages/slate/test/operations/index.js:35:22)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:64:40)
      at GeneratorFunctionPrototype.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:299:22)
      at GeneratorFunctionPrototype.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:116:21)
      at step (packages/slate/test/operations/index.js:19:191)
      at packages/slate/test/operations/index.js:19:437
      at Context.<anonymous> (packages/slate/test/operations/index.js:19:99)
```

What are you seeing?

skogsmaskin:
Sorry @justinweiss - guess who forgot to build before running with the removed code... It's looking great!

packages/slate/src/operations/apply.js
```diff
@@ -287,12 +287,20 @@ const APPLIERS = {
     const { anchorKey, focusKey, anchorOffset, focusOffset } = selection
     let node = document.assertPath(path)
 
-    if (anchorKey == node.key && anchorOffset >= rangeOffset) {
-      selection = selection.moveAnchor(-length)
+    if (anchorKey == node.key) {
```
skogsmaskin: There now two of these if's, should be moved into one if:

```
    if (anchorKey == node.key) {
      if (anchorOffset >= rangeOffset) {
        selection = selection.moveAnchor(-length)
      } else if (anchorOffset > offset) {
        selection = selection.moveAnchorTo(anchorKey, offset)
      }
      if (focusOffset >= rangeOffset) {
        selection = selection.moveFocus(-length)
      } else if (focusOffset > offset) {
        selection = selection.moveFocusTo(focusKey, offset)
      }
    }
```

justinweiss:
Thanks for checking it out!

packages/slate/src/operations/apply.js
justinweiss: Wouldn't that change the behavior? In that case, we'd only check anchorKey, not focusKey.

skogsmaskin: No, just move them into one parent if. There are now two of the same if.

skogsmaskin: Think I need glasses. Read them as the same if. Please ignore :)

skogsmaskin:
@justinweiss (deleted the last review issue, I thought it was the same if but it was of course not, I think and need glasses or shouldn't review code at 03 AM :) )

This is perfect now in my opinion. Would like to hear a statement about the status of operation tests from @ianstormtaylor before we merge though.

ianstormtaylor:
Good catch @justinweiss! 

@skogsmaskin you're correct, there are currently no tests for the lower-level operations. I'd love for some to be added though, if anyone wanted to get started on just adding a few sanity checks. Although I think we can merge this for now until that happens.

Either one of you, feel free to merge!

Issue #1513, opened 2018-01-09
Screen Jumps/Scrolls cursor out of view during editing on iOS (iPad, iPhone)


thesunny:
#### Do you want to request a *feature* or report a *bug*?

Report a Bug

#### What's the current behavior?

Using the rich text editor on the SlateJS demo site (or any of the demos for that matter):

http://slatejs.org/#/rich-text

If you start editing and the cursor goes below the virtual keyboard, the text doesn't scroll up properly. Instead, it jumps all the way to the top of the screen.

![image-2](https://user-images.githubusercontent.com/469587/34737118-0d39abbe-f52a-11e7-9024-df606951ab0b.GIF)

Note that the jump happened as soon as I typed anything. In this case, I hit the SPACE on the virtual keyboard.

This is on an iPad iOS 11 but the same happens on my iPad. The latest version of Slate as on the demo as of today.

From what I can remember, this did not happen before; however, it may have been quite an old version where it worked properly.

#### What's the expected behavior?

Normally, the screen will stay scrolled to the part of the editor we are editing.

zGrav:
I cannot seem to replicate this on my iPhone nor Simulator.

Will investigate.

zGrav:
@ianstormtaylor are you using a whiteSpace styler in the demos?

thesunny:
@zGrav @ianstormtaylor Additional details:

- The jump happens no matter what character is typed. It can be a letter, space or carriage return for example.

- I can confirm that it does not appear in the simulator (initially tried to replicate it there and failed)

- I've been trying to diagnose, fiddling with scroll positions and such, but what seems to completely break this is that `window.scrollTo(x, y)` line ALWAYS scrolls to the top no matter what the value of y is. So, for example, I hard-coded 100, -100, and various other values and they all scroll to the top.

Somehow it seems unlikely that `window.scrollTo` is broken in the newest version of iOS but if I remove that line, the bug does goes away. Of course it doesn't scroll the cursor into position though. I can only theorize that the `scrollTo` is triggering a scroll event and the function attached to the scroll event is not working properly. That's just a guess at this point though.

I feel like a decent temporary fix is simply doing:

```js
function scrollToSelection(selection) {
  if (IS_IOS) return
// ...
```

With the matching `import` statement for `IS_IOS` at the top of course.

[...]

PR #1515, opened 2018-01-10
Fixed scroll-to-selection jumping to top on iOS

thesunny:
This code simply disables the scroll to selection on all iOS devices.

The reason why is because on the iPad, any typing in the editor causes a scroll back to the top of the page. On other iOS devices, it appears the scroll to selection might not do anything anyways.

It's a bit brute force but it makes the editor usable if not necessarily convenient. Right now, it is unusable on the iPad.

The full discussion is here:

https://github.com/ianstormtaylor/slate/issues/1513

### DIFF
--- a/packages/slate-react/src/utils/scroll-to-selection.js
+++ b/packages/slate-react/src/utils/scroll-to-selection.js
@@ -1,7 +1,7 @@
 
 import getWindow from 'get-window'
 import isBackward from 'selection-is-backward'
-import { IS_SAFARI } from '../constants/environment'
+import { IS_SAFARI, IS_IOS } from '../constants/environment'
 
 /**
  * CSS overflow values that would cause scrolling.
@@ -58,6 +58,7 @@ function findScrollContainer(el, window) {
  */
 
 function scrollToSelection(selection) {
+  if (IS_IOS) return
   if (!selection.anchorNode) return
 
   const window = getWindow(selection.anchorNode)
--- a/packages/slate-react/src/utils/scroll-to-selection.js
+++ b/packages/slate-react/src/utils/scroll-to-selection.js
@@ -16,6 +16,12 @@ const OVERFLOWS = [
 ]
 
 /**
+ * Detect whether we are running IOS version 11
+ */
+
+const IS_IOS_11 = IS_IOS && !!window.navigator.userAgent.match(/os 11_/i)
+
+/**
  * Find the nearest parent with scrolling, or window.
  *
  * @param {el} Element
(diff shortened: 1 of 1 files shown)


skogsmaskin:
As it is confirmed working fine on several iOS devices, I think this should be investigated some more. At least debugged via USB-developer console and see if there are any errors thrown on your iPad.

thesunny:
Limited the PR so that it aborts scroll-to-selection only on iOS 11.

@ianstormtaylor I wasn't quite sure where the IS_IOS_11 check should occur and how generic it should be. In order to keep the code small and to keep the check close to where it is used I declared an IS_IOS_11 constant directly in the `scroll-to-selection` file.

I considered (and wrote) alternate versions to to derive major, minor and patch versions. I also considered putting it in `constants/environment` but I did it this way for a few reasons:

- By making the test specific to 11, the code was extremely succint. Just one line in fact. The other versions all used functions and were several lines long and could possibly introduce parsing bugs (in fact, the iOS version detectors I saw online had potential bugs in them I could see)
- By putting the detection code in `scroll-to-selection` it indicates that this check is really only used in this one file. This means that if the OS test needs to change (e.g. if the bug is solves in a later subversion or it continues into version 12), we know we can modify or obliterate the constant whereas if it was in the environment, it may end up as dead code. No longer used, but it's not easy to see that it is no longer referenced.

Let me know if you'd like me to change it though.

packages/slate-react/src/utils/scroll-to-selection.js
```diff
@@ -15,6 +15,12 @@ const OVERFLOWS = [
   'scroll',
 ]
 
+/**
+ * Detect whether we are running IOS version 11
+ */
+
+const IS_IOS_11 = IS_IOS && !!window.navigator.userAgent.match(/os 11_/i)
```
ianstormtaylor: This will throw in server-side rendering I think. Maybe we can just put the `11` check inside the `scrollToSelection` function instead?

thesunny: Mentioned this in our Slack conversation. It shouldn't throw an error because of the `IS_IOS` check before the window userAgent check. `IS_IOS` will only return true if we are in a browser. That check is done in the `constants/environment.js` file.

Because the check was in IS_IOS already, I took out the check here.

ianstormtaylor: Ah gotcha, that makes sense. It's slightly confusing maybe, but I think it sounds good. Please ignore me!

thesunny: Yeah, I agree that it is confusing. If this was my own code, I usually put comments right next to it but I noticed that SlateJS code is pretty sparse with comments so I tried to follow that commenting style.

juangl:
Hey, @ianstormtaylor, is there an estimated date to have this published? Thank!

Issue #1113, opened 2017-09-11
Marks and Selection no longer works correctly in some cases


matmat94:
I didn't include a JSFiddle neither a GIF because you can reproduce the bug on slatejs.org.
I'm on Debian Stretch using Chrome Version 59.0.3071.115 (Official Build) (64-bit).

I'd like to know if you have the same issue as me: I took the rich-text example to work on it, but there seems to be a bug in Slate since I have the same problems on the examples page (http://slatejs.org/#/rich-text and http://slatejs.org/#/iframes).
If you click at the end of the text or at the beginning, not selecting any text, and enable bold then disable it without typing any text, Selection becomes buggy (however it's not the case if I type some text before disabling bold).
Then whenever I move the cursor to a portion of text that was bold, the button does not become active anymore and if I select this portion of text and try to remove the bold mark, it does not work anymore. After reproducing this bug, other mark buttons such as italic do not become active when they should.
Doing my own toggleMark() function with some() to do some debugging, there seems to be no more bold mark on the selected portion of text (which is bold of course), so it tries to add it, indefinitely.

Thank you

danburzo:
@matmat94 Hmm, sounds like it *may* have something to do with the changes recently merged, https://github.com/ianstormtaylor/slate/pull/990 

Unfortunately I cannot reproduce (on Chrome 61/MacOS) the issue by following (to the best of my knowledge) the steps you provided. Can you please create a short video that demonstrates the steps?

matmat94:
No problem, very nice plugin btw.
Here's a gif as asked.
![slate](https://user-images.githubusercontent.com/6900842/30332793-8928b92e-97db-11e7-9da5-4b3c43e9eab9.gif)

TimYi:
@matmat94 select a collapsed range, and then make a toggleMark change, selection will be wrong.
By the way, the selection mechanism is the combination of onSelect in content model and Changes.select in on-selection model, the code is not very clear for me to understand.

TimYi:
These codes may have something wrong, in at-current-range, line 229.
When the selection is a collapsed range, it will be called.
```
const marks = document.getActiveMarksAtRange(selection).remove(mark)
const sel = selection.set('marks', marks)
change.select(sel)
```

bengotow:
[...]

PR #1517, opened 2018-01-10
Abort select() if only change is “forgetting” current marks (#1113)

bengotow:
This is a small fix for the issue discussed in #1113 that restores the behavior of `toggleMark`, et. al. on collapsed selections for me. I'm totally open to other solutions if there's a better approach!

### DIFF
--- a/packages/slate/src/changes/on-selection.js
+++ b/packages/slate/src/changes/on-selection.js
@@ -56,6 +56,13 @@ Changes.select = (change, properties, options = {}) => {
     return
   }
 
+  // If the only change is blowing away the current selection's marks, abort.
+  // To remove all marks, clear the set, don't set it to null.
+  const onlyMarks = Object.keys(props).length == 1 && props.hasOwnProperty('marks')
+  if (onlyMarks && props.marks == null) {
+    return
+  }
+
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
@bengotow I'm worried about changing the transform logic to essentially ignore a direct request. But two potential ideas for you, and maybe you can tell me which would work... 

- Change [this line](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/content.js#L312) to compare only the "positional" characteristics of a selection when deciding if it was a "duplicate", and ignore the `marks`? 

- Or, add [something here](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/after.js#L571) to change the `onSelect` logic to preserve the `marks` if everything else about the selection matched the existing one exactly?

That way we are catching this case more specifically, instead of restricting the change methods abilities for what might be a valid use case.

rebirthtobi:
is this going to be ever addressed, slate is not usable as its not useful for rich text at the moment cause of a bug https://github.com/ianstormtaylor/slate/issues/2736

ianstormtaylor:

PR #1520, opened 2018-01-11
Fix double-handling of iOS return and backspace keypresses

### LINKED ISSUE #1519: I think master was recently broken with enter creating a double-enter on iOS 11 (possibly other versions as well)

Bug

on an iPad if I hit enter on a line I get a double enter

Should insert a single paragraph, not two.

I was trying to submit my PR and couldn't for the life of me figure out why my code kept inserting a double ENTER. Turns out, after cloning the repo from scratch, that this bug is on master.

Anybody that accepted a PR recently, please check on iOS 11 iPad. Probably affects other devices as well.

rgrove:
Fixes #1519, which was introduced by 5f991645.

I'm honestly not sure how I missed this when I tested 5f991645. It's possible I was testing stale code without realizing it. Sorry about that!

This fix is (I think) a better alternative to my original approach of modifying the hotkey constants. Instead of preventing the hotkeys from being matched on iOS, we just avoid handling them when they shouldn't be handled. Turns out this is easy because it only needs to happen in one place, and doing it there also makes the intent more clear.

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -6,6 +6,7 @@ import React from 'react'
 import getWindow from 'get-window'
 import { Block, Inline, Text } from 'slate'
 
+import { IS_IOS } from '../constants/environment'
 import EVENT_HANDLERS from '../constants/event-handlers'
 import HOTKEYS from '../constants/hotkeys'
 import Content from '../components/content'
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Beautiful, good call @rgrove that is much easier to understand. Thank you!

davidlibland:
@rgrove Sorry, I was on vacation so I missed a lot of the activity surrounding these PRs. Thanks for all the work to fix these issues.
One comment: I think that backspace and enter should be handled in `after.js`'s `KeyDown` handler rather than in `content.js`'s `beforeInput`. The reason for this is that other plugins (such as slate-edit-list) expect it to happen there; and they introduce novel functionality by overriding the `KeyDown` events. What are your thoughts?

rgrove:
> One comment: I think that backspace and enter should be handled in after.js's KeyDown handler rather than in content.js's beforeInput. The reason for this is that other plugins (such as slate-edit-list) expect it to happen there; and they introduce novel functionality by overriding the KeyDown events. What are your thoughts?

@davidlibland `keydown` events don't provide the information necessary to preserve iOS's native autocorrect functionality. That information is only provided by native `beforeinput` events.

If you want to perform custom handling of <kbd>Backspace</kbd> and <kbd>Return</kbd>, calling `e.preventDefault()` in your plugin's `keydown` handler will prevent the `beforeinput` event from happening and allow you to do whatever you want, but that means you won't get native iOS autocorrect behavior. If you want native autocorrect behavior, you'll have to allow the `beforeinput` event to happen.

One solution might be to add `onNativeBeforeInput` to the plugin lifecycle, but then plugins would still need to know whether to handle events in `onKeyDown` or `onNativeBeforeInput`, which means each plugin would need to be aware of which handler Slate itself is using to update the editor for any given keypress.

PR #1524, opened 2018-01-11
Expose core plugins

czechdave:
In some scenarios we would like to use the core plugins inside tests with `slate-simulator`. That is currently not possible. This PR exposes the core plugins so we can import them on case by case basis.


ianstormtaylor:
Beautiful PR @czechdave, thank you!

Issue #1454, opened 2017-12-06
Bug in search-highlight example: `setValue({decoration})` will disable redo


zhujinxuan:
**Do you want to request a feature or report a bug?**

Report a bug.

**What's the current behavior?**

In the search-highlight example, http://slatejs.org/#/search-highlighting , in each time the search-box `onChange` is triggered, the `redo` will be cleaned `setValue({decoration})`.  Then `redo` will be disabled by the search-box change.

** What's the expected **

Just wondering if it is possible to re-rendering `decorationNode` or `forceUpdate` from the outside?

SmilinBrian:
#1458 shows more clearly why a different way to trigger re-building the `decoration`s is useful, but this issue more clearly requests the functionality.

@zhujinxuan would you consider putting your animated GIF and explanation here and close 1458?

zhujinxuan:
@SmilinBrian   Sorry I forgot to close the request after know setOperationFlag...  I will simply send a PR to add `setOperationFlag` to the example and close this commit.  

Thank you~

PR #1526, opened 2018-01-11
Update search-highlighting example per #1454

bengotow:
(Per #1454)

Hey! I saw this change was intended by @zhujinxuan but he hadn't had time to add it yet. I copied the search highlighting example today and ran into the same issue after a few hours so I think it should definitely be changed.


ianstormtaylor:
Thanks @bengotow!

PR #1529, opened 2018-01-13
Fix broken links on docs

Kornil:
@ianstormtaylor plugins + typo in core-plugins link

Scope creep:
- `slate-mentions` and `slate-suggestions` referenced in the root `README.md` are no longer maintained/broken cause of the last months breaking changes, I did not remove the links cause at the moment I'm not sure I can offer an alternative solution (I have built a working one with latest specs but it is stored on a private repository for my company, I need to check if it can be open sourced).


ianstormtaylor:
Thank you @Kornil!

I'm actually okay if we remove no longer-maintained ones, as long as it was a clear decision that it was no longer maintained. If it's just that the author is lagging, I think people can figure that out for themselves maybe.

Issue #1415, opened 2017-11-16
Export schema violation reason constants


CameronAckermanSEL:
#### Do you want to request a *feature* or report a *bug*?
Feature

#### What's the current behavior?

The schema violation reason constants are private to the slate package

#### What's the requested behavior?

The schema violation reason constants are exported for consumption by consumers of the API so that they can be used in normalization handlers.

Could we export something like

```
export const SCHEMA_VIOLATIONS = Object.freeze({
  'CHILD_KIND_INVALID': CHILD_KIND_INVALID,
  ....
});

Then consumers can use this way:

`case SCHEMA_VIOLATIONS.CHILD_KIND_INVALID: `

CameronAckermanSEL:
closed in #1532

PR #1532, opened 2018-01-14
added export constant enum for schema violations

DamareYoh:


DamareYoh:
For #1415

DamareYoh:
Was a little unsure about good naming conventions :)

schneidmaster:
@DamareYoh Thanks! I think this is a good improvement. Would you mind also updating the [documentation](https://github.com/ianstormtaylor/slate/blob/master/docs/reference/slate/schema.md#normalize) and the other usages e.g. the [tests](https://github.com/ianstormtaylor/slate/tree/master/packages/slate/test/schema/custom)?

DamareYoh:
@schneidmaster thanks very much for the feedback and direction! Let me know if there's anything else I can improve

schneidmaster:
LGTM 👍  thanks for the contribution!

ianstormtaylor:
Does anyone know of prior art on this type of enum export? I'm wondering whether other libraries (especially in the React ecosystem) export these. And whether they use `PascalCase` or `CONSTANT_CASE` for these type of exports too. Thanks!

DamareYoh:
Hey @ianstormtaylor, i found this stack overflow question on the topic. https://stackoverflow.com/questions/13233137/javascript-enum-pattern-naming-convention 

But totally up for changing the naming convention!

ianstormtaylor:
I was just thinking about this some more. I think as written right now, these aren't going to guard against typos actually be the export is the object instead of the individual reasons for invalid states?

In other words, you can still write:

```js
import { SchemaViolations } from 'slate'

if (reason == SchemaViolations.ChildTypeInvaliddd) {

}
```

And you won't get any feedback as to whether that is correct or not? Wouldn't that end up making this not that useful?

cc @DamareYoh @schneidmaster @CameronAckermanSEL @Soreine

CameronAckermanSEL:
This is a good point. I think that's kind of status-quo for what we have at the moment  (because the same thing will happen if someone had `reason === 'child_type_Invalid' for instance)  but there's definitely room for improvement.

ianstormtaylor:
@CameronAckermanSEL yeah but if it's status quo then aren't we gaining nothing by just converting it into a dictionary exported instead of using strings?

CameronAckermanSEL:
I think so. So we just want a way to get this to assert or throw it the user has attempted to retrieve a property of the SchemaViolations object that doesnt exist?

schneidmaster:
A lot of editors will help you out with the object key, so I think this is decidedly better than having ad-hoc strings throughout. But I agree there's more that can be done here. We could instead export each constant to give a build error if you typo. Such as (rough):

```js
// src/schema-violations.js
export const ChildTypeInvalid = 'child_type_invalid';
export const NodeTextInvalid = 'node_text_invalid';

// userland file
// would error at build if you misnamed the import
import { ChildTypeInvalid } from 'slate/schema-violations';
```

The downside to that approach is then you have things like `ChildTypeInvalid` in your code where it's not especially clear that it's a constant, so that would push me in the direction of ALL_CAPS_CONSTANTS. (I don't believe `import * as SchemaViolations from 'slate/schema-violations';` will error at build time if you try to use an invalid `SchemaViolations.SomeTypo`.)

ianstormtaylor:
@schneidmaster yeah I think the individual exports is the way to go. I'll see about making a commit that changes to that on master!

CameronAckermanSEL:
Hey @ianstormtaylor, I went poking around and found this that might help: https://www.nczonline.net/blog/2014/04/22/creating-defensive-objects-with-es6-proxies/

Sorry for leaving this a bit incomplete and creating more work for you :\

CameronAckermanSEL:
I got something like this to work:

```
export const SchemaViolations = new Proxy(schemaViolations, {
  get: (target, property) => {
    if (property in target) {
      return target[property]
    } else {
      throw new ReferenceError(`Property '${property}' does not exist.`)
    }
  }
})
```

Based on this article: https://www.nczonline.net/blog/2014/04/22/creating-defensive-objects-with-es6-proxies/

It builds and test pass, but there's a worry that it doesnt work with babel https://stackoverflow.com/questions/34960094/when-i-use-babel-to-compile-js-proxy-is-not-defined

CameronAckermanSEL:
Hey @ianstormtaylor and @schneidmaster , even if we use named exports in this way and get the import wrong, it still doesn't assert like we want it to. It'll just return undefined if we try to import something that isn't exported.

ianstormtaylor:
I implemented a slightly different version of this PR, but as a `slate-schema-violations` package that you can import the named constants from directly. That way if you're using eslint there are ways to get it to complain, and it should fail regardless when bundled/run.

CameronAckermanSEL:
Thanks! Sorry about that, hope it wasn't too much trouble for you

PR #1533, opened 2018-01-15
Append copied fragment to editor DOM node instead of page body

### LINKED ISSUE #1484: CSS from body is being copied when copying text from slate, even if slate div has css that should override this
As reported in slack, when the body has any CSS, this is used when copying text out of slate. See gif of issue. I reproduced it by:

going to http://slatejs.org/#/rich-text
Opening inspector, adding red background to body element of the page
Copying everything in the editor
Pasting into a outlook new email

Actual behavior:
Copied with red background which I put on body

Expected behavior: 
Copied with css only from slate editor (white background).

Conversation here: https://slate-js.slack.com/archives/C1RH7AXSS/p1513711945000264, Jinxuan Zhu seemed to acknowledge that this is a slate bug here: https://slate-js.slack.com/archives/C1RH7AXSS/p1513717102000279

There is a gif of this in slack, I can't seem to upload it here. Feel free to ask questions on slate slack, `@steven.white`

jennaplusplus:
Fixes #1484, which caused pasted material to retain styles (such as background color) from the page body.

Instead of appending the cloned fragment to the `<body>`, append it to the editor element instead.

Before:
before

After:
fixed

### DIFF
--- a/packages/slate-react/src/utils/clone-fragment.js
+++ b/packages/slate-react/src/utils/clone-fragment.js
@@ -92,7 +92,7 @@ function cloneFragment(event, value, fragment = value.fragment) {
   attach.setAttribute('data-slate-fragment', encoded)
 
   // Add the phony content to the DOM, and select it, so it will be copied.
-  const editor = window.document.querySelector('[data-slate-editor]');
+  const editor = window.document.querySelector('[data-slate-editor]')
   const div = window.document.createElement('div')
   div.setAttribute('contenteditable', true)
   div.style.position = 'absolute'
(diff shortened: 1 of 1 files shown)


schneidmaster:
Thanks @jennaplusplus!

jennaplusplus:
Woot! @schneidmaster when can I expect the npm package to be updated with this?

schneidmaster:
Not sure exactly -- @ianstormtaylor handles npm publishing. Generally the package has been published roughly once a week but it depends on commit/PR volume.

PR #1534, opened 2018-01-15
Document: add `change.snapshotSelection` in the change.md

zhujinxuan:
Hi, I think `change.snapshotSelection` is useful in developing some packages with `removeNode`, for example, in the case of https://github.com/GitbookIO/slate-edit-table/issues/40

Therefore, I think it might be better if we have list the api in the document.

### DIFF
--- a/docs/reference/slate/change.md
+++ b/docs/reference/slate/change.md
@@ -474,3 +474,9 @@ Move forward one step in the history.
 `undo() => Change`
 
 Move backward one step in the history.
+
+
+### `snapshotSelection`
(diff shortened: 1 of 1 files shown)


schneidmaster:
@zhujinxuan Thanks for the PR! Sorry to nitpick -- would you mind inserting a newline between lines 480 and 481, and capitalizing "snapshot" in the following sentence? That should make the new entry match the rest of the documentation format 👍 

screenshot 2018-01-16 22 40 37

zhujinxuan:
@schneidmaster Thank you. updated~

Issue #1537, opened 2018-01-16
slate tries to get deleted node in `with-schema.js`


zhujinxuan:
#### Do you want to request a *feature* or report a *bug*?

Perhaps a bug:

#### What's the current behavior?

https://jsfiddle.net/JinxuanZhu/711y5rLe/
The link trys:
1. if a child of `bad-table` is not `bad-row`; unwrap the child; by the default logic of `unwrapNodeByKey`, if `bad-table.nodes.size === 1`, bad-table is deleted
2.  if a child of `bad-row` is not `bad-cell`; unwrap the child; by the default logic of `unwrapNodeByKey`, if `bad-row.nodes.size === 1`, bad-row is deleted

However, it is conflicting with the `with-schema.js` in this part, where `node=refindNode` is undefined

```
    if (change.operations.size > size) {
      node = refindNode(change, node);

      // Add any new children back onto the stack.
      node.nodes.forEach(function (n) {
        if (set.has(n.key)) return;
        keys.unshift(n.key);
      });
    }
```
It seems that the problem is caused by visiting the bad-row after the bad-table normalized...

CameronAckermanSEL:
Strong probability of being a duplicate of https://github.com/ianstormtaylor/slate/issues/1528

Issue #1528, opened 2018-01-13
normalizeNodeAndChildren does not rsync node with change object at enough intervals


CameronAckermanSEL:
#### Do you want to request a *feature* or report a *bug*?

Bug

#### What's the current behavior?

With certain very complicated schema rules, normalizeNodeAndChildren can become badly out of sync with the current document state, even though there are some safeguards against this sort of situation.

The root cause is the code here:
https://github.com/ianstormtaylor/slate/blob/5444a300b8fe86edbe2250da86f11e8a9d0ae776/packages/slate/src/changes/with-schema.js#L86
```
    // Unwind the stack, normalizing every child and adding it to the set.
    while (key = keys[0]) {
      const child = node.getChild(key)
      normalizeNodeAndChildren(change, child, schema)
      set.add(key)
      keys.shift()
    }
```

Here the normalization will recurse into child blocks of the node being validated. Now, if the recursive calls here spawn normalization that causes the descendant structure of the node to change, then the node is out of sync of the change object within this loop. Children of this node are no longer guaranteed to be in sync with the change object, and may not even exist within the node under validation.

Changing the code to this resolves the issue:

```
function normalizeNodeAndChildren(change, node, schema) {
  if (node.object == 'text') {
    normalizeNode(change, node, schema)
    return
  }

  // We can't just loop the children and normalize them, because in the process
  // of normalizing one child, we might end up creating another. Instead, we
  // have to normalize one at a time, and check for new children along the way.
  // PERF: use a mutable array here instead of an immutable stack.
  const keys = node.nodes.toArray().map(n => n.key)

  let { size } = change.operations
  let key

  // PERF: use a mutable set here since we'll be add to it a lot.
  let set = new Set()

  // Unwind the stack, normalizing every child and adding it to the set.
  while (key = keys[0]) {
    set = set.add(key)
    keys.shift()
    const child = node.getChild(key)
    if (child) {
      normalizeNodeAndChildren(change, child, schema)
      // PERF: Only re-find the node and re-normalize any new children if
      // operations occured that might have changed it.
      if (change.operations.size > size) {
        size = change.operations.size
        node = refindNode(change, node)
        // if the node is no longer in the document
        // then break the loop, dont need to normalize
        // it anymore
        if (!node) break

        // Add any new children back onto the stack.
        node.nodes.forEach((n) => {
          if (set.has(n.key)) return
          keys.unshift(n.key)
        })
      }
    }
  }

[...]

Issue #1363, opened 2017-11-01
nested validation stacks break because they've already been normalized


ianstormtaylor:
#### Do you want to request a *feature* or report a *bug*?

Bug.

#### What's the current behavior?

If you write a custom `normalize` function for a schema, and it happens to do something like `change.moveNodeByKey`, there's a chance that calling that change method will initialize its own validation stack, fixing problems that the parent validation stack assumes are still there.

For example, consider that you have two child text nodes adjacent to each other, which is invalid in Slate. If in your `normalize` function you make any change to the parent of those text nodes, that change will end up normalizing internally, and the text nodes will be combined. However, the parent validation stack still thinks that an invalid rule that needs to be fixed, so it tries to normalize them again, and fails.

#### What's the expected behavior?

Not sure.

The ideal would be to prevent the double-checking from happening. I haven't dug into this yet, so I'm not sure if it's impossible to do or not.

Another option would be to automatically pass `options.normalize: false` to all change methods called while performing a validations. The "core" `validateNode` handlers already do this automatically. But this feels like it might be too magic-y?

Have others run into this? Thoughts?

Soreine:
I think I just ran into this https://github.com/ianstormtaylor/slate/issues/1378
In my example, adding `options.normalize: false` to the change done in the custom normalization does not fix the issue.

> However, the parent validation stack still thinks that an invalid rule that needs to be fixed, so it tries to normalize them again, and fails.
What precisely makes it "thinks that it still needs to be fixed"? Is it because we are not "refinding" the nodes correctly after a normalization happened?
Why this problem never arose in the previous implementation?

Should we simply restart validation from the document after every normalization happening ? (heavy...)

alanctkc:
Pretty sure I'm running into this with `wrapBlockByKey`. Here's an example trace (pardon webpack indirection) that shows the multiple normalization passes with `wrapBlockByKey` calling `moveNodeByKey` internally. In the trace, it's isolated down to a single schema rule and a single validation failure:

```
Uncaught Error: Could not find a node with key "3".
    at Document.assertNode (http://localhost:3000/static/js/bundle.js:204111:15)
    at Document.getPath (http://localhost:3000/static/js/bundle.js:205363:24)
    at Document.object.(anonymous function) [as getPath] (http://localhost:3000/static/js/bundle.js:210941:30)
[...]

PR #1538, opened 2018-01-16
Continue normalize next child when and only when the node exists still

zhujinxuan:
Hi, this is a PR regarding:
1. https://github.com/ianstormtaylor/slate/issues/1537 and https://github.com/ianstormtaylor/slate/issues/1528
2. https://github.com/ianstormtaylor/slate/issues/1363 Not try to get the invalid node in parent stack
3. A bit change of code to make the `normalizeNodeAndChildren` shorter; Perhaps this change makes this function a bit more friendly to readers

If you like to have tests for the two issues; I would like to add one~

Thank you very much for your reviewing~

### DIFF
--- a/packages/slate/src/changes/with-schema.js
+++ b/packages/slate/src/changes/with-schema.js
@@ -1,5 +1,4 @@
 
-import { Set } from 'immutable'
 
 /**
  * Changes.
@@ -67,42 +66,28 @@ function normalizeNodeAndChildren(change, node, schema) {
     normalizeNode(change, node, schema)
     return
   }
+  const normalizedKeys = []
+  let child = node.nodes.first()
+
+  let path = change.value.document.getPath(node.key)
 
   // We can't just loop the children and normalize them, because in the process
   // of normalizing one child, we might end up creating another. Instead, we
   // have to normalize one at a time, and check for new children along the way.
-  // PERF: use a mutable array here instead of an immutable stack.
-  const keys = node.nodes.toArray().map(n => n.key)
+  while (node && child) {
+    const lastSize = change.operations.size
+    normalizeNodeAndChildren(change, child, schema)
+    normalizedKeys.push(child.key)
 
-  // While there is still a child key that hasn't been normalized yet...
-  while (keys.length) {
-    const { size } = change.operations
-    let key
-
-    // PERF: use a mutable set here since we'll be add to it a lot.
-    let set = new Set().asMutable()
-
-    // Unwind the stack, normalizing every child and adding it to the set.
-    while (key = keys[0]) {
-      const child = node.getChild(key)
-      normalizeNodeAndChildren(change, child, schema)
-      set.add(key)
-      keys.shift()
-    }
-
-    // Turn the set immutable to be able to compare against it.
-    set = set.asImmutable()
-
-    // PERF: Only re-find the node and re-normalize any new children if
-    // operations occured that might have changed it.
-    if (change.operations.size > size) {
-      node = refindNode(change, node)
-
-      // Add any new children back onto the stack.
-      node.nodes.forEach((n) => {
-        if (set.has(n.key)) return
-        keys.unshift(n.key)
-      })
+    // PERF: if size is unchanged, then no operation happens and we can simply normalize the next child
+    if (lastSize === change.operations.size) {
+      const nextIndex = node.nodes.indexOf(child) + 1
+      child = node.nodes.get(nextIndex)
+    } else {
+      const result = refindNode(change, node, path)
+      path = result.path
+      node = result.node
+      child = node ? node.nodes.find(c => normalizedKeys.indexOf(c.key) === -1) : null
     }
   }
 
@@ -121,12 +106,25 @@ function normalizeNodeAndChildren(change, node, schema) {
  * @return {Node}
  */
 
-function refindNode(change, node) {
+function refindNode(change, node, previousPath) {
   const { value } = change
   const { document } = value
-  return node.object == 'document'
-    ? document
-    : document.getDescendant(node.key)
+
+  if (node.object === 'document') {
+    return { node: document, path: previousPath }
+  }
+
+  // PERF: if we find the right node by the old path, return it; otherwise find the right node by key
+  const tryNode = document.getDescendantAtPath(previousPath)
+  if (tryNode && tryNode.key === node.key) {
+    return { node: tryNode, path: previousPath }
+  }
+  node = document.getDescendant(node.key)
+  if (!node) {
+    return { node, path: [] }
+  }
+  const path = document.getPath(node.key)
+  return { node, path }
 }
 
 /**
@@ -146,11 +144,14 @@ function normalizeNode(change, node, schema) {
     if (!normalize) return
 
     // Run the `normalize` function to fix the node.
+    let path = c.value.document.getPath(n.key)
     normalize(c)
 
     // Re-find the node reference, in case it was updated. If the node no longer
     // exists, we're done for this branch.
-    n = refindNode(c, n)
+    const result = refindNode(c, n, path)
+    n = result.node
+    path = result.path
     if (!n) return
 
     // Increment the iterations counter, and check to make sure that we haven't
--- a/packages/slate/src/changes/with-schema.js
+++ b/packages/slate/src/changes/with-schema.js
@@ -66,9 +66,9 @@ function normalizeNodeAndChildren(change, node, schema) {
     normalizeNode(change, node, schema)
     return
   }
+
   const normalizedKeys = []
   let child = node.nodes.first()
-
   let path = change.value.document.getPath(node.key)
 
   // We can't just loop the children and normalize them, because in the process
@@ -79,15 +79,20 @@ function normalizeNodeAndChildren(change, node, schema) {
     normalizeNodeAndChildren(change, child, schema)
     normalizedKeys.push(child.key)
 
-    // PERF: if size is unchanged, then no operation happens and we can simply normalize the next child
+    // PERF: if size is unchanged, then no operation happens
+    // Therefore we can simply normalize the next child
     if (lastSize === change.operations.size) {
       const nextIndex = node.nodes.indexOf(child) + 1
       child = node.nodes.get(nextIndex)
     } else {
-      const result = refindNode(change, node, path)
-      path = result.path
-      node = result.node
-      child = node ? node.nodes.find(c => normalizedKeys.indexOf(c.key) === -1) : null
+      node = change.value.document.refindNode(path, node.key)
+      if (!node) {
+        path = []
+        child = null
+      } else {
+        path = change.value.document.refindPath(path, node.key)
+        child = node.nodes.find(c => normalizedKeys.indexOf(c.key) === -1)
+      }
     }
   }
 
@@ -95,36 +100,6 @@ function normalizeNodeAndChildren(change, node, schema) {
   if (node) {
     normalizeNode(change, node, schema)
   }
-}
-
-/**
- * Re-find a reference to a node that may have been modified or removed
- * entirely by a change.
- *
- * @param {Change} change
- * @param {Node} node
- * @return {Node}
- */
-
-function refindNode(change, node, previousPath) {
-  const { value } = change
-  const { document } = value
-
-  if (node.object === 'document') {
-    return { node: document, path: previousPath }
-  }
-
-  // PERF: if we find the right node by the old path, return it; otherwise find the right node by key
-  const tryNode = document.getDescendantAtPath(previousPath)
-  if (tryNode && tryNode.key === node.key) {
-    return { node: tryNode, path: previousPath }
-  }
-  node = document.getDescendant(node.key)
-  if (!node) {
-    return { node, path: [] }
-  }
-  const path = document.getPath(node.key)
-  return { node, path }
 }
 
 /**
@@ -149,10 +124,10 @@ function normalizeNode(change, node, schema) {
 
     // Re-find the node reference, in case it was updated. If the node no longer
     // exists, we're done for this branch.
-    const result = refindNode(c, n, path)
-    n = result.node
-    path = result.path
+    n = c.value.document.refindNode(path, n.key)
     if (!n) return
+
+    path = c.value.document.refindPath(path, n.key)
 
     // Increment the iterations counter, and check to make sure that we haven't
     // exceeded the max. Without this check, it's easy for the `validate` or
--- a/packages/slate/src/changes/with-schema.js
+++ b/packages/slate/src/changes/with-schema.js
@@ -91,7 +91,7 @@ function normalizeNodeAndChildren(change, node, schema) {
         child = null
       } else {
         path = change.value.document.refindPath(path, node.key)
-        child = node.nodes.find(c => normalizedKeys.indexOf(c.key) === -1)
+        child = node.nodes.find(c => !normalizedKeys.includes(c.key))
       }
     }
   }


CameronAckermanSEL:
Alright so I spent about 3 days researching this, which culminated in this code snippet that is closer to the existing implementation in case there are any style or performance concerns that led to the specific usage of the key cache approach.

```
function normalizeNodeAndChildren(change, node, schema) {
  if (node.object == 'text') {
    normalizeNode(change, node, schema)
    return
  }

  // We can't just loop the children and normalize them, because in the process
  // of normalizing one child, we might end up creating another. Instead, we
  // have to normalize one at a time, and check for new children along the way.
  // PERF: use a mutable array here instead of an immutable stack.
  const keys = node.nodes.toArray().map(n => n.key)

  let { size } = change.operations
  let key

  // PERF: use a mutable set here since we'll be add to it a lot.
  let set = new Set()

  // Unwind the stack, normalizing every child and adding it to the set.
  while (key = keys[0]) {
    set = set.add(key)
    keys.shift()
    const child = node.getChild(key)
    if (child) {
      normalizeNodeAndChildren(change, child, schema)
      // PERF: Only re-find the node and re-normalize any new children if
      // operations occured that might have changed it.
      if (change.operations.size > size) {
        size = change.operations.size
        node = refindNode(change, node)
        // if the node is no longer in the document
        // then break the loop, dont need to normalize
        // it anymore
        if (!node) break

        // Add any new children back onto the stack.
        node.nodes.forEach((n) => {
          if (set.has(n.key)) return
          keys.unshift(n.key)
        })
      }
    }
  }

  // Normalize the node itself if it still exists.
  if (node) {
    normalizeNode(change, node, schema)
  }
}
```

So whichever the author and the reviewers want to use, just thought I throw it out there so my time was at least well spent.

packages/slate/src/changes/with-schema.js
```diff
@@ -67,42 +66,27 @@ function normalizeNodeAndChildren(change, node, schema) {
...
+      child = node.nodes.get(node.nodes.indexOf(child) + 1)
+    } else {
       node = refindNode(change, node)
-
-      // Add any new children back onto the stack.
-      node.nodes.forEach((n) => {
-        if (set.has(n.key)) return
-        keys.unshift(n.key)
-      })
+      if (!node) {
```
CameronAckermanSEL: this whole conditional can be reduced to 

```
child = node ? node.nodes.find(c => checkedKeys.indexOf(c.key) === -1) : null
```

zhujinxuan: yes;  I prefer to set child null (though un-necessary) when node is void;  Personally I think it is more semantic

CameronAckermanSEL: The recommended conditional does that and removes the unnecessary break statement.

zhujinxuan: Thanks, done~

zhujinxuan:
Update: Speed up refindNode by getPath and getDescandentAtPath
Related commits: 8e146ac and 0bb852a

zhujinxuan:
Benchmark info for path cache update:

zhujinxuan:
> support/benchmark-compare
> 
> 
> 
>   benchmarks
>     html-serializer
>       deserialize
>         7.66 → 9.27 ops/sec
>       serialize
>         826.21 → 1231.49 ops/sec (49% faster)
>     plain-serializer
>       deserialize
>         575.97 → 560.60 ops/sec
>       serialize
>         9888.73 → 10597.30 ops/sec
>     rendering
>       normal
>         81.47 → 109.23 ops/sec (34% faster)
>     changes
>       delete-backward
>         122377.14 → 141842.41 ops/sec
>       delete-forward
>         30732.13 → 23778.45 ops/sec
>       insert-text-by-key-multiple
>         117.89 → 111.57 ops/sec
>       insert-text-by-key
>         1170.36 → 1462.73 ops/sec
>       insert-text
>         1044.16 → 1284.89 ops/sec
>       normalize
>         1989.26 → 2318.63 ops/sec
>       split-block
>         26.89 → 22.89 ops/sec
>     models
>       from-json
>         63.60 → 148.00 ops/sec (133% faster) 🙌
>       get-blocks-at-range
>         862977.63 → 1256478.35 ops/sec (46% faster)
>       get-blocks
>         4587416.59 → 4636871.15 ops/sec
>       get-characters-at-range
>         840762.80 → 985222.91 ops/sec
>       get-characters
>         3165792.71 → 3410707.92 ops/sec
>       get-inlines-at-range
>         885826.57 → 1060447.15 ops/sec
>       get-inlines
>         3471842.60 → 3483210.02 ops/sec
>       get-leaves
>         3490694.82 → 3043713.89 ops/sec
>       get-marks-at-range
>         1019412.74 → 1016667.94 ops/sec
>       get-marks
>         3130672.59 → 3405080.53 ops/sec
>       get-path
>         1204061.92 → 1304849.61 ops/sec
>       get-texts-at-range
>         980118.04 → 993377.17 ops/sec
>       get-texts
>         3231410.99 → 3319266.51 ops/sec
>       has-node-multiple
>         165304.41 → 160362.06 ops/sec
>       has-node
>         1296729.43 → 1137963.17 ops/sec
>       to-json
>         3245.77 → 2737.85 ops/sec
>       update-node
>         18759.41 → 15179.42 ops/sec
> 
> ✨  Done in 62.28s.


Benchmark information after cache nodePath in normalization;

ianstormtaylor:
@zhujinxuan can you explain what's going on here?

zhujinxuan:
@ianstormtaylor Sorry about my poor explanation in the first thread...  

Please forgive my poor English in the following explanation:

## First Problem

The existing problem is that in normalization we must use `{normalize: fase}` in `change.doChange(... {normalize: false})` or `validateNode` because: 
1. everytime a child is normalized, the `normalizeNodeAndChildren` tries to get next normalized child by `node.nodes`.  But 
   1.1 `node` can be non-existent by the `change.operation(...{normalize: true})`.
   1.2  functions like `unwrapNodeByKey`, with `{normalize:false}` can delete the node still
2. If following child is removed, it is possible that trying to normalize the following children causes an error.

## Second Problem

This function pushes all normalized childKeys in a stack, and , if the parent is still valid, normalize the next child of the parent which is not in the stack.

Another thing this PR is trying to resolve is about the refind(change, parent.key):
1. everytime the parent is re-get from document.getDescendant(parent.key); it is not a fast search and refind is called many times (as many as the number of children, or the number of all nodes under parent considering the recursion)
2. the parent of the parent are unlikely to be removed in the normalization

Therefore, the refind function is re-designed as:
1. Trying to reget the parent from the old path
2. If succesfully refind by path, return
3. otherwise, refind by key and update the path

CameronAckermanSEL:
@ianstormtaylor @zhujinxuan Sorry to sort of barge in. I thought it'd be helpful to point out the problems in the original code which were causing some exceptions.

This PR fixes two uncaught exceptions that happen when normalizing deep document node trees.

First error is caused by this loop not synchronizing the local `node` variable after `normalizeNodeAndChildren` is called.

```
  // Unwind the stack, normalizing every child and adding it to the set.
  while (key = keys[0]) {
  const child = node.getChild(key)
  normalizeNodeAndChildren(change, child, schema)  // <-- after this code, the equivalent node in the change object may not have the same children as what is in the `node` local variable, so `child` may not actually exist in the change object when we hit the next iteration. We needed to have another refindNode() call here after normalizeNodeAndChildren. 
  set.add(key)
  keys.shift()
}
```

The second exception happens if `refindNode` returns null.
```
if (change.operations.size > size) {
  node = refindNode(change, node) //<-- in some cases, refindNode might return null here which would cause the following line to explode
 
  // Add any new children back onto the stack.
  node.nodes.forEach((n) => {
    if (set.has(n.key)) return
    keys.unshift(n.key)
  })
}
```

CameronAckermanSEL:
I can't speak to the refindNode performance improvements, but I've been rolling with this on a private branch since the PR was opened and the exceptions have been alleviated at least.

zhujinxuan:
@CameronAckermanSEL @ianstormtaylor 
Thanks for @CameronAckermanSEL 's explanation...

Just want to mention it also fixes a problem (mentioned in https://github.com/ianstormtaylor/slate/issues/1363)
```
  while (key = keys[0]) {
  const child = node.getChild(key)
  normalizeNodeAndChildren(change, child, schema)  // <-- local `node` variable might have oudated children than what the change has after this call, the next loop gives us children that dont exist
```
that:
if a child normalization removes or moves a following child, then the next loop `child = node.getChild(key)` can be null; then it throw an error in `normalizeNodeAndChildren(change, child, schema)`

packages/slate/src/changes/with-schema.js
```diff
@@ -67,42 +66,26 @@ function normalizeNodeAndChildren(change, node, schema) {
     normalizeNode(change, node, schema)
     return
   }
+  const checkedKeys = []
+  let child = node.nodes.first()
+
+  let nodePath = change.value.document.getPath(node.key)
```
ianstormtaylor: Can you change this variable to be just `path`?

packages/slate/src/changes/with-schema.js
```diff
@@ -67,42 +66,26 @@ function normalizeNodeAndChildren(change, node, schema) {
...
-      // Add any new children back onto the stack.
-      node.nodes.forEach((n) => {
-        if (set.has(n.key)) return
-        keys.unshift(n.key)
-      })
+  while (node && child) {
+    const lastSize = change.operations.size
+    normalizeNodeAndChildren(change, child, schema)
+    checkedKeys.push(child.key)
+    if (lastSize === change.operations.size) {
```
ianstormtaylor: Can you add a comment (and a line break) above this line, explaining for future people why we do this check here? It seems like an optimization, so add `PERF: ` before the comment to make that clear.

packages/slate/src/changes/with-schema.js
```diff
@@ -67,42 +66,26 @@ function normalizeNodeAndChildren(change, node, schema) {
...
+      child = node.nodes.get(node.nodes.indexOf(child) + 1)
+    } else {
+      const cacheNodeAndNodePath = refindNode(change, node, nodePath)
```
ianstormtaylor: Can you change this variable to `result`?

packages/slate/src/changes/with-schema.js
```diff
@@ -121,12 +104,22 @@ function normalizeNodeAndChildren(change, node, schema) {
  * @return {Node}
  */
 
-function refindNode(change, node) {
+function refindNode(change, node, lastPath) {
```
ianstormtaylor: Can you change this variable to `previousPath`?

packages/slate/src/changes/with-schema.js
```diff
@@ -121,12 +104,22 @@ function normalizeNodeAndChildren(change, node, schema) {
...
+  const tryNode = document.getDescendantAtPath(lastPath)
+  if (tryNode && tryNode.key === node.key) {
+    return { node: tryNode, nodePath: lastPath }
+  }
+  node = document.getDescendant(node.key)
+  if (!node) {
+    return { node, nodePath: [] }
+  }
+  const nodePath = document.getPath(node.key)
+  return { node, nodePath }
```
ianstormtaylor: Can you change these return values to `{ node, path }`?

packages/slate/src/changes/with-schema.js
```diff
@@ -146,11 +139,14 @@ function normalizeNode(change, node, schema) {
     if (!normalize) return
 
     // Run the `normalize` function to fix the node.
+    let nPath = c.value.document.getPath(n.key)
```
ianstormtaylor: Can you change this variable to `path`?

packages/slate/src/changes/with-schema.js
```diff
@@ -146,11 +139,14 @@ function normalizeNode(change, node, schema) {
...
     normalize(c)
 
     // Re-find the node reference, in case it was updated. If the node no longer
     // exists, we're done for this branch.
-    n = refindNode(c, n)
+    const cacheNAndPath = refindNode(c, n, nPath)
```
ianstormtaylor: Can you change this variable to `result`?

packages/slate/src/changes/with-schema.js
```diff
@@ -67,42 +66,26 @@ function normalizeNodeAndChildren(change, node, schema) {
...
   }
+  const checkedKeys = []
+  let child = node.nodes.first()
+
+  let nodePath = change.value.document.getPath(node.key)
 
   // We can't just loop the children and normalize them, because in the process
   // of normalizing one child, we might end up creating another. Instead, we
   // have to normalize one at a time, and check for new children along the way.
   // PERF: use a mutable array here instead of an immutable stack.
```
ianstormtaylor: I think this last line with the `PERF:` comment might be outdated and we can remove it?

packages/slate/src/changes/with-schema.js
```diff
@@ -67,42 +66,26 @@ function normalizeNodeAndChildren(change, node, schema) {
     normalizeNode(change, node, schema)
     return
   }
+  const checkedKeys = []
```
ianstormtaylor: Can you change this variable to `keys`?

packages/slate/src/changes/with-schema.js
```diff
@@ -67,42 +66,26 @@ function normalizeNodeAndChildren(change, node, schema) {
...
-      node.nodes.forEach((n) => {
-        if (set.has(n.key)) return
-        keys.unshift(n.key)
-      })
+  while (node && child) {
+    const lastSize = change.operations.size
+    normalizeNodeAndChildren(change, child, schema)
+    checkedKeys.push(child.key)
+    if (lastSize === change.operations.size) {
+      child = node.nodes.get(node.nodes.indexOf(child) + 1)
```
ianstormtaylor: Can you store this in an interim variable of `nextIndex`, so that it's more clear?

packages/slate/src/changes/with-schema.js
```diff
@@ -67,42 +66,26 @@ function normalizeNodeAndChildren(change, node, schema) {
...
+    } else {
+      const cacheNodeAndNodePath = refindNode(change, node, nodePath)
+      nodePath = cacheNodeAndNodePath.nodePath
+      node = cacheNodeAndNodePath.node
+      child = node ? node.nodes.find(c => checkedKeys.indexOf(c.key) === -1) : null
```
ianstormtaylor: Can you use `Array.includes` here instead for terseness?

packages/slate/src/changes/with-schema.js
```diff
@@ -121,12 +104,22 @@ function normalizeNodeAndChildren(change, node, schema) {
...
  */
 
-function refindNode(change, node) {
+function refindNode(change, node, lastPath) {
   const { value } = change
   const { document } = value
-  return node.object == 'document'
-    ? document
-    : document.getDescendant(node.key)
+  if (node.object === 'document') {
```
ianstormtaylor: Can you add breaks about these `if` statements to match the rest of the code?

packages/slate/src/changes/with-schema.js
```diff
@@ -121,12 +104,22 @@ function normalizeNodeAndChildren(change, node, schema) {
...
+    return { node: document, nodePath: lastPath }
+  }
+  const tryNode = document.getDescendantAtPath(lastPath)
```
ianstormtaylor: Can you add a `PERF: ` comment above this line, explaining why this is a shortcut that speeds things up lots of the time?

packages/slate/src/changes/with-schema.js
zhujinxuan: Hi, is there any special reasons about changing the var name?  Personally I would prefer a bit adjective on the var name to help me remember what I have done when I read the code again.

packages/slate/src/changes/with-schema.js
zhujinxuan: Yes, I will change it.

packages/slate/src/changes/with-schema.js
zhujinxuan: Yes, I will delete it.

packages/slate/src/changes/with-schema.js
zhujinxuan: Yes, I will add the comment

packages/slate/src/changes/with-schema.js
zhujinxuan: Yes, I will change it.

packages/slate/src/changes/with-schema.js
zhujinxuan: Yes, I will.

packages/slate/src/changes/with-schema.js
zhujinxuan: Yes, I will.

packages/slate/src/changes/with-schema.js
zhujinxuan: `Array.includes` is not supported by IE, and I do not know whether babel fixes this problem. 
If babel can resolve the problem, I will use `Array.includes`

See browser support by:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

packages/slate/src/changes/with-schema.js
zhujinxuan: Yes, I will

packages/slate/src/changes/with-schema.js
zhujinxuan: Yes, I will.

packages/slate/src/changes/with-schema.js
zhujinxuan: Yes, I will.

packages/slate/src/changes/with-schema.js
zhujinxuan: Yes, I will.

packages/slate/src/changes/with-schema.js
zhujinxuan: Yes, I will

zhujinxuan:
Commit fixed:
About `Array.includes`, the function is not supported by IE.  I am not sure whether babel supported the convection from includes => indexOf by default.

packages/slate/src/changes/with-schema.js
ianstormtaylor: Fair enough, in that case can you call it `normalizedKeys` instead?

zhujinxuan: Fixed by directly update on the last commit~

CameronAckermanSEL:
Hello! It's been a few days so I thought I'd check in. This one is a pretty big deal for us because it's breaking our paste handling. 

Thanks again for seeing this to completion. Can't wait to use it!

zhujinxuan:
Give me a second... I mistakenly deleted my branch

zhujinxuan:
Update: use the `refindNode` and `refindPath` in the merged PR.

packages/slate/src/changes/with-schema.js
```diff
@@ -67,42 +66,32 @@ function normalizeNodeAndChildren(change, node, schema) {
     normalizeNode(change, node, schema)
     return
   }
+  const normalizedKeys = []
+  let child = node.nodes.first()
+
+  let path = change.value.document.getPath(node.key)
```
ianstormtaylor: Can you fix these line breaks to match the rest of the code?

packages/slate/src/changes/with-schema.js
```diff
@@ -67,42 +66,32 @@ function normalizeNodeAndChildren(change, node, schema) {
...
-      node.nodes.forEach((n) => {
-        if (set.has(n.key)) return
-        keys.unshift(n.key)
-      })
+  while (node && child) {
+    const lastSize = change.operations.size
+    normalizeNodeAndChildren(change, child, schema)
+    normalizedKeys.push(child.key)
+
+    // PERF: if size is unchanged, then no operation happens and we can simply normalize the next child
```
ianstormtaylor: Can you fix this comment style to match the rest of the code?

packages/slate/src/changes/with-schema.js
```diff
@@ -67,42 +66,32 @@ function normalizeNodeAndChildren(change, node, schema) {
     normalizeNode(change, node, schema)
     return
   }
+  const normalizedKeys = []
```
ianstormtaylor: Can you add a line break here to match the rest of the code?

packages/slate/src/changes/with-schema.js
```diff
@@ -146,12 +118,14 @@ function normalizeNode(change, node, schema) {
...
     // Run the `normalize` function to fix the node.
+    let path = c.value.document.getPath(n.key)
     normalize(c)
 
     // Re-find the node reference, in case it was updated. If the node no longer
     // exists, we're done for this branch.
-    n = refindNode(c, n)
+    n = c.value.document.refindNode(path, n.key)
     if (!n) return
+    path = c.value.document.refindPath(path, n.key)
```
ianstormtaylor: Can you add a line break after this early return to match the rest of the code?

zhujinxuan: Fixed. Thank you

packages/slate/src/changes/with-schema.js
zhujinxuan: Fixed. Thank you

packages/slate/src/changes/with-schema.js
zhujinxuan: Fixed. Thank you

packages/slate/src/changes/with-schema.js
zhujinxuan: Fixed. Thank you

zhujinxuan:
@ianstormtaylor Hi, I have changed that to `array.includes`, and fix the styling problems.

ianstormtaylor:
Thanks @zhujinxuan!

PR #1539, opened 2018-01-17
Upgrade to lerna 2.7 to correct peer dependency management

### LINKED ISSUE #1414: peer dependency management is not correct
Right now when a new minor version of Slate is published, the `peerDependencies` of all of the other packages are also bumped to that minor. This doesn't make sense, because the peer dependencies should stay as loose as possible. 

For instance, in most cases when `slate` is upgraded, `slate-prop-types` is unaffected and can happily keep using older versions. But it is forced to upgrade currently regardless. This causes extra, unnecessary dependency upgrades to be forced on users.

This is actually due to a bug in Lerna — https://github.com/lerna/lerna/issues/1018

If anyone was up for getting a pull request merged into Lerna to fix this behavior I would be super, super grateful. It would save a lot of headaches in the Slate ecosystem.

schneidmaster:
[lerna 2.7](https://github.com/lerna/lerna/releases/tag/v2.7.0) included a fix for peer dependency management (not bumping peer dependencies when a minor version is published)

Fixes #1414


ianstormtaylor:
Thanks @schneidmaster!

PR #1540, opened 2018-01-17
Use absolute links in changelog so they render correctly in GitBook.

### LINKED ISSUE #1507: Changelog links don't work
https://docs.slatejs.org/general/changelog

schneidmaster:
Fixes #1507

### DIFF
--- a/docs/Readme.md
+++ b/docs/Readme.md
@@ -27,7 +27,7 @@
 - [Plugins](./general/plugins.md)
 - [Resources](./general/resources.md)
 - [Contributing](../Contributing.md)
-- [Changelog](../Changelog.md)
+- [Changelog](./general/changelog.md)
 - [FAQ](./general/faq.md)
 
 


ianstormtaylor:
@schneidmaster good catch! I think changing the link in `./docs/Readme.md` to `./docs/general/changelog.md` would fix this too actually, and potentially be a better solution?

schneidmaster:
To be honest I don't know anything about how Gitbook works, I was just following their suggestion. I'll change to your solution though and we'll see if it does the trick 👍

schneidmaster:
@ianstormtaylor Actually, I'm not sure if your solution will do the trick. The changelog in the docs sidebar already links to https://docs.slatejs.org/general/changelog which I assume is the one under the `docs/general` folder (not the root `Changelog.md`).

schneidmaster:
Huh, I do see that clicking `Edit on GitHub` on the current `Changelog` does go to the root-level one. Maybe I'm just missing something in how Gitbook generates the directory structure. Anyway, updated per your suggestion.

SamyPesse:
The new GitBook can handle entries in the `./docs/Readme.md` that point to files outside the `./docs` folder.

The issue with #1507 is a bug that has been fixed month ago (GitBook is still in beta), but GitBook only import changes and `Changelog.md` has not changed since then.

Any small changes to this file should import it correctly. Basically both of your solutions will work.

If you don't want to merge this PR, I can do some work to run a full import of the slate repository.

ianstormtaylor:
Thanks @schneidmaster and @SamyPesse for the explanation! I'll merge this then.

Issue #1449, opened 2017-12-05
Can't copy single void block 


tobiasandersen:
#### Do you want to request a *feature* or report a *bug*?
Bug

#### What's the current behavior?
If you select a single void block, and press cmd+c (mac) to copy, and then try to paste it somewhere, it won't be pasted. However, it works if it's part of a larger selection range. 

**To reproduce:**
Go to http://slatejs.org/#/images, click the image (which should highlight that it's selected), press cmd+c and try to paste it somewhere. 

I'm on Chrome Version 62.0.3202.94 (Official Build) (64-bit) on mac.

dmitrizzle:
I feel like this may be because of different types of selections. When you select text, that's native. While clicking void block that's something else, which causes this bug. I think...

zGrav:
In my case, it was possible to "bypass" this by doing a custom copy/paste handler... a bit annoying but it was the only solution we found.

dmitrizzle:
Would love to see a snippet @zGrav - I think it might be a valid solution for Slate core (since void blocks are special case either way) - or at least as a patch in my code...

zGrav:
@dmitrizzle https://gist.github.com/zGrav/440fdbe83d7eca8455362ccfd2e58662

dmitrizzle:
Thank you!

skogsmaskin:
This is a bug which has been there for long as I can remember. It would be wonderful to get this fixed properly!

PR #1543, opened 2018-01-17
make sure void nodes are always fully copied

tobiasandersen:
This is an attempt at solving https://github.com/ianstormtaylor/slate/issues/1449. 

The problem is that if we select only a single void node, browsers will (for some reason I haven't been able to find out) copy only its text – rather than the whole node. This becomes a problem down the line [when we try to find that fragment again](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/get-event-transfer.js#L53) when pasting. 

I tried a bunch of different stuff, but the approach I went with in the end is to surround the content we want to copy with two empty textareas. So instead of having:
```
<div>[contents]</div>
```
we now have:
```
<div>
  <textarea />
  [contents]
  <textarea />
</div>
```

This works for me in latest Chrome, Firefox and Safari (on mac). In Chrome and Firefox it was actually enough to simply add a space character to the "phony" div before appending the contents:

```
div.innerHTML = '  '
div.appendChild(contents)
```

But this failed in Safari, and would in some cases (depending on the void node's content) copy a space to the default clipboard. 

I've tested this on the example site, and in our app where we do a lot of void node copying, and I haven't seen any problems with it so far. And what I especially like about it is that we don't have to add any extra text that could potentially mess up the standard clipboard. 

Even though I've tested it quite a bit already, I'd love it if other people with this problem (cc @skogsmaskin) could give it a try as well.

### DIFF
--- a/packages/slate-react/src/utils/clone-fragment.js
+++ b/packages/slate-react/src/utils/clone-fragment.js
@@ -4,6 +4,9 @@ import getWindow from 'get-window'
 import findDOMNode from './find-dom-node'
 import { ZERO_WIDTH_SELECTOR, ZERO_WIDTH_ATTRIBUTE } from './find-point'
 import { IS_CHROME, IS_SAFARI } from '../constants/environment'
+import TRANSFER_TYPES from '../constants/transfer-types'
+
+const { FRAGMENT, HTML, TEXT } = TRANSFER_TYPES
 
 /**
  * Prepares a Slate document fragment to be copied to the clipboard.
@@ -95,38 +98,17 @@ function cloneFragment(event, value, fragment = value.fragment) {
 
   attach.setAttribute('data-slate-fragment', encoded)
 
-  // Add the phony content to the DOM, and select it, so it will be copied.
-  const editor = event.target.closest('[data-slate-editor]')
+  // In order to populate the HTML clipboard register, we need a string
+  // representation of the content. We can get this by adding the phony content
+  // to a div and then reading its innerHTML.
   const div = window.document.createElement('div')
-  div.setAttribute('contenteditable', true)
-  div.style.position = 'absolute'
-  div.style.left = '-9999px'
-
-  // COMPAT: In Firefox, the viewport jumps to find the phony div, so it
-  // should be created at the current scroll offset with `style.top`.
-  // The box model attributes which can interact with 'top' are also reset.
-  div.style.border = '0px'
-  div.style.padding = '0px'
-  div.style.margin = '0px'
-  div.style.top = `${window.pageYOffset ||
-    window.document.documentElement.scrollTop}px`
-
   div.appendChild(contents)
-  editor.appendChild(div)
 
-  // COMPAT: In Firefox, trying to use the terser `native.selectAllChildren`
-  // throws an error, so we use the older `range` equivalent. (2016/06/21)
-  const r = window.document.createRange()
-  r.selectNodeContents(div)
-  native.removeAllRanges()
-  native.addRange(r)
-
-  // Revert to the previous selection right after copying.
-  window.requestAnimationFrame(() => {
-    editor.removeChild(div)
-    native.removeAllRanges()
-    native.addRange(range)
-  })
+  // Add appropriate data to the clipboard registers.
+  event.preventDefault()
+  event.clipboardData.setData(TEXT, native.toString())
+  event.clipboardData.setData(FRAGMENT, encoded)
+  event.clipboardData.setData(HTML, div.innerHTML)
 }
 
 export default cloneFragment


dmitrizzle:
Hey @tobiasandersen this is awesome! Have you tested it with images that have editable captions? In my Editor config my void blocks are `<Figure />` elements that have `<picture />` and `<figcaption><textarea/></figcaption>` - I have the same issue of them being selectable but not copyable to clipboard.

tobiasandersen:
No I haven't tried that exact use case. But we do have editable content in our void nodes. If you could try this in your app I'd really appreciate it.

dmitrizzle:
Will definitely do that tonight or tomorrow morning 👍

dmitrizzle:
OK I've tested it with my setup. Works without errors. Here are some things I've noticed:

Interestingly enough, copying single void element was working for me before applying your patch (I guess there were updates to Slate recently that have fixed the issue and I missed them). I presume exactly because I have had a textfield attached there. Here's what's also interesting: I have button overlays on my `<figure/>` elements that would end up being copied to clipboard and be added to the editor before applying your patch: https://github.com/dmitrizzle/Analog.Cafe/issues/836 -- however your code seems to have fixed that issue (which is great! but I'm not sure why that worked).

One minor issue with this method is that pasting images adds an empty block right on top, which seems identical to how `slate-drop-or-paste-images` behaves: https://github.com/dmitrizzle/Analog.Cafe/issues/804

dmitrizzle:
Any updates on this PR?

tobiasandersen:
Please hold off on merging this, it adds adds extra line breaks before and after the content. I'll see if I can come up with another solution... 🙃

tobiasandersen:
Been looking into it some more. Safari behaves completely different than Chrome and Firefox. I'll try to sum up my findings below.

**The problem**
Currently when copying, we wrap the copied content in a "phony"  div like this: 
```
<div contenteditable="true">
  [contents]
</div>
```
This isn't enough (in any browser) when only a void node is selected. In order to force browsers to copy the void node, we need to add some content around it. 

**Safari**
The only way I've got it to work in Safari is to wrap the content within two textareas (or inputs):
```
<div contenteditable="true">
  <textarea />
    [contents]
  <textarea />
</div>
```
I have yet to find any drawbacks with this in Safari. 

**Chrome and Firefox**
The "textarea trick" above works in Chrome and Firefox as well, but it has a major drawback — now **everything** you copy will be wrapped in newlines. So this trick clearly isn't good enough. What works pretty well though is to simply add a span width a zero width space character before the content:
```
<div contenteditable="true">
  <span>\u200B</span>
  [contents]
</div>
```
The only problem I've found with this is that the zero width space character also will be copied. So everything will always be prepended with this character. 

-------------------

I'm not really sure what the way forward is from here, but here're a few alternatives:

1. Only do weird stuff like this when the selection contains only a single void node. 
2. Go through the selection and "fix" void nodes by wrapping them in something.
3. Accept the extra zero width space character in Chrome and Safari.

No matter what, I think we'll have to handle Safari differently.

benjycui:
Also close: https://github.com/ianstormtaylor/slate/issues/1693

zhujinxuan:
@tobiasandersen Perhaps we can consider Plain.serialize for `\n`.

ianstormtaylor:
@tobiasandersen I think the `clipboardData.setData` API solution is super nice. The only thing that's unfortunate is that I think both IE11 and Edge don't support it? (Although Edge is supposedly working on it?) Which means those both break.

I personally don't care about them, but right now Slate has been acting to support them.

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/commit/f19cde92c735a83d003d3cb4c82a15d9c815bdac. Thank you for taking the time to do all this research @tobiasandersen, it made me realize that other potential solution.

zhujinxuan:
@ianstormtaylor  f19cde9 solves the startVoid problem.  But there are two other bugs fixed by this PR but not fixed by  f19cde9;

The two are https://github.com/ianstormtaylor/slate/issues/1693 and https://github.com/ianstormtaylor/slate/issues/1799
Perhaps we need `event.preventDefault()` ?

Perhaps we can do things like this:

```
if ( ! is_IE && !is_Edge) {
  event.preventDefault()
   event. clipboardData.setData.....
} else {
  // append div solution
}
```

tobiasandersen:
Totally forgot about support for those browsers (we don't support them). 

While the fix in https://github.com/ianstormtaylor/slate/commit/f19cde92c735a83d003d3cb4c82a15d9c815bdac will help in a lot of cases, it'll still fails in some. E.g. our void nodes look kind of like this:

```html
<div className='container'>
  <div className='gutter'></div>
  <div className='content'></div>
  <div className='gutter'></div>
</div>
```

Now, with your fix, the attachment node will be set to `div.container` (which is what I'd expect): 
```js
if (startVoid) {
  attach = contents.childNodes[0].childNodes[1].firstChild  
}

```
But Chrome will for some reason copy only `div.content` into the `text/html` register — which in turn makes [this check](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/get-event-transfer.js#L43) fail (and thus the whole paste). 

Personally I'd prefer doing what @zhujinxuan suggested, and use the Clipboard API for browsers that support it (since we eliminate all these weird quirks), and then use the current solution for other browsers. But then again, we don't support IE/Edge so I'm quite biased. 

Another solution I could think of (not sure how nice it is though), is to make `cloneFragment` return everything that's useful. Then users who don't care about IE/Edge could just do this:
```js
 const { text, fragment, html } = cloneFragment()

 event.preventDefault()
 event.clipboardData.setData(TEXT, text)
 event.clipboardData.setData(FRAGMENT, fragment)
 event.clipboardData.setData(HTML, html)
```

@ianstormtaylor I'd really appreciate your thoughts here!

ianstormtaylor:
I’m down with having it use setData for newer browsers and move the existing logic into a fallback!

tobiasandersen:
I've created https://github.com/ianstormtaylor/slate/pull/1815 for this now (couldn't reopen this PR).

PR #1545, opened 2018-01-18
[WIP] fix expanded range changes when last block is empty

tobiasandersen:
This should fix https://github.com/ianstormtaylor/slate/issues/1452 and https://github.com/ianstormtaylor/slate/issues/1447. 

The PR isn't ready yet, but I need som eyes on it before continuing. I'm trying to write tests for all these, but have some trouble. Please see my comments below.

### DIFF
--- a/packages/slate/test/changes/at-current-range/insert-block/block-end-empty.js
+++ b/packages/slate/test/changes/at-current-range/insert-block/block-end-empty.js
@@ -1,0 +1,30 @@
+/** @jsx h */
+
+import h from '../../../helpers/h'
+
+export default function (change) {
+  change.insertBlock('quote')
+}
+
+export const input = (
+  <value>
+    <document>
+      <paragraph>
+        <anchor />word
+      </paragraph>
+      <paragraph>
+        <focus />
+      </paragraph>
+    </document>
+  </value>
+)
+
+export const output = (
+  <value>
+    <document>
+      <quote>
+        <cursor />
+      </quote>
+    </document>
+  </value>
+)
--- a/packages/slate/test/changes/at-current-range/insert-fragment/end-block-empty.js
+++ b/packages/slate/test/changes/at-current-range/insert-fragment/end-block-empty.js
@@ -1,0 +1,36 @@
+/** @jsx h */
+
+import h from '../../../helpers/h'
+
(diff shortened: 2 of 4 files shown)


packages/slate/test/changes/at-current-range/insert-fragment/end-block-empty.js
```diff
@@ -0,0 +1,36 @@
...
+      </paragraph>
+    </document>
+  </value>
+)
+
+export const output = (
+  <value>
+    <document>
+      <quote>
+        fragment<cursor />
```
tobiasandersen: This currently fails. What works is:
```
<cursor />fragment
```
But is that really how it should be?

packages/slate/test/changes/at-current-range/insert-block/block-end-empty.js
```diff
@@ -0,0 +1,30 @@
...
+      </paragraph>
+    </document>
+  </value>
+)
+
+export const output = (
+  <value>
+    <document>
+      <quote>
+        <cursor />
```
tobiasandersen: This test is failing. I end up with two blocks rather than one. Am I doing something wrong here?

packages/slate/test/changes/at-current-range/split-block/block-end-empty.js
```diff
@@ -0,0 +1,31 @@
+/** @jsx h */
+
+import h from '../../../helpers/h'
+
+export default function (change) {
```
tobiasandersen: This test is working...

tobiasandersen:
https://github.com/ianstormtaylor/slate/pull/1545 is also related. My changes prevents it from throwing, but it still isn't handled correctly.

tobiasandersen:
We've been using these fixes for some time now without any problems. The test results seems to be a bit off(?), so maybe it's better to move those into their own issues, and merge the fixes without the failing tests?

ianstormtaylor:
@tobiasandersen I'm not sure I understand these.

It seems like setting to `change.value.selection` is tightly coupling `*AtRange` methods, which are supposed to work on any arbitrary range in the document, to only work on the specific use case of the current "selection" range.

Issue #1408, opened 2017-11-15
consider using paths instead of keys


ianstormtaylor:
This is just an idea.

Right now we use the concept of unique "keys" to keep track of the `anchor` and `focus` of a node in a range. And we use them to look up a node by key in the document, for updating, etc.

But keys are local-only, and once we convert the change information into operations, we convert "keys" into "paths", which are just an array of indices that locate a node in the document—like `[2, 1, 2]`. Because these don't rely on local information.

But I was thinking... would it be an improvement if we actually just used paths everywhere? and did away with the idea of "keys" entirely?

---

###### PROS

- For many things, like "find node by key", the "find node by path" equivalent would actually be much quicker, since you wouldn't have to search the entire tree, you'd just know exactly where to descend by just following the path.
- It would also make operations expressed in the same terms as the local changes are applied.
- When inserting fragments or nodes, we wouldn't have to do the current behavior where we iterate all the new nodes, insuring that their keys aren't already in the document—which would save performance.
- I think a lot of the logic in the "at current range" transforms that determines what the selection will be after the change has applied could be simplified, because path changes are deterministic, compared to key changes which cannot be guessed ahead of time since they are unique.

###### CONS

- Finding a node in the DOM currently searches for its "key", which doesn't change across renders. But if paths were used instead then whenever a node was inserted, each node after it would have to re-render to update DOM attributes. This could potentially be mitigated by moving the key generation into the view layer, as a view-level concern instead of the data layer. Keeping it around only for locating nodes in the DOM.
- To update a specific node, you no longer have the `key` reference. This could be solved by either searching for the `node` instance itself (with the same performance). Or it could be by passing the `path` to places that need the node. The path approach is awkward though, since it means the information needed can change without the node itself changing. But that's okay, because comparing by instance would work still I think, so not really an issue.

---

Would love to hear people's thoughts on this.

justinweiss:
I like the idea of using paths -- they're self-describing and, like you said, will probably be faster.

If we lost keys, this is the first possible problem that comes to my mind: Let's say I have a list block. I insert a paragraph block before the list block. Now any reference to the list's block key will point at the paragraph, instead. If keys aren't immutable references anymore, how can you refer to the same node across changes?

ianstormtaylor:
@justinweiss can you describe your use case a bit more? I think there's a few scenarios for that...

[...]

PR #1548, opened 2018-01-22
Faster (but also safe) method about re-find node by path

zhujinxuan:
Hi, this is related to https://github.com/ianstormtaylor/slate/issues/1408, last two commits by https://github.com/ianstormtaylor/slate/pull/1538;  The path provides two methods to update node/path after change happens.

Cheers,
Jinxuan Zhu

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1352,6 +1352,41 @@ class Node {
   }
 
   /**
+   * Refind the path of node if path is changed.
+   *
+   * @param {Array} path
+   * @param {String} key
+   * @return {Array}
+   */
+
+  refindPath(path, key) {
+    const node = this.getDescendantAtPath(path)
+    if (node && node.key === key) {
+      return path
+    }
+
+    return this.getPath(key)
+  }
+
+  /**
+   *
+   * Refind the node with the same node.key after change.
+   *
+   * @param {Array} path
+   * @param {String} key
+   * @return {Node|Void}
+   */
+
+  refindNode(path, key) {
+    const node = this.getDescendantAtPath(path)
+    if (node && node.key === key) {
+      return node
+    }
+
+    return this.getDescendant(key)
+  }
+
+  /**
    * Get the placeholder for the node from a `schema`.
    *
    * @param {Schema} schema


ianstormtaylor:
@zhujinxuan can you explain what these are for? why they're good?

zhujinxuan:
When we develop normalize or range-related functions, there is always functions like
```
parent =.....
// some changes happens
// refind the parent after change 
parent = change.value.document.getNode(parent.key)
```
But the path of parent is often not changed after change; then we can 
```
parent = ...
path = document.getPath(parent.key)
// some changes happens
// refind the parent after change 
parent = change.value.document.refindNode(path, parent)
path = change.value.document.refindPath(path, parent)
```

I think the two functions can help plugin developers speed up some normalize functions.

zhujinxuan:
Commit fixed

packages/slate/src/models/node.js
```diff
@@ -1351,6 +1351,41 @@ class Node {
...
 
+  /**
+   * refind the path of node if path is changed
+   *
+   * @param {Array} path
+   * @param {Node|String} node
+   * @return {Array}
+   */
+
+  refindPath(path, node) {
```
ianstormtaylor: These methods should take `key` instead of `node`, to match how we handle it elsewhere I think.

packages/slate/src/models/node.js
```diff
@@ -1351,6 +1351,41 @@ class Node {
...
+    const key = (typeof node === 'string') ? node : node.key
+    const tryNode = this.getDescendantAtPath(path)
+    if (tryNode && tryNode.key === key) {
```
ianstormtaylor: Can you add line breaks around this to match the rest of the codebase?

packages/slate/src/models/node.js
```diff
@@ -1351,6 +1351,41 @@ class Node {
     return path
   }
 
+  /**
+   * refind the path of node if path is changed
```
ianstormtaylor: Can you capitalize this comment and add a period to the end to match the rest of the codebase?

packages/slate/src/models/node.js
```diff
@@ -1351,6 +1351,41 @@ class Node {
...
+   * refind the node with the same node.key after change
+   *
+   * @param {Array} path
+   * @param {Node|String} node
+   * @return {Node|Void}
+   */
+
+  refindNode(path, node) {
+    const key = (typeof node === 'string') ? node : node.key
+    const tryNode = this.getDescendantAtPath(path)
```
ianstormtaylor: Once `node` is changed to `key`, then this variable can be called `node` instead of `tryNode`.

packages/slate/src/models/node.js
```diff
@@ -1351,6 +1351,41 @@ class Node {
...
+  /**
+   *
+   * refind the node with the same node.key after change
+   *
+   * @param {Array} path
+   * @param {Node|String} node
+   * @return {Node|Void}
+   */
+
+  refindNode(path, node) {
```
ianstormtaylor: Can you change these methods to be called `refresh` instead of `refind`? I want to avoid confusion with the existing `find` methods which take iterators instead.

zhujinxuan:
@ianstormtaylor Most comments are fixed and I update on the original commit directly.

But there is still one thing I am feeling unsure about: about function name of `refind` or `refresh` or other potential names.  Personally I think refresh is not a good name because it does not appear to be a getter function returning a node or path.  

I chose the `refind` function name because this function was originally named in the `with-schema.js` as `refind(change, key)`.  Perhaps we can think of some better names.  Personally, I do not think refind is a very bad name, because its return behaves much alike the node.get... .

ianstormtaylor:
Fair enough @zhujinxuan, I think that makes sense as an argument. We can merge this for now and always deprecate it later if we find a better name.

PR #1549, opened 2018-01-22
Improve normalize suppression to make less verbose and safer 

### LINKED ISSUE #1544: Would it be good to have `setOperationFlag('normalize', false)`

features;

In transforming node from one type to another type, or in customizing methods of delete/paste fragment at range, `{normalize: false}` is often applied until the final stage.  Because in the middle of the operations(for example, set type in a nested chain of blocks), normalization is often bad. 

However, writing `{normalize: false}` everywhere is a bit wordy. Would you think whether it is good to have a PR about `setOperationFlag('normalize', false)`?

CameronAckermanSEL:
Fixes #1544 

Added a function called `change.withMutations()`, which mirrors immutable's `withMutations` api. It accepts a changeFunction which has one or more change transformation calls, and suppresses normalization until all changes are complete. Once the changeFunction returns, normalization is enabled and then executed against the entire document.

Should solve the following problems:

1. normalization suppression is a hidden, undocumented behavior for safety reasons, but it is very important because unexpected/unintentional validation can break slate.
2. Users who have been suppressing normalization may be subjecting themselves to defects caused by the document being in an unnormalized state at the end of a validation function. `{ normalize: false}` used at the final call of a normalization function may leave the document in unknown state in certain control flows and this is dangerous.
3. `{ normalize: false }` can create a lot of unweildly boiler plate over time.


packages/slate/src/changes/by-key.js
```diff
@@ -12,6 +12,26 @@ import Node from '../models/node'
...
+/**
+ * Obtains the normalization setting to use for the change based on the
+ * change object flags and the change operation options.
+ *
+ * @param {Object} flags - flags object for the current change object
+ * @param {Object} options - options for the change operation
+ * @return {bool} - the normalization configuration to use. Defauls to true
+ *   if flags or options does not have a normalization configuration
+ */
+
```
CameronAckermanSEL: order is important here because we do not want the user to override the normalize flag with { normalize: false }. This may be a debatable decision, but this achieves the desirable effect of withMutations having full suppression of normalization within its scope. Allowing options.normalize to override flags would allow the user to enable normalization early within withMutations and break the contract of that function/create defects

packages/slate/src/models/change.js
```diff
@@ -48,7 +48,7 @@ class Change {
     const { value } = attrs
     this.value = value
     this.operations = new List()
-    this.flags = pick(attrs, ['merge', 'save'])
+    this.flags = pick(attrs, ['merge', 'save', 'normalize'])
   }
```
CameronAckermanSEL: the normalize flag can be set with `change.setOperationFlag()` or in the constructor. This is dangerous for several reasons so it remains an undocumented feature. We are hoping that `withMutations` will be a pleasant enough interface that people will not be tempted to use `change.setOpeartionFlag` for this purpose anyway.

packages/slate/src/changes/by-key.js
```diff
@@ -12,6 +12,26 @@ import Node from '../models/node'
 
 const Changes = {}
 
+/**
+ * Obtains the normalization setting to use for the change based on the
+ * change object flags and the change operation options.
+ *
+ * @param {Object} flags - flags object for the current change object
+ * @param {Object} options - options for the change operation
+ * @return {bool} - the normalization configuration to use. Defauls to true
```
YurkaninRyan: typo here `Defauls`

YurkaninRyan:
The more I look over this PR the more that I love it :+1:  Can't wait to have a more standard way of doing something like this.

Would be cool to get some documentation for it rolling to help newer people understand the purpose of it, I'm gonna think about what that should look like over the next few days if this gets in soon

CameronAckermanSEL:
I'm glad you brought the learning curve up because I've been thinking about it for a while. We can probably add something under `guides/schemas` but I haven't come up with any great ideas. I'll post some samples here in the conversation and we can fiddle with it and once we get something  we like I'll add it to the PR

CameronAckermanSEL:
How about this? ## Multi-step Normalizations

Some normalizations will require multiple `change` function calls in order to complete. 

Consider the following validation function that merges adjacent text nodes together. 

Note: This functionality is already correctly implemented in slate-core so you don't need to put it in yourself!

```
/**
  * Merge adjacent text nodes.
  *
  * @type {Object}
  */
validateNode(node) {
  if (node.object != 'block' && node.object != 'inline') return

  const invalids = node.nodes
    .map((child, i) => {
      const next = node.nodes.get(i + 1)
      if (child.object != 'text') return
      if (!next || next.object != 'text') return
      return next
    })
    .filter(Boolean)

  if (!invalids.size) return

  return (change) => {
    // Reverse the list to handle consecutive merges, since the earlier nodes
    // will always exist after each merge.
    invalids.reverse().forEach((n) => {
      change.mergeNodeByKey(n.key)
    })
  }
}
```

There is actually a problem with this code. Because each `change` function call will cause nodes impacted by the mutation to be normalized, this can cause interruptions to carefully implemented sequences of `change` functions and may create performance problems or errors. The normalization logic in the above example will merge the last node in the invalids list together, but then it'll trigger another normalization and start over!

How can we deal with this? Well, normalization can be suppressed temporarily for multiple `change` function calls by using the `change.withMutations` function. `withMutations` accepts a function that takes a `change` object as a parameter, and executes the function while suppressing normalization. Once the function is done executing, the entire document is then normalized to pick up any unnnormalized transformations and ensure your document is in a normalized state.

The above validation function can then be written as below

```
/**
  * Merge adjacent text nodes.
  *
  * @type {Object}
  */
validateNode(node) {
  ... 
  return (change) => {
    change.withMutations((c) => {
      // Reverse the list to handle consecutive merges, since the earlier nodes
      // will always exist after each merge.
      invalids.reverse().forEach((n) => {
        change.mergeNodeByKey(n.key)
      })
    });
  }
}
```

Soreine:
There's something that this approach currently ignore, is that normalize operations were optimized depending on the changes made.

See for example this case, were [only the `parent` of the affected node is normalized](https://github.com/ianstormtaylor/slate/pull/1549/files#diff-e889bf99ac368f13306c6c00e1d897b3R91)

It would be great if the [final normalization](https://github.com/ianstormtaylor/slate/pull/1549/files#diff-0f25865aa4f7f8b2db06856c8eff1568R160) was only done were it needed to happen. Maybe changes should "request" normalizations. These normalization requests would be delayed when using `withMutations`. And at the end, we make the smallest normalization call needed to answer all these requests.

This is just optimization though. It can be added at a later stage. For now, the few uses of `withMutations` will be less than optimal.

Also, `withMutations` is not a correct name since this has nothing to do with immutability. `change.withDelayedNormalization` for example is a bit longer but clearer. We could as well talk about a "atomic" changes, or "transactions", to express that the encapsulated changes should happen together...

CameronAckermanSEL:
The name was suggested by ian.

docs/reference/slate/change.md
```diff
@@ -50,6 +50,33 @@ function onSomeEvent(event, change) {
...
+ */
+validateNode(node) {
+  if (node.object != 'document') return
+  const invalids = node.nodes.filter(n => n.object != 'block')
+  if (!invalids.size) return
+
+  return (change) => {
+    change.withMutations((c) => {
+      invalids.forEach((child) => {
+        change.removeNodeByKey(child.key)
```
ianstormtaylor: This should be `c.removeNodeByKey`

docs/guides/schemas.md
```diff
@@ -112,3 +112,67 @@ This validation defines a very specific (honestly, useless) behavior, where if a
 When you need this level of specificity, using the `validateNode` property of the editor or plugins is handy.
 
 However, only use it when you absolutely have to. And when you do, you need to be aware of its performance. `validateNode` will be called **every time the node changes**, so it should be as performant as possible. That's why the example above returns early, so that the smallest amount of work is done as possible each time it is called.
+
+## Multi-step Normalizations
+
+Some normalizations will require multiple `change` function calls in order to complete.
```
ianstormtaylor: Add after this sentence:

> But after calling the first change function, the resulting document will be normalized, changing it out from under you. This can cause intended behaviors.

packages/slate/src/utils/get-normalize-setting.js
```diff
@@ -0,0 +1,27 @@
...
+ * Obtains the normalization setting to use for the change based on the
+ * change object flags and the change operation options.
+ *
+ * @param {Object} flags - flags object for the current change object
+ * @param {Object} options - options for the change operation
+ * @return {bool} - the normalization configuration to use. Defaults to true
+ *   if flags or options does not have a normalization configuration
+ */
+
+function getNormalizeSetting(flags, options) {
```
ianstormtaylor: I think we should change this from being a util to being a method on `Change` objects. So the calling code can change from:

```js
const normalize = getNormalizeSetting(change.flags, options)
```

Into...

```js
const normalize = change.getFlag('normalize', options)
```

(Later we can even change `setOperationFlag` and `unsetOperationFlag` to just be `un/setFlag` to match the terser API too.)

packages/slate/src/utils/get-normalize-setting.js
```diff
@@ -0,0 +1,27 @@
...
+  let normalize = true
```
ianstormtaylor: I think this value should then be set as the default on `this.flags` in the `Change` objects? So that there isn't a hardcoded one here?

packages/slate/test/models/change/with-mutations-no-option-override.js
```diff
@@ -0,0 +1,69 @@
...
+
+export const input = (
+  <value>
+    <document>
+      <paragraph />
+    </document>
+  </value>
+)
+
+export const output = {
```
ianstormtaylor: Any specific reason we're using the JSON version of the output here? It might be nicer to write tests using the JSX version instead (like the `./test/changes` tests do) for terseness if we can.

packages/slate/src/utils/get-normalize-setting.js
CameronAckermanSEL: Great idea

docs/reference/slate/change.md
CameronAckermanSEL: Sure should, there was a similar mistake in the the other example too

packages/slate/test/models/change/with-mutations-no-option-override.js
CameronAckermanSEL: It was what was familiar, happy to change it to the other style

ianstormtaylor: @CameronAckermanSEL cool, yeah let's change it. It'll make adding future tests easier I think too.

packages/slate/src/utils/get-normalize-setting.js
CameronAckermanSEL: Hm, this one might require a little bit more careful examination (and might cause some complications with the lovely `getFlag` idea).

The default here comes from the `const { normalize = true } = options` which would set normalize to `true` if it was absent from `options`. 

The existing implementation of this function is set up carefully to have the following logic

1. Assume the normalize flag is true.
2. Check if normalize flag is set in operation flags. If so return that value. This is to allow flags to have the highest precedence for this normalization configuration.
3. Finally check if options has set normalize. If so, use that value only if normalize is not specified in flags.

This is important because we don't want folks doing something odd like setting `{ normalize: true }` in a change call and breaking the contract of `withoutNormalization`

CameronAckermanSEL: @ianstormtaylor  alright! I think I got it sorted out. Going to write some comments on the new getFlag method and start a discussion for one of the calls

packages/slate/src/changes/at-range.js
```diff
@@ -28,7 +28,7 @@ const Changes = {}
 Changes.addMarkAtRange = (change, range, mark, options = {}) => {
   if (range.isCollapsed) return
 
-  const { normalize = true } = options
+  const normalize = change.getFlag('normalize', { normalize: true, ...options })
```
CameronAckermanSEL: I'm not sure if there's a better way to write this, but it is important that the logic is equivalent here. We need to only use this default value of `true` if `normalize` is not in `flags`  and not in `options`. It's a little difficult to express it concisely

ianstormtaylor: If the `Change` model's `constructor` set `this.flags.normalize = true` then I think we could avoid this?

CameronAckermanSEL: Sorry, I noticed that comment before and I forgot to address it! ;_;

The reason why that won't work in this existing implementation is that all `{ normalize: false }` option objects will fail because of the way `getFlag` operates. In the current implementation the assumption is that the `normalize` flag will override the `normalize` option, because that allows `withoutNormalization` to globally turn off normalization regardless of what callers pass in options

Does that make sense?

ianstormtaylor: @CameronAckermanSEL Ah, I think the `options` should be able to override the `flags`. And then `withoutNormalization` would just be a quick shortcut for changing the flags default value for all the change methods inside the callback. But if a user really wanted to (for some reason) override with `options` then that's fine—although strange.

CameronAckermanSEL: Alrighty, that's not a problem. I'll go rework that and hopefully have something for you to look at shortly :)

CameronAckermanSEL: ok! @ianstormtaylor got the desired behavior of options overriding flags, and adjusted tests to make sure it works properly. Tests should also now look at the case where someone unsets the normalize flag to make sure nothing blows up and the behavior for `withoutNormalization` still works correctly

packages/slate/src/models/change.js
```diff
@@ -153,6 +177,21 @@ class Change {
...
+   * Get the `value` of the specified flag by its `key`. Optionally accepts an `options`
+   * object with override flags.
+   *
+   * @param {String} key
+   * @param {Object} options
+   * @return {Change}
+   */
+
+  getFlag(key, options = {}) {
+    return options[key] !== undefined ?
```
CameronAckermanSEL: now override the flag if the provided option object (if any) has the flag

packages/slate/src/models/change.js
```diff
@@ -48,7 +48,10 @@ class Change {
     const { value } = attrs
     this.value = value
     this.operations = new List()
-    this.flags = pick(attrs, ['merge', 'save'])
```
CameronAckermanSEL: This seemed like the best way to have a default set here

packages/slate/test/models/change/without-normalization-normalize-flag-true.js
```diff
@@ -0,0 +1,49 @@
+/** @jsx h */
+
+import h from '../../helpers/h'
+
+export const normalize = true
```
ianstormtaylor: I think exporting the `normalize` here is confusing, since it doesn't appear to be used directly? Instead just the `flags` exports here would suffice I think.

CameronAckermanSEL: It gets used in the assert that verifies that the normalize flag got reset to the expected value

CameronAckermanSEL: Some tests deliberately do not specify a normalize flag in the settings to verify that the correct default of `true` is applied in the constructor and the normalize flag is `true` at the end.

ianstormtaylor: I think we need to remove it, it's too confusing. If we really need to assert that it got properly reset, we can import `assert` into this test. But I seems like something that isn't necessary? If we want that level of testing we can add unit tests later for the change methods I think.

CameronAckermanSEL: I'm ok with that :)

ianstormtaylor:
Thanks @CameronAckermanSEL! Looks good to me.

Issue #1547, opened 2018-01-22
Feature request - Change method: setTextByKey


bergalv:
A Feature request

There is a missing and needed Change method: "setTextByKey"
Currently there is no straight way to replace text in a Node.
I can append text to Node (insertTextByKey) or remove text from Node (removeTextByKey). However, for replacing the text, one can not merely remove the text and re-insert it. This is because removing the text of the Node will cause for its deletion (empty Nodes gets deleted automatically).

zhujinxuan:
How about `deleteAtRange(Range.create().moveToRangeOf(text.key), {normalize: false})).insertTextAtRange(Range.create().collapseToStartOf(text.key), "some texts")`

zGrav:
Hey @bergalv , I believe I built what you wanted 👍

ianstormtaylor:
I'd be open to a `replaceTextByKey` change method if someone wants to add it, it should just call into the existing `removeTextByKey` and then `addTextByKey`, but it does seem like a nice convenience.

benjycui:
> How about deleteAtRange(Range.create().moveToRangeOf(text.key), {normalize: false})).insertTextAtRange(Range.create().collapseToStartOf(text.key), "some texts")

The problem is, when users try to undo, they need to undo twice to revert back to previous text.

zhujinxuan:
@benjycui Yes, I forgot that....  It should be `removeTextByKey(....).insertTextByKey(...) `
but I think it would be much better to have an `replaceTextByKey`. I will submit a PR about that.

ianstormtaylor:
Actually, I think having both `replaceTextByKey` (which takes a `offset, length, text`) and `setTextByKey` (which takes just a `key`) would both be useful. Open to PR's that implement either of them.

zhujinxuan:
@ianstormtaylor How about another thing:
for insertTextAtRange and insertText, if range is expanded but startKey and endKey is the same, use replaceTextByKey and do not use snapshot and deleteAtRange

zhujinxuan:
@ianstormtaylor

About marks of setTextByKey, there are two choices.  Which one would you prefer?
   1. `setTextByKey(key: string , text: string|Text) `
   2. `setTextByKey(key: string, text: string, marks?:  Set<Marks>)`

Another question is shall we inherit `activeMarks` in repalceTextByKey and insertTextByKey?

ianstormtaylor:
[...]

PR #1550, opened 2018-01-22
add setTextByKey function/test

zGrav:
Should add functionality required for #1547 

@ianstormtaylor , this is my first time implementing a 'full functionality' into slate, sorry if this is wrong way to approach it, would appreciate some pointers 👍

### DIFF
--- a/packages/slate/src/changes/by-key.js
+++ b/packages/slate/src/changes/by-key.js
@@ -160,6 +160,34 @@ Changes.insertTextByKey = (change, key, offset, text, marks, options = {}) => {
 }
 
 /**
+ * Sets `text` in node by `key`.
+ *
+ * @param {Change} change
+ * @param {String} key
+ * @param {String} text
+ * @param {Object} options
+ *   @property {Boolean} normalize
+ */
+
+Changes.setTextByKey = (change, key, text, options = {}) => {
+  const { normalize = true } = options
+  const { value } = change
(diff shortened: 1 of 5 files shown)


zhujinxuan:
How to undo the operation?

justinweiss:
I would be _very_ hesitant to add new primitive operations into Slate -- each one has to support undo / inverse, and will likely add a lot of complexity when Slate supports collaborative editing. 

Is this something that could be handled using `insert_text` / `remove_text` instead? It wouldn't be atomic, but there are a lot of change operations (like `wrapBlockByKey`) that are built out of several primitive operations.

CameronAckermanSEL:
IMO this will be a lot harder to implement an undo because the current `invert` feature doesn't have completely solid approaches for undoing changes where the original state is not somehow cached in the operation. 

slate-auto-replace has some code that shows how to delete and @zhujinxuan posted (what I'm really sure is) a functional code snippet that achieves this.

https://github.com/ianstormtaylor/slate-plugins/blob/master/packages/slate-auto-replace/src/index.js#L49

@zGrav  thanks for taking the initiative to provide this feature

zGrav:
@justinweiss since @zhujinxuan has given a solution that might work in the issue itself and due to the complexity of adding primitive ops into Slate, I'll close this issue as of now.

I appreciate the look anyway!

PR #1557, opened 2018-01-26
Convert to babel-preset-env and upgrade build and linting packages.

schneidmaster:
Based on #936, I wanted to start by converting to babel-preset-env and establishing a list of supported browsers to help keep the build slim and stay up to date.

The [browserslist](https://github.com/schneidmaster/slate/blob/babel-preset-env/browserslist) I established is mostly just my own sense of what we might reasonably support: the last 2 versions of major browsers, IE 11, and anything with >1% global usage share.

While I was at it, I upgraded the rest of our babel packages for consistency. This in turn required me to upgrade eslint to work out some inconsistencies with the babel-eslint parser and eslint 3. I also started seeing some complaints in the test environment from `babel-polyfill`, so I converted the test build to use `babel-plugin-transform-runtime` and `babel-runtime`. Finally, I fixed some linting errors that cropped up -- not sure why they weren't happening in master. (I didn't want to rewrite a bunch of the examples so I ignored the errors about [react/jsx-no-bind](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md)).

Let me know if you have thoughts or feedback! 🙂


browserslist
```diff
@@ -0,0 +1,6 @@
+last 2 Chrome versions
+last 2 Firefox versions
+last 2 Safari versions
+last 2 Edge versions
+IE 11
+>1%
```
ianstormtaylor: Do we know what `>1%` globally translates too? I think we should get rid of this if not, because it seems like something that can change without us formally committing to it?

browserslist
```diff
@@ -0,0 +1,6 @@
+last 2 Chrome versions
+last 2 Firefox versions
+last 2 Safari versions
+last 2 Edge versions
+IE 11
```
ianstormtaylor: Let's remove IE 11 here, because I don't want to commit to formally supporting it with the exported files. We've already been using things like `Array.includes` which I think are unsupported?

schneidmaster: That's where this gets tricky. Currently it is possible to use Slate in IE11 by including `babel-polyfill` to account for unsupported primitives -- any ES6 syntax is fully converted to ES5. And also, Slate seems to generally work in IE11 (and there's a small handful of open issues which might get worked at some point to fix the bugs that do exist). If we remove IE 11 here then Slate will almost certainly break completely in IE since we wouldn't be transpiling ES6 syntax that IE doesn't support. So I strongly feel that we should keep it in to allow continued development there.

Perhaps it would be helpful to add a comment to the `browserslist` to indicate that this is not a formal list of browsers we officially support but rather a broader set of browsers that we're willing to transpile ES6 features for? (Kind of verbose but hopefully you get what I mean.)

browserslist
schneidmaster: You can preview `>1%` (or any other `browserslist` config) [here](http://browserl.ist/?q=%3E1%25). Without it, I would have to add verbose entries for browsers like iOS Safari. I can move in that direction though if you'd prefer it to be more explicit.

ianstormtaylor:
Thanks @schneidmaster! I agree with you on both of those points, looks like a great setup.

PR #1568, opened 2018-01-28
Rollup

### LINKED ISSUE #1566: switch to rollup for bundling packages
This should allow us to reduce the bundle size a decent amount, as well as the startup time, since it will have less closures I believe.

schneidmaster:
I saw #1566 open up and figured I'd take a crack at it as a weekend project (have been meaning to get into rollup). This is nearly complete; all of the builds work correctly and for each package output a CommonJS module, an ES module (newly possible with rollup and useful for tree-shaking etc. with webpack), a browser UMD build, and a minified browser UMD build. As a consequence of this change, we would no longer ship the complete babelified `lib` directory, just the built output.

There are two outstanding items at the moment: 1) ~need to test the browser UMD builds and make sure they don't break anywhere (i.e. if you're using unpkg and include react-dom, immutable, react, then slate and slate-react, are there any problems)~; and 2) the top-level `watch` does not work at the moment, because for some reason `lerna run --parallel watch` will build the source once in each package and then exit, rather than keeping the process alive in each package to watch for changes (as would happen if you ran `yarn watch` from the package directory). It's not really clear to me why this is happening, might need to dig some more into lerna.

Other than configuration, I needed a few changes to the actual examples/source code:

* `examples/check-lists/value.json` was not valid JSON (it had trailing commas)
* `examples/images/index.js` was not updated for the new extracted schema violations. It also used an `is-image` npm library that requires the core nodejs `path` module -- I'm not sure how it ever worked in the browser (does browserify do some sort of magic there?) but it's performing a fairly simple task (checking if a URL has an image extension) so I just rewrote it with a small helper.
* `packages/slate/src/index.js` did not export the `Change` model (even though `packages/slate-prop-types/src/index.js` imports it)
* `packages/slate/src/models/value.js` oddly had `const Change = require('./change').default` about halfway down the file, rather than having it as an import at the top

Feel free to pull it down and check it out, let me know if you have any thoughts or feedback 👍 

Edit: I tested out the UMD builds and they seem to work fine. The UMD doc was a good bit out of date (it says to install `slate@0.11.12`!) so I updated it. Still working on the lerna/rollup watch issue; I opened up https://github.com/rollup/rollup/issues/1919 to see if I'm missing something or if there might be a rollup bug.

**Bundle size savings**

Comparing [current size badges](https://github.com/ianstormtaylor/slate/tree/master/packages) vs. the size of gzipping the package's UMD `min.js` on my local machine.

| package | browserify/master | rollup | % reduction |
| --- | --- | --- | --- |
| **slate** | **49.3kB** | **36.3kB** | **26.4%** |
| slate-base64-serializer | 731B | 433B | 40.8% |
| slate-html-serializer | 2.72kB | 2.34kB | 13.9% |
| slate-hyperscript | 2.39kB | 1.84kB | 23.0% |
| slate-plain-serializer | 931B | 675B | 27.5%
| slate-prop-types | 1.08kB | 811B | 24.9% |
| **slate-react (*)** | **50.1kB** | **20.4kB** | **59.3%** |
| slate-schema-violations | 731B | 423B | 42.7% |
| slate-simulator | 1.07kB | 752B | 29.7% |

**(*)** In master, we are (presumably accidentally) bundling immutable.js into the slate-react UMD build, so part of the enormous savings is in fixing that. The minified build in master would weigh 31.1kB if that mistake were fixed, meaning that rollup would then provide a more believable (but still great) 34.4% savings. 

Closes #1566


docs/walkthroughs/using-the-bundled-source.md
```diff
@@ -17,8 +17,9 @@ And then look in the `node_modules` folder for the bundled `slate.js` file:
 node_modules/
   slate/
     dist/
-      slate.js
-      slate.min.js
+      umd/
```
ianstormtaylor: Let's change this to be not `dist/umd/` but instead just `umd/`, to match other React libraries. (Unless this is just a typo!)

package.json
```diff
@@ -4,21 +4,19 @@
   "workspaces": [
     "packages/*"
   ],
+  "browser": "./examples/build.dev.js",
+  "browserMin": "./examples/build.prod.js",
```
ianstormtaylor: What are these two for? I thought that `browser` was a field for including a browserified version of something. But this package is private, so why do we need it?

package.json
```diff
@@ -69,19 +73,17 @@
     "benchmark:save": "mkdir -p ./tmp && babel-node ./node_modules/.bin/_matcha --reporter ./support/benchmark-reporter ./packages/*/benchmark/index.js > ./tmp/benchmark-baseline.json",
     "bootstrap": "lerna bootstrap && yarn run build",
     "build": "yarn run build:packages",
-    "build:examples": "yarn run build:examples:dev && yarn run build:examples:prod",
-    "build:examples:dev": "browserify --debug --transform babelify ./examples/index.js > ./examples/build.dev.js",
-    "build:examples:prod": "NODE_ENV=production browserify --global-transform envify --global-transform uglifyify --transform babelify ./examples/index.js > ./examples/build.prod.js",
+    "build:examples": "rollup -c",
```
ianstormtaylor: Can you use the long-form CLI flags here?

package.json
```diff
@@ -69,19 +73,17 @@
...
     "gh-pages": "yarn run build && yarn run build:examples && gh-pages --dist ./examples",
-    "lint": "eslint packages/*/src packages/*/test examples/*/*.js examples/dev/*/*.js",
+    "lint": "eslint packages/*/src packages/*/test packages/*/rollup.config.js examples/*/*.js examples/dev/*/*.js",
     "open": "open http://localhost:8080/dev.html",
     "release": "yarn run test && yarn run lint && lerna publish && yarn run gh-pages",
     "start": "http-server ./examples",
     "test": "BABEL_ENV=test mocha --require babel-core/register ./packages/*/test/index.js",
     "watch": "npm-run-all --parallel --print-label watch:examples watch:packages start",
-    "watch:examples": "watchify --debug --transform babelify ./examples/index.js -o ./examples/build.dev.js -v",
+    "watch:examples": "rollup -c -w",
```
ianstormtaylor: Can you use the long-form CLI flags here?

packages/slate-base64-serializer/package.json
```diff
@@ -4,34 +4,37 @@
...
   "repository": "git://github.com/ianstormtaylor/slate.git",
-  "main": "./lib/index.js",
+  "main": "./dist/cjs/slate-base64-serializer.js",
+  "module": "./dist/es/slate-base64-serializer.js",
+  "browser": "./dist/umd/slate-base64-serializer.js",
+  "browserMin": "./dist/umd/slate-base64-serializer.min.js",
   "files": [
-    "dist/",
-    "lib/"
+    "dist/cjs/slate-base64-serializer.js",
```
ianstormtaylor: Why not just `"dist/"` here instead of having to list them all out?

packages/slate-base64-serializer/package.json
```diff
@@ -4,34 +4,37 @@
...
-    "build": "babel --out-dir ./lib ./src",
-    "build:max": "mkdir -p ./dist && NODE_ENV=production browserify ./src/index.js --transform babelify --transform envify --transform [ browserify-global-shim --global ] --standalone SlateBase64Serializer > ./dist/slate-base64-serializer.js",
-    "build:min": "mkdir -p ./dist && NODE_ENV=production browserify ./src/index.js --transform babelify --transform envify --transform [ browserify-global-shim --global ] --transform uglifyify --standalone SlateBase64Serializer | uglifyjs > ./dist/slate-base64-serializer.min.js",
-    "clean": "rm -rf ./dist ./lib ./node_modules",
-    "prepublish": "yarn run build && yarn run build:max && yarn run build:min",
-    "watch": "babel --watch --out-dir ./lib ./src --source-maps inline"
-  },
-  "browserify-global-shim": {
-    "slate": "Slate"
+    "build": "rollup -c",
```
ianstormtaylor: Can you use the long-form CLI flags here, and everywhere else/

packages/slate-hyperscript/rollup.config.js
```diff
@@ -0,0 +1,67 @@
...
+umdConfigMin.plugins = umdConfig.plugins.slice(0).concat(uglify())
+
+// CommonJS (for Node) and ES module (for bundlers) build.
+const moduleConfig = {
+  input: 'src/index.js',
+  external: [
+    'is-empty',
+    'is-plain-object',
+    'slate',
+    'slate-dev-logger',
```
ianstormtaylor: Why are these ones other than `slate` external? They aren't peer dependencies?

packages/slate-simulator/rollup.config.js
```diff
@@ -0,0 +1,65 @@
+import resolve from 'rollup-plugin-node-resolve'
+import commonjs from 'rollup-plugin-commonjs'
+import babel from 'rollup-plugin-babel'
+import uglify from 'rollup-plugin-uglify'
+import sourcemaps from 'rollup-plugin-sourcemaps'
+import pkg from './package.json'
+
+// UMD build for browsers
+const umdConfig = {
```
ianstormtaylor: Is there a better way for us to share all of these configuration logic somewhere, instead of having to duplicate it in each new monorepo package here? It seems like it could be a few util functions in a top-level helper I think.

docs/walkthroughs/using-the-bundled-source.md
schneidmaster: Hmm, currently the built source in the slate library is at `/dist/slate.js` and `/dist/slate.min.js`, so for the new output I used `/dist/es/slate.js`, `/dist.cjs/slate.js`, and `/dist/umd/slate.js` (slash `/dist/umd/slate.min.js`). So you want to drop the `/dist` and have three top-level output folders (`/es`, `/cjs`, and `/umd`)? (Doesn't matter a ton to me, I just thought it was somewhat customary to use a `/dist` folder, in addition to being a bit closer to what we currently ship)

package.json
schneidmaster: These two are [imported and used](https://github.com/schneidmaster/slate/blob/rollup/rollup.config.js#L10) by the top-level `rollup.config.js` when building the examples. (I could hardcode it inside the rollup config, I was just keeping a consistent pattern with all the sub packages.)

packages/slate-base64-serializer/package.json
schneidmaster: Currently the way I have it configured results in sourcemaps also being put in the `dist/` folder when in development mode, so I listed it out to avoid accidentally shipping sourcemaps along with the package

packages/slate-hyperscript/rollup.config.js
schneidmaster: In this context, `external` tells rollup that the package is an external npm dependency rather than something that it is supposed to bundle into your built output file.

packages/slate-simulator/rollup.config.js
schneidmaster: I looked at this a bit and I'm not sure. The problem is that despite all the similarities, there's a decent amount of variance for the different config files:

* the output name
* the UMD globals
* the dependencies to be considered `external` both for UMD and CJS/ES
* the plugins (a few of them need `alias` or `replace`, for example, but not all)

I didn't see a great way to condense the configuration logic. I think the abstraction is worse than what I've got here, because this is clearly comprehendible as normal rollup options, whereas I'd have weird ad-hoc exports/configuration to try to extract out shared bits.

packages/slate-hyperscript/rollup.config.js
ianstormtaylor: Is there really no way other than manually managing each dependency here? This seems like a big downside to using Rollup?

schneidmaster: Hmm, I can use [rollup-plugin-auto-external](https://www.npmjs.com/package/rollup-plugin-auto-external), that'll reduce this bit for the CJS/ES sections. It's still probably needed for the UMD builds though because it's not so cut-and-dry (we actually WANT to bundle in some of the smaller packages).

schneidmaster:
Hey @ianstormtaylor -- I was trying to work around the lerna/rollup issue and I developed an alternate approach that uses a single `rollup.config.js` in the project root. It all works fine (no watch bug) and also reduces the configuration a bit further. The tradeoff is that it's no longer possible to run `yarn build` or `yarn watch` from within an individual package directory, only at the top level of the whole project. Check out the [diff here](https://github.com/ianstormtaylor/slate/compare/master...schneidmaster:rollup-centralized?expand=1) and let me know what you think, if you like this approach then I'll merge it into my branch/PR here.

ianstormtaylor:
@schneidmaster that seems like a decent tradeoff to me—maintaining those scripts was kinda annoying anyways 😄  

Once you merge it in I'll go through again and make comments. Anything you can do to clean up the `rollup.config` files to make them super easy to read and consistent with decisions elsewhere in the codebase would be awesome!

schneidmaster:
Cool, merged it in 👍will keep an eye out for additional comments!

docs/walkthroughs/using-the-bundled-source.md
ianstormtaylor: I agree, I much prefer `/dist` to `/umd`, I was just thinking of following React's lead. I think `react`, `react-dom` and `react-router` now are all using `/umd` for some reason. That said, I think for now let's go with:

```
dist/
  slate.js
  slate.min.js
lib/
  slate.js
  slate.es.js
```

Where `/dist` holds the UMD files, and `/lib` holds the CJS/ES6 files. This way I believe existing links to Slate's UMD files will remain unbroken.

package.json
ianstormtaylor: I think the `browser` has special meaning as a `package.json` field? I believe it's used in Browserify and others as the browser-preferred variant for packages. We might want to call it `umd` and `umdMin` instead I think to avoid this breaking stuff?

packages/slate-base64-serializer/package.json
```diff
@@ -4,33 +4,30 @@
...
+  "module": "dist/es/slate-base64-serializer.js",
+  "browser": "dist/umd/slate-base64-serializer.js",
+  "browserMin": "dist/umd/slate-base64-serializer.min.js",
   "files": [
-    "dist/",
-    "lib/"
+    "dist/cjs/slate-base64-serializer.js",
+    "dist/es/slate-base64-serializer.js",
+    "dist/umd/slate-base64-serializer.js",
+    "dist/umd/slate-base64-serializer.min.js"
```
ianstormtaylor: Same goes for all of these files in the other packages too, to match the `dist/` and `lib/` structure.

packages/slate-base64-serializer/package.json
```diff
@@ -4,33 +4,30 @@
...
   "dependencies": {
     "isomorphic-base64": "^1.0.2"
   },
   "scripts": {
-    "build": "babel --out-dir ./lib ./src",
-    "build:max": "mkdir -p ./dist && NODE_ENV=production browserify ./src/index.js --transform babelify --transform envify --transform [ browserify-global-shim --global ] --standalone SlateBase64Serializer > ./dist/slate-base64-serializer.js",
-    "build:min": "mkdir -p ./dist && NODE_ENV=production browserify ./src/index.js --transform babelify --transform envify --transform [ browserify-global-shim --global ] --transform uglifyify --standalone SlateBase64Serializer | uglifyjs > ./dist/slate-base64-serializer.min.js",
-    "clean": "rm -rf ./dist ./lib ./node_modules",
-    "prepublish": "yarn run build && yarn run build:max && yarn run build:min",
-    "watch": "babel --watch --out-dir ./lib ./src --source-maps inline"
```
ianstormtaylor: I love that we get to eliminate all of this inside these modules!

packages/slate-plain-serializer/package.json
```diff
@@ -17,24 +22,16 @@
...
-    "clean": "rm -rf ./dist ./lib ./node_modules",
-    "prepublish": "yarn run build && yarn run build:max && yarn run build:min",
-    "watch": "babel --watch --out-dir ./lib ./src --source-maps inline"
+    "clean": "rm -rf ./dist ./node_modules"
   },
-  "browserify-global-shim": {
+  "umdGlobals": {
     "immutable": "Immutable",
-    "slate": "Slate"
+    "slate": "slate"
```
ianstormtaylor: Should be capital `"Slate"` I think?

packages/slate-prop-types/package.json
```diff
@@ -17,23 +22,15 @@
...
-    "clean": "rm -rf ./dist ./lib ./node_modules",
-    "prepublish": "yarn run build && yarn run build:max && yarn run build:min",
-    "watch": "babel --watch --out-dir ./lib ./src --source-maps inline"
+    "clean": "rm -rf ./dist ./node_modules"
   },
-  "browserify-global-shim": {
+  "umdGlobals": {
     "immutable": "Immutable",
-    "slate": "Slate"
+    "slate": "slate"
```
ianstormtaylor: Also should be capitalized?

packages/slate-react/package.json
```diff
@@ -32,27 +37,19 @@
...
+    "clean": "rm -rf ./dist ./node_modules"
   },
-  "browserify-global-shim": {
+  "umdGlobals": {
+    "immutable": "Immutable",
     "react": "React",
     "react-dom": "ReactDOM",
-    "react-dom/server": "ReactDOMServer",
-    "slate": "Slate"
+    "slate": "slate"
```
ianstormtaylor: Also here?

packages/slate-hyperscript/package.json
```diff
@@ -18,21 +23,13 @@
...
-    "build:max": "mkdir -p ./dist && NODE_ENV=production browserify ./src/index.js --transform babelify --transform envify --transform [ browserify-global-shim --global ] --standalone SlateHyperscript > ./dist/slate-hyperscript.js",
-    "build:min": "mkdir -p ./dist && NODE_ENV=production browserify ./src/index.js --transform babelify --transform envify --transform [ browserify-global-shim --global ] --transform uglifyify --standalone SlateHyperscript | uglifyjs > ./dist/slate-hyperscript.min.js",
-    "clean": "rm -rf ./dist ./lib ./node_modules",
-    "prepublish": "yarn run build && yarn run build:max && yarn run build:min",
-    "watch": "babel --watch --out-dir ./lib ./src --source-maps inline"
+    "clean": "rm -rf ./dist ./node_modules"
   },
-  "browserify-global-shim": {
+  "umdGlobals": {
     "slate": "Slate"
```
ianstormtaylor: Or should this one not be capitalized?

packages/slate/src/models/value.js
```diff
@@ -4,6 +4,7 @@ import logger from 'slate-dev-logger'
 import { Record, Set, List, Map } from 'immutable'
 
 import MODEL_TYPES from '../constants/model-types'
+import Change from './change'
```
ianstormtaylor: Nice solution to that recursive import problem 😄

rollup.config.js
```diff
@@ -0,0 +1,26 @@
...
+import pkgSlateHtmlSerializer from './packages/slate-html-serializer/package.json'
+import pkgSlateHyperscript from './packages/slate-hyperscript/package.json'
+import pkgSlatePlainSerializer from './packages/slate-plain-serializer/package.json'
+import pkgSlatePropTypes from './packages/slate-prop-types/package.json'
+import pkgSlateReact from './packages/slate-react/package.json'
+import pkgSlateSchemaViolations from './packages/slate-schema-violations/package.json'
+import pkgSlateSimulator from './packages/slate-simulator/package.json'
+
+export default [
+  ...factory(pkgSlate),
```
ianstormtaylor: Can you remove `pkg` from all of these imports, just to make it slightly more readable at a glance? Since there's no worry about overlap.

rollup.config.js
```diff
@@ -0,0 +1,26 @@
...
+  ...factory(pkgSlateDevLogger),
+  ...factory(pkgSlateHtmlSerializer),
+  ...factory(pkgSlateHyperscript),
+  ...factory(pkgSlatePlainSerializer),
+  ...factory(pkgSlatePropTypes),
+  ...factory(pkgSlateReact),
+  ...factory(pkgSlateSchemaViolations),
+  ...factory(pkgSlateSimulator),
+  ...examplesConfig,
+]
```
ianstormtaylor: This file is great!

support/rollup/examples.js
```diff
@@ -0,0 +1,66 @@
...
+const configurations = []
+
+const umdConfig = {
+  input: 'examples/index.js',
+  output: {
+    file: pkg.browser,
+    name: 'slate-examples',
+    format: 'umd',
+    exports: 'named',
+    sourcemap: process.env.ROLLUP_WATCH,
```
ianstormtaylor: Can you collect all `env` variables near the top of the file somewhere, so we can reference them just as `ROLLUP_WATCH`, etc. from here on out?

Should this be `ROLLUP_WATCH`, or should it actually be `NODE_ENV != 'production'`? I think using Rollup's watch feels slightly incorrect?

support/rollup/examples.js
```diff
@@ -0,0 +1,66 @@
...
+  },
+  plugins: [
+    alias({
+      debug: path.resolve(__dirname, 'node_modules', 'debug', 'src', 'browser'),
```
ianstormtaylor: Can you use `'debug'` (quoted) to keep them the same?

support/rollup/examples.js
```diff
@@ -0,0 +1,66 @@
...
+  input: 'examples/index.js',
+  output: {
+    file: pkg.browser,
+    name: 'slate-examples',
+    format: 'umd',
+    exports: 'named',
+    sourcemap: process.env.ROLLUP_WATCH,
+  },
+  plugins: [
+    alias({
```
ianstormtaylor: Can you add a comment above this `alias` usage to explain what it does? It's not clear from looking at this why just these two modules are aliased?

support/rollup/examples.js
ianstormtaylor: I don't think the `, , ,` is necessary? Can't you just do `path.resolve(__dirname, 'node_modules/debug/src/browser')` instead? It would be a lot more readable.

support/rollup/examples.js
```diff
@@ -0,0 +1,66 @@
...
+      debug: path.resolve(__dirname, 'node_modules', 'debug', 'src', 'browser'),
+      'react-dom/server': path.resolve(__dirname, 'node_modules', 'react-dom', 'cjs', 'react-dom-server.browser.production.min'),
```
ianstormtaylor: Same here to eliminate `, , ,`

support/rollup/examples.js
```diff
@@ -0,0 +1,66 @@
...
+    }),
+    resolve(),
```
ianstormtaylor: I think actually it would be nice to have a comment on top of each of the Rollup plugins quickly summarizing what it does / why it's needed. It's hard to come back to these kinds of files later and understand them.

support/rollup/examples.js
```diff
@@ -0,0 +1,66 @@
...
+    commonjs({
+      exclude: ['examples/**'],
+      namedExports: {
+        esrever: ['reverse'],
```
ianstormtaylor: Quote `'esrever'` here too for consistency.

support/rollup/examples.js
```diff
@@ -0,0 +1,66 @@
...
+    include: [
+      'examples/**',
+      'packages/*/dist/**',
+    ],
+  },
+}
+
+configurations.push(umdConfig)
+
+if (process.env.ROLLUP_WATCH) {
```
ianstormtaylor: Yeah it's unclear why this is dependent on `ROLLUP_WATCH`?

support/rollup/examples.js
```diff
@@ -0,0 +1,66 @@
...
+  umdConfig.plugins.push(sourcemaps())
+} else {
+  const umdConfigMin = Object.assign({}, umdConfig)
```
ianstormtaylor: Can you use `lodash/cloneDeep` here instead of needing to continually ensure that things aren't mutated at each level?

support/rollup/examples.js
```diff
@@ -0,0 +1,66 @@
...
+  umdConfigMin.output = Object.assign({}, umdConfig.output, { file: pkg.browserMin })
+  const prodReplace = replace({ 'process.env.NODE_ENV': JSON.stringify('production') })
+  umdConfigMin.plugins = umdConfig.plugins.slice(0).concat(prodReplace, uglify())
```
ianstormtaylor: Can you change this to not have `prodReplace` as an interim variable, to match the other usages?

support/rollup/factory.js
```diff
@@ -0,0 +1,99 @@
...
+import replace from 'rollup-plugin-replace'
+import uglify from 'rollup-plugin-uglify'
+import sourcemaps from 'rollup-plugin-sourcemaps'
+
+export default (pkg) => {
+  // UMD build for browsers
+  const umdConfig = {
+    input: `packages/${pkg.name}/src/index.js`,
+    output: {
+      file: `packages/${pkg.name}/${pkg.browser}`,
```
ianstormtaylor: Same concept here, can you store all of the `pkg` props as interim variables, so that we don't need to keep repeating `pkg.`?

support/rollup/factory.js
```diff
@@ -0,0 +1,99 @@
+import path from 'path'
```
ianstormtaylor: Same stuff from `examples.js` for this file.

support/rollup/factory.js
```diff
@@ -0,0 +1,99 @@
...
+        exports: 'named',
+        sourcemap: process.env.ROLLUP_WATCH,
+      },
+      {
+        file: `packages/${pkg.name}/${pkg.module}`,
+        format: 'es',
+        sourcemap: process.env.ROLLUP_WATCH,
+      },
+    ],
+    external: id => !!dependencies.find(dep => dep === id || id.startsWith(`${dep}/`)),
```
ianstormtaylor: What is this doing? I think it needs a comment.

package.json
schneidmaster: Well in the child packages I think `"browser"` is appropriate (because the fields do point to the UMD/browser-preferred variant). And I don't think it matters a ton in the top-level package because it's private, so I opted for consistency with the child packages. I don't really care though, I can change it if you want 🙂

packages/slate-plain-serializer/package.json
schneidmaster: Yep they should all be capitalized, I'll fix

support/rollup/examples.js
schneidmaster: Yeah, kinda six one/half a dozen the other (because we only build non-production when also watching), but I agree yours is probably a little more clear

package.json
ianstormtaylor: I think `browser` isn't supposed to be UMD though, I think it is usually just an alternate, browser-compatible CommonJS version of the code.

schneidmaster: Oh fair enough, I guess I was under the wrong impression of what `browser` is for

ianstormtaylor:
Looks great! Thank you @schneidmaster for taking the time to make this happen.

ianstormtaylor:
@schneidmaster what build times were you seeing for this? I'm seeing `30s` build times for and `15s` watch times using Rollup. It's feeling really slow so I'm wondering if I'm doing something wrong.

schneidmaster:
@ianstormtaylor The development build (i.e. yarn start) takes about 15s initial. Rebuilds take 50-500ms for the package itself and about 8s for the full examples.dev.js. I have a decently spec'd MBP so maybe the difference is in processor? What times did you see previously with browserify?

ianstormtaylor:
@schneidmaster okay that seems consistent with me as well—the examples are the slow part. I feel like I used to see almost instantaneous rebuilds of the examples with `watchify` because it was caching stuff. Not the end of the world though I guess.

ianstormtaylor:
@schneidmaster I think we might want to end up using Webpack for bundling the examples, from my quick research it sounds like Rollup doesn't handle incremental builds well.

schneidmaster:
@ianstormtaylor Yeah makes sense. I was trying to keep to one packaging library but at the end of the day rollup is optimized for libraries so having webpack would make for a nicer dev experience. We could also likely take advantage of hot reloading in the examples.

ianstormtaylor:
@schneidmaster totally, I appreciate the aiming for simplicity!

Issue #1518, opened 2018-01-10
Shift-select with the arrow keys fails when selecting toward an inline void node


bengotow:
Hey folks! This one's a bit odd. GIF below! In the GIF, I am holding down the shift key the whole time. (I haven't attached a JSFiddle because it's reproduced on http://slatejs.org/#/emojis.)

![jan-10-2018 14-18-04](https://user-images.githubusercontent.com/1037212/34798479-1bf4221e-f611-11e7-96be-f42451887560.gif)

Observations:
- Happens in both Firefox Beta and stable Chrome.
- You can shift-select with the arrow keys *away* from an inline void node. Eg: "shift right arrow" past the last void node, or "shift left arrow" before the first void node works. 
- You cannot shift-select with the arrow keys *toward* an inline void node.

CameronAckermanSEL:
Might be similar to https://github.com/ianstormtaylor/slate/issues/1502

ms88privat:
Also related?
https://github.com/ianstormtaylor/slate/pull/913

conorcussell:
This is happening because [`nextText`](https://github.com/ianstormtaylor/slate/blob/44eccca60a60908c88f022b1303452fc8467fde9/packages/slate-react/src/plugins/after.js#L444) is the zero-width text node slate inserts. As a result, the condition is true and we collapse the selection forward instead of extending it. 

I've made a PR which deals with this issue, I still need to check it deals with the other cases and doesn't cause any regressions.

PR #1570, opened 2018-01-29
Fix selecting with shift+(left|right) key

conorcussell:
Fixes https://github.com/ianstormtaylor/slate/issues/1518

I still need to verify it fixes both the issues that may be linked, and check it doesn't cause any regressions. Will report back!

slate-selection

### DIFF
--- a/packages/slate-react/src/constants/hotkeys.js
+++ b/packages/slate-react/src/constants/hotkeys.js
@@ -48,8 +48,10 @@ const DELETE_WORD_FORWARD_PC = isKeyHotkey('ctrl+delete')
 const DELETE_WORD_BACKWARD = e => IS_APPLE ? DELETE_WORD_BACKWARD_MAC(e) : DELETE_WORD_BACKWARD_PC(e)
 const DELETE_WORD_FORWARD = e => IS_APPLE ? DELETE_WORD_FORWARD_MAC(e) : DELETE_WORD_FORWARD_PC(e)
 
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
@conorcussell can you explain what the PR is doing? I'm not sure I follow the comments in there.

(Also can you change the comments to be `COMPAT:` and be sentence case too? Thanks.)

conorcussell:
Hey @ianstormtaylor I've realised that this PR doesn't actually make much sense... apologies.

https://github.com/ianstormtaylor/slate/issues/1518 is caused because we are collapsing instead of extending.

This happens because `HOTKEYS.COLLAPSE_CHAR_FORWARD` and `HOTKEYS.EXTEND_CHAR_FORWARD` are both true when selecting with `shift+right` and as a result when selecting towards void nodes we are collapsing rather than extending. https://github.com/ianstormtaylor/slate/blob/28450334a7e10bafddedd3dbf8839071ffbaacb6/packages/slate-react/src/plugins/after.js#L442-L449

Should we actually be changing https://github.com/ianstormtaylor/slate/blob/28450334a7e10bafddedd3dbf8839071ffbaacb6/packages/slate-react/src/constants/hotkeys.js#L51 
to check whether `shift` is currently pressed so we don't execute the `COLLAPSE_CHAR_FORWARD` block?

ianstormtaylor:
Great call @conorcussell, thanks!

PR #1574, opened 2018-01-30
Edge browser selection edge case

kellyjosephprice:
The Edge browser (41) has a weird edge case when selecting. If you double click to select the last word in a block, and it's the last text in the block, it sets the `endContainer` of the selection range to the surrounding node. The `selection-is-backward` lib doesn't handle cases where node's are contained. 

The current effect is that `findRange` get's `isBackward` wrong, and the range collapses to the start when creating a native `Range`.

example


ianstormtaylor:
Thanks @kellyjosephprice!

Issue #1500, opened 2018-01-04
`change.unwrapBlock` sometimes incorrectly modifies block selection


CameronAckermanSEL:
#### Do you want to request a *feature* or report a *bug*?

Bug

#### What's the current behavior?

When `change.unwrapBlock` is called when there is an expanded selection on a block within a list, sometimes the selection will be modified by this operation, and subsequent operations will not apply to the original selection.

There's a specific case where this is a problem. If the user has highlighted a list item within a list and tries to change the list type for the selected list item, then we see the odd behavior below where the blocks preceding the original selection are selected after the `change.unwrapBlock` operation and then get wrapped in a list (so many words!!) 

Here's the gif to accompany my word heap :)

![](http://g.recordit.co/LzVU9xiPci.gif)

#### What's the expected behavior?

`change.unwrapBlock` should not modify the selection in this way.

darkyshiny:
I've been messing with this quite a bit in a project that I am working on and will share what I have gathered so far.

As far as I have been able to tell during my testing this bug only occurs when doing a backwards selection on a single row in a list **that is not the last item in the list**. When triggered `change.unwrapBlock` will extend the selection all the way to the begging of the list in question.

![slate-unwrapblock-bug](https://user-images.githubusercontent.com/16773900/35323576-473d0516-00ee-11e8-9d97-79e12b50351d.gif)

A hacky workaround that anyone facing this can use until it is properly fixed is checking if the `selection.isBackward` and simply calling `change.flip()` before and after unwrapping/rewrapping to avoid having a backwards selection at the time of unwrapping the list.

I might take a crack at looking into a PR for this later this week if I find the time for it. 👍

PR #1575, opened 2018-01-30
Fix range.moveToRangeOf when range is a backward range;

zhujinxuan:
@CameronAckermanSEL   Can you have a test whether unwrapBlock is good after this PR?

It should fix https://github.com/ianstormtaylor/slate/issues/1500
The problem is that moveToRangeOf does not consider the case `Range.isBackward`.  I think there can be two options:
1. `moveToRangeOf` always return with `isBackward: false`
2. `moveToRangeOf` keep `isBackward`, but switch anchor and focus if necessary.

In this PR, I choose the second solution

### DIFF
--- a/packages/slate/src/models/range.js
+++ b/packages/slate/src/models/range.js
@@ -657,11 +657,8 @@ class Range extends Record(DEFAULTS) {
    */
 
   moveToRangeOf(start, end = start) {
-    if (this.isBackward) {
-      return this.moveAnchorToEndOf(end).moveFocusToStartOf(start)
-    }
-    return this
-      .moveAnchorToStartOf(start)
+    const range = this.isBackward ? this.flip() : this
+    return range.moveAnchorToStartOf(start)
       .moveFocusToEndOf(end)
   }
 
--- a/packages/slate/src/models/range.js
+++ b/packages/slate/src/models/range.js
@@ -658,7 +658,8 @@ class Range extends Record(DEFAULTS) {
 
   moveToRangeOf(start, end = start) {
     const range = this.isBackward ? this.flip() : this
-    return range.moveAnchorToStartOf(start)
+    return range
+      .moveAnchorToStartOf(start)
       .moveFocusToEndOf(end)
   }
 
--- a/packages/slate/src/models/range.js
+++ b/packages/slate/src/models/range.js
@@ -657,6 +657,9 @@ class Range extends Record(DEFAULTS) {
    */
 
   moveToRangeOf(start, end = start) {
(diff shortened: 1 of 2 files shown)


CameronAckermanSEL:
I'm really confident that this will fix the defect. Nice job @zhujinxuan !

ianstormtaylor:
Hey @zhujinxuan I'm not totally sure I understand, but I think that (1) makes more sense a solution? That way you know what to expect no matter what when you call `moveToRangeOf`?

ererik:
Here's a test that covers the fail condition:

`packages\ slate\test\changes\at-current-range\unwrap-block\middle-child-block-reversed.js`

```
/** @jsx h */

import h from '../../../helpers/h'

export default function (change) {
  change.unwrapBlock('quote')
}

export const input = (
  <value>
    <document>
      <quote>
        <paragraph>
          one
        </paragraph>
        <paragraph>
          t<focus />w<anchor />o
        </paragraph>
        <paragraph>
          three
        </paragraph>
      </quote>
    </document>
  </value>
)

export const output = (
  <value>
    <document>
      <quote>
        <paragraph>
          one
        </paragraph>
      </quote>
      <paragraph>
        t<focus />w<anchor />o
      </paragraph>
      <quote>
        <paragraph>
            three
        </paragraph>
      </quote>
    </document>
  </value>
)

```

zhujinxuan:
@ianstormtaylor The problem is caused by the following code and alike codes in `by-key.js`
```
  const { value } = change
  ....
  const range = selection.moveToRangeOf(first, last)
```

The problem is that, when the selection is backward,  the original `moveToRangeOf` cannot switch the anchor and focus wisely.

zhujinxuan:
@ererik Thanks. The test is added

packages/slate/src/models/range.js
```diff
@@ -657,6 +657,10 @@ class Range extends Record(DEFAULTS) {
    */
 
   moveToRangeOf(start, end = start) {
+    if (this.isBackward) {
+      return this.flip().moveAnchorToStartOf(start)
+        .moveFocusToEndOf(end)
+    }
```
ianstormtaylor: Can you change this to only have a single code path, so we're not duplicating the `moveAnchorToStartOf` stuff? For example:

```
let range = this

if (this.isBackward) {
  range = this.flip()
}

return range
  ...
```

zhujinxuan: Fixed. thanks

ianstormtaylor:
Thanks @zhujinxuan!

Issue #1402, opened 2017-11-14
Unbounded Cache causing some Memory Issues


YurkaninRyan:
Currently we memoize calls to select functions, but there is no limit to how much we store.  Our caches uses `immutable` which is a bit of a memory hog.  Eventually, the size of the cache can grow to be so large that it causes an out of memory error in V8.

![screen shot 2017-11-14 at 10 55 18 am](https://user-images.githubusercontent.com/8742831/32789645-5c40b36e-c92a-11e7-96a8-74670d1d3d16.png)

doing some light memory profiling, the amount of memory that is created by normalizing is pretty heavy.  At any rate, if we let the cache grow and grow, eventually it will hit a size that the browser can't collect and it will just crash. 

Couple questions:
  1. Can we be more selective on the functions we memoize?
  2. Can we be more selective of when we memoize those functions?  Should normalizing cause a memoize?
3. Can we prune the cache over time?

YurkaninRyan:
Here's a better drill down on the object that has the normalize.

![screen shot 2017-11-14 at 11 14 57 am](https://user-images.githubusercontent.com/8742831/32790699-22c26b3e-c92d-11e7-93f4-d2bcd928de66.png)

Here is a graph displaying what the memory consumption looks like over time,  after a minor gc, it just never seems to be able to gc again

![screen shot 2017-11-13 at 2 53 39 pm](https://user-images.githubusercontent.com/8742831/32790800-60a7ea50-c92d-11e7-88eb-1c7808a6ec16.png)

doodlewind:
Quite concerned about this. Wondering are there some other factors causing the memory leak? e.g., document size, schema rules size, operations size? And can we observe some other side effects like slower re-render /  page crashed?

ianstormtaylor:
Off the top of my head, some random ideas for things that could lead to the larger-than-ideal memory usage:

- The `memoize` implementation being incorrect and leaky somehow...
- Memoizing too many under-used methods that aren't actually giving us enough performance benefits to be worth it. (We can absolutely get rid of any memoized methods that don't actually give us real performance improvements for the 90% use cases, but research needs to show which they are.)
- Memoizing methods that often get called with highly-variable arguments, resulting in poor cache hits and lots of cached values building up in memory. (Right now memoized methods are split into argument-less and argument-ed methods, so that would be a place to start looking.)
[...]

PR #1576, opened 2018-01-31
Store operations as plain JS objects in History

bengotow:
Hey! This PR is a suggested mitigation for #1402 and reduces the memory overhead of keeping a large undo stack by converting the `Operation` objects in `History` to plain JS (and stripping the `__cache`  field used for memoization.)

In order to make this change, I removed the selection `paths` from the JSON representation in favor of the `keys` in the underlying `Range`. I'm not sure what the motivation is for storing them as paths (maybe an experiment toward #1408? maybe support for collaborative editing?) Because the serialized and deserialized Selection objects no longer have `value.document`, they can't perform the coercion from path to key and vice versa.

To implement this and keep the added benefits of using `paths`, I think the best move (long term at least) is to fully implement #1408, since using paths everywhere would also eliminate the need for `value.document` and the conversion.

**Outcomes**

Test Scenario:
- Launch app
- Open editor
- Start profiler
- Type "A" followed by "Backspace" 20 times
- Stop profiler

Note that in both results below the memory usage is growing because we haven't hit the undo limit, so we're storing more and more data with every keystroke. It's just that in the after scenario we're storing so little it appears it's not growing.

Before:
screen shot 2018-01-30 at 4 29 16 pm

After:
screen shot 2018-01-30 at 4 27 02 pm

### DIFF
--- a/packages/slate/src/changes/on-history.js
+++ b/packages/slate/src/changes/on-history.js
@@ -1,5 +1,6 @@
 
 import invert from '../operations/invert'
+import Operation from '../models/operation'
 import omit from 'lodash/omit'
 
 /**
@@ -30,7 +31,7 @@ Changes.redo = (change) => {
   undos = undos.push(next)
 
   // Replay the next operations.
-  next.forEach((op) => {
+  next.map(json => Operation.fromJSON(json)).forEach((op) => {
     const { type, properties } = op
 
     // When the operation mutates the selection, omit its `isFocused` value to
@@ -69,7 +70,7 @@ Changes.undo = (change) => {
   redos = redos.push(previous)
 
   // Replay the inverse of the previous operations.
-  previous.slice().reverse().map(invert).forEach((inverse) => {
+  previous.slice().reverse().map(json => invert(Operation.fromJSON(json))).forEach((inverse) => {
     const { type, properties } = inverse
 
     // When the operation mutates the selection, omit its `isFocused` value to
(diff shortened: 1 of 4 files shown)


justinweiss:
This is really great, but it doesn't seem quite right to serialize keys, because keys only make sense on a specific live version of a document. All of the other operations only carry around paths, so it would make more sense to me to turn the Selection keys into paths when it's serialized.

I don't think that would affect the behavior -- if you're not working collaboratively, paths will always be valid on the document they generated. Well, unless you ran some changes with `save:false`, but all bets are pretty much off at that point :-)

bengotow:
Hmmm I think you're right - with this change in place, it becomes impossible to save the editor state with the undo history, restore it later from JSON, and re-create selection changes, since the keys would be different when the document was re-created.

I think a possible solution would be for the `undo` and `redo` Change methods to provide the Operation with a `document`, allowing it to do the mapping from the saved `anchorPath` to an `anchorKey` in the new document. Come to think of it, I'm not sure when/where the operations in the stack get a `value` when the document is re-inflated. Not sure how this worked before :-)

justinweiss:
I don’t think they do! Luckily, you can reconstruct it if you really wanted to — the top operation on the stack has the value, `apply(current value, invert(operation))`, the next one has `apply(just calculated value, invert(operation))`, and so on. The way I remember it is, the operation should always contain the value it was sourced from, and you should always be able to apply an operation to its own value. 

Actually, if I remember right, Slate doesn’t recalculate the value when it inverts the rest of the operation, which could cause some problems.

ianstormtaylor:

PR #1578, opened 2018-01-31
Warn during normalization when range offsets are not numbers

majelbstoat:
Feel free to take or leave this, but I just spent a fair amount of time trying to understand why `change.addMarkAtRange()` was extending my mark to the end of each text node. Turns out, I was reading the offsets straight into a Range from window.location, and in `change.addMarkByKey()`, it was doing `offset + length` as `"13" + 158` giving `"13158"` instead of the expected `171` for the end of the range.  JavaScript 🙄.

Figured that normalizing offsets in `range.normalize` was a reasonable approach, and that it might save someone the same trouble later.

Happy to write tests for this, but couldn't see any locations that models were being tested, nor a specific `tests/changes/on-document` section either. If you let me know the right location, I can write one.

### DIFF
--- a/packages/slate/src/models/range.js
+++ b/packages/slate/src/models/range.js
@@ -674,6 +674,12 @@ class Range extends Record(DEFAULTS) {
     const range = this
     let { anchorKey, anchorOffset, focusKey, focusOffset, isBackward } = range
 
+    const anchorOffsetType = typeof anchorOffset
+    const focusOffsetType = typeof focusOffset
+    if (anchorOffsetType != 'number' || focusOffsetType != 'number') {
+      logger.warn(`The range offsets should be numbers, but they were of type "${anchorOffsetType}" and "${focusOffsetType}".`)
+    }
+
     // If the range is unset, make sure it is properly zeroed out.
     if (anchorKey == null || focusKey == null) {
       return range.merge({


ianstormtaylor:
@majelbstoat can you show the example code you were using? I'm a little wary of doing this here for people because it may be unexpectedly covering up other issues.

majelbstoat:
Yeah, I hear that. Simplified version is:

```
      // query is the nextjs object that gives you a map of strings from window.location query params.
      // e.g. ?highlight=2,20,2,45
      const [anchorKey, anchorOffset, focusKey, focusOffset] = query.highlight.split(',')
      change.addMarkAtRange(
        new Range({
          anchorKey,
          anchorOffset,
          focusKey,
          focusOffset,
        }),
        'foo'
      )
    })
```

with the above example, in addMarkByKey, it created a mark from `(2, 20)` to `(2, 2025)`, because length is `"45" - "20"`, which is `25`, but the calculated mark range is `"20" + 25`, which is `"2025"`. (That `"2025"` gets implicitly cast to a number later, and capped at the end of the text.)

Definitely programmer error, but required a bunch of spelunking. A reasonable alternative might be to throw a warning if they're not Numbers, if you want to be more explicit, which would be inline with the behaviour of other badly-defined ranges in range.normalize.

ianstormtaylor:
I'd be open to throwing a warning in `range.normalize`.

I'd also be open to throwing an error in `Range.create`, which we encourage using instead of `new Range` since Immutable doesn't give us a nice chance to override the constructor last I checked.

majelbstoat:
Oh yeah, I should totally be using Range.create 😅.

I'll modify this to add a warning to normalize, as that would catch this error and one in Range.create.

(Question about tests for this still stands - are the models specifically tested, other than implicitly through change tests?)

ianstormtaylor:
@majelbstoat nope, no real model tests right now, so just the warnings and manually testing sounds good to me.

majelbstoat:
Updated to warn and confirmed it's logging as expected when run locally. (And updated my own code to use `Range.create` 😉)

packages/slate/src/models/range.js
```diff
@@ -674,6 +674,12 @@ class Range extends Record(DEFAULTS) {
     const range = this
     let { anchorKey, anchorOffset, focusKey, focusOffset, isBackward } = range
 
+    const anchorOffsetType = typeof anchorOffset
+    const focusOffsetType = typeof focusOffset
+    if (anchorOffsetType != 'number' || focusOffsetType != 'number') {
+      logger.warn('The range offsets should be numbers, but they were of type', anchorOffsetType, 'and', focusOffsetType)
```
ianstormtaylor: Can you make this a `` style string, with the type variables interpolated? And add `"` around each of them, that way it'll show up as:

> Range offsets should be numbers, but they were of type "string" and "string".

Which will hopefully be easier to read quickly. And add a period, thanks!

majelbstoat: done :) I appreciate the call for consistency!

ianstormtaylor:
Thank you @majelbstoat, great PR!

PR #1583, opened 2018-02-01
Documentation: Wrong onKeyDown event condition

wasabigeek:
Use `&&` instead of `||` for `control-\`` to work


wasabigeek:
My bad. It doesn't seem to work on Edge, hence I thought it was an error.

PR #1585, opened 2018-02-01
Docs: Add that Edge does not support `CTRL+...`

wasabigeek:
The CTRL-` example fails in Edge due to lack of support for CTRL+... events. Reference: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/742263/.

Added a note so other Edge users won't be tripped up by the walkthrough.


ianstormtaylor:
Thanks @wasabigeek!

PR #1586, opened 2018-02-01
Fix broken "Characters" link in docs

andreasmischke:
Link was pointing to `./mark.md`, now points to `./character.md`


schneidmaster:
@andreasmischke Good catch, thanks!

PR #1592, opened 2018-02-03
fixed build for windows

DamareYoh:
The lovely rollup update caused the `yarn build` command to break for windows, because the environment cannot be set in this way.

I fixed this by adding a module called cross-env to the devDepdendencies. It helps alleviate this cross platform incompatibility.

See stack overflow question here https://stackoverflow.com/a/37131308


ianstormtaylor:
Ah interesting, thanks @DamareYoh!

Issue #1455, opened 2017-12-06
Bug in table undo: The last cell is not restored after delete->undo change


zhujinxuan:
#### Do you want to request a *feature* or report a *bug*?

 A Bug

#### What's the current behavior?

Unexpected `undo` behavior in re-inserting a table;  See the gif below

![tfpwgw4hyb](https://user-images.githubusercontent.com/746159/33671576-f898ae36-da75-11e7-895b-b4d2712b667e.gif)

#### What's the expected behavior?

The last cell should be in the cell.

zGrav:
Possibly could be fixed with #1456 if a solution is developed for it.

zhujinxuan:
@zGrav  I am not sure whether bundle change could solve the problem.  I try undo multiple times but the last cell is still not restored

CameronAckermanSEL:
This should be resolved in #1594

zhujinxuan:
@CameronAckermanSEL  Fixed.  Thank you.

Issue #1394, opened 2017-11-09
Block type of last block is wrong after undo change


czechdave:
#### Do you want to request a *feature* or report a *bug*?
BUG

#### What's the current behavior?
https://monosnap.com/file/OLDUdIkmmyi1AwDtoAYQJ0AtscoCB7
Go to http://slatejs.org/#/forced-layout
Select all text (mouse or cmd+a)
Press backspace/delete or type something
Press cmd+z (undo button)
The last block will have the same type as the first block.

#### What's the expected behavior?
The last block has the same type as it had initially after undo.

CameronAckermanSEL:
This should be resolved in #1594

PR #1594, opened 2018-02-03
Bugfix/undo merge node

### LINKED ISSUE #1485: Undo deletion of list restores final list item as a paragraph

Bug

After deleting a list and then undoing the change, the list is recreated in the editor, but the final list item is restored as a paragraph.

http://recordit.co/eUi7ix9PLW

When list is restored, all list items have the correct type

PS. Apologies if this is a duplicate, there seem to be some other flavors of this bug already reported.

DamareYoh:
This fixes #1485 (and others that I will dig up later :))

In this defect, if nodes were merged (specifically via deletion of a range), then undoing this change would not properly restore the data and type of the node because the `split_node` operation resulting from the invert of `merge_node` did not have enough information to do so.

I fixed this by adding an extra property to the `merge_node` operation object called `original` that stored the block that was merged into the other one.

I'm definitely not married to any names here :), and I wasn't exactly sure what the implications of this would be on the operation feature. Hopefully this will be a sufficient fix without causing too much disruption to other features.


packages/slate/test/history/undo/delete-across-blocks.js
```diff
@@ -18,9 +18,9 @@ export const input = (
       <paragraph>
         o<anchor />ne
       </paragraph>
```
DamareYoh: this test would have caught the defect if the types were different. 😆

packages/slate/src/operations/apply.js
```diff
@@ -405,6 +405,20 @@ const APPLIERS = {
 
     // Split the node by its parent.
     parent = parent.splitNode(index, position)
```
DamareYoh: hopefully the names are ok and the comments not too verbose! :)

packages/slate/src/constants/operation-attributes.js
```diff
@@ -29,6 +29,7 @@ const OPERATION_ATTRIBUTES = {
     'value',
```
DamareYoh: property called 'original' added to the merge_node and split_node operations

DamareYoh: While I was writing documentation, I concluded that the name 'original' is bad here. I'm not sure what a good name is, so I'll wait for your feedback before I go pick one :)

I think maybe if we just have `type` and `data` that's clear, easy to document, and probably more lightweight

justinweiss:
Is there a reason to carry around the entire node, instead of just `data` and `type`?

DamareYoh:
Not in particular,  I can make the adjustment. Should the properties just be called 'data' and 'type'?

justinweiss:
Good question. `set_node` hangs them off of `properties`, as does the documentation, but I'm not sure how that would interact with the other operations that use `properties`. I'm guessing it would be fine, though.

justinweiss:
Yeah, it looks like a few operations use properties, and just switch behavior based on the type: https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/models/operation.js#L133

DamareYoh:
Ok. So that'd be kinda similar to the existing approach, except the name is much better and we don't have to pass it along the child nodes. Should I hard code the merge_node operation created in `mergeNodeByKey` to have a properties object that has `type` and `data` only then?

justinweiss:
Yeah, that seems to make sense. You could even add them to splitNodeByKey if you wanted to -- it wouldn't hurt anything, and might make things more consistent during split-undo-redo.

DamareYoh:
Great. I think that's a major improvement, and now the interface makes sense. 

Thanks very much :)

justinweiss:
Awesome! That's true (about splitNode), but looking at https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/models/node.js#L1898, it duplicates the original node -- so I wonder if we can assume when we generate the `split_node` operation, that it will carry over the `type` and `data` from the original node?

DamareYoh:
Some time ago I asked about the behavior of `split_node` passing along the type and the data, and the response was that it was by design behavior, so I think as long as we leave this behavior in tact from the `splitNodeByKey` end we shouldn't be breaking anything and it will be consistent with how users consume the API now.

If the `split_node` operation is being generated manually, the caller would be able to provide their own properties object and have it merge anything they wanted to the new node created from the split similar to how set node does. So we sort of got a free enhancement for the folks who are in the nitty-gritty operations level of things.

From `splitNodeByKey`, unless we want to change the interface now and all the user to specify a properties object as  an argument, this behavior is unfortunately hidden. At the very least, the properties object from `mergeNodeByKey` can get to the `split_node` operation because the `invert` function will pass it along for us for free

Does this address the concern?

DamareYoh:
This fixes:
#1485 
#1394

This PROBABLY fixes:
#1455

justinweiss:
Yeah, this is definitely an improvement. For `splitNodeByKey`, I'm having trouble seeing where the change in behavior would come from -- I don't think we'd have to change the interface of that method at all.

Internally, it could generate a `split_node` operation with the `data` and `type` keys set to the same values as the node it split off from (for non-text nodes, at least). With the rest of your code exactly the way it is right now, I don't think there'd be any change in interface or behavior -- things would just work. When the operation is applied, it'd set `data` and `type` to the values set in `split_node`, which came from the original node. Today, when the operation is applied, it doesn't touch `data` or `type` when cloning the node, so they're also set to the same value as the original.

I could totally be missing something, though.

Beyond that, the only other thing that would be cool, would be to serialize / deserialize properties in Operation's `fromJSON`.

CameronAckermanSEL:
Oh ok I think I understand what you're saying now. I'll give it a shot!

DamareYoh:
Ok, I think I understand what you're saying. Let me give it a shot!

justinweiss:
I just pulled this down to try it out, and it feels _so_ much better. Thanks for the contribution!

One last question I had -- In https://github.com/ianstormtaylor/slate/pull/1594/files#diff-61e2b3d4761a8c79aed160d90bf78b43R412, you have the check to make sure that it's a `block` node -- would it break without that check? If not, I don't think you need it. If so, it might also need to check for inline nodes, since they also have data and type.

Thanks again!

DamareYoh:
Hey @justinweiss, I updated one of the tests that involved deleting nodes across inlines, and the existing code passed, which was a bit surprising to me. I ran the tests with the check removed enturely, and a ton of errors happened associated with text nodes, so I just changed the check to `splitNode.object !== 'text'`.

I think this is the most correct logic and the manual test still behaves correctly after this change.

justinweiss:
Awesome, thanks for the fix!

DamareYoh:
@justinweiss thanks for all the help making this better! :)

PR #1595, opened 2018-02-03
Improvements to developer experience with building packages

### LINKED ISSUE #1593: yarn build and yarn test improvements with rollup

Not really sure XD

The current behavior (and please correct me if I am wrong :)) is that 

`yarn build` defaults to building production data
`yarn test` tests the production build
`yarn start` attempts to open a dev example site that does not get built anymore by `yarn build`

The consequences of this are that (test) development on slate is slower because, one has to run `yarn build` to get `yarn test` to function. `yarn build` also takes a very long time because it has to do extra steps for a production build.

`yarn start` also now yields a 404 because it tries to run against the dev site, which does not get built in `yarn build`

It seems to me that `yarn build` should default to building in a development environment, and tests should be run against that for the purposes of having quicker development.

Alternatively, we should have a family of commands that run development equivalents of the aforementioned commands.

schneidmaster:
Based on the conversation in #1593 

1. Default to emitting both CJS and ES even in development/watch mode, as CJS is used by the tests so this allows for auto rebuilds that the tests can be run against

2. Move the `http-server` command to `yarn server` and change `yarn start` to both start the rollup watch and also start the http server. This is for a slightly more intuitive experience getting up and running -- you expect `yarn start` to be the default command to get started in development, rather than needing to either build first or run `yarn watch` instead.

3. Add a `watch:packages` command that builds & watches the packages (but not the examples), for super fast auto rebuilds when you are developing on a package and running tests against it but not using the examples.

Closes #1593

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -72,6 +72,8 @@ To keep the source rebuilding on every file change, you need to run an additiona
 yarn run watch
 ```
 
+If you're not using the examples in-browser (and only running the tests against the watched source), you can instead run `yarn watch:packages` for faster rebuilds.
+
 If you need to debug something, you can add a `debugger` line to the source, and then run `yarn run test debug`. 
 
 If you only want to run a specific test or tests, you can run `yarn run test --fgrep="slate-react rendering"` flag which will filter the tests being run by grepping for the string in each test.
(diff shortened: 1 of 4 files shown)


DamareYoh:
Works great! thank you :)

ianstormtaylor:
Nice, thanks @schneidmaster!

PR #1597, opened 2018-02-03
[Doc] Adding an unknown mark doesn't produce an error

stefk:
The *Applying custom formatting* guide states that Slate outputs an error in the console when a mark is applied without a corresponding renderer but AFAIK, that's not the case in the current version. I'm guessing this is an old behaviour (I'm completely new to this library), hence the doc update.


schneidmaster:
Thanks @stefk, good catch! Confirmed I see the same thing for an unknown mark.

Issue #1590, opened 2018-02-02
Walkthrough - metaKey example fails on Windows


wasabigeek:
Referring to: https://docs.slatejs.org/walkthroughs/applying-custom-formatting, the `meta+B` shortcut doesn't work, although `meta+backtick` does. I believe this is because Windows already maps that [particular shortcut](https://support.microsoft.com/en-sg/help/12445/windows-keyboard-shortcuts). 

Suggest to reuse the CTRL key example from the previous!

schneidmaster:
@wasabigeek The help please label generally just means that a PR would be welcomed to address the issue. In this case I think a PR changing the example to use `ctrl` instead of `meta` (assuming that works cross-platform) would be great 👍

schneidmaster:
Fixed in #1601

PR #1601, opened 2018-02-05
Examples: Use ctrlKey instead of metaKey (for Windows compatibility)

wasabigeek:
As discussed in https://github.com/ianstormtaylor/slate/issues/1590, Windows already maps the `win-B` (metaKey) shortcut, which makes the walkthrough example fail. This changes the examples to use ctrlKey throughout.


ianstormtaylor:
Thanks @wasabigeek!

PR #1603, opened 2018-02-05
fix(updateSelection): make sure there is a node to extend from

tobiasandersen:
`Selection.collapse()` can fail at setting the anchor point, but it does so without throwing. So when we do the following in `updateSelection`:
```
native.collapse(range.startContainer, range.startOffset)
native.extend(range.endContainer, range.endOffset)
```

`extend()` will throw if `collapse()` wasn't successful (since there's nothing to extend from). I haven't been able to figure out why it fails, but we can reproduce it consistently in our app (on Chrome). I can't find any info on when it could fail etc.

This PR just adds a safety check before running `extend` so that an error isn't thrown. What's interesting in our app is that it behaves as expected with this even when it can't set the anchor point.

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -213,10 +213,16 @@ class Content extends React.Component {
       // we need to check and do the right thing here.
       if (isBackward) {
         native.collapse(range.endContainer, range.endOffset)
-        native.extend(range.startContainer, range.startOffset)
+
+        if (native.anchorNode) {
+          native.extend(range.startContainer, range.startOffset)
+        }
       } else {
         native.collapse(range.startContainer, range.startOffset)
-        native.extend(range.endContainer, range.endOffset)
+
+        if (native.anchorNode) {
+          native.extend(range.endContainer, range.endOffset)
+        }
       }
     } else {
       // COMPAT: IE 11 does not support Selection.extend, fallback to addRange
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -207,22 +207,24 @@ class Content extends React.Component {
     this.tmp.isUpdatingSelection = true
     native.removeAllRanges()
 
-    // COMPAT: IE 11 does not support Selection.extend
-    if (native.extend) {
+    // COMPAT: IE 11 does not support Selection.setBaseAndExtent
+    if (native.setBaseAndExtent) {
       // COMPAT: Since the DOM range has no concept of backwards/forwards
       // we need to check and do the right thing here.
       if (isBackward) {
-        native.collapse(range.endContainer, range.endOffset)
-
-        if (native.anchorNode) {
-          native.extend(range.startContainer, range.startOffset)
-        }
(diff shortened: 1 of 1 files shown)


CameronAckermanSEL:
I thought I'd update this PR with a discussion I was having with @tobiasandersen so we wouldn't lose any info.

For the referenced issue #1609 , I tried this PR and unfortunately it did not alleviate the problems I was experiencing because this PR fixes an exception coming out of `native.expand()`, not one coming out of `native.collapse()`. The `native.collapse()` exception only happens in firefox (I'm using latest firefox 58.x). Reverting the `collapse/expand` calls back to `setBaseAndExtent()` resolves the issues in firefox, and also works properly in latest chrome.

That being said, there is said to be an issue with `setBaseAndExtent` in chrome (unsure what version) that causes the `native` object to have a null `anchorNode` and causes some error that eventually results in the selection being correctly applied. I am unable to reproduce this problem, but perhaps the fix is to try `setBaseAndExtent` and handle the case where `anchorNode` is null after that separately

tobiasandersen:
Thanks for updating @CameronAckermanSEL. So concretely, that would mean the following change instead:

**Before:**
```
native.collapse(range.startContainer, range.startOffset)
native.extend(range.endContainer, range.endOffset)
```

**After:**
```
native.setBaseAndExtent(
  range.startContainer, 
  range.startOffset, 
  range.endContainer, 
  range.endOffset
)
```

And just to add/re-iterate: I did try this as well. But for us (on Chrome 63), it had the exact same result (visually and functionally), as the changes in this PR. And I felt what I did was a smaller change with lower risk. But since `setBaseAndExtent` seems to solve the issues that @CameronAckermanSEL are seeing, I'm definitely open to changing it!

CameronAckermanSEL:
@tobiasandersen yep! That is the proposal!

There was about a 1 month window where setBaseAndExtent was used in slate-react, but it got changed to the expanded form for IE compatibility (but IE doesnt support the selection.expand()  call anyway so thats why there's the check for addRange)

Here's the code I've been using on my private fork:

```
    // COMPAT: IE 11 does not support Selection.setBaseAndExtent
    if (native.setBaseAndExtent) {
      // COMPAT: Since the DOM range has no concept of backwards/forwards
      // we need to check and do the right thing here.
      if (isBackward) {
        native.setBaseAndExtent(
          range.endContainer,
          range.endOffset,
          range.startContainer,
          range.startOffset)
      } else {
        native.setBaseAndExtent(range.startContainer,
          range.startOffset,
          range.endContainer,
          range.endOffset)
      }
    } else {
      // COMPAT: IE 11 does not support Selection.setBaseAndExtent, fallback to addRange
      native.addRange(range)
    }

```

tobiasandersen:
Okay I've updated the PR according to our discussion above.

CameronAckermanSEL:
Thank you so much! Very excited to use this! :)

ianstormtaylor:
Sounds good, thank @tobiasandersen and @CameronAckermanSEL!

PR #1605, opened 2018-02-05
Fix triple selection extending to next block

Kornil:
Fixes https://github.com/ianstormtaylor/slate/issues/752

Explanation: on triple mouse click, not only you select one block, but the next one even tho it is visually not specified and an issue from UX perspective (unexpected result). When setting a different block to a selection made this way, the next block gets included as well. Gif explains it even better:

(Look at the block under the selected block)

slate triple selection bug #1
slate triple selection bug #2

I basically took the code to detect hanging selection from `deleteAtRange` and added a simple `if` checking hanging selection(triple click). if `true`, simply removes last block from selection.

New behaviour is now tested and working nicely with the other tests.

~I have an issue with the tests now, they are failing and I'm not 100% sure of how to fix them, if someone could help or point me in the right direction I'd appreciate it.~


zhujinxuan:
Why the hanging is related to `startOffset === 0`?

zhujinxuan:
BTW, there is another method like;
```
  if (range.isBackward) {
    return change.setBlockAtRange(range.flip())
  }
  if (!endBlock.isVoid && range.collapseToEnd().isAtStartOf(endBlock) && !endBlock.getDescendant(startKey)) {
    return change.setBlockAtRange(range.moveFocusToEndOf(document.getPreviousText(endKey)))
  }
```

Kornil:
Got it from here https://github.com/Kornil/slate/commit/cf0745a178d2bb1fd1aaf45af7bbda8cbd2eafe0#diff-dc8d1609b4f86c7c528f714e5ec5d3c6R78 code @ianstormtaylor wrote to detect hanging selection

zhujinxuan:
@Kornil I know that, but I am not quite understanding the original `deleteAtRange` function.   I think the `startOffset === 0` is to keep the `startText.key` not deleted, while this feature is not quite necessary in `setBlockAtRange`

CameronAckermanSEL:
set-block/across-blocks
   Failing because the test expects the paragraph block that has the focus at the start to also change to code. This is testing the behavior being removed by this change

set-block/data-only
set-block/single-block
set-block/nested-block
set-block/single-block-string-shorthand
set-block/with-data-as-map
set-block/with-data-as-object
set-block/with-is-void
  failing because the cursor is at the start of the block, means the logic is not handling the case where the cursor is collapsed at the start of the block. We definitely want the set to work here.

packages/slate/test/changes/at-current-range/set-block/across-blocks.js
```diff
@@ -26,7 +26,7 @@ export const output = (
         <anchor />word
       </code>
       <code>
-        <focus />another
+        a<focus />nother
```
CameronAckermanSEL: I think maybe this calls for an additional test.

1. That checks that the cursor being expanded but ending at the start of the block does NOT result in the focus block being changed
2. Basically what you implemented here.

Kornil:
Thanks @CameronAckermanSEL for the input on why tests were failing, I fixed them by adding a isCollapsed check on the function to detect hanging selection.

@zhujinxuan I removed the unused code, and as a matter of fact, does not change anything, you were correct.

I modified the `across-blocks` test because it would always fail with the new hanging detection, what do you think? should I test the new behaviour as well?

packages/slate/test/changes/at-current-range/set-block/across-blocks.js
Kornil: I was not sure how to proceed here so I put that logic in its own commit in case we wanted to revert it, I'll add a new test for the new behaviour (meaning restoring the old test but with the new behaviour).

zhujinxuan:
What will happen if someone, a very mouse expert, consciously select to the `offset=0` of the next block and then wrapBlock or setMark?

zhujinxuan:
I do not think 
```
startKey == startBlock.getFirstText().key
```
is necessary. Because even the startKey is not the first Text, we still should not wrap the endBlock

`deleteAtRange` is dealing a quite complex normalization, we would like to keep simple in wrapBlockByKey

CameronAckermanSEL:
@zhujinxuan I may be mistaken, but I think the change in this PR only impacts `setBlockAtRange`

Kornil:
The scope of this PR is limited to fix the `setBlock` behaviour on triple click, we could deal with the other behaviour in a separate one to preserve scope.

zhujinxuan:
@CameronAckermanSEL Ohhhh... I understand `startOffset === 0 && startKey === firstText.key` happens only with the triple click.

I did not notice that.

zhujinxuan:
@Kornil  Please ignore one of my previous comment.  When I said about deleting `startOffset ===0`, I do not notice that it is an identified behavior of triple-click

Kornil:
@zhujinxuan done && solved conflicts

packages/slate/src/changes/at-range.js
```diff
@@ -948,9 +948,32 @@ Changes.setBlockAtRange = (change, range, properties, options = {}) => {
...
+  const isHanging =
+    isCollapsed == false &&
+    startOffset == 0 &&
+    endOffset == 0 &&
+    isStartVoid == false &&
+    startKey == startBlock.getFirstText().key &&
+    endKey == endBlock.getFirstText().key
+
+  // If it's a hanging selection, ignore the last block.
+  if (isHanging) {
```
ianstormtaylor: Hey @Kornil can you change this logic to eliminate the extra code path here? so it would instead become like:

```js
const sets = isHanging ? blocks.slice(0, -1) : blocks
sets.forEach( ...
```

Thanks!

Kornil: Done, thanks for the feedback!

ianstormtaylor:
Thank you very much @Kornil for tackling this. This PR looks great!

Slapbox:
Perhaps related to this is the following issue with using CTRL and arrow keys to navigate a document. You can't move to the very end of a line this way. `Slate 0.33.4`.

http://recordit.co/s5ijpzxQPR

Soreine:
@Slapbox I don't think this is related to this PR. This PR only affects what is done when applying changes to the selected blocks through `setBlocksAtRange`.

Soreine:
Due to the fact that this PR does not changes how the selection changes, the root of the problem is not addressed.
If I'm writing any custom logic, like some plugin transforms, I need to be aware of this issue, and detect hanging selections as well.

zhujinxuan:
@Slapbox Perhaps you can have a look with `localstorage.debug = *` in devtools.  I think CTRL+ARROW is handled by onSelect in plugin/after.js in slate-react.

Slapbox:
Whoops I thought I updated here, but actually `Ctrl`+`Arrow` does not move to the end of the line on Windows, except in code editors. I'm just living in code editors, so I didn't realize.

zhujinxuan:
@Slapbox Perhaps we shall make a plugin of emacs-mapping key binding  for moving~

Slapbox:
This is an issue in the newest version of Slate. Triple-click selection grabs the linebreak, but it shouldn't.

LearnedVector:
Hello, I am still having this issue with hovering menu using `slate 0.44.12`. Any work around?

Slapbox:
@LearnedVector your best bet is to create a new issue (link to this one from there) and add a visual demonstration of the issue or reproducible repo as well.

kripod:
Unfortunately, this still seems to be an issue within Slate v0.59.0, but it doesn’t seem to affect Firefox. Are there any workarounds available?

_Related recent issue: #3871_

Issue #1563, opened 2018-01-26
[Discussion] Should validateNode functions have access to Value?


CameronAckermanSEL:
#### Do you want to request a *feature* or report a *bug*?

Feature

#### What's the current behavior?

Current behavior is that `validateNode` only accepts the node undergoing validation as an argument.

#### What's the expected behavior?

I'm wondering how many other folks have found themselves wishing that they could access the value undergoing change. I have some complicated validation rules that depend on where a block sits in the document hierarchy, and this requires me to implement validations at the `object: document` level. 

That's ok, but it would be a large optimization if I could have a `validateNode` rule that operated against a `object: block` and was able to analyze the block's parents. One nice way to achieve this might be to also pass in the Value object undergoing validation.

For example consider a validation where for some reason I only want to add a property to the data of certain blocks based on their position in the document and type.

```
// assume that there's some logic somewhere else that ensures this is only called on a document node
validateSomeBlocksHaveSpecialData(document) {
  let invalids = document.nodes.filter(b => isSpecialType(b));
  // lets say list items are special, but stuff inside of table cells are not, so it doesnt serve me to arbitrarily recurse
  document.nodes.filter(b => isList(b)).forEach(li=> {
    li.nodes.filter(c => isSpecialType(c)).forEach(child => {
       invalids.push(child);
    });
  });
  if (invalids.size) {
    // return change function
  }
}
```
This has to be run at the document level now because that's the only way to get an idea of the document hierarchy.

But what if I also had the value on hand? Then I could have validateNode functions that operate on the block level and look like

```
// assume there is some logic somewhere that ensures this is only called on a block node and block 
// and value are basically passed down to this fn
validateSomeBlocksHaveSpecialData(block, value) {
  // block is special if its not in a table
  if(isSpecialType(b) && !value.document.getClosest(p => p.type !== 'table')) {
    // return change function
  }  
}
```

I can then have a block validation function that operates against the block as it's modified, and I don't have to iterate over the entire document everytime something is changed

Is that a reasonable feature? Or have I wandered into some sort of antipattern?

zhujinxuan:
Because the `validateNode` and `decorateNode` are called by the ancestor chain of the changed node, for example` cell->row->table->document`

I would like to have `ancestors` instead of `value` as the optional argument (in both validateNode and decorateNode).

for the value; I remember we can return 
```
change => ( const {value} = change)
```
, and use the default stack size to prevent un-stopped recursion.

[...]

PR #1611, opened 2018-02-07
Pass value down to validateNode functions

DamareYoh:
(I apologize for reposting this, but I totally FUBAR'd the other PR by doing a merge XD)

This is a proof of concept for an issue #1563 where we pass down the value to `validateNode` calls. It is definitely up for debate whether this would be a positive change or not.

The primary use case for this feature is that it would allow for validation functions that need to analyze the hierarchy of the document. Currently, this is achievable but only if the document node is under validation (or if a block node that contains the blocks that one actually wants to validate is under validation, which is not always possible). Such validate node calls get called every single time any sort of change is made to state, which seems avoidable. It would be nice to be able to have similar rules that operate on the block that changed only, and allowed it to analyze siblings.

Thanks for your consideration :)

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1970,8 +1970,8 @@ class Node {
    * @return {Function|Null}
    */
 
-  validate(schema) {
-    return schema.validateNode(this)
+  validate(schema, value) {
+    return schema.validateNode(this, value)
   }
 }
 
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1967,6 +1967,7 @@ class Node {
    * Validate the node against a `schema`.
    *
    * @param {Schema} schema
+   * @param {Value} value
    * @return {Function|Null}
    */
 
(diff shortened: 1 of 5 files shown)


packages/slate/src/models/node.js
```diff
@@ -1967,11 +1967,12 @@ class Node {
    * Validate the node against a `schema`.
    *
    * @param {Schema} schema
+   * @param {Value} value
    * @return {Function|Null}
    */
 
-  validate(schema) {
-    return schema.validateNode(this)
+  validate(schema, value) {
```
Soreine: Passing the value here defeat the memoization done on `node.validate(schema)`, because the value always change, you can never be sure if a node is still valid for the same schema.


Currently, `validate` says whether a node (and its children) is valid against a schema or not, which is limited and reasonable. I think it should stay that way. Maybe the `schema` should have a way to tell which node in particular will need to inspect the whole value.

Normalizing a node would become:

```
const requiresValue = schema.requiresValue(node);
if (requiresValue) {
  node.validate(schema, value);
} else {
  node.validate(schema);
}
```

CameronAckermanSEL: Thanks @Soreine , :) 

Ian gave me very similar feedback a while ago and I totally forgot to clean up this PR. I greatly appreciate the well thought out feedback, and I hope I didn't waste your time 😆

PR #1612, opened 2018-02-07
Lint JSON, CSS and Markdown files with Prettier

renchap:
Follow up for #1589

Prettier supports more languages than JS(X), so I added support for those. This allows a more consistent style over the repo, and improves readability. For example it makes MD files more readable (tables are correctly spaced, …), fixes some errors (bad numbering in ordered lists) and also formats the JS code embedded in markdown files, making it consistent with the rest of the codebase.

As with the previous PR, there are two commits:
- The first one changes `package.json` to add `.md`, `.json` and `.css` to the Prettier invocations
- The second one is the result of `yarn prettier`, reformatting those files.


ianstormtaylor:
Beautiful, thank you @renchap!

Issue #1613, opened 2018-02-07
[bug] Backspace at one end of inline will lose the selection


zhujinxuan:
#### Do you want to request a *feature* or report a *bug*?

bug

#### What's the current behavior?

If we backspace and then insert like the following:
![](http://g.recordit.co/XMNxz7XYEK.gif)

PR #1614, opened 2018-02-07
Fix getInsertMarksAtRangeAsArray for empty Text Node

zhujinxuan:
When `Text.text === ''`, for example after delete or by normalization around inline, the `getInsertMarksAtRangeAsArray` will throw an error.

This PR fixes that.
Fix https://github.com/ianstormtaylor/slate/issues/1613

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1109,6 +1109,9 @@ class Node {
 
     const text = this.getDescendant(range.startKey)
     const char = text.characters.get(range.startOffset)
+    if (!char) {
+      return []
+    }
     return char.marks.toArray()
   }
 


packages/slate/src/models/node.js
```diff
@@ -1109,6 +1109,11 @@ class Node {
 
     const text = this.getDescendant(range.startKey)
     const char = text.characters.get(range.startOffset)
+    // PREF: if text.characters.size === 0; then no char is got
+    // empty text by normalization or delete at range
```
ianstormtaylor: Hey @zhujinxuan can you remove this comment? Usually `PERF:` comments should be things that could be removed without any breakage and are only for performance improvements. In this case that doesn't apply.

zhujinxuan:
@ianstormtaylor Fixed~

ianstormtaylor:
Thank you @zhujinxuan!

PR #1619, opened 2018-02-08
add ability to click multiple emojis in example

AndreiRailean:
currently, after clicking one emoji, user needs to click in the editor before being able to add another emoji
this fixes that by jumping to next text block and focusing selection to make it obvious where next insertion will take place.

## Before
current-behaviour-screencast 2018-02-08 12-40-13

## After
slatejs-fix-screencast 2018-02-08 12-38-13


ianstormtaylor:
Nice one, thanks @AndreiRailean!

PR #1620, opened 2018-02-08
update example readme with dev server info

AndreiRailean:
`yarn watch` doesn't start the server


ianstormtaylor:
Thanks @AndreiRailean!

PR #1621, opened 2018-02-08
Fix drag and dropping an inline node "forward" in the document

julienp:
When drag and dropping an inline node to a position earlier in the
document, the node is inserted and then removed right after.
Regenerate the key to make sure the old node is removed instead of the
new one.

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -264,11 +264,11 @@ function AfterPlugin() {
     }
 
     if (type == 'node' && Block.isBlock(node)) {
-      change.insertBlock(node).removeNodeByKey(node.key)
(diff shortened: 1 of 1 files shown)


zhujinxuan:
Can we `change.removeNode(node.key).insertNode(node)` ?

packages/slate-react/src/plugins/after.js
```diff
@@ -268,7 +268,7 @@ function AfterPlugin() {
     }
```
ianstormtaylor: Should this happen here for this `Block` handling logic too @julienp? Could you make that fix too while we're at it to stay consistent?

julienp: Indeed! I updated the commit.

ianstormtaylor:
Thanks @julienp!

PR #1622, opened 2018-02-08
Restore Selection after replaceNodeByKey

zhujinxuan:
Problem:
When someone calculate a new node with `Immutable` methods, and then update the node with `replaceNodeByKey`, the selection is not restored:

For example: erasing all text contents in a Row:
```
const nextRow = row.set('nodes', row.nodes.map(cell => {
      const cellTexts = cell.getTexts().map( t => t.set('characters', List()));
      return cell.set('nodes', cellTexts);
   }))
change.replaceNodeByKey(nextRow.key, nextRow, {normalize: false})
```

However, the current `replaceNodeByKey` will move cursor around the the cursor is inside the block. Therefore, I add some codes to restore the cursor~

### DIFF
--- a/packages/slate/src/changes/by-key.js
+++ b/packages/slate/src/changes/by-key.js
@@ -417,16 +417,70 @@ Changes.removeTextByKey = (change, key, offset, length, options = {}) => {
 Changes.replaceNodeByKey = (change, key, newNode, options = {}) => {
   newNode = Node.create(newNode)
   const normalize = change.getFlag('normalize', options)
+  const { cursorPerserve = ['key', 'path'] } = options
   const { value } = change
-  const { document } = value
+  const { document, selection } = value
   const node = document.getNode(key)
   const parent = document.getParent(key)
   const index = parent.nodes.indexOf(node)
   change.removeNodeByKey(key, { normalize: false })
   change.insertNodeByKey(parent.key, index, newNode, options)
+  newNode = change.value.document.getDescendant(newNode.key)
+
+  const { anchorKey, focusKey } = selection
+  if (newNode && anchorKey !== change.value.anchorKey) {
+    const { anchorOffset } = selection
+    const anchorText = refindEdgeTextAfterReplace(
+      anchorKey,
+      node,
+      newNode,
+      cursorPerserve
+    )
+    if (anchorText && anchorText.object === 'text') {
+      const nextAnchorOffset = Math.min(anchorText.text.length, anchorOffset)
+      change.moveAnchorTo(anchorText.key, nextAnchorOffset)
+    }
+  }
+  if (newNode && focusKey !== change.value.focusKey) {
+    const { focusOffset } = selection
+    const focusText = refindEdgeTextAfterReplace(
+      focusKey,
+      node,
+      newNode,
+      cursorPerserve
+    )
+    if (focusText && focusText.object === 'text') {
+      const nextFocusOffset = Math.min(focusText.text.length, focusOffset)
+      change.moveFocusTo(focusText.key, nextFocusOffset)
+    }
+  }
+
   if (normalize) {
     change.normalizeNodeByKey(parent.key)
   }
+}
+
+function refindEdgeTextAfterReplace(edgeKey, oldNode, newNode, methods) {
+  if (!oldNode.getDescendant(edgeKey)) return null
+  let result = null
+  for (const key of methods) {
+    if (result) {
+      return result.object === 'text' ? result : result.getFirstText()
+    }
+    switch (key) {
+      case 'key':
+        result = newNode.getDescendant(edgeKey)
+        break
+      case 'path':
+        if (oldNode.object === 'text') {
+          result = newNode
+        } else {
+          result = newNode.getDescendantAtPath(oldNode.getPath(edgeKey))
+        }
+        break
+    }
+  }
+  return result
 }
 
 /**
--- a/packages/slate/src/changes/by-key.js
+++ b/packages/slate/src/changes/by-key.js
@@ -424,11 +424,13 @@ Changes.replaceNodeByKey = (change, key, newNode, options = {}) => {
   const parent = document.getParent(key)
   const index = parent.nodes.indexOf(node)
   change.removeNodeByKey(key, { normalize: false })
-  change.insertNodeByKey(parent.key, index, newNode, options)
+  change.insertNodeByKey(parent.key, index, newNode, { normalize })
+
+  // Restore the selection after replaceNode
   const { anchorKey, focusKey } = selection
   newNode = change.value.document.getDescendant(newNode.key)
-
-  if (newNode && anchorKey !== change.value.anchorKey) {
+  if (!newNode) return undefined
+  if (anchorKey !== change.value.anchorKey) {
     const { anchorOffset } = selection
     const anchorText = refindByKey(anchorKey, node, newNode)
     if (anchorText) {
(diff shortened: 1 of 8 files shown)


packages/slate/src/changes/by-key.js
```diff
@@ -422,8 +422,23 @@ Changes.replaceNodeByKey = (change, key, newNode, options = {}) => {
   const node = document.getNode(key)
   const parent = document.getParent(key)
   const index = parent.nodes.indexOf(node)
+  const { selection } = change.value
```
ianstormtaylor: Can you move this line up to the one earlier:

```js
```

packages/slate/src/changes/by-key.js
```diff
@@ -422,8 +422,23 @@ Changes.replaceNodeByKey = (change, key, newNode, options = {}) => {
...
   change.removeNodeByKey(key, { normalize: false })
   change.insertNodeByKey(parent.key, index, newNode, options)
+  if (selection !== change.value.selection) {
```
ianstormtaylor: Can we express this instead as `if (anchorKey == key || focusKey == key)`? I think that would be correct still while being more clear why we're doing this. Also could you add a quick comment above this line to explain it too.

packages/slate/src/changes/by-key.js
```diff
@@ -422,8 +422,23 @@ Changes.replaceNodeByKey = (change, key, newNode, options = {}) => {
...
+      const nextAnchorOffset = Math.min(anchorText.text.length, anchorOffset)
+      change.moveAnchorTo(anchorKey, nextAnchorOffset)
+    }
+    const { focusKey, focusOffset } = selection
+    const focusText = newNode.getDescendant(focusKey)
+    if (focusText) {
+      const nextFocusOffset = Math.min(focusText.text.length, focusOffset)
+      change.moveFocusTo(focusKey, nextFocusOffset)
+    }
+  }
```
ianstormtaylor: I think there's another case we're actually not handling here too which is if the node is replaced and the key is different, then it won't be selected at all? But ideally it would be selected in the same spot. Would you be down to solve that as well?

zhujinxuan: Do you mean we do not handle the case `focusText.text.length < focusOffset`, and assuming that the text, with the same key,, is unchanged?

My opinion is that, we should be defensive about the case `focusText.text.length < focusOffset` that the text is potentially changed.  There are two options:
1. Do not restore when `focusText.text.length < focusOffset`
2. Restore to the text end

The second solution has one small, even tiny, advantage.  If the {edge}Key of selection is in 
`<emoji isVoid>'SPC'<focus/></emoji>` Then the focusText is reasonable to convert from 'SPC' to ''.  Then the second solution is still good to restore the selection.

packages/slate/src/changes/by-key.js
zhujinxuan: I am not sure what you mean by `if (anchorKey == key || focusKey == key)` . 
Do you mean `if (selection.anchorKey !== change.value.selection.anchorKey || selection.focusKey !== change.value.selection.focusKey)`? 
Or Do you mean 
```
if (selection.anchorKey !== change.value.selection.anchorKey) {
  // restore anchorKey
 }
if (selection.focusKey !== change.value.selection.focusKey) {
  // restore focusKey
 }
```
If you means one of these two, I would prefer the second one~

packages/slate/src/changes/by-key.js
zhujinxuan: Fixed~

packages/slate/src/changes/by-key.js
ianstormtaylor: I agree with you there, I like your solution of doing the `Math.min`.

I think there's another case for this:

```js
<block key="a">
  fo<cursor/>o
</block>

<block key="b">
  ba<cursor/>r
</block>
```

Where the `replaceNodeByKey` is called with a new node that doesn't even share the same key as the previous node. It would be nice if the selection-porting logic preserved that one as well for consistency I think.

zhujinxuan: I think the `bar` does not have the same Text Key with the `foo`; the anchorKey and focusKey are textKeys gotten from the selection before `removeNodeByKey` happens.  Therefore, I think it seems it does not matter at this case?

zhujinxuan: If the text key is not shared, then I think: 
```
+    const anchorText = newNode.getDescendant(anchorKey)
+    if (anchorText) {
```
will prevent the selection restore.  And I think that is perhaps a reasonable logic.

ianstormtaylor: @zhujinxuan I think if you're calling `replaceNodeByKey` you're most likely going to want the selection to be preserved in the same way. I believe it will be a lot easier for people to choose to collapse/move than to have to implement preserving it themselves.

zhujinxuan: @ianstormtaylor Do you mean that
```
<block key="a">
<text key="text1">Some<cursor/> </text>
</block>
```
should be converted to 
```
<block key="b">
<text key="text2">Some<cursor/></text>
</block>
```
even the two text keys "text1" and "text2" are different?

I can do it with an optional choice (default turn on preserving by themselves, but allow to disable the preserving if different key )

zhujinxuan:
@ianstormtaylor Do you mean you should restore the selection even with different text key.  If the text key is different, the logic is not very intuitive.  I add an option (by default turn on) to restore the selection with different text keys.  However, if by path there is no text node, then I think trying to restore selection by the child or elsewhere is a bit dangerous.

zhujinxuan:
@ianstormtaylor any idea about the options?  I think, at current commit, the option names/interface for restoring selection is very ugly...

ianstormtaylor:
@zhujinxuan yup, I mean that if the node is replaced with one with a new key we should preserve the selection still. This is the most reasonable thing I think for a method called `replaceNodeByKey`—and it's easy for a user to opt-out by changing the selection themselves if they want, but it's hard for them to implement that logic.

The current PR is very messy I agree. I'd like for the `cursorPreserve` option to be removed, and for that to just always happen.

I can't really tell when it got messy, I seem to remember it was less messy before. (I think you might have squashed the commits which removes the history? This makes it harder to see what happened in the PR over time, so I'd recommend not going it. Some maintainers used to be strict about requiring users to do this, but since GitHub lets us squash in the UI now I personally prefer not to squash manually.)

zhujinxuan:
@ianstormtaylor Thank you.  I will not squash commits in future updates

The major reason for the messiness is the `cursorPreserve` option.  After removal, the code is much cleaner.  

I also add a specialization for the inline node, though not sure whether we shall consider the nested inline. (For example, someone put a link in the old node, then wrap the link with comment in new node)

Mayeu:
Hello :wave:

I was trying to solve this today before finding this PR. Any ETA for the merge to happen?

Thank you for your hard work on this :pray:

ianstormtaylor:
@zhujinxuan I'll review this once the code looks like the existing code in Slate. Right now it's really crazy and hard to read, and very hard to reason about what's going on.

zhujinxuan:
@ianstormtaylor How about this version of replaceNodeByKey?  I drop some crazy logic about inlines and now it seems much more clearer.

zhujinxuan:
Since delete and insert is O(log(N)), is worthwhile to provide replace_node operation to speed up replaceNodeByKey and some other operations.  I will do it in another PR

> Since delete() re-indexes values, it produces a complete copy, which has O(N) complexity.
> Since insert() re-indexes values, it produces a complete copy, which has O(N) complexity.
> Since splice() re-indexes values, it produces a complete copy, which has O(N) complexity.
https://facebook.github.io/immutable-js/docs/#/List/delete

Issue #1624, opened 2018-02-09
Paste on a hanging Selection cause an error:


zhujinxuan:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

If we copy anything, then select a hanging block by triple click, then `cmd+v`. We see an error not find a key 

![](http://g.recordit.co/U8PfB8dlz6.gif)

I guess the problem is due to deleteAtRange removes the startText. A naive solution is to move the startText with empty text content to the next block.

zhujinxuan:
Code becomes much uglier by the fix than I thought considering the `normalize`

@ianstormtaylor Would you mind me rewrite the deleteAtRange function? 

My personal solution for delete for reference: 
https://github.com/zhujinxuan/slate-bind-copy-paste/blob/master/lib/deleteAtRange/rules/atDifferentText.js
https://github.com/zhujinxuan/slate-bind-copy-paste/blob/master/lib/deleteAtRange/rules/atSameText.js
It will works in this way:
1. delete all nodes between startText and endText
2. then decide how to deal with startText and endText

zhujinxuan:
Could I know what the `OT user cases` are?

I will provide two PRs for the `deleteAtRange` (rewrite or a small ugly fix), you can pick up the one you think it is fine.

zhujinxuan:
@ianstormtaylor   I have submitted two PRs to fix the problem:  selecting hanging block and paste would put an error that `startKey` is not found

In the PR#1625, the solution is to add startText when isHanging, but I cannot permit that whether it is possible that 
In the PR#1634, the solution is about, under a rewriting of `deleteAtRange`, we put the fix `startKey` existence by the end of the function. 

Personally, I would prefer PR#1634, because, when more bugs of copy/paste is found, the PR#1634 is much easier to be changed than PR#1625

PR #1625, opened 2018-02-09
Fix insertAtRange on Hanging Selections

zhujinxuan:
This aims to fix the problem that `cmd+v` on a hanging selection, because the startBlock and startText is removed, the `insertFragmentAtRange` find no place to add. (Explained in https://github.com/ianstormtaylor/slate/issues/1624)

A problem of this PR is that, the normalization part is very very ugly.  I will provide a neater `deleteAtRange` in another PR.

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -3,6 +3,7 @@ import logger from 'slate-dev-logger'
 
 import Block from '../models/block'
 import Inline from '../models/inline'
+import Text from '../models/text'
 import Mark from '../models/mark'
 import Node from '../models/node'
 import String from '../utils/string'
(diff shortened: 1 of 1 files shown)


zhujinxuan:
I submit the neater PR on fixing the same problem: https://github.com/ianstormtaylor/slate/pull/1634.

zhujinxuan:
@ianstormtaylor How do you think about this PR?  This PR is small change only to fix the missing startKey in paste at a hanging selection.

ianstormtaylor:
@zhujinxuan I'm open to it. I think as currently written this is very confusing though, and no one is going to be able to remember why these new lines of code are in there when reading through the code next time they go to read it. So I'd like to see more clear variable names and a comment explaining what logic this is handling.

zhujinxuan:
@ianstormtaylor I make some annotation and variable name fixes, and hope the last file changes looks much better.

PR #1626, opened 2018-02-09
Change Validation for Plugins

pvande:
From Slack:

> ianstormtaylor: @pvande but your point still stands, trying to prevent all possible merges/deletes is actually impossible purely from the event handler layer, when you consider that any other plugin could introduce new key bindings for the same behaviors and you’d never know it

> ianstormtaylor: I’m not sure if there’s a nicely-abstracted but still simple solution to that issue, but it doesn’t hurt to try to think one up. I’d imagine it as a layer right underneath the event handlers, actually one that feels similar to the current change methods

> ianstormtaylor: @pvande I’d be open to a PR for `validateChange` if you want to make one

During that conversation, I proposed a strawman based on a plugins providing an `onChange` function that could "rewrite" changes on the fly, and abort application of a change when that change would cause problems.  This strawman had at least two flaws:

* It was made without awareness that plugins could already supply an `onChange` handler.
* It was based on the assumption that Changes were immutable.

The conversation concluded after having discussed Schema-driven solutions, changes to the Change class, and a `validateNode` analog for Changes.

After looking through the code and shaking free a few misunderstandings, I realized that the `validateChange` function I was pursuing was already perfectly handled by the existing `onChange` functionality:

* `onChange` is already called on every change, and the resulting value is consumed.
* `onChange` leverages the same Stack handling as other callbacks, which can easily be short-circuited.
* Changes are mutable, so it's trivial to make additional changes to them.
* "Rewrites" are just a reset + additional changes.
* "Aborts" are just a reset + a non-null return.

The only substantial difference between the `validateChange` interface and the `onChange` interface is that `validateChange` (as discussed) would only be invoked for Changes with Operations.  Considering that we cannot easily deprecate the `onChange` interface, and adding `validateChange` would necessitate _at minimum_ walking through the plugin stack an additional time for every operational Change, it seemed reasonable to fall back to `onChange` as the appropriate place for this.

As discussed earlier, the biggest piece missing from this solution was the ability to "rollback" the Change object to its creation state.  The first commit in this PR handles this case by adding a `reset` method to instances of Change.  I chose to implement this by closing over the constructor arguments, ensuring that the reset state was immutable, and keeping the object free of bookkeeping data (though the original value is still exposed, for convenience).  The tradeoff is that the constructor is marginally more complicated, and populates a method in a less common way; I'm certainly willing to discuss the merits of this approach.

The second piece that was strongly desirable for this feature was a way to make snap decisions about whether to further validate a change.  Walking through every change's list of operations inside (potentially) every plugin sounds like a recipe for disaster, so the second commit in this PR exposes metadata about the Change.  This is simultaneously an area where I feel like I may have gone too far, and yet not far enough – Changes absolutely _must_ be responsible for handling bookkeeping about their contents, and there's so much more that would be useful to track (e.g. separate Operation lists for each type and/or subject), but I'm not sure whether the proliferation of computed "convenience" properties will be welcome.

### "Demonstration"

The existing Table example has an `onKeyDown` handler for `Backspace`, `Delete`, and `Enter`, in order to prevent table destruction through backwards- and forwards-deletions and Slate's default Block splitting functionality.  This handler, however, doesn't block the Apple-specific `Ctrl+D` binding in core, nor does it handle the automatic erasure when typing with a selection – both of which break the demo.

While JSFiddle obviously doesn't have this PR's code, [the example here](https://jsfiddle.net/0onkq9bg/) _does_ demonstrate the kind and volume of changes required to implement a substantially more robust protection using this PR's changes.  (I've opted to not include changes to the example in this PR, for easier consideration.)

This example, while simple, also showcases a couple of the remaining pain points (and likely my own lack of expertise with Slate).  In particular, Operations are still a bit tricky / verbose to work with – pre-filtered Operation lists (e.g. `change.moveNodeOperations`) could go a long way to helping with this.

### DIFF
--- a/packages/slate/src/models/change.js
+++ b/packages/slate/src/models/change.js
@@ -147,23 +147,22 @@ class Change {
   }
 
   get changesText() {
-    return this.insertsText ||
-           this.removesText
+    return this.insertsText || this.removesText
   }
 
   get changesMarks() {
-    return this.addsMarks ||
-           this.removesMarks ||
-           this.setsMarks
+    return this.addsMarks || this.removesMarks || this.setsMarks
   }
 
   get changesNodes() {
-    return this.insertsNodes ||
-           this.mergesNodes ||
-           this.movesNodes ||
-           this.removesNodes ||
-           this.setsNodes ||
-           this.splitsNodes
+    return (
+      this.insertsNodes ||
+      this.mergesNodes ||
+      this.movesNodes ||
+      this.removesNodes ||
+      this.setsNodes ||
+      this.splitsNodes
+    )
   }
 
   /**
--- a/packages/slate/src/models/change.js
+++ b/packages/slate/src/models/change.js
@@ -44,12 +44,18 @@ class Change {
 
   constructor(attrs) {
     const { value } = attrs
-    this.value = value
-    this.operations = new List()
-    this.flags = {
-      normalize: true,
-      ...pick(attrs, ['merge', 'save', 'normalize']),
+
+    this.reset = () => {
+      this.original = value
+      this.value = value
+      this.operations = new List()
+      this.flags = {
+        normalize: true,
+        ...pick(attrs, ['merge', 'save', 'normalize']),
+      }
     }
+
+    this.reset()
   }
 
   /**
--- a/packages/slate/src/models/change.js
+++ b/packages/slate/src/models/change.js
@@ -53,6 +53,24 @@ class Change {
         normalize: true,
         ...pick(attrs, ['merge', 'save', 'normalize']),
       }
+
+      this.metadata = {
+        operations: {
+          insert_text: false,
+          remove_text: false,
+          add_mark: false,
+          remove_mark: false,
+          set_mark: false,
+          insert_node: false,
+          merge_node: false,
+          move_node: false,
+          remove_node: false,
+          set_node: false,
+          split_node: false,
+          set_selection: false,
+          set_value: false,
+        },
+      }
     }
 
     this.reset()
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @pvande, thank you for the super detailed and well thought out PR!

The `reset` method looks great to me, I like the closure approach for keeping it simple. And even if we did eventually want `validateChange` for who knows what reason, we'd probably want `reset` anyways, so this is great!

I'd like to keep the metadata a bit simpler, and expose less of an API that we have to commit to. Instead, I think I'd like to do it as:

```js
change.hasOperation('insert_text')
```

Which we can model internally in a private `change.__hasOperation.insert_text = true` fashion, but that way we don't have to commit to any sort of `metadata` object API. And if we wanted to in the future we could change it to just loop the `operations` array with slightly worse performance without changing the API itself.

In addition to that, I'd like to remove the `changesMarks, etc.` getters. Instead, I think it would be reasonable to let `hasOperation` take (hardcoded, for performance/simplicity) globs, like:

```js
change.hasOperation('*_text')
```

It seems like that handles most all of the cases well enough.

Let me know what you think.

pvande:
I agree with all of this.  I'll get a change wrangled together a little later this weekend!

pvande:
@ianstormtaylor This latest change strips the getters in favor of an `operationsByType` hash of Lists.

In terms of API commitment, this takes us from…

* An ill-specified `metadata` hash.
* One-off predicate getters for each operation type.
* One-off predicate getters for some operation group.

… to …

* An `operationsByType` hash with Lists for…
  * Operation type.
  * Operation group as a glob (i.e. `'*_node').

The predicate behaviors can be replicated by calling `isEmpty()` on the associated List, and the most common List filtering cases have ready access to pre-filtered data.  This also implies that for many common uses, the predicates are now unnecessary (since they'd often only be short-circuiting effectively zero-cost property lookups and an iteration over an empty list).

The plugin described in my last JSFiddle now shrinks to just this:

``` javascript
{
  onChange(change) {
    const originalDoc = change.original.document
    const nodeOps = change.operationsByType['*_node']
    const changedTableCell = nodeOps.some(op => {
      const node = originalDoc.getDescendantAtPath(op.path)
      return node && node.type === 'table-cell'
    })

    if (changedTableCell) {
      return change.reset()
    }
  }
}
```

While the indirect interface discussed previously (e.g. `operationsByType(name) { return this.__operationsByType[name] }`) would give us a little more flexibility in the implementation, I'm not sure it would buy us anything we want: lazy computation would make simple plugins like the one written above far more expensive, more complex `name` parsing does likewise, and the biggest cost we are currently biting off for the feature is a couple additional Immutable.List pushes per operation (which is a cost that seems highly unlikely to grow much).  Even so, let me know if you'd like to see me change it.

ianstormtaylor:
Hey @pvande thanks for the update.

I'd like to see it implemented as the `hasOperation(type)` (or better yet `hasOperationType(type)`) from above. It keeps the API smaller still, and is more memorable IMO when working with it. We can always expand to expose more information about which operations are in the list if needed, but not until we really need to.

I don't want to commit to a more complex object API that is pre-computed to this, because it's still an experiment and it feels pretty edge case.

ianstormtaylor:

PR #1628, opened 2018-02-12
Fix skip this in getFurtherstLonelyAncestor

zhujinxuan:
node.getFurtherstLonelyAncestor intends to skip `this` in the return, but the code failed to do this because
1. `skipLast()`  is equal to `skipLast(0)` which skips nothing
2. `this` is in the start of ancestor, and therefore it should be `skip(1)` or `shift()`

This PR intends to fix that bug, with a text to verify.


ianstormtaylor:
Hey @zhujinxuan please check out the way the other tests are setup, and change the test in PR to match them. It looks like many things (including the file name) are inconsistent.

zhujinxuan:
@ianstormtaylor Fixed~ . Thank you.

packages/slate/test/models/by-key/get-furthest-only-child.js
```diff
@@ -0,0 +1,20 @@
...
+)
+const pNode = document.nodes.first()
+const textNode = pNode.getFirstText()
+
+export const actual = [
+  document.getFurthestOnlyChildAncestor(pNode.key),
+  pNode.getFurthestOnlyChildAncestor(textNode.key),
+  document.getFurthestOnlyChildAncestor(textNode.key),
+]
+export const expected = [null, null, pNode]
```
ianstormtaylor: Instead of structuring this as `actual` and `expected`, can you change it to import the `assert` module directly and expose a single `test` function instead that does the assertions directly?

zhujinxuan: @ianstormtaylor  Fixed~ Thank you.

packages/slate/test/models/by-key/get-furthest-only-child.js
```diff
@@ -0,0 +1,22 @@
+/** @jsx h */
```
ianstormtaylor: Can you change this file to be `/node/get...` instead of `/by-key/get...`

zhujinxuan:
Fixed about the test path~

ianstormtaylor:
@zhujinxuan I think you force-pushed over the other commit that I made cleaning up the logic.

zhujinxuan:
@ianstormtaylor Lol, it seems that I did not notice you make the commit.  Can I know on which branch your commits are?

zhujinxuan:
Thank you a lot.  I have pull the two new commits to my local repo.

ianstormtaylor:
Awesome, thank you very much @zhujinxuan!

PR #1630, opened 2018-02-13
Remove redundant memoization

julienp:
This implements some of suggestions from https://github.com/ianstormtaylor/slate/issues/1402

- Remove memoization from node.getKeys

- Remove memoization from methods that are directly backed by AsArray variants: In addition to the methods listed in https://github.com/ianstormtaylor/slate/issues/1402#issuecomment-362035319, I removed memoization from 
	- getBlocks
	- getBlocksByType
	- getCharacters
	- getInlines
	- getOrderedMarks
	- getOrderedMarksAtRange
	- getOrderedMarksByType
	- getMarksByType

I also think hasChild, hasDescendant and hasNode gain very little from memoization at the cost of increased memory, since they are thin wrappers around the memoized get-variants.


zhujinxuan:
I am thinking that we perhaps shall remove the memorization of `getPath`. Because someone may
```
const path = document.getPath(key)
path.pop()
```

julienp:
It might be worth clarifying which methods are private and which are public. At least the *AsArray methods are not in the docs and it’s somewhat clear that they are an internal optimization. getPath is not documented either, so I would assume it’s a private method? 
In any case it does indeed seem slightly dangerous to memoize functions that return an array that can be modified in place.

zhujinxuan:
A potential problem:
ancestor.getXXXAsArray takes advantage of the cache of child.getXXXAsArray.  If we remove that cache, the getXXXAsArray is slow, and so is the getXXX

I think we should remove the getXXXAsArray and only use getXXX to save space and keep the cache advantage still.

ianstormtaylor:
Thank you very much for doing this @julienp! Looks good.

PR #1631, opened 2018-02-13
Use pageXOffset instead of scrollX in hover menu example

tomasz-sodzawiczny:
`scrollX`/`scrollY` do not work on some browsers, using `pageXOffset`/`pageYOffset` instead: https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollX#Notes.

This makes the hovering menu example work on IE11:
screen shot 2018-02-13 at 16 18 30

Before:
screen shot 2018-02-13 at 16 19 35


ianstormtaylor:
Awesome, thanks @tomasz-sodzawiczny!

PR #1632, opened 2018-02-13
[Feature] Semantic `isStartByKey` and `isEndByKey`

zhujinxuan:
Considering the empty text nodes surrounding the inlines, or some un-normalized empty text nodes, it is helpful to provide two function that 
```
 <paragraph key=p1>
    <text />
    <inline key=i2> .... </inline>
  </paragraph>
```

Because of the empty text nodes surrounding the inline. we cannot do it like `p1.getFirstText() === i2.getFirstText()`.  This PR provides a convenient way to check whether the inline is at the "semantic" start of the paragraph like
```
document.getNode('p1').isStartByKey('p2') === true
```

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1775,6 +1775,50 @@ class Node {
   }
 
   /**
+   * Check whether the node is start by another node, considering the empty text before and after inlines
+   * @param {string} key
+   * @returns {Boolean}
+   */
+
+  isStartByKey(key) {
+    const child = this.getDescendant(key)
+    if (!child) return false
+    if (child.object !== 'text') {
+      return this.isStartByKey(child.getFirstText().key)
+    }
+    if (child === this.getFirstText()) return true
+    const firstValid = this.nodes.find(
+      n => n.object !== 'text' || n.text.length > 0 || n.key === key
+    )
+    if (!firstValid) return false
+    if (firstValid.key === key) return true
+    if (firstValid.object === 'text') return false
+    return firstValid.isStartByKey(key)
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @zhujinxuan, this seems like a very specific and hard to reason about method. I'm not really sure where this is used, and why it would be so highly used to be added to core.

zhujinxuan:
It is used in this case:
```
<paragraph key = 'p1'>
  <link><anchor/>Some Text</link>
  Some middle text
  <link>Some Text<focus/></link>
</paragraph>
```

Due to the normalization, it is indeed
```
<paragraph key = 'p1'>
  <text/>
  <link><anchor/>Some Text</link>
  Some middle text
  <link>Some Text<focus/></link>
  <text/>
</paragraph>
```

Then we cannot have it like
```
expect(document.getNode('p1').getFirstText().key === startKey) // Error
expect(document.getNode('p1').isStartByKey(startKey))//True
```

ianstormtaylor:
Yeah but where are we needing to do...

```js
expect(document.getNode('p1').getFirstText().key === startKey) // Error
expect(document.getNode('p1').isStartByKey(startKey))//True
```

...and how often for it to merit being in core?

zhujinxuan:
@ianstormtaylor   I think it is designed for simplify some special operations in the start/end of a block.  For example, disable `backspace` in a table_cell and the table_cell can have links or emoji inside.

ianstormtaylor:
Closing for the reasons described in https://github.com/ianstormtaylor/slate/pull/1634#issuecomment-367534527, but I'd be happy to reconsider fresh PR's that are small and well-explained.

PR #1633, opened 2018-02-13
[Feature] change.deleteNodesBetween: a change function to delete between nodes

zhujinxuan:
This change function is designed to

1. simplify the deleteAtRange logic, a pre-procedure for rewriting the deleteAtRange to make that function
2. simplify the common process deleting a lot of nodes like that n.nodes.slice(begin, end).forEach(c => change.removeNodeByKey(c.key, {normalize: false})
3. Provides another method for plugin developers, rather than `deleteAtRange`, to delete a lot of nodes without considering about hanging and snapshot.

This PR depends on https://github.com/ianstormtaylor/slate/pull/1632

### DIFF
--- a/packages/slate/src/changes/by-key.js
+++ b/packages/slate/src/changes/by-key.js
@@ -347,6 +347,129 @@ Changes.removeNodeByKey = (change, key, options = {}) => {
 }
 
 /**
+ * Delete Nodes between startKey and endKey, including nodes of startKey and endKey
+ *
+ * @param {Change} change
+ * @param {string} startKey
+ * @param {string} endKey
+ * @param {Object} options
+ *   @property {Boolean} normalize
+ */
+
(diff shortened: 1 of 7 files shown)


ianstormtaylor:
Hey @zhujinxuan, thanks for taking the time to do this. I like the idea of adding more `delete*` change methods, since it seems like a good way to encapsulate some of the complexity in `delete{,Forwards,Backwards}AtRange` in a more reusable way.

That said, this pull request is really hard to read... Because the code style / structure doesn't match any of the existing patterns for Slate. And because the method names are not clear at all as to what the mean.

I'm thankful for you submitting this PR in the first place. I'd be happy to accept one that added a single, small `delete*` helper and explicitly showed where the existing methods could call it instead, resulting in simpler code.

Right now it seems like this adds a new method that none of the core methods depend on, just purely increasing the API surface/complexity.

PR #1634, opened 2018-02-13
[Improvement & Bug Fix]More modularized change.deleteAtRange and fix problem of pasting on a hanging selection

zhujinxuan:
This PR aims to:
1. fix the problem that paste in hanging selection or other things cause on error  (Issue: https://github.com/ianstormtaylor/slate/issues/1624).  The method is, when the normalize is false, try everything possible to keep the startText.
2. delete on hanging selected block will lose selection
3. modularize the `change.deleteAtRange`, makes it easier to be modified for future debugging

This PR depends on the PR of https://github.com/ianstormtaylor/slate/pull/1633 and of https://github.com/ianstormtaylor/slate/pull/1632

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -1,6 +1,7 @@
 import { List } from 'immutable'
 
 import Block from '../models/block'
+import Text from '../models/text'
 import Inline from '../models/inline'
 import Mark from '../models/mark'
 import Node from '../models/node'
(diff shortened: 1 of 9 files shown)


ianstormtaylor:
@zhujinxuan please, don't make PR's depend on other PR's. It makes it super, super complicated to figure out what's going on. And then if I don't like something in one of the PR's, it's super complicated to figure out what is from where and what depends on what.

I appreciate you taking the time to write these PR's, and I think your intuition for splitting up the `deleteAtRange` method is good, but I can't accept any of these as-is because they are very hard to reason about and are going to add complexity. I'd be happy to reconsider fresh PR's that don't have these issues. I hope that makes sense, and thank you very much.

Issue #1636, opened 2018-02-15
The Plugins example is broken


tashburn:
http://slatejs.org/#/plugins

conorcussell:
Its caused by an old version of `slate-soft-break` and `slate-collapse-on-escape` being bundled. PR on its way.

PR #1637, opened 2018-02-15
Update plugins used in plugins example

conorcussell:
Fixes https://github.com/ianstormtaylor/slate/issues/1636

- updates `slate-collapse-on-escape` and `slate-soft-break` plugins to new version


ianstormtaylor:
Thanks @conorcussell!

PR #1642, opened 2018-02-16
Set up webpack configuration for building examples

schneidmaster:
Use webpack to build examples in development and prod. Faster rebuilds and a nicer dev experience.


examples/App.js
```diff
@@ -0,0 +1,135 @@
+import React from 'react'
```
ianstormtaylor: Can you change this file to be `app.js` to match the rest?

webpack.config.js
```diff
@@ -0,0 +1,82 @@
+const path = require('path')
```
ianstormtaylor: Can you move this file into `./support` like the others?

schneidmaster:
@ianstormtaylor updated!

ianstormtaylor:
Beautiful, thanks!

ianstormtaylor:
@schneidmaster I think this removes the `index.html` that GitHub pages uses?

schneidmaster:
@ianstormtaylor webpack is now generating its own `index.html` via [HtmlWebpackPlugin](https://github.com/ianstormtaylor/slate/blob/master/support/webpack.config.js#L55-L63) (this makes things easier because webpack can then inject its own script tags with the current hash for the JS/CSS)

schneidmaster:
(You should see this in `examples/dist/index.html` after running the production build)

ianstormtaylor:
@schneidmaster ah yup, you're absolutely right.

PR #1643, opened 2018-02-16
Fix marks of undefined on char.marks

isubasti:
`char` could be `undefined` when accessing `char.marks`, this PR make sure that char exists otherwise `return []` .
ideally, any `operations` should not result in `char` being `undefined`


zhujinxuan:
Please do not delete the 
```
if (!char) return []
```
in the `getInsertMarksAsArray`

Because that is also prepared for the following situation
```
<link>aaaa<anchor/></link><focus/>bbbb
```

ianstormtaylor:
Thanks @isubasti!

Issue #1635, opened 2018-02-14
Newlines aren't preserved when copying


tobiasandersen:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
If I copy text from within the editor, all empty blocks are stripped away. So if I write a paragraph, press enter a few times, then write a new paragraph, ending up with e.g:
```
A line of text in a paragraph.


A line of text in a paragraph.
```

And now copy all that, my `plain/text` clipboard will be only:
```
A line of text in a paragraph.
A line of text in a paragraph.
```

Note that it's only for the `plain/text` clipboard, so pasting inside the slate editor works fine.

To reproduce, copy the content in [this fiddle](https://jsfiddle.net/832672pe/1/), and paste somewhere outside of slate. 

#### What's the expected behavior?
Newlines should be preserved when copying.

tobiasandersen:
We didn't have this problem before (haven't been able to pinpoint exactly when it started to occur, but it worked a few months ago). I can get it to work if I remove [this line in `cloneFragment`](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/clone-fragment.js#L68):

```
zws.forEach(zw => zw.parentNode.removeChild(zw))
```

But i suspect it's useful for other stuff?

ianstormtaylor:
@tobiasandersen sounds like we need to be able to differentiate between zero-width nodes that are there for cross-browser fixing, and those that are there for keeping empty blocks from collapsing? And then instead of removing them all, some of them should be replaced with normal spaces?

tobiasandersen:
Indeed! There's no COMPAT comment above this line though (and no real explanation on what it's for). I can see why it would be useful to do this for void nodes, but not really for anything else. Am I missing something? Do you remember why it was added in the first place? It seems it was added before [this](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/leaf.js#L123-L126):
```
// COMPAT: If the text is empty, it's because it's on the edge of an inline
// void node, so we render a zero-width space so that the selection can be
// inserted next to it still.
if (text == '') return <span data-slate-zero-width>{'\u200B'}</span>
```

ianstormtaylor:
I think we use zero-width spaces in three(?) places:

1. Inside void node's as spacers to catch the cursor.
2. Inside empty text nodes next to inline nodes, to catch the cursor.
[...]

PR #1644, opened 2018-02-16
Replace zero width blocks with newline before copy

blakeembrey:
Add back empty blocks as a newline behaviour by swapping the text out at copy and paste time. It needs to stay as a zero width space because otherwise the browser swaps the `\n` out with nothing when inputting multiple IME characters. This results in Slate blowing up like before, so keeping the zero width behaviour is the simplest approach while using the copy and paste swap to get the plain text behaviour that's desired.

Happy to replace `z` and `n` with something else too, those are just short placeholders. I initially tried with `parent.kind` or `node.kind` as the values, but it resulted in undesirable behaviour because these nodes appear in a lot of places.

Fixes https://github.com/ianstormtaylor/slate/issues/1635.

### DIFF
--- a/packages/slate-react/src/components/leaf.js
+++ b/packages/slate-react/src/components/leaf.js
@@ -116,14 +116,20 @@ class Leaf extends React.Component {
   renderText() {
     const { block, node, parent, text, index, leaves } = this.props
 
+    // COMPAT: If the text is empty and the only child, we need to render a
+    // line break when copying and pasting to support expected plain text.
+    if (text == '' && parent.kind == 'block' && parent.text == '') {
+      return <span data-slate-zero-width="n">{'\u200B'}</span>
+    }
+
     // COMPAT: Render text inside void nodes with a zero-width space.
     // So the node can contain selection but the text is not visible.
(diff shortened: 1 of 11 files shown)


tobiasandersen:
I really like this solution. Rendering a `\n` character for empty blocks is actually problematic even for "normal text editing" (non-IME). Before this PR was created, I tried adding this back to leaf nodes:
```
if (text == '' && parent.object == 'block' && parent.text == '') {
  return <span data-slate-empty-block>{'\n'}</span>
}
```
But saw similar problems as https://github.com/ianstormtaylor/slate/issues/1452 with it.

I've tried this PR (actually already deployed it) and it does solve https://github.com/ianstormtaylor/slate/issues/1635. Haven't seen any other problems with it either. Great work @blakeembrey!

isubasti:
nice implementation, the only downside of this is that slate does not trigger onCopy if it is readOnly.It would be nice if `onSelect` and `onCopy` can work on a readOnly slate document so that it behaves the same way as the read/write document. I will try to get that in another PR sometimes soon unless someone else wants to tackle it.

ianstormtaylor:
Beautiful solution, thank you @blakeembrey!

Issue #1561, opened 2018-01-26
Marks always "bleeding" into new line by default


thomaswhyyou:
#### Do you want to request a *feature* or report a *bug*?

Bug

#### What's the current behavior?

Tested with v0.32.1 on Chrome/Firefox/Safari locally as well as in the examples page (below gif recorded as of today 1/24/2018): It appears that, with given two paragraphs, if there are marks on the ending text of the previous paragraph, they always carry over to the following paragraph by default. 

Below are a couple issues reproduced from this bug, using the examples page.
![zvpylyfylr](https://user-images.githubusercontent.com/4471723/35426216-40179bc4-022f-11e8-8a5d-afcf526b1607.gif)
(http://recordit.co/zvPYlyFYlr)

* Styling the end of the text with bold, then hitting enter to hard break leads the following paragraph to be bolded by default as well. I did search through the previous issues and from #1269 I see there was some discussion around this behavior; but I think the decision was ultimately not to support the auto carry over and let users to decide for themselves (#1273). Given this, it leads me to believe this is a bug.
* Another unintuitive behavior is when you manipulate the previous paragraph (by deleting `!` character), it automatically affects the following paragraph even though the edit operation was not related and done only to the first paragraph. I imagine this is rather unexpected for most use cases. (I'm not certain, but back in the days Slate used to use zero-width space to deal with empty text and such, but I don't think the latest version does that anymore- maybe related?)

Omitting JSFiddle, because reproducible with the example page.

#### What's the expected behavior?

* Hard breaking on text with marks does not automatically carry over the style to the following split paragraph
* Editing marks/style of a paragraph in isolation should not affect the following paragraph.

I took a quick look at `splitBlock` and other associated internals but I must admit, I was pretty lost in it 😅.. Really appreciate Slate and the work everyone has put in - let me know in any ways I can help.

dmitrizzle:
I'd actually expect marks to carry over into next paragraph. Sometimes it's undesirable but that's what I would expect and that's what most editors seem to do.

thomaswhyyou:
I don't personally have too strong of opinions as to what the correct behavior here should be, since it is subjective and dependent on one's situation.. I'd add, from my observations, more consumer oriented products like Dropbox Paper and Notation tend to not carry over marks/styles, but If this is the agreed behavior for Slate, I can live with this. 

[...]

PR #1645, opened 2018-02-16
Fixing the bug that getMarks steals mark from the previous block

zhujinxuan:
Fix https://github.com/ianstormtaylor/slate/issues/1561 that when the end of previous block is marked, then cursor at the block start is also marked.

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1130,6 +1130,11 @@ class Node {
     if (startOffset == 0) {
       const previous = this.getPreviousText(startKey)
       if (!previous || previous.text.length == 0) return []
+      if (
+        this.getClosestBlock(startKey) !== this.getClosestBlock(previous.key)
+      ) {
+        return []
+      }
       const char = previous.characters.get(previous.text.length - 1)
       return char.marks.toArray()
     }


zhujinxuan:
TODO Fix: set mark when splitBlock at current range.

Thanks @CameronAckermanSEL

CameronAckermanSEL:
Double/Triple clicking also applies the mark to the next block, but this might be deliberate? Unsure. It's not really a desirable behavior, but it might be consistent with google docs and draft.

ianstormtaylor:
Thank you very much @zhujinxuan!

zhujinxuan:
@ianstormtaylor @CameronAckermanSEL said about a related case:
```
<paragraph>
xxx<b>BoldText</b><cursor/>YYY
</paragraph>
```
Then press 'Enter' to splitBlock and press a to insert 'a';
Then the document becomes 

```
<paragraph>
xxx<b>BoldText</b>
</paragraph>
<paragraph>
a<cursor/>YYY
</paragraph>
```
,while in the unfixed getMark:
```
<paragraph>
xxx<b>BoldText</b>
</paragraph>
<paragraph>
<b>a</b><cursor/>YYY
</paragraph>
```

Whch one do you think it is better? the first one '`a` without bold' or the second '`<b>a</b>`'?  If you prefer the second one, I can make a PR on splitBlock to set marks in the selection.

Issue #1657, opened 2018-02-21
Not every node affected by a change is revalidated


marian-c:
#### Do you want to request a _feature_ or report a _bug_?
BUG

#### What's the current behavior?

When a change affects multiple adjacent inlines, validateNode prop only triggers for one of them.
To be more specific: adjacent inlines in a common immediate parent.

Steps to reproduce:
1: select text spanning 2 inlines
2: press the space bar
3: observe the console

Note that the problem seems to be more general, when the inlines don't have the same parent, the validateNode prop gets triggered on both of them, but then only on one parent.

```
├─ slate-react@0.11.4
└─ slate@0.32.4
```
http://recordit.co/pUupvM4Xh9
https://jsfiddle.net/fj9dvhom/141/

#### What's the expected behavior?

Every node that was affected by the change should be re-validated.

Edit:
It's worth mentioning that if instead of the space key, you press the delete key (mac) then the expected behavior happens, all nodes affected are re-validated

zhujinxuan:
I think it is a bug existing in both insertTextAtRange and insertFragmentAtRange;

The deleteAtRang shall normalize the commonAcestor of startKey and endKey. However, insertFragmentAtRange normalize the parent of startBlock, and insertTextAtRange normalize the parent of startKey.

PR #1659, opened 2018-02-21
Fix normalize range of the insertTextAtRange

zhujinxuan:
`insertTextAtRange` normalize the parent of startText, regardless of where the endText is.  (https://github.com/ianstormtaylor/slate/issues/1657) . This PR fixes the normalize range to normalize all affected nodes.

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -896,8 +896,19 @@ Changes.insertTextAtRange = (change, range, text, marks, options = {}) => {
   if (normalize !== undefined) {
     normalize = range.isExpanded
   }
+  change.insertTextByKey(key, offset, text, marks, { normalize: false })
 
-  change.insertTextByKey(key, offset, text, marks, { normalize })
+  if (normalize) {
+    // normalize in the narrowest existing block that originally contains startKey and endKey
+    const commonAncestor = document.getCommonAncestor(startKey, range.endKey)
+    const ancestors = document
+      .getAncestors(commonAncestor.key)
+      .push(commonAncestor)
+    const normalizeAncestor = ancestors.findLast(n =>
+      change.value.document.getDescendant(n.key)
+    )
+    change.normalizeNodeByKey(normalizeAncestor.key)
+  }
 }
 
 /**


ianstormtaylor:
@zhujinxuan can you please read the rest of the code in Slate and fix your pull requests (this one and others) to follow its conventions. In this case, your change is one huge block of code with no line breaks, whereas most every block `if` statement in Slate is broken up by line breaks.

zhujinxuan:
@ianstormtaylor I make some fixes.  Do you mean add line breaks and annotation like that?

packages/slate/src/changes/at-range.js
```diff
@@ -896,8 +896,27 @@ Changes.insertTextAtRange = (change, range, text, marks, options = {}) => {
...
+      .push(commonAncestor)
+    const normalizeAncestor = ancestors.findLast(n =>
+      change.value.document.getDescendant(n.key)
+    )
+    if (normalizeAncestor) {
+      change.normalizeNodeByKey(normalizeAncestor.key)
+      return
+    }
+
+    // If for some strange reason, the commonAncestor of startKey and endKey does not
```
ianstormtaylor: When does this "strange reason" occur? Can you change this comment to include `(eg. when ` or whatever explaining when this might be the case.

ianstormtaylor: (Or actually can you just add a single comment that explains this whole piece of logic clearly and concisely.)

zhujinxuan: @ianstormtaylor I am not sure about one thing: If the document includes one block and the block is void.  Will the document.key be regenerated after deleting the void block?

zhujinxuan: I read about the `deleteAtRange`.  The document.key is always in the ancestors and should be unchanged even with the scenario of select-all(document has only only void blocks) and delete all. Then I delete that line and fix with comment

zhujinxuan:
Fixed~

zhujinxuan:
Is there something I should do about this PR?

ianstormtaylor:
Thanks @zhujinxuan!

PR #1666, opened 2018-02-22
Use source-map-loader for more fine-grained sourcemaps.

schneidmaster:
Currently when an error happens, the console log looks like this:

screenshot 2018-02-21 21 29 47

Note the `slate.es.js:3166` on the right. Not too bad, it's the unminified ES build of Slate so we can read it pretty well and see what's going on. But it's all of Slate rolled up together, so it takes some hunting to figure out which original source file it came from. Also, if you click on the source, it looks like this:

screenshot 2018-02-21 22 12 59

which isn't exactly what the source looks like -- it's the source after it's been run through babel (so for example we get the `key/value` thing instead of an ES6 class).

There's a webpack loader, [source-map-loader](https://github.com/webpack-contrib/source-map-loader), which can help us do better. It consumes any sourcemaps in your "source" JS (in our case, the built `slate.es.js`) and rolls them together into the final sourcemaps output by your webpack build. So the above console log changes to:

screenshot 2018-02-21 21 39 21

where `node.js` is the actual original filename. And clicking on the line gives this:

screenshot 2018-02-21 22 13 58

so we get the actual original code (in this case, an ES6 class).

To support this change, I changed the rollup build to always output sourcemaps for the module variants, not just when in development. (I figured we might as well ship the sourcemaps on npm so that end users can use `source-map-loader` in the same fashion if they'd like.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "slate-collapse-on-escape": "^0.6.0",
     "slate-soft-break": "^0.6.0",
     "slate-sugar": "^0.6.1",
+    "source-map-loader": "^0.2.3",
     "source-map-support": "^0.4.0",
     "style-loader": "^0.20.2",
     "to-camel-case": "^1.0.0",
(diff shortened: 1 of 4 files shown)


ianstormtaylor:
Thanks @schneidmaster this is awesome!

PR #1667, opened 2018-02-22
Ignore the build file in examples/

zhujinxuan:
The exmaples/build.{dev, prod}.js are not in .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 build/
 dist/
 lib/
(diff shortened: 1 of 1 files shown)


schneidmaster:
This was an intentional change in #1642. We are not building to `examples/build.{dev, prod}.js` anymore (webpack builds to a `build/` directory in the project root) so I removed them from the `.gitignore` as well to keep it tidy.

PR #1670, opened 2018-02-23
Update change.md to include setValue

thesunny:
Okay, sorry this is a little poor; however, there are some things I'm still not understanding like

- why is {save: false} required
- In the code, I can see it calls Value.createProperties but Value.createProperties is not documented so I can't refer to it.

### DIFF
--- a/docs/reference/slate/change.md
+++ b/docs/reference/slate/change.md
@@ -78,6 +78,15 @@ validateNode(node) {
 }
 ```
 
+## Full Value Change
+
+### `setValue`
+
+`setValue(properties: Object, [options: Object]) => Change` <br/>
+`setValue(value: Value, [options: Object]) => Change`
+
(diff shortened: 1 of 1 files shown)


zhujinxuan:
Because save flag is about whether the operation is written to the redo/undo stack.  However, I am not sure whether it is intended behavior to pass `{save: false}` by function options.  What I am doing right now is alike 
```
change.setOperationFlag('save', false).setValue(...).setOperationFlag('save', true)
```

ianstormtaylor:
The `{ save: false }` shouldn't be recommended. It sounds like there's just a bug in the `setValue` undo logic instead.

thesunny:
Okay, I will hold off on this PR and submit an issue for the `setValue` undo bug. Maybe accept this PR after that bug is fixed?

thesunny:
Here is the issue for fixing the bug https://github.com/ianstormtaylor/slate/issues/1675

Issue #1658, opened 2018-02-21
Keyboard selection won't extend across inlines


marian-c:
#### Do you want to request a _feature_ or report a _bug_?
BUG

#### What's the current behavior?

You cannot keyboard select across inlines.

1: start the document with a paragraph containing two inline nodes
2: put the cursor in the first inline node
3: press `shift` + `->` enough times so that the selection should also move to the next inline.

Note that selecting with the mouse has no issue.

http://recordit.co/JKTVPna6qN
https://jsfiddle.net/fj9dvhom/141/

slate-react 0.11.4
slate 0.32.4
chrome
OS X

#### What's the expected behavior?
Selection also covers the second inline

Soreine:
Is this related ? https://github.com/ianstormtaylor/slate/issues/1502

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #1677, opened 2018-02-25
Fix Shift+Right selection behaviour on empty text nodes

tomasz-sodzawiczny:
Attempt to fix the <kbd>Shift</kbd> + arrows bahaviour. See: https://github.com/ianstormtaylor/slate/issues/1658


ianstormtaylor:
Hey @tomasz-sodzawiczny it's hard for me to understand the context here. Can you explain what's going on here with the solution? And maybe add a comment for future people reading this code?

PR #1679, opened 2018-02-26
Do not split void node by default

zhujinxuan:
This PR add an option that by default `splitNodeByKey` will not split void blocks; it avoid duplication of void blocks in careless splitting.


ianstormtaylor:
@zhujinxuan what's the reason for this?

zhujinxuan:
The reason is, when we use splitDescendantsByKey , we have to be careful if we do not want to split an image or an emoji.  I think it is perhaps good to have an option to avoid split void nodes

ianstormtaylor:
Can you give a code sample for some use cases for this?

PR #1689, opened 2018-03-06
Handle readOnly inside contentEditable check list

vilic:
The check list example is a very important example for me to understand how to handle similar scenarios. So even it's already working as this example itself is not read-only, I think it still worth pointing out we may need to have the value of `contentEditable` accordingly to the `readOnly` property.


ianstormtaylor:
Sounds good to me, thanks @vilic!

PR #1691, opened 2018-03-07
Export Hotkeys from 'slate-react' as an external package 'slate-hotkeys'

zhujinxuan:
Hi, here is a small PR about getting `HOTKEYS` exported, to simplify plugin development in rewriting key behaviors.


zhujinxuan:
@schneidmaster Hi, do you know how to get umd under `dist/` compiled?  I tried `yarn build`, but `packages/slate-hotkeys/dist` is still empty.

isubasti:
@zhujinxuan need to set `NODE_ENV` to 'production' to build `umd` as well try running `NODE_ENV=production yarn build` that should do the trick

zhujinxuan:
@isubasti Got it.  Thank you~

ianstormtaylor:
I'm not sure I want to export these hotkeys for now, since it just adds API surface. I could be convinced I guess. But doing it as this PR does it doesn't make sense because it's adding a bunch of other exported things like environment checks that shouldn't be grouped together.

zhujinxuan:
How about exports hotkeys from slate-react?
These hotkeys are useful when someone wants to do some tricks about `deleteAtRange`. For example, select a range in a table crossing cells and then press a-z, then this hotkey exports are perhaps useful when they have their own logic of deleteAtRange

ianstormtaylor:
@zhujinxuan can you give a specific example (with code) for when these are useful? (Or multiple examples ideally.)

zhujinxuan:
@ianstormtaylor I will provide that a few days later with a demo about table.

zhujinxuan:
My mistake.  I find I do not need this file to overwrite a-z behaviors.  They are in onBeforeInput

Issue #1275, opened 2017-10-23
beforeInput loose selection and inserting in random place


isubastiCadmus:
#### What's the current behavior?
using browser spell check on rich text example on a text that has mark on chrome seems to loose selection and inserting the text on weird place. there are two behaviours on this depending on where it is performed.

-  text got replaced(looks like its correct) but it is not inside a span and when you try to delete it, it will break
![Text outside span](https://giphy.com/gifs/xT9IgwGh4eqYN0tmE0)
-  possibly related to the first one as well and as it loose the selection when it should technically have a mark, it encounters https://github.com/ianstormtaylor/slate/issues/1271
![cannot read mark of undefined](https://giphy.com/gifs/3o7aD8eg5UUDUL1Sak)

tested on safari and it works fine which means it's a different implementation between ```onNativeBeforeInput and onBeforeInput```

#### What's the expected behavior?

text got replaced while preserving current active marks

isubastiCadmus:
related to https://github.com/ianstormtaylor/slate/issues/909

isubastiCadmus:
@ianstormtaylor is this potentially related to a compatibility issues that affect firefox mentioned in https://github.com/ianstormtaylor/slate/issues/714 which now affect chrome as well?

isubastiCadmus:
@ianstormtaylor this seem to only happens in chrome.
It works fine on safari.
I also tried removing the spellcheck overwrite for firefox and it seems to works fine as well.
Do you remember what's the firefox's compatibility issue was since it might be the same one and it has been fixed on newer version of firefox?

ianstormtaylor:
Hey @isubastiCadmus I'm honestly not sure. Is this the same as #909?

doodlewind:
Seems that the `findPoint` util for deriving native selection returns wrong result, I'm not confident enough for a PR, while I may bring a writeup explaining this.

isubastiCadmus:
@ianstormtaylor not sure if it is the same as #909 but it insert the next outside the span instead of replacing the text(first gif). @doodlewind yes it is either findPoint deriving native selection returns wrong result or chrome change the behaviour of context menu and resulting in slate loosing its selection(onBlur got triggered?)

isubastiCadmus:
@doodlewind any pointers on where I should start if I want to make a PR?

CameronAckermanSEL:
This now crashes in `findDOMRange` instead. http://recordit.co/VdwvtBxDXQ

[...]

PR #1695, opened 2018-03-08
Fix correct-spell problem that crashing the slate

zhujinxuan:
https://github.com/ianstormtaylor/slate/issues/1275 . 
## The description of Problem
Spell check can cause slateJS crash

## Reason and solution of the problem
The original onInput after.js assume the spell check works this way:
```html
<span data-offset-key="a:b"> Cat i cute </span>
```
after input event, before render
```html
<span data-offset-key="a:b"> Cat is cute </span>
```

But with marks, the html content after input event and before React rendering can be
#### spell check inside a mark:
```html
<span data-key="a"><span data-offset-key="a:b"><b>Cat i cute</b></span></span>
<!-- Becomes -->
<span data-key="a"><span data-offset-key="a:b"><b>Cat </b>i <b>cute</b></span></span>
<!-- or -->
<span data-key="a"><span data-offset-key="a:b"><b>Cat </b></span>i <span data-offset-key="a:b"><b>cute</b></span></span>
```
#### spell check replacing a mark:
```html
<span data-key="a"><span data-offset-key="a:0">Cat </span><span data-offset-key=a:1><b>i</b></span><span data-offset-key='a:2'> cute</span></span>
<!-- Becomes -->
<span data-key="a"><span data-offset-key="a:0">Cat </span>is<span data-offset-key='a:2'> cute</span></span>
```
The `span data-offset-key=a:1` is deleted

#### spell check at end of a mark
```html
<span data-key="a"><span data-offset-key="a:0"><b>Cat i</b></span><span data-offset-key='a:1'> cute</span></span>
<!-- Becomes -->
<span data-key="a"><span data-offset-key="a:0"><b>Cat </b></span>is <span data-offset-key='a:1'> cute</span></span>
```
`is` is outside of `span data-offset-key`

#### spell check at begin of a mark
```html
<span data-key="a"><span data-offset-key="a:0">Cat </span><span data-offset-key='a:1'><b>i cute</b></span></span>
<!-- Becomes -->
<span data-key="a"><span data-offset-key="a:0">Cat </span>is <span data-offset-key='a:1'><b>cute</b></span></span>
```
`is` is outside of `span data-offset-key`

4. spell check 

In this PR, I did 
1. write a `findContent` function to get `textContent` and `leaf` for the following scenarios
2. Force remount by regenerate React key, because anchorNode and splitted dom are not controlled by react;
3. remove anchorNode because it is not controlled by react

### DIFF
--- a/packages/slate-react/src/components/leaf.js
+++ b/packages/slate-react/src/components/leaf.js
@@ -71,6 +71,7 @@ class Leaf extends React.Component {
 
   componentWillReceiveProps(props) {
     const { ref, firstChild } = this.leafRefs
+    // check if ref or its all contents are deleted by spell check
     if (!ref || ref.firstChild !== firstChild) {
       this.forceRegeneration()
       return
--- a/packages/slate-react/src/components/leaf.js
+++ b/packages/slate-react/src/components/leaf.js
@@ -71,6 +71,10 @@ class Leaf extends React.Component {
     return false
   }
 
+  setRef = ref => {
+    this.spanRef = ref
+  }
+
   /**
    * Render the leaf.
    *
@@ -86,7 +90,11 @@ class Leaf extends React.Component {
       index,
     })
 
-    return <span data-offset-key={offsetKey}>{this.renderMarks()}</span>
+    return (
+      <span ref={this.setRef} data-offset-key={offsetKey}>
+        {this.renderMarks()}
+      </span>
+    )
   }
 
   /**
--- a/packages/slate-react/src/components/leaf.js
+++ b/packages/slate-react/src/components/leaf.js
@@ -40,7 +40,7 @@ class Leaf extends React.Component {
   constructor(props) {
     super(props)
     this.state = {
-      regenerateNum: 0,
+      regenerateKey: 0,
     }
   }
 
@@ -57,7 +57,7 @@ class Leaf extends React.Component {
 
   forceRegeneration = () => {
     this.setState(state => ({
-      regenerateNum: state.regenerateNum + 1,
+      regenerateKey: state.regenerateKey + 1,
     }))
   }
 
@@ -87,7 +87,7 @@ class Leaf extends React.Component {
     const queryString = `[data-offset-key="${offsetKey}"]`
     if (!window.document.querySelector(queryString)) {
       this.setState(state => ({
-        regenerateNum: state.regenerateNum + 1,
+        regenerateKey: state.regenerateKey + 1,
       }))
       return
     }
@@ -141,7 +141,7 @@ class Leaf extends React.Component {
       key: node.key,
       index,
     })
-    const key = `${offsetKey}:${this.state.regenerateNum}`
+    const key = `${offsetKey}:${this.state.regenerateKey}`
 
     return (
       <span key={key} ref={this.setRef} data-offset-key={offsetKey}>
--- a/packages/slate-react/src/components/leaf.js
+++ b/packages/slate-react/src/components/leaf.js
@@ -84,6 +84,7 @@ class Leaf extends React.Component {
       this.forceRegeneration()
       return
     }
+
     for (let index = 0; index < childNodes.length; index++) {
       const child = childNodes[index]
       if (child.nodeName === '#text') {
@@ -91,6 +92,7 @@ class Leaf extends React.Component {
         return
       }
     }
+
     const { node, index } = this.props
     const offsetKey = OffsetKey.stringify({
       key: node.key,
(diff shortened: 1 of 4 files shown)


zhujinxuan:
Problem: after onInput, and after textContent is gotten.  How to restore to the DOM before the input event.  input will create doms uncontrolled by React

zhujinxuan:
Thank @YurkaninRyan 's idea of forcing re-mount by changing react key. It is close to be solved

zhujinxuan:
I think it is almost solved~

@isubasti  @CameronAckermanSEL Can you have tests on different browsers?

zhujinxuan:
Update: Fixing the problem of nested marks, lol

zhujinxuan:
Done~

zhujinxuan:
This PR does not support marks in part of a word like:
screen shot 2018-03-12 at 1 01 15 pm

I can support it if you think it is necessary.

isubasti:
@zhujinxuan I don't think the browser spellcheck work with that kind of case anyway, it will be treated as misspelled `h` and misspelled `el`by browsers

packages/slate-react/src/components/leaf.js
```diff
@@ -49,20 +55,59 @@ class Leaf extends React.Component {
     debug(message, `${this.props.node.key}-${this.props.index}`, ...args)
   }
 
+  forceRegeneration = () => {
+    this.setState(state => ({
+      regenerateNum: state.regenerateNum + 1,
+    }))
```
CameronAckermanSEL: how is this different than `this.forceUpdate()`?

zhujinxuan: @CameronAckermanSEL Thank you. I just find the re-mount in leaf.js is un-necessary. The remount is necessary in text.js where the dom element of leaf can be removed.

The difference of forceUpdate and set a different key is that, by setting a different key, the react will unmount the current dom without re-using the childNodes.  It is useful in spell check when the childNodes of text.js can be removed without notifying react.

zhujinxuan: My mistake. The renerageKey of leaf.js is useful.

See this image:
screenshot_20180307_235202

under the DOM generated after spell check.  The `try it out` and `help` are not controlled by react.  We have to regenerate key to unmount this whole leaf dom and mount a new leaf dom.

zhujinxuan:
screen shot 2018-03-26 at 4 40 06 pm

I will not support this.

zhujinxuan:
Hi, @ianstormtaylor .  Would you have some time to review this PR on fixing spell checking crash?

ianstormtaylor:
@zhujinxuan I appreciate you taking the time to try to solve this. Your description of the problem in your opening comment is good, and the solution you describe sounds right.

But, I've said this to you before... **follow the existing style of the Slate codebase**. I mean this in terms of basic code styling, and also in terms of how decisions are made to abstract/write functions. I'm not reviewing or merging these kinds of sprawling PR's that add tons of new logic in a complex, poorly commented, high LOC way. If I merged this PR it would reduce the quality of the codebase and make it harder for everyone to contribute.

zhujinxuan:
@ianstormtaylor Thank you. I have a bit problem in understanding the existing code style:
1. what does LOC mean?
2. what do you mean by `how decisions are made to abstract/write functions`

zhujinxuan:
@ianstormtaylor Got it.  I have a way to shorter the part `find-content.js` from 190 lines to 120 lines (though it would becomes a bit harder to understand).  But I do not think I can make it shorter than 120 lines.  

I also feel the change of `leaf.js` and `text.js` are crazy.  Spell check creates remove React-controlled doms, split react-controlled doms.  `forceUpdate` and alike will throw an error in children's unmount, and fail to remove the uncontrolled doms by spell check split.  Then we have to regenerate react key on the leaf or text to remount them without messing up with children.

If the key regeneration is a must. I can only find three ways avoiding duplicating code in leaf.js and text.js.  
1. HOC, but I think the hoc only makes the problem more complex. 
2.  write a customized `findComponentByDom` by collecting all refs recursively, and remount the children if its ref is missing. And I think it is crazily too complex.
(Unfortunately, the findCompoentByDom is under `ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`, an unfriendly position)
3. put all related functions in `components/utils/`. And import them in leaf.js and node.js.  

Do you think whether it is good to have `components/utils/` to fix this bug?

zhujinxuan:
Close this PR because I find the right solution:
add 
`data-text: true` to renderMarks

I will submit the right PR soon.

PR #1698, opened 2018-03-09
Fix normalize empty inline

urugator:
This PR attempts to fix 2 problems related to normalization of empty inlines:

1) [Nested empty inlines](https://github.com/urugator/slate/commit/64e84a14dfc8628c5ec52f22286742d63136de2c) not being removed  by normalization.

2) Since `0.33.0` void nodes can contain empty string `''`. This means that `inline.text === ''` is no longer a valid indicator of "emptiness". 
Consider: `<a><void/></a>` then `a.text === ''` , but `<a>` isn't actually empty and shouldn't be removed by normalization. This PR [changes the way](https://github.com/urugator/slate/commit/ddfb71f68f57a04e057ac00bb4daf6d3b31cb32e) how empty inlines are detected.

### DIFF
--- a/packages/slate/src/constants/core-schema-rules.js
+++ b/packages/slate/src/constants/core-schema-rules.js
@@ -92,7 +92,7 @@ const CORE_SCHEMA_RULES = [
     },
   },
 
- /**
+  /**
    * Ensure that inline non-void nodes are never empty.
    *
    * This rule is applied to all blocks and inlines, because when they contain an empty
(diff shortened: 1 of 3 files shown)


SmilinBrian:
We are facing issues with an essentially identical use case -- thank you for this!

urugator:
yea... unfortunately I am fighting with running liner locally, due to a different line endings...

ianstormtaylor:
Good call, thank you @urugator!

urugator:
@ianstormtaylor Btw I think we should reimplement `node.isEmpty` to perform similar check, instead of being an alias for `text === ''`

ianstormtaylor:
@urugator I'd love a PR that implements that, I totally agree.

PR #1699, opened 2018-03-12
Allow null return values, which will skip serializing that node

justinweiss:
Sometimes, I have nodes that only make sense temporarily, so it doesn't make sense to serialize them. Following the pattern in React, explicitly returning `null` from a serialization rule should result in that node (and its children) not making it into the serialized document.

### DIFF
--- a/packages/slate-html-serializer/src/index.js
+++ b/packages/slate-html-serializer/src/index.js
@@ -325,7 +325,7 @@ class Html {
 
   serialize = (value, options = {}) => {
     const { document } = value
-    const elements = document.nodes.map(this.serializeNode)
+    const elements = document.nodes.map(this.serializeNode).filter(el => el)
     if (options.render === false) return elements
 
     const html = renderToStaticMarkup(<body>{elements}</body>)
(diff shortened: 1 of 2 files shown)


zhujinxuan:
I remember `return []` can skip serializing.

BTW, I think there is a problem about this PR; there are existing codes using null values to skip the step, rather than skip the serializing.

I would suggest use `NaN` to skip serializing, which is to change `filter(el=>el)` to `filter(el=>el===el)`

justinweiss:
`[]` doesn't work during serializing, and most other "like empty" things won't either -- ReactDOM raises a `TypeError: Cannot read property 'toLowerCase' of undefined` when rendering.

>  there are existing codes using null values to skip the step, rather than skip the serializing.

Makes sense. The documentation says:

> The serialize function should return a React element representing the serialized HTML, or nothing if the rule in question doesn't know how to serialize the object, in which case the next rule in the stack will be attempted.

Which I read as "return undefined," but looking more closely, I can definitely see how someone could read that as "return null."

That said, if we're still OK with making breaking changes in Slate, `null` makes the most sense here, just from its parallels with React's `render` method. A close second would be `return false`, for its parallels with event handling. `NaN` doesn't make sense to me in this scenario, other than as a "no code that exists today would possibly be returning `NaN`" -- long term, it doesn't seem like the right decision.

ianstormtaylor:
@justinweiss nice, I agree with you here, I'm down to make this breaking change. Thanks for the clean PR!

PR #1700, opened 2018-03-13
Fix #474

### LINKED ISSUE #474: change key logic to allow for keys with dashes in them
`setKeyGenerator` docs should state that the generated ids shouldn't contain dashes `-`.

Mine did (I used UUIDs), and this broke slate with very confusing errors, until I noticed slate appends `-0` to Leaf nodes and assumes the ids themselves don't contain dashes.

vilic:
https://github.com/ianstormtaylor/slate/commit/e0978a31e93923c88cee9ab831e5a889dc9baa4f#commitcomment-28060594

### DIFF
--- a/packages/slate-react/src/utils/offset-key.js
+++ b/packages/slate-react/src/utils/offset-key.js
@@ -4,7 +4,7 @@
  * @type {RegExp}
  */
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Nice, thank you @vilic!

PR #1704, opened 2018-03-14
Add key prop to <br /> elements in html-serializer

nschlehe:
Get rid of "Each child in an array or iterator should have a unique "key" prop" warning when serializing with html-serializer.


zhujinxuan:
Great catch!  Thanks.

ianstormtaylor:
Thanks @nschlehe!

PR #1705, opened 2018-03-15
Make serialize/deserialize dev friendlier

Kornil:
It just switches the `!=` to `==`, thus removing an empty `return` for the default return in case the `if` is not matched, looks friendlier, easier to read, less confusion.

### DIFF
--- a/docs/walkthroughs/saving-and-loading-html-content.md
+++ b/docs/walkthroughs/saving-and-loading-html-content.md
@@ -103,27 +103,29 @@ const rules = [
     // Switch deserialize to handle more blocks...
     deserialize(el, next) {
       const type = BLOCK_TAGS[el.tagName.toLowerCase()]
-      if (!type) return
-      return {
-        object: 'block',
-        type: type,
-        nodes: next(el.childNodes),
+      if (type) {
+        return {
+          object: 'block',
+          type: type,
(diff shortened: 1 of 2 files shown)


tobiasandersen:
I guess this is just about taste — but I don't agree that this is easier to read. I prefer returning early over nesting in `if` blocks. It's a common pattern throughout the codebase.

Kornil:
In the end, it's just a matter of opinions so I understand if it doesn't get accepted, I've been working on a complex deserializer lately and all the `!=` were confusing me all the time while looking for the right rule.

zhujinxuan:
I am very curious there are so many != and == rather than === and !==. 
Though it is reasonable for !=null, but I would think !==null || !== undefined is much better than !=null

Kornil:
@zhujinxuan agree with you there, that's something I've been thinking about a lot but never asked, strict comparison in a lot of places would make more sense than the loose equivalent.
I think it was more of a stylistic decision (as the absence of `;`).

zhujinxuan:
@Kornil more than stylistic; there are places in slate-hyperscripts about 
```
(result != null)
```
where result is always undefined, null or valid node
however true === (undefined == null)

ianstormtaylor:
Thanks @Kornil, this sounds good to me.

PR #1711, opened 2018-03-19
IME - remove extra force re-renders on onCompositionStart/End

### LINKED ISSUE #879: [IME] Should not be re-rendered when use input method

### LINKED ISSUE #1708: Cant't delete character which is placed just behind transformed character
bug report

can t delete character after kanji sometimes

OS: macOS High Sierra
browser: Firefox nightly v61

and

OS: OSX El Capitan
browser: Chrome v64
slate version: 0.32.2

delete works correctly

### LINKED ISSUE #1710: IME - Japanese input - some characters get erased

_bug_

MacOS High Sierra (10.13.3)
Chrome 64.0.3282.186 using input method: **Hiragana (Google)**

**expected behavior:** if you type the characters: nn[space]wo, the expected output is: ンを

**current behavior:** currently in slate, the output is: ンお  (character sequence: nn[space]o, w gets erased)

It looks like the editor's forced re-render causes the issue in the compositionStart and compositionEnd event handlers.

Related bug: #1708

illesg:
fixes #1708 
fixes #1710 
**edit**: looks like it fixes #879 too

i've removed the tmp.key solution from content.js, because i think it isn't necessary anymore (the before.js > onCompositionStart/End handler triggers the force re-render by calling setState() when `isComposing` changed)

**edit**: I'm not a regular IME user, it would be great if someone could test it in case I introduced new issues with IME input handling

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -72,7 +72,6 @@ class Content extends React.Component {
   constructor(props) {
     super(props)
     this.tmp = {}
-    this.tmp.key = 0
     this.tmp.isUpdatingSelection = false
 
     EVENT_HANDLERS.forEach(handler => {
@@ -283,12 +282,6 @@ class Content extends React.Component {
   onEvent(handler, event) {
     debug('onEvent', handler)
 
-    // COMPAT: Composition events can change the DOM out of under React, so we
-    // increment this key to ensure that a full re-render happens. (2017/10/16)
-    if (handler == 'onCompositionEnd') {
-      this.tmp.key++
-    }
-
     // Ignore `onBlur`, `onFocus` and `onSelect` events generated
     // programmatically while updating selection.
     if (
@@ -497,7 +490,6 @@ class Content extends React.Component {
       <Container
         {...handlers}
         data-slate-editor
-        key={this.tmp.key}
         ref={this.ref}
         data-key={document.key}
         contentEditable={readOnly ? null : true}
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -143,7 +143,10 @@ function BeforePlugin() {
       // HACK: we need to re-render the editor here so that it will update its
       // placeholder in case one is currently rendered. This should be handled
       // differently ideally, in a less invasive way?
-      editor.setState({ isComposing: false })
+      // (apply force re-render if isComposing changes)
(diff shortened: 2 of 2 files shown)


pizn:
Force re-renders is required, because IME is hard to control. I had try to remove tmp.key to fix #879 but it bring other IME bug.

illesg:
@pizn do you remember what was the problem with removing tmp.key? The force re-render still happens when isComposing changes. I couldn't find any new bugs after that change, but i'm not a regular IME user.

pizn:
@illesg like this https://github.com/ianstormtaylor/slate/issues/967.

illesg:
@pizn #967 is an 8 months old open issue, which is still an issue after this change. Could you be more specific please?

21-Hidetaka-Ko:
Same as me. Thanks for issuing. I'm Japanese. So, this is a critical problem for us. We can't launch by you fix this issues. Could you solve this if you can ?

illesg:
I'm already using it in production, no issues from our Japanese costumers since. If it's urgent, try applying these changes, I'm not sure this PR gets merged soon, because it's hard to test IME input behaviour for non-native Japanese people.

21-Hidetaka-Ko:
@illesg Thanks. You are right. How can we fork it? we already tried to fork these branch and PR.
is this right like below?

illesg:
I don't know how to add a monorepo as a npm dependency, i don't think it's easy. For a quick test you can clone the whole repository with the fix implemented, copy it into your project (eg. lib/slate, lib/slate-react), and import e.g. slate-react as ```import { Editor } from "<your path>/slate-react/src";```
it's a bit ugly, but it works :)

illesg:
@21-Hidetaka-Ko I think there is a problem with slate-reacts peer dependency (slate) when it's npm linked. I'm not sure what is the problem, but a quick search gave me this: http://codetunnel.io/you-can-finally-npm-link-packages-that-contain-peer-dependencies/ 
hope it helps

ianstormtaylor:
Thanks @illesg!

PR #1713, opened 2018-03-20
 Improve memorize interface

zhujinxuan:
1. [DONE] use `func.length` to replace `takesArgument` 
2.  [DONE] Add `post` interface to infer and set cache in other nodes


zhujinxuan:
with both #1714 and this post interface, the benchmark is 
```
benchmarks
    html-serializer
      deserialize
        9.08 → 9.31 ops/sec
      serialize
        706.17 → 1081.84 ops/sec (53% faster)
    plain-serializer
      deserialize
        391.95 → 570.18 ops/sec (45% faster)
      serialize
        7887.13 → 10603.34 ops/sec (34% faster)
    rendering
      normal
        91.71 → 126.01 ops/sec (37% faster)
    changes
      delete-backward
        119634.72 → 140803.69 ops/sec
      delete-forward
        23448.94 → 36411.25 ops/sec (55% faster)
      insert-text-by-key-multiple
        95.94 → 133.02 ops/sec (39% faster)
      insert-text-by-key
        919.46 → 1245.49 ops/sec (35% faster)
      insert-text
        899.50 → 1156.37 ops/sec
      normalize
        1971.23 → 2239.59 ops/sec
      split-block
        28.18 → 31.09 ops/sec
    models
      from-json
        90.44 → 114.11 ops/sec
      get-blocks-at-range
        15827.09 → 18976.65 ops/sec
      get-blocks
        660651.56 → 576647.34 ops/sec
      get-characters-at-range
        9457.75 → 7250.64 ops/sec
      get-characters
        9768.90 → 9683.52 ops/sec
      get-inlines-at-range
        339931.46 → 280573.95 ops/sec
      get-inlines
        1558645.38 → 1283291.08 ops/sec
      get-leaves
        1541913.39 → 3466946.24 ops/sec (125% faster) 🙌
      get-marks-at-range
        1465.62 → 1631.63 ops/sec
      get-marks
        2280.98 → 1990.81 ops/sec
      get-path
        884555.59 → 702686.34 ops/sec
      get-texts-at-range
        565170.54 → 464544.12 ops/sec
      get-texts
        1030002.46 → 997138.77 ops/sec
      has-node-multiple
        120752.77 → 93097.92 ops/sec
      has-node
        923524.68 → 753612.76 ops/sec
      to-json
        3990.51 → 3697.66 ops/sec
      update-node
        20440.03 → 17384.67 ops/sec

✨  Done in 57.15s.
```

zhujinxuan:
Reason of faster getLeaves:
1. each time of getLeaves(decoration=[]); the decoration is associated with a new empty array.  Then it would not take advantage of cache
2. solution. when use the default decorations, takeArguments is considered false by 
```
original.length === 0 && args.length === 0
```

zhujinxuan:
awaiting the yarn problem fixed in travis CI...

zhujinxuan:
Update only use last 3 keys, and benchmark as when all possible apps are closed:

only this PR:

this PR and #1714

ianstormtaylor:
@zhujinxuan please don't combine separate features in pull requests like this. I want to merge the removal of `takesArguments` immediately, but I'm unsure about the other one. And since they're combined I now can't do anything. If you split these into two separate pull requests I can consider them. Thanks

zhujinxuan:
@ianstormtaylor Fixed.  The second feature is removed. Thank you.

packages/slate/src/utils/memoize.js
```diff
@@ -108,11 +125,47 @@ function memoize(object, properties, options = {}) {
...
         this.__cache = setIn(this.__cache, keys, v)
       } else {
-        this.__cache.set(property, v)
+        this.__cache_no_args[property] = v
       }
 
       return value
     }
+
+    // Keep fn.length unchanged after cache binding
```
ianstormtaylor: What is this stuff for?

zhujinxuan: The problem of 
```
object["getAncestors"] === function(...args) 
```
is 

object["getAncestors"].length will be 0 regardless of the original function length.
So I have to hardcode that staff to reset the function length

Another trick here is about 
```
function getLeaves(decoration = [])
// getLeaves.length === 0
```
The argument with default params are not counted by function length.  Then I need to pass (...args) in the following in case there are functions with default param length.

ianstormtaylor: Gotcha. I think this sounds like a reason to not simplify the API then, and to keep the existing code as-is? Because we're dealing with a leaky abstraction.

zhujinxuan: @ianstormtaylor The existing code does not keep the func.length either, because the existing `function (...args)` always have `object[property].length === 0`

I can remove this part if you think keeping func.length is not necessary.

zhujinxuan: One advantage of removing takeArguments is that we can allow getLeaves use no arguments cache when with the default params. This is about 100% benchmark speed up in the benchmark test.

ianstormtaylor: Ah I see what you mean. I think we're good to ignore handling the function length here, since we weren't doing it before.

zhujinxuan: @ianstormtaylor Fixed~

packages/slate/src/utils/memoize.js
```diff
@@ -52,6 +52,11 @@ const UNDEFINED = {}
 
 const UNSET = undefined
 
+/**
+ * A set of methods to expose setIn, getIn, set and get
+ *
+ */
+
```
ianstormtaylor: Can you remove this extra comment?

packages/slate/src/models/node.js
```diff
@@ -2023,77 +2023,65 @@ function assertKey(arg) {
...
+  'getLastText',
+  'getMarksAsArray',
+  'getText',
+  'getTextDirection',
+  'getTextsAsArray',
+  'isLeafBlock',
+  'isLeafInline',
+])
+
+memoize(Node.prototype, [
```
ianstormtaylor: Can you combine these into a single one now, alphabetized?

packages/slate/src/utils/memoize.js
```diff
@@ -79,12 +82,26 @@ function memoize(object, properties, options = {}) {
...
       }
+      if (!this.__cache_no_args) {
+        this.__cache_no_args = {}
+      }
+
+      // The default parameter of text.getLeaves would have an arg with fn.length===0
+      const takesArguments = original.length !== 0 || args.length !== 0
+      if (IS_DEV && takesArguments) {
+        if (original.length > args.length) {
+          throw new Error(
```
ianstormtaylor: Can you remove this error? I'm not sure why we need it, but it seems weird for this part of the code to enforce this restriction.

packages/slate/src/utils/memoize.js
```diff
@@ -79,12 +82,26 @@ function memoize(object, properties, options = {}) {
...
 
       if (!this.__cache) {
         this.__cache = new Map() // eslint-disable-line no-undef,no-restricted-globals
       }
+      if (!this.__cache_no_args) {
+        this.__cache_no_args = {}
+      }
+
+      // The default parameter of text.getLeaves would have an arg with fn.length===0
+      const takesArguments = original.length !== 0 || args.length !== 0
```
ianstormtaylor: Do we need to check the `original` at all here actually? Could it just be that argument-less always kicks in when no arguments are passed?

zhujinxuan: Thank you for pointing out. Fixed~

packages/slate/src/utils/memoize.js
zhujinxuan: Yes, I remove it in the recent commit.

packages/slate/src/models/node.js
zhujinxuan: Fixed~

packages/slate/src/utils/memoize.js
zhujinxuan: Fixed~

packages/slate/src/utils/memoize.js
```diff
@@ -79,12 +77,19 @@ function memoize(object, properties, options = {}) {
...
       }
 
       if (!this.__cache) {
         this.__cache = new Map() // eslint-disable-line no-undef,no-restricted-globals
       }
+      if (!this.__cache_no_args) {
+        this.__cache_no_args = {}
+      }
+
+      // The default parameter of text.getLeaves would have an arg with fn.length===0
```
ianstormtaylor: Can you remove this comment that is no longer relevant?

zhujinxuan: Fixed~ Thank you.

ianstormtaylor:
Thanks @zhujinxuan, a very clean PR in the end with minimal changes but that got the solution implemented well!

PR #1604, opened 2018-02-05
Use getAncestors in getDescendant

zhujinxuan:
Both `getAncestors` and `getDescendant` are DFS, but the `getDescendant` cannot take advantage by the cache by `getAncestors`.   This PR makes `getDescendant` taking advantage from `getAncestors` cache.

PR #1714, opened 2018-03-20
Restore #1604: Use getAncestors in getDescendant

zhujinxuan:
PR #1589 overwrites #1604 PR.  I think it is because git text comparison fails to warn the conflict between PR 1589 and other PRs.  

This PR is to restore PR #1604 , which is to use getAncestors in getDescendant to speed up the getDescendant when getAncestors is cached.

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -351,6 +351,7 @@ class Node {
 
     if (key == this.key) return List()
     if (this.hasChild(key)) return List([this])
+    if (!this.getDescendant(key)) return null
 
     let ancestors
     this.nodes.find(node => {
@@ -359,11 +360,7 @@ class Node {
       return ancestors
     })
 
-    if (ancestors) {
-      return ancestors.unshift(this)
-    } else {
-      return null
-    }
+    return ancestors.unshift(this)
   }
 
   /**
@@ -651,10 +648,20 @@ class Node {
 
   getDescendant(key) {
     key = assertKey(key)
-    // Use the cache by getAncestors
-    const ancestors = this.getAncestors(key)
-    if (!ancestors || ancestors.size === 0) return null
-    return ancestors.last().getChild(key)
+    let descendantFound = null
+
+    const found = this.nodes.find(node => {
+      if (node.key === key) {
+        return node
+      } else if (node.object !== 'text') {
+        descendantFound = node.getDescendant(key)
+        return descendantFound
+      } else {
+        return false
+      }
+    })
+
+    return descendantFound || found
   }
 
   /**
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -651,20 +651,10 @@ class Node {
 
   getDescendant(key) {
     key = assertKey(key)
-    let descendantFound = null
-
-    const found = this.nodes.find(node => {
-      if (node.key === key) {
-        return node
-      } else if (node.object !== 'text') {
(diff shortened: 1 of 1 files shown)


zhujinxuan:
I looked around other PRs in Feb 6.  Other PRs are not influenced by #1589

zhujinxuan:
benchmark comparison at this PR
```
  benchmarks
    html-serializer
      deserialize
        9.08 → 9.59 ops/sec
      serialize
        706.17 → 1115.16 ops/sec (58% faster)
    plain-serializer
      deserialize
        391.95 → 561.18 ops/sec (43% faster)
      serialize
        7887.13 → 8441.90 ops/sec
    rendering
      normal
        91.71 → 95.56 ops/sec
    changes
      delete-backward
        119634.72 → 120250.14 ops/sec
      delete-forward
        23448.94 → 21054.85 ops/sec
      insert-text-by-key-multiple
        95.94 → 104.87 ops/sec
      insert-text-by-key
        919.46 → 1164.11 ops/sec
      insert-text
        899.50 → 1146.82 ops/sec
      normalize
        1971.23 → 1752.37 ops/sec
      split-block
        28.18 → 28.49 ops/sec
    models
      from-json
        90.44 → 159.98 ops/sec (77% faster)
      get-blocks-at-range
        15827.09 → 27666.13 ops/sec (75% faster)
      get-blocks
        660651.56 → 814328.16 ops/sec
      get-characters-at-range
        9457.75 → 13116.15 ops/sec (39% faster)
      get-characters
        9768.90 → 14603.87 ops/sec (49% faster)
      get-inlines-at-range
        339931.46 → 466801.99 ops/sec (37% faster)
      get-inlines
        1558645.38 → 1859080.49 ops/sec
      get-leaves
        1541913.39 → 1770915.22 ops/sec
      get-marks-at-range
        1465.62 → 2023.92 ops/sec (38% faster)
      get-marks
        2280.98 → 1215.76 ops/sec (47% slower)
      get-path
        884555.59 → 831902.28 ops/sec
      get-texts-at-range
        565170.54 → 596404.45 ops/sec
      get-texts
        1030002.46 → 1036032.50 ops/sec
      has-node-multiple
        120752.77 → 128937.99 ops/sec
      has-node
        923524.68 → 954112.59 ops/sec
      to-json
        3990.51 → 3939.42 ops/sec
      update-node
        20440.03 → 19089.29 ops/sec

✨  Done in 62.42s.
```

ianstormtaylor:
Why? It doesn't appear to make things faster?

zhujinxuan:
I think there are two reasons.  
1. using getAncestors get code shorter
2. it is normal in plugins to use getClosestXXX or other ancestors related methods after getDescendants.  For example, whether a text is in void inline, void image or table.

And a potential reason I am not sure about:
I observe a strange benchmark speed up:  getLeaves is speed up when both this PR and the post interface of memorize is added. 

Let us postpone it.  I will add some benchmark files about delete/insert with nested structures to see whether this PR is good.

ianstormtaylor:
Awesome, that sounds good to me. Thanks!

If we do decide we want to go with this technique, then I think it would be worth adding a `getDescendants(key)` or something similar that returns ancestors **and** the child itself. Then `getAncestors` and `getDescendant` and anything else can all call into a single method. Only reason I say that is because I think it's more clear than using `getAncestors().last().getChild(key)` which is hard to follow.

zhujinxuan:
Close this PR; Because https://github.com/ianstormtaylor/slate/pull/1763 is a much better idea ~

PR #1716, opened 2018-03-22
Don't stop propagation of drop event

lxcid:
We mirrored the fix in PR #1278 (meant for #1277) but for drop event.
This fixed an issue I faced with Slate swallowing drop event, kind of breaking compatibility with react-dnd.

/cc @AlbertHilb


ianstormtaylor:
Thanks @lxcid!

PR #1718, opened 2018-03-22
Fix node.isEmpty

urugator:
Since 0.33.0 void nodes can contain empty string `''`. This means that `node.text === ''` is no longer a valid indicator of "emptiness". Consider: `<a><void/></a>` then `a.text === ''` , but `<a>` isn't actually empty.
This PR changes the implementation of `isEmpty` in the following way:
- `value.isEmpty` returns true if selected document fragment `isEmpty` (see next)
- `document.isEmpty` returns true if all document's children nodes are empty
- `block.isEmpty` returns true if block is **not** void and all it's children nodes are empty
- `inline.isEmpty` returns true if inline is **not** void and all it's children nodes are empty
- `text.isEmpty` - unchanged, text is empty if it represents empty string `''`

Therefore any void node is considered non-empty, regardless of actual content.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -121,7 +121,7 @@ To get a sense for how you might use Slate, check out a few of the examples:
 
 * [**Plain text**](./examples/plain-text) — showing the most basic case: a glorified `<textarea>`.
 * [**Rich text**](./examples/rich-text) — showing the features you'd expect from a basic editor.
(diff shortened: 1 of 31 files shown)


ianstormtaylor:
Thank you @urugator, great call here.

PR #1721, opened 2018-03-23
Make cache management available in non-dev environment

Zhouzi:
We are facing heap memory errors due to memoization taking more and more space. We are paying close attention to the progress made on that side and will be investing in the efforts as well. In the meantime, and as long as the issues are not fixed for good, it would be great to be able to opt-out of memoization.


ianstormtaylor:
Sounds good to me, thanks @Zhouzi!

PR #1727, opened 2018-03-25
add missing ```js instead of ``` on two code blocks

larskarbo:
Just two simple fixes I noticed

before: ```
after: ```js


ianstormtaylor:
Awesome, thank you @larkskarbo!

Issue #1722, opened 2018-03-24
cloneFragment is broken when multiple <Editor /> instances are present


czechdave:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?
If we have multiple instances of <Editor /> on one page, it is impossible to copy text from any but the first instance in the DOM.

#### What's the expected behavior?
We should be able to copy from any editor instance.

This bug is caused by the `cloneFragment` function always selecting the first instance in the DOM regardless of which instance the copy event was triggered from:
https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/clone-fragment.js#L99

tobiasandersen:
Fixed by @czechdave in #1728 🎉

PR #1728, opened 2018-03-27
Fix cloneFragment to work with multiple Editor instances

czechdave:
Currently when multiple instances of `<Editor />` are on one page `onCopy` silently fails for all but the first instance in the DOM. This PR makes the `cloneFragment` function use the DOM node of the editor instance it's being triggered in which solves this issue.

(Issue #1722)


ianstormtaylor:
Nice, thanks @czechdave!

Issue #1709, opened 2018-03-19
Slate crashes on ALT+SHIFT+BACKSPACE


pferdone:
#### What's the current behavior?

- A JSFiddle that minimally reproduces the issue: http://slatejs.org/#/rich-text
- A GIF showing the issue in action: https://imgur.com/rjbcsqS
- Information about your OS, browser, Slate version, etc: MacOS, Chrome 64.0.3282.186

#### What's the expected behavior?

No exception on ALT+SHIFT+BACKSPACE

PR #1729, opened 2018-03-27
fix: add missing hotkey for delete word backward, close: #1709

benjycui:
Close: #1709 

On Mac, both of `shift+option+backspace` and `option+backspace` are *delete word backward*.


ianstormtaylor:
Good call @benjycui, should this happen for delete word forward too? And can you fix the linter?

tobiasandersen:
This fix works great, and should be added for delete word forward as well (it has the exact same problem as #1709).

benjycui:
Lint fixed.

> This fix works great, and should be added for delete word forward as well (it has the exact same problem as #1709).

@tobiasandersen could you help me to test *delete word forward* ? There is only one `delete` button on my MacBook, and it means `backspace`

tobiasandersen:
I can confirm that this fixes it for delete on mac (using `fn + backspace` for delete) 👍

ianstormtaylor:
Beautiful, thank you @benjycui!

Issue #1726, opened 2018-03-24
Remove draggable attribute in nodes.


chilijung:
#### Do you want to request a _feature_ or report a _bug_?

Now every node will wrap with a `div` container with draggable attribute, this will conflict with customized dnd features, source: https://github.com/ianstormtaylor/slate/blob/3339d088e14ad0bf04a674c14834179403ae369f/packages/slate-react/src/components/void.js#L79

![](https://user-images.githubusercontent.com/3375461/37860520-4a44dee0-2ee4-11e8-8129-f8020aaeae90.png)

Code source: https://github.com/Canner/slate-editor-icons/blob/master/packages/slate-editor-renderer/src/ImageNode.js#L128

You could try it live here: https://canner.github.io/slate-editor-icons/

#### What's the current behavior?

Wrap every node with draggable attribute.

#### What's the expected behavior?

The attribute should be remove.

pizn:
Handle image resizing can use `mouseDown`/`mouseMove`/`mouseUp` event. And void node support `draggable` attribute is needed.

chilijung:
@pizn, I think slate shouldn't default add `draggable` attribute, there are many dnd libraries that do better jobs. Right now it's interfering other libraries.

pizn:
Maybe can support custom defined `draggable` attribute, default is true.

tobiasandersen:
Yeah I've also found myself work around this. Couldn't people just set the `draggable` attribute themselves if they want it, but also where they want it?

czechdave:
Just ran into an issue with the forced `draggable` attribute too. What is the reason Slate does anything related to draggability? As far as I can tell that attribute should just be removed.

ianstormtaylor:
I’m down for it to be opt-in. If whoever submits the PR can also make it so that the images and emoji examples remain draggable that would be perfect.

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #1731, opened 2018-03-27
Remove draggable attribute from void nodes

tobiasandersen:
As stated in https://github.com/ianstormtaylor/slate/issues/1726, having the `draggable` attribute set on the containers of void nodes can be a bit cumbersome when e.g. integrating with third party libs. This PR simply removes that attribute. I'm thinking it should be easy enough for people needing this to just add it to the actual nodes instead. 

As for the examples, DnD still works in the image example since the `` element is draggable. Emojis though, are no longer draggable in the emojis example. I've opted not to add the `draggable` attribute to the emoji nodes since neither the dragging nor the dropping currently work properly (not sure it ever did?). For reference, this is what the dragging looks like before this PR:

slate-emoji-dnd

### DIFF
--- a/packages/slate-react/src/components/void.js
+++ b/packages/slate-react/src/components/void.js
@@ -72,10 +72,7 @@ class Void extends React.Component {
     )
 
     const content = (
-      <Tag
-        contentEditable={readOnly ? null : false}
-        draggable={readOnly ? null : true}
-      >
+      <Tag contentEditable={readOnly ? null : false}>
         {children}
       </Tag>
     )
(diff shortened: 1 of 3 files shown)


mbrnwsk:
Any chance for merging soon? Quite urgent thing :)

marek-baranowski:
This PR is really needed, @ianstormtaylor do you plan to merge it anytime soon?

ianstormtaylor:
Hey @tobiasandersen can you update the rendering tests so that they pass with this change? Thanks!

tobiasandersen:
Oh sorry, missed that. They should pass again now.

ianstormtaylor:
Thank you @tobiasandersen, this is great :D

PR #1732, opened 2018-03-28
Both the iOS and Android Development Guide

thesunny:
Ignore the old PR. This one has a cleaner directory structure and I resized all the images to be at non-retina size so as to avoid bloat in the repo.


thesunny:
This PR has the Android Development Guide as well. A much better PR overall.

We should probably write somewhere to not to use GitBook for updating docs.

I'm failing the lint checks but am not sure how to get it to pass. I can run `yarn lint` but it's not giving me any useful output to know what I need to fix. It just lists the files that I presume are failing.

thesunny:
Fixing lint problems...

thesunny:
Sorry, I just noticed that the HTML => markdown conversion removed my image comments. Adding them back in.

thesunny:
Okay, I think I'm done. Please check and merge if it looks good. Let me know if you spot anything.

ianstormtaylor:
Hey @thesunny thanks for taking the time to write this up, the guides look really good for getting people up to speed.

I think that these guides should be something that is maintained outside of Slate, since they don't need to be Slate-specific at all, and adding them to the codebase adds extra maintenance burden for something that shouldn't be tied to Slate at all. If you publish them as another repo (or in whatever way you'd like) then we can link to them from the contributing documentation to help folks.

thesunny:
@ianstormtaylor Okay, I'll think about how to maintain these outside this repo.

I do think we still need a Mobile Development Guide though because there are certain aspects of the guide that is in this PR that relates specifically to Slate and/or ContentEditable.

* iOS and Android versions to support
* iOS and Android devices to support
* Gotchas specifically related to SlateJS like disabling of the hardware keyboard

ianstormtaylor:
@thesunny I think the versioning questions can be loose for now, the last two OS versions sounds fine to me. And the flagship devices sounds fine to start. This is something that we can figure out as we go, since even just getting solid support on the flagship devices would be great to start, instead of overreaching.

I'm open to other specific gotchas being added to the `Contributing.md` docs or to the FAQ, or some new document if one really needs to be created.

That said, I really want to avoid documentation bloat for things that don't absolutely need to be documented. Maintaining documentation in open-source is already hard enough, so we should minimize the footprint as much as possible.

I think the biggest thing prevent Android support is not documentation, but is for Android people to be incentivized enough to learn the repository and contribute fixes. We don't have super in depth browser documentation, or IE documentation, yet those fixes are still contributed.

thesunny:
I understand your point of view. I'd (and this is with great respect to you) disagree about the documentation.

For me, the depth of documentation has a huge effect on whether I start contributing to it.

As an example, it took a lot to get me to dive into the iOS bug fix. It took me a few days (not continuously but still) just to get my development environment setup for mobile with iOS. It's much hairier with mobile because I didn't know anything whereas with desktop, everybody using Slate already knows how to debug in the desktop browser. For iOS, I bought an iPad just to do debugging. Then found out there was a simulator and I didn't need it but the documentation on setting it up was poor.

If I had the documentation I just wrote, I would have been involved sooner because it would have taken me about an hour to set up and more importantly, I wouldn't have had the big issue of not knowing how long it would take to set up the dev environment.

The SlateJS documentation itself gives me a lot of confidence to use Slate. It is extremely well written and I feel like I understand, from a 50,000 ft view, how everything ties together and what its trade offs are and why.

I feel like the internals could use fleshing out like this and, at least for me, would have improved my likelihood to contribute. For example, I would love (absolutely love) to understand how the events move from key, through plugins, and update `Value`. I'm kind of dreading stepping through that source code to figure out the 50,000 ft view. Would be a lot easier getting the high view, and then seeing how it ties together in code.

With Android too, I didn't know if there was an emulator and the documentation for this is poor. I basically had to ask around which, if I wasn't doing this Mobile roadmap, I 100% wouldn't have taken the time to try out. Once I had the environment though, I could replicate the bug in seconds.

Even a few comments about how to contribute to the documentation, how not to use GitBook, how to run the linter and how to use prettier to fix any failures on `.md` files would have saved me a lot of time (in my case a few hours). I was planning to contribute these actually but won't if you don't want them in there.

I think some things to consider as well are:

* Starting in mobile web development is of significant difficulty compared to debugging Slate on desktop browsers because most people using Slate are probably not familiar with starting up devices and what gotchas there are. Most people are probably familiar with mobile screen sizing and media queries but that's it.
* IMHO, your familiarity of the Slate code (since you wrote it) may give you a blind spot in how hard it is to piece how everything fits together for somebody that isn't you. You've already invested the time so the documentation adds little value to you and perhaps adds more overhead.

Personally, though, I think the project could benefit from more documentation, especially on the internals and for contributing.

ianstormtaylor:
Hey @thesunny, thanks for that, I hear where you are coming from.

I definitely think it would be good to document more of Slate's internals, and explain how the architecture is designed. This would absolutely help with people submitting pull requests, and feeling more comfortable taking on some of the bigger features/issues that need more refactoring. The only reason we don't have more of this is because of me not having time to write it. I've started with the "guides". And if others want to write more of the higher level architecture documentation that would be extremely welcome.

But what I don't want to do is add lots of documentation for things that aren't Slate specific, and that I don't have a good way of keeping up to date—or even knowing when the information is out of date. For me, detailed instructions on how to get an iOS Simulator or Android Simulator running for different OS versions falls into this category. It's absolutely something that's important and very helpful, I just don't think it belongs in Slate's GitHub repository because it's a lot of work to maintain, and we don't have the bandwidth.

I think it's the perfect thing to go in it's own dedicated place. It would be awesome if there was a single resource web developers could go to see how to debug/reproduce things in any browser—be it Chrome, Safari, Opera, Firefox, iOS, Android, etc. These could maybe be the start of that project. And putting it in it's own dedicate place would both cause it to get more visibility, and increase the chance that it was well maintained.

thesunny:
Thanks for the clarification @ianstormtaylor 

The choice makes sense. As I mentioned in Slack, my concern was that we'd have less docs on internals and contributing but I like your clarification that you're fine with more Slate specifics and keeping generic information separate but possibly linked to.

ianstormtaylor:
Thanks for understanding @thesunny! If you end up publishing the mobile guides in another repository, I'm definitely down for us to add links to them from our contributing docs.

Issue #1669, opened 2018-02-22
Autofocus does not always result in the cursor being placed in the document


CameronAckermanSEL:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

The autofocus prop currently focused the editor, but a cursor does not appear.

This is because autofocus does not actually select a block! So no block is selected and therefore the cursor does not appear.

#### What's the expected behavior?

The autofocus prop should by default try to figure out a sensible block to select (usually the 0 index of the first block in the document), unless the `value` already has a selection defined. 

This would allow the user to turn on the autofocus behavior, but still be able to control where the cursor shows up. 

Alternatively, the user could also provide a function for resolving selection during autofocus that would allow for more complex default selection rules (for instance, in editors with a Title block, users typically want the cursor to show up after the title).

benjycui:
See: https://github.com/ianstormtaylor/slate/pull/1733

PR #1733, opened 2018-03-28
fix: Editor[autoFocus] should work, close: #1669

benjycui:
Close: https://github.com/ianstormtaylor/slate/issues/1669

Currently, Editor will cache the value which is passed from outside

https://github.com/ianstormtaylor/slate/blob/4346ad0e3f4a956e40a71f3cfd36682be6f90f4a/packages/slate-react/src/components/editor.js#L92

And then flush it when mount

https://github.com/ianstormtaylor/slate/blob/4346ad0e3f4a956e40a71f3cfd36682be6f90f4a/packages/slate-react/src/components/editor.js#L152

This will make any changes between Editor's initialization and componentDidMount lost. Such as `.focus` in Content's componentDidMount

https://github.com/ianstormtaylor/slate/blob/4346ad0e3f4a956e40a71f3cfd36682be6f90f4a/packages/slate-react/src/components/content.js#L109-L111

### DIFF
--- a/packages/slate-react/src/components/editor.js
+++ b/packages/slate-react/src/components/editor.js
@@ -145,11 +145,16 @@ class Editor extends React.Component {
   }
 
   /**
-   * When the component first mounts, flush any temporary changes.
+   * When the component first mounts, flush any temporary changes,
+   * and then, focus the editor if `autoFocus` is set.
    */
 
   componentDidMount = () => {
     this.flushChange()
+
+    if (this.props.autoFocus) {
+      this.focus()
+    }
   }
 
   /**
--- a/examples/rich-text/index.js
+++ b/examples/rich-text/index.js
@@ -263,6 +263,7 @@ class RichTextExample extends React.Component {
           renderNode={this.renderNode}
           renderMark={this.renderMark}
           spellCheck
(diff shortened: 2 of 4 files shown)


packages/slate-react/src/components/editor.js
```diff
@@ -145,11 +145,16 @@ class Editor extends React.Component {
   }
 
   /**
-   * When the component first mounts, flush any temporary changes.
+   * When the component first mounts, flush any temporary changes,
+   * and then, focus the editor if `autoFocus` is set.
    */
 
   componentDidMount = () => {
     this.flushChange()
```
benjycui: I fixed the bug by hoisting `.focus` to Editor's componentDidMount.

But I don't understand why we need to cache initial value and then flush it again. @ianstormtaylor

ianstormtaylor:
Nice solution @benjycui, thanks!

The reason for the caching/flushing is because when a schema is applied to a `value` it can result in changes being applied, and we need to flush after construction so that it's in a "normal" phase for `onChange` to be called. (Or so I think, open to ideas.)

benjycui:
Make sense, this will make Editor more *controlled*.

Issue #1639, opened 2018-02-15
Cannot delete the whole document content when the document starts with a void block


zhujinxuan:
#### Do you want to request a _feature_ or report a _bug_?

bug
#### What's the current behavior?

Select all and hit a-z in the following gif:
![](http://g.recordit.co/vVfdrBEfnu.gif)

I also tried backspace and backspace is not working.

benjycui:
Slate's void node will be render as `contentEditable=false`, which will affect behavior of browser's `window.getSelection`.

You can try to debug in `onSelect` of Slate's after plugin, and then `mod+a`, you will see what I mean.

benjycui:
https://github.com/ianstormtaylor/slate/pull/1734 will fix this issue.

PR #1734, opened 2018-03-28
fix: make void node selectable for edge case, close: #1639

benjycui:
Close: https://github.com/ianstormtaylor/slate/issues/1639

In Slate, we render void node as `contentEditable=false`, it works fine under most circumstances. 

However, if the void node is the first child of Editor, the behavior of selection will be broken, e.g. the void node will not be select when we try to select all.

So, I just make void node's wrapper `contentEditable=true`, and make its content `contentEditable=false`, and then everything works fine.

### DIFF
--- a/packages/slate-react/src/components/void.js
+++ b/packages/slate-react/src/components/void.js
@@ -66,26 +66,24 @@ class Void extends React.Component {
     }
 
     const spacer = (
-      <Tag
-        contentEditable
-        data-slate-spacer
-        suppressContentEditableWarning
-        style={style}
-      >
+      <Tag data-slate-spacer style={style}>
         {this.renderText()}
       </Tag>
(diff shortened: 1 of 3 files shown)


benjycui:
You can checkout the branch and test it locally 😄

ianstormtaylor:
Interesting solution @benjycui! Can you verify that all of the selection before (moving across, up/down, left/right, etc.) works properly with void nodes in Chrome, Safari & Firefox.

benjycui:
Of course, I had tested it. And it works fine.

Actually, I just change wrapper's contentEditable to true, and others are the same.

benjycui:
ping~

ianstormtaylor:
Thanks @benjycui!

Issue #1671, opened 2018-02-23
Plugin schemas not combining properly


tashburn:
If I add two plugins to the same Editor with these schemas --

```
  {
    schema: {
      document: {
        nodes: [ 
          { types: [ 'paragraph' ] } 
        ],
      },
      blocks: {
        paragraph: {
          nodes: [{ objects: ['text','inline'] }],
        }
      }
    }
  }
```

```
  {
    schema: {
      document: {
        nodes: [ 
          { types: [ 'quote' ] } 
        ],
      },
      blocks: {
        quote: {
          nodes: [{ objects: ['text','inline'] }],
        }
      }
    }
  }
```

The combination should be --

```
  {
    schema: {
      document: {
        nodes: [ 
          { types: [ 'paragraph', 'quote' ] } 
        ],
      },
      blocks: {
        paragraph: {
          nodes: [{ objects: ['text','inline'] }],
        },
        quote: {
          nodes: [{ objects: ['text','inline'] }],
        }
      }
    }
  }
```

But is actually (wrongly) --

```
  {
    schema: {
      document: {
        nodes: [ 
          { types: [ 'paragraph' ] } 
        ],
      },
      blocks: {
        paragraph: {
          nodes: [{ objects: ['text','inline'] }],
        },
        quote: {
          nodes: [{ objects: ['text','inline'] }],
        }
      }
    }
  }
```

What's wrong is `schema.document.nodes.types`. It should have both block types, but it only has `paragraph`.

CameronAckermanSEL:
There's probably a limitation where the schema plugin logic does not merge conflicting rules. I think this is probably very difficult to get right.

CameronAckermanSEL:
One other thought. I think it might be an antipattern for a plugin to dictate where it fits in the document schema, because it creates conflicting rules where the consumer of the plugin might have a different set of expectations on where the component provided by the plugin sits within the document. To get this functionality correct, I think we'd be looking at a pattern where the plugin that provides the block basically "registers" where the block sits in the document with the document.nodes schema on top of the other encapsulated rules for the schema rules of the block. Delegating control over the block's relationship with the document and blocks to the plugin makes the plugin less flexible.

tashburn:
In order to get this working now, I did the following to compose my plugin schemas together and feed into the `Editor` as one schema, which seems to work (in Typescript) --

```
function combineSchemas(plugins : Plugin[]) {
  let schema = {
    document: {
      nodes: [] as any[],
    },
[...]

PR #1736, opened 2018-03-28
[WIP] Schema issues and possible improvements

Zhouzi:
We're currently upgrading GitBook's Slate version and thus rewriting our schemas with the new definition. I'm opening this PR to document the issues we are facing along with possible improvements.

* ~`min: 1` doesn't work well for empty blocks when combined with another restriction (such as `types` or `objects` set to something else than "text") ([test](https://github.com/ianstormtaylor/slate/pull/1736/commits/f970b71c362779d6ad0fb6e819f6f7c74d14ce41))~
* ~similar to the issue above, `max: 1` also leads to empty blocks being removed due to their empty text node not matching the provided `types`. To be more precise, the text node is not removed because of the `max: 1` but the `types`, it's just that the `max: 1` makes it obvious it's optional ([test](https://github.com/ianstormtaylor/slate/pull/1736/commits/0c59286859c505667f9bb728c6e90ade6ffefae3))~
* when enforcing a node's children order, `child_type_invalid` is raised before `child_required` which leads to the node being removed although it could have been moved to the correct position ([test](https://github.com/ianstormtaylor/slate/pull/1736/commits/4c8bc05d142e2fc320e9dd18816beda8afb0bd58))
* `child_type_invalid` has a higher priority than `parent_type_invalid` while there are probably more chances that the user will want to normalize the parent before the child as we tend to think "top down" rather than "bottom up" (just my opinion though, I could be wrong here). Perhaps the order is fine as it is, it just needs to be decided/documented ([test](https://github.com/ianstormtaylor/slate/pull/1736/commits/8163e14fb3cd570d8daf1429836d15f6b75ab233))
* plugins are not properly merged, resulting in some rules being lost ([test](https://github.com/ianstormtaylor/slate/pull/1736/commits/e38baa5139d67e755265144554a54eadb36051d8)) **[related discussion](https://github.com/ianstormtaylor/slate/issues/1671)**
* `validateNode` will always have the higher priority, no matter the plugin's order. This is definitely not an issue but an illustration of why applying the plugins one after the other can help in certain situations (and thus left to the user, there's nothing to be changed in Slate). ([test](https://github.com/ianstormtaylor/slate/pull/1736/commits/69c505e9edb909098482ad3d822f7f6d68adf45d))

**Note:** I am discussing my findings with the team and crossing invalid items (things I first seen as an issue and then understood to be a conceptual misunderstanding).


Soreine:
> child_type_invalid has a higher priority than parent_type_invalid while there are probably more chances that the user will want to normalize the parent before the child as we tend to think "top down" rather than "bottom up"

Slate's normalization works from the bottom up, so we normalize the child before the parent, otherwise higher level rules have a lot more work to do, because they are given invalid children.
Concerning those two rules, 'parent_type_invalid' should be called first (normalization of a child) rather than `child_type_invalid` (normalization of a parent).

ianstormtaylor:
Hey @Zhouzi all of your points here sound smart to me. I'd be down for PRs that fix them one by one if that's something you'd be willing to do? Either that or let me know how you want to proceed here!

Zhouzi:
@ianstormtaylor we ended up working around those minor issues. To be honest, they're no big deal and mostly nitpicks. We'll open pull requests if we feel it's worth it 👍

PR #1737, opened 2018-03-28
Fixed syntax mismatch at normalize method

finethanks:
`switch` operator was wrapped around `case` statements


ianstormtaylor:
Ah, thanks @finethanks!

Issue #1562, opened 2018-01-26
Enter on selected link causes error


svenadlung:
#### Do you want to request a *feature* or report a *bug*?
Bug

#### What's the current behavior?
Selecting a link and pressing Enter results in an error. 
http://slatejs.org/#/links

![jan -26-2018 15-46-02](https://user-images.githubusercontent.com/11420688/35444850-15b34cd0-02b0-11e8-877d-682f02f9301e.gif)

![slate](https://user-images.githubusercontent.com/11420688/35444732-bb8f6626-02af-11e8-9c2d-9353720fb659.png)

#### What's the expected behavior?
It should behave the same like regular text (split).

benjycui:
See: https://github.com/ianstormtaylor/slate/pull/1739

PR #1739, opened 2018-03-29
fix: get the right range after delete, close: #1562

benjycui:
Close: https://github.com/ianstormtaylor/slate/issues/1562

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -990,7 +990,7 @@ Changes.splitBlockAtRange = (change, range, height = 1, options = {}) => {
 
   if (range.isExpanded) {
     change.deleteAtRange(range, { normalize })
-    range = range.collapseToStart()
+    range = change.value.selection
   }
 
   const { startKey, startOffset } = range


packages/slate/src/changes/at-range.js
```diff
@@ -990,7 +990,7 @@ Changes.splitBlockAtRange = (change, range, height = 1, options = {}) => {
 
   if (range.isExpanded) {
     change.deleteAtRange(range, { normalize })
-    range = range.collapseToStart()
+    range = change.value.selection
```
benjycui: Sometimes, start block may be deleted. So that we cannot just collapse selection to start, but get the new selection from changed value.

ianstormtaylor: I think as written here this will only work when the `range` to be deleted is also the currently selected range (ie. `range === selection`).

Instead, it seems like we'd need to check if the start key still exists in the document after the delete, and if it doesn't then select the key of the text node before it?

benjycui: Actually, this work even though the start block is not been deleted.

> Instead, it seems like we'd need to check if the start key still exists in the document after the delete, and if it doesn't then select the key of the text node before it?

`change.deleteAtRange` had did this for us, so what we need to do is just read the latest selection.

tobiasandersen:
This looks very similar to what I did in https://github.com/ianstormtaylor/slate/pull/1545, where @ianstormtaylor had some concerns. I haven't had the time to look into it yet though...

packages/slate/src/changes/at-range.js
ianstormtaylor: The problem here is that `*AtRange` methods doesn't act on the current user's selection, they act at an arbitrary range in the document. So they can't use `change.value.selection`, because someone may be splitting a block at a range programmatically that isn't the current selection.

zhujinxuan:
My fix for insertText if it is helpful:
https://github.com/ianstormtaylor/slate/pull/1741

packages/slate/src/changes/at-range.js
benjycui: OK, I understand.

PR #1741, opened 2018-03-29
Fix insertTextAtRange

zhujinxuan:
## Description:
change insertTextAtRange logic to insert and then delete

## Why:
startKey can be deleted in deleteAtRange in hanging selection

### DIFF
--- a/packages/slate/src/changes/at-current-range.js
+++ b/packages/slate/src/changes/at-current-range.js
@@ -199,12 +199,34 @@ Changes.insertText = (change, text, marks) => {
   const { value } = change
   const { document, selection } = value
   marks = marks || selection.marks || document.getInsertMarksAtRange(selection)
-  change.insertTextAtRange(selection, text, marks)
+  change.insertTextAtRange(selection, text, marks, { normalize: false })
+
+  // Restore the cursor if startKey is deleted
+  if (!change.value.document.getDescendant(selection.startKey)) {
+    change.collapseTo(selection.endKey, text.length)
+  }
 
   // If the text was successfully inserted, and the selection had marks on it,
   // unset the selection's marks.
   if (selection.marks && document != change.value.document) {
     change.select({ marks: null })
+  }
+
+  // normalize in the narrowest existing block that originally contains startKey and endKey
+  const commonAncestor = document.getCommonAncestor(
+    selection.startKey,
+    selection.endKey
+  )
+  const ancestors = document
+    .getAncestors(commonAncestor.key)
+    .push(commonAncestor)
+  const normalizeAncestor = ancestors.findLast(n =>
+    change.value.document.getDescendant(n.key)
+  )
+  if (normalizeAncestor) {
+    change.normalizeNodeByKey(normalizeAncestor.key)
+  } else {
+    change.normalizeNodeByKey(change.value.document.key)
   }
 }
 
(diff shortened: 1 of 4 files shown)


zhujinxuan:
Update: 
In recent commit, I switch the logic to insert and then delete; it seems that method solves the problem in a simpler way.

packages/slate/src/changes/at-current-range.js
```diff
@@ -199,7 +199,7 @@ Changes.insertText = (change, text, marks) => {
   const { value } = change
   const { document, selection } = value
   marks = marks || selection.marks || document.getInsertMarksAtRange(selection)
-  change.insertTextAtRange(selection, text, marks)
+  change.insertTextAtRange(selection, text, marks, { normalize: true })
```
zhujinxuan: Remember to remove `{normalize:true}` and add back PREF in at-range.

zhujinxuan: Fixed

ianstormtaylor:
@zhujinxuan why not use the same trick you used in the other pull request, so that we still delete first (since I think it's simpler?) but then right after the delete we properly fix the range if the `startKey` was deleted?

zhujinxuan:
@ianstormtaylor Because considern this:
```
<paragraph><anchor/> Cat is Cute </paragraph>
<quote><focus/> Something else </quote>
```

I want the result to be 
```
<paragraph>x<cursor/></paragraph>
<quote> Something else </quote>
```
instead of 
```
<quote>xSomething else </quote>
```

I think it is surprising users when they find the next text are going to the next block, especially when the next block is a table cell or a list (or void blocks).

I think it is better for users to keep the insert text under the same block type, rather than using the next block.

ianstormtaylor:
I'd actually like to have it the other way. Google Docs, Dropbox Paper and Quill.js all do it the other way.

ianstormtaylor:
@zhujinxuan you make good points!

In that case, I feel like we should instead get rid of the "hanging" concept altogether, since you're right it does add lots of complexity. The only way I can think to do this though is to fix triple-click to nudge the selection back into the previous block. (Since that's the main source of hanging selections.)

Do you know if that's possible?

zhujinxuan:
@ianstormtaylor Shall we remove block when we hit delete in a whole selected paragraph like this:

```
<paragraph><anchor/>aaaaaa<focus/></paragraph>
```
If we can delete that paragraph, then I think we can do some methods to avoid the triple selection in Slate.  (I think we can use range.normalize() to solve the problem)

ianstormtaylor:
I think **1.** is okay.

I think **2.** is okay, but it can't be always when hanging. Instead, we need to be able to only nudge when a triple-click has occurred. Because otherwise you won't be able to select "through" that invalid selection if it auto-normalizes you.

For the inserting table case, I think that's not an issue? I think if people are using the "replace or insert" logic they'll be fine.

zhujinxuan:
@ianstormtaylor Great. But then I just come up with another problem:
```
<paragraph><anchor/>aaaaaa<focus/></paragraph>
```
press enter
1. with normalize range
```
<paragraph></paragraph>
<paragraph><cursor/></paragraph>
```
2. without normalize range
```
<paragraph><cursor/></paragraph>
```
Do you think the double paragraph blocks are good?

ianstormtaylor:
Yeah I think it’s fine. It’s kinda unclear what enter should do there, and using delete instead is more intuitive I think.

zhujinxuan:
@ianstormtaylor Great. I will close this PR, and submit another PR with nudging triple-clicks and fixed insertTextAtRange

PR #1742, opened 2018-03-29
Fix hanging selection paste in insertFragmentAtRange

zhujinxuan:
## Problem:
We cannot paste on the hanging selection because deleteAtRange removes the startKey 
## Solution:
Use range.collapseToEnd() if startKey does not exists.


ianstormtaylor:
Thanks @zhujinxuan!

PR #1745, opened 2018-03-30
Feature: setTextByKey and replaceTextByKey

zhujinxuan:
## Problem:
There is not a easy way to replace a length of text

## Fix:
Provides two functions to replace a length of text


ianstormtaylor:
Thanks @zhujinxuan, these look great!

PR #1747, opened 2018-03-30
Fix splitBlock bug with hanging selection

zhujinxuan:
## Problem:
With hanging selection, splitBlock cannot find startKey.

## Fix:
split and then delete.


ianstormtaylor:
Thanks @zhujinxuan!

PR #1748, opened 2018-03-31
[Fix] Copy with empty Block as first-child

jvdsande:
Hi,

I've been using Slate for a personal project recently, and I'm loving it.

I've stumbled across an issue in which a copy starting by an emtpy block (for instance, leaving an empty paragraph between two paragraphs... Or selecting a complete line up to the preceding one) would loose all formating.

Looking into the code, I found that the [data-slate-fragment] attribute would get lost in such case, as it is attached to the first node, but an emtpy node will not get transfered. Anyway, here is the PR fixing it!

Thanks for Slate!


jvdsande:
Fixes https://github.com/ianstormtaylor/slate/issues/1775

ianstormtaylor:
Great call @jvdsande, and nice fix! 

I'm sorry to do this, but as I was merging other PRs it looks like there's a conflict now. Could you fix that? If so, I'm good to merge.

jvdsande:
Hi, I'll look into it as soon as I have the chance, probably on Monday!

jvdsande:
Hi @ianstormtaylor ,

Here is the update PR, in sync with the current `master` branch.

ianstormtaylor:
Thanks @jvdsande!

PR #1753, opened 2018-04-03
Fix spell check bug

zhujinxuan:
I find there is a solution in draft.js by https://github.com/facebook/draft-js/blob/master/src/component/contents/DraftEditorTextNode.react.js#L36

which is to add `data-*` in each layer of marks

This PR fixes the slate spellcheck problem in the same way.

After this fix, renderMark should have props.attributes to append `data-text:true`

Cheers,
Jinxuan Zhu

### DIFF
--- a/examples/rich-text/index.js
+++ b/examples/rich-text/index.js
@@ -302,16 +302,16 @@ class RichTextExample extends React.Component {
    */
 
   renderMark = props => {
-    const { children, mark } = props
+    const { children, mark, attributes } = props
     switch (mark.type) {
       case 'bold':
-        return <strong>{children}</strong>
+        return <strong {...attributes}>{children}</strong>
       case 'code':
-        return <code>{children}</code>
+        return <code {...attributes}>{children}</code>
       case 'italic':
-        return <em>{children}</em>
+        return <em {...attributes}>{children}</em>
       case 'underlined':
-        return <u>{children}</u>
+        return <u {...attributes}>{children}</u>
     }
   }
 }
--- a/packages/slate-react/src/components/leaf.js
+++ b/packages/slate-react/src/components/leaf.js
@@ -5,6 +5,10 @@ import SlateTypes from 'slate-prop-types'
 
 import OffsetKey from '../utils/offset-key'
 
+const attributes = {
+  'data-text': true,
+}
+
 /**
  * Debugger.
  *
@@ -86,7 +90,11 @@ class Leaf extends React.Component {
       index,
     })
 
-    return <span data-offset-key={offsetKey}>{this.renderMarks()}</span>
+    return (
+      <span ref={this.setRef} data-offset-key={offsetKey}>
+        {this.renderMarks()}
+      </span>
+    )
   }
 
   /**
@@ -101,7 +109,16 @@ class Leaf extends React.Component {
     const leaf = this.renderText()
 
     return marks.reduce((children, mark) => {
-      const props = { editor, mark, marks, node, offset, text, children }
+      const props = {
+        editor,
+        mark,
+        marks,
+        node,
+        offset,
+        text,
+        children,
+        attributes,
+      }
       const element = stack.find('renderMark', props)
       return element || children
     }, leaf)
--- a/packages/slate-react/src/components/leaf.js
+++ b/packages/slate-react/src/components/leaf.js
@@ -5,10 +5,6 @@ import SlateTypes from 'slate-prop-types'
 
 import OffsetKey from '../utils/offset-key'
 
-const attributes = {
-  'data-text': true,
-}
-
 /**
  * Debugger.
  *
@@ -90,11 +86,7 @@ class Leaf extends React.Component {
       index,
     })
 
-    return (
-      <span ref={this.setRef} data-offset-key={offsetKey}>
-        {this.renderMarks()}
-      </span>
-    )
+    return <span data-offset-key={offsetKey}>{this.renderMarks()}</span>
   }
 
   /**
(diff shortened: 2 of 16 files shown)


packages/slate-react/src/plugins/after.js
```diff
@@ -305,8 +305,8 @@ function AfterPlugin() {
 
     // Get the selection point.
     const native = window.getSelection()
-    const { anchorNode, anchorOffset } = native
-    const point = findPoint(anchorNode, anchorOffset, value)
+    const { anchorNode } = native
+    const point = findPoint(anchorNode, 0, value)
```
ianstormtaylor: Why this change?

packages/slate-react/src/plugins/after.js
```diff
@@ -323,7 +323,7 @@ function AfterPlugin() {
       leaves.find(r => {
         start = end
         end += r.text.length
-        if (end >= point.offset) return true
+        if (end > point.offset) return true
```
ianstormtaylor: Why this change?

packages/slate-react/src/components/leaf.js
```diff
@@ -86,7 +90,11 @@ class Leaf extends React.Component {
       index,
     })
 
-    return <span data-offset-key={offsetKey}>{this.renderMarks()}</span>
+    return (
+      <span ref={this.setRef} data-offset-key={offsetKey}>
```
ianstormtaylor: Why this `ref=`?

packages/slate-react/src/components/leaf.js
```diff
@@ -101,7 +109,16 @@ class Leaf extends React.Component {
...
+      const props = {
+        editor,
+        mark,
+        marks,
+        node,
+        offset,
+        text,
+        children,
+        attributes,
+      }
```
ianstormtaylor: Can you just define `attributes` inline here instead of pulling it up above, since it's only used in one place?

packages/slate-react/src/components/leaf.js
```diff
@@ -5,6 +5,10 @@ import SlateTypes from 'slate-prop-types'
 
 import OffsetKey from '../utils/offset-key'
 
+const attributes = {
+  'data-text': true,
+}
```
ianstormtaylor: Please see comment below to remove this.

packages/slate-react/src/components/leaf.js
```diff
@@ -5,6 +5,10 @@ import SlateTypes from 'slate-prop-types'
 
 import OffsetKey from '../utils/offset-key'
 
+const attributes = {
+  'data-text': true,
```
ianstormtaylor: Can you rename this to `data-slate-leaf: true` to match the other data attributes we've got?

examples/rich-text/index.js
```diff
@@ -302,16 +302,16 @@ class RichTextExample extends React.Component {
...
+        return <strong {...attributes}>{children}</strong>
       case 'code':
-        return <code>{children}</code>
+        return <code {...attributes}>{children}</code>
       case 'italic':
-        return <em>{children}</em>
+        return <em {...attributes}>{children}</em>
       case 'underlined':
-        return <u>{children}</u>
+        return <u {...attributes}>{children}</u>
```
ianstormtaylor: Can you update the other examples that use `renderMark` so that all of them mix in the `attributes` now?

packages/slate-react/src/plugins/after.js
zhujinxuan: Consider a leaf can be extended by a spell check:
```
<b>hel</b> yourself
```
will be 
```
<b>help</b> yourself
```
after spell check.  Because `anchorOffset` is measured after the spell check.  It will tell that anchorOffset is `4`.  However, during `onInput`, the text is not yet changed.  `offset =4` will tell the slate the spell corrected leaf is `yourself`, not `hel`

packages/slate-react/src/plugins/after.js
zhujinxuan: The same reason of the previous explanation:
Consider a leaf can be extended by a spell check:
```
<b>hel</b> yourself
```
will be
```
<b>help</b> yourself
```
after spell check. Because anchorOffset is measured after the spell check. It will tell that anchorOffset is 4. However, during onInput, the text is not yet changed. offset =4 will tell the slate the spell corrected leaf is yourself, not hel

packages/slate-react/src/components/leaf.js
zhujinxuan: I forgot to remove it. I will remove that.

examples/rich-text/index.js
zhujinxuan: Fixed~

packages/slate-react/src/components/leaf.js
zhujinxuan: Fixed, thank you

zhujinxuan: If we want to define attributes in `reduce` function; how about `data-slate-mark:index`; then we may be able to find leaf-mark by index in future development if necessary.

packages/slate-react/src/components/leaf.js
zhujinxuan: Fixed~

packages/slate-react/src/components/leaf.js
zhujinxuan: Fixed~

zhujinxuan:
Could I know if there is any updates?

ianstormtaylor:
Thanks @zhujinxuan, great find here.

PR #1755, opened 2018-04-04
fix copying void nodes in Chrome

bagilevi:
When the node is a void node, it uses the `ClipboardEvent.clipboardData` API to directly copy the HTML markup to the clipboard.

Fixes https://github.com/ianstormtaylor/slate/issues/1449

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -10,6 +10,7 @@ import EVENT_HANDLERS from '../constants/event-handlers'
 import HOTKEYS from '../constants/hotkeys'
 import Content from '../components/content'
 import cloneFragment from '../utils/clone-fragment'
+import copyVoidNode from '../utils/copy-void-node'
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Hey @bagilevi thanks for taking the time to write this up. Is there a way to fix this without using the `setData` API? I believe that API is not supported in some browsers (IE, not sure about others). Can we solve it by doing something more similar to what we do for regular nodes?

bagilevi:
Hi @ianstormtaylor, if setData is not available, then it should fall back to the original behaviour, at least in  theory, I have only tested it with Chrome and Firefox. I'm not sure how to solve it without setData, you would somehow need to trick Chrome into copying the exact HTML fragment without stripping the wrapping div.

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/commit/f19cde92c735a83d003d3cb4c82a15d9c815bdac! Thank you for investigating this @bagilevi!

PR #1756, opened 2018-04-04
Issues in IE 11

ikorgik:
Fixes for several issues in IE 11 we found during slate integration:
```
Objects are not valid as a React child (found: object with keys {size, _origin, _capacity, _level, _root, _tail, ownerID, hash, __altered}). If you meant to render a collection of children, use an array instead. in span (at text.js:130)
```

```
Unable to get property 'childNodes' of undefined or null reference
```

```
Objects are not valid as a React child (found: List [ [object Object] ]). If you meant to render a collection of children, use an array instead.
```

### DIFF
--- a/packages/slate-html-serializer/src/index.js
+++ b/packages/slate-html-serializer/src/index.js
@@ -121,6 +121,8 @@ class Html {
    */
 
   deserialize = (html, options = {}) => {
+    // Provides default value to make it working in IE 11.
+    html = html || '<p></p>'
     const { toJSON = false } = options
     const { defaultBlock, parseHtml } = this
     const fragment = parseHtml(html)
@@ -327,8 +329,8 @@ class Html {
     const { document } = value
     const elements = document.nodes.map(this.serializeNode).filter(el => el)
     if (options.render === false) return elements
-
-    const html = renderToStaticMarkup(<body>{elements}</body>)
+    // Converts List to Array to support IE 11.
+    const html = renderToStaticMarkup(<body>{elements.toArray()}</body>)
     const inner = html.slice(6, -7)
     return inner
   }
@@ -343,14 +345,17 @@ class Html {
   serializeNode = node => {
     if (node.object === 'text') {
       const leaves = node.getLeaves()
-      return leaves.map(this.serializeLeaf)
+      const children = leaves.map(this.serializeLeaf);
+      // Converts List to Array to support IE 11.
+      return children.toArray()
     }
 
     const children = node.nodes.map(this.serializeNode)
 
     for (const rule of this.rules) {
       if (!rule.serialize) continue
-      const ret = rule.serialize(node, children)
+      // Converts List to Array to support IE 11.
+      const ret = rule.serialize(node, children.toArray())
       if (ret === null) return
       if (ret) return addKey(ret)
     }
--- a/packages/slate-react/src/components/text.js
+++ b/packages/slate-react/src/components/text.js
@@ -124,9 +124,10 @@ class Text extends React.Component {
       return child
     })
 
+    // Converts List to Array to support IE 11.
     return (
       <span data-key={key} style={style}>
-        {children}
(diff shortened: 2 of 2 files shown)


zhujinxuan:
Have you tried `babel-polyfill` in building the web?

I remember there are several includes methods requiring `babel-polyfill` for IE in final build

ikorgik:
Hi @zhujinxuan, didn't try it in this case, but use some other polyfills in my project like: 
es6-shim, some packages from core-js (es7/array, symbol) and polyfill for closest

bsouthga:
would be awesome to get this in, looks like the build error is just from prettier not liking a semi colon.

bsouthga:
for more context, we are seeing the same issues in IE 11 with default `core-js`

packages/slate-html-serializer/src/index.js
```diff
@@ -327,8 +329,8 @@ class Html {
     const { document } = value
     const elements = document.nodes.map(this.serializeNode).filter(el => el)
     if (options.render === false) return elements
-
-    const html = renderToStaticMarkup(<body>{elements}</body>)
+    // Converts List to Array to support IE 11.
```
sethmcleod: Could this and similar comments be updated to say `Compat: Converts List to Array for IE11 support` for consistency with other compatibility comments?

packages/slate-html-serializer/src/index.js
```diff
@@ -343,14 +345,17 @@ class Html {
   serializeNode = node => {
     if (node.object === 'text') {
       const leaves = node.getLeaves()
-      return leaves.map(this.serializeLeaf)
+      const children = leaves.map(this.serializeLeaf);
```
sethmcleod: Looks like this semicolon needs to be removed.

packages/slate-html-serializer/src/index.js
```diff
@@ -121,6 +121,8 @@ class Html {
    */
 
   deserialize = (html, options = {}) => {
+    // Provides default value to make it working in IE 11.
```
sethmcleod: Could we update this comment to something along these lines: `Compat: Provide a default value for IE11 support`

ianstormtaylor:
Hey @ikorgik, thank you for this. Could you make the changes to the comments to append `COMPAT:` to them, to match the rest of the codebase.

For the serializer change, does it need to be `'<p></p>
` or can it be just an empty string `''`? I'd rather not force the paragraph default on folks to be hardcoded here.

terenceodonoghue:
I'm hitting the same problem in IE11

`Unable to get property 'childNodes' of undefined or null reference`

Is there an interim fix that I might be able to use in the meantime?

enzoferey:
Hi, we are facing issues in IE11 as well. We are using the slate-html-serializer, which we could fix it problems by having a default initial state of `"<p></p>"` instead of `""`. Regarding Slate itself, we have noticed that the placeholder prop was introducing errors. We will be working on fixing everything along next week and I will update you here based on our advancements. Have a nice weekend and thanks for this amazing library ! 🙌

enzoferey:
One of my team mates reported me this:

"The problem is that IE11 lets users select placeholder text, which is disabled with css on other browsers.
By selecting I mean it should have pointer-events disabled"

Dunno if that is helpful for you in any way. I will take a look at it pretty soon anyways ;)

danielsarin:
I think I managed to track down the `Unable to get property 'childNodes' of undefined or null reference` error and submitted a pull request. But for now you can also fix it by supplying the modified parser as a custom parseHtml to the serializer. Ex.

```
const parseHtml = (html) => {
  const parsed = new DOMParser().parseFromString(html, 'text/html')
  const { body } = parsed
  // COMPAT: in IE 11 body is null if html is an empty string
  return body || window.document.createElement('body')
}

const html = new Html({ rules, parseHtml });
```

ianstormtaylor:

PR #1760, opened 2018-04-05
Extract hotkey and environment detection into separate packages

justinweiss:
Overriding default behavior in plugins can be hard, because you have to match all of the keybindings that Slate uses. By exporting hotkeys as its own package, both core Slate plugins and custom plugins can use the same key detection logic.

This is a first stab at pulling this logic out -- let me know what you think!


ianstormtaylor:
@justinweiss this looks great! Thank you for doing this and for carrying over the changes to meticulously.

I'm thinking we should maybe name the hotkeys functions `Hotkeys.is*` with the `is` prefix to parallel some of the other boolean logic we've got. Are you okay to make that change?

justinweiss:
@ianstormtaylor Yep! Not a problem.

tobiasandersen:
This is awesome @justinweiss, looking great!

ianstormtaylor:
Thank you @justinweiss, this PR is amazing.

PR #1763, opened 2018-04-07
Node use a parent map for performance

Soreine:
For fun, I wanted to explore a different approach for node finding and parent/ancestors computations on Nodes, and start a discussion around it.

The idea is to implement a general-purpose method `node.getParentMap` that returns an `OrderedMap<key, parentNode>` of all their descendants. Then a lot of other methods reuse this map and become trivial or more efficient.

I was running the benchmarks, but then gave up on them once I realized that running the same benchmark twice could randomly show ±80% changes with no reason.

What I like about this approach is that we can memoize this single method, and skip memoization for a lot of methods that become trivial after relying on `getParentMap`.

`getParentMap` could return a `Map` instead of `OrderedMap`, which has better performance, but then `getKeysAsArray` cannot benefit from it.

### DIFF
--- a/benchmark-results.txt
+++ b/benchmark-results.txt
@@ -1,0 +1,67 @@
+
+  benchmarks
+    html-serializer
+      deserialize
+        7.52 → 7.59 ops/sec
+      serialize
+        799.19 → 770.50 ops/sec
+    plain-serializer
+      deserialize
+        254.29 → 243.54 ops/sec
+      serialize
+        6361.44 → 4317.70 ops/sec
+    rendering
+      normal
+        76.80 → 57.52 ops/sec
+    changes
+      delete-backward
+        33566.02 → 26689.58 ops/sec
+      delete-forward
+        4732.58 → 3950.77 ops/sec
+      insert-text-by-key-multiple
+        65.16 → 92.48 ops/sec (42% faster)
+      insert-text-by-key
+        871.62 → 572.61 ops/sec (34% slower)
+      insert-text
+        1095.37 → 952.57 ops/sec
+      normalize
+        71389.81 → 75804.18 ops/sec
+      split-block
+        63.31 → 76.51 ops/sec
+    models
+      from-json-big
+        2.19 → 2.22 ops/sec
+      from-json
+        115.77 → 117.47 ops/sec
+      get-blocks-at-range
+        24593.08 → 24263.87 ops/sec
+      get-blocks
+        573926.72 → 547490.89 ops/sec
+      get-characters-at-range
+        11343.84 → 10889.83 ops/sec
+      get-characters
+        11810.52 → 11572.59 ops/sec
+      get-inlines-at-range
+        161176.72 → 169514.27 ops/sec
+      get-inlines
+        1017884.49 → 925047.30 ops/sec
+      get-leaves
+        11034836.47 → 10723100.04 ops/sec
+      get-marks-at-range
+        1391.25 → 1243.05 ops/sec
+      get-marks
+        1126.99 → 1345.93 ops/sec
+      get-path
+        429087.40 → 554933.62 ops/sec
+      get-texts-at-range
+        311569.89 → 265181.51 ops/sec
+      get-texts
+        662274.94 → 501818.13 ops/sec
+      has-node-multiple
+        72346.99 → 60184.15 ops/sec
+      has-node
+        565016.94 → 560211.79 ops/sec
+      to-json
+        2604.46 → 2289.03 ops/sec
+      update-node
+        14278.60 → 12166.05 ops/sec
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1,7 +1,7 @@
 import direction from 'direction'
 import isPlainObject from 'is-plain-object'
 import logger from 'slate-dev-logger'
-import { List, OrderedSet, Set } from 'immutable'
+import { List, OrderedSet, Set, Map } from 'immutable'
 
 import Block from './block'
 import Data from './data'
@@ -198,6 +198,7 @@ class Node {
     first = assertKey(first)
     second = assertKey(second)
 
+    // TODO optimize getKeysAsArray
     const keys = this.getKeysAsArray()
     const firstIndex = keys.indexOf(first)
     const secondIndex = keys.indexOf(second)
@@ -244,6 +245,7 @@ class Node {
 
   /**
    * Assert that a node's tree has a node by `key` and return it.
+   * Includes the node itself
    *
    * @param {String} key
    * @return {Node}
@@ -350,20 +352,21 @@ class Node {
     key = assertKey(key)
 
     if (key == this.key) return List()
-    if (this.hasChild(key)) return List([this])
 
-    let ancestors
-    this.nodes.find(node => {
-      if (node.object == 'text') return false
-      ancestors = node.getAncestors(key)
-      return ancestors
-    })
+    const parentMap = this.getParentMap()
 
-    if (ancestors) {
-      return ancestors.unshift(this)
-    } else {
+    let parent = parentMap.get(key)
+    if (!parent) {
       return null
     }
+
+    const ancestors = []
+    while (parent) {
+      ancestors.unshift(parent)
+      parent = parentMap.get(parent.key)
+    }
+
+    return List(ancestors)
   }
 
   /**
@@ -586,33 +589,27 @@ class Node {
   /**
    * Get the common ancestor of nodes `one` and `two` by keys.
    *
-   * @param {String} one
-   * @param {String} two
+   * @param {String} a
+   * @param {String} b
    * @return {Node}
    */
 
-  getCommonAncestor(one, two) {
-    one = assertKey(one)
-    two = assertKey(two)
+  getCommonAncestor(a, b) {
+    a = assertKey(a)
+    b = assertKey(b)
 
-    if (one == this.key) return this
-    if (two == this.key) return this
+    if (a == this.key) return this
+    if (b == this.key) return this
 
-    this.assertDescendant(one)
-    this.assertDescendant(two)
-    let ancestors = new List()
-    let oneParent = this.getParent(one)
-    let twoParent = this.getParent(two)
+    const aAncestors = this.getAncestors(a)
+    const bAncestors = this.getAncestors(b)
 
-    while (oneParent) {
-      ancestors = ancestors.push(oneParent)
-      oneParent = this.getParent(oneParent.key)
+    if (!aAncestors || !bAncestors) {
+      // Fail
+      return this.assertDescendant(a) && this.assertDescendant(b)
     }
 
-    while (twoParent) {
-      if (ancestors.includes(twoParent)) return twoParent
-      twoParent = this.getParent(twoParent.key)
-    }
+    return aAncestors.findLast(ancestor => bAncestors.contains(ancestor))
   }
 
   /**
@@ -651,20 +648,10 @@ class Node {
 
   getDescendant(key) {
     key = assertKey(key)
-    let descendantFound = null
 
-    const found = this.nodes.find(node => {
-      if (node.key === key) {
-        return node
-      } else if (node.object !== 'text') {
-        descendantFound = node.getDescendant(key)
-        return descendantFound
-      } else {
-        return false
-      }
-    })
-
-    return descendantFound || found
+    const parentMap = this.getParentMap()
+    const parent = parentMap.get(key)
+    return parent && parent.getChild(key)
   }
 
   /**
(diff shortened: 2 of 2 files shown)


zhujinxuan:
Great idea, this is what I want for a long time! 
Could you post some benchmark test result?

Soreine:
As I said, I could, but benchmarks look unreliable to me for now.
Let's see what comes out of https://github.com/ianstormtaylor/slate/pull/1765

zhujinxuan:
Hi, I am currently working on the same file:
https://github.com/ianstormtaylor/slate/pull/1771

I do not find logical conflicts between our two PRs(thought there would be some conflicts in the memorize part).  But to ensure there is really no conflicts, could you have a look to ensure that our changed functions would not influence each other?

zhujinxuan:
I think we shall use a chain-list for getAncestors.  Calling getAncestors includes a lot of insert (and construction new List by List().unshift()).  I am not sure what happens to the real performance of that, but, in benchmark tests, they are slow (theoretically it is O(logN), but see https://github.com/facebook/immutable-js/issues/584#issuecomment-131883133 and https://hypirion.com/musings/understanding-persistent-vector-pt-1).

benjycui:
I don't think this PR will improve the performance.

Before this PR, we don't need to traverse the whole tree to get parent of a node(only edge case need it). But after this PR, we must traverse the whole tree, even in this use case:

```jsx
<value>
  <document>
    <paragraph>
      ...
    </paragraph>
    ...
  </document>
<value>

document.getParent(paragraph)
```

benjycui:
This PR may improve performance of some use cases, but it will slow down more use cases, too.

So, I don't think that it is a good solution.

Soreine:
> Before this PR, we don't need to traverse the whole tree to get parent of a node

If you're lucky, yes. But on average you need to traverse half the tree. I would even say that, statistically, when editing, you're most likely to edit nodes at the end of the document, which would push the average closer to traversing the whole tree most of the time.

If we always compute the parent map, we are being 2 times slower on average, because we always traverse the whole tree. But all future method calls can then have access to that memoized parent map. So the next calls to `getParent` for example, even for different keys, will only have to do `parentMap.get(key)` which is `O(log32 N)` instead of traversing the tree again which is `O(N)`.

image

So this PR should improve performance overall, by adding some cost upfront. It should:

- Reduce performance for single calls on a particular value. This is not a big deal, since doing a single call only takes a few milliseconds anyway.
- Increase performance for several calls on the same value (or values that share a lot of identical nodes).

Soreine:
I have run @zhujinxuan new benchmarks...

```diff
  benchmarks
    changes
      delete-backward
        user : 462.55 → 370.12 ops/sec
        hr : 746.15 → 598.01 ops/sec
      delete-forward
        user : 141.86 → 119.07 ops/sec
        hr : 197.00 → 168.89 ops/sec
      insert-text-by-key-multiple
+        user : 68.47 → 122.27 ops/sec (44% faster)
+        hr : 80.80 → 156.86 ops/sec (48% faster)
      insert-text-by-key
        user : 555.40 → 694.99 ops/sec
        hr : 775.49 → 1015.33 ops/sec
      insert-text
-        user : 2331.45 → 1538.00 ops/sec (52% slower)
-        hr : 3382.93 → 2035.99 ops/sec (66% slower)
      normalize
        user : 651.39 → 803.23 ops/sec
        hr : 883.62 → 1111.58 ops/sec
      split-block
        user : 327.76 → 400.85 ops/sec
        hr : 431.63 → 541.32 ops/sec
    models
      from-json-big
        user : 4.52 → 4.68 ops/sec
        hr : 4.71 → 4.92 ops/sec
      from-json
        user : 175.53 → 209.49 ops/sec
        hr : 222.36 → 273.70 ops/sec
      get-active-marks-at-range
        user : 317.50 → 368.49 ops/sec
        hr : 449.51 → 467.64 ops/sec
      get-blocks-at-range
        user : 781.68 → 989.03 ops/sec
        hr : 1441.46 → 1700.61 ops/sec
      get-blocks
+        user : 5145.54 → 8294.02 ops/sec (38% faster)
+        hr : 8961.49 → 12672.82 ops/sec (29% faster)
      get-characters-at-range
        user : 218.58 → 268.61 ops/sec
        hr : 292.59 → 341.02 ops/sec
      get-characters
        user : 3966.61 → 4085.31 ops/sec
        hr : 5413.69 → 5334.25 ops/sec
      get-inlines-at-range
+        user : 676.48 → 1459.90 ops/sec (54% faster)
+        hr : 828.10 → 1972.26 ops/sec (58% faster)
      get-inlines
        user : 10713.96 → 10552.63 ops/sec
        hr : 17128.01 → 17359.54 ops/sec
      get-leaves
        user : 9522.91 → 9213.89 ops/sec
        hr : 17629.79 → 16738.93 ops/sec
      get-marks-at-range
        user : 209.06 → 202.95 ops/sec
        hr : 278.97 → 257.19 ops/sec
      get-marks
        user : 660.75 → 720.29 ops/sec
        hr : 854.61 → 933.75 ops/sec
      get-path
        user : 2356.74 → 2142.12 ops/sec
        hr : 3228.52 → 2843.75 ops/sec
      get-texts-at-range
-        user : 3970.62 → 2248.78 ops/sec (77% slower)
-        hr : 5928.97 → 2956.28 ops/sec (101% slower) 😟
      get-texts
        user : 10403.75 → 9926.92 ops/sec
        hr : 16988.23 → 15712.33 ops/sec
      has-node-multiple
        user : 2098.78 → 2601.75 ops/sec
        hr : 2560.69 → 3265.02 ops/sec
      has-node
-        user : 8150.63 → 3271.16 ops/sec (149% slower) 😟
-        hr : 12856.79 → 4253.41 ops/sec (202% slower) 😟
      to-json
        user : 949.26 → 924.66 ops/sec
        hr : 1299.58 → 1236.17 ops/sec
      update-node
-        user : 3375.53 → 2387.29 ops/sec (41% slower)
-        hr : 4786.83 → 3146.95 ops/sec (52% slower)
    html-serializer
      deserialize
        user : 9.29 → 8.32 ops/sec
        hr : 10.21 → 9.03 ops/sec
      serialize
        user : 306.36 → 362.49 ops/sec
        hr : 437.62 → 506.03 ops/sec
    plain-serializer
      deserialize
        user : 472.93 → 452.47 ops/sec
        hr : 511.15 → 497.17 ops/sec
      serialize
        user : 2747.93 → 2727.83 ops/sec
        hr : 3352.30 → 3277.58 ops/sec
    rendering
      normal
        user : 82.52 → 76.63 ops/sec
        hr : 101.56 → 93.15 ops/sec
```

The results are not there. Looking into this one specifically:

```diff
      has-node-multiple
        user : 2098.78 → 2601.75 ops/sec
        hr : 2560.69 → 3265.02 ops/sec
      has-node
-        user : 8150.63 → 3271.16 ops/sec (149% slower) 😟
-        hr : 12856.79 → 4253.41 ops/sec (202% slower) 😟
```

`has-node` is a lot slower because it needs to build the parent map.
But `has-node-multiple` should be way faster, after having built the parent map once. It is not that fast though... I need to do more research

Soreine:
```diff
  benchmarks
    changes
      delete-backward
        user : 462.55 → 449.65 ops/sec
        hr : 746.15 → 692.48 ops/sec
      delete-forward
        user : 141.86 → 132.96 ops/sec
        hr : 197.00 → 192.18 ops/sec
      insert-text-by-key-multiple
+        user : 68.47 → 127.82 ops/sec (46% faster)
+        hr : 80.80 → 171.74 ops/sec (53% faster)
      insert-text-by-key
+        user : 555.40 → 820.74 ops/sec (32% faster)
+        hr : 775.49 → 1226.87 ops/sec (37% faster)
      insert-text
-        user : 2331.45 → 1504.49 ops/sec (55% slower)
-        hr : 3382.93 → 2150.45 ops/sec (57% slower)
      normalize
        user : 651.39 → 803.43 ops/sec
        hr : 883.62 → 1087.30 ops/sec
      split-block
+        user : 327.76 → 462.33 ops/sec (29% faster)
+        hr : 431.63 → 648.28 ops/sec (33% faster)
    models
      from-json-big
        user : 4.52 → 4.67 ops/sec
        hr : 4.71 → 4.88 ops/sec
      from-json
        user : 175.53 → 201.11 ops/sec
        hr : 222.36 → 254.81 ops/sec
      get-active-marks-at-range
+        user : 317.50 → 2749.18 ops/sec (88% faster)
+        hr : 449.51 → 4158.74 ops/sec (89% faster)
      get-blocks-at-range
+        user : 781.68 → 2840.59 ops/sec (72% faster)
+        hr : 1441.46 → 3910.59 ops/sec (63% faster)
      get-blocks
+        user : 5145.54 → 9593.77 ops/sec (46% faster)
+        hr : 8961.49 → 14101.32 ops/sec (36% faster)
      get-characters-at-range
+        user : 218.58 → 3745.46 ops/sec (94% faster)
+        hr : 292.59 → 5004.47 ops/sec (94% faster)
      get-characters
        user : 3966.61 → 4318.90 ops/sec
        hr : 5413.69 → 5263.51 ops/sec
      get-inlines-at-range
+        user : 676.48 → 2862.92 ops/sec (76% faster)
+        hr : 828.10 → 4448.66 ops/sec (81% faster)
      get-inlines
        user : 10713.96 → 11699.45 ops/sec
        hr : 17128.01 → 19913.59 ops/sec
      get-leaves
        user : 9522.91 → 11672.89 ops/sec
        hr : 17629.79 → 17964.79 ops/sec
      get-marks-at-range
+        user : 209.06 → 2728.10 ops/sec (92% faster)
+        hr : 278.97 → 4098.63 ops/sec (93% faster)
      get-marks
        user : 660.75 → 712.19 ops/sec
        hr : 854.61 → 959.57 ops/sec
      get-path
+        user : 2356.74 → 3276.68 ops/sec (28% faster)
+        hr : 3228.52 → 4835.35 ops/sec (33% faster)
      get-texts-at-range
        user : 3970.62 → 4005.19 ops/sec
        hr : 5928.97 → 5255.94 ops/sec
      get-texts
        user : 10403.75 → 11235.53 ops/sec
        hr : 16988.23 → 19409.07 ops/sec
      has-node-multiple
+        user : 2098.78 → 4724.41 ops/sec (56% faster)
+        hr : 2560.69 → 6144.56 ops/sec (58% faster)
      has-node
-        user : 8150.63 → 5589.01 ops/sec (46% slower)
-        hr : 12856.79 → 7955.02 ops/sec (62% slower)
      to-json
        user : 949.26 → 1054.04 ops/sec
        hr : 1299.58 → 1417.61 ops/sec
      update-node
        user : 3375.53 → 3778.33 ops/sec
        hr : 4786.83 → 5258.33 ops/sec
    html-serializer
      deserialize
        user : 9.29 → 9.00 ops/sec
        hr : 10.21 → 9.76 ops/sec
      serialize
        user : 306.36 → 347.80 ops/sec
        hr : 437.62 → 501.82 ops/sec
    plain-serializer
      deserialize
        user : 472.93 → 443.71 ops/sec
        hr : 511.15 → 480.37 ops/sec
      serialize
        user : 2747.93 → 3002.24 ops/sec
        hr : 3352.30 → 3722.05 ops/sec
    rendering
      normal
        user : 82.52 → 75.36 ops/sec
        hr : 101.56 → 91.60 ops/sec
```

If I just use a `Map` instead of `OrderedMap`, it seems better. Since we found that we can get rid of `getKeysAsArray`, that might be a good approach.

Soreine:
I haven't pull master for this benchmark :) I'll do when I have the time.

zhujinxuan:
To some extent, I agree with @benjycui 's concern that `O(n*h)` is not good.

There are two things in concern:
1. immutable list is with log32(N) complexity, but when N is small, the time consuming (at least in this stage) is larger than array operation.  Indeed, the O(N)~O(log32(N)) comparison is meaningful when the first layer children is bigger than 32 (this.nodes.size > 32, which happens not so often)
2. It will slow down the loading process.  In the first load, people often call change.normalize().  It gives a O(n*h) complexity to built the parent map.


I think we can consider tempting with __cache; when a node is considered as a hot point, build the parentMap of that node with Promise.  
However, there is another problem about benchmark tests.  Current benchmark test is just to test one change function, then remove the memory.  If we want to see the power of parentMap, we shall find a way to mock real editing with different change function. 

@Soreine I have a different idea about getAncestors and getParent, which causes some conflicts between our two PRs.  

The following two commits shows my current idea of getParent and getDescendant: 
https://github.com/ianstormtaylor/slate/pull/1771/commits/fe51443f004dfe57a6c4af40740c31f5f3514014 https://github.com/ianstormtaylor/slate/pull/1771/commits/d72e7d4d4c597736bcbc40ff4e2d390c304b1abc

I am thinking about a path prior method to reconstruct the models/node.js, and it turns out a significant improvement when switching to getPath(AsString).  Perhaps it is a bit better to consider about a method like `set(__cache, 'getPathAsString' , {children: "all"})` to improve performance of hot point.

Issue #1291, opened 2017-10-26
Highlight example shows wrong highlight after delete


twobit:
#### Do you want to request a *feature* or report a *bug*?

bug

#### What's the current behavior?

In the highlight example, type a word in the search box, then delete that word (or change it slightly) in the paragraph. The highlight will still be visible applied to the wrong characters.

![search_slate](https://user-images.githubusercontent.com/2778/32033021-f08df962-b9be-11e7-93b8-1dc84fb46056.gif)

#### What's the expected behavior?

The highlight should be removed if the text changes. I think only a minor tweak to the [onInputChange]( https://github.com/ianstormtaylor/slate/blob/5b2e53039a3d905c497b0f3bd603aa60361dd702/examples/search-highlighting/index.js#L54) method is needed. When I have some extra time this week I can look into it.

ianstormtaylor:
Ah good call. Actually if we just extracted the body of `onInputChange` out into an `updateDecorators` method, and called it from both `onInputChange` and `onChange`, I think that would fix it.

ianstormtaylor:
I just realized...

This is due to `state.decorations` not being updated in the same way that `state.selection` is updated in `./operations/apply.js`. Every time an operation is applied, the logic checks to see if the selection range needs to be shifted/removed based on what was inserted/removed. If the same thing happened for decorations this would be fixed.

I think we should fix it at the source, instead of in the example.

roundrobin:
@ianstormtaylor any update on this bug?

ianstormtaylor:
Not from me personally, because I’m not affected by it right now. But if anyone else wanted to PR a fix that would be awesome!

PR #1777, opened 2018-04-15
slate-hyperscript tests and decorations

jasonphillips:
This PR first adds tests to `slate-hyperscript`, covering all its basic functionality as well as its handling of selection (by cursor, and by anchor / focus).

It then adds the ability to similarly notate decorations in hyperscript, using either a syntax similar to marks (if it's a simple range within a continuous section of text), or using a `[decoration]Anchor` / `[decoration]Focus` syntax.

For the simple syntax, [see this test](https://github.com/jasonphillips/slate/blob/hyperscript-tests-and-decorations/packages/slate-hyperscript/test/decorations/single.js)

For the Anchor / Focus syntax (must supply matching key since can be multiple), [see this test](https://github.com/jasonphillips/slate/blob/hyperscript-tests-and-decorations/packages/slate-hyperscript/test/decorations/across-blocks.js)

I dove into these improvements mainly as a preparation for being able to more rigorously test decorations (I'm working on solving https://github.com/ianstormtaylor/slate/issues/1291 ), which is greatly helped when `slate-hyperscript` can easily include decorations in its readable notation.


jasonphillips:
I just noticed @Soreine 's PR on slate-hyperscript https://github.com/ianstormtaylor/slate/pull/1379 -- after looking through it, however, none of those changes will conflict with these. The couple of inline tests he added can also very easily be incorporated into the nested directory of tests here upon merging these two PRs.

packages/slate-hyperscript/test/decorations/across-blocks.js
```diff
@@ -0,0 +1,87 @@
...
+  decorators: {
+    highlight: 'highlight',
+  },
+})
+
+export const input = (
+  <value>
+    <document>
+      <paragraph>
+        This is one <highlightAnchor key="a" />block.
```
Soreine: What are the decoration `key` used for? 
Cannot you use the decoration `key` as a way to pair the start/end of the decoration?

If not, I would suggest the following syntax:

```jsx
<highlight anchor key="a"/>
<highlight focus key="a" />
```

jasonphillips:
```js
<highlight anchor key="a"/>
<highlight focus key="a"/>
```

Thanks, that's a much better syntax suggestion, and I've updated to match it. 

These two updated tests help illustrate why the `key` attribute is needed to distinguish overlapping versus nested decorations in complex cases:
  - [overlapping](https://github.com/jasonphillips/slate/blob/2ddebd36d5cfb32c781ce1c9b3aae1628ac4536c/packages/slate-hyperscript/test/decorations/overlapping.js) versus
  - [nested](https://github.com/jasonphillips/slate/blob/2ddebd36d5cfb32c781ce1c9b3aae1628ac4536c/packages/slate-hyperscript/test/decorations/nested.js)

jasonphillips:
Huh, the CI build failed because of this error in upath (on the latest version of node) which suddenly hit us, unrelated to the PR:

https://github.com/anodynos/upath/issues/14

Soreine:
Well it looks like we can still tell which decoration edges pair together,
just using the key. Following the document order, the first with key a is
the anchor, the second with key a is the focus

>

Soreine:
@jasonphillips do you know what should be done about `upath` ?
Should we tell Travis to not use Node 10 ? Or should we tell Travis to do `yarn install --ignore-engines` as suggested in the upath issue ?

jasonphillips:
I suppose since this is really a browser-focused library (except perhaps for the de/serialization packages that I do use on the server a bit), "officially" testing it on cutting-edge node versions in CI isn't necessarily something that matters? In which case, just setting Travis to use the prior stable Node certainly seems reasonable to me for now.

ianstormtaylor:
@jasonphillips this is absolutely amazing. These tests look great, and the ability to model decorators with hyperscript is super smart!

I didn't realize that the two PRs would conflict, and I've already merged the other. Would you be able to resolve the conflicts and then I can merge this?

Thanks so much for your help.

jasonphillips:
No problem -- latest master merged & conflicts resolved. I moved the 3 inline tests from the other PR into the test directories of this one.

ianstormtaylor:
@jasonphillips thank you! Moving those is above and beyond, thank you very much.

PR #1778, opened 2018-04-16
Operations update decorations

jasonphillips:
Note: this PR depends upon the changes in https://github.com/ianstormtaylor/slate/pull/1777 (it makes testing decorations via hyperscript possible) and is intended to be merged after that one. It therefore includes its commits, but you can see just the relevant diff from that prior point here: https://github.com/jasonphillips/slate/compare/08d0cb25f4685fdd5b683dad0944c977f790c3d0...fc0ca10bc161e33c16828647ee46f0880d4efacc

This PR handles two limitations / problems of the prior decorations:

  1. Decorations were not updated upon operations in the way that selection is updated, which means that their ranges quickly go out of sync while typing / backspacing / etc. See https://github.com/ianstormtaylor/slate/issues/1291 

   2. Decorations had no way to designate an 'atomic' type that would be invalidated when the internal content changes, which seems necessary for something like that search demo and other use cases.

This PR fixes (**1**) by having operations apply the same logic to decorations as they do to selections (both are just ranges as far as it is concerned), and then makes (**2**) possible by establishing an `atomic` property that can be optionally set on decorations. 

The results on the search demo make the case for both of these fixes (updating in sync with operations, and allow atomic decorations):

kapture 2018-04-16 at 0 27 29

Tests have also been added for the new decorations behavior. I believe this solution makes good sense and handles my use case, but am interested in your reaction.


ianstormtaylor:
@jasonphillips this PR is absolutely incredible. Genius idea to have the ranges themselves be atomic too. As soon as we merge the other one I'm good to merge this.

Can you change the `atomic` to `isAtomic` to match the conventions elsewhere? Thank you so much!

zhujinxuan:
A question: what is atomic range? do you mean decorations are not overlapped?

jasonphillips:
@zhujinxuan In this case, 'atomic' is meant to indicate decoration ranges that are irreducible as a single unit, so that if their contents would change they are just invalidated. Analogous to [immutable entities in draftjs](https://github.com/facebook/draft-js/blob/35b3605f1b7a6714be0159f2b111eadc854963ac/docs/Advanced-Topics-Entities.md#immutable), which are also described as atomic.

@ianstormtaylor I've updated this branch now to resolve conflicts, good to see Travis going green again

zhujinxuan:
Another question, though I think it is perhaps occurs rare:

What will happen if we have 
```
<paragraph>
   <text-key = 1>1</text>
   <mark b><text-key = 1> Something </text></mark>
   <text-key = 1>1</text>
</paragraph>
```
with decoration `{startKey:1, startOffset:1, endKey:1, endOffset:3}`; as the tags are not closed.

jasonphillips:
Oh -- the PR adds `isAtomic` as a boolean option to Ranges (and therefore decorations) which still defaults to false, so it's just an opt-in for those cases that need it and otherwise doesn't change current behavior of mutability of ranges.

jasonphillips:
@zhujinxuan 

> What will happen if we have [...]

I can't quite figure out the last example you posted -- are you saying it would be difficult to notate the decorations in this case in hyperscript, or that there would be some issue updating a range in operations?

zhujinxuan:
@jasonphillips Yes.  I mean this PR is great, but I am wondering if there would be a way to annotate decoration in a case that decoration is in same text but crossing different leaves.

jasonphillips:
@zhujinxuan 

Oh - I included a way to match up two distant (self-closed) decoration tags by manually supplying a matching `key` prop on each, similar to using the `<anchor/>` and `<focus/>` endpoints for a selection; see one typical use in the tests here:

https://github.com/jasonphillips/slate/blob/operations-update-decorations/packages/slate-hyperscript/test/decorations/across-blocks.js

So cases like your example above should look something like:

```js
<paragraph>
    <mydecoration  key="a"/>some text before 
    <bold> the decoration ends<mydecoration  key="a"/> inside this mark</bold>
    some text after
</paragraph>
```

zhujinxuan:
@jasonphillips Great! Thank you.

ianstormtaylor:
Thank you so much for this @jasonphillips! Sorry for the delay here.

RXminuS:
@jasonphillips @ianstormtaylor You guys are legends! Just ran into this problem...and it's already fixed! This project is feeling more amazing by the minute :-)

PR #1779, opened 2018-04-16
Ensure cursor is shown after delete hanging selection

zhujinxuan:
Problem:
We cannot see a cursor blink after delete hanging selection with backspace.

Before fix:

After fix:


ianstormtaylor:
Nice one, thank you @zhujinxuan!!

PR #1780, opened 2018-04-16
Fix cursor restore in removeNodeByKey

zhujinxuan:
## Problem
Cursor is not restored after removeNodeByKey.  

## Solution
This is simply solved by moving the cursor restore logic from operation part to the change.removeNodeByKey part.

There is perhaps another solution by passing previousSelection to operations, but I do not want to change insert node logic.

Perhaps we will no longer need snapshotSelection in deleteAtRange any more.

### DIFF
--- a/packages/slate/src/changes/by-key.js
+++ b/packages/slate/src/changes/by-key.js
@@ -330,8 +330,52 @@ Changes.removeNodeByKey = (change, key, options = {}) => {
   const normalize = change.getFlag('normalize', options)
   const { value } = change
   const { document } = value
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
@zhujinxuan I'm not sure we can do this, because handling the selection tweaks at the operation level is necessary for collaborative editing, where the change functions aren't called, and instead the operations are applied directly.

zhujinxuan:
@ianstormtaylor We cannot fix it in this way.  I will submit another PR with another way to fix it.  So I close this one

PR #1781, opened 2018-04-16
Update debug package in package.json

zhujinxuan:
This PR updates the `debug` package in package.json.

According to `https://github.com/visionmedia/debug/blob/master/CHANGELOG.md`, the newer debug package has 256 color, better browser compatability and better format.


ianstormtaylor:
Thanks @zhujinxuan!

PR #1782, opened 2018-04-16
Add polyfills to example site for IE11 support

sethmcleod:
This PR adds the `babel-polyfill` and `element-closest` polyfills to the example site so the page will load in IE11.

I remember there being a refactor related to bringing your own polyfills #1252 but I don't believe that included the example site build; correct me if I'm wrong though.

Here is a quick screen recording showing the example site working in IE11 with no console errors 🎉

https://www.useloom.com/share/f29d1dbbbf804314bca20de34aff679f


ianstormtaylor:
Hey @sethmcleod, thanks for pointing this out! I've added the Polyfill.io polyfills to the example now on `master`, which lets us avoid messing with Webpack.

sethmcleod:
Nice, that works too!

Opty1712:
image

it doesn't work in IE11 now

Issue #1768, opened 2018-04-10
Pasting very huge document freezes the editor, taking more than 10 minutes to finally paste


chitezh:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Pasting a very huge document like the one contained in this file [https://www.dropbox.com/s/wpefhhwrivvajvc/huge.text?dl=0](https://www.dropbox.com/s/wpefhhwrivvajvc/huge.text?dl=0) on slatejs.org demo app freezes the app. I didn't see the need to make a jsfiddle example as it's happening on slate demo

#### What's the expected behavior?
It appears within seconds like it does on facebook.com, draft.js, http://prosemirror.net/, https://quilljs.com/

CameronAckermanSEL:
Was this using the paste html example?

chitezh:
I tried it on richtext and huge document examples(clearing existing contents first).

Note, I pasted the content on the linked file [https://www.dropbox.com/s/wpefhhwrivvajvc/huge.text?dl=0](https://www.dropbox.com/s/wpefhhwrivvajvc/huge.text?dl=0) not the dropbox page

CameronAckermanSEL:
I don't know that those examples have paste handlers, so maybe check what happens with the Paste HTML example because that's specifically implemented to showcase the paste handling and how that works.

It might have the same performance problems, but it also might not.

chitezh:
Just tried pasting on `Paste HTML` example; same result

zhujinxuan:
I do not find a easy way to solve the problem.  It seems we have to async the serializer to solve the problem.

Slapbox:
Woah yeah that document destroys my editor. It's been several minutes now since the cursor even blinked. I eventually had to restart, as I have no idea how much longer that might have gone on for.

zhujinxuan:
I have a hypothesis.  The slate creates a list of characters when creating texts.  If we change the structure of text, use leaf rather than characters as the base of text.  It would solve the problem.

However, I would begin on performance issue when and only when my finishing https://github.com/ianstormtaylor/slate/pull/1765

benjycui:
The bottle neck is `node.mapDescendants`, not serializer. See: https://github.com/ianstormtaylor/slate/pull/1783

zhujinxuan:
@benjycui Do you mean the paste problem is fixed after that PR?

chitezh:
Thanks, will do

chitezh:
[...]

PR #1783, opened 2018-04-17
refactor: remove useless .indexOf to improve perf

benjycui:
Ref: https://github.com/ianstormtaylor/slate/issues/1768

When we try to paste something into editor, the call stack:

```bash
plugin.onPaste => change.insertFragment => change. insertFragmentAtRange => node.mapDescendants
```

We need `node.mapDescendants` to re-generate keys. But the implementation of `node.mapDescendants` is extremely slow:

https://github.com/ianstormtaylor/slate/blob/a943eada8534ccdff05a770ad84ac599bd7b3682/packages/slate/src/models/node.js#L1854-L1868

If we paste a long list, the algorithm complexity will be almost O(n^2), for we need to `.forEach` n nodes and `.indexOf` n nodes.

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1854,13 +1854,12 @@ class Node {
   mapDescendants(iterator) {
     let { nodes } = this
 
-    nodes.forEach((node, i) => {
+    nodes.forEach((node, index) => {
       let ret = node
       if (ret.object != 'text') ret = ret.mapDescendants(iterator)
-      ret = iterator(ret, i, this.nodes)
+      ret = iterator(ret, index, this.nodes)
       if (ret == node) return
 
-      const index = nodes.indexOf(node)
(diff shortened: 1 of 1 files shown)


packages/slate/src/models/node.js
```diff
@@ -1854,13 +1854,12 @@ class Node {
...
 
-    nodes.forEach((node, i) => {
+    nodes.forEach((node, index) => {
       let ret = node
       if (ret.object != 'text') ret = ret.mapDescendants(iterator)
-      ret = iterator(ret, i, this.nodes)
+      ret = iterator(ret, index, this.nodes)
       if (ret == node) return
 
-      const index = nodes.indexOf(node)
```
benjycui: `index` will be the same as `i`, so we don't need to re-calculate it.

ianstormtaylor:
Good call, thanks @benjycui!

PR #1793, opened 2018-04-24
Configure our forks to be published on GitHub

Soreine:
For now, we only fork `slate-react` and `slate` (alias `slate-core`). I don't think we will need forks for others.

This PR adds instructions and a script to build and publish the packages on GitHub.

I already published an initial build for the current versions of `slate-core` and `slate-react`. But I reverted the commit so that this PR diff is simple.


Soreine:
Sorry, wrong repo

PR #1796, opened 2018-04-26
Remove unused dependencies

Soreine:
I started by removing `slate-sugar`.
Then I ran npm-check and removed some others I was confident were no longer used.


ianstormtaylor:
Awesome, thanks @Soreine!

PR #1801, opened 2018-04-29
Speed up decorations rendering

### LINKED ISSUE #1788: Decorations are slow
We are investigating huge performance issues since we updated our app to Slate 0.33, and have noticed that using a syntax highlighting plugin, like `slate-prism`, slows everything a lot.

- Slate version: 0.33.3 on Chrome
- Example: visit https://gitbookio.github.io/slate-prism/ paste a non-trivial code snippet (~100 lines) and start typing inside it

I have measured our the time to `React.renderToString` our app on a reasonable page containing code blocks [like this one](https://redux.js.org/~/revisions/-LAU3Ba0tMiCaOgxoUMb/faq/immutable-data). Here are the times:

- with highlighting: `16437ms` to render the app and document
- without: `899ms` to render the app and document

I also measured the time spent in the calls of the [`decorateNode`](https://github.com/GitbookIO/slate-prism/blob/master/lib/index.js#L15) of the plugin, which amounts to `188ms`. So the issue most likely comes from Slate and the way decorations are applied and rendered.

Soreine:
This PR greatly improves performance for rendering decorations.

Fix #1788 

This PR takes from @jasonphillips work on decoration filtering, as well as @zhujinxuan work on text decoration rendering.

See the past discussion around it, in the issue first #1788  then in https://github.com/GitbookIO/slate/pull/3

In practice, this PR makes Slate usable again for reasonably decorted documents. Perfs _might_ be improved further. My suggestion is to go with this solution for now, which has a clean API, and look for further improvements if needed.


Soreine:
Ready to review/merge

packages/slate-react/src/utils/get-children-decorations.js
```diff
@@ -0,0 +1,129 @@
...
+  orderChildDecorations(node, decorations).forEach(item => {
+    if (item.isRangeStart) {
+      // Item is a decoration start
+      activeDecorations.add(item.decoration)
+    } else if (item.isRangeEnd) {
+      // item is a decoration end
+      activeDecorations.remove(item.decoration)
+    } else {
+      // Item is a child node
+      childrenDecorations.push(activeDecorations.toList())
```
zhujinxuan: Why to list? or why were decorations a List?

Soreine: The List type originates from the implementation of `node.getDecorations` and `Range.createList`. Do you think changing it to a Set or Array is better?

zhujinxuan: No, I made a mistake here.  I thought creating Set is faster than creating List. But it is wrong, creating List is indeed faster without checking same member.

ianstormtaylor:
Thank you @Soreine, this is really great! Thanks for taking the time to fix this.

PR #1802, opened 2018-04-30
remove spellcheck force disable on firefox

isubasti:
spellCheck was disabled on `firefox` due to some compatibility issues. The comment for the compat sounds to be the same problem as what happened to `safari` and `chrome` on the most recent released of the browser that is fixed by https://github.com/ianstormtaylor/slate/pull/1753 therefore this PR is to remove the IS_FIREFOX check for disabling spellCheck by force

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -443,7 +443,15 @@ class Content extends React.Component {
 
   render() {
     const { props } = this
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
@isubasti did you test it works in Firefox with this removed?

isubasti:
@ianstormtaylor I did and looks good to me

ianstormtaylor:
Awesome, thanks @isubasti!

Issue #1800, opened 2018-04-27
Active-State for List-Items


signalwerk:
The [Rich-Text Example](http://slatejs.org/#/rich-text) doesn't show the correct status (active/inactive) if the current selection/curser is in a list. It shows as [inactive](https://github.com/ianstormtaylor/slate/blob/master/examples/rich-text/index.js#L203-L204)

PR #1803, opened 2018-04-30
[examples] fix rich text editor active block logic

davidchang:
Just a small fix to the rich text editor example - when you use the numbered list/unordered list block level buttons, then focus within the block, the buttons were not previously properly reflecting an active state.

This is outlined in https://github.com/ianstormtaylor/slate/issues/1800

Thanks! Digging Slate so far.


ianstormtaylor:
Thanks @davidchang!

Issue #1579, opened 2018-01-31
[Discussion] More precisely handle deferred normalization using queue


DamareYoh:
#### Do you want to request a *feature* or report a *bug*?

Feature. 

#### What's the current behavior?

In my PR #1549, @Soreine pointed out that the approach `withoutNormalization` uses to ensure the document is normalized at the end of the callback is not as performant as it should be, because it uses `normalizeDocument` as opposed to normalizing the specific nodes that were modified in the callback.

#### What's the expected behavior?

The desired behavior is that the nodes that would have normally been normalized as part of the changes executed in the callback are normalized specifically as opposed to the entire document. I was hoping to resolve this problem because I feel like it's possible to still get good performance out of this feature, but I think a discussion is required to shake out a good solution.

#### Proposal

If we detect that normalization is suppressed, we add the nodes that would normally be normalized to a queue, and normalize them in order of entry when 1) normalization is resumed or 2) some other appropriate slate lifecycle event (very unsure about what a good one would be).

The biggest problem with this idea is that if normalization is never resumed though a mistaken workflow (such as someone playing with setOperationFlag), we could potentially wind up with a massive pile of unnormalized nodes. If we're not careful, the nodes in the queue might get left behind by other changes that are triggered on the side that cause them to become outdated by other normalizations or value changes. Also, if there is a large build up of deferred normalize nodes, this could lead to performance issues. What this approach is missing is a solid event for normalizing the nodes. It would be easier to safely handle this in `withoutNormalization`

`normalizeNodeByKey` is a tempting place to resume normalization, but this would only adequately handle deferred normalization that happens in the schema or in `validateNode` and would not cover deferred normalization elsewhere, like in toolbar buttons or such. 

Alternatively, we could localize the normalization queue behavior to `withoutNormalization` because within that scope we have a clear event where we need to resolve our queue of normalization changes, and it becomes easy to manage. We don't have to worry about leaking deferred normalization outside of `withoutNormalization`, and if someone wants to continue using `{ normalize: false }` or `setOperationFlag('normalize', false)` we could assume they know what they're doing. But I like this approach less because it sort of feels half baked and incomplete, and like we'd be allowing people to shoot themselves in the foot.

Another way to deal with it might be to embrace the idea that some items aren't being normalized (because it seems that `{ normalize: false }` has become kind of rampant and people are leaving their documents unnormalized at the end of a sequence of changes), and give them some functions to clear out the queue manually. 

Anyway, I hope this is a useful discussion :)

PR #1805, opened 2018-05-01
[On Development] Do not call nested normalization in a cycle of normalization

zhujinxuan:
We still needs to call `{normalize: false}` when writing validateNode, because normalization inside normalization is hard to understand the sequence.  

For example, we want to have a validateNode to set table alignment:
```
validateNode() {
  presetAlign = [...] // calculate presetAlign
return change => {
change.setNodeByKey(table.key, {data:presetAlign}) // Forget normalize: false here
table.nodes.forEach(row => row.nodes.forEach( (cell, index) => cell.data.normalizeNodeByKey(presetAlign[index], {normalize:false))
}
}
```
if the first {normalize:false} is missing, then the `valdiateNode` will run on this schema again and again if presetAlign has different array reference each time.

<del>This PR prevents nested normalize in a normalization, unless `{normalize: true}` is given.</del>
This PR use a FIFO to organize all normalizeNodeByKey tasks

@CameronAckermanSEL  Question: shall we still call normalizeDocument if change.flags.normalize is false before withoutMutations?

### DIFF
--- a/packages/slate/src/changes/with-schema.js
+++ b/packages/slate/src/changes/with-schema.js
@@ -40,13 +40,22 @@ Changes.normalizeNodeByKey = (change, key) => {
   let { document, schema } = value
   if (change.__nextNormalizedKeys.length !== 0) {
     change.__nextNormalizedKeys.push(key)
+    return
   }
+  let firstRun = true
   change.__nextNormalizedKeys = [key]
   try {
     while (change.__nextNormalizedKeys.length !== 0) {
       key = change.__nextNormalizedKeys.shift()
-      const node = document.assertNode(key)
-      normalizeNodeAndChildren(change, node, schema)
+      document = change.value.document
+      if (firstRun) {
+        document.assertNode(key)
+        firstRun = false
+      }
+      const node = document.getNode(key)
+      if (node) {
+        normalizeNodeAndChildren(change, node, schema)
+      }
       document = change.value.document
       const ancestors = document.getAncestors(key)
       if (!ancestors) return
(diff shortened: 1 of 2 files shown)


CameronAckermanSEL:
I think based on some feedback @Soreine gave for the `withoutNormalization` function calling normalization on the entire document as opposed to the nodes that were modified within the `withoutNormalization` enclosure, we should probably not involve `withoutNormalization` in this feature for the time being.

zhujinxuan:
@CameronAckermanSEL By performance decrease, do you mean `try catch` blocks?

zhujinxuan:
If we want to allow nested normalization, but in a sequence rather than breaking a normalize process. I think we can use nextTick idea?

zhujinxuan:
I decide not to use withoutNormalizations, since there are something I am not sure about it.  Here I use a seque for normalizeNodeByKey

CameronAckermanSEL:
Should revert the change to `withoutNormalization` then. The point of it is to call normalization after all of the changes in the function enclosure have been completed so as to not require manual suppression.

CameronAckermanSEL:
To your performance question, I think this captures the concern more adequately. https://github.com/ianstormtaylor/slate/issues/1579

Would be happy to elaborate more. 

Basically what `withoutNormalization` needs is a way to keep track of the nodes that were changed but didnt undergo normalization, and then do the normalization of those nodes specifically instead of calling `normalizeDocument`.

Seems like you've set up a queue of block keys that were normalized, a similar concept and then a loop that normalizes each in order would be all we need to make `withoutNormalization` performant.

Also, I think we should probably rename it to `deferNormalization`

zhujinxuan:
@CameronAckermanSEL Agree, I think that function behaves more like normalizeNodeByKey is deferred. rather than changed.

PR #1811, opened 2018-05-02
Optimize with getPath; always try to use paths to get and update

zhujinxuan:
This is a part of #1771 .  This PR provides a getPathAsString to take advantage from the speed of string concat, and also try to use path to get, search and update


packages/slate/src/models/node.js
```diff
@@ -1370,17 +1378,55 @@ class Node {
...
+   * @returns {Array}
+   */
 
-    return path
+  getPathAsString(key) {
+    key = assertKey(key)
+    if (this.key === key) return ''
+    if (this.nodes.size === 0) return null
+    let result = null
+    // PERF: often the node keys is generated with sequences;
```
benjycui: After users edited the document, this assumption might be wrong.

Before:

```diff
[
  '1',
  '2'
]
```

And then, user inserts a new line:

```diff
  [
    '1',
+   '3',
    '2'
  ]
```

zhujinxuan: @benjycui Yes, I agree the assumption can be wrong.  But it is like a costless try that "If the guess is right, then it is good; If the guess is wrong, it does not make things worse".  

If we have no idea about users behavior, `findLast` is not expected worse than `find`.  In this case, this methods is worse the previous one only by two read keys (which seems to be costless). 

If key is not a number string, the key is NaN, then `Math.abs(lastKey - searchKey) < Math.abs(searchKey - firstKey)` is false, then the method will be findIndex anyway~

zhujinxuan: Perhaps I need to add an annotation to elaborate that~

ianstormtaylor:
@zhujinxuan this change looks like too much magic for the sake of performance to me that will actively make the codebase harder to understand. Encoding paths as strings, parsing them as integers, etc. It all seems like a convoluted way to get this.

If I understand correctly, all of this would be unnecessary if we used paths in the selections instead of (or in addition to) keys?

I'd rather not take on a lot of complicated logic if it could be avoided by making a better architectural change to the library.

zhujinxuan:
@ianstormtaylor We still have a great (not such great) improvement if we use getPath() in getAncestors, getFurthestAncestor, getFurthest...  I use string just for an extra 50% improvement, while only use getPath without string magic can have 800% already. 

I will remove the string magic and submit commits.

zhujinxuan:
@ianstormtaylor Indeed, it is related to selection/range.  It is alike: 
1. Do not do DFS twice
2. If it is possible, do not create an Immutable Object. Because Immutable takes a longer time to get itself created.

All are switched to getPath to ensure DFS is not duplicated.  And creating array is faster than creating Immutable List.

benjycui:
Hope that you can hold this PR.

Actually, I had done something like this in our forked slate, but I cannot PR now for I have to apply first.

After my PR, the benchmark(base on https://github.com/ianstormtaylor/slate/pull/1765):

```bash
benchmarks
    changes
      delete-backward
        user : 530.27 → 819.12 ops/sec (54.4720511289997% faster)
        hr : 912.60 → 1287.85 ops/sec (41.11925192135757% faster)
      delete-forward
        user : 142.23 → 228.58 ops/sec (60.7065909325922% faster)
        hr : 235.44 → 366.03 ops/sec (55.46557149243505% faster)
      insert-node-by-key
        user : 314.85 → 807.03 ops/sec (156.31846797327427% faster) 🙌
        hr : 538.00 → 1243.08 ops/sec (131.05496517171622% faster) 🙌
      insert-text-by-key-multiple
        user : 79.64 → 190.47 ops/sec (139.1550052283623% faster) 🙌
        hr : 114.91 → 301.94 ops/sec (162.75320053456548% faster) 🙌
      insert-text-by-key
        user : 576.61 → 1672.88 ops/sec (190.1241191631047% faster) 🙌
        hr : 896.01 → 2816.23 ops/sec (214.30660136689013% faster) 🙌
      insert-text
        user : 2377.44 → 2940.70 ops/sec
        hr : 4158.32 → 5221.40 ops/sec
      normalize
        user : 882.67 → 931.35 ops/sec
        hr : 1433.59 → 1418.38 ops/sec
      remove-node-by-key
        user : 415.29 → 970.73 ops/sec (133.74660793652228% faster) 🙌
        hr : 685.31 → 1552.40 ops/sec (126.5232320035702% faster) 🙌
      split-block
        user : 556.96 → 691.33 ops/sec
        hr : 886.48 → 1043.11 ops/sec
    models
      from-json-big
        user : 4.70 → 4.75 ops/sec
        hr : 6.66 → 6.74 ops/sec
      from-json
        user : 209.83 → 186.56 ops/sec
        hr : 380.03 → 340.77 ops/sec
      get-active-marks-at-range
        user : 480.36 → 449.10 ops/sec
        hr : 639.31 → 544.03 ops/sec
      get-ancestors
        user : 5472.29 → 30735.45 ops/sec (461.65649037140025% faster) 🙌
        hr : 8218.20 → 74928.36 ops/sec (811.7364539092894% faster) 🙌
      get-blocks-at-range
        user : 905.50 → 1568.01 ops/sec (73.16469292465477% faster)
        hr : 1796.80 → 2832.79 ops/sec (57.65788195885708% faster)
      get-blocks
        user : 8456.70 → 8796.07 ops/sec
        hr : 13770.19 → 12915.79 ops/sec
      get-characters-at-range
        user : 304.38 → 283.00 ops/sec
        hr : 387.57 → 333.03 ops/sec
      get-characters
        user : 4276.38 → 3765.22 ops/sec
        hr : 5671.35 → 5117.25 ops/sec
      get-common-ancestor
        user : 1172.95 → 31975.12 ops/sec (2626.039633262033% faster) 😱
        hr : 1527.20 → 60860.45 ops/sec (3885.089857942521% faster) 😱
      get-furthest-ancestor
        user : 9174.95 → 98356.25 ops/sec (972.0089992082658% faster) 🙌
        hr : 14504.19 → 259868.42 ops/sec (1691.67872889726% faster) 😱
      get-inlines-at-range
        user : 734.96 → 2373.07 ops/sec (222.88369262836073% faster) 🙌
        hr : 925.96 → 3385.41 ops/sec (265.61168992987535% faster) 🙌
      get-inlines
        user : 11932.45 → 11690.45 ops/sec
        hr : 19018.06 → 18908.81 ops/sec
      get-leaves
        user : 10595.06 → 10224.20 ops/sec
        hr : 20530.07 → 18697.95 ops/sec
      get-marks-at-range
        user : 199.28 → 223.86 ops/sec
        hr : 287.74 → 300.91 ops/sec
      get-marks
        user : 735.03 → 710.85 ops/sec
        hr : 1082.48 → 1111.09 ops/sec
      get-parent
        user : 6711.79 → 34013.90 ops/sec (406.7785689207298% faster) 🙌
        hr : 8569.09 → 80001.76 ops/sec (833.6088349983935% faster) 🙌
      get-path
        user : 2400.40 → 256104.97 ops/sec (10569.271371374982% faster) 😱
        hr : 3499.55 → 579564.64 ops/sec (16461.107761650317% faster) 😱
      get-texts-at-range
        user : 4371.11 → 6576.90 ops/sec (50.46289262370672% faster)
        hr : 6499.21 → 10157.09 ops/sec (56.28180427844705% faster)
      get-texts
        user : 10660.73 → 11297.84 ops/sec
        hr : 15847.01 → 18445.46 ops/sec
      getDescendant
        user : 9553.02 → 44818.60 ops/sec (369.1565709271796% faster) 🙌
        hr : 14032.07 → 131351.23 ops/sec (836.0788587363576% faster) 🙌
      has-node-multiple
        user : 2093.88 → 9774.22 ops/sec (366.7990937667193% faster) 🙌
        hr : 2386.29 → 12934.66 ops/sec (442.0398514115642% faster) 🙌
      has-node
        user : 9871.90 → 44356.45 ops/sec (349.3205556074771% faster) 🙌
        hr : 14706.70 → 119986.85 ops/sec (715.8654576617625% faster) 🙌
      to-json
        user : 1013.53 → 627.81 ops/sec (61% slower)
        hr : 1465.05 → 898.23 ops/sec (63% slower)
      update-node
        user : 3724.43 → 6708.29 ops/sec (80.11586139756857% faster)
        hr : 5524.59 → 11372.22 ops/sec (105.8473597097442% faster) 🙌
    html-serializer
      deserialize
        user : 9.81 → 25.96 ops/sec (164.7090172892424% faster) 🙌
        hr : 13.10 → 33.85 ops/sec (158.2912669871731% faster) 🙌
      serialize
        user : 298.81 → 316.10 ops/sec
        hr : 493.48 → 488.24 ops/sec
    plain-serializer
      deserialize
        user : 489.34 → 501.56 ops/sec
        hr : 814.25 → 836.74 ops/sec
      serialize
        user : 2587.39 → 2458.94 ops/sec
        hr : 3938.02 → 3678.04 ops/sec
    rendering
      normal
        user : 88.65 → 76.63 ops/sec
        hr : 120.21 → 99.16 ops/sec
```

benjycui:
My solution will not assume key is number string or sequential, and it works fine without memoization(which may cause memory leaks https://github.com/ianstormtaylor/slate/issues/1402)

zhujinxuan:
@benjycui I have just added you to my repo, so you can edit on branches directly.  Can you add your `get-ancestors`, `get-parent` and other benches to the branch of #1765 (benchmark-consistence branch)?

Thank you.

zhujinxuan:
@benjycui I think it is fine to hold this PR.  
Can you comment on the code panel if you feel something is not ideal?  

I am very surprised about your getPath optimization (100x faster).  Can I know what your solution is?

benjycui:
Ref: https://github.com/ianstormtaylor/slate/pull/1818

zhujinxuan:
Hi, there are already too much changes in node.js and operation.js.  I think it is not optimistic to wait for @benjycui  #1818 .  I will submit this PR this weekend with resolving the node conflict.  

For the path refactor, I will do it in future PR with a different interface.

zhujinxuan:
@ianstormtaylor The conflict is resolved.  Can you have a review on this branch?

Thank you,
Jinxuan Zhu

packages/slate/src/models/node.js
```diff
@@ -1371,41 +1389,54 @@ class Node {
...
 
   getPath(key) {
-    let child = this.assertNode(key)
-    const ancestors = this.getAncestors(key)
-    const path = []
+    key = assertKey(key)
+    if (this.key === key) return []
+    if (this.nodes.size === 0) return null
+    let result = null
+    // PERF: often the node keys is generated with sequences;
```
ianstormtaylor: This is not really the kind of performance improvement I'd like to add here, since it's so implementation-specific. People can swap out their key generator with whatever they want. I'd need to see this in an isolated pull request to see if it's really quick enough to merit adding.

packages/slate/src/models/node.js
```diff
@@ -1748,7 +1779,7 @@ class Node {
    */
 
   hasNode(key) {
-    return !!this.getNode(key)
+    return Array.isArray(this.getPath(key))
```
ianstormtaylor: What is this doing now? This is really hard to understand.

packages/slate/src/models/node.js
```diff
@@ -2024,19 +2055,10 @@ class Node {
...
-      child = parent
-      nodes = nodes.set(index, node)
-      parent = parent.set('nodes', nodes)
-      node = parent
-    })
-
-    return node
+    this.assertDescendant(node.key)
+    const path = this.getPath(node.key).map(x => ['nodes', x])
+    // PERF: setIn is faster than recursive assign
```
ianstormtaylor: Nice, this is a cool one!

packages/slate/src/models/node.js
```diff
@@ -943,25 +957,29 @@ class Node {
...
+        keys[child.key] = true
+        return
+      }
 
-  /**
-   * Return a set of all keys in the node.
-   *
-   * @return {Set<String>}
-   */
+      // PERF: use the logic of utils._extend; becuase
```
ianstormtaylor: Can you remove this and add it back in a later pull request to prove that it makes a big enough difference to merit adding?

packages/slate/src/models/node.js
```diff
@@ -198,12 +198,18 @@ class Node {
...
-    return firstIndex < secondIndex
+    if (first === second) return false
+    if (parseInt(first, 10) > parseInt(second, 10)) {
+      // Ensure areDescendantSorted(second, first) is also cached
+      // Always prefer newer node in second argument, for potential
+      // futher optimization
+      return !this.areDescendantsSorted(second, first)
+    }
+    if (first === this.key) return true
+    if (second === this.key) return false
```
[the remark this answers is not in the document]
zhujinxuan: Because they have different return; we can write 
```
  if (first === this.key || second === this.key) return first === this.key
```

But I am not sure whether it is easier to understand

packages/slate/src/models/node.js
```diff
@@ -198,12 +198,18 @@ class Node {
...
     second = assertKey(second)
 
-    const keys = this.getKeysAsArray()
-    const firstIndex = keys.indexOf(first)
-    const secondIndex = keys.indexOf(second)
-    if (firstIndex == -1 || secondIndex == -1) return null
-
-    return firstIndex < secondIndex
+    if (first === second) return false
+    if (parseInt(first, 10) > parseInt(second, 10)) {
```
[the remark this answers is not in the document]
zhujinxuan: To ensure `this.areDescendantsSorted("100", "3")` can use the cache of `this.areDescendantsSorted("3", "100")`

packages/slate/src/models/node.js
```diff
@@ -250,14 +253,13 @@ class Node {
    */
 
   assertNode(key) {
-    const node = this.getNode(key)
+    key = assertKey(key)
 
-    if (!node) {
-      key = assertKey(key)
+    if (!this.hasNode(key)) {
```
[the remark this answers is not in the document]
zhujinxuan: Because `hasNode` is faster than `getNode`.  `hasNode` is only for whether path exists, while `getNode` additionally contains visit node by path.

Therefore we use `hasNode` is better here.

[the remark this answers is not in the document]
zhujinxuan: Another reason is `hasNode` is more verbally alike with `assertNode`

packages/slate/src/models/node.js
zhujinxuan: Fixed~

packages/slate/src/models/node.js
zhujinxuan: Fixed~

packages/slate/src/models/node.js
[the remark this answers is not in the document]
zhujinxuan: This part is removed I think.

packages/slate/src/models/node.js
zhujinxuan: Because I move everything as path based;  Then is this.getPath(key) returns an array, then it will have a node, otherwise (null or undefined) the node doest not exists.

zhujinxuan: Replace it with `!!this.getPath(key)` in rencent PR.

ianstormtaylor:
Hey @zhujinxuan thank you for this (and for all of your PRs)! As I mentioned in #1997 I borrowed a lot of concepts from this pull request to make that one. I think most of the changes to the core `Node.get*` methods are accounted for, as in #1997 they almost all use paths instead of keys now. With the similar caching method by using a `key -> path` mapping, which was based on your `key -> true` mapping here.

Because of that, I'm going to close this pull request out since lots of its behavior is handled by #1997. If I've forgotten any specific piece that you think is important or more performant, please feel free to open another pull request with it.

PR #1812, opened 2018-05-02
Faster search when taking no arguments

zhujinxuan:
Due to nature of JS, JS will search on the cache__no_arguments.prototype when a value is not cached.  This PR prevents the prototype search.

### DIFF
--- a/packages/slate-react/src/components/void.js
+++ b/packages/slate-react/src/components/void.js
@@ -72,9 +72,7 @@ class Void extends React.Component {
     )
 
     const content = (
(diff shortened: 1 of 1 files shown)


Soreine:
Have you run your benchmarks against this?

zhujinxuan:
@Soreine No, I have not.  All benchmarks in the current slate reset memory after each bench.  Perhaps we need some benchmark tests for caches.

ianstormtaylor:
Does this actually speed things up? Without benchmarking it how do we know?

Issue #1799, opened 2018-04-27
clipboard fragment content is lost


zhujinxuan:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

Select something, do two split block(or some other operations), then paste.  We can see `getEventTransfer(html).type === "html"` rather than "fragment", and fragment content is missing.

![](http://g.recordit.co/zabncoSAFO.gif)

zhujinxuan:
```
  // Revert to the previous selection right after copying.
  window.requestAnimationFrame(() => {
    editor.removeChild(div)
    native.removeAllRanges()
    native.addRange(range)
  })
```
The problem happens here. Any operation after copy will destroy the copy. I think we shall restore range in window.requestAnimationFrame, but do not removeChild in this frame. We shall remove child in paste on another copy.

zhujinxuan:
Note: workaround solution if someone is also struggling here.

```
onCopy(....) {
  event.clipBoardData.setData('text/html', html.serialize(....))
  event.clipBoardData.setData('data/base64', base64.serialize(....))
  event.preventDefault()
}
```

ianstormtaylor:
I don't understand what's happening here. Can you explain @zhujinxuan? Why wouldn't the `requestAnimationFrame` be running before the split?

zhujinxuan:
I am not sure about the reason either.  I thought requestAnimationFrame is the problem because if we immediately paste, then we still have `getEventTransfer(html).type === "html"`.

I will be happy to investigate when then event transfer data is lost, but the problem is that I do not know how to get the clipboard info from console without a clipboard event.

ianstormtaylor:
I don't think I can replicate this actually.

zhujinxuan:
@ianstormtaylor Here is my method to reproduce in rich-text example:

1. set a break point at 

```
    const { type, fragment, text } = transfer
```
2. select a range, copy
3. press double enter
4. check the value of `transfer.type`

ianstormtaylor:
Interesting, if the range you select includes a mark then it doesn't seem to happen for me. But if it doesn't include a mark then I can reproduce it.

zhujinxuan:
Nice, we are sure now it is not related to requestAnimation. Perhaps we need a span or alike in the div.innerHTML for copy to ensure it is a text/HTML?

Issue #1693, opened 2018-03-08
Copy of a word within a marked set of text copies all characters prior to and including the selected word.


CameronAckermanSEL:
#### Do you want to request a _feature_ or report a _bug_?
Bug 

#### What's the current behavior?

When copying a word from a block that has a `mark` starting at the beginning of the text, all of the text prior to the selected word is also applied to the clipboard.

http://recordit.co/8dapGX4bRF

#### What's the expected behavior?

Only the selected word is copied.

zhujinxuan:
I do some logs in the devTool.  I am sure that the problem is the fragment contains no marks in insertFragmentAtRange

I think the problem is by getFragmentAtRange

benjycui:
The bug relate to browser compatibility.

In Firefox, what we copy is:

```html
<div style=\"position: absolute; left: -9999px; border: 0px none; padding: 0px; margin: 0px; top: 52px;\" contenteditable=\"true\"><span style=\"white-space: pre;\" data-slate-fragment=\"JTdCJTIyb2JqZWN0JTIyJTNBJTIyZG9jdW1lbnQlMjIlMkMlMjJkYXRhJTIyJTNBJTdCJTdEJTJDJTIybm9kZXMlMjIlM0ElNUIlN0IlMjJvYmplY3QlMjIlM0ElMjJibG9jayUyMiUyQyUyMnR5cGUlMjIlM0ElMjJwYXJhZ3JhcGglMjIlMkMlMjJpc1ZvaWQlMjIlM0FmYWxzZSUyQyUyMmRhdGElMjIlM0ElN0IlN0QlMkMlMjJub2RlcyUyMiUzQSU1QiU3QiUyMm9iamVjdCUyMiUzQSUyMnRleHQlMjIlMkMlMjJsZWF2ZXMlMjIlM0ElNUIlN0IlMjJvYmplY3QlMjIlM0ElMjJsZWFmJTIyJTJDJTIydGV4dCUyMiUzQSUyMmJvbGQlMjIlMkMlMjJtYXJrcyUyMiUzQSU1QiU3QiUyMm9iamVjdCUyMiUzQSUyMm1hcmslMjIlMkMlMjJ0eXBlJTIyJTNBJTIyYm9sZCUyMiUyQyUyMmRhdGElMjIlM0ElN0IlN0QlN0QlNUQlN0QlNUQlN0QlNUQlN0QlNUQlN0Q=\">bold</span></div>
```

[...]

PR #1815, opened 2018-05-03
Set clipboard data programmitcally on copy when supported

tobiasandersen:

Fixing bugs.

#### What's the new behavior?

This should make copying much more reliable for most browsers. See https://github.com/ianstormtaylor/slate/pull/1543 for the full discussion.

For browsers that support it (I think all except IE and Edge), we use `clipboardData.setData()` to set all clipboard data. Also removed some old `COMPAT` code for Firefox (since it will use the Clipboard API).

Fixes: https://github.com/ianstormtaylor/slate/issues/1449, https://github.com/ianstormtaylor/slate/issues/1693 and https://github.com/ianstormtaylor/slate/issues/1799 (for non IE/Edge).

Reviewers: @ianstormtaylor


packages/slate-react/src/utils/clone-fragment.js
```diff
@@ -32,7 +35,7 @@ function cloneFragment(event, value, fragment = value.fragment) {
 
   // Make sure attach is a non-empty node, since empty nodes will not get copied
   contents.childNodes.forEach(node => {
-    if (node.innerText.trim() !== '') {
+    if (node.innerText && node.innerText.trim() !== '') {
```
tobiasandersen: This was [added recently](https://github.com/ianstormtaylor/slate/pull/1748/) and throws for plain text since they don't have `node.innerText`.

packages/slate-react/src/utils/clone-fragment.js
```diff
@@ -107,31 +110,30 @@ function cloneFragment(event, value, fragment = value.fragment) {
...
+
+  // COMPAT: For browser that don't support the Clipboard API's setData method,
+  // we must rely on the browser to natively copy what's selected.
+  // So we add the div (containing our content) to the DOM, and select it.
+  const editor = event.target.closest('[data-slate-editor]')
   div.setAttribute('contenteditable', true)
   div.style.position = 'absolute'
   div.style.left = '-9999px'
-
-  // COMPAT: In Firefox, the viewport jumps to find the phony div, so it
```
tobiasandersen: Removed this since Firefox support `clipboardData.setData()`.

packages/slate-react/src/utils/clone-fragment.js
```diff
@@ -107,31 +110,30 @@ function cloneFragment(event, value, fragment = value.fragment) {
...
-  div.style.border = '0px'
-  div.style.padding = '0px'
-  div.style.margin = '0px'
-  div.style.top = `${window.pageYOffset ||
-    window.document.documentElement.scrollTop}px`
-
-  div.appendChild(contents)
   editor.appendChild(div)
-
-  // COMPAT: In Firefox, trying to use the terser `native.selectAllChildren`
```
tobiasandersen: Removed this since Firefox support clipboardData.setData().

packages/slate-react/src/utils/clone-fragment.js
jvdsande: Great catch!
Sorry for that =/ Seems like I'm not using any plain texts in my projects.

ianstormtaylor:
Thank you @tobiasandersen, this is awesome. 

Almost a negative diff too 😄

packages/slate-react/src/utils/clone-fragment.js
gggdomi: @tobiasandersen I noticed that empty blocks now contains a "zero-width space" character when copied and pasted in a notepad.

Replacing the highlighted line with the following effectively remove the zero-width spaces - but there's probably a better fix, I'm not familiar with this problem.
`event.clipboardData.setData(TEXT, native.toString().replace(/\u200B/g, ''))`

PR #1817, opened 2018-05-03
Faster is`Type`

zhujinxuan:
Improve a feature

#### What's the new behavior?

The previous isXX use `any[Type]`, and it will try to search on the prototype chain when any does not have the type.  In this change, we use `hasOwnProperty` instead to ensure the search does not recursively go through the prototype chain, but search exactly on the prototype object.  It should make the `isXXX` a bit faster

Here I intentionally left Text.isText, Leaf.isLeaf unchanged, because I do not want to cause conflict with on working characters/leaf PRs

### DIFF
--- a/packages/slate/src/models/block.js
+++ b/packages/slate/src/models/block.js
@@ -129,7 +129,11 @@ class Block extends Record(DEFAULTS) {
    */
 
   static isBlock(any) {
-    return !!(any && any[MODEL_TYPES.BLOCK])
+    if (!any || !any.__proto__) return false
+    if (any.__proto__.hasOwnProperty) {
+      return any.__proto__.hasOwnProperty(MODEL_TYPES.BLOCK)
+    }
+    return !!any[MODEL_TYPES.BLOCK]
   }
 
   /**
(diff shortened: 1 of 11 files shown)


ianstormtaylor:
Hey @zhujinxuan is there any evidence of this code path slowing anything down? 

It seems like this is will be incredibly fast compared to most everything either way. In which case I'd like to keep the simpler check that we borrowed from Immutable.js since it's easier to read/maintain.

zhujinxuan:
@ianstormtaylor For slate operations, it is not significantly faster, only theoretically fast, because the isType are not frequently called in Slate.  (we often use x.object instead)

zhujinxuan:
A potential (though never occurs in real practice) is that we may have `Block.isBlock(Block.prototype) === true` in Immutable's way.

zhujinxuan:
I will submit a PR to immutable-js, and see what is their ideas about that.

zhujinxuan:
I see how the immutable-js did the type checking that way.  They do things like

```
export class OrderedSet extends Set 
```

They want make Set, OrderedSet and other things extendable, and thus they use all prototype chain search.  It seems we do not need to extends things like that.

ianstormtaylor:
Hey @zhujinxuan thanks for this, but I think we should wait for this to become a problem before we make a change like this that is complex to understand.

PR #1818, opened 2018-05-04
refactor: use cached paths to improve performance

benjycui:

To improve performance.

#### What's the new behavior?

No behavior of API will be changed.

1. Generate paths' cache for each node in a document tree.
1. Use path to locate a node to get/modify it.
    > If the API accept keys, just map key to path with `this.getPath(key)`
1. Update paths' cache if the structure of document tree is changed.
1. The new document tree will inherit updated paths' cache.

After this PR, all the methods that won't modify the structure of document tree will improve from O(n) to O(d), e.g. all the getters and `add_mark` `set_mark` `remove_mark` `set_node` `insert_text` `move_text`. 

And we only need to traverse the sub tree when its structure is changed.

1. n => the number of nodes of a document
1. d => the depth of a node in document
In real world, d << n.

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1236,13 +1236,18 @@ class Node {
   getNextSibling(key) {
     key = assertKey(key)
 
-    const parent = this.getParent(key)
-    const after = parent.nodes.skipUntil(child => child.key == key)
+    const path = this.getPath(key)
+    if (path) {
+      const isLast = index => index === path.length - 1
+      const nextSiblingPath = path.map((n, i) => (isLast(i) ? n + 1 : n))
 
-    if (after.size == 0) {
-      throw new Error(`Could not find a child node with key "${key}".`)
+      const sibling = this.getDescendantAtPath(nextSiblingPath)
+      if (sibling) {
+        return sibling
+      }
     }
-    return after.get(1)
+
+    throw new Error(`Could not find a child node with key "${key}".`)
   }
 
   /**
@@ -1717,7 +1722,7 @@ class Node {
    */
 
   hasDescendant(key) {
-    return !!this.getDescendant(key)
+    return !!this.getPath(key)
   }
 
   /**
@@ -1728,7 +1733,8 @@ class Node {
    */
 
   hasNode(key) {
-    return !!this.getNode(key)
+    return !!this.getDescendant(key)
+    // return !!this.getNode(key)
   }
 
   /**
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1,7 +1,7 @@
 import direction from 'direction'
 import isPlainObject from 'is-plain-object'
 import logger from 'slate-dev-logger'
-import { List, OrderedSet, Set } from 'immutable'
+import { List, Map, OrderedSet, Set } from 'immutable'
 
 import Block from './block'
 import Data from './data'
@@ -352,18 +352,14 @@ class Node {
     if (key == this.key) return List()
     if (this.hasChild(key)) return List([this])
 
-    let ancestors
-    this.nodes.find(node => {
-      if (node.object == 'text') return false
-      ancestors = node.getAncestors(key)
-      return ancestors
-    })
-
-    if (ancestors) {
-      return ancestors.unshift(this)
-    } else {
-      return null
+    const path = this.getPath(key)
+    if (path) {
+      return path.slice(0, -1).reduce((ancestors, index) => {
+        const parent = ancestors.last()
+        return ancestors.push(parent.nodes.get(index))
+      }, List([this]))
     }
+    return null
   }
 
   /**
@@ -598,21 +594,23 @@ class Node {
     if (one == this.key) return this
     if (two == this.key) return this
 
+    // Both of one and two are descendants of `this`,
+    // so at least `this` is the commom ancestor
     this.assertDescendant(one)
     this.assertDescendant(two)
-    let ancestors = new List()
-    let oneParent = this.getParent(one)
-    let twoParent = this.getParent(two)
 
-    while (oneParent) {
-      ancestors = ancestors.push(oneParent)
-      oneParent = this.getParent(oneParent.key)
-    }
+    const onePath = this.getPath(one)
+    const twoPath = this.getPath(two)
 
-    while (twoParent) {
-      if (ancestors.includes(twoParent)) return twoParent
-      twoParent = this.getParent(twoParent.key)
+    // Find the first different ancestor in path
+    let index = 0
+    for (; index < onePath.length; index += 1) {
+      if (onePath[index] !== twoPath[index]) {
+        break
+      }
     }
+    const commonAncestorPath = onePath.slice(0, index)
+    return this.getDescendantAtPath(commonAncestorPath)
   }
 
   /**
@@ -651,20 +649,12 @@ class Node {
 
   getDescendant(key) {
     key = assertKey(key)
-    let descendantFound = null
 
-    const found = this.nodes.find(node => {
-      if (node.key === key) {
-        return node
-      } else if (node.object !== 'text') {
-        descendantFound = node.getDescendant(key)
-        return descendantFound
-      } else {
-        return false
-      }
-    })
-
-    return descendantFound || found
+    const path = this.getPath(key)
+    if (path) {
+      return this.getDescendantAtPath(path)
+    }
+    return null
   }
 
   /**
@@ -725,32 +715,27 @@ class Node {
     // Split at the start and end.
     let child = startText
     let previous
-    let parent
 
-    while ((parent = node.getParent(child.key))) {
-      const index = parent.nodes.indexOf(child)
+    while (node.getParent(child.key)) {
       const position =
         child.object == 'text' ? startOffset : child.nodes.indexOf(previous)
 
-      parent = parent.splitNode(index, position)
-      node = node.updateNode(parent)
-      previous = parent.nodes.get(index + 1)
-      child = parent
+      node = node.splitNode(node.getPath(child.key), position)
+      previous = node.getNextSibling(child.key)
+      child = node.getParent(child.key)
     }
 
     child = startKey == endKey ? node.getNextText(startKey) : endText
 
-    while ((parent = node.getParent(child.key))) {
-      const index = parent.nodes.indexOf(child)
+    while (node.getParent(child.key)) {
       const position =
         child.object == 'text'
           ? startKey == endKey ? endOffset - startOffset : endOffset
           : child.nodes.indexOf(previous)
 
-      parent = parent.splitNode(index, position)
-      node = node.updateNode(parent)
-      previous = parent.nodes.get(index + 1)
-      child = parent
+      node = node.splitNode(node.getPath(child.key), position)
+      previous = node.getNextSibling(child.key)
+      child = node.getParent(child.key)
     }
 
     // Get the start and end nodes.
@@ -823,11 +808,12 @@ class Node {
 
   getFurthestAncestor(key) {
     key = assertKey(key)
-    return this.nodes.find(node => {
-      if (node.key == key) return true
-      if (node.object == 'text') return false
-      return node.hasDescendant(key)
-    })
+
+    const path = this.getPath(key)
+    if (path) {
+      return this.getDescendantAtPath(path.slice(0, 1))
+    }
+    return null
   }
 
   /**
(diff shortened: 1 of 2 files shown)


benjycui:
Both of @zhujinxuan and me agree to map `key` to `path` to improve the performance of get/modify.

The only difference is that we use different way to improve the performance of `getPath`.

packages/slate/src/models/node.js
```diff
@@ -1370,17 +1394,15 @@ class Node {
...
-      const index = ancestor.nodes.indexOf(child)
-      path.unshift(index)
-      child = ancestor
-    })
+    if (!this.getPathsCache()) {
+      this.regeneratePathsCache()
+    }
 
-    return path
+    return this.getPathsCache().get(key, null)
```
benjycui: @zhujinxuan this is why `getPath` is so fast.. OK, it is a cheat.

packages/slate/src/models/node.js
```diff
@@ -1982,14 +2018,34 @@ class Node {
...
+  updateNode(node, willTreeChanged = true, changedChildIndex = 0) {
     if (node.key == this.key) {
       return node
     }
 
     let child = this.assertDescendant(node.key)
-    const ancestors = this.getAncestors(node.key)
 
+    let pathsCache = this.getPathsCache()
+    if (pathsCache && willTreeChanged) {
```
benjycui: We need to update paths' cache if the tree structure is changed.

We still need to traverse the whole tree if we just insert a node before `document.nodes.first()`, but we won't do that frequently.

Under most circumstances, we just need to update a sub tree's path cache.

zhujinxuan:
A small question: why use cache as Map() not `{}.__proto__ =null`; I remember, at least before last July, the Map is slower in get when the key is all string.

zhujinxuan:
Great, we only need to update with `updateNode` and `insertNode` and thus, then we can always have good path.

Only one problem is that shall we recursively update the path?  If we shall, how to recursively set path without losing efficiency?

zhujinxuan:
```
    changes
      delete-backward
        user : 328.72 → 552.30 ops/sec (68.01259702935722% faster)
        hr : 411.24 → 909.28 ops/sec (121.10671342770662% faster) 🙌
      delete-forward
        user : 111.04 → 164.88 ops/sec (48.47736408392881% faster)
        hr : 163.67 → 271.09 ops/sec (65.63535241091316% faster)
      insert-node-by-key
        user : 223.22 → 626.18 ops/sec (180.5201210854309% faster) 🙌
        hr : 332.78 → 989.49 ops/sec (197.3412813963255% faster) 🙌
      insert-text-by-key-multiple
        user : 69.15 → 165.07 ops/sec (138.70555589487432% faster) 🙌
        hr : 91.77 → 259.41 ops/sec (182.68057222144358% faster) 🙌
      insert-text-by-key
        user : 366.01 → 1422.02 ops/sec (288.51988951531894% faster) 🙌
        hr : 407.11 → 2343.71 ops/sec (475.69453689890497% faster) 🙌
      insert-text
        user : 1685.80 → 2466.68 ops/sec (46.32065235296292% faster)
        hr : 1981.42 → 4206.84 ops/sec (112.31377436845244% faster) 🙌
      normalize
        user : 658.01 → 834.76 ops/sec
        hr : 981.88 → 1283.84 ops/sec
      remove-node-by-key
        user : 286.73 → 788.50 ops/sec (174.9940279541578% faster) 🙌
        hr : 368.83 → 1314.08 ops/sec (256.2861357698833% faster) 🙌
      split-block
        user : 379.03 → 479.18 ops/sec
        hr : 479.96 → 805.65 ops/sec (67.85534893571239% faster)
    models
      from-json-big
        user : 3.61 → 4.21 ops/sec
        hr : 3.89 → 6.02 ops/sec (54.73709720074705% faster)
      from-json
        user : 171.05 → 184.17 ops/sec
        hr : 244.77 → 308.29 ops/sec
      get-active-marks-at-range
        user : 237.30 → 476.64 ops/sec (100.85639437241885% faster) 🙌
        hr : 283.34 → 580.91 ops/sec (105.02146443489802% faster) 🙌
      get-ancestors
        user : 4026.25 → 31138.55 ops/sec (673.388731106515% faster) 🙌
        hr : 5688.91 → 63763.42 ops/sec (1020.8362576537926% faster) 😱
      get-blocks-at-range
        user : 663.70 → 1125.98 ops/sec (69.65123588575764% faster)
        hr : 1085.43 → 2309.03 ops/sec (112.72907185785827% faster) 🙌
      get-blocks
        user : 6288.14 → 8458.17 ops/sec (34.509706892742734% faster)
        hr : 7691.27 → 11956.95 ops/sec (55.46146506613449% faster)
      get-characters-at-range
        user : 223.19 → 287.74 ops/sec
        hr : 212.90 → 346.63 ops/sec (62.814437149237904% faster)
      get-characters
        user : 3576.32 → 3666.84 ops/sec
        hr : 3615.92 → 5178.32 ops/sec (43.20911949710848% faster)
      get-common-ancestor
        user : 693.82 → 29848.73 ops/sec (4202.10177012649% faster) 😱
        hr : 580.76 → 52352.47 ops/sec (8914.438069644988% faster) 😱
      get-furthest-ancestor
        user : 7001.33 → 60471.01 ops/sec (763.7080848820091% faster) 🙌
        hr : 8924.66 → 173701.55 ops/sec (1846.310076945164% faster) 😱
      get-inlines-at-range
        user : 575.41 → 1851.25 ops/sec (221.72900530343247% faster) 🙌
        hr : 689.16 → 2788.99 ops/sec (304.6952019842218% faster) 🙌
      get-inlines
        user : 10796.53 → 10307.06 ops/sec
        hr : 14964.08 → 15498.44 ops/sec
      get-leaves
        user : 8513.67 → 9982.92 ops/sec
        hr : 14370.33 → 16258.30 ops/sec
      get-marks-at-range
        user : 155.80 → 180.06 ops/sec
        hr : 171.91 → 246.56 ops/sec (43.427610027276145% faster)
      get-marks
        user : 507.37 → 496.00 ops/sec
        hr : 787.88 → 870.32 ops/sec
      get-parent
        user : 6092.50 → 34265.24 ops/sec (462.41660627684666% faster) 🙌
        hr : 6893.33 → 66451.77 ops/sec (864.0014752019548% faster) 🙌
      get-path
        user : 2136.64 → 123532.77 ops/sec (5681.646459986303% faster) 😱
        hr : 2850.94 → 309911.71 ops/sec (10770.52453020746% faster) 😱
      get-texts-at-range
        user : 4183.46 → 6814.66 ops/sec (62.89533621168808% faster)
        hr : 5315.87 → 9748.23 ops/sec (83.37990384638904% faster)
      get-texts
        user : 10717.81 → 9654.29 ops/sec
        hr : 15128.62 → 14774.64 ops/sec
      getDescendant
        user : 8863.47 → 45246.80 ops/sec (410.4863865126614% faster) 🙌
        hr : 10989.88 → 105839.97 ops/sec (863.0677839232625% faster) 🙌
      has-node-multiple
        user : 1907.49 → 9454.65 ops/sec (395.6587471357242% faster) 🙌
        hr : 2165.12 → 11884.54 ops/sec (448.9090326390179% faster) 🙌
      has-node
        user : 7527.37 → 41278.57 ops/sec (448.3799985897573% faster) 🙌
        hr : 9577.25 → 93003.25 ops/sec (871.0846629235726% faster) 🙌
      to-json
        user : 827.97 → 677.72 ops/sec
        hr : 1147.06 → 1104.63 ops/sec
      update-node
        user : 2909.93 → 5569.27 ops/sec (91.38808219878119% faster)
        hr : 3483.29 → 10554.17 ops/sec (202.99471083297814% faster) 🙌
    html-serializer
      deserialize
        user : 8.28 → 26.33 ops/sec (217.99988521337679% faster) 🙌
        hr : 10.06 → 35.49 ops/sec (252.8654389926683% faster) 🙌
      serialize
        user : 203.20 → 252.72 ops/sec
        hr : 325.46 → 411.25 ops/sec
    plain-serializer
      deserialize
        user : 405.47 → 420.02 ops/sec
        hr : 647.01 → 684.20 ops/sec
      serialize
        user : 2520.71 → 2379.43 ops/sec
        hr : 3628.95 → 3599.25 ops/sec
    rendering
      normal
        user : 66.02 → 77.60 ops/sec
        hr : 80.76 → 106.41 ops/sec

✨  Done in 61.09s.
```

Crazy improvement

packages/slate/src/models/node.js
```diff
@@ -1927,52 +1955,59 @@ class Node {
...
-   * Split a child node by `index` at `position`.
+   * Split a descendant node by `path` at `position`.
    *
-   * @param {Number} index
+   * @param {Array} path
    * @param {Number} position
    * @return {Node}
    */
 
-  splitNode(index, position) {
```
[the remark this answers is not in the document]
zhujinxuan: Oh.... I see, you can re-use path if splitNode(path...)

packages/slate/src/models/node.js
```diff
@@ -1722,28 +1743,28 @@ class Node {
...
+   * @param {Array} path
    * @param {Node} node
    * @return {Node}
    */
 
-  insertNode(index, node) {
-    const keys = this.getKeysAsArray()
-
-    if (keys.includes(node.key)) {
+  insertNode(path, node) {
```
ianstormtaylor: I like this change! Good for them to take a `path` and to make the operation logic simpler. Nice job.

packages/slate/src/models/node.js
```diff
@@ -1982,14 +2017,34 @@ class Node {
    * @return {Node}
    */
 
-  updateNode(node) {
+  updateNode(node, willTreeChanged = true, changedChildIndex = 0) {
```
ianstormtaylor: I'm not really keen on these extra arguments that are hard to remember, and very hard to use if you aren't deeply familiar with the code. Is there any way to do this using the existing recursive, top-down memoization techniques?

packages/slate/src/models/node.js
```diff
@@ -1346,20 +1332,57 @@ class Node {
...
+        ret = child.forEachDescendantWithPath(iterator, path)
+        return ret
       }
     })
 
-    return node
+    return ret
+  }
+
+  regeneratePathsCache() {
```
ianstormtaylor: I'm not keen on these types of methods being exposed, and generally what they do to the complexity of the code. Is there a way to restructure this to use the existing memoization technique, for top-down memoization?

benjycui:
> A small question: why use cache as Map() not {}.__proto__ =null; I remember, at least before last July, the Map is slower in get when the key is all string.

@zhujinxuan we will reuse paths' cache, an immutable Map means that we don't need to shallow copy the cache. See: https://github.com/ianstormtaylor/slate/pull/1818/files#diff-54c7838b3ea52077be8fa46cee479d9bR2057

benjycui:
> Only one problem is that shall we recursively update the path? If we shall, how to recursively set path without losing efficiency?

In real world, a document tree will look like this:

image

1. The depth is small, for users are not going to write a deep nested document which is hard to read. 
    > This is why I said O(d) << O(n)
1. There are many branches in first layer of document tree. So that we only need to traverse a small subtree unless we had insert/remove node to `document.nodes`.
1. Yep, it is a common case to insert/remove node to `document.nodes`(e.g. insert new line, table), but users won't do that frequently. What users may do frequently is `isnert_text` `remove_text` which will not change the document tree.

benjycui:
> If you can get it to all use the implicit existing memoization techniques I'm all for it.

The existing memoization will lead to memory leak. In my use case, user found a Chrome tab will use more that 1GB memory, but if he just refresh the page, only need 300MB memory.

And the paths' cache is also implicit. Developers don't need to care about it, just call nodes' API, we will maintain the right cache.

> Otherwise, I think this is similar to the other path caching pull requests in that it wouldn't be necessary if we just switched to having selections track paths instead of keys. And in that case, I'd rather not merge interim behavior that's going to be erased in the switch.

Theoretically, this PR has no difference with https://github.com/ianstormtaylor/slate/pull/1811 , but in real world, it is different, see: https://github.com/ianstormtaylor/slate/pull/1818#issuecomment-386771674

And I think developers should not use path directly, for it is difficult to maintain.

zhujinxuan:
@benjycui Can I know what is the memory leak when we remove getCharacters cache?  I have a hypothesis that most memories are use by getCharacters and getCharactersAtRange.   

Another thing I am not sure is that, though shallow copying has a greater complexity, but single utils._extends and Object.assign is much faster than Immutable operations (and saves more memory).  I think I will have a test comparing shallow copying and Immutable.Map after this PR merged.

Soreine:
Oops, did not mean to push upon your branch @benjycui 
I wanted to push to my own fork, but with many remotes, I ended up making a mistake :) Feel free to revert my commit

truongdinhquoc:
Yuur

Soreine:
We have been using this PR as is for some time, and it worked mostly fine. We have noticed two bugs though, that forced us to stop using it.

The first bug occurred during normalization. We normalize a node by replacing it `change.replaceNodeByKey(context.node.key ...)`, and then the same normalization function is called again for that node, but this time `change.value.document.getDescendant(context.node.key)` returns `null`, which is wrong.

The second bug happened when simply setting some data on a node using `change.setNodeByKey`, which failed with `Node.create only accepts objects or nodes but you passed it: null`

If needed, we can try to provide failing tests for this.

PR #1825, opened 2018-05-06
fix: prevent IE11 from throwing if selection contains tables

### LINKED ISSUE #1824: If selection contains table, IE11 throws "error 800a025e" on calling content.updateSelection()
If the content contains tables, IE11 will throw the following error:

```
unable to complete the operation due to error 800a025e
```

Based on this SO answer: https://stackoverflow.com/a/37580789 I was able to narrow down the culprit to this line: 

https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/content.js#L154

Adding  the solution linked above seemed to work:

```javascript
/**
 * COMPAT: if we are in <= IE11 and the selection contains
 * tables, `removeAllRanges()` will throw 
 * "unable to complete the operation due to error 800a025e"
 * 
 * @param {Selection} selection 
 */
function removeAllRanges(selection) {
  if (document.body.createTextRange) { // All IE but Edge
    const range = document.body.createTextRange()
    range.collapse()
    range.select()
  } else {
    selection.removeAllRanges()
  }
}
```

Will open a PR shortly

bsouthga:
resolves #1824 by adding a compatibility helper function.

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -16,6 +16,7 @@ import Node from './node'
 import findDOMRange from '../utils/find-dom-range'
 import findRange from '../utils/find-range'
 import scrollToSelection from '../utils/scroll-to-selection'
+import removeAllRanges from '../utils/remove-all-ranges'
 
 /**
  * Debug.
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Thanks @bsouthga!

PR #1831, opened 2018-05-09
Fix slate-html-serializer: applyMark

finethanks:

Fix the bug when a nested in mark node a "void" inline node is the cause of error in the applyMark method because always trying to get children nodes without any checking.

For example, if we copy text from MS Word/Pages, HTML in clipboard will look like `<strong>first line<br/>second line</strong>`. Probably, break line(as well as `image`) node will be a void inline node without children and it'll break deserializer in this case.

I added test case `mark-void-inline.js` and new inline type `linebreak` in helpers.

### DIFF
--- a/packages/slate-html-serializer/src/index.js
+++ b/packages/slate-html-serializer/src/index.js
@@ -299,7 +299,7 @@ class Html {
           leaf.marks.push({ type, data })
           return leaf
         })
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Thanks @finethanks, this is a great looking PR!

PR #1836, opened 2018-05-11
Remove legacy COMPAT code from cloneFragment

tobiasandersen:
Fixing a bug and cleaning up legacy stuff. I should probably have done this already in my [previous PR](https://github.com/ianstormtaylor/slate/pull/1815)...

#### What's the new behavior?
We currently select more html than we need when copying marks (to work around browser quirks). This isn't needed anymore (since #1815), and also messes up the contents of the html clipboard register (see https://github.com/ianstormtaylor/slate/issues/1693). 

Since we now have full control of what's put in each clipboard register, we should be able to remove all COMPAT code for Chrome, Safari and Opera.

Fixes: https://github.com/ianstormtaylor/slate/issues/1693
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/utils/clone-fragment.js
+++ b/packages/slate-react/src/utils/clone-fragment.js
@@ -1,6 +1,4 @@
 import Base64 from 'slate-base64-serializer'
-import { IS_CHROME, IS_SAFARI, IS_OPERA } from 'slate-dev-environment'
-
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
I love this kind of PR, thank you @tobiasandersen!

PR #1837, opened 2018-05-14
avoid innerText

caub:
better to avoid innerText https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#Differences_from_innerText

Fixes: #
Reviewers: @


ianstormtaylor:
Sounds good, thanks @caub!

Issue #1527, opened 2018-01-12
Would you like to add flow type check in some files?


zhujinxuan:
#### Do you want to request a *feature* or report a *bug*?
feature;

#### What's the current behavior?
Some plugins are using flow to type check with the slate.  However, flow is not smart enough to intercept some types inside the slate. 

I am just wondering whether or not it is appreciated of submitting PRs of adding flow annotation for type checking in some files?

Thank you
Jinxuan Zhu

ianstormtaylor:
Hey @zhujinxuan, good question. I don't want to add Flow to the codebase though. If we were to consider one of the type systems down the road it would be TypeScript, but even that I think should be punted for nearer to a 1.0 release.

I think plugins can do whatever they please on this front, but it doesn't need to impact the core codebase.

PR #1838, opened 2018-05-14
Fix insertFragment: undefined newText

Soreine:

Fixing a bug introduced in https://github.com/GitbookIO/slate/commit/f07fc92f2e31a355a91f19ca080c2e380f883dae

In production, we have tracked this error several times already.

```
TypeError: Cannot read property 'object' of undefined
  at getLast(./node_modules/slate/lib/slate.es.js:1756:1)
  at apply(./node_modules/slate/lib/slate.es.js:1208:1)
  at apply(./node_modules/slate/lib/slate.es.js:1687:1)
  at collapseToEndOf(./node_modules/slate/lib/slate.es.js:1733:1)
  at apply(./node_modules/slate/lib/slate.es.js:6167:1)
  at call(./node_modules/slate/lib/slate.es.js:13116:1)
  at insertFragment(./node_modules/slate/lib/slate.es.js:13220:1)
  at apply(./node_modules/slate-react/lib/slate-react.es.js:3191:1)
  at run(./node_modules/slate/lib/slate.es.js:9592:1)
  at apply(./node_modules/slate-react/lib/slate-react.es.js:4160:1)
  at call(./node_modules/slate/lib/slate.es.js:13116:1)
  at change(./node_modules/slate-react/lib/slate-react.es.js:4142:1)
  at apply(./node_modules/slate-react/lib/slate-react.es.js:4159:1)
  at handler(./node_modules/slate-react/lib/slate-react.es.js:3910:1)
  at onEvent(./node_modules/slate-react/lib/slate-react.es.js:2084:1)
  at apply(./node_modules/slate-react/lib/slate-react.es.js:1967:1)
```

I could not write a failing test for the existing code. But the error happens when `newText` is undefined here https://github.com/ianstormtaylor/slate/blob/f07fc92f2e31a355a91f19ca080c2e380f883dae/packages/slate/src/changes/at-current-range.js#L171
Any idea to reproduce @ianstormtaylor ?

### DIFF
--- a/packages/slate/src/changes/at-current-range.js
+++ b/packages/slate/src/changes/at-current-range.js
@@ -168,7 +168,7 @@ Changes.insertFragment = (change, fragment) => {
   const newTexts = document.getTexts().filter(n => !keys.includes(n.key))
   const newText = isAppending ? newTexts.last() : newTexts.takeLast(2).first()
 
-  if ((newText && lastInline) || isInserting) {
+  if (newText && (lastInline || isInserting)) {
     change.select(selection.collapseToEndOf(newText))
   } else if (newText) {
     change.select(


SamyPesse:
@ianstormtaylor What do you think about adding [Flow](https://flow.org/) to the codebase ? It looks like a bug that could have been easily notified by some typing.

zhujinxuan:
@SamyPesse https://github.com/ianstormtaylor/slate/issues/1527  I asked ian about that.  It seems ian prefer to add when apis are stable.

ianstormtaylor:
@Soreine thanks for catching this. I'm not totally sure on reproduction, but potentially this case would trigger it:

```js
const fragment = (
  <document>
    <paragraph>
      <link>text</link>
    </paragraph>
  </document>
)

const input = (
  <value>
    <document>
      <paragraph>te<cursor/>xt</paragraph>
    </document>
  </value>
)

```

In that case a fragment that only results in a single text node is added in a non-appending way, resulting in `newTexts.takeLast(2).first()` returning undefined? Let me know if that helps for adding a test case.

---

@SamyPesse @zhujinxuan yeah I'd rather not commit to adding Flow, since I've never used it before and although I'm sure it would catch these kinds of bugs, it would also add a lot of extra work to the codebase and dissuade more users from contributing. If we were to choose a typing system, I'd rather go with TypeScript which seems to have a lot more community support. But even that would still add a large barrier to contributing for lots of folks.

tobiasandersen:
This can happen if you drop something onto the first node — if that node is void. The error then gets triggered by [this line](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/get-event-range.js#L40), since there's no previous node: 
```
range.moveToEndOf(document.getPreviousText(text.key))
```
Not sure what the best solution is though. I guess if there's no previous node, we could return the range of the next node instead. But what do we return if the only thing in document is a void node?

tobiasandersen:
I've opened https://github.com/ianstormtaylor/slate/pull/1849 for the case I described above.

SamyPesse:
@tobiasandersen Is this PR still needed since #1849 has been merged ? It looks like you've fixed a situation where it may occurs, but @Soreine in this PR prevents a potential crash from similar situations.

tobiasandersen:
I agree with what you just said. #1849 should fix one situation where this can occur, while this PR could prevent other potential cases.

ianstormtaylor:
I'm definitely down for fixing this bug, I just need this PR to include a test since otherwise it's very confusing to what it's fixing and will probably regress.

PR #1839, opened 2018-05-15
update schema customizer to merge marks

isubasti:
improving a feature

#### What's the new behavior?
document's marks in schema are now extendable

let say we have two plugins where each of them defines some marks to exist in the document such as
```js
const pluginA = {
  schema: {
    document: {
      marks: [{ type: 'bold' }, { type: 'italic'}]
    },
  }
}
const pluginB = {
  schema: {
    document: {
      marks: [{ type: 'underline' }, { type: 'strikethrough'}]
    },
  }
}

render() {
  return (
    <Editor plugins={[pluginA, pluginB]} />
  )
}
```

previously the Editor's schema would be:

```js
schema =  {
  document: {
    marks: [{ type: 'bold' }, { type: 'italic'}]
  },
}
```

with this PR the resulting schema would be:

```js
schema =  {
  document: {
    marks: [{ type: 'bold' }, { type: 'italic'}, { type: 'underline'}, { type: 'strikethrough' }]
  },
}
```

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate/src/models/schema.js
+++ b/packages/slate/src/models/schema.js
@@ -567,7 +567,7 @@ function resolveNodeRule(object, type, obj) {
  */
 
 function customizer(target, source, key) {
-  if (key == 'objects' || key == 'types') {
+  if (key == 'objects' || key == 'types' || key == 'marks') {
     return target == null ? source : target.concat(source)
   } else {
     return source == null ? target : source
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @isubasti!

PR #1842, opened 2018-05-15
Schema objects types interaction

Soreine:
## Is this adding or improving a _feature_ or fixing a _bug_?

Improving the `schema` feature.

Currently, there is no way to allow a given node type to live either under the `document` or under certain other types. 

For example, writing this ...

```jsx
// Nested lists schema
const schema = {
  list: {
    parent: {
      objects: ['document'],
      types: ['list']
    }
  }
}
```

... does not work. Because if a list is under another list (nested), the `objects` check does not pass since it is not under a `document`. You could add `block` to the `objects` list, but it would allow lists to live under any block.

Here is what the current implementation does:

|               | no objects          | objects valid       | objects invalid       |
| ------------- | ------------------- | ------------------- | --------------------- |
| **no types**      | ✅                  | ✅                  | PARENT_OBJECT_INVALID |
| **types valid**   | ✅                  | ✅                  | PARENT_OBJECT_INVALID |
| **types invalid** | PARENT_TYPE_INVALID | PARENT_TYPE_INVALID | PARENT_OBJECT_INVALID |

## What's the new behavior?

Now, instead of considering `objects` and `types` rules as conjunctive (AND), we consider them as disjunctive (OR), unless nothing was specified.

|               | no objects     | objects valid | objects invalid |
| ------------- | -------------- | ------------- | --------------- |
| **no types**      | ✅             | ✅            | PARENT_INVALID  |
| **types valid**   | ✅             | ✅            | ✅              |
| **types invalid** | PARENT_INVALID | ✅            | PARENT_INVALID  |

## How does this change work?

The new logic is also simplified, and just yields a `PARENT_INVALID` violation, without distinguishing between `PARENT_TYPE_INVALID` and `PARENT_OBJECT_INVALID`. I can keep `PARENT_TYPE_INVALID` and `PARENT_OBJECT_INVALID` if we want to add more specific `PARENT_X_INVALID` reasons later. Something like:

|               | no objects     | objects valid | objects invalid |
| ------------- | -------------- | ------------- | --------------- |
| **no types**      | ✅             | ✅            | PARENT_OBJECT_INVALID  |
| **types valid**   | ✅             | ✅            | ✅              |
| **types invalid** | PARENT_TYPE_INVALID | ✅            | PARENT_OBJECT_INVALID  |


ianstormtaylor:
Hey @Soreine, very interesting use case. I agree with you that this is something that should be supported.

I'm a little nervous about having `parent` be different than the other rules, in that it uses `OR` instead of `AND`. I wonder if there is another way we can do it.

My first thought is, what if we allowed for all of the top-level schema validation rule properties to also be defined as functions, which would get called with the target to validate. So in this case we could have functions of:

```js
parent: parent => parent.object === 'document' || parent.type === 'list_item'
```

Alternatively, we could change things to allow for arrays instead of just objects for the definitions. For example:

```js
parent: [
  { object: 'document' },
  { type: 'list_item' },
]
```

This one actually sounds nice to me, since it matches how AND vs. OR is handled in other programming areas. If we did do that, I think we'd need to change the `nodes` rule to not be an order-dependent array, and instead make it an AND array too. (This would eliminate the ability to specify exact ordering, but we can find another way to do this.)

What do you think?

Soreine:
> allow for arrays instead of just objects for the definitions

I like this solution better as well. I would like to avoid functions as much as possible for the schema. We should reserve the flexibility for `validateNode` for now, and converge to a good schema structure that we can manipulate easily, like ProseMirror's one.

I'm not sure how to specify nodes order differently though. Maybe we should have a concept of 'match':

```jsx
{
  nodes: [
    { match: [
       { type: 'image' },
       { type: 'video' }
      ],
      min: 1,
      max: 3 },
    { match: [
       { type: 'paragraph' }
      ],
      min: 0 
    },
  ]
}
```

And `parent` is itself a match array:

```jsx
{
  parent: [
    { type: 'image' },
    { object: 'document' }
  ]
}
```

> make [the `nodes`] rule an AND array too

Did you mean an OR array ? 😄

CameronAckermanSEL:
I don't want to dispute or diminish this PR, but one thing I think is worth pointing out is that the existing schema you describe is currently possible with 

```
// Nested lists schema
const schema = {
  list: {
    parent: {
      objects: ['document', 'block'],
      types: ['list', undefined]
    }
  }
}
```

Whether this is right or not is a different question, but seems like this PR makes the whole logistics of achieving this a bit better..

CameronAckermanSEL:
Yeah, and having to use 'undefined' here is a bit counter intuitive. I think an improvement to this area would be nice.

ianstormtaylor:
@Soreine Ah, I like the `match` solution! That's a nice way to handle it, and not really much more verbose at all. I'd love a PR that implemented arrays and changed the existing `objects/types` logic to use `match` instead everywhere. I like that it makes it super explicit about what the specific matchable nodes are.

> Did you mean an OR array ? 😄

Yup, I totally did :smile:

---

I agree with you about ideally finding a nice way to do it without functions. It's also potentially nice if schemas maintain JSON-serializability. Although not required.

Soreine:
I'm on it

ianstormtaylor:
@Soreine @SamyPesse any update on this?

Soreine:
Sorry I don't have the time right now to implement this.

On a side note, the code that applies the schema rules could also benefit from being refactored. I think it can be made lighter to digest (but identical in logic). This could be done in this PR, or in another PR. But I would feel bad to bluntly patch something onto the existing code.

PR #1846, opened 2018-05-18
Fix a hyperscript bug in perserving text-key

zhujinxuan:

Fix a bug;
For example, the following hyper-script 
```
<block>
"cat"
<inline/>
<text key="a">is cute</text>
</block>
```
will produce result like
```<block>
<text key="a">"cat"</text>
<inline/>
<text key="a">is cute</text>
</block>
```

#### What's the new behavior?
Fix that bug

### DIFF
--- a/packages/slate-hyperscript/test/default/perserve-keys-in-right-text.js
+++ b/packages/slate-hyperscript/test/default/perserve-keys-in-right-text.js
@@ -12,7 +12,8 @@ export const input = (
     </block>
   </document>
 )
-export function assertation() {
+
+export function test() {
   const block = input.nodes.first()
   assert.notEqual(block.nodes.first().key, 'a')
   assert.equal(block.nodes.last().key, 'a')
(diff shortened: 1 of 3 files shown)


packages/slate-hyperscript/test/default/perserve-keys-in-right-text.js
```diff
@@ -0,0 +1,72 @@
...
+
+export const input = (
+  <document>
+    <block type="paragraph">
+      Cat <inline type="link">is</inline>
+      <text key="a"> cute</text>
+    </block>
+  </document>
+)
+export function assertation() {
```
SamyPesse: I think `function test()` is a better name than `function assertation()`.

zhujinxuan: Great suggestion.  I make the rename in 98884c7.

ianstormtaylor:
Thank you @zhujinxuan for this. And thank you @SamyPesse for the review! Looks good now with that change made.

PR #1847, opened 2018-05-18
Perhaps Faster stack

zhujinxuan:

feature

#### What's the new behavior?

The old stack render use slice().reverse(), which costs time creating new array and reverse it.   This PR use reduceRight instead.


ianstormtaylor:
Thanks @zhujinxuan!

PR #1849, opened 2018-05-18
Fix invalid ranges in getEventRange

tobiasandersen:
Fixing [a bug](https://github.com/ianstormtaylor/slate/pull/1838#issuecomment-389915570):

kapture 2018-05-18 at 12 11 15

#### What's the new behavior?
When dropping something onto the first node's top part, when that node is void, we currently try to return the previous text node's range. This text node doesn't exist, and results in an error being thrown. Now we instead try to return the next text node's range. And if that doesn't exist, we return null (unless you have a better idea?). 

kapture 2018-05-18 at 12 26 32


ianstormtaylor:
Great call, thank you @tobiasandersen!

PR #1851, opened 2018-05-18
Faster search text in example

zhujinxuan:

bug

1. when `text===''` in search box, we will have a lot of empty decorations
2. text.split is slower than indexOf


ianstormtaylor:
Hey @zhujinxuan this example is just to give people an idea for what's possible, not to have the best performance possible. I'd rather keep it simple and easy to maintain rather than tweak for performance.

PR #1854, opened 2018-05-20
Allow html serialiser rules to return strings and arrays.

ara4n:
Currently the HTML serialiser seems to assume that serialisation rules
will always return a React element, which it then tries to clone and
add a `key` to.

However, it can be useful to have rules which return a plain
string (or array of strings and elements) - for instance serializing
fancy inlines (e.g. svg emoji) which when expressed as plain HTML should
just be text.  This changes the logic to only ever try to add keys to
React elements, and pass through strings/arrays/etc to renderToStaticMarkup
otherwise without changes.

Somewhere in between. It feels pretty buggy that a serialiser can't serialise nodes in your document to plain text, but it's probably a missing feature.

#### What's the new behavior?

You can now return strings (or arrays of strings & elements) from the `serialize` method of your custom HTML serializer rules - so they behave more like React render methods, and don't break mysteriously if you try to serialize a node to a plain string.

The change is trivial.

### DIFF
--- a/packages/slate-html-serializer/src/index.js
+++ b/packages/slate-html-serializer/src/index.js
@@ -337,7 +337,7 @@ class Html {
    * Serialize a `node`.
    *
    * @param {Node} node
-   * @return {String}
+   * @return {React.Element|Array<React.Element|String>|String}
    */
 
   serializeNode = node => {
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @ara4n thanks for this. I agree with you that the end goal of this should be possible. Is this not solveable though by using `React.Fragment` in React 16?

ara4n:
ah, quite possibly; i am still stuck on react 15. feel free to close if so.

ianstormtaylor:
I believe this is possible now that in React 16, so I'm going to close this.

Issue #1858, opened 2018-05-24
Margins are note collapsed because of `slate-data-spacer`


Banb4n:
#### Do you want to request a _feature_ or report a _bug_?

- Bug

Example to reproduce : https://jsfiddle.net/BanBn/gszke21n/2/
Informations: MacOS, Google Chrome, Slate 0.33

#### What's the current behavior?
![capture d ecran 2018-05-24 a 15 49 10](https://user-images.githubusercontent.com/25232681/40489754-24c2a6b4-5f6a-11e8-9209-6aa2c5b8132e.png)

The `data-slate-spacer` div prevents the collapsing of margin.

#### What's the expected behavior?
![capture d ecran 2018-05-24 a 15 49 28](https://user-images.githubusercontent.com/25232681/40489747-1d8bdb90-5f6a-11e8-8901-30433c3940d8.png)

At GitBook, we fixed this by adding `position: absolute` to the `data-slate-spacer` element. But we are not confident about this solution, since we are not sure to understand the purpose of this `spacer`.

ianstormtaylor:
Ah interesting. We've had it `position: absolute` in the past I think, and I'm open to going back to it again. It's honestly pretty hard to keep track of the exact CSS properties that are required to get it functioning cross-browser. But if `position: absolute` has no adverse effects with selecting into/across the void nodes I'm down with it.

ianstormtaylor:
The spacer's purpose is to capture the cursor when people move it with the arrow keys for instance, otherwise the entire void node in the tree is non-selectable.

Banb4n:
Hi @ianstormtaylor thx for the answer 👍, so I tested the moving of the cursor with the `spacer` position set to `absolute` and that has no negative effect. I tested it in: Chrome, Safari and Firefox

Banb4n:
Hi @ianstormtaylor thx for the answer 👍, so I tested in Chrome, Safari and Firefox the moving of the cursor with the spacer position set to absolute and that has no negative effect. 
I've edited the jsfiddle so you can test it to

ianstormtaylor:
@Banb4n nice, that's good news 😄 in that case I'd be happy to have a PR with that tweak!

Banb4n:
I made the change on this PR https://github.com/ianstormtaylor/slate/pull/1862, so you can check it

freeplant:
[...]

PR #1862, opened 2018-05-25
Set position data-slate-spacer block

Banb4n:
 - Fixing a bug 

#### What's the new behavior?

Now the `data-slate-spacer` block have a `position: absolute` to fix margins collapse

View issue #1858 to more details.

Fixes: #1858
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/void.js
+++ b/packages/slate-react/src/components/void.js
@@ -63,7 +63,7 @@ class Void extends React.Component {
       height: '0',
       color: 'transparent',
       outline: 'none',
-      position: 'absolute'
+      position: 'absolute',
     }
 
     const spacer = (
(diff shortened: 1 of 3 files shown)


freeplant:
In our project, we render Markdown documents using Slate. The images in a document are rendered as inline nodes. With this change (adding `position: 'absolute'`), when an inline image is clicked, the view will jump to an incorrect scroll position.

In `scrollToSelection()`  https://github.com/ianstormtaylor/slate/blob/1923888af1b07b6c21ea20b0a9f054a5cbb27654/packages/slate-react/src/utils/scroll-to-selection.js#L74

```
let cursorRect = range.getBoundingClientRect()
```

returns incorrect cursorRect.

willdawsonme:
@freeplant Try adding `position: relative;` to the Editor. It seems to have improved things for me

PR #1864, opened 2018-05-28
Fix the characters count in deleteForward

zhujinxuan:
bug

The deleteForwardLine is using a wrong characters number.  This PR fixes the problem.


ianstormtaylor:
Thanks @zhujinxuan and @SamyPesse! Looking good now.

PR #1865, opened 2018-05-29
Use cleaned up text content when copying into plain/text

tobiasandersen:
Fixing a bug.

#### What's the new behavior?
I was previously using the natively selected text when putting content into the `plain/text` register. That's problematic though since that text can contain zero width space characters. Since we already clean those up in our "dummy div", I just use the `textContent` of that div instead.


ianstormtaylor:
Good call, thanks @tobiasandersen!

Issue #1852, opened 2018-05-18
Impossible to select a void inline in Safari 11.1


justinweiss:
#### Do you want to request a _feature_ or report a _bug_?

Report a bug.

#### What's the current behavior?

In Safari 11.1, you can't select a void inline by clicking on it. You also can't navigate through it from the left side with the arrow keys.

I'm pretty sure this used to work, but I can't tell whether it was a bug introduced by Safari, or if it was a result of the change to empty text nodes on void nodes, since I made both changes in my demo app around the same time.

It appears that Safari will always force the selection to the end of the preceding text node, which causes this problem. I wonder if it's not considering zero-width spaces when deciding on the size of the inline.

Works fine in Chrome / Firefox.

Fiddle here (open your console, and it will print the list of selected inlines on every change): https://jsfiddle.net/oyerecen/

#### What's the expected behavior?

Clicking on the inline void node should select the node, and you should be able to use the keyboard to navigate through it.

justinweiss:
A few things I’ve seen so far:

When positioning a cursor at offset 0 of a node, Safari will normalize the native selection to be the end of the previous node. Chrome will allow offset 0 as a valid position. 

It seems that during the process of updating the selection, onNativeSelectionChange is called twice. I think this is because of the throttle, that it’s called on the trailing edge of the 100ms. This will almost certainly happen too late to be blocked by tmp.isSelectionUpdating.

So I think this is what’s happening. We have a native selection update on offset 0 of the inline node. This triggers onNativeSelectionChange, which triggers the onSelect stack. That sets the native selection to offset 0, which gets normalized to offset `previousNode.length`. This triggers another onNativeSelectionChange, which gets throttled but still happens later. This will go back through the plugin stack, setting the Slate selection to offset `previousNode.length`, which registers as outside of the inline. 

I don’t know for sure that this is what’s happening, but it seems to match what I’m seeing. I’ll dig into it more.

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #1866, opened 2018-05-30
Skip onNativeSelectionChange if Slate changed the selection

justinweiss:
Just like the `onSelect` handler, the plugin `onSelect` shouldn't be called if the selection was changed through Slate's selection API.

Otherwise, depending on how a browser normalizes a selection, the following can occur:

- Slate changes `value.selection`
- `updateSelection` updates the browser selection with the Slate selection
- The browser normalizes the browser selection, changing it
- `onNativeSelectionChange` is called with the new browser selection, which is now different than `value.selection`
- `after.js` `onSelect` is called with that new selection, and calls `.select`
- `updateSelection` updates the browser selection
- ... and so on

In Safari, which normalizes offset 0 of a node to be offset `previousNode.length`, this can cause the selection to creep leftward across a void inline, making that node impossible to select.

Fixing a bug.

#### What's the new behavior?

When clicking on a void inline in Safari, `anchorInline` will return the correct inline node.

By blocking `onNativeSelectionChange` if `updateSelection` caused the native selection change, we avoid the native selection update from cycling back through into the Slate `Value`'s selection. This avoids problems with different browser selection normalization from affecting the Slate `Value`.

Fixes: #1852
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -9,7 +9,6 @@ import {
   SUPPORTED_EVENTS,
 } from 'slate-dev-environment'
 import logger from 'slate-dev-logger'
-import throttle from 'lodash/throttle'
 
 import EVENT_HANDLERS from '../constants/event-handlers'
 import Node from './node'
@@ -427,7 +426,8 @@ class Content extends React.Component {
    * @param {Event} event
    */
 
-  onNativeSelectionChange = throttle(event => {
+  onNativeSelectionChange = event => {
+    if (this.tmp.isUpdatingSelection) return
     if (this.props.readOnly) return
 
     const window = getWindow(event.target)
@@ -435,7 +435,7 @@ class Content extends React.Component {
     if (activeElement !== this.element) return
 
     this.props.onSelect(event)
-  }, 100)
+  }
 
   /**
    * Render the editor content.


justinweiss:
There are a few different ways to fix the problem. I chose to remove the `throttle`, because I thought it would be easier to debug if something went wrong, and onNativeSelectionChange wasn't called all that frequently in my tests. If we brought back the throttle, we'd also probably have to defer resetting `isUpdatingSelection`, otherwise we wouldn't be able to successfully block the selection change. And if we defer the reset, we might miss valid, non-programmatic selection changes. It's unlikely, but it could happen.

packages/slate-react/src/components/content.js
```diff
@@ -9,7 +9,6 @@ import {
   SUPPORTED_EVENTS,
 } from 'slate-dev-environment'
 import logger from 'slate-dev-logger'
-import throttle from 'lodash/throttle'
```
SamyPesse: I don't know about the specific fo this PR and if it's the right solution, I'll let @ianstormtaylor reviews it.

But if we remove this include, we can remove `lodash/throttle` from the dependencies in `package.json` (I'm always happy to see lodash being removed from dependencies 😍).

justinweiss:
There might be another option for this fix, I'll look into it if I have time:

I think this started to happen when void nodes changed to allow any content in their text, instead of restricting it to a single space. Last time I checked, inline voids render a single space to the Browser DOM, but internally, Slate uses `text.size` to determine the length of the inline when figuring out range boundaries. This makes it easy for Slate and the browser to have very different ideas about which part of the inline is selected -- for example, if the void has 0 characters, and you click to the right of the inline, the browser might try to select offset 1, which doesn't exist in the Slate model.

Again, not super sure about this, but it's around the time I started to notice the problem.

packages/slate-react/src/components/content.js
justinweiss: @SamyPesse Looks like we're including lodash as a single package, and not as individual pieces I can remove, unfortunately.

justinweiss:
It looks like this might have been fixed at the void node level, and even if it hasn't, it probably needs to be rethought for newer versions of Slate. So I'll close it for now.

PR #1867, opened 2018-05-30
[On Development] Prepare for React-v17

zhujinxuan:
This PR changes the `componentWillRecieveProps` to make it more alike `getDerivedStateFromProps`

Then when react v17 is released or when we decide to move to React v16 and drop v15 support, then we can easily change the `componentWillRecieveProps` to `getDerivedStateFromProps`

### DIFF
--- a/packages/slate-react/src/components/editor.js
+++ b/packages/slate-react/src/components/editor.js
@@ -73,24 +73,29 @@ class Editor extends React.Component {
 
   constructor(props) {
     super(props)
-    this.state = {}
-    this.tmp = {}
-    this.tmp.updates = 0
-    this.tmp.resolves = 0
+
+    this.tmp = {
+      updates: 0,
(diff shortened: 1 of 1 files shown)


zhujinxuan:
@SamyPesse I did not know that polyfill.  I think that polyfill works better.

I will colse this PR;

SamyPesse:
@zhujinxuan I was not suggesting that we should close this Pr, but instead switch the code to use the React16/17 APIs (such as `getDerivedStateFromProps`) now in this PR and setup the polyfill.

zhujinxuan:
@SamyPesse Ohhhh, my mistake.  I reopen this PR and will finish this three weeks later.

zhujinxuan:
Close this PR; the lifecycle works will be in https://github.com/ianstormtaylor/slate/pull/1975

PR #1869, opened 2018-05-31
Allow user to use immutable 4.0.0-rc without yarn warning

zhujinxuan:
Feature

Some developers or users may starts to use immutable 4.0.0-rc with Slate, to prepare for the future upgrade or for some new features.

This changes the semver in package json, so using immutable 4.0.0-rc will not result to a `yarn check` error.


packages/slate-plain-serializer/package.json
```diff
@@ -16,7 +16,7 @@
     "slate-dev-logger": "^0.1.39"
   },
   "peerDependencies": {
-    "immutable": "^3.8.0",
+    "immutable": "^3.8.0 | ^4.0.0-rc",
```
ianstormtaylor: I think this `|` single bar will fail? And needs to be a `||` double bar?

packages/slate-html-serializer/package.json
```diff
@@ -17,7 +17,7 @@
     "type-of": "^2.0.1"
   },
   "peerDependencies": {
-    "immutable": "^3.8.1",
+    "immutable": "^3.8.1 || ^4.0.0-rc",
```
ianstormtaylor: This is a good call @zhujinxuan. I think the ideal thing we should be doing is to instead make all of the peer dependencies use `>=` so that we never really have to update them like this. So this would change to:

```json
"immutable": ">=3.8.1",
```

And even the other ones simply:

```json
"react": ">=0.14.0",
"react-dom": ">=0.14.0",
...
```

Would you be down to make that change for all of them? I think it would end up making the codebase easier to manage for us and less restricting for everyone.

zhujinxuan:
@ianstormtaylor Hi, it is fixed~

zhujinxuan:
@ianstormtaylor Hi, versions of immutable, react and react-dom are fixed~

ianstormtaylor:
Thank you @zhujinxuan this is great!

denis-sokolov:
This does not work in yarn 1.12.0 and npm 6.9.0:

```
error "slate#immutable@>=3.8.1" doesn't satisfy found match of "immutable@4.0.0-rc.12"
error "slate-react#immutable@>=3.8.1" doesn't satisfy found match of "immutable@4.0.0-rc.12"

npm ERR! peer dep missing: immutable@>=3.8.1, required by slate@0.44.12
npm ERR! peer dep missing: immutable@>=3.8.1, required by slate-react@0.21.20
```

[Per npm docs](https://docs.npmjs.com/misc/semver#prerelease-tags), pre-release versions only satisfy constrains in the same per-release version collection.

In order to satisfy, the requirement must manually opt-in: `>=3.8.1 | >4.0.0-rc`.

zhujinxuan:
@denis-sokolov Thank you.  I will make a PR to fix tomorrow.

denis-sokolov:
Thank _you_ for the changes!

zhujinxuan:
@denis-sokolov I make a PR about that. I think the version problem shall be fixed after merging.
https://github.com/ianstormtaylor/slate/pull/2695

Cheers,
Jinxuan Zhu

Issue #1792, opened 2018-04-24
Invalid marks in schema are not removed


PierBover:
#### Bug report

With this schema it shouldn't be possible to have any marks in the `heading-one` block:
```
const schema = {
  blocks: {
    'heading-one': {
      nodes: [{ objects: ['text'] }],
      marks: []
    }
  }
};
```
But as you can see I'm able to insert bold text in it:
![image](https://user-images.githubusercontent.com/226483/39166782-2aca079a-4751-11e8-98d5-b989999c9889.png)

#### What's the current behavior?
You can see this behaviour in action here:
https://youtu.be/mfI3U3KX9ug

To reproduce I'm just toggling a mark without actually making a selection and then simply writing.

CameronAckermanSEL:
Try `marks: ['']`

CameronAckermanSEL:
Also there's a [slack channel](https://slate-js.slack.com/messages/C1RH7AXSS/), which is a better place to ask support questions like this:

PierBover:
Thanks @CameronAckermanSEL but I'm getting the same result with either `[]` or `['']`.

CameronAckermanSEL:
The `['']` should definitely work so there's something else in play. 

Should probably go to the slack channel for some support.

CameronAckermanSEL:
Sorry, misunderstood. Able to reproduce this bug, specifically when inserting text. Appears to be because the insertion of text doesn't cause normalization so it doesn't pick up schema violations like this anymore.

PierBover:
Is this a good first task? I'd like to contribute to Slate.

zhujinxuan:
@PierBover Yes, you can fix it here: packages/slate/src/changes/by-key in insertTextAtRange
```
  if (normalize !== undefined) {
    normalize = range.isExpanded
  }
```
to something like
```
  if (normalize !== undefined) {
    normalize = range.isExpanded && marks.size !== 0
  }
```

PierBover:
Ok I'll take a look tonight.

SmilinBrian:
A discussion with "bupkis" on Slack on May 29th shows that this issue also applies to cases where the editor is intended to automatically convert certain sequences of characters typed into macros or something.

I made a JSFiddle showing this issue so that this other aspect of this problem can be fixed/tested as well:  https://jsfiddle.net/SmilinBrian/d5qrhtgx/2/

zhujinxuan:
[...]

PR #1875, opened 2018-06-04
InsertTextAtRange normalize

isubasti:
Fixing a bug

#### Curent Behaviour
inserttext not normalized mov

#### What's the new behavior?
inserttext fix normalization mov

previously the *PERF* normalization check isn't checking the right thing and also text that has marks that potentially needs to be normalized due to schema isn't normalized as well.This change is basically fix the *PERF* check and putting the correct normalization flag when needed.

Fixes: #1792 

Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate/src/changes/at-range.js
+++ b/packages/slate/src/changes/at-range.js
@@ -863,7 +863,7 @@ Changes.insertTextAtRange = (change, range, text, marks, options = {}) => {
   }
 
   // PERF: Unless specified, don't normalize if only inserting text.
-  if (normalize !== undefined) {
+  if (normalize === undefined) {
     normalize = range.isExpanded
   }
   change.insertTextByKey(key, offset, text, marks, { normalize: false })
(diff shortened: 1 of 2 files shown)


gggdomi:
I started a similar PR (#1823) almost a month ago, but have been off before finishing it.
If this one is merged instead of #1823, you should take a look at the discussion about getFlag and not overriding user-specified options.

Also, I had troubles writing tests for normalization, I'll add them to #1823 so you can see what's happening

isubasti:
@gggdomi thanks for the linked PR, I think I know how to fix the test. Will get to it on the weekend

isubasti:
don't merge, will add test to make sure it won't regressed

ianstormtaylor:
Thank you @isubasti and @gggdomi! This change looks great, thanks for catching that `!== undefined` issue.

@gggdomi does this fix the same issue as your PR? or do they overlap but not quite fix it? If you want to rebase yours after I merge this one, in case there is extra logic needed to fix your case, I'd be happy to check it out then.

packages/slate/src/changes/at-range.js
```diff
@@ -863,8 +863,8 @@ Changes.insertTextAtRange = (change, range, text, marks, options = {}) => {
   }
 
   // PERF: Unless specified, don't normalize if only inserting text.
-  if (normalize !== undefined) {
-    normalize = range.isExpanded
+  if (normalize === undefined) {
+    normalize = range.isExpanded && marks.size !== 0
```
t3chguy: marks needs a null-guard, given its not even in the docs. https://docs.slatejs.org/slate-core/change#inserttextatrange

PR #1876, opened 2018-06-04
Fix #1463: don't split inline tags when wrapping a range in an inline

### LINKED ISSUE #1463: Do we need a `wrapSelectionInline()` method?
Feature

Just stumbled into this unexpected behavior with `wrapInline()` method inside a single inline parent. The current behavior splits the parent inline into 3 inlines and wraps the middle into the one supplied to the method. This is apparently very much intentional. I'm failing to come up with any use case that I would want this to happen.

I would expect my selection to be wrapped by the inline and not have side effects on parts that I haven't selected. Here's the expected use case:
https://github.com/ianstormtaylor/slate/pull/1460/files

So my question here is obvious: Is there a different method to achieve this? How are you guys using `wrapInline()`? How do I wrap my selection in an inline without destroying the parent structure?

crisward:

Allows inline tags not to be split when wrapping a range. Fixes #1463 

#### What's the new behavior?

This allows `wrapInlineAtRange` to do just that. If the selection is in the same inline element, it just wraps it without splitting.

It checks if the startInline matches the endInline keys, then skips the spliting and wraps the selection with the new inline tag.

Fixes: #1463
Reviewers: @


packages/slate/src/changes/at-range.js
```diff
@@ -1366,7 +1368,20 @@ Changes.wrapInlineAtRange = (change, range, inline, options = {}) => {
   const startIndex = startBlock.nodes.indexOf(startChild)
   const endIndex = endBlock.nodes.indexOf(endChild)
 
-  if (startBlock == endBlock) {
+  if (startInline && startInline == endInline) {
+    const text = startInline.text.substring(startOffset, endOffset)
+    inline = inline.set('nodes', List([Text.create(text)]))
```
SamyPesse: It may result in lost marks since you are taking the string and reconstructing a `Text`, right ?

Maybe on `Node`, we should have a method `Node.getSubText(startOffset: number, endOffset: number): Text` ?

crisward: Thanks. I'll add a test with overlapping marks and see if I can somehow construct the element with marks in place.

crisward:
Got this working with marks too, also added a test case for that. Can you resolve conflicts or do you need me to rebase.

crisward:
Api has changed - will need to work out a different way doing this.

crisward:
Got it working with the new leaf nodes. Bit more going on than there would have been with characters, but all tests are passing.

zhujinxuan:
Hi, you can use splitLeaves method from Leaf.  See example at https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/models/text.js#L271

BTW, the .eslintrc have changed.  Would you like to re-lint again? https://github.com/ianstormtaylor/slate/pull/1924

crisward:
Will take a look at splitNodeByKey and update eslint.
Thanks.

crisward:
Any reason not to merge this?

zhujinxuan:
Just because ian does not have time to review this PR in last month (lol)

crisward:
@zhujinxuan time aside, who really likes reviewing pull requests? Especially as many contributions are a bit off target to say the least. (probably including mine). Hopefully not though. I may have to work from my fork with crossed finger for now.

packages/slate/src/changes/at-range.js
```diff
@@ -1389,7 +1392,24 @@ Changes.wrapInlineAtRange = (change, range, inline, options = {}) => {
   const startIndex = startBlock.nodes.indexOf(startChild)
   const endIndex = endBlock.nodes.indexOf(endChild)
 
-  if (startBlock == endBlock) {
+  if (startInline && startInline == endInline) {
+    const text = startBlock
+      .getTextsAtRange(range)
+      .get(0)
+      .splitText(startOffset)[1]
+      .splitText(endOffset - startOffset)[0]
```
zhujinxuan: Could you add a blank line after the multi-lines expression?

packages/slate/src/changes/at-range.js
```diff
@@ -1389,7 +1392,24 @@ Changes.wrapInlineAtRange = (change, range, inline, options = {}) => {
...
+    inline = inline.set('nodes', List([text]))
+    Changes.insertInlineAtRange(change, range, inline, { normalize: false })
+    const inlinekey = inline.getFirstText().key
+    const rng = {
+      anchorKey: inlinekey,
+      focusKey: inlinekey,
+      anchorOffset: 0,
+      focusOffset: endOffset - startOffset,
+      isFocused: true,
+    }
```
zhujinxuan: Could you add a blank line after multi-lines expression here, too?

zhujinxuan:
@crisward I can help you review it.  But it is up to ian to merge PRs.

ianstormtaylor:
Thanks @crisward, looks good! Thanks for writing those tests.

Issue #1571, opened 2018-01-29
Bug: insert value to an empty table cell does not work


amitm02:
#### What's the current behavior?

In the table demo [link](http://slatejs.org/#/tables), it is impossible to write a value to a table cell if the cell is empty

![pw62t2kgsq](https://user-images.githubusercontent.com/1888927/35520610-298da2ec-0520-11e8-910b-455d6205801f.gif)

schneidmaster:
Thanks for the report, I can reproduce this as well (macOS/Chrome 64)

zhujinxuan:
Strange.  I can reproduce it in my macbook.  And there is no error thrown in the console.

In the debug mode; we have that `slate:editor onChange change.opearations === 0`

zhujinxuan:
The bug is caused by 
```
  onKeyDown = (event, change) => {
    const { value } = change
    const { document, selection } = value
    const { startKey } = selection
    const startNode = document.getDescendant(startKey)

    if (selection.isAtStartOf(startNode)) {
      const previous = document.getPreviousText(startNode.key)
      const prevBlock = document.getClosestBlock(previous.key)

      if (prevBlock.type == 'table-cell') {
        event.preventDefault()
        return true
      }
    }

```

Where 
```
      if (prevBlock.type == 'table-cell') {
        event.preventDefault()
        return true
      }
```
disable the text insert;

I think we shall only disable delete operations; or at least pass the insert operations

schneidmaster:
Yeah I agree, that seems to be correct. Looks like it's only a problem with the example implementation and not with slate itself. @zhujinxuan want to send a PR fixing that?

zhujinxuan:
@schneidmaster Yes, give me some minutes~

zhujinxuan:
A bug happens when insert
Strangely, the text is bolded

http://g.recordit.co/jT3HKSS6ME.gif

schneidmaster:
Hmm, looks like it's carrying over the bold mark from the previous cell? (Maybe due to #1561)

zhujinxuan:
Let us fix it after https://github.com/ianstormtaylor/slate/issues/1561

I think perhaps there is a bug of carrying mark in selection normalization.

zhujinxuan:
We can fix that after https://github.com/ianstormtaylor/slate/pull/1645

amitm02:
Any updates regarding this bug?

PR #1880, opened 2018-06-07
Enable insert in the start of table cell

zhujinxuan:

bug; In the table-cell example, users cannot insert at the beginning of the cell.
Fix: https://github.com/ianstormtaylor/slate/issues/1571

#### What's the new behavior?
People can insert in the start of the cell

Only blocks enter, backspace and delete at the beginning of the cell

Fixes: #1571
Reviewers: @amitm2


ianstormtaylor:
Good call @zhujinxuan, thanks!

PR #1881, opened 2018-06-08
[On Development] Fix cycle reference in slate-core 

zhujinxuan:

There exists a cycle reference in the `packages/slate/src/models`, where `node` depends on `block`, `inline` and `document`, which of all three depneds on `node` as well.

Currently, the node is hacking the cycle reference by `alleviating function`, but I am not sure whether this hacking will be always working in future ES and babel.

#### What's the new behavior?
Remove the cycle reference

put the fromJSON and isType logic in another file.

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate/src/constants/model-types.js
+++ b/packages/slate/src/constants/model-types.js
@@ -22,6 +22,18 @@ const MODEL_TYPES = {
 }
 
 /**
+ * Export type identification function
+ *
+ * @param {string} type
+ * @param {any} any
+ * @return {boolean}
+ */
+
+export function isType(type, any) {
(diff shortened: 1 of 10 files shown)


ianstormtaylor:
Hey @zhujinxuan, I like the idea of eliminating this weirdness. But I'd like to find a solution that makes the `./block.js`, `./inline.js` and `./document.js` directly requireable, instead of making those things exported from the `./node.js` file, since that is going to be confusing.

zhujinxuan:
We can import Block from ./block;  I export them from `node.js` because the mix-in happens in the node.js; but it is fine to export from block.js (because module.exports namespace is just a reference).

Update in recent commit.

Issue #1884, opened 2018-06-09
change void nodes to pass the spacer as `props.children`


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now when rendering a void node, Slate wraps the node in a `<div>` or `<span>` so that it can put a "spacer" next to it. This is needed to capture the browser's native selection behavior, since the void node has `contenteditable="false"`.

However, this also means that you can't control the top-level element that is rendered in the editor. Which is unfortunate in many use cases. For example one where you want to style adjacent void nodes. You can't just use the `& + &` CSS selector, since they aren't actually adjacent in the DOM.

#### What's the expected behavior?

I think...

We could solve this by passing the spacer as `props.children`, and requiring that users render it in their void node. (This is even more consistent with how we render non-void nodes actually.)

Similarly, we'd add the `contenteditable="false"` to the `props.attributes`, which would ensure that the top-level element of the void node remains ineditable.

bengotow:
Hey folks—it seems like this issue has been merged and can probably be closed!

ianstormtaylor:
@bengotow I don't think it has yet.

bengotow:
Hey thanks for the reply! It looks like my problem was that I was upgrading from an older version of Slate to the latest version and didn't realize that `isVoid: true` must now be specified in the schema and not on Block objects directly. The void node I was debugging was actually behaving as non-void and had a single child text node.  Thanks!

adjourn:
Experimented a little bit with this, see [this commit](https://github.com/ianstormtaylor/slate/compare/master...adjourn:feat/new-voids).

It works wonderfully like this but I don't see any ways to decrease the amount of properties user has to add/spread because selection works properly only with the following structure:

```jsx
// in renderBlock
return (
  <div {...props.voidAttrs}> {/* data-slate-void="true" data-key="1871" */}
    {props.spacer}
    <div contentEditable={false}> {/* has to be here like everything else */}
      <img
        {...attributes} {/* data-slate-object="block" data-key="1871" */}
        src={src}
        className={css`
          display: block;
          max-width: 100%;
          max-height: 20em;
          box-shadow: ${isFocused ? '0 0 0 2px blue;' : 'none'};
        `}
      />
    </div>
  </div>
)
```

If you remove any elements, selection is messed up. 
I guess it could be achieved by changing selection logic somewhere else.

ianstormtaylor:
[...]

PR #1886, opened 2018-06-10
Inline Void block cursor

isubasti:
fixing cursor being rendered on an inline void block where you won't be able to type anything inside

#### Current behavior?
contenteditablenull mov

#### What's the new behavior?
cursor is not rendered on an inline void block
contenteditablefalse mov

adding `contentEditable={false}` on inline void node only. Previously this was the case for both block and inline void. It had problem mentioned in #1639 that was fixed on #1734 but introduced unnecessary cursor on an inline void. as this only do `contentEditable={false}` only on inline void, issue from #1639 is not there as well 👍 

a check from @benjycui would be great to make sure that #1639 issue is not reggressed
Fixes: #
Reviewers: @benjycui

### DIFF
--- a/packages/slate-react/src/components/void.js
+++ b/packages/slate-react/src/components/void.js
@@ -78,7 +78,11 @@ class Void extends React.Component {
     this.debug('render', { props })
 
     return (
-      <Tag data-slate-void data-key={node.key} contentEditable={false}>
+      <Tag
+        data-slate-void
+        data-key={node.key}
+        contentEditable={readOnly || node.object == 'block' ? null : false}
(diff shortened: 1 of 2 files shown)


isubasti:
@SamyPesse it droves me and my UX/UI designer crazy as well. This will do for now util we can get something like what @ianstormtaylor describe on https://github.com/ianstormtaylor/slate/issues/1884. With something like that we'll have more control over it although need to be properly document to make sure that spacer is there

ianstormtaylor:
Thanks @isubasti!

PR #1891, opened 2018-06-11
Add styling rules for having blank lines before and after multiline blocks

zhujinxuan:

feature

#### What's the new behavior?

It is very easy to forgot some styling things during PR.  One common problem is that we forgot to add padding lines before and after multi-line blocks;  This PR will help to automatically add blank lines before and after multi-line blocks

This PR also puts all styling thing into a new package `slate-dev-lints`; then we can extend config from that package.

Fixes: #
Reviewers: @


zhujinxuan:
Close this PR;  I will try to submit this rule to eslint.

It is not clean to maintain rules in the same package.

Issue #1890, opened 2018-06-11
Code coverage analysis


zhujinxuan:
#### Do you want to request a _feature_ or report a _bug_?

feature

#### What's the current behavior?

SlateJS is complex enough, and is requiring tests.  But we cannot clearly know which part of code are not tested yet.  Code coverage can help us find out untested code in `packages/src/slate`

#### What's the expected behavior?
I think it will be good to have coveralls.io introduced in slate.  However, only maintainer/owner can set up code coverage analysis in coverall.io. @ianstormtaylor @Soreine @SamyPesse

zhujinxuan:
I find how to do that.  I will submit a code coverage PR this weekend.

ianstormtaylor:
Thanks @zhujinxuan!

PR #1896, opened 2018-06-13
Add cover coverage tests

zhujinxuan:

feature
1. add code coverage
2. In configuring code coverage, I switch the test tool from mocha to jest.  Because we need path alias to make slate-hyperscript use source files instead of lib files
3. faster test; we do not need to run `yarn build` for test.  We only build when lint code.
4. faster linting CI.  Becuase yarn $LINT_TYPE is run in parallel, I split the `yarn lint` with `yarn lint:code` and `yarn lint:document` then it can be run faster.   And it is less confusing when contributors see an error by `lint:document` than just `lint`

#### What's the new behavior?

Sometimes we make careless mistakes just by forgetting to test, or decide to add tests later but forgot.  (for example, PR https://github.com/ianstormtaylor/slate/pull/1864 is an example that a test is missing)

Code analysis can help us eliminate these problems, and suggests us which tests are perhaps missing.  And new PR contributors can look at the code coverage difference by code.

Add codeCov as code coverage service.  The following is an example about which files/lines are tested and which are not yet tested.  When submitting PR, we can check which newly added lines are not yet tested. 

screen shot 2018-06-13 at 6 11 21 pm
screen shot 2018-06-13 at 6 18 28 pm

For example

Fixes: #1890
Reviewers: @


zhujinxuan:
Here is a demo after code coverage is activated https://github.com/zhujinxuan/slate/pull/4.  


Here is what we need we do before merging this PR: 
1. Go to https://codecov.io/gh/zhujinxuan/slate and register an account with github
2. activate ianstormtaylor/slate in 
3. merge this PR; or close and reopen this PR to see how codecov works

Kornil:
lots to improve coverage wise :)

ianstormtaylor:
Hey @zhujinxuan, thanks for this!

I personally prefer to use Mocha over Jest. But I'd be okay with the change as long as we maintain the same level of useful-ness in error messages. Right now it looks like they get kind of meaningless:

image

And only if we can maintain the usefulness without hacks, and without having to use the `expect` library which I don't think is good.

zhujinxuan:
@ianstormtaylor   I agree that the `expect` interface is bad, and totally agree with your suggestion in https://github.com/facebook/jest/issues/2237.  
I think a solution is to wrap `expect` with `assert` interface.  So in the recent commit, I use `https://github.com/jungomi/jest-t-assert` as a method to use asset interface.

Do you think that is good?

zhujinxuan:
Close this PR because some travis.ci is absorbed by this PR

zhujinxuan:
@ianstormtaylor Hi, I replace assert as `jest-t-assert` in the recent commit.  Is that fine?

ianstormtaylor:
@zhujinxuan could you screenshot a comparison of what a breaking test looks like with `master` vs. this pull request?

zhujinxuan:
The below is for this PR:
screen shot 2018-06-16 at 9 40 00 pm

The below is for master
screen shot 2018-06-16 at 9 42 43 pm

packages/slate-hyperscript/test/index.test.js
```diff
@@ -2,7 +2,7 @@
  * Dependencies.
```
ianstormtaylor: Can we change the config to not have to add the `.test.js` suffix?

packages/slate/test/index.test.js
```diff
@@ -2,25 +2,25 @@
...
  */
 
 describe('slate', () => {
+  /**
+   * Reset Slate's internal key generator state before each text.
+   */
+
+  beforeEach(() => {
+    resetKeyGenerator()
+  })
```
ianstormtaylor: Can you remove the comment here, and simplify to:

```js
beforeEach(resetKeyGenerator)
```

packages/slate-hyperscript/test/index.test.js
zhujinxuan: I have not yet found out a way of that unfortunately.  We have to have either `__test__` or `.test.js`

packages/slate/test/index.test.js
zhujinxuan: Fixed~

.gitignore
```diff
@@ -7,6 +7,9 @@ lib/
 # Temporary files.
 tmp/
 
+# coverage files
+coverage
```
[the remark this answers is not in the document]
zhujinxuan: Fixed~

package.json
```diff
@@ -84,13 +87,16 @@
     "build:production": "cross-env NODE_ENV=production rollup --config ./support/rollup/config.js && cross-env NODE_ENV=production webpack --config support/webpack/config.js",
     "clean": "lerna run clean && rm -rf ./node_modules ./dist ./build",
     "gh-pages": "gh-pages --dist ./build",
-    "lint": "eslint packages/*/src packages/*/test examples/*/*.js examples/dev/*/*.js && prettier --list-different '**/*.{js,jsx,md,json,css}'",
+    "lint": "yarn lint:code && yarn lint:document",
+    "lint:code": "eslint packages/*/src packages/*/test examples/*/*.js examples/dev/*/*.js",
+    "lint:document": "prettier --list-different '**/*.{js,jsx,md,json,css}'",
```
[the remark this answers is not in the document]
zhujinxuan: Fixed~

zhujinxuan:
Hi, @ianstormtaylor Hi @ianstormtaylor . Would you like to have a look on this PR when you have time?

We will have several feature PRs coming, 
1. move focus to `value`
2. point model
3. decoration model
4. restruct getPath and related node methods.

I want to ensure the newly added functions have an about `100%` branches coverage when submitting PRs.

zhujinxuan:
Rerun the travis CI by reopen

packages/slate-hyperscript/test/index.test.js
ianstormtaylor: @zhujinxuan I think it can happen with the `testMatch` configuration option? https://jestjs.io/docs/en/configuration#testmatch-array-string

zhujinxuan: Great!  That is what I am looking for.  I will fix it minutes later.

zhujinxuan: Fixed~  Thank you

zhujinxuan:
Fix file conflicts~

zhujinxuan:
Hi, @ianstormtaylor .  The conflict of this PR is resolved after https://github.com/ianstormtaylor/slate/pull/2021.  I make a small dev-only change on KeyUtils and dev-fixtures because `jest` supports neither `this.skip` nor `describe(name, async)`

ianstormtaylor:
Hey @zhujinxuan, thank you for all your help here researching how to get code coverage reports. I'm sorry for my poor communication on this pull request. I really like the goal of having code coverage, but I'm just hesitant about switching from Mocha to Jest for lots of small reasons that add up...

### Diffs still aren't as good

image

image

There are now extra `Object` `Array` prefixes all throughout the diffs when looking at the JSON output of Slate. It's a small thing, but it's just a lot of noise.

### Filtering tests isn't as good

image

Whereas Mocha has the `--fgrep` option, jest has `-t` but every suite that it doesn't find a match for fails, and then it exists with a `1` error code. This isn't that nice because it's the main way to debug individual tests.

### Assertions aren't as good

image

Since Jest doesn't support the default `assert` module properly, we have to hack around it with the `jest-t-assert` module, but it doesn't have the exact same API as `assert`. Again it's a small thing, but just extra confusion.

### Configuration isn't as good

image

Jest requires more configuration that's going to be more to maintain over time to figure out why `lerna-aliases` isn't working, or whatever it ends up being. Mocha by contrast was very simple, just point it at an `index.js` file.

### Bailing early doesn't work

There's a Jest CLI flag called `--bail`, but it doesn't actually bail out of the tests early like Mocha does... it appears to be broken?

### Others feel similarly

I was thinking about this yesterday when I say [Tim Oxley tweet](https://twitter.com/secoif/status/1025303928705961985) about his troubles with Jest, that are some of the same "buildup of small issues" type worries I have.

---

All that to say, that's why I haven't merged this sooner. I'm just slightly worried about it. That's not to say I won't merge it... I'd just like to be more sure.

Why is it that this is currently not possible with Mocha? Why do we need to switch to Jest? From what I can tell, Coveralls (or others) work with Mocha too?

zhujinxuan:
I will have a try on mocha this weekend.  I forgot the jest thing.  When I
was beginning with this PR, I knew little about coverage workflow.  It was
easier to do it with jest. I will try to switch back to mocha this weekend.

ianstormtaylor:
@zhujinxuan thank you, I really appreciate it. I totally understand that feeling of forgetting why in the first place. Thank you for all of your help!

Issue #1794, opened 2018-04-24
this.document.getPreviousInline is not a function


kalley:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?
When calling `value.previousInline`, and error gets thrown because `this.document.getPreviousInline` is not a function

fiddle: https://jsfiddle.net/qmxu8ocy/ (type anything)

OS: mac
browser: Chrome
slate version: 0.33.4

#### What's the expected behavior?
It would return the previous inline if it existed.

I'm getting around it right now by using `value.document.getPreviousSibling` and checking if it's inline or not. I looked through history and I'm not sure that `getPreviousInline` ever existed :man_shrugging:

davidchang:
seems like https://github.com/ianstormtaylor/slate/blob/9e50bc6b1fd881e7a913ae890481316714c7bf59/packages/slate/src/models/value.js#L472-L490 should be removed from the value API, though that might warrant a minor version bump? looks like neither `previousInline` nor `nextInline` in the Slate codebase itself. unless we want to implement `getPreviousInline` and `getNextInline` on document

ianstormtaylor:
Open to a PR that adds those two.

ianstormtaylor:
As of https://github.com/ianstormtaylor/slate/pull/3093 (which was just merged), I believe this issue is no longer applicable, because a lot has changed. I'm going through and closing out any potential issues that are not out of date with the overhaul. Thanks for understanding.

PR #1897, opened 2018-06-13
[models/node] implement getPreviousInline and getNextInline APIs

davidchang:

Fixing a bug by adding previously undefined APIs.

#### What's the new behavior?

Previously, trying to use `document.getPreviousInline` and `document.getNextInline` were not defined, so using `value.previousInline` and `value.nextInline`, which depend on those methods, would cause a runtime error. Since the APIs were previously undefined, I don't think I should need to worry too much about backwards compatibility.

From https://github.com/ianstormtaylor/slate/issues/1794, https://jsfiddle.net/qmxu8ocy/ should start working. I also just wrote unit tests, which is how I developed/tested this.

I used the preexisting getPreviousBlock/getNextBlock APIs as examples, which needed only minor adjustments to work for inlines.

Fixes: #1794

### DIFF
--- a/packages/slate/src/models/node.js
+++ b/packages/slate/src/models/node.js
@@ -1226,7 +1226,7 @@ class Node {
   }
 
   /**
-   * Get the block node before a descendant text node by `key`.
+   * Get the block node after a descendant node by `key`.
    *
    * @param {String} key
    * @return {Node|Null}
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Hey @davidchang I think this only works when inline happen to be right next to each other? I guess I'm not really sure what the ideal logic would be... but it seems like it should work across the entire document, across blocks. Which makes it a bit harder to write I think.

davidchang:
ahh, eek - yeah. yeah, i'm not sure i really know how to tackle this, then, without doing some sort of reverse depth first search. i might give it some time to think about. going to close for now, will reopen if i make any progress.

Issue #1841, opened 2018-05-15
renderEditor only works in plugin, not in Editor


klis87:
#### Do you want to request a _feature_ or report a _bug_?
Potential bug or not update doc

#### What's the current behavior?
`renderEditor` doesn't work when defined as `Editor` prop, only as plugin.
See https://jsfiddle.net/fj9dvhom/352/

#### What's the expected behavior?
`renderEditor` should be added to implicit `editor` plugin and called, currently it is ignored

luolonghao:
these tags of Toolbar should be inserted before Editor.

```
render() {
    return (
    	<div>
        <div>Toolbar</div>
        <Editor
          placeholder="Enter some text"
          onChange={change => this.onChange(change)}
          value={this.state.value}
        />
  	  </div>
    )
  }
```

klis87:
I could, but then I wont have access to editor related props, and anyway according to docs using `renderEditor` should be possible.

PR #1899, opened 2018-06-14
[slate-react] Pass renderEditor prop through into editor plugin

davidchang:

Fixing a bug

#### What's the new behavior?

Previously, the `renderEditor` prop was only applied when defined on a plugin, not when it was passed as a prop to the top-level Slate Editor component. Now, the `renderEditor` prop will be invoked.

There was just a whitelist of plugin props that gets applied to the Editor component, `renderEditor` needed to be on the list

Fixes: #1841


ianstormtaylor:
Good save, thanks @davidchang!

Issue #897, opened 2017-06-21
Can't paste into editor in IE 11


jwoldan:
A bug exists which prevents pasting text into the editor on IE11.  When attempting, nothing is pasted and an error occurs in the `getTransferData` method.

 #### To Reproduce:
1. Copy text from anywhere outside the editor
2. Click in the editor and paste, either using a keyboard shortcut, or by right-clicking and selecting "Paste"

#### Expected behavior: 
The copied text is pasted into the editor

#### Actual behavior:
Nothing is pasted and a console error occurs.

ianstormtaylor:
Hey @jwoldan thanks, I think this is related to this PR: https://github.com/ianstormtaylor/slate/pull/882

jwoldan:
@ianstormtaylor Thanks, looks like it!

bendulum:
Pasting text seems to work in IE11 (Version 11.1176.10586.0), Chrome (Version 62.0.3202.94), and Firefox (Version 56.0) on Windows 10 when removing the lines:

```
  if (event.nativeEvent) {
    event = event.nativeEvent
  }
```
from `get-event-transfer.js`.

DylanPaulusSEL:
Getting this error, and came to the same conclusion as @bendulum. Is `event.nativeEvent` something we need to be checking for? I'll be willing to work on this bug if no one else is currently!

Let me know if this seems to be on the right path. I see two solutions:

1. Remove the whole if-statement if checking for `event.nativeEvent` isn't needed.
2. Using `IS_IE` from `slate-dev-environment`, change the if-statement to ignore IE instances.

PR #1901, opened 2018-06-14
Fix issue with pasting in IE11

DylanPaulusSEL:

Fixing a bug.

#### What's the new behavior?

Fixing the issue where dataTransfer or clipboardData are not populated in IE 11 under event.NativeEvent. Instead, if we are using IE11, use the base event object which contains a valid dataTransfer/clipboardData object.

Adding to the `event.nativeEvent` check in `getEventTransfer.js` to prevent reassigning `event` with `nativeEvent` in IE11.

Fixes: #897


ianstormtaylor:
Thanks @DylanPaulusSEL!

PR #1902, opened 2018-06-15
add hyperprint to test diffs

ianstormtaylor:

Improvement.

#### What's the new behavior?

Uses `slate-hyperprint` to make the diff for test failures easier to read.

Waiting on https://github.com/GitbookIO/slate-hyperprint/issues/3 for selection support, since a bunch of the tests need selections to be serialized.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "rollup-plugin-sourcemaps": "^0.4.2",
     "rollup-plugin-uglify": "^3.0.0",
     "slate-collapse-on-escape": "^0.6.0",
+    "slate-hyperprint": "^2.2.4",
     "slate-soft-break": "^0.6.0",
     "source-map-loader": "^0.2.3",
     "source-map-support": "^0.4.0",
--- a/packages/slate/test/changes/index.js
+++ b/packages/slate/test/changes/index.js
@@ -1,7 +1,7 @@
-import assert from 'assert'
 import fs from 'fs-promise' // eslint-disable-line import/no-extraneous-dependencies
 import toCamel from 'to-camel-case' // eslint-disable-line import/no-extraneous-dependencies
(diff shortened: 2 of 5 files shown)


zarv1k:
Hey @ianstormtaylor, I've already [implemented](https://github.com/GitbookIO/slate-hyperprint/pull/17) selection+decorations support for slate@^0.40. But seems my PR won't be merged, because you know that guys from Gitbook forked Slate at ^0.34 so they aren't going to update to ^0.40 at all.

Anyway hyperscript+hyperprint would be really useful for tests. I'm using it in my project and found it very useful for dumping Slate Values in hyperscript jsx string and immediately use it in my new test. Diffing in jsx format also much more understandable than in JSON.

I think it would be great if the fork of hyperprint package is supported by Slate community with the latest hyperscript features. What do you think?

Thank you.

ianstormtaylor:
Hey @zarv1k, I would love to have it as part of core actually. I was thinking about it, and I think it might make sense to have the `hyperscript` package expose the printer as well. That way you can do:

```js
const h = createHyperScript({ ... })`
const string = h.print(...)
```

And both would use the same set of shorthands. What do you think? Would you be down to add a PR that adds it to core? (I realize it would be a lot, because it would require removing the Flow logic from that other PR.)

Side note, do you have a screenshot of what the test diffs look like? I'm just curious, because it sounds really nice.

ianstormtaylor:
Now that we have JSON support, I think this is not necessary anymore.

zarv1k:
@ianstormtaylor , just FYI, there is my fork of hyperprint, which is compatible with 0.47.x - [@zarv1k/slate-hyperprint](https://www.npmjs.com/package/@zarv1k/slate-hyperprint)

PR #1904, opened 2018-06-15
fix decoration rendering

isubasti:
fixing a bug

#### Current behaviour?
only last decorations of leaves are getting applied
broken-decoration mov

#### What's the new behavior?
All decorations of leaves are getting applied properly
fix-decoration mov

Previously `getLeaves` of text only applied the last decoration of each leaves as on every single loop it reapplied decorations to the original leaves instead of the updated leaves

Fixes: #
Reviewers: @zhujinxuan @ianstormtaylor


zhujinxuan:
That is my mistake.  Thank you for figuring it out.

ianstormtaylor:
Thank you @isubasti!

Issue #1908, opened 2018-06-16
introduce a `Point` model


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now ranges have properties that represent their two points:

```
anchorKey
anchorOffset
focusKey
focusOffset
```

This is alright, but this idea of a "point" is only inferred. There are already places in the codebase that refer to "points" informally, like [in the `findPoint` utility](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/find-point.js#L81-L84). And we've [thought about](https://github.com/ianstormtaylor/slate/pull/1582#discussion_r195614042) adding even more of them.

#### What's the new behavior?

We'd introduce a new model called a `Point`:

```js
Point {
  key
  offset
}
```

And ranges would just have two points, `range.anchor` and `range.focus`.

This would standardize the data structure, making it easier to use in some cases since we can provide the utility functions on the `Point` model directly, and make it more of a known entity for userland code to build with.

zhujinxuan:
@ianstormtaylor  Each time we use a point is used temporally.  If we make a point a class with immutable-js, then it will slows down the creation of Point. (though it will not be the bottleneck of performance)

Because we do not have functions for `Point`, perhaps we can make point only a plain object as `{offset, key}`, implementing `getAnchorPoint`, `getFocusPoint` into Range, `getPointByOffset` into node, and do not cache point functions.  Do you think that a good idea?

ianstormtaylor:
@zhujinxuan I think I’d like to see it actually become a performance issue before making that change, since having the data models be mixed between immutable and mutable is not ideal.

zhujinxuan:
@ianstormtaylor Great. I will add a Point model after #1896 , to ensure newly added the Point model is tested.

ianstormtaylor:
Hey @zhujinxuan thanks for trying that. It looks like #1912 doesn't actually implement the `Point` model in the core use cases of Slate, and instead just provides it as an add-on. That's not really what I was after.

Instead, the `Range` object should implement it's own `anchor/focus` with the point:

```
Range {
  anchor: Point,
  focus: Point,
  isFocused: Boolean,
  isBackward: Boolean,
}
```
```
Point {
  key: String,
  offset: Number,
}
```

And then accessing it would be done via: `range.anchor.key` instead of the current `range.anchorKey`, etc. This way points are used as a core concept all throughout the codebase.

[...]

PR #1912, opened 2018-06-18
[Work in Progress] Provide Point Model

zhujinxuan:

feature
#### What's the new behavior?

See https://github.com/ianstormtaylor/slate/issues/1908

Provides a point model 

Fixes: #1908
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-hyperscript/src/index.js
+++ b/packages/slate-hyperscript/src/index.js
@@ -198,7 +198,7 @@ const CREATORS = {
     }
 
     if (!isEmpty(props)) {
-      selection = selection.merge(props).normalize(document)
+      selection = selection.loadProps(props).normalize(document)
     }
 
     let value = Value.fromJSON({ data, document, selection }, { normalize })
--- a/packages/slate/src/models/range.js
+++ b/packages/slate/src/models/range.js
@@ -229,7 +229,7 @@ class Range extends Record(DEFAULTS) {
    */
 
   get isSet() {
-    return this.anchorPoint.isSet() && this.focusPoint.isSet()
+    return this.anchorPoint.isSet && this.focusPoint.isSet
   }
 
   /**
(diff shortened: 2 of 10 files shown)


zhujinxuan:
Hi @ianstormtaylor , this PR is ready for review.

packages/slate-hyperscript/src/index.js
```diff
@@ -202,7 +202,7 @@ const CREATORS = {
     }
 
     if (!isEmpty(props)) {
-      selection = selection.merge(props).normalize(document)
+      selection = selection.loadProps(props).normalize(document)
```
ianstormtaylor: Why do we call this `loadProps`? Why not just overload `merge` if we need to do custom logic? I assume it's for cases like:

```js
const next = range.merge({ anchorKey: '1' })
```

I think I'd rather overload `merge` for now with deprecation warnings when those key names are used, and prevent updating by those key names in the future.

packages/slate/src/models/range.js
```diff
@@ -129,11 +130,14 @@ class Range extends Record(DEFAULTS) {
       isAtomic = false,
     } = object
 
+    const {
```
ianstormtaylor: Here we need to accept `anchorKey/Offset` style old names, but log a deprecation warning when they are used. And instead move to using the newer `anchor.key/offset` objects.

packages/slate/src/models/range.js
```diff
@@ -227,7 +229,7 @@ class Range extends Record(DEFAULTS) {
    */
 
   get isSet() {
-    return this.anchorKey != null && this.focusKey != null
+    return this.anchorPoint.isSet && this.focusPoint.isSet
```
ianstormtaylor: Can you change these to be `this.anchor/focus` instead of `this.anchorPoint/focusPoint`. 

(I realize this will break `this.focus()` currently, but I will instead deprecate that in another pull request before merging this one, so that we have a migration path.)

packages/slate/src/models/range.js
```diff
@@ -240,6 +242,46 @@ class Range extends Record(DEFAULTS) {
     return !this.isSet
   }
 
+  /**
+   * Get the anchor key.
+   *
+   * @return {String}
+   */
+
+  get anchorKey() {
```
ianstormtaylor: We need to put deprecation warnings here, and encourage people to use `this.anchor.key` instead I think.

packages/slate/src/models/range.js
```diff
@@ -302,8 +361,7 @@ class Range extends Record(DEFAULTS) {
    */
 
   hasAnchorAtEndOf(node) {
```
ianstormtaylor: I think we can deprecate these methods now. Previously they were required in "anchor" and "focus" were owned by the range, but now we can just have people call into the `Point` methods instead. So instead of:

```js
range.hasAnchorAtEndOf(node)
```

People can just as easily do:

```js
range.anchor.isAtEndOf(node)
```

And it keeps us from having to duplicate a lot of logic.

packages/slate/src/models/range.js
```diff
@@ -599,7 +616,8 @@ class Range extends Record(DEFAULTS) {
    */
 
   moveToAnchor() {
-    return this.moveFocusTo(this.anchorKey, this.anchorOffset)
+    const { anchorPoint: focusPoint } = this
```
ianstormtaylor: This is a kind of confusing notation. It would be better to make these just use `focus: this.anchor` instead of the tricky renaming.

zhujinxuan:
@ianstormtaylor Great. 
Could you put a waiting tag on this PR?  I will put another PR about the focus, then we can move back to this one.

zhujinxuan:
@ianstormtaylor Another problem. Our decoration is by Range interface as `anchorKey, anchorOffset`.  Shall we switch decoration to the interface as `anchor, focus`?

zhujinxuan:
@ianstormtaylor  I am thinking of merging https://github.com/ianstormtaylor/slate/pull/1952 before this PR.  Then we can use `focus` and `anchor` directly~

Issue #1916, opened 2018-06-19
Backspace doesn't remove the first character of any inline formatted content


kevinguard:
#### Do you want to request a _feature_ or report a _bug_?

Report a bug.

#### What's the current behavior?

[JSFiddle](https://jsfiddle.net/2jw7ozdn/3/)
To reproduce the bug, click anywhere close to word **rich** and press backspace. The first letter **r** cannot be removed. Trying the same with anything that is formatted inline, same issue can be seen.

#### What's the expected behavior?

On pressing backspace, anywhere, a character should be removed regardless of inline formatting.

PR #1917, opened 2018-06-20
fix removeText with Mark of length 1

isubasti:
fixing a _bug_

#### What's the new behavior?
text with mark of length 1 get removed properly on deletion
text with mark deleted

previously, the `searchLeafAtOffset` got the wrong start point therefore returning previous leaves on leaves with length === 1

Fixes: https://github.com/ianstormtaylor/slate/issues/1916
Reviewers: @zhujinxuan


zhujinxuan:
Great!  Thank you.

ianstormtaylor:
Thanks @isubasti!

Issue #1914, opened 2018-06-19
Pass selectionPosition Prop To Nodes


travigd:
#### Do you want to request a _feature_ or report a _bug_?
Feature

#### What's the current behavior?
Currently, nodes get an `isSelected` prop which is `true` if and only if the selection resides within the node. It seems that the selection for inline nodes is considered within the node if the cursor falls within `[startOfNode, endOfNode)` (i.e. if the cursor is at the right edge of the node, the selection is considered to be at the start of the next node). This makes it impossible for nodes to know if the cursor lives at their right edge (short of always updating in `shouldComponentUpdate` and checking the current value of `editor.value.selection`, which is performance prohibitive for large documents).

#### What's the expected behavior?

Something like this was considered in #1125 (the proposal was making `isSelected` an enum of sorts).

I'd like to add a prop along the same lines named `selectionPosition` (or a different name) that takes on enum values to signify where the selection lies.

```typescript
const enum SelectionPosition {
  LeftEdge,
  RightEdge,
  Within, // includes when selection is expanded but start and/or end is at edge
  Outside, // or pass prop as null
}
```

This would leave the existing behavior of `isSelected` in place.

I'd be happy to create a pull request if you think that this sounds okay. It'd be a few extra lines of code around the [`render` and `renderNode` methods of `Content`](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/content.js#L465), plus `PropTypes` for `Node`.

P.S. Seems like this would only actually be useful for inline nodes since the selection is still contained within a block node when the caret is at the right-most (assuming LTR) position.

ianstormtaylor:
Is it really `[startOfNode, endOfNode)`? I wasn't aware of that. Might be a bug somewhere, or something else going on here.

travigd:
See [this JSFiddle](https://jsfiddle.net/0x7th53y/). I actually notice (in Chrome on Ubuntu) that the cursor is still blue because Chrome considers the cursor to still be inside the span but slate considers the selection to be in the text after the inline.

It seems to me like sensible behavior. The alternative (as far as I see) would be to give two nodes the `isSelected: true` prop which doesn't really make a lot of sense if the selection is collapsed.

Also seems very much related to #1905.

travigd:
Actually, in that example, it seems that the focus is (to abuse interval notation...)
```
[ text ]( inline )[ text ]
```
[...]

PR #1918, opened 2018-06-20
[WIP] Add selectionPosition prop on Node components.

travigd:
Feature

#### What's the new behavior?
Add `selectionPosition` prop to Node component.

[Example Video](https://drive.google.com/file/d/1Oox36l7pepixHBb8TnCzVaS8k4ThQrhC/view) (apologies, no easy GIF video screenshot recorder for Linux that I know of)

Sample Code:
```jsx
class LatexNode extends React.Component {
  render() {
    const {selectionPosition} = this.props;
    if (selectionPosition === "outside") {
      return this.renderKatex();
    } else {
      // selectionPosition is one of leftEdge, rightEdge, or within
      return this.renderEditor();
    }
  }

  renderKatex() {
    // Render Latex via Katex
    const {attributes, children} = this.props;
    return (
       <span {...attributes}>
        <KatexRenderingComponent {...}  contentEditable={false}/>
        <span>{children}</span>
       </span>
    );
  }

  renderEditor() {
    const {attributes, children} = this.props;
    return <code {...attributes}>{children}</code>;
  }
}
```

Fixes: #1914 
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/node.js
+++ b/packages/slate-react/src/components/node.js
@@ -38,6 +38,7 @@ class Node extends React.Component {
     node: SlateTypes.node.isRequired,
     parent: SlateTypes.node.isRequired,
     readOnly: Types.bool.isRequired,
+    selectionPosition: Types.string,
   }
 
   /**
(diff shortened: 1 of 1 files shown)


travigd:
Questions:
* Is the way I've done this (with several `isSelectionAt{Start|End}Of`'s) the _best_ way to do that?
* The string literals (`"leftEdge"`, `"within"`, etc.) should be constants. Where? In `constants/selection-position.js`? They should also _probably_ be renamed to use start and end rather than left and right for consistency with the rest of Slate.
* Tests?

ianstormtaylor:
Closing in favor of the discussion in #1914.

PR #1921, opened 2018-06-20
Fix typo in Docs

RXminuS:
the change() function doesn't exist on a change and given the context setValue() is probably the function that was meant.


ianstormtaylor:
@RXminuS ah good call!

Is Donna using Slate for the editor in the product? Just curious, if so if I could add it to: https://github.com/ianstormtaylor/slate/blob/master/docs/general/resources.md#products

PR #1927, opened 2018-06-22
fix(Safari): Set a text value in dataTransfer to have the cursor on Safari when dragging a text node

qgerome:
bug

#### What's the new behavior?
While dragging some text or a node, it displays the caret where the drop will happen.

#### How to test it ?
1. Go to the rich text example
2. Select words/sentences and move your selection somewhere else in the same slate editor
3. While dragging it, you'll see the "drop point" thanks to a caret

I suppose that Safari does not like to have a value for _application/x-slate-fragment_ in _dataTransfer_ and no _text_ value

Fixes: no issues found


ianstormtaylor:
Wow, great catch @qgerome, thank you very much!

PR #1928, opened 2018-06-22
Fix a linter bug in a semi-colon case

zhujinxuan:

bug

#### What's the new behavior?
Previously we have to disable padding-lines rule when
```
if (...) {
 ..
}

;[].forEach((x) => {
...
})
```
becuase `;` is considered an empty block.  This PR fixes that problem

Fixes: #
Reviewers: @


ianstormtaylor:
Thanks @zhujinxuan!

Issue #1871, opened 2018-06-01
Drag and drop does not remove original text/block


CameronAckermanSEL:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

When a selected fragment is dragged then dropped into another location of the editor, the original selected fragment is not removed.

http://recordit.co/6NMojTC5Cp

 The current behavior results in duplication of text, and in some cases the extra blocks left behind can cause problems if there are any unique identifier keys stored in the data property.

#### What's the expected behavior?

The original fragment is removed.

CameronAckermanSEL:
Can currently be suppressed with a plugin that does

```
{
  onDrop: (event, change) => {
    return change;
  },
},
```

gggdomi:
```
if (isDraggingInternally) {
  change.delete()
}
```
 in `slate-react/src/plugins/after.js` line 234 must be the culprit, coupled with the `isDraggingInternally = false` in `onDragOver` (line 167) that disable deletion

PR #1929, opened 2018-06-22
fix(DragAndDrop): isDraggingInternally should not be reset by the onDragOver handler

qgerome:
bug

#### What's the new behavior?
Drag and drop works again.

This bug exists since : [Removed early return in onDragOver](https://github.com/ianstormtaylor/slate/commit/4712e64c308e68895f2601b5572b367aec2c4001). This commit removed the _return true_ that was stopping the call of `onDragOver` for the plugins stack. Since now the after plugin `onDragOver` is called, the internal dragging flag (`isInternallyDragging`) is reset to null which prevents the `onDrop` handler to remove the selection when it's an internal drag operation.

Fixes: #1871

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -163,8 +163,6 @@ function AfterPlugin() {
 
   function onDragOver(event, change, editor) {
     debug('onDragOver', { event })
-
-    isDraggingInternally = false
   }
 
   /**


ianstormtaylor:
Good call, thanks @qgerome!

CameronAckermanSEL:
Not all heroes wear capes

PR #1933, opened 2018-06-23
Added slate-instant-replace to list of plugins.

enzoferey:
Added a plugin I made to the plugins list.

#### What's the new behavior?
New entry in the table.

--

Fixes: #
Reviewers: @


enzoferey:
Not sure why linting test failed... No insightful error to know what to fix.

davidchang:
not directly related, but i'd love to see this doc point to a search page on npm or yarn (search would be something like "slate" AND "plugins"), rather than package authors having to manually add their plugin to this page each time.

enzoferey:
@zhujinxuan I'm doing it in the next few days ;)

PR #1935, opened 2018-06-25
RFC: Fix for Android devices

### LINKED ISSUE #725: Input is wonky on Android devices
I just did some testing on Chrome on a Samsung Galaxy J3, and the input was pretty wonky (iPhone was ok). It's hard to nail down exactly what's happening, but I see a couple things:

- Often the first space press while at the end of a word doesn't add a space, instead selecting a default autocorrect
- Sometimes words are duplicated or appear and disappear while typing

This project looks great so far, and I'd love to use it, but I need Android support. I don't know anything about the nuances of input on contenteditable and/or Android, but if you have a rough direction/hypothesis for why this may be happening and don't have time to fix it now, I can dig in more and see what I can do.

Thanks!

### LINKED ISSUE #1857: Chrome android: when long pressed delete error
On chrome on android device.

Open slatejs.org, or another editor which use it, like: https://framesia.com/try-editor

try to type some text then delete it with long press delete. until `text.length === 0`

The error happens in each node with the same message:
```
Error: Unable to find a DOM Node for "N".
This often because of forgetting to add `props.attributes` to a custom component
```

mattkrick:

Fixing a bug

#### What's the new behavior?

It works

Proof: https://photos.google.com/share/AF1QipO2uaYQFgq8PuYPq2ojGLokBHhqPaLoFQif8fQFyq220giwtcfz6x1FFdl8HWp7XQ/photo/AF1QipMrSTxFBM5GeXc98E_2gh7p9SKveXRmphrDQfJ1?key=Q19FaENxc01ka3VKZnZINXgzNGxfZThhSG1XYXhn

Device is an OG Pixel, using GBoard on Android 8.1.0. YMMV. PLEASE TEST ON YOUR DEVICES!

1. It doesn't hijack events. it just lets android do its thing
2. it debounces updates to slate so the IME auto-suggest isn't constantly refreshing when the selection gets reset every render. Without a debounce, you can't type faster than ~3 keys per second and holding the backspace doesn't work. 

NOTE: 
- I use snapshotting, so react version is pinned at 16.3, making this a breaking change. you can use cWU if you want, but the package will need to be made react 17-ready soon anyways...
- Before merging, please consider using snapshotting to replace `updateSelection` for all devices, not just android.

fix #725 fix #1857 

Reviewers: @Slapbox @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -129,8 +129,32 @@ class Content extends React.Component {
    * On update, update the selection.
    */
 
-  componentDidUpdate = () => {
-    this.updateSelection()
+  getSnapshotBeforeUpdate() {
+    // Android changes are debounced and the range needs to be cached
+    if (!IS_ANDROID) return null
+    const selection = window.getSelection()
+    if (selection.rangeCount !== 1) return null
+    const range = selection.getRangeAt(0)
+    return {
+      startOffset: range.startOffset,
+      endOffset: range.endOffset,
+      startContainer: range.startContainer,
+      endContainer: range.endContainer,
+    }
+  }
+
+  componentDidUpdate = (prevProps, prevState, snapshot) => {
+    if (snapshot) {
+      const selection = window.getSelection()
+      if (selection.rangeCount === 1) {
+        const range = selection.getRangeAt(0)
+        const { startContainer, endContainer, startOffset, endOffset } = snapshot
+        range.setStart(startContainer, startOffset)
+        range.setEnd(endContainer, endOffset)
+      }
+    } else {
+      this.updateSelection()
+    }
   }
 
   /**
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -146,9 +146,15 @@ class Content extends React.Component {
   componentDidUpdate = (prevProps, prevState, snapshot) => {
     if (snapshot) {
       const selection = window.getSelection()
+
       if (selection.rangeCount === 1) {
         const range = selection.getRangeAt(0)
-        const { startContainer, endContainer, startOffset, endOffset } = snapshot
+        const {
+          startContainer,
+          endContainer,
+          startOffset,
+          endOffset,
+        } = snapshot
         range.setStart(startContainer, startOffset)
         range.setEnd(endContainer, endOffset)
       }
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -1,7 +1,7 @@
 import Base64 from 'slate-base64-serializer'
 import Debug from 'debug'
 import Plain from 'slate-plain-serializer'
-import { IS_IOS } from 'slate-dev-environment'
+import { IS_IOS, IS_ANDROID } from 'slate-dev-environment'
 import React from 'react'
 import getWindow from 'get-window'
 import { Block, Inline, Text } from 'slate'
@@ -25,6 +25,7 @@ import setEventTransfer from '../utils/set-event-transfer'
  */
 
 const debug = Debug('slate:after')
+const ANDROID_KEYSTROKE_DEBOUNCE = 300
 
 /**
  * The after plugin.
@@ -34,7 +35,7 @@ const debug = Debug('slate:after')
 
 function AfterPlugin() {
   let isDraggingInternally = null
-
+  let stagedChange = null
   /**
    * On before input, correct any browser inconsistencies.
    *
@@ -45,7 +46,9 @@ function AfterPlugin() {
 
   function onBeforeInput(event, change, editor) {
     debug('onBeforeInput', { event })
-
+    // onBeforeInput is used as a cheaper insert
+    // Android events are debounced and a full diff is required
+    if (IS_ANDROID) return
     event.preventDefault()
     change.insertText(event.data)
   }
@@ -349,7 +352,22 @@ function AfterPlugin() {
       .moveFocusTo(point.key, end)
 
     // Change the current value to have the leaf's text replaced.
-    change.insertTextAtRange(entire, textContent, leaf.marks).select(corrected)
+
+    if (IS_ANDROID) {
+      // Controlled contentEditables lose their selection range every render
+      // and must be updated manually (see Content#componentDidUpdate)
+      // Updating the selection causes the IME auto-suggest to recompute
+      // synchronously, causing lost input events during that recompute
+      // If device turns off IME auto-suggest, this is not needed
+      clearTimeout(stagedChange)
+
+      stagedChange = setTimeout(() => {
+        change.insertTextAtRange(entire, textContent, leaf.marks).select(corrected)
+        editor.onChange(change)
+      }, ANDROID_KEYSTROKE_DEBOUNCE)
+    } else {
+      change.insertTextAtRange(entire, textContent, leaf.marks).select(corrected)
+    }
   }
 
   /**
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -36,6 +36,7 @@ const ANDROID_KEYSTROKE_DEBOUNCE = 300
 function AfterPlugin() {
   let isDraggingInternally = null
   let stagedChange = null
+
   /**
    * On before input, correct any browser inconsistencies.
    *
@@ -362,11 +363,16 @@ function AfterPlugin() {
       clearTimeout(stagedChange)
 
       stagedChange = setTimeout(() => {
-        change.insertTextAtRange(entire, textContent, leaf.marks).select(corrected)
+        change
+          .insertTextAtRange(entire, textContent, leaf.marks)
+          .select(corrected)
+
         editor.onChange(change)
       }, ANDROID_KEYSTROKE_DEBOUNCE)
     } else {
-      change.insertTextAtRange(entire, textContent, leaf.marks).select(corrected)
+      change
+        .insertTextAtRange(entire, textContent, leaf.marks)
+        .select(corrected)
     }
   }
 
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -131,6 +131,8 @@ function BeforePlugin() {
    */
 
   function onCompositionEnd(event, change, editor) {
+    // triggering a forced update here will close the Android IME
+    if (IS_ANDROID) return
     const n = compositionCount
 
     // The `count` check here ensures that if another composition starts
@@ -161,6 +163,8 @@ function BeforePlugin() {
    */
 
   function onCompositionStart(event, change, editor) {
+    // Overriding Android native composition logic results in errors
+    if (IS_ANDROID) return
     isComposing = true
     compositionCount++
 
@@ -370,6 +374,14 @@ function BeforePlugin() {
 
   function onKeyDown(event, change, editor) {
     if (editor.props.readOnly) return true
+
+    // ignore garbage 229 events sent during composition
+    // event.isComposing doesn't exist in react pseudo events
+    // https://github.com/facebook/react/issues/13104
+    if (event.key === 'Unidentified') {
+      event.preventDefault()
+      return true
+    }
 
     // When composing, we need to prevent all hotkeys from executing while
     // typing. However, certain characters also move the selection before


Slapbox:
@mattkrick, this looks very interesting.

I'll give it a test.

Do you know if this fixes the issue with GBoard where the selection can get stuck in one spot no matter where you try to move it? It seems like you covered the bases on IME input, but I've suspected that the selection state getting out of sync is a separate issue.

mattkrick:
@Slapbox I never came across that problem. I'm guessing that would happen if the IME autocorrect changed the length (eg correct "Well" to "We'll") without telling slate about the change. The rabbit hole goes deep if you try to guess all the things the IME can do, which is why I just let it do its thing.

The only downfall of this approach that i've noticed is the autosuggestions will briefly flash whenever you stop typing for 300ms. I think that's good enough to claim android support :wink: 

I also haven't tested it with long docs, so the scroll position may break on re-render. This is intentional because I think `updateSelection` and `scrollToSelection`  are needlessly complex & can be greatly simplified by using `getSnapshotBeforeUpdate`. The goal of this PR is to present the minimal change necessary to get it working, then we can worry about fine tuning & optimization later.

thesunny:
Woe! That's crazy if this works. Thanks for putting the work in.

You may want to verify against a few Android devices. I wrote up this guide on how to simulate different Android devices here: https://thesunny.gitbook.io/slate/v/doc-dev-guides/contributing/android-dev-guide

Slapbox:
This is a huge step in the right direction. There are still numerous bugs, but if this doesn't break any existing code, and it seems it should not, I recommend integrating these changes.

With the changes from @mattkrick the editor becomes a burden to use on Android, as opposed to an outright impossibility. (Android 8.10 with GBoard, Pixel 2, Chrome 67.0.3396.87)

Here's the bugs I saw:

1. The selection resets to the beginning of a _line_ after typing a word and hitting space, or after roughly 3 characters. There may be something more precise to this.
2. Replacing words with identical words of differing capitalization resets the selection to the beginning of the _word_. For example, Changing 'i' to 'I' will bring the cursor to the first spot prior to the word. changing 'aM' to 'am' has the same effect.
3. It is still possible to crash the editor with "Unable to find DOM node for "[some number]"
image
4. Pressing enter in the middle of a word will

## This is a tremendous improvement.

@thesunny @SamyPesse @Soreine @zhujinxuan @AlbertHilb @tobiasandersen @YurkaninRyan @danburzo @oyeanuj 

**Slate + Android/GBoard Issues Addressed:**

1. Using spacebar mangles or erases preceding words - **FIXED**
2. Cursor can get irreversibly stuck in a spot and refuse to leave it - **IMPROVED** (resets to a more predictable location and can be moved manually)
3. Fragments of words and entire words are no longer repeated when using autocorrect, though there is still some weird behavior - **IMPROVED**
4. Unable to find DOM Node errors - **POSSIBLY IMPROVED** (seem to be less frequent)

zhujinxuan:
Please use https://github.com/reactjs/react-lifecycles-compat to ensure support of old react versions.

thesunny:
@mattkrick Do you know how this affects collaboration and undo? @Slapbox since you are doing some testing, that might be something to test as well.

Even if it breaks collaboration and undo, I'd suggest we still merge if it's the difference between no Android and weak Android. We should, however, put a strong caveat on the documentation if that is the case.

Slapbox:
@thesunny I've not worked on collaboration at all so I can't test in that regard.  I can loop back later and test undo/redo though.

@mattkrick regarding context, I think that the context for swiping text is determined differently. I'm going by gut feeling from my usage of it over the years though and not any explicit evidence. I could definitely be wrong. In any event, your code is an enormous step forward and I support merging it.

mattkrick:
@thesunny a change event is fired every 300ms, so the operation is the diff that occurred in that time span. It'd be easy enough to make it per input.

@slapbox check out the video proof, I swipe in it, swipes are different because they don't send a composition end until you finish.

Slapbox:
@mattkrick I'll need to do some more testing. Honestly, I didn't even see your demo link. Talk about selective blindness.

**Regarding the 300ms debounce:** does it seem possible that's related to the selection resetting itself? I'm interested to try it without debouncing and with a much longer debouncing period to try to narrow in on the source of the selection resetting issue.

**Regarding autosuggest context:** I think the issue I'm talking about needs to be tested with some poorly input swipes that leave the correct choice more ambiguous. In your example the suggestions were already close, but my question is more, will it maintain the context of the entered swipe if the swipe can't figure out whether we mean `feather` or `feature` or something like that. Words that are substantially different, but both suggested due to the imprecise nature of swiping. When you click back on a word like that later in normal operation, it has lost that context and makes suggestions based only on the word that's actually on the page, and not at all based on the context of the entry. You're probably right about this working. I'll do some more testing and if there's any issues I'll report them.

thesunny:
@mattkrick, I found an issue.

We get incorrect cursor selection consistently if you do keypresses within that 300ms interval (guessing that that is the cause but not certain).

For example, go in the middle of any line and type one character. Cursor will be fine. Now type two in quick succession (e.g. "oo"). The cursor will be offset one to the right of its usual position. If you type three characters in quick succession, the cursor will be offset two to the right of the original position. You basically always get one additional offset per character.

The same happens for backspaces. Hit it once and it's fine. If you hit it twice, the offset is one too far to the left. etc.

I'm guessing this is related to the debounce? Even if it slows performance, I think the slow performance may be a better trade off than the cursor moving around for fast typers.

If I type slowly, it works perfectly.

This test was done under Android Studio with the "Nexus 5X API 26" device.

thesunny:
Okay, some more testing revealed:

The cursor movement bug I just mentioned appears to be the same as the resetting cursor to the beginning of the line bug mentioned by SlapBox above (his bug #1). It doesn't do if you type slowly.

A fatal bug is if you hit backspace at the beginning of a line (i.e. to merge it into the previous line).

image

thesunny:
More testing reveals that this bug was NOT introduced by your code @mattkrick 

I can confirm that it appears on www.slatejs.org as well. I'm not sure if there is an issue for this yet.

thesunny:
@mattkrick

I tried setting `ANDROID_KEYSTROKE_DEBOUNCE = 0` and all of the selection movement issues disappeared no matter how quickly I typed.

I don't have an actual Android device (I'm using an emulator) so I'm not sure what this does to real life performance but it does solve all the cursor issues. I hope with that, if we fix the backspace issue (which is separate from your code) we may have a decently working Android version.

While we're not out of the woods quite yet, thank you so much for this contribution. So many people have been waiting so long for Android support.

Also, thank you personally. I was planning to spend a week or two in August to see if I can fix this but your solution is so much simpler.

thesunny:
Some more insights:

* I tested on a real world Nexus 6P of a co-workers (about two years old) with a recent version of Android and the performance was good.
* I tested on several virtual devices in Android Studio and the patch works starting from Android 7.1.1 Nougat and the entire demo site fails to load for some reason on Android 6 Marshmallow and earlier.
* There is another bug. If you type a space in the middle of a word, it moves the cursor one to the right of its expected cursor position. It's a livable bug though. This bug happens in Android 8+ but does not appear in Android 7.1.1 which highlights the issue of the Chrome Android API changing unpredictably.

Slapbox:
Great test reporting @thesunny. I added your bug to the list I have above which I'll try to maintain so we have a comprehensive overview of where we stand.

thesunny:
Sorry for this relentless barrage of comments. I'm just trying to add as much value as I can in as short a time span as I can.

So an issue I'm seeing is that setting `ANDROID_KEYSTROKE_DEBOUNCE = 0` does work perfectly when I'm typing on the virtual keyboard (i.e. the keyboard that is on the screen); however, if I type *quickly* using a connected physical keyboard in Android Studio, I get bad input and cursor issues. This probably won't be an issue on a mobile phone with decent performance but it will be an issue for tablets with connected keyboards and possibly slower mobile devices.

Example: I typed "This is a test to see how well it works"
image

This happens both with a debounce of 0 (i.e. next tick) and completely removing the debounce code.

Slapbox:
No need to apologize when you're commenting in the name of progress!

@thesunny Have you tried any other debounce values? Does setting something very low to the point of being nearly useless, like 20, do anything to mitigate the physical keyboard issue? Does it bring back the selection issue with software keyboards?

thesunny:
@Slapbox @mattkrick I had tried 10 and it didn't help.

Hooked up a Samsung Tablet. Unfortunately this issue does present itself on a physical tablet with an attached keyboard. It only happens on gboard. Initially we thought it was fine but I realized it was a Samsung device which uses a different keyboard by default. When we switched to using gboard, the problem came back.

My guess is that all of these selection/input error issues are all related to one issue which is that Slate doesn't get a chance to update the DOM through React prior to a new input event coming in that has already updated the DOM on its own. @mattkrick do you think that's what's happening?

I feel like the right solution might be one of two things:

* Tell React not to update the DOM and assume that it is correct after input events (not including backspace, delete, etc).
* Tell React to *only* update the DOM if a new `input` event hasn't come in since we calculated the last update. For example if `input` => `calculate insertion` => `NEW input` => `ignore`. But if there is no `NEW input` then we `input` => `calculate insertion` => `flush insertion to DOM`. This way we put back in the setTimeout as well to improve performance. Note: this does feel tricky though because we have to make sure that no new input events are allowed to update the DOM through its default behavior between the time Slate says to start flushing to the DOM and the time React actually does the flush which I'm guessing is asynchronous. There is a `forceUpdate` function in React but it's unclear to me if it's synchronous https://reactjs.org/docs/react-component.html#forceupdate .

mattkrick:
fantastic QA, thank you!

@Slapbox oh, you're talking about the sloppiness autocorrect! (eg if i swipe "bread" twice the suggestions are distinct, first may be "bad", second may be "bed"). These go away if you hit enter, or blur the field, or change the selection. The only way to prevent these from going away is to delay updating the component, so let's do that. 

@thesunny removing the debounce is a slippery slope. No matter how fast the device is, android WILL drop events that fire while the IME is creating the suggestions. When react rerenders, it will reset the selection range. So just like above, *the goal is update the component as little/late as possible.* (PS forceUpdate is async, it's identical to this.setState({})).



To solve both problems, here are a couple ideas:
1. move the debounce logic from the `after` plugin to `Editor#shouldComponentUpdate`. Then, we can decide there if it should update (ie did a plugin fire to open a menu? did a plugin trigger an auto correct? etc). Using logic that's more sophisticated than "after 300ms has passed" would be nice.

2. Don't call setState so gosh darn always. 90% of the time, the default behavior of the input event will result in the correct state. Heck, `event.preventDefault()` isn't even called in the `After#onInput` handler, which means we're basically saying "Hey DOM, what's the answer? 3? Nope, you're way off. It's 3!". Reducing setState calls will  improve performance like crazy because we won't have to reset the caret & scroll on every input.

I haven't actually used slate yet (i need it to support android before i can use it for work, hence the PR), but i imagine setting a flag like `value.pushUpdate = true` and then sCU would look like `return value.pushUpdate`. that way, any plugin can force an update.

Hope that helps!

Slapbox:
@mattkrick your PR really made my day. Knowing you've not even used Slate blows my mind.
I encourage you to join us on the Slack channel https://slate-js.slack.com/

thesunny:
@mattkrick 

1. That's a great idea to move the debounce logic into `#shouldComponentUpdate`! It of course totally makes sense that it would be called immediately before an actual update.

2. I think somebody said that Slate used to be selective about when to update the DOM but after doing some performance testing, they removed the selectivity because it simplified the code and it performed well enough. Of course, that was before this Android issue.

mattkrick:
Added #1938 to discuss selective rendering. I probably can't get more done this week.
@thesunny If there's a "Good first PR" label you could attach to it, or some other way to attract someone to PR it soon-ish, that'd be amazing!

Slapbox:
I reviewed the requirements for `react-lifecycles-compat` and it looks like there should be no problem with dropping that in when matt has a chance to do it or someone has a chance to PR it.

thesunny:
@mattkrick I think we can pull the PR but I think we should do it without the debounce code. Do you mind pulling it out? With the debounce code out, it seems to work fine for me in all Android browsers that I tested as long as I don't attach a hardware keyboard.

It may be slower on some devices (though it worked fine with an older Nexus 6P) but with my testing it still works noticeably better. With the debounce code in, I get issues in everything I tested.

Slapbox:
If someone hosts this version of Slate examples on a public URL I can test with a Samsung Galaxy S4 and an HTC 10. Getting remote debugging working for multiple (especially older) devices is just too much of a pain.

renchap:
@Slapbox it is done automatically on PRs. Check the Netlify link in checks below: https://deploy-preview-1935--slatejs.netlify.com/

Slapbox:
@renchap truly the future is now! Thanks for the tip, I'll try to find some time to test it in the next day or so.

thesunny:
Just a heads up that I'm heading on vacation starting tomorrow for a little over a week.

If any other maintainers are checking in, the pr looks good to me with two changes:

* The debounce removed or reduced to `0` (i.e. next tick) for now.
* Adding the `react-lifecycles-compat` polyfill to maintain compatibility with older versions of React

@mattkrick I wonder if you aren't seeing the debounce issue on your device since I haven't seen you confirm this issue.

thesunny:
Before I go (on vacation), just wanted to say thanks @mattkrick for this amazing work!

Can't believe you aren't even using Slate yet and you have made the most progress on this issue to date. I'm really looking forward to having Android support. My colleagues here are always asking me when Android support is coming. This is the best news on this front in a long time.

Also to @Slapbox for staying on top of everything, putting a lot of work in and giving a lot of encouragement (more important than I think many realize).

Feels like we are getting close to Android support.

mattkrick:
hey, sorry for the delay.
on my device, this PR without the debounce is almost unusable:
- if i press a key before the IME autosuggest has updated (~200ms), the keypress isn't registered.
- i can't press and hold backspace

to reproduce, make sure you turn on every IME "Text Correction" feature, as each one will make it incrementally worse (personalize suggestions, auto-correct, autospace, auto-capitalization, etc). There may be a delta between the emulator and a physical device.
 
I think we both agree the debounce is a hack & needs to go away, but until a PR to fix #1938 lands, it makes the experience darn near perfect for my device.

I'd like to experience the bugs you're seeing with the debounce. i can only reproduce the one where you press & hold backspace until it hits the beginning, can you please provide some detailed repro steps for the others? we can then those into issues & use em as a burndown.

thesunny:
@mattkrick That's interesting because it's the reverse for me on the devices and the Android simulator (multiple versions of Android) in which I tested. It's near perfect without the debounce and almost unusable with the debounce on. If I type before the debounce fires, there are problems with cursor position. I don't have IME on.

Can you tell me what mobile device and what version of Android you're using? When I get back, I can do some more testing to find out when debounce helps and when it doesn't. It might be simply when IME is on. It's possible we can dynamically set the debounce on/off depending on when it causes problems versus when it solves them.

I don't think it's a good idea to accept the PR as is since it seems to be not working for the majority use case.

Either way though this is promising. As long as we can identify the working/not-working use case, we can at least toggle the behavior.

I can take a closer look when I get back. I'm typing this on vacation (couldn't help looking in).

zaaack:
@mattkrick Why is this closed? This PR looks really promising.

Slapbox:
I personally think it is worth accepting this PR in spite of its shortcomings, as it is a DRAMATIC improvement over the current implementation and gives us a foundation to build on. Unfortunately many people feel it is not worth integrating though because it has remaining issues.

I personally think we should be accepting this and merely not advertising any form of Android support. **I think we would be more likely to get PRs to fix the remaining issues with this PR if we accept it as is. It breaks nothing. It only improves the current functionality.**

mattkrick:
I'll post a better PR in a week or 2 based on #1938.
I guess i didn't explain it clear enough in that issue, so I decided to just built it myself. I spent a couple hours to prototype it this weekend & it works as intended. 300% faster input (on all platforms), full mobile support without IME lag, etc. lot of low hanging fruit.

i've slated (ha) to start using it at work in a month, so i'll have more time then.

Slapbox:
@mattkrick, you're blowing my mind.

Slapbox:
@mattkrick any update on this? I'm super intrigued in what you're working on.

mattkrick:
i'm no longer using slate.
here's the branch i was working on if you want to take it over:  https://github.com/mattkrick/slate/tree/slate-fast

oyeanuj:
@mattkrick Bummer! Is there a summary anywhere of your findings and proposed solutions, for others to take it on?

thesunny:
Will be working on this the week of August 13-17 and created a separate issue to track progress:
https://github.com/ianstormtaylor/slate/issues/2047

mattkrick:
@oyeanuj Here's everything you'll need to fix:

- zero-width spaces need to be handled in a DOM-native fashion by using `br`. 
- the `before` plugin needs to go away because arbitrary event canceling is bad
- the `content` component shouldn't handle iOS events because spaghetti is bad
- the `after` plugin needs to handle events based on the InputEvents level supported instead of one-off browser/device flags. There are 3 levels: no support, level 1, level 2: https://w3c.github.io/input-events/index.html#dom-inputevent-inputtype. As you sunset browser support, sunset the fallbacks.
- Avoid canceling events & use DOM-native behavior if text is being added/removed
- Avoid re-renders unless non-DOM-native events occur (see PR for a naive implementation, you'll need to move this down 1 component to make it work with per-keystroke plugins)

Doing this will add support for all browsers on android in addition to full IME support.
Cheers

PR #1942, opened 2018-06-28
Added missing paste event handler in slate-simulator docs

enzoferey:
Acording to https://github.com/ianstormtaylor/slate/blob/master/packages/slate-simulator/src/index.js onPaste event is suported by slate-simulator. Tested locally and the onPaste function of plugins gets called if simulator.paste() is called.

### DIFF
--- a/docs/reference/slate-simulator/index.md
+++ b/docs/reference/slate-simulator/index.md
@@ -83,6 +83,12 @@ Simulator a `focus` event with an `event` object.
 
 Simulator a `keyDown` event with an `event` object.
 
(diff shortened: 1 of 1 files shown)


enzoferey:
I will improve this PR in the next few days with more slate-simulator examples. The entry barrier right now is pretty high in my opinion.

Slapbox:
@enzoferey that would be excellent. I spent most of a day trying to figure that thing out to no avail.

ianstormtaylor:
Thanks @enzoferey!

Issue #1941, opened 2018-06-28
CloneFragment throws errors on IE11


Dtphelan1:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Copying plain text in IE causes multiple errors, rooted in code found in `slate-react/utils/cloneFragment` that is not cross-browser compatible. Specifically: 
1.  Line 36: `contents.childNodes.forEach` and Line 62: `contents.querySelectorAll(ZERO_WIDTH_SELECTOR)).forEach` produce the error `SCRIPT438: Object doesn't support property or method 'forEach'`.  This is because NodeList.forEach is [not supported on any versions of IE](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach#Browser_Compatibility).  
2. Line 93: ` event.clipboardData.setData(TEXT, div.textContent)` produces the error `SCRIPT65535: Unexpected call to method or property access.`. This is because, while setData is defined for IE, it's only valid arguments are 'Text' and 'Url'; we pass 'text/plain'. 

Neither the JSFiddle nor the example pages loaded for me on IE11, so I can't include a link to that, but I imagine any rudimentary fiddle implementing the editor would expose these bugs, as I've traced them back to the `cloneFragement` function present in slate-react.
 
Information re: the system that reproduces the bug: 
- OS: Windows 10, version 10.0.16299 
- Browser: IE11
- Slate: 0.34.0

#### What's the expected behavior?
Basic copying of the text.

#### Possible Solutions:
1. Wrap any NodeLists in the `Array.from()` function to ensure that all these objects are arrays, which do have a [forEach function defined in IE>=9](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#Browser_compatibility). *Concern:* this may have some performance impacts, but I'm not familiar with how resource intensive this operation is.
2. On Line 91 we have a condition check for deciding whether or not we should use the `setData` function. Since this function throws an error on IE, we should change the condition to ensure the browser isn't IE. That is, add this import to the top of the file: `import { IS_IE } from 'slate-dev-environment'` and replace Line 91 with `if (event.clipboardData && event.clipboardData.setData && !IS_IE) {`.

ianstormtaylor:
[...]

PR #1943, opened 2018-06-28
Initial fixes for IE compat of cloneFragment

Dtphelan1:

Fixing a bug

#### What's the new behavior?

Allows basic copy and pasting on IE11.

Change is pretty self-explanatory. Avoids using functions not defined by IE11 or lower by a) converting a NodeList object to an Array, and b) by checking for browser compatibility before using setData functions with arguments that may throw errors.

Fixes: #1941 
Reviewers: No one in particular


packages/slate-react/src/utils/clone-fragment.js
```diff
@@ -88,7 +89,7 @@ function cloneFragment(event, value, fragment = value.fragment) {
 
   // For browsers supporting it, we set the clipboard registers manually,
   // since the result is more predictable.
-  if (event.clipboardData && event.clipboardData.setData) {
+  if (event.clipboardData && event.clipboardData.setData && !IS_IE) {
```
tobiasandersen: Does this mean IE has those methods, but they don't work?

Dtphelan1: It supports the method, but in a very restricted sense. Specifically, IE doesn't support arbitrary MIME types or even common ones like 'text/plain'; it only accepts the strings "Text" (which gets mapped to 'text/plain') and "Url" (mapped to 'text/url-list') as types for calls to `DataTransfer.setData()`. That's why I suggest we skip the following block if we are in IE.

Dtphelan1:
I'm having trouble getting the linter to pass (n.b. I'm on a Windows machine). When I run `yarn lint` on my end I get a litany of errors across files I haven't touched; that said, the build seems to only fail on the code I've added, and it seems prettier is indicating the error is with line endings (which would make sense). Could someone with experience on this front offer some help?

packages/slate-react/src/utils/clone-fragment.js
ianstormtaylor: Sounds like this one can't be polyfilled, so I'd be down to accept it alone. Although can you add a `COMPAT:` comment (check other parts of codebase to see one) to explain why. Thanks!

Dtphelan1:
@ianstormtaylor Would you take a look now and let me know if there is anything else you'd like me to include?

ianstormtaylor:
Thanks @Dtphelan1.

PR #1944, opened 2018-06-30
Clarified the behavior of Paste-HTML Example

Slapbox:
Added a quick note to the bottom with explicit directions on how to use the example.

### DIFF
--- a/examples/paste-html/value.json
+++ b/examples/paste-html/value.json
@@ -62,7 +62,7 @@
             "kind": "text",
             "leaves": [
               {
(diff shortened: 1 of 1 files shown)


Slapbox:
Can someone explain to me what's wrong here?  I''m not sure why the checks fail here.

I had a stray comma in the first commit in this PR, which is corrected by the second commit.

Slapbox:
I have no idea why this PR failed to pass the checks. In any event, it's now resolved by #2344

PR #1950, opened 2018-07-03
Add prop "isFocused" / "isSelected" for custom nodes

SamyPesse:
This PR is a proposal for a breaking change. It changes the definition of the prop `isSelected` for custom nodes, and add a prop `isFocused`.

Basically `isFocused` is equal to the previous `isSelected`.

- **isSelected**: true when the node is in the selection (focused or blurred)
- **isFocused**: true when the node is in the focused selection

In my opinion, these names better match the other concept in slate: something can be selected without being focused.

### Upgrade path

Developers depending on `isSelected` will have to rename it to `isFocused`. It should be easy (Search & replace).

### TODOs

- [x] Add `isSelected` and `isFocused`
- [x] Document it
- [ ] Write some rendering unit tests in `slate-react`


ianstormtaylor:
Hey @SamyPesse this looks good to me. Can you update the examples that use `isSelected` so people don't get confused? Thanks!

examples/embeds/video.js
```diff
@@ -37,10 +37,12 @@ class Video extends React.Component {
    */
 
   render() {
+    const { isSelected } = this.props
+
     return (
       <div {...this.props.attributes}>
         {this.renderVideo()}
-        {this.renderInput()}
+        {isSelected ? this.renderInput() : null}
```
SamyPesse: I've adapted this example to show the input only when the node is selected.

ianstormtaylor:
Thanks @SamyPesse, this looks great!

PR #1952, opened 2018-07-05
Move `isFocused` from Range to Value

zhujinxuan:

feature, preparing for https://github.com/ianstormtaylor/slate/pull/1912

This change will move `isFocus` property, and `focus` and `blur` properties from Range to Value. 

Fixes: #
Reviewers: @


zhujinxuan:
This PR is ready for review~

packages/slate-hyperscript/test/selections/blurred-block.js
```diff
@@ -0,0 +1,51 @@
+/** @jsx h */
+
+import h from '../..'
+
+export const input = (
+  <value isFocused={false}>
```
ianstormtaylor: I think in keeping with JSX standards, the `isFocused` prop in hyperscript should be renamed to just `focused` instead.

packages/slate-hyperscript/src/DecoratorPoint.js
```diff
@@ -0,0 +1,44 @@
+import { Range } from 'slate'
```
ianstormtaylor: Can you rename this file to not use uppercase characters? `decorator-point` instead.

packages/slate-hyperscript/src/createChildren.js
```diff
@@ -0,0 +1,122 @@
+import { Node, Text } from 'slate'
```
ianstormtaylor: Same with this file, it should be `create-children`

zhujinxuan:
@ianstormtaylor The advantage of moving `isFocused` from `Range` to `Value` is that we can have `focus` property in the `Range` (because Range does not handle focus anyone).  Therefore, we can use `anchor` and `focus` instead of `anchorPoint` and `focusPoint` in the Point model.

Another reason is for the consistency of DOMRange.  The Range of DOM does not have the focus feature, the focus is handled by DOM themselves.  Therefore, I think perhaps we shall move focus to the Value.

packages/slate-hyperscript/test/selections/blurred-block.js
zhujinxuan: Fixed~

zhujinxuan:
For the disadvantage, perhaps it disables range.focus() and range.isFocused API.  But according to my knowledge (my plugins, gitbook plugins), selection.isFocused is not used.

zhujinxuan:
Hi, @ianstormtaylor . Do you have anything else on working related `models`?  If you are working on things unrelated to `models`, I will begin to rebase this PR.

ianstormtaylor:
@zhujinxuan good question, I'm not working on any model-related changes right now, so rebasing would be great, thank you!

ianstormtaylor:
I had another thought...

Right now having `isFocused` and `marks` on the `selection` object itself is nice because it means that `set_selection` operations are contained to all of the selection-changing behavior. 

If `isFocused` was to move to `value`, then you'd need a combination of `set_value` and `set_selection` to detect when a user's selection has changed.

zhujinxuan:
@ianstormtaylor I think you are right.  I think isFocus is better to be stay in selection.  Do you think that it is good to close this PR and use `focusPoint` and `anchorPoint` as keys in Range?

ianstormtaylor:
@zhujinxuan I think it makes sense to wait on this one for now, until we come to a clear conclusion that separating `isFocused` is better. Right now I see pros/cons to each side, so I would rather keep it the same until we see stronger pros for changing.

I've just opened #2035 which adds the `Point` model, while refactoring and standardizing a **lot** of the range-based logic to make it less confusing for people. So it might be best to close and revisit this later with that new logic merged.

PR #1953, opened 2018-07-05
More tests for `addMarks` and `Leaf.splitLeaves`

Soreine:
This is just a PR adding more tests to `Text.addMarks` and `Leaf.splitLeaves`.

While upgrading to the latest Slate, I started hunting down a bug, only to realize that my code was accessing a `Text` length with `text.length` instead of `text.text.length`. And using `undefined` was actually working with the old code of `Text.addMark`. The tests here are the results of my hunt. Feel free to close.


ianstormtaylor:
Awesome, thanks @Soreine!

PR #1956, opened 2018-07-06
Allow onContextMenu event handler

RXminuS:
Right clicks aren't handled by onClick in a content-editable component. So we need access to a onContextMenu handler to handle these events properly.

My use case is to be able to open a context menu with actions when the user right-clicks a specific inline node.

Let me know if you think we need tests for this, but I made the change in GitHub since it was so small.

### DIFF
--- a/packages/slate-react/src/constants/event-handlers.js
+++ b/packages/slate-react/src/constants/event-handlers.js
@@ -8,6 +8,7 @@ const EVENT_HANDLERS = [
   'onBeforeInput',
   'onBlur',
   'onClick',
+  'onContextMenu',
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @RXminuS, good idea.

vamshi9:
How can onContextMenu pass event with default params `(props,editor,next)` ?
As in :
```<Editor onContextMenu={(e,props,editor,next)=>this.handleContextMenu(e,props,editor,next)} />
handleContextMenu(e,props,editor,next){
      e.preventDefault();
     //...do some stuff
}```

PR #1957, opened 2018-07-07
plugins reference: update available properties

wmertens:

Feature

#### What's the new behavior?

Doc change only: add some missing information to the Plugins reference.

It is only rudimentary since I don't really know the codebase, but it's better than nothing :)


ianstormtaylor:
Thanks @wmertens, this is great!

PR #1958, opened 2018-07-08
add null-guard for insertTextAtRange when marks is undefined

t3chguy:

fixing a recent bug that slipped in during #1875 

#### What's the new behavior?

No error is thrown.


ianstormtaylor:
Thanks @t3chguy!

PR #1967, opened 2018-07-11
Fix drag-and-drop issues in IE11

justinweiss:

Fixing a bug.

#### What's the new behavior?

You can now drag-and-drop files into an editor that supports it, without console errors or uncaught exceptions.

There were three problems preventing drag-and-drop in IE:

* IE [doesn't consider contentEditables as inputs](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/913982/) when it comes to calling `onDrop`. So you have to `preventDefault` in `onDragOver`, which prevents the cursor from being displayed. Turns out _this is fine_, because:
* You can only get a cursor position from a point with `createTextRange` and `moveToPoint`. Also, `moveToPoint` raises an error if it's called during a dropEvent. So you wouldn't be able to drop into a specific point, even if you wanted to.
* Setting `dropEffect` causes an `unspecified error`.

### DIFF
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -4,6 +4,7 @@ import { findDOMNode } from 'react-dom'
 import Hotkeys from 'slate-hotkeys'
 import {
   IS_FIREFOX,
+  IS_IE,
   IS_IOS,
   IS_ANDROID,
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thank you @justinweiss, this is a quality PR!

Issue #1959, opened 2018-07-09
JSON format for storing


wmertens:
#### Do you want to request a _feature_ or report a _bug_?

feature

#### What's the current behavior?

`value.toJSON()` is a rather elaborate data structure, optimized for editing rather than storage.

There are also no guarantees regarding future compatibility.

#### What's the expected behavior?

Return a data structure that is optimized for storage.

## Some ideas

In previous APIs I made, I noticed that arrays parsed faster than objects. Furthermore, the data is highly predictable. So instead of

```json
{
  "object": "value",
  "document": {
    "object": "document",
    "data": {},
    "nodes": [
      {
        "object": "block",
        "type": "paragraph",
        "isVoid": false,
        "data": {},
        "nodes": [
          {
            "object": "text",
            "leaves": [
              {
                "object": "leaf",
                "text": "Lorem ipsum ",
                "marks": []
              },
              {
                "object": "leaf",
                "text": "dolor sit amet",
                "marks": [
                  {
                    "object": "mark",
                    "type": "bold",
                    "data": {}
                  }
                ]
              },
              {
                "object": "leaf",
                "text": ", consectetuer adipiscing elit.",
                "marks": []
              }
            ]
          }
        ]
      },
      {
        "object": "block",
        "type": "heading-one",
        "isVoid": false,
        "data": {},
        "nodes": [
          {
            "object": "text",
            "leaves": [
              {
                "object": "leaf",
                "text": "Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem.",
                "marks": []
              }
            ]
          }
        ]
      },
      {
        "object": "block",
        "type": "paragraph",
        "isVoid": false,
        "data": {},
        "nodes": [
          {
            "object": "text",
            "leaves": [
              {
                "object": "leaf",
                "text": "Nulla ",
                "marks": [
                  {
                    "object": "mark",
                    "type": "bold",
                    "data": {}
                  }
                ]
              },
              {
                "object": "leaf",
                "text": "consequat ",
                "marks": [
                  {
                    "object": "mark",
                    "type": "bold",
                    "data": {}
                  },
                  {
                    "object": "mark",
                    "type": "italic",
                    "data": {}
                  }
                ]
              },
              {
                "object": "leaf",
                "text": "massa",
                "marks": [
                  {
                    "object": "mark",
                    "type": "bold",
                    "data": {}
                  },
                  {
                    "object": "mark",
                    "type": "italic",
                    "data": {}
[...]

PR #1970, opened 2018-07-13
Add slate-slt1-serializer

wmertens:

feature

#### What's the new behavior?

Provide a serialization format that is hopefully future-proof. See #1959 

It provides a simple, compact, and easily extensible format to store documents. It seems to be 7-10x faster than the html (de)serializer.

The format is explained in [the Readme](https://github.com/wmertens/slate/blob/a082765d64f66367d4968cc81bf8ec324c279de0/packages/slate-slt1-serializer/Readme.md).

* [?] The new code matches the existing patterns and styles.

Fixes: #1959 
Reviewers: @ianstormtaylor 

## Questions

* Currently, the format includes a tag for the document, so that other things from the value can be serialized. Will that ever be useful?
* I noticed that the `<document>` hyperscript tag doesn't accept attributes to pass to `data`. Intended?
* Would it be more useful to serialize to a string using jsurl2? Maybe add a `.toJSON` method on the result that stringifies it?

### DIFF
--- a/packages/slate-slt1-serializer/Changelog.md
+++ b/packages/slate-slt1-serializer/Changelog.md
@@ -1,0 +1,9 @@
+# Changelog
+
+This document maintains a list of changes to the `slate-slt1-serializer` package with each new version. Until `1.0.0` is released, breaking changes will be added as minor version bumps, and smaller changes won't be accounted for since the library is moving quickly.
+
+---
+
+### `0.1.0` — September 17, 2017
+
+:tada:
--- a/packages/slate-slt1-serializer/Readme.md
+++ b/packages/slate-slt1-serializer/Readme.md
@@ -1,0 +1,70 @@
+# slate-slt1-serializer
+
+This package contains the slt1 serializer for Slate documents.
+
+The Slate Data Model is good for editing, but not so good for storing. Furthermore, it might change in the future.
+
+This serializer is meant to provide a stable and compact format for storing Slate documents in a lossless manner.
+
+As an example example, these are equivalent:
+
+slt1:
+
+```json
+["slt1", 1, [2, "line", [["bold", ["italic", { "very": true }], "one"]]]]
+```
+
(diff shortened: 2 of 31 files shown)


DamareYoh:
Im not quite sure what opinion I have on this feature set is, largely because I have my own custom serialization format that I use to save documents. 

However, a lot of this format seems to deal with values being located in very specific indexes, and some “magic number” integers to express node types. At a minimum, for this to have success or robustness, these concepts need to be conveyed with a set of exported constants that indicate
1. The information that goes in a given index
2. The integer used to indicate a type

On the second concern, Im unsure whether the usage of an integer as opposed to having a string to indicate type is truly a meaningful optimization, because without a standard reader library, its very easy to lose the meaning of those numbers by just reading the serialized text.

Just my 2c. Thanks for the contribution

DamareYoh:
(Ah uh Im @CameonAckermanSEL and this is my outside of work account)

ianstormtaylor:
Replied here: https://github.com/ianstormtaylor/slate/issues/1959#issuecomment-404926235

wmertens:
Just a note: I'm working on making (de)serialization easier to write by letting to/fromJSON take functions to transform the data, much like JSON.parse/stringify.

Right now, the plan is that toJSON accepts `transform(node, options)` and `process(json, options)` as options. `transform` results will be used instead of the default if it returns anything, and `process` can access the "json" data (whatever that serialized format is) and change it if required. This even allows implementing caching.

With this approach you could also combine transformation by just wrapping the functions. It's not really suitable for plugins though, because you might not always want the same transforms when calling `toJSON`, and maybe a parent `transform` of one plugin expects different output than what a child `process` of another plugin returns.

I'm leaving this PR open as a reminder until I get that done and then I'll make slt1 into a package.

ianstormtaylor:
@wmertens I'm going to close this out to stay sane.

PR #1972, opened 2018-07-14
Improved slate-simulator docs.

enzoferey:
Following up https://github.com/ianstormtaylor/slate/pull/1942#issuecomment-401518963, I improved the docs adding a DataTransfer mockup class and adding some details about core plugins to make it easier to understand their role for beginners.


ianstormtaylor:
Thanks @enzoferey!

PR #1975, opened 2018-07-17
Upgrade to React v16 lifecycle

zhujinxuan:

feature

#### What's the new behavior?

1. use `memoize-one` instead of state as recommended in reactJS https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization
2. value is handled by and only `props.onChange`, then no state.value is provided.
3. Remove flushChange in componentDidUpdate.  After mount, all update will go to onEvent->change->onChange->`props.onChange`

Fixes: #1426 #1427 
Reviewers: @


packages/slate-react/src/components/editor.js
```diff
@@ -71,100 +72,61 @@ class Editor extends React.Component {
...
     this.flushChange()
+    if (prevProps === this.props) return
+    // Increment the updates counter as a baseline.
+    this.tmp.updates++
+
+    // If we've resolved a few times already, and it's exactly in line with
+    // the updates, then warn the user that they may be doing something wrong.
+    if (this.tmp.resolves > 5 && this.tmp.resolves == this.tmp.updates) {
+      logger.warn(
+        'A Slate <Editor> is re-resolving `props.plugins` or `props.schema` on each update, which leads to poor performance. This is often due to passing in a new `schema` or `plugins` prop with each render by declaring them inline in your render function. Do not do this!'
```
wmertens: how about making that === 5 so you only warn once?

zhujinxuan: This warning only appears in dev mode. We want the warn appears in each update, so the developer cannot miss it.

zhujinxuan:
It is ready for review now~

zhujinxuan:
This PR is ready for review~

packages/slate-react/src/components/content.js
```diff
@@ -468,10 +466,7 @@ class Content extends React.Component {
       return this.renderNode(child, isSelected, childrenDecorations[i])
     })
 
-    const handlers = EVENT_HANDLERS.reduce((obj, handler) => {
-      obj[handler] = this[handler]
-      return obj
-    }, {})
+    const handlers = this.getHandlers()
```
jtadmor: im not sure why the initial implementation is the way it is, but this is actually creating new behavior (new function was being created before on every render, now its not).

are we sure this doesn't subtly change anything?

packages/slate-react/src/components/editor.js
```diff
@@ -228,17 +189,61 @@ class Editor extends React.Component {
...
+  /**
+   * Get stack by plugins, it is a memoized function and therefore do not need
+   * to re-evaluate for the same plugins
+   *
+   * @param {Array<Object>} plugins
+   * @return {Stack}
+   */
+
+  getStackWithMemoization = memoize(plugins => {
+    this.tmp.resolves++
```
jtadmor: should we move `this.tmp.resolves++` into `this.resolvePlugins` just for consistency?

packages/slate-react/src/components/editor.js
```diff
@@ -228,17 +189,61 @@ class Editor extends React.Component {
...
+   *
+   * If the value is produced by onChange with the same stack, then will immediately return
+   * this value.
+   *
+   * @param {Value} value
+   * @return {Schema}
+   */
+
+  processValueOnChange = memoize((value, stack) => {
+    if (this.tmp.change && value === this.tmp.change.value) return value
```
jtadmor: if we are going to memoize this, should we explicitly pass in `this.tmp.change.value` as an argument?

or are there other protections against `this.tmp.change.value` changing between calls to this method while `value` and `stack` stay the same?

packages/slate-react/src/components/editor.js
zhujinxuan: Yes.  Thank you.  Fixed~

packages/slate-react/src/components/editor.js
```diff
@@ -260,12 +265,11 @@ class Editor extends React.Component {
...
     debug('onChange', { change })
-
-    this.stack.run('onChange', change, this)
-    const { value } = change
+    if (this.value === change.value) return
+    const value = this.processValueOnChange(change.value, this.stack)
     const { onChange } = this.props
-    if (value == this.value) return
-    onChange(change)
+    if (value === this.value) return
```
jtadmor: isnt `this.value` just a getter that resolves to the same thing as like `269`?

zhujinxuan: Yes.

packages/slate-react/src/components/editor.js
zhujinxuan: This part is to ensure  `processValueOnChange(processValueOnChange(value)) === processValueOnChange(value)`

The reason is that, in each update, we will have `stack.run('onChange', change)` passed to the parent container, then the parent container will pass `stack.run('onChange', change)` back as `props.value`.  And I do not want to have the `stack.run('onChange', change)` run again when receiving props.

I do not see a neat way to solve the problem.  The only thing I can do is to ensure all value are updates are passed via `processValueOnChange`

zhujinxuan: @jtadmor Hi, I have just removed this part of logic in recent commit.   I agree with you that we shall leave the memoized function pure.

Cheers,

packages/slate-react/src/components/leaf.js
```diff
@@ -60,9 +60,9 @@ class Leaf extends React.Component {
     // If any of the regular properties have changed, re-render.
     if (
       props.index != this.props.index ||
-      props.marks != this.props.marks ||
+      !props.marks.equals(this.props.marks) ||
       props.text != this.props.text ||
-      props.parent != this.props.parent
```
ianstormtaylor: What are these changes for?

packages/slate-react/src/components/editor.js
```diff
@@ -228,17 +163,59 @@ class Editor extends React.Component {
...
 
+  /**
+   * Get stack by plugins, it is a memoized function and therefore do not need
+   * to re-evaluate for the same plugins
+   *
+   * @param {Array<Object>} plugins
+   * @return {Stack}
+   */
+
+  getStackWithMemoization = memoize(plugins => {
```
ianstormtaylor: Can you name these `resolveStack/resolveSchema` to match the existing `resolvePlugins`?

Why not have these call `resolvePlugins` internally themselves instead of needing to take in plugins? (Might be a good reason for it, I'm just quickly reading.)

packages/slate-react/src/components/editor.js
```diff
@@ -260,12 +237,11 @@ class Editor extends React.Component {
...
     debug('onChange', { change })
-
-    this.stack.run('onChange', change, this)
-    const { value } = change
+    if (this.value === change.value) return
+    const value = this.processValueOnChange(change.value, this.stack)
     const { onChange } = this.props
-    if (value == this.value) return
-    onChange(change)
+    this.tmp.value = value
```
ianstormtaylor: Why this?

zhujinxuan: All our values are updated by `this.props.onChange` and `stack.run('onChange')`.  During the mount, we pass the `stack.run('onChange', this.props.value)` back to the onChange, and the the parent container feed use the value after onChange.  

But we do not want to have onChange called again on the value that we returned to the parent container and then passed back to me.  Because calling onChange again and again on the same value will dangerously lead to infinite loop. 

Therefore I use a tmp.value and tell the get value() not to run onChange if it is already processed by onChange(https://github.com/ianstormtaylor/slate/pull/1975/files#diff-0f1805ffc056f7c54e6a6f98ba0259a1R174)

packages/slate-react/src/components/leaf.js
zhujinxuan: Because marks can be the same value with different reference after characters-to-leaves.  Therefore I think it is perhaps better with immutable equals

packages/slate-react/src/components/editor.js
zhujinxuan: Fixed~

zhujinxuan: Because we need a argument to make the function working with right cache.  Therefore we would like to pass the plugins.  If we do not have plugins as argument, then it will always return the same value regardless of plugins change

packages/slate-react/src/plugins/after.js
```diff
@@ -595,10 +594,7 @@ function AfterPlugin() {
    */
 
   function renderEditor(props, editor) {
-    const handlers = EVENT_HANDLERS.reduce((obj, handler) => {
-      obj[handler] = editor[handler]
-      return obj
-    }, {})
+    const handlers = editor.getHandlers()
```
ianstormtaylor: I like how handlers have been simplified in other parts of this pull request. But I think here it feels strange. This is probably because it's a bit weird that we're rendering `<Content>` here, instead of passing it through as a child in the `Editor.render`. Could you change that? Then there will be no reaching for those handlers.

packages/slate-react/src/components/editor.js
```diff
@@ -71,130 +72,64 @@ class Editor extends React.Component {
...
-
-    // Resolve the plugins and create a stack and schema from them.
-    const plugins = this.resolvePlugins(props.plugins, props.schema)
-    const stack = Stack.create({ plugins })
-    const schema = Schema.create({ plugins })
-    this.state.schema = schema
-    this.state.stack = stack
-
-    // Run `onChange` on the passed-in value because we need to ensure that it
-    // is normalized, and queue the resulting change.
```
ianstormtaylor: Where has this logic gone? This is important to ensure that the editor never renders a `value` that is invalid according to the schema.

zhujinxuan: @ianstormtaylor This part of logic has gone to `get value`

`get value` use the resolved stack, and use the resolved stack to get value after `onChange` function.

packages/slate-react/src/components/editor.js
```diff
@@ -228,17 +163,37 @@ class Editor extends React.Component {
...
+   * this value.
+   *
+   * @param {Value} value
+   * @return {Schema}
+   */
+
+  normalizeOnChangeValue = memoize((value, stack) => {
+    const change = value.change()
+    stack.run('onChange', change, this)
+    return change.value
```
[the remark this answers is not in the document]
zhujinxuan: Hi, can I know how users get `change.operations` in current slate-react;  It seems users still cannot get `change.operations` after `onChange`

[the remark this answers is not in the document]
zhujinxuan: Do you mean this part 
```
  onChange = change => {
    debug('onChange', { change })
    if (this.value === change.value) return
    const value = this.normalizeOnChangeValue(change.value, this.stack)
```
the change.operations shall not be erased by `    const value = this.normalizeOnChangeValue(change.value, this.stack)` ?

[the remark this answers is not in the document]
zhujinxuan: Oh, I see. `onChange` must return a `change` with all operations in this cycle.  I am working on that.

packages/slate-react/src/components/editor.js
zhujinxuan: Oh, I see.  Do you mean if stack changes and value is unchanged, we shall still run `onChange`?

packages/slate-react/src/components/editor.js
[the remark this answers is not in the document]
zhujinxuan: Oh, I see.  I will use this memoize function only in `get value`; As for `onChange` function, we will run `stack.run('onChange', change, this)` manually.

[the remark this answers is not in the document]
zhujinxuan: Fixed~ this memoize function will be only used in `get value`

packages/slate-react/src/components/editor.js
ianstormtaylor: With the current logic, if the `<Editor>` is passed an invalid `value`, it will normalize it against the schema before rendering. And then after mounting it will ensure that the operations in that extra normalization pass are propagated to `onChange` so that the view layer is aware.

zhujinxuan:
Hi, @ianstormtaylor .  I think I am missing some of your points.  Can I know what the current problem is with this PR?

ianstormtaylor:
@zhujinxuan it's a bit hard to follow, but it looks better now. But I think it's still not compatible with the current behavior. Specifically this case:

1. Create an invalid `value` object, according to their schema.
2. Render an `<Editor>` with the invalid `valid`.
3. The editor should automatically normalize the value to make it value.
4. The editor should only ever render with a valid value.
5. Once mounted, the editor should call its own `onChange` with the new value.
6. The `change` object should include the operations that normalization took.

It looks like now this PR does everything but **6.** which is required for collaborative editing implementations to be able to know what operations were performed. It looks like it's creating a new change object instead as `onChange(value.change())` which means that the `change.operations` is empty.

zhujinxuan:
@ianstormtaylor How about this version?

ianstormtaylor:
I honestly don't understand what's going on there. What's the identity function for? Why are all the changes being cached in a `Map` at all? Why does that code path need to change in the first place?

This isn't something I'm likely to merge if it makes the logic (which already wasn't obvious) even more confusing than before.

zhujinxuan:
@ianstormtaylor I see. I just find that using `id` function and `Map` is unnecessary.  I make an update that works in this way: 
1. in `get value`, triggered by a change of props.  
    1.1 If the value is same as the value from the last lifecycle, use the value; otherwise 
    1.2 queue the change when get value is triggered 

2. In `componentDidMount`, `componentDidUpdate` and `onEvent`, call `this.change` to update change:
   2.1 Get change from queued change.
   2.2 If no operation happens since `this.props.value`, do not call props.onChange (when an event is triggered
   2.3 queue the value and stack.  Therefore the `get value` do not re-evaluate `stack.onChange` on a value which is already run with `onChange`.

packages/slate-react/src/components/editor.js
```diff
@@ -71,130 +72,59 @@ class Editor extends React.Component {
...
-    this.flushChange()
-
+  componentDidMount() {
     if (this.props.autoFocus) {
       this.focus()
+    } else {
+      this.change(change => change)
     }
+
+    this.tmp.updates++
```
ianstormtaylor: Why does `updates` increment here?

packages/slate-react/src/components/editor.js
```diff
@@ -228,15 +176,51 @@ class Editor extends React.Component {
...
+    this.queueChange(change)
+    return change.value
+  })
+
+  /**
+   * Queue a change object with value, to save operations of an value
+   * @param {Change} change
+   */
+
+  queueChange(change) {
```
ianstormtaylor: Can you remove this function and inline the logic instead since it's only used in one place now?

packages/slate-react/src/components/editor.js
```diff
@@ -252,22 +236,6 @@ class Editor extends React.Component {
     })
   }
 
-  /**
-   * On change.
-   *
-   * @param {Change} change
-   */
-
-  onChange = change => {
```
ianstormtaylor: We can't remove this `editor.onChange(change)` function because it's part of the public API that people are depending on. At least for now.

packages/slate-react/src/components/editor.js
zhujinxuan: This updates calculates the number of completed life cycles.  It will be compared to tmp.resolves to decide whether to warn in componentDidUpdate that schema is re-created too many times.

It is ++ in DidMount because in the first mount, resolvePlugins will increase the tmp.resolves.

packages/slate-react/src/components/editor.js
zhujinxuan: `onChnage` is restored in 37502ee9

packages/slate-react/src/components/editor.js
zhujinxuan: Fixed~

packages/slate-react/src/components/editor.js
```diff
@@ -228,17 +201,44 @@ class Editor extends React.Component {
...
+   *
+   * If the value is produced by onChange with the same stack, then will immediately return
+   * this value.
+   *
+   * @param {Value} value
+   * @param {Stack} stack
+   * @return {Change}
+   */
+
+  associateStackAndValue = memoizeOne((value, stack) => {
```
ianstormtaylor: Can you call this `resolveValue` to be consistent with the other places?

packages/slate-react/src/components/editor.js
```diff
@@ -228,17 +201,44 @@ class Editor extends React.Component {
...
 
   get stack() {
-    return this.state.stack
+    const plugins = this.resolvePlugins(this.props.plugins, this.props.schema)
+    return this.resolveStack(plugins)
   }
 
   get value() {
-    return this.state.value
+    if (this.tmp.value === this.props.value && this.tmp.stack === this.stack) {
```
ianstormtaylor: Can you add a comment here to explain what this is doing?

packages/slate-react/src/components/editor.js
```diff
@@ -71,140 +72,112 @@ class Editor extends React.Component {
...
    */
 
-  change = (...args) => {
-    const change = this.value.change().call(...args)
-    this.onChange(change)
+  onChange = change => {
+    debug('onChange', { change })
+    const { value } = change
+    if (value == this.props.value && change.operations.size === 0) return
+    this.stack.run('onChange', change, this)
```
ianstormtaylor: Thanks for re-adding this `onChange` method. 

We've now got another issue with this logic though, which is that there are three different places that `stack.run('onChange'` is called from, which I don't think is good because it makes things hard to follow.

Right now, it seems like one of the reasons for the duplication is that in the `componentDidMount` we're doing `this.change(change => change)` to actually flush any existing changes. This feels like too much magic, it isn't clear. And if that line was refactored to be more clear and not depend on the same code path, I think it would allow us to remove the duplcation.

For example (pseudocode):

```js
if (this.tmp.change) {
  this.flushChange()
}
```

Ideally `editor.change` should be reduced to:

```js
change = (...args) => {
  const { value } = this
  const change = value.change()
  change.call(...args)
  this.onChange(change)
}
```

Otherwise it probably means we're doing something too confusing.

packages/slate-react/src/components/editor.js
```diff
@@ -71,140 +72,112 @@ class Editor extends React.Component {
...
+    this.updateChange(change)
+  }
+
+  /*
+   * Update by change, and cache the current stack and value for update
+   * @param {Change} change
+  */
+
+  updateChange = change => {
```
ianstormtaylor: Similarly, I think this `updateChange` method should be only called from a single place, and then it wouldn't need to be its own method.

packages/slate-react/src/components/editor.js
zhujinxuan: Hi, @ianstormtaylor .  How about 3257e3e?  The problem is that `onChange(change)` can take an arbitrary change that not restricted to the inner state of the editor.  So we have to decide whether to run `stack.run(`onChange`)` in difference places.  One for inner change state `this.tmp.change` in `this.change`, the other for arbitrary change in `this.onChange`

zhujinxuan:
Update: this PR shall be able to close https://github.com/ianstormtaylor/slate/issues/1426 and https://github.com/ianstormtaylor/slate/pull/1427

ianstormtaylor:
Thanks @zhujinxuan! I've added a single commit there which refactors things a bit, and cleans up some of the comments and logging. It will hopefully make the code slightly easier to follow. (Allow it will still be confusing because memoization is inherently confusing I think.)

Thanks for working on this!

Issue #1977, opened 2018-07-17
toJSON can export history but fromJSON cannot instantiate it


t3chguy:
#### Do you want to request a _feature_ or report a _bug_?
Arguably a feature request

#### What's the current behavior?

`Value.toJSON` can be provided with `{preserveHistory: true}` and the output object will contain a `history` field but this field gets entirely ignored by `Value.fromJSON`

#### What's the expected behavior?

Ability to serialize and deserialize Value with History using `toJSON` and `fromJSON`

PR #1979, opened 2018-07-17
allow fromJSON to restore history

t3chguy:
Arguably adding a feature

#### What's the new behavior?

`Value.fromJSON` can now restore `history` from the given object

Trivial

Fixes: #1977

### DIFF
--- a/packages/slate/src/models/history.js
+++ b/packages/slate/src/models/history.js
@@ -54,6 +54,24 @@ class History extends Record(DEFAULTS) {
   }
 
   /**
+   * Create a list of `Operations` from `operations`.
+   *
+   * @param {Array<Object>|List<Object>} operations
+   * @return {List<Object>}
+   */
+
+  static createList(operations = []) {
+    if (List.isList(operations) || Array.isArray(operations)) {
+      const list = new List(operations)
+      return list
+    }
+
+    throw new Error(
+      `\`History.createList\` only accepts arrays or lists, but you passed it: ${operations}`
+    )
+  }
+
+  /**
    * Create a `History` from a JSON `object`.
    *
    * @param {Object} object
@@ -64,8 +82,8 @@ class History extends Record(DEFAULTS) {
     const { redos = [], undos = [] } = object
 
     const history = new History({
-      redos: new Stack(redos),
-      undos: new Stack(undos),
+      redos: new Stack(redos.map(this.createList)),
+      undos: new Stack(undos.map(this.createList)),
     })
 
     return history
--- a/packages/slate/src/models/history.js
+++ b/packages/slate/src/models/history.js
@@ -60,7 +60,7 @@ class History extends Record(DEFAULTS) {
    * @return {List<Object>}
    */
 
-  static createList(operations = []) {
+  static createOperationsList(operations = []) {
     if (List.isList(operations)) {
       return operations
     }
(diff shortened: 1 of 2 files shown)


packages/slate/src/models/history.js
```diff
@@ -64,8 +82,8 @@ class History extends Record(DEFAULTS) {
     const { redos = [], undos = [] } = object
 
     const history = new History({
-      redos: new Stack(redos),
-      undos: new Stack(undos),
+      redos: new Stack(redos.map(this.createList)),
```
jtadmor: should this just be `this.createList(redos)` / `this.createList(undos)`?

packages/slate/src/models/history.js
```diff
@@ -53,6 +53,24 @@ class History extends Record(DEFAULTS) {
...
+  /**
+   * Create a list of `Operations` from `operations`.
+   *
+   * @param {Array<Object>|List<Object>} operations
+   * @return {List<Object>}
+   */
+
+  static createList(operations = []) {
+    if (List.isList(operations) || Array.isArray(operations)) {
+      const list = new List(operations)
```
jtadmor: probably not a huge performance issue, but could this be:

```
if (List.isList(operations)) {
  return list
} else if (Array.isArray(operations)) {
  return new List(operations)
}
// error
```

packages/slate/src/models/history.js
t3chguy: @jtadmor nope, the stack is fine being generated from any iterable, but this class expects the stack to contain lists so we have to map all the things that will be in the stack to lists so that we can use `.last`

packages/slate/src/models/history.js
t3chguy: @jtadmor I agree but I was conforming to existing identical constructs within the code as the PR template commanda

packages/slate/src/models/history.js
jtadmor: ah i see, thanks.

packages/slate/src/models/history.js
t3chguy: fwiw I'd be happy to switch it over to the above ^

t3chguy: based on things like https://github.com/ianstormtaylor/slate/blob/b26535fd947cb54153800799a1dfc5d4b7729218/packages/slate/src/models/block.js#L65-L74

packages/slate/src/models/history.js
t3chguy: its needed for when the stored object has been serialized and `List` gets converted to an `Array` and `History` previously did not re-box it

packages/slate/src/models/history.js
t3chguy: I went ahead and did this based on your suggestion

packages/slate/src/models/history.js
```diff
@@ -53,6 +53,27 @@ class History extends Record(DEFAULTS) {
...
   }
 
+  /**
+   * Create a list of `Operations` from `operations`.
+   *
+   * @param {Array<Object>|List<Object>} operations
+   * @return {List<Object>}
+   */
+
+  static createList(operations = []) {
```
ianstormtaylor: Hey @t3chguy can you rename this to `createOperationsList`?

t3chguy: Sure

t3chguy:
@ianstormtaylor back with you, PTAL

ianstormtaylor:
Thanks @t3chguy!

Issue #1978, opened 2018-07-17
Value.fromJSON doesn't regenerate {anchor,focus}Key from {anchor,focus}Path


t3chguy:
#### Do you want to request a _feature_ or report a _bug_?
Arguably a feature request

#### What's the current behavior?

`fromJSON` requires the `selection` field to have `anchorKey` and `focusKey` even though by default if you `preserveSelection` on `toJSON` it omits those fields and instead has `anchorPath` and `focusPath` which in conjunction with the document can be used to regenerate the required fields. With those fields missing `selection.isSet` is always false so `fromJSON` resets the selection to the start of the composer.
I believe code from https://github.com/ianstormtaylor/slate/blob/e4ff1972d79911baf8a6aed683765b66107b09a4/packages/slate/src/operations/apply.js#L446-L454 is required in `Value.fromJSON`

#### What's the expected behavior?

Ability to import a Value with a stored selection without maintaining keys

PR #1980, opened 2018-07-17
rebuild selection from anchorPath and focusPath if keys were dropped

t3chguy:
Arguably adding a feature

#### What's the new behavior?

`Value.fromJSON` can rebuild the selection `anchorKey` and `focusKey` from `anchorPath` and `focusPath` respectively if they were dropped at `Value.toJSON`

Uses existing code from https://github.com/ianstormtaylor/slate/blob/e4ff1972d79911baf8a6aed683765b66107b09a4/packages/slate/src/operations/apply.js#L446-L454

Fixes: #1978

### DIFF
--- a/packages/slate/src/models/value.js
+++ b/packages/slate/src/models/value.js
@@ -100,6 +100,18 @@ class Value extends Record(DEFAULTS) {
     let data = new Map()
 
     document = Document.fromJSON(document)
+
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Good call, thanks @t3chguy!

PR #1987, opened 2018-07-20
Move `insertBlock`s outside of void inlines

justinweiss:
When inserting a block, we shouldn't split void inlines -- otherwise,
they're duplicated. Instead, move the insert to either before or after
the void node.

Fixing a bug.

#### What's the new behavior?

When inserting a block while a void inline is selected, the block will be inserted before or after the block, rather than splitting the void node.


ianstormtaylor:
Beautiful PR, thank you @justinweiss!

PR #1990, opened 2018-07-23
Accept legacy "ranges" additionally to "leaves"

Kronuz:
Accept legacy "ranges" additionally to "leaves"

#### Feature to improve compatibility

In our migration from old Slate v0.25.3 to v0.34.6, we had a whole bunch of documents in our database using the old JSON structure, changing those will mean we'll have to do a lot of planing ahead of time. Knowing what was changed, and doing some testings, we found out the new Slate could *almost* read old documents, if it wasn't for a single thing which didn't read: `ranges`, from `Text` nodes. Everything else seems to work.

#### What's the new behavior?

In order to maintain compatibility with old JSON documents and in the same spirit as currently already supported old properties (`kind`, for example), additionally to the new "leaves" property for `Text` nodes, also accept "ranges" with a clear deprecation warning.

Increases compatibility with a very small change to makes it simpler to upgrade to newer Slate.


packages/slate/src/models/text.js
```diff
@@ -88,6 +88,15 @@ class Text extends Record(DEFAULTS) {
     const { key = generateKey() } = object
     let { leaves = List() } = object
 
+    if (leaves.size === 0 && object.ranges) {
```
ianstormtaylor: I don't think this will work as written for JSON arrays, since they need `length === 0` instead.

Kronuz: @ianstormtaylor that is correct! I guess we could use this instead, if you have no objections, I’ll gladly update the pull request :

```javascript
    let { leaves } = object

    if (!leaves) {
      if (object.ranges) {
        logger.deprecate(
          'slate@0.27.0',
          'The `ranges` property of Slate objects has been renamed to `leaves`.'
        )
        leaves = object.ranges
      } else {
        leaves = List()
      }
    }
```

ianstormtaylor:
Thanks @Kronuz, great PR! 😄

Issue #1923, opened 2018-06-20
refactor `Schema` to differentiate "validating" and "normalizing"


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Feature.

#### What's the current behavior?

Right now the `schema.validateNode` method returns a change object that will "fix" the invalid state, keeping the editor in a valid state. This is required for building complex behaviors without letting the editor get into invalid states.

However, there's another type of validation that it doesn't cater to: user-facing validations, for example "your tweet is too long". These aren't things that should be auto-fixed. They should instead be presented to the user as errors, however the developer sees fit.

But this is currently hard to do, since `validateNode` returns a black box.

#### What's the expected behavior?

I think it would be good to instead, rename `validateNode` to `normalizeNode`, and have it call into a new `validateNode` function under the covers.

The new `validateNode` would simply return a `SlateError` (or not) with whatever the reason for the State being invalid is.

Similarly, the schema would expose two functions: `schema.normalize(value)` which would do the auto-fixing, and `schema.validate(value)` which would return errors.

With these in place, I think userland would have everything it needs to be able to build complex user-facing errors using the existing `Schema` definitions.

PR #1993, opened 2018-07-24
Refactor schema

ianstormtaylor:

Improvement.

#### What's the new behavior?

- Tweaking the declarative schema definition syntax to make it easier to represent more complex states, as well as enable it to validate previously impossible things.
- Rename `validateNode` to `normalizeNode` for clarity.
- Introduce `validateNode`, `checkNode`, `assertNode` helpers for more advanced use cases, like front-end API validation of "invalid" fields that need to be fixed before they are sent to the server.

The `schema.blocks/inlines/document` entries are now a shorthand for a more powerful `schema.rules` syntax. For example, this now allows for declaratively validating by a node's data, regardless of type:

```js
{
  rules: [
    {
      match: {
        data: { id: '2kd293lry' },
      },
      nodes: [
        { match: { type: 'paragraph' }},
        { match: { type: 'image' }},
      ]
    }
  ]
}
```

Previously you'd have to use `validateNode` for this, since the syntax wasn't flexible enough to validate nodes without hard-coding their `type`.

This also simplifies the "concatenation" of schema rules, because under the covers all of them are implemented using the `schema.rules` array, so they simply take effect in order, just like everything else in plugins.

Fixes: #1842
Fixes: #1923


Soreine:
Hi @ianstormtaylor :)

> - Tweaking the declarative schema definition syntax to make it easier to represent more complex states, as well as enable it to validate previously impossible things.

Great :)

> - Rename validateNode to normalizeNode for clarity.

Why not

> - Introduce validateNode, checkNode, assertNode helpers for more advanced use cases, like front-end API validation of "invalid" fields that need to be fixed before they are sent to the server.

I haven't seen these in the code changes. How would one use them ? Are they just functions we can run on a given document, but which don't result in an actual normalization ?

--- 

> The schema.blocks/inlines/document entries are now a shorthand for a more powerful schema.rules syntax.

That sure is more powerful. But this change is essentially reverting to the old schema in terms of performance. There are two performance benefits to the current schema:

1. To normalize a given node, we only need to lookup its type in the schema to get the corresponding rule. No need to run every `validate` function on it, only the ones relevant to this node type.
2. With the schema encoded as data instead of function, we lose flexibility, but we gain the ability to merge several rules together into a single object. This allow plugins to extend the schema without necessarily grow a list of `validate` functions.

Your proposed change is correct in term of abstractions, but it completely defeats 1. and 2.
To me, it looks like the old schema (a list of `{ match, validate, normalize }` functions triplet) but encoded as an object. Whereas the current code trades off syntax complexity with performance, the new code trades off syntax complexity just for the sake of writing less line of code.

I'd be happy if you prove me wrong. But right now, I'm scared of the performance issues that would come with this PR.

ianstormtaylor:
Hey @Soreine, thanks for the review!

> I haven't seen these in the code changes. How would one use them ? Are they just functions we can run on a given document, but which don't result in an actual normalization ?

Yup, exactly. We will probably want to introduce even more helpers here (potentially taking some of the `normalizeNodeAndChildren` logic from the `with-schema` changes and moving it here), so that you can end up calling just `schema.validate(value)` yourself and receive an error for what is invalid. But this is just the start of that. The methods in question start here: https://github.com/ianstormtaylor/slate/pull/1993/files#diff-ca226d74d2860e82cf177ee1e086dd13R142

---

As for your performance concerns, I think we'd need to answer a more fundamental question about the architecture, and then the answer might be a bit more clear...

Right now, the declarative schema validation rules can only be matched by `object+type`. Which allows for doing the most common, simple things like:

- Validating `block.list_item` nodes to ensure they're in a list.
- Validating `document` nodes to ensure they contain only specific children.
- etc.

My question (which I'm slightly unsure of still) is whether this is where the schema should stop. Or whether it is a valid, first-class use case to be able also take `node.data` into account (in addition to `node.object` and `node.type`) when deciding how to match schema validation rules. Enabling use cases like...

- Validating `block.heading.level=3` nodes differently.
- Validating `block.field.id=xxxxxxx` nodes based on a user-defined schema.

The heading example is contrived, but the real place this comes into effect is in configuration that is user-defined that happens to affect people's editor schema. 

If the user-defined configuration is an infinite realm, it isn't really possible to only use `object+type` when validating it. (Unless you made your type strings `${type}_${id}` which doesn't seem like a good solution?) Trying to solve for this case nicely is what prompted these changes to the validation architecture.

To me, it seems like the matching `object+type+data` gives people the most flexibility, and feels like it is the "right" call from Slate's point of view to stay un-opinionated but still powerful enough to build complex features.

---

Another way to try to figure out the answer to this problem, is to imagine we've gotten rid of `node.isVoid` and that that logic is now handled by the schema instead. It would be present as an `isVoid` property of a schema rule, just like it currently is.

And then the question is what level of complexity should you be able to determine void-ness by. In this case I'd probably go with just `object+type` personally, which is what makes me question the earlier decision.

---

As for the potential performance issues...

> To normalize a given node, we only need to lookup its type in the schema to get the corresponding rule. No need to run every validate function on it, only the ones relevant to this node type.

This is true, it is strictly slower than the pure `object+type` lookup. However, I'm not sure this was ever a true performance issue? I've just re-read all of the old schema-related pull requests and it was never mentioned as one of the reasons for changing from the old schema system to the current one. The big reasons for the change was (a) to separate out rendering logic, and (b) to make schemas more expressive.

However, making the schemas expressive actually did have significant performance gains, due to rules like `parent` being able to be collapsed into a single iteration. But this pull request retains these gains.

> With the schema encoded as data instead of function, we lose flexibility, but we gain the ability to merge several rules together into a single object. This allow plugins to extend the schema without necessarily grow a list of validate functions.

Again, I'm not really sure this is a true performance bottleneck. Unless the actual filtering of the schema rules is the biggest performance issue this isn't really an issue. And it's much more important then to allow for the necessary flexibility. (Further, since the matching objects are declarative, we can choose to implement short `PERF:` shortcuts later if needed, since they aren't blackbox functions.)

Even further... the current "core" schema rules are implemented as `validateNode` (now `normalizeNode`) functions because the existing expressive schema rules weren't flexible enough to be able to handle them. I'd guess that if making the schema rules more flexible means we're able to convert those core functions into declarative rules we'll actually see a performance increase instead of a decrease.

Not only that, I'm not convinced that the "merging" of rules was a good thing. I think it makes the rules very hard to reason about, since they merge together in complex ways. Especially when you take the `normalize` functions into account, it's way too complex. This new architecture is much more clear, and mimics how all the other plugin-related logic works using a stack.

> Your proposed change is correct in term of abstractions, but it completely defeats 1. and 2. To me, it looks like the old schema (a list of { match, validate, normalize } functions triplet) but encoded as an object. Whereas the current code trades off syntax complexity with performance, the new code trades off syntax complexity just for the sake of writing less line of code.

It is similar (which actually is a good thing for flexibility), but I think its declarative-object nature is where we gain the most performance benefits, not by artificially restricting the matching flexibility.

---

Another another note...

> On a side note, the code that applies the schema rules could also benefit from being refactored. I think it can be made lighter to digest (but identical in logic). This could be done in this PR, or in another PR. But I would feel bad to bluntly patch something onto the existing code.

...I definitely agree with you. I didn't have enough time here to find a better abstraction and make it more readable, but I agree that the schema validation logic is a clusterfuck right now. Maybe even splitting it into separate functions would be better to start.

zhujinxuan:
> Rename validateNode to normalizeNode for clarity.
I am thinking whether we shall put together the logic of matcher and fixer in either `validateNode` and `normalizeNode`

Perhaps we can improve the interface as `validateNode` throws an message, then `fixNode/normalizeNode/...` catches the message and fix it (like the action-reducer in redux).  

To keep a compatable interface, If `validateNode` throws an function, we do it like redux-thunk, call the function as `callback(change)`.

Perhaps we can fix code concentration in this way?

zhujinxuan:
But perhaps we can have normalizeNode as another interface:
`normalizeNode(node: Node) => void|Node`

If normalizeNode returns a new node, then replace the old node with new node and restore the cursor.  The only problem is that creating new node often depends our internal APIs.

ianstormtaylor:
@zhujinxuan indeed, this pull request makes it sort of work like that. The `validateNode(node) => Error|Void` and then the `normalizeNode(node) => Function|Void` based on whether validation returned an error. Is that what you're describing?

As for the `normalizeNode(node) => Node|Void`, that would be pretty nice since it would make it easier to memoize normalizations. But the problem I see is that it means we'd not be using operations to make those changes, so it breaks the requirement to have all changes to the document performed by the low-level operations. Maybe there's another similar way?

zhujinxuan:
@ianstormtaylor Almost Yes.  But I am thinking about something like:

```
validateNode(node) => Object | Void
normalizeNode(Object) => Function
```


We can take the return of `validateNode(node)` to get information obtained by the matcher, and makes the normalizeNode neater.  It is more alike the `fail` and `schema` you did in slate-core.  


As for `fixNode(node) => Node|Void`, I think we can:
```
normalize() {
  const newNode = fixNode(node);
  const {selection} = change.value
  if (Node.isNode(result) {
      change.replaceNodeByKey(node, newNode, {normalize: false})
      change.restoreSelection(selection, node, newNode) 
  }
}
```
We are still using operations to update.

zhujinxuan:
As for memoize, I think we only need to memoize whether a node shall be normalized.  Because invalid node will be soon removed from the document, how a node is normalized is not important.

The advantage of `node => NewNode` interface, I think, is everything in function is pure, and has a shorter history stack eventually (only two operation by replaceNode and one operation by reset selection).  I feels that some logics `node.nodes.forEach(n => change....)` can be simpler if we can just write as `node.nodes.map` `node.nodes.reduce` (Just some intuitive thoughts, not yet explored in detail)

ianstormtaylor:
@zhujinxuan Using `replaceNodeByKey` doesn't work because it ignores the "intent" that is necessary for collaborative editing implementations. If the document is invalid, and a new document is returned, the entire thing will be replaced, clobbering anyone who's currently adding their own edits. It needs to use the low-level operations to make changes so that they can be transformed against each other.

ianstormtaylor:
@Soreine following up on performance concerns. I've created a simple test case and used the new `--inspect` flag to use the Chrome profiler. Given a simple profiling case of:

```js
const { Value } = require('slate')

console.profile('profile')

for (let i = 0; i < 10; i++) {
  Value.fromJSON({
    document: {
      nodes: Array.from(Array(10)).map(() => ({
        type: 'list',
        object: 'block',
        isVoid: false,
        data: {},
        nodes: Array.from(Array(5)).map(() => ({
          type: 'list-item',
          object: 'block',
          isVoid: false,
          data: {},
          nodes: [
            {
              type: 'link',
              object: 'inline',
              isVoid: false,
              data: {
                id: 1,
              },
              nodes: [
                {
                  leaves: [
                    {
                      object: 'leaf',
                      marks: [],
                      text: 'Some text for a link',
                    },
                  ],
                  object: 'text',
                },
              ],
            },
          ],
        })),
      })),
    },
  })
}

console.profileEnd('profile')
debugger
```

Here's a flamegraph of a single pass of the `Value.fromJSON` method:

image

Those little dark blue squares are the time spent in the native `Array.filter` method which are what is used to do `this.rules.filter` (as opposed to using the previous system of direct object access). It's a very small amount of time overall (`< 5%`), and could probably be optimized further if necessary. There are much bigger places though that we stand to gain performance.

Here's an alternate compiled view of the places where time is spent:

- **85%** `Value.fromJSON`
    - **80%** `Change.normalize`
        - **65%** `normalizeNodeAndChildren`
            - **37%** `iterate`
                - **22%** `rule.normalize`
                    - **20%** `Change.insertNodeByKey`
                        - **17%** `Value.insertNode`
                            - **12** `Value.mapRanges`
                            - **4%** `Node.insertNode`
                                - **2%** `Node.replaceNode`
                                    - **1%** `flatMap`
                                    - **1%** `setIn`
                                - **1%** `splice`
                        - **1%** `Operation.create`
                    - **1%** `Text.create`
                    - **8%** `Schema.normalizeNode`
                        + **4%** `new SlateError`
                        + **2%** `validateRules`
                        + **2%** `filter`
                    - **6%** `iterate`
                        + **6%** `Schema.normalizeNode`
                            * **4%** `validateRules`
                            * **2%** `filter`
            - **22%** `rule.normalize`
                - **20%** `Change.insertNodeByKey`
                    - **17%** `Value.insertNode`
                        - **12** `Value.mapRanges`
                        - **5%** `Node.insertNode`
                            - **3%** `Node.replaceNode`
                                - **2%** `flatMap`
                                - **1%** `setIn`
                            - **1%** `splice`
                    - **1%** `Operation.create`
                - **1%** `Text.create`
            - **4%** `Node.getFirstInvalidDescendant`
                - **3%** `Schema.normalizeNode`
                    - **1%** `validateRules`
                    - **1%** `filter`
                - **1%** `Node.getFirstInvalidDescendant`
        - **15%** `Node.getFirstInvalidDescendant`
            - **9%** `Node.getFirstInvalidDescendant`
                - **6%** `Schema.normalizeNode`
                    - **5%** `new SlateError`
                    - **1%** `validateRules`
                - **2%** `Node.getFirstInvalidDescendant`
                    - ...
            - **6%** `Schema.normalizeNode`
                - **4%** `validateRules`
                - **1%** `filter`

I think the biggest wins, from dissecting that would be:

- Tweaking `Node.getFirstInvalidDescendant` to use the new `Schema.validateNode` method instead of the slower `Schema.normalizeNode`. This is a small fix that should have already been applied, but you can see how much time is spent in `new SlateError` which it eliminates (`~ 9%`).

- Allowing `rule.normalize` to have access to the `path` of the node, so that they can use `insertNodeByPath` instead of `insertNodeByKey` which avoids have to recompute the `key->path` map for each pass of normalization, which could be huge since it spends a lot of time there (`~ 40%`).

- Making the `value.*` mutation methods smarter about how they use `mapRanges`. Right now lots of them set the `anchorPath/focusPath` to `null`, to allow them to be re-synced with the keys when they make changes. Specifically in this case the `Value.insertNode` method. This isn't necessary because we could write a bit more logic to compute what the updated paths should be, which would avoid having to recompute the `key -> path` mapping on each normalization pass. This could be fairly significant too (`~ 24%`).

- Removing the core requirement of `inline non-void nodes are never empty` which requires a `match` argument that checks the text content of child nodes. This is slow because calculating text is slow, and because filtering by child nodes is slow. And this shouldn't really be something the core library needs to do. But this is a small concern, since this is only affecting that `< 5%` that is in the `rules.filter` path.

---

So, with all that in mind, I think changing the schema rules to be stack-based with filtering is okay to do. It will make things a lot easier to reason about, and make Slate more flexible, with seemingly not that big of a performance impact.

zhujinxuan:
Hi, @ianstormtaylor . Can you inspect how much of `Value.mapRanges` is consumed by `Node.getKeysToTableDict`?  I try to implement re-use path logic in `Value.mapRanges`,  but I cannot find any significant improvement.

I think if we do not call `Node.getKeysToTableDict` in `Value.mapRanges`, then the next `insertNodeByKey` will call `Node.getKeysToTableDict`.  As a result, the efficiency is not improved even we tries to re-use path in `Node.getKeysToTableDict`

ianstormtaylor:
@zhujinxuan you can check it yourself by using the Chrome Dev Tools and recording a profile. It's very possible that a subsequent call to `getKeysToPathsTable` will negate the gains from using paths in `mapRanges`. But that just means that the second place where keys are used that paths should be used instead too. It's because we're using keys in too many places, and we need to switch more of them to see the proper performance improvement. 

But any switches from keys to paths are very welcome, even if they don't improve performance themselves because of a bottleneck after them.

Issue #1567, opened 2018-01-26
Store paths instead of keys on selection data for `set_selection` operations


justinweiss:
#### Do you want to request a *feature* or report a *bug*?

Feature.

#### What's the current behavior?

Currently, `set_selection` operations store some selection details in the `selection` property, and other details in the `properties` property. `selection` is a Range, and only stores keys, not paths.

In order to support transforming selections against other operations, we need a way to store `anchorPath`  and `focusPath` on Range -- either instead of, or in addition to, the key.

Here's what I'm trying to do with transformations:

1. Convert `operation.selection` into an plain JS object
2. Turn `selection.anchorKey, selection.focusKey` into `anchorPath, focusPath`
3. Merge `properties` with my plain JS `selection` object (because properties might only contain the difference from selection)
4. Transform my `selection` JS object against the incoming op
5. Transform `properties` against the incoming op
6. Turn the new `anchorPath, focusPath` on the selection object back into keys
7. Attach the transformed objects back onto the `set_selection` operation

The problem happens in Step 6. In Step 2, to turn the original anchor/focusKey into a path, I can use the `document` on `operation.value`. But in Step 6, those transformed anchor/focusPaths only apply to the new document. 

This usually isn't a problem -- you can get the new state by `Operation.apply`ing the incoming operation to the selection op's value. But the keys you get from that value are going to be different than the keys that eventually show up in the document, even though the paths will be the same. For example, if you call 

```
value1 = Operation.apply(value, split_node_op)
value2 = Operation.apply(value, split_node_op)
```

the new node might have a different key in `value1` and `value2`, even though it's the same node in each document.

The only way I can think of to fix this would be to have Range allow you to use paths, and either allow you to also set key, or have key be a calculated property based on path and value.

What do you think? I'm happy to take a stab at a PR, if we agree on a direction.

#### What's the expected behavior?

In a collaborative environment, keys can appear and disappear, but paths can be transformed and stay valid. In order to support both collaboration and undo / redo, `set_selection` needs to store both the old selection and the new selection as paths, instead of keys.

bengotow:
Hey! I was just looking to do the same thing for #1402 (trying to serialize the operation history as JSON and not as ImmutableJS objects with memoized cached values). It seems like the selection operation should serialize it's `selection` property / Range, and when #1408 is implemented, the`Range` object will be switched to always having paths?

justinweiss:
[...]

PR #1997, opened 2018-07-26
add paths to ranges

ianstormtaylor:

Feature.

#### What's the new behavior?

This pull request adds paths to `Range` objects, including the selection. The paths and keys are kept in sync automatically, so that you can use whichever is ideal for your use case.

This should allow us to use paths for lots of the internal logic, which are much quicker to work with than keys since they avoid having to lookup the key in the document and can just traverse right to the node in question.

`Range` objects have two new properties:

```js
range.anchorPath
range.focusPath
```

(Eventually these will be `range.anchor.path` and `range.focus.path` when points are introduced.)

When operations occur and whenever ranges are created/normalized, the paths are updated and kept in sync with the keys.

Fixes: https://github.com/ianstormtaylor/slate/issues/1408
Fixes: https://github.com/ianstormtaylor/slate/issues/1567


Slapbox:
This sounds awesome!

packages/slate/src/utils/path-utils.js
```diff
@@ -0,0 +1,201 @@
...
+ * Compare paths `a` and `b` to see which is before or after.
+ *
+ * @param {List} a
+ * @param {List} b
+ * @return {Number|Null}
+ */
+
+function compare(a, b) {
+  a = a.toArray()
+  b = b.toArray()
```
jtadmor: any reason we are calling `toArray` here? does it make the `for` iteration below much faster?

it seems we could at least do it after the check, `if (a.size !== b.size) return null` then coerce to make this faster in the base case.

ianstormtaylor: @jtadmor good call. Nope no reason. I actually had originally written the logic for arrays and then realized they needed to be lists instead and didn't feel like re-writing it 😄

packages/slate/src/models/node.js
```diff
@@ -969,14 +976,15 @@ class Node {
    */
 
   getMarksAsArray() {
-    // PERF: use only one concat rather than multiple concat
-    // becuase one concat is faster
     const result = []
 
     this.nodes.forEach(node => {
       result.push(node.getMarksAsArray())
```
jtadmor: i think if we're trying for full perf, we should consider:

`result.push(...node.getMarksAsArray())`
`return result`

https://jsperf.com/slate-concat-vs-push/1

jtadmor: (might be out of scope for this PR, but if we are making changes here...)

ianstormtaylor: For me the `concat` in that JSPerf runs faster actually

zhujinxuan:
Great!  After this PR, will we have both key and path in range?

packages/slate/src/models/node.js
ianstormtaylor: @zhujinxuan thanks!

These types of perf improvements are kind of minor, and can be changed/improved in subsequent pull requests. I prefer to always go with the most idiomatic, native version of the API when possible because browsers are constantly getting faster in these regards.

ianstormtaylor:
@zhujinxuan yup! We'll have `range.anchorKey` and `range.anchorPath`.

I'm sorry for making your work on #1908 harder, since this probably touches lots of that code. I was just working on the schema/normalization and realizing that this "add paths" architectural change was important, and wanted to get it done as soon as possible since it impacts so many other things.

Right now there is still lots of low-hanging fruit in the implementation, since most of the `*AtRange` change methods still use keys instead of paths for lookups. I think most could be converted to using paths, and if so then it's possible that all of Slate's core changes would use paths and the key map would be recomputed much less, which should speed things up.

zhujinxuan:
@ianstormtaylor No problem.  I will wait for your schema and range work and then go back to the `Point` work.

Do you think your current work is conflicting with https://github.com/ianstormtaylor/slate/pull/1811?  It seems that they are not conflicting.

ianstormtaylor:
@zhujinxuan ah yup, I think it does conflict with #1811. I was going to update #1811, #1763 and #1818 to mention that. I think all of those pull requests will need to be closed (since I pulled lots of inspiration from each of them in doing this one) and additional improvements can be made in subsequent ones. Lots of the internals have become more streamlined so some of the tweaks in those pull requests might not be necessary. Either way the library has gotten a lot cleaner.

PR #2018, opened 2018-08-01
fix react-portal import

### LINKED ISSUE #2017: Wrong import for 'react-portal'
The `Portal` is imported as a default import: https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/editor.js#L2

This probably works with the non-es version of the package, however if you use the es-bundle this will crash the build. The correct way to import 'react-portal' is (https://github.com/tajo/react-portal/blob/master/README.md):

```
import { Portal } from 'react-portal';
```

Changing this, will fix the build.

jfrolich:
fixes #2017

### DIFF
--- a/packages/slate-react/src/components/editor.js
+++ b/packages/slate-react/src/components/editor.js
@@ -1,5 +1,5 @@
 import Debug from 'debug'
-import Portal from 'react-portal'
+import { Portal } from 'react-portal'
 import React from 'react'
 import SlateTypes from 'slate-prop-types'
 import Types from 'prop-types'


jfrolich:
The first commit didn't work, as previous versions of react-portal didn't have a es-modules entrypoint, so rollup didn't detect the named export by default. Updating to this version also has advantages due to leveraging the react 16 public API for portals.

jfrolich:
Hey any thoughts about merging this? This is still an open issue.

ianstormtaylor:
@jfrolich sorry, I'm down to merge now as long as it doesn't break compat with versions of React under 16. Can you resolve the merge conflicts? Thanks!

ianstormtaylor:
Actually @jfrolich, I'm going to solve this in #2113 by removing the deprecated `renderPortal` in the first place, and letting people use proper fragments and portals from React 16 to solve this instead.

jfrolich:
Ah great @ianstormtaylor, makes sense!

PR #2032, opened 2018-08-03
Docs: Fix typo in guides/rendering

FWirtz:
@ianstormtaylor found another typo, see below.
~If you want you can keep this PR open (in case I find another typo) and I'll ping you(/whoever else) when I am done reading the docs?~ **edit** nvm think I read all articles

Really loving the docs so far and looking forward to working with Slate! 👏 👍 ⚡️ 

---

In [guides/rendering#Placeholders](https://docs.slatejs.org/guides/rendering#placeholders):
> That will render a simple placeholder element inside all of **~the~ your** caption blocks until someone decides to write in a caption.

---

Fixing a bug/typo in the docs.

#### What's the new behavior?
This fixes a typo in the docs.

See above.

No.

* [X] ~The new code matches the existing patterns and styles.~
* [X] ~The tests pass with `yarn test`.~
* [X] ~The linter passes with `yarn lint`. (Fix errors with `yarn prettier`.)~
* [X] ~The relevant examples still work. (Run examples with `yarn watch`.)~

No, I don't think so.

### DIFF
--- a/docs/guides/rendering.md
+++ b/docs/guides/rendering.md
@@ -120,7 +120,7 @@ function renderPlaceholder(props) {
 />
 ```
 
-That will render a simple placeholder element inside all of the your `caption` blocks until someone decides to write in a caption.
+That will render a simple placeholder element inside all of your `caption` blocks until someone decides to write in a caption.
 
 ## The Editor Itself
 


FWirtz:
oh one question: in the `Hotkey` definition in [XXX]():
```js
function Hotkey(hotkey, fn) {
  return {
    onKeyDown(event, change, editor) {
      if (isHotkey(hotkey, event)) {
        change.call(fn)
        // return true here?
      }
    },
  }
}
```

Is it correct that, as there is no `return true`, if two different plugins define the same hotkey, **both** would be evaluated?

ianstormtaylor:
Thanks @FWirtz! And yup, that’s correct.

FWirtz:
thanks!

Issue #1932, opened 2018-06-23
New lines are removed on copy


thetutlage:
#### Do you want to request a feature or report a bug?
Bug

#### What's the current behavior?
This has been reported earlier https://github.com/ianstormtaylor/slate/issues/1635 and then a related PR was created https://github.com/ianstormtaylor/slate/pull/1644.

Somehow it doesn't work as expected with `slate@0.34.2`.

**In Action**

![new-lines-skipped](https://user-images.githubusercontent.com/1706381/41805253-f2f6d94a-76c3-11e8-96eb-67455d1e154b.gif)

#### What's the expected behavior?
The new lines must be copied over

thetutlage:
After doing bit of debugging. This is not a problem with the empty nodes. The `\n` is not preserved even between 2 text nodes.

![missing-line-breaks](https://user-images.githubusercontent.com/1706381/41805367-b4cd2a0e-76c6-11e8-82a0-b3e30799d213.gif)

Also the `textContent` property from output of `range.cloneContents()` doesn't preserve line breaks.

https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/clone-fragment.js#L32

```js
contents.textContent // no line breaks
```

thetutlage:
One way I can think of is to add a new `br` element inside this `forEach` loop https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/clone-fragment.js#L36.

I tried it and it works fine. Not sure if this is the best way to do it.

```
contents.childNodes.forEach(function (node) {
 if (node.textContent && node.textContent.trim() !== '') {
   attach = node;
 }
+ node.appendChild(document.createElement('br'))
});
```

ianstormtaylor:
@thetutlage thanks for investigating! We’re we always using `textContent`? Might be something to check in Git blame. I don’t recall this issue in the past. Wonder if it was a recent regression.

thetutlage:
Yes added in this commit https://github.com/ianstormtaylor/slate/commit/ee8bc1b695930fbd79c17b256511744b3c9b54f9 `13 days ago`

ianstormtaylor:
Ah okay, should be an easy fix then. Maybe does `innerText` solve our needs? Not sure.

tobiasandersen:
[...]

PR #2040, opened 2018-08-06
Adds Unicode LF character to line endings for clipboard plaintext

Slapbox:
This addresses https://github.com/ianstormtaylor/slate/issues/1932

I do not believe this will break any existing functionality. In my testing I could not find any issues with this solution so far.

### DIFF
--- a/packages/slate-react/src/utils/clone-fragment.js
+++ b/packages/slate-react/src/utils/clone-fragment.js
@@ -1,4 +1,6 @@
 import Base64 from 'slate-base64-serializer'
+import Plain from 'slate-plain-serializer'
+import { Value } from 'slate'
 import TRANSFER_TYPES from '../constants/transfer-types'
 import getWindow from 'get-window'
 import findDOMNode from './find-dom-node'
@@ -82,16 +84,15 @@ function cloneFragment(event, value, fragment = value.fragment) {
 
   attach.setAttribute('data-slate-fragment', encoded)
   
-  //  Builds plaintext object for clipboard line by line using top-level nodes.
-  //  Adds '\u000A' (unicode LF) to end of each line except the last.
-  //  This prevents line breaks from being lost when pasting to targets
-  //  which don't support html input.
-  let plainText = ''
-
-  for (let i = 0; i < contents.childNodes.length; i++) {
-    plainText += contents.childNodes[i].innerText
-    if (i < contents.childNodes.length - 1) plainText += '\u000A'
-  }
+  //  Creates value from only the selected blocks
+  //  Then gets plaintext for clipboard with proper linebreaks for BLOCK elements
+  //  Via Plain serializer
+  const valFromSelection = Value.create({
+    document: {
+      nodes: value.blocks,
+    },
+  })
+  const plainText = Plain.serialize(valFromSelection)
 
   // Add the phony content to a div element. This is needed to copy the
   // contents into the html clipboard register.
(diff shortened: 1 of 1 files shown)


zhujinxuan:
Does this work with table-cell blocks?

Slapbox:
No it does not, **it also doesn't work with list items. Working on list items now.**

I don't think this method will work with table items, though I'm not sure.

zhujinxuan:
@Slapbox Perhaps we can find the dom by node and check the tagName to solve the table and list problem?  But I am afraid that it is a bit too complex

packages/slate-react/src/utils/clone-fragment.js
```diff
@@ -81,6 +81,17 @@ function cloneFragment(event, value, fragment = value.fragment) {
...
+  
+  //  Builds plaintext object for clipboard line by line using top-level nodes.
+  //  Adds '\u000A' (unicode LF) to end of each line except the last.
+  //  This prevents line breaks from being lost when pasting to targets
+  //  which don't support html input.
+  let plainText = ''
+
+  for (let i = 0; i < contents.childNodes.length; i++) {
+    plainText += contents.childNodes[i].innerText
+    if (i < contents.childNodes.length - 1) plainText += '\u000A'
```
ianstormtaylor: Hey @Slapbox any reason you're using `'\u000A'` instead of just `'\n'`?

Slapbox: No good one. We can change it to `\n`.

I just wanted to be extra-explicit for myself about what characters were in use while I was comparing the unicode outputs of Slate vs other major editors.

Slapbox:
Is there some way to differentiate between these 3 elements for the sake of recursion?

image

I'm trying to expand this to cover cases like content in list items, but I'm not sure how to determine with certainty that an element is a block from the regular DOM. A seemingly easy approach was to add `\n` to the end of each element that wasn't a `span`, but inlines break this idea.

Slapbox:
I'm not really sure.

The serializer requires a `value` object and then peels off the `document` from that.  I looked at how the plain serializer works though and I think it would work, if I can pass it a `Value`.

I'm open to explore this, but I don't want to make a mess of the utility file. I'm not sure what the cleanest method for creating a `Value` is here, one which only encompasses the selection. Suggestions?

https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/clone-fragment.js#L19

Slapbox:
@ianstormtaylor I took your advice and solved the problem very easily that way once I understood how to use `Value.create()`

@zhujinxuan This is what the output for tables looks like in plaintext:

image

packages/slate-react/src/utils/clone-fragment.js
```diff
@@ -82,6 +84,16 @@ function cloneFragment(event, value, fragment = value.fragment) {
 
   attach.setAttribute('data-slate-fragment', encoded)
 
+  //  Creates value from only the selected blocks
+  //  Then gets plaintext for clipboard with proper linebreaks for BLOCK elements
+  //  Via Plain serializer
+  const valFromSelection = Value.create({
```
isubasti: I believe you can do `Value.create({document: fragment})` instead as fragment is a slate's document

Slapbox: This is accurate. I'll make that change now. Thank you

zhujinxuan:
@Slapbox 
What will happen for blocks like
```
<table-cell><paragraph> Text </paragraph></table-cell>
```
and 
```
<list-item><paragraph> Text </paragraph></list-item>
```

`slate-edit-table` and `slate-edit-list` wrap the children in blocks as well.  I am not sure whether it will render an extra `\n` in this case.

Slapbox:
@zhujinxuan I only tested this by manually wrapping table cells with `p` tags, but it seems to make no difference in the output. 

Even if this test is inaccurate (I don't believe it is,) surely the output will be dramatically better than the existing method which has no linebreaks or even spaces between blocks.

ianstormtaylor:
Thanks @Slapbox!

Slapbox:
@ianstormtaylor thanks to you! :heart: Slate

PR #2046, opened 2018-08-07
Fixes inability to Ctrl+Backspace at line start

Slapbox:
I don't know why my previous commits are also shown as changes, since these commits have been merged. Perhaps someone can shine some light on that for me?

### DIFF
--- a/packages/slate-react/src/utils/clone-fragment.js
+++ b/packages/slate-react/src/utils/clone-fragment.js
@@ -1,4 +1,6 @@
 import Base64 from 'slate-base64-serializer'
+import Plain from 'slate-plain-serializer'
+import { Value } from 'slate'
 import TRANSFER_TYPES from '../constants/transfer-types'
 import getWindow from 'get-window'
 import findDOMNode from './find-dom-node'
@@ -82,16 +84,15 @@ function cloneFragment(event, value, fragment = value.fragment) {
 
   attach.setAttribute('data-slate-fragment', encoded)
   
-  //  Builds plaintext object for clipboard line by line using top-level nodes.
-  //  Adds '\u000A' (unicode LF) to end of each line except the last.
-  //  This prevents line breaks from being lost when pasting to targets
-  //  which don't support html input.
-  let plainText = ''
-
-  for (let i = 0; i < contents.childNodes.length; i++) {
-    plainText += contents.childNodes[i].innerText
-    if (i < contents.childNodes.length - 1) plainText += '\u000A'
-  }
+  //  Creates value from only the selected blocks
+  //  Then gets plaintext for clipboard with proper linebreaks for BLOCK elements
+  //  Via Plain serializer
+  const valFromSelection = Value.create({
+    document: {
+      nodes: value.blocks,
+    },
+  })
+  const plainText = Plain.serialize(valFromSelection)
 
   // Add the phony content to a div element. This is needed to copy the
   // contents into the html clipboard register.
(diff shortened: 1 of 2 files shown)


zhujinxuan:
Hi, I think ctrl+backspace is not handled by slate with slate-hotkeys.

See https://github.com/ianstormtaylor/slate/pull/1888

Slapbox:
Ah well it works fine everywhere in Slate except here. This should be a painless way to enable proper functionality at line-starts. 

I think this is also the most proper way to do it, as it's entirely independent of the hotkey used and relies on whatever hotkeys Slate says to call `deleteWordBackwardAtRange`

zhujinxuan:
About previous changes, 
1. Use other branch submit PR rather than master
2. synchronize your master with slate's master
3. rebase to master

Slapbox:
Found the hotkey in slate-hotkeys: https://github.com/ianstormtaylor/slate/blob/master/packages/slate-hotkeys/src/index.js#L47

zhujinxuan:
Can you rename it to `Fixes inability to deleteWordBackward at line start`

Slapbox:
@ianstormtaylor is this PR okay with you, specifically the fact that it shows some inaccuracies in the changes? If this is an issue I can re-PR this.

ianstormtaylor:
@Slapbox I think if you pull the latest `master` on your fork, and then switch to this PR's branch and do `git merge master`, and push that up, it'll fix the diff. (I'm not worried about the extra commits showing up, cuz I'll squash it when I merge.) That should help.

But then I think this will require adding some tests, because I'm not sure exactly what the behavior will result in, or what it's fixing. 

Do you also have a copy of the stack trace for the error? That might help me get my head around it. Thanks!

Slapbox:
@ianstormtaylor whoops I had meant to reference an issue.  It seems this issue might not be submitted yet, or perhaps I just cannot find it. 

There's no errors. Just a lack of expected functionality. When deleting by _word_ instead of by _character_ the editor cannot currently continue up to the previous block.

Here's a video of the current behavior when using `Ctrl+Backspace` (deleteWordBackward on Windows) http://recordit.co/PzOmuW6las

After studying the code that surrounds this I don't believe there will be any additional tests necessary, though I am admittedly less familiar with the test plan used in Slate.

Basically the current behavior checks for remaining words to delete on the current line. If there are no more words, it sends an `n` of `0` to the `deleteBackwardAtRange`, which immediately aborts the function. This PR modifies the `deleteWordBackward` function to return a minimum `n` of `1` for the purpose of executing a normal, single character backwards deletion, instead of doing nothing.

This should have no potential to break anything as far as I can tell. The only edge case I could see was the possibility of deleting backwards at the beginning of the document when there is no additional block to continue to, but this won't be an issue due to the existing check here: https://github.com/ianstormtaylor/slate/blob/abc647463599b43e5550bb60dbd47ea7116a730d/packages/slate/src/changes/at-range.js#L369

ianstormtaylor:
> After studying the code that surrounds this I don't believe there will be any additional tests necessary, though I am admittedly less familiar with the test plan used in Slate.

What do you mean? This should have either change the behavior of an existing test, or if there is no existing test then it should include one with the proposed effect tested.

It would be useful to have a GIF of what the behavior is when using the shortcut on another editor, like Google Docs, to test against what the "expected" should be. Otherwise it's hard to tell.

Slapbox:
@ianstormtaylor 

I had thought this would be covered under another test case (deleteBackwardAtRange) - but this function is not tested directly as far as I can see, so I've thrown together a test (I just copied it from https://github.com/ianstormtaylor/slate/blob/master/packages/slate/test/changes/at-current-range/delete-backward/join-blocks.js)

Here's a video of the expected behavior on Google Docs: http://recordit.co/EAXNcszolz

The newly added test and the video below of Google Docs `deleteWordBackwards` behavior should hopefully clear things up. If there are additional tests you'd like to see, let me know.

PR #2052, opened 2018-08-08
Do not use requestAnimation if DataTransfer is available

zhujinxuan:

feature

#### What's the new behavior?

We are using dataTransfer for chrome, firefox and other modern browsers in clone-fragment.  Therefore, for these browsers, we do not need to call editor.change with `window.requestAnimation`.  

Calling `window.requestAnimation` sometimes makes the code hard to debug, and we cannot ensure there is no another React state change between `window.requestAnimation` in React v16 (as React.requestIdleCallback + queue are very unpredictable)

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -6,8 +6,8 @@ import React from 'react'
 import getWindow from 'get-window'
 import { Block, Inline, Text } from 'slate'
 import Hotkeys from 'slate-hotkeys'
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Good call @zhujinxuan, I like this. Could you change it so that `cloneFragment` takes a last `callback` argument instead of returning it?

zhujinxuan:
@ianstormtaylor Fixed

ianstormtaylor:
Awesome, thanks @zhujinxuan!

PR #2053, opened 2018-08-08
fix pointForward and pointBackward

### LINKED ISSUE #2036: Cannot read property 'key' of undefined
It happens when the first word in paragraph is link. 

1. select link
2. press right arrow
3. it crashes

screen shot 2018-08-06 at 02 49 30

screen shot 2018-08-06 at 02 49 42

### LINKED ISSUE #2050: Bugs related to migration to point system
**Issues:**
- Left arrow from empty line selects final character of preceding line (No error)
- Right arrow from empty line does not move to following block ([Error](https://gist.github.com/Slapbox/6b9dbfec422b1cfb5ee8073417dce252))

Maybe an off by 1 error somewhere?

**Demo:** http://recordit.co/PPDalpUVHO

bryanph:
Fixing a bug

#### What's the new behavior?
For pointForward, existance of next was not checked. Also, the method is called for both focus and anchor; updated the code to reflect this.

The same implementation as pointBackward()

#### Issues
Fixes #2036 
Fixes #2050

### DIFF
--- a/packages/slate/src/changes/on-selection.js
+++ b/packages/slate/src/changes/on-selection.js
@@ -662,6 +662,7 @@ function pointBackward(change, point, n = 1) {
   const p = selection[point]
   const isInVoid = document.hasVoidParent(p.path)
 
+  // what is this?
   if (!isInVoid && p.offset - n >= 0) {
     const range = selection[`move${Point}Backward`](n)
     change.select(range)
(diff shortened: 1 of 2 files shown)


Slapbox:
This works. I don't understand how it fixes the issue with pointForward, but it does. Barring any issues anyone sees, I'd love to see this published ASAP. The bug in the current version is pretty substantial, though by no means a showstopper.

bryanph:
Might want to add some tests so this doesn't occur again.
@Slapbox the problem in pointForward was simply a variable being referenced before it was defined. Afterwards I also had to change the methods being called for moving the selection; the complete selection was moved instead of the anchor and focus separately causing the selection to move twice.
@ianstormtaylor I add some comments to parts of code that I don't understand and I feel like can be clarified a bit with a comment indicating what it is for, is this ok?

ianstormtaylor:
Good catch, thanks @bryanph! I agree with you about those comments too, I'll leave them there to fill in when I get a chance.

PR #2054, opened 2018-08-08
Update Contributing.md

bryanph:
Add `yarn bootstrap` to the docs as a prerequisite for running tests


ianstormtaylor:
@bryanph I think this shouldn't be required, what Yarn version are you using?

bryanph:
@ianstormtaylor you're right I was running an earlier version of Yarn when installing, thanks!

PR #2056, opened 2018-08-08
Use run-p instead of npm-run-all --parallel

zhujinxuan:

feature

#### What's the new behavior?

`run-p` is alias of `npm-run-all --parallel` provided in `npm-run-all`.  Perhaps `run-p` is a bit better here because it is shorter

Fixes: #
Reviewers: @


ianstormtaylor:
Hey @zhujinxuan thanks for this, but I think the visible `--parallel` flag is more clear to people coming in, and the `npm-run-all` name matching the dependency is also helpful.

PR #2057, opened 2018-08-08
docs(plugins): update signature docs

ericedem:
Just a simple doc change: `onChange()` also receives the [editor as an argument](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/before.js#L106-L124).


ianstormtaylor:
Thanks @ericedem!

ericedem:
@ianstormtaylor Looks like this didn't quite make it in. Anything I need to do on my end? Thanks!

Issue #2043, opened 2018-08-07
deprecate `deleteCharBackward/Forward`


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now there are both `deleteBackward/Forward` and `deleteCharBackward/Forward`. These `*Char*` variants should be removed because the `deleteBackward/Forward` changes should move by one character at a time anyways.

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/3093.

PR #2073, opened 2018-08-11
[changes] deprecate deleteCharBackward and forward APIs

davidchang:

An Improvement, by deprecating extra APIs

#### What's the new behavior?

Removing `deleteCharBackward` and `deleteCharForward` in their entirety, since users can just use `deleteBackward` and `deleteForward` to achieve the same behavior.

Wasn't sure if the issue that was opened was wanting to remove in their entirety, or just add a `logger.deprecate` into the functions. But if it's a major, breaking change, with a simple migration path (find/replace), I think it's okay just to remove in its entirety.

Fixes: #2043
Reviewers: @ianstormtaylor


ianstormtaylor:
Hey @davidchang thanks for taking this on! We'll actually want to add deprecations, to make migrating to the new version easier for people. And we'd want to preserve the existing tests to ensure that the current behaviors of the `*Char*` methods is preserved.

I think the one issue is that the `*Char*` methods are the only place that it was looking at the actual text to determine the distance to delete (emojis have length greater than `1` sometimes), so we'd need that logic to be preserved in `deleteBackward/Forward` still.

ianstormtaylor:

Issue #1996, opened 2018-07-25
rename hyperscript `decorators` option to `decorations`


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Debt.

#### What's the expected behavior?

Right now the option is called `decorators`. But this doesn't match with the noun being "decoration" everywhere else. It should be `decorations` instead.

PR #2074, opened 2018-08-11
[hyperscript] rename decorators option in params to decorations

davidchang:

Debt.

#### What's the new behavior?

`createHyperscript` takes an options object that now looks for the `decorations` property instead of the `decorators` property

Pretty straightforward rename of an option, then all of its uses and wherever the term decorators showed up in docs

Fixes: #1996
Reviewers: @ianstormtaylor

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -9,12 +9,12 @@
 <br/>
 
 <p align="center">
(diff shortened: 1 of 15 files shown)


ianstormtaylor:
Thanks @davidchang this is great!

Issue #1757, opened 2018-04-04
Issues in IE 11


ikorgik:
Hi, we found several issues in IE 11 during slate integration:

```
Objects are not valid as a React child (found: object with keys {size, _origin, _capacity, _level, _root, _tail, ownerID, hash, __altered}). If you meant to render a collection of children, use an array instead. in span (at text.js:130)
```

```
Unable to get property 'childNodes' of undefined or null reference
```

```
Objects are not valid as a React child (found: List [ [object Object] ]). If you meant to render a collection of children, use an array instead.
```
Attached a pull request with changes that work well to me: 
https://github.com/ianstormtaylor/slate/pull/1756

Slapbox:
I can't even load http://slatejs.org/ in IE11.

**Error:**
```
SCRIPT5009: 'Symbol' is undefined
File: main-01e8c59ea6c6fde8ea58.js, Line: 6, Column: 81245
```

renchap:
This is because the official website is not transpiled for IE11 compat nor includes the needed polyfills (like Symbol)

sethmcleod:
On a related note, I just opened a PR for adding the polyfills needed to support IE11 on the example site: https://github.com/ianstormtaylor/slate/pull/1782

ianstormtaylor:
Hey, thanks for using Slate! Unfortunately, we can't offer support for usage questions in the issues here because it becomes overwhelming to maintain the project if the issues are filled with questions. 

However, we do have a Slack channel and people are constantly asking and answering questions in there. So I'm going to close this issue, but I definitely recommend joining the Slack channel if you want to find people that might be able to help. 

Thanks for understanding!

PR #2087, opened 2018-08-16
IE11 html serializer parser fix

danielsarin:

This should fix `Unable to get property 'childNodes' of undefined or null reference` error in IE11 when using slate-html-serializer and deserializing an empty string (#1757, #1756).

#### What's the new behavior?

The `defaultParseHtml` now returns an empty body element if the parsed body is null. 

I tested this in the latest Chrome, Firefox and Safari, and they all seemed to work similarly. If the html param was an empty string, `parseFromString` returned a document element, which had a body property which was an empty body element. In IE11 though the body property was null, which caused the undefined or null reference later.


ianstormtaylor:
Thanks @danielsarin, great catch!

Issue #1850, opened 2018-05-18
Drag dropping a node between branches crashes


alaindresse:
#### Do you want to request a _feature_ or report a _bug_?

A bug

#### What's the current behavior?

Dragging a node between two editors generates the error "Could not find a node with id ...".

See https://codesandbox.io/s/q9lyy6k5mq for a codesandbox reproducing the issue. See also the following gif illustrating the issue.

I first show drag drop works within same editor, then I try to drag the image node between editors.

![slateissue](https://user-images.githubusercontent.com/1415531/40233525-a5b4f50e-5aa3-11e8-94b5-49181518e0a2.gif)
 

#### What's the expected behavior?

Dropped node should be copied to the new editor.

bryanph:
I know draft.js solves this by having a unique key for every draft editor and checking onDrop whether the keys match (in which case the block should be moved).

As I see it there are these two cases:
1. Drag within editor: slate-editor-key is equal => move_node
2. Drag between editors: slate-editor-key is different => insert_node at the target (clone the node at the target)

PR #2090, opened 2018-08-17
use fragment instead of node transfer type on drag

bryanph:
Fixing a bug

#### What's the new behavior?
Creates a fragment EventTransfer instead of a node EventTransfer object. 

Previously only the direct node that was dragged was stored in the EventTransfer, now the complete fragment is, allowing for dragging of a selection with multiple blocks.

Fixes: #1850
Reviewers: @

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -266,9 +266,11 @@ function AfterPlugin() {
     isDraggingInternally = true
 
     const { value } = change
-    const { document } = value
+    const { document, schema } = value
     const node = findNode(event.target, value)
-    const isVoid = node && (node.isVoid || document.hasVoidParent(node.key))
+    const ancestors = document.getAncestors(node.key)
+    const isVoid =
+      node && (schema.isVoid(node) || ancestors.some(a => schema.isVoid(a)))
     const selectionIncludesNode = value.blocks.some(
       block => block.key === node.key
     )
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -4,7 +4,7 @@ import Plain from 'slate-plain-serializer'
 import { IS_IOS } from 'slate-dev-environment'
 import React from 'react'
 import getWindow from 'get-window'
-import { Block, Inline, Text } from 'slate'
+import { Text } from 'slate'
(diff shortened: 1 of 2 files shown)


bryanph:
This actually causes local drags to duplicate nodes because this code here doesn't actually do anything: https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/after.js#L317

ianstormtaylor:
@bryanph thanks for looking into this! One of the reasons we use `node` instead of `fragment` is for the individual void node dragging case, where you want to move the node inside the editor. It looks like this PR currently changes that behavior. (Try dragging the single image in the `images` example to see what I mean.)

bryanph:
@ianstormtaylor this distinction causes problems when dragging to another editor. I think it is best if we just rely on fragments because then the behaviour for dragging within the editor is the same as dragging to another one. The last commit I added handles the case where just the void node is dragged; it selects the void node if it isn't in the selection already.

This behaviour fixes two bugs: #1850 and an unreported one where a selection of multiple blocks can't be drag+dropped if the drag is initiated from the void block.

packages/slate-react/src/plugins/after.js
```diff
@@ -348,12 +350,14 @@ function AfterPlugin() {
       change.insertFragment(fragment)
     }
 
+    // todo: this never gets called anymore because we just work with fragments
```
ianstormtaylor: Can you remove these blocks since we no longer handle fragments?

packages/slate-react/src/plugins/after.js
```diff
@@ -267,15 +267,17 @@ function AfterPlugin() {
...
+    )
 
-    if (isVoid) {
-      const encoded = Base64.serializeNode(node, { preserveKeys: true })
-      setEventTransfer(event, 'node', encoded)
-    } else {
-      const { fragment } = value
-      const encoded = Base64.serializeNode(fragment)
-      setEventTransfer(event, 'fragment', encoded)
+    if (isVoid && !selectionIncludesNode) {
```
ianstormtaylor: Can you add a comment to explain what's going on here, why this case would ever come up?

bryanph:
@ianstormtaylor done, i also aligned it with the new way of handling void nodes.

ianstormtaylor:
Thanks @bryanph, this is great!

PR #2091, opened 2018-08-17
fix images example schema

bryanph:
bugfix

use new schema changes in the images example

Fixes: #
Reviewers: @


ianstormtaylor:
Oops, thanks for catching that @bryanph!

PR #2100, opened 2018-08-20
Refactor the deprecation of Range

zhujinxuan:

feature

1. Use generated refactor for deprecation interface.  Therefore, the class definition is stayed clean
2. Put deprecation interface in a separate file.  Then exclude that file in code coverage analysis

Fixes: #
Reviewers: @


zhujinxuan:
I think we do the same thing in Value in another PR.

ianstormtaylor:
@zhujinxuan why is this needed? It seems like it's going to make maintaining these things much more complicated. Soon we can just remove the deprecations anyways if they are a big issue?

zhujinxuan:
@ianstormtaylor This is for a better coverage estimate in codecov.

Our current `Range` and `Value` includes a lot of deprecated functions that we do not want to test, and that gives a falsified estimation of coverage percentage.  And we have about 500 lines of code for deprecated interface, while the core function is only about 800 lines.  Almost half of the file is for the deprecated interface.  Somehow it make people read and find available range methods.

Therefore I think perhaps it is better to keep all deprecation interface in a separate file, and exclude that file from the code coverage analysis.  Then the range.js includes mostly only the currently maintained interface. 

I think most complexity of current PR is the alias generator.  I think I can remove these generators and replace with the original definition.  I will submit a new commit on that this evening.

ianstormtaylor:
@zhujinxuan this is not something I want to do. I don't want to make the codebase harder to maintain just to fix coverage reporting accuracy.

These deprecations live in the file right now where people would expect to find them, and they will be removed in the future. Because of how they're currently structured, the removal will be a very simple diff. When they are removed we'll get more accurate code coverage, but I don't think it's worth adding extra complexity to get that sooner.

zhujinxuan:
@ianstormtaylor I agree.  Let us close this PR.

Issue #1969, opened 2018-07-13
Clicking on a zero-width node and dragging does not select text


thallada:
#### What's the current behavior?

Clicking on an empty paragraph or next to a inline link (both have `data-slate-zero-width` nodes) and then dragging to select text only moves the cursor and does not select text.

![slate_bug](https://user-images.githubusercontent.com/1505923/42695277-db8bb036-8682-11e8-8e00-490369436375.gif)

#### What's the expected behavior?

I should be able to click and drag anywhere on the document to select text.

willbryk720:
I have the same problem with the cursor. Sometimes it just doesn't highlight. It happens randomly about every 1 in 4 attempts to highlight. For me, it's not just inline nodes, it happens anywhere. It more often happens when I start highlighting from within an already highlighted section. So if I place my cursor inside a highlighted portion and try to highlight, it will not highlight.

![strangecursor](https://user-images.githubusercontent.com/16409551/43867306-c95deab4-9b36-11e8-89df-e6ddf6b34306.gif)

zhujinxuan:
Is this problem reproducible right now?  It seems I cannot reproduce it in current demo.

willbryk720:
Interestingly, after investigating the problem more deeply, it seems the cause of my problem was completely different from @thallada. 

I wrap my editor so that I can access its value from a parent component. The onChange function passed to the editor called a function that passed the editor value to the parent component. Every time the editor changed, it passed it to the parent component. I tested out the same editor without any passing to parent function, and the editor stopped having this problem. This bad cursor behavior then must be due to the parent component re-rendering so often.

For anyone who has a similar problem, instead of passing the value to the parent on every onChange, you can just pass the value to the parent only when you need it.

ianstormtaylor:
@willbryk720 thank you! That is actually super helpful to figuring out the original issue too.

I think it might be the case that the original issue is caused by re-rendering too, where clicking a zero-width space happens to result in `updateSelection` being triggered, cancelling the draggability. Which means we might be able to solve it without needing to listen to `mousemove` events? cc @zhujinxuan

zhujinxuan:
@ianstormtaylor I think the re-rendering is perhaps the reason.  We shall not have the re-rendering just after selection change.  

I will have a try whether udpateSelection in onChange and change `updateSelection` works

zhujinxuan:
[...]

PR #2104, opened 2018-08-21
[WIP] Add onMouseMove method; Fix incorrect selection when click and drag move

zhujinxuan:

bug

#### What's the new behavior?

If user click and drag, and an selection event is fired during dragging.  Then we shall expect an extended range instead of an collapsed range.  When the browser accidentally causes an collapsed range, then we shall extend it to the new selection instead of select.

Fixes: https://github.com/ianstormtaylor/slate/issues/1969
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -273,7 +273,9 @@ class Content extends React.Component {
    */
 
   onEvent(handler, event) {
-    debug('onEvent', handler)
+    if (handler !== 'onMouseMove') {
+      debug('onEvent', handler)
+    }
 
     // Ignore `onBlur`, `onFocus` and `onSelect` events generated
     // programmatically while updating selection.
--- a/packages/slate-react/src/constants/event-handlers.js
+++ b/packages/slate-react/src/constants/event-handlers.js
@@ -22,6 +22,7 @@ const EVENT_HANDLERS = [
   'onDrop',
   'onInput',
   'onFocus',
+  'onMouseMove',
   'onKeyDown',
   'onKeyUp',
   'onPaste',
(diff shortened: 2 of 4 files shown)


ianstormtaylor:
Hey @zhujinxuan, this sounds like a good call. Can you explain exactly what's going on here in terms of the events that are firing in the browser? I'm not sure I understand.

zhujinxuan:
@ianstormtaylor Sometimes, when you press and drag from a empty block start.  The browser will fire onSelection events with collapsed selection, rather than selection extended from the mouse down selection.  (https://github.com/ianstormtaylor/slate/issues/1969)

This PR tries to fix the problem that, for any `onSelection` event during onMouseMove, if the left key is pressed when mouse moves (click and drag) and if the previous selection and current selection is both collapsed,  we identify that the bug occurs.  Then we fix the selection with extending between  previous and current selection.

zhujinxuan:
@ianstormtaylor I can only reproduce it in the slate.  I can only reproduce https://github.com/ianstormtaylor/slate/issues/1969#issue-341029949 but unable to reproduce https://github.com/ianstormtaylor/slate/issues/1969#issuecomment-411570384.  (only able to reproduce the zero-width condition but not the second one). 

Can we just change this PR to add onMouseMove handlers (without adding the timer to the core)? 
 Then a temporary solution can be provided as a plugin.

zhujinxuan:
@ianstormtaylor BTW, the timer is initialized only on (pressed & mouse move).  Because the setTimeout is called only when (event.buttons % 2 == 1) (which means left mouse button is pressed)

But I agree the timer solution is an ugly solution.

zhujinxuan:
Will use another PR to fix the problem

Issue #1922, opened 2018-06-20
remove the restriction of "no empty inlines"


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now there is a core schema rule that enforces that inline nodes must have text that is not an empty string, or they will be removed. But this is not ideal, it's not great to force this on users, since it's fairly arbitrary.

There are valid use cases for wanting empty inlines.

The main reason this was done was to eliminate the possibility of having "leftover" inlines in the document if a user has deleted the last character of link (or similar). But this was added in a much older iteration of Slate, before the schema was solidified.

#### What's the expected behavior?

These days, I think we can achieve the same goal in userland with a fairly simple schema rule with a regex for `text`, like:

```js
const schema = {
  inlines: {
    link: {
      text: /.+/,
    }
  }
}
```

Which by default will be removed, or in edge cases can be customized.

zhujinxuan:
> There are valid use cases for wanting empty inlines.

Can I know the user case of wanting empty inline without `isVoid`?

ianstormtaylor:
There was someone in chat wanting to essentially use inlines like a `<input />` element behaves, but they always start with an empty value. (Although they may have been wanting `isVoid`, I'm not sure.)

Either way though I think it would be nice if possible if core was unrestrictive about this.

nkoder:
Hi 👋 

I think it was me who asked about requirement for non-empty `text`. On the one hand [documentation](https://docs.slatejs.org/guides/data-model#documents-and-nodes) says:

> Inlines can't contain no text. Any inline node whose text is an empty string ('') will be automatically removed. This makes sense when you think about a user backspacing through an inline. When they delete that last character, they'd expect the inline to be removed. And when there are no characters, you can't really put your selection into the inline anymore. So Slate removes them from the document automatically, to simplify things.

On the other hand I have code like this one:
```
const EditableFieldNode = {
  nodeType: NODE_TYPE,
  createNode: ({ name, placeholder }) => ({
    object: "inline",
    type: NODE_TYPE,
    data: {
      name,
      placeholder,
    },
    nodes: [
      {
        object: "text",
        leaves: [
          {
            // TODO PoC: what to put here? It's not visible, but required to be non-empty
            text: name,
          },
        ],
      },
    ],
  }),
  plugins: [renderPlugin()],
};
```

I can do it like that:
```

          {
[...]

PR #2105, opened 2018-08-21
Remove is void

ianstormtaylor:

Improvement.

#### What's the new behavior?

This removes the `isVoid` property from `Node` models, and changes the behavior to be driven by the schema instead.

This is the second stage to https://github.com/ianstormtaylor/slate/pull/2102.

While accessing the `node.isVoid` property was deprecated in the first stage, its value was still being inferred from the `isVoid` boolean passed into nodes when they were created. However, with this pull request the `isVoid` property of nodes is no longer used when determining a nodes "void" treatment. Now that behavior is determined by checking the editor's `schema`.

Fixes https://github.com/ianstormtaylor/slate/issues/1922


ianstormtaylor:
Superseded by https://github.com/ianstormtaylor/slate/pull/2113.

PR #2114, opened 2018-08-22
Enable event logging in development mode

zhujinxuan:

featrue

#### What's the new behavior?
When localestorage.ENABLE_EVENT_LOG is true, `event.persist()` will be called to understand {event} in `debug('onEvent', {event})`

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -6,7 +6,7 @@ import React from 'react'
 import getWindow from 'get-window'
 import { Text } from 'slate'
 import Hotkeys from 'slate-hotkeys'
-
+import eventLogger from './event-logger'
 import Content from '../components/content'
 import cloneFragment from '../utils/clone-fragment'
 import findDOMNode from '../utils/find-dom-node'
@@ -23,7 +23,7 @@ import setEventTransfer from '../utils/set-event-transfer'
  * @type {Function}
  */
 
-const debug = Debug('slate:after')
+const debug = eventLogger(Debug('slate:after'))
 
 /**
  * The after plugin.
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
@zhujinxuan I don't really want to add a separate logger just for the events here. It feels like it adds complexity for not much gain. What are you trying to access on the events in the debugger? Maybe we can mask the properties themselves instead.

zhujinxuan:
Hi, @ianstormtaylor 

There are two reasons for trying to access events in console:
1. currently we have `debug('onEvent', {event})` in before.js and after.js.  But all events attributes are not logged. If we read event properties in console, then we only have a lot of null.
2. On development, we may need to track a series of events to understand the browser behavior.  For example, in IME problem, it is good to read the `event.data` of `onCompositionEnd`, `onBeforeInput` and `onInput` in console. 

Can I know what you mean by `mask the properties`?  It seems finding all meaningful properties is much harder than calling `evnet.persist()`

zhujinxuan:
For example, before this plugin, if we want to read a `onKeyDown` event in console:
screen shot 2018-08-22 at 9 08 45 pm
All attributes are null.

After this plugin, with `localStorage.ENABLE_EVENT_LOG` we can:
screen shot 2018-08-22 at 9 10 38 pm

ianstormtaylor:
@zhujinxuan very fair. Let's go with always persisting in development mode then, instead of adding an extra local storage flag. And then if that becomes an issue we can figure out something else. I just want to avoid having lots of little flags that are hard to document and use for people.

zhujinxuan:
@ianstormtaylor Fixed in 7b59eb2

Issue #2097, opened 2018-08-19
Slate-React - Unable to focus editor programmatically


Slapbox:
This is going to be a vague bug report for now as I can't track down exactly how to reproduce this. Programmatic refocus works in all ways in my app except one circumstance.

This bug first appears in Slate-React 0.15.6 and remains in 0.15.9.

Slapbox:
I just spent 2 hours tracking down this bug to 0.15.6, then when I rolled back to 0.15.6 again, it seems to work... Will continue to update.

Slapbox:
**It's 0.15.8** I had accidentally used `^0.15.6` for my version, which led to me getting `0.15.9`, which has the same problem from `0.15.8`

zhujinxuan:
Can you make a code example?  I remember there are some conflicts between react update process and focus.  Perhaps you need `window.requestAnimation` (only works before React v16.3) or use componentDidUpdate to do `editor.focus()`

Slapbox:
@zhujinxuan I've tried both passing in a `change` object with a selection with `isFocused: true` as well as using the `myEditor.focus()`

I've tried using `forceUpdate` but no dice there either. The issue is 100% related to `0.15.8` and I tried everything under the sun, including `window.requestAnimationFrame` and `setTimeout`.

bryanph:
I am getting the same issue, It seems to happen quite randomly; something I can focus and something I can't. Haven't found the cause yet.

Slapbox:
@bryanph what version are you using?

I've yet to try versions later than 0.15.9.

I am 100% certain 0.15.8 is where the problem begins for me. I've got no idea how to go about reproducing the issue.

Just a note that 0.15.8 was released August 16th, so we can hopefully find the commit that this problem originates with.

Slapbox:
The only commit between 0.15.7 and 0.15.8 is this one, so the issue must originate here: https://github.com/ianstormtaylor/slate/pull/1975

Slapbox:
I'm trying to understand what exactly these changes do, but I'm afraid I'm a bit too new to the internals of Slate to really grasp the finer points here. @zhujinxuan is there anything in that commit you can identify as a possible cause?

bryanph:
@Slapbox the issue has something to do with https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/changes/on-selection.js#L17-L19 this method not being called. When I add a change.focus() in the before onFocus plugin, it works fine.

[...]

PR #2130, opened 2018-08-26
handle focus case where .focus() is called directly on the DOM Node

bryanph:
Fixing a bug

#### What's the new behavior?
.focus() being called on the DOM Node did not update the selection isFocused state, now it does because another call is done to change.focus().

Calling change.focus() in onFocus in plugins/before.js

Fixes: #2097
Reviewers: @

### DIFF
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -354,6 +354,9 @@ function BeforePlugin() {
       return true
     }
 
+    // This handles the case where .focus() was called directly on the DOM Node instead of via the editor
+    change.focus();
+
     debug('onFocus', { event })
   }
 


zhujinxuan:
Why it is in before.js instead of after.js?

bryanph:
@zhujinxuan no reason really, wasn't sure where to put it.

I actually find that I am still getting some issues when i am using the "smooth" option, these become apparent. Calling change.focus() multiple times seems to interrupt the animation in chrome (not in firefox). I should probably set up a panel example, which shows multiple panels with editors and shows the focus between them. I'll see if I can find some time to set that up this week.

quentez:
I've tested that solution and I'm not sure it fully fixes the problem:

1. Il solves the problem where multiple clicks are required to focus an editor, but when focused for the first time marks can't be applied before some content is added.
2. After the editor is focused/blurred programmatically multiple times, it randomly stops workings.

bryanph:
@quentez you're right, this didn't fix the underlying issue. Apparently this got fixed now, see #2097. No need to merge this anyway!

Issue #2125, opened 2018-08-24
Enter deletes paragraph and replaces with new one in iOS Safari


ericedem:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Hitting Enter in iOS Safari deletes the current paragraph and replaces it with an empty one, instead of starting a new paragraph.

Try any of the examples: https://www.slatejs.org/#/rich-text

![kapture 2018-08-24 at 15 36 44](https://user-images.githubusercontent.com/1060669/44610950-92e0f100-a7b3-11e8-9338-d5e5ef197fbb.gif)

Slate version: 0.40.2
Can reproduce on a real iPhone 7 v11.4.1
And in simulator iPhone X (pictured)

#### What's the expected behavior?

It should create a new paragraph.

Happy to contribute a fix if someone can point me in the right direction.

janjiss:
Not sure if this is helpful, but after some investigation, I've found that when you remove `!IS_IOS` check from the following line https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/after.js#L464, it works as expected.

dmitrizzle:
Does this still happen in slate 0.39 for you?

ericedem:
I upgraded from `0.36.2` where it was working. It is broken in `0.37.7`.

I ran a `git bisect` and it appears that the code broke in this commit, though I don't see anything directly related to iOS behavior: https://github.com/ianstormtaylor/slate/commit/877dea16bf11271bf3232ccfb20cc1f0147834a4

dmitrizzle:
@zhujinxuan any ideas on what may have gone wrong here? I really am clueless.

zhujinxuan:
@dmitrizzle No idea.  I plan to work on desktop IMEs, but I have no experience about mobile part.

ericedem:
@janjiss It looks like the `!IS_IOS` checks still need to be in there due to autocomplete issues on iOS. Described in #1471 and fixed in #1475.

Playing around with this a little bit more in the rich-text example, this bug seems a little more nuanced:

* Hitting 'Enter' in unmarked text deletes backward until hitting a mark.
* Hitting 'Enter' in marked text deletes backward until hitting unmarked text.
* Hitting 'Enter' exactly at the boundary of a mark and unmarked text, the line will be split.

ericedem:
I did some debugging in the content editable component, and here are the events we are getting:

[...]

Issue #2129, opened 2018-08-25
Bug: Backspacing in empty line on iOS causes an error


janjiss:
#### What's the current behavior?

Pressing backspace on iOS Chrome and Safari an empty line causes an error.

![aug-25-2018 21-59-12](https://user-images.githubusercontent.com/600203/44621642-226cc980-a8b2-11e8-93e9-680478453a2e.gif)

```
[Error] Error: Unable to find a DOM node for "10". This is often because of forgetting to add `props.attributes` to a custom component.
```

#### What's the expected behavior?

The expected behavior is not to error :) 

#### What caused it

Removing `!IS_IOS` check from [here](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/before.js#L398) and [here](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/after.js#L470) seems to fix the issue, however, I'm not sure what kind of consequences this entails. 

P.S - It seems that those `!IS_IOS` checks don't serve their purpose and just cause havoc on the behavior of the editor on the iOS. See similar issue https://github.com/ianstormtaylor/slate/issues/2125

PR #2131, opened 2018-08-27
fix(content): use handlers that actually exist in event listeners

ericedem:

Fixing a _bug_

In a recent change, handlers now only exist in `Content#handlers` and not directly on the instance itself. There were still a couple of places where they were being accessed directly from the instance for event listeners. This was causing some side effects in iOS.

More details in #2125 

Fixes: #2125 #2129 
Reviewers: @zhujinxuan @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -113,12 +113,15 @@ class Content extends React.Component {
     if (window) {
       window.document.removeEventListener(
         'selectionchange',
-        this.onNativeSelectionChange
+        this.handlers.onNativeSelectionChange
       )
     }
 
     if (HAS_INPUT_EVENTS_LEVEL_2) {
-      this.element.removeEventListener('beforeinput', this.onBeforeInput)
+      this.element.removeEventListener(
+        'beforeinput',
+        this.handlers.onBeforeInput
+      )
     }
   }
 
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -91,13 +91,13 @@ class Content extends React.Component {
 
     window.document.addEventListener(
       'selectionchange',
-      this.onNativeSelectionChange
+      this.handlers.onNativeSelectionChange
     )
 
     // COMPAT: Restrict scope of `beforeinput` to clients that support the
     // Input Events Level 2 spec, since they are preventable events.
     if (HAS_INPUT_EVENTS_LEVEL_2) {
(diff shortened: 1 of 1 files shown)


zhujinxuan:
Hi, @ericedem  Can you also update the componentWillUnmout's removeEventListener in https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/content.js#L116

ericedem:
@zhujinxuan nice catch! Whoops.

It looks like this PR is blocked because the code coverage hasn't changed. With this patch, what is the best way to deal with this? I don't see any tests for `content.js`.

zhujinxuan:
@ericedem codecov does not block merging process.  We do not care for the codecov with slate-react right now because it is not able to be tested with jsdom.  Hopefully we may have slate-react tested with chrome headless next year.

isubasti:
The same reason seems to break safari on OSX as well. if you create a new block and start typing and try going back to the very first character using arrow keys, it has an extra cursor position. This PR seems to fix it

ianstormtaylor:
Thank you @ericedem for catching this! Sorry for the delay.

ericedem:
@ianstormtaylor No problem, thanks for maintaining! 🍻

zarv1k:
@ianstormtaylor Seems, this PR (slate-react@0.18.4) breaks focus in empty editor:
1. Select all and delete all content https://www.slatejs.org/#/rich-text
2. Blur editor (click outside the editor)
3. Try to add some text into the editor. The editor loses focus immediately.

There is no such issue in slate-react@0.18.3.

ericedem:
@zarv1k Thanks for reporting. Can you file a separate issue for it for tracking the issue? Something to check is whether you see the issue in 0.36.x, since this PR fixes a bug introduced in 0.37.x that may have covered it up.

zarv1k:
@ericedem, The issue I described above is related only to this PR, it is not related to slate package at all, it's related to slate-react package only. There is no issue in slate@0.40.2+slate-react@0.18.3, but as you can see it is reproduced on top versions of slate packages with slate@0.40.2+slate-react@0.18.4. Just try to reproduce using in my comment above).
I can create the separate issue in a few hours if needed.

zarv1k:
I've created the issue: https://github.com/ianstormtaylor/slate/issues/2147

briandk:
👍 

I second @zarv1k - I was dealing with this very bug this morning. I was just about to file an Issue when I saw they had already filed one: #2147 😇

Issue #560, opened 2017-01-17
Typescript definitions


frederickfogerty:
There are no typescript definitions for this project that I can find, either here or on DefinitelyTyped/@types. Does anyone have any definitions? 

It would be great if the definitions could be exported from the module directly, it makes adoption a lot easier, and keep the definitions close to the source. 

🙌🏻

ianstormtaylor:
For context, I'm open to adding type definitions to the library in the future, if we start using typed source code, or after it's very stable. But until one of those things happens I'm avoiding adding type definitions purely for the maintenance burden.

bsouthga:
As a stopgap, I created this https://github.com/CrossLead/slate-dts which generates the typings from the source code and publishes scoped typings packages (for example `@crosslead/slate-typings`)

styu:
@ianstormtaylor I was going to add type definitions to `DefinitelyTyped` and then I saw this issue - is this something you're open to now, or do you still want to wait?

andykent:
I started work on some handwritten typedefs this week (just `slate` and `slate-react` for now)...

https://gist.github.com/andykent/202e819c35bd720e73302849c76fab33

It's actually my first typedef for an external lib like this so I'm sure I've made some huge errors but it seems to be working for me. Also, wow, Slate maybe wasn't the easiest project to start with, there's a pretty large API and some quite rich types.

I've put them in a Gist for now in case they are useful to others as I think it's a good step forward from the autogenerated ones mentioned above but it would be great to get something in Slate itself or DefinitelyTyped if not. 

Note: I largely based the types on the public docs but noticed there is quite a lot more undocumented APIs in the source, @ianstormtaylor are things not in the public website docs to be considered private? or are the docs just lagging behind?

JanLoebel:
Thank you @andykent and @majelbstoat! I created two basic repositories to store the files and add some information how to use them in a current project to find other bugs and improve the definition further. Please feel free to create any Pull-Request or add issues. Thank you for the great work so far :)

https://github.com/JanLoebel/types-slate-react
https://github.com/JanLoebel/types-slate

chilbi:
[https://github.com/chilbi/slate-demo/tree/master/src/types](https://github.com/chilbi/slate-demo/tree/master/src/types)

mattcopas:
@chilbi Thanks for your link! 

[...]

PR #2144, opened 2018-08-31
[WIP] Typescript Typings

RXminuS:
There was a long discussion in https://github.com/ianstormtaylor/slate/issues/560 if we should include Typescript typings in the repo itself or not. The core team seamed geared towards keeping typings separate.

However, with recent big API changes the other attempts at external typings have almost fully broken. Personally I feel that the effort required to maintain a set of solid base typings while making a PR is much easier than a constant back and forth to an external repo. The auto-generated typings out there miss a lot of details (especially because of how changes are added to prototype classes)

So I've started to hand-type the entire library. My goal was to keep it as symbolically close to the source code as possible so that any contributers to the project should easily find what they are looking for. Sometimes it was actually super unclear from the source and docs what the code is supposed to do so I even think that types can help newcomers more easily figure out what is happening.

As a disclaimer, I'm very new to Typescript and am by no means an expert. Therefore there might be some patterns that can be polished a bit nicer, but am hoping for a bit of community help. My hope is that once these typings are done we can include them in the project. I'm even willing to commit myself to maintain them as part of PR's or donate some money for others to maintain them if it's really that big a hassle.

On a side-note: I'm also finding a few bugs because of the typings, I'll post fixes as separate PR's

- [x] Core Typings
- [ ] Packages
- [ ] Tests
- [ ] Build Pipeline

### DIFF
--- a/packages/slate/index.d.ts
+++ b/packages/slate/index.d.ts
@@ -1,0 +1,1384 @@
+declare module "slate" {
+    import * as Immutable from "immutable"
+  
+    type Partial<T> = { [P in keyof T]?: T[P] }
+  
+    type Primitive = undefined | null | boolean | string | number | Function
+  
+    type Immutable<T> = T extends Primitive
+      ? T
+      : T extends Array<infer U>
+        ? ReadonlyArray<U>
+        : T extends Map<infer K, infer V> ? ReadonlyMap<K, V> : Readonly<T>
+  
+    type DeepImmutable<T> = T extends Primitive
+      ? T
+      : T extends Array<infer U>
+        ? DeepImmutableArray<U>
+        : T extends Map<infer K, infer V> ? DeepImmutableMap<K, V> : DeepImmutableObject<T>
+  
+    interface DeepImmutableArray<T> extends ReadonlyArray<DeepImmutable<T>> {}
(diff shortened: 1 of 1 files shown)


lichenglu:
@RXminuS Would love to help if you want to take something off your plate!

zhujinxuan:
Hi, I would suggest not to document `getBlocksAsArray` and other `...AsArray` functions (or docunmenting them as private).  Because these functions are internal functions to speed up functions like `getBlock`

RXminuS:
Thank you @zhujinxuan I've fixed that. Also added a bunch of missing methods and used the https://github.com/piotrwitek/utility-types package to not re-invent the wheel for DeepImmutable etc. Just need to cleanup and add a few test and then will push later this week. Also added typings for a few of the other side packages (html-serializer and react) @lichenglu if you want to help you could add typings for the other side-packages? They should be fairly straight forward and for inspiration you could have a look at the current ones on DefinitelyTyped?

zhujinxuan:
Hi, I am not sure whether this is doable in ts checking:
```
  const child = document.getNodeByKey(key);
  if (child) return;
  const parent :: (NotNULL)= document.getParentByKey(key);
  const path :: (NotNULL)= parent.getPath(key)
```

Sometimes, we want to assert non-null according to our knowledge, but I am not sure whether ts allow us to assert non-null.

ianstormtaylor:

Issue #2147, opened 2018-09-02
slate-react@0.18.4 breaks editor focus on desktop (Safari, Chrome, Firefox)


zarv1k:
#### Do you want to request a _feature_ or report a _bug_?
bug
#### What's the current behavior?
Current latest slate stack:
slate@0.40.2
slate-react@0.18.4

Empty editor can't take focus: https://jsfiddle.net/zarv1k/vcdpn74j/
So it's impossible to add anything into the editor.

The issue is not only reproduced in empty editor. After clicking in blurred editor on the area after text content, focus loses immediately.

#### What's the expected behavior?
Empty editor can take focus, user can add anything into the editor:
slate@0.40.2
slate-react@0.18.3
https://jsfiddle.net/zarv1k/wv5ky9fq/1/

briandk:
Anyone have thoughts on how to fix this? If so, I'm happy to work on a PR.

dmitrizzle:
Same issue in Chrome 68.0.3440.106 macOS (looking at jsFiddle).

ericedem:
Oh nope, this is definitely because of `onNativeSelectionChange()` getting added.

ericedem:
For anyone looking for a workaround for the time being, you can do this:

```
<Editor
  ...
  onFocus={(event, change) => change.focus()}
/>
```

We were already doing this to workaround focus breaking when tabbing into the editor, which seems like it is probably a related issue.

zarv1k:
@dmitrizzle can't agree with you about FF in my FF v.61.0.2 on macOS 10.13.6. It looks like editor has focus (you can see the caret), but you can't add anything into the editor.

zarv1k:
> @zarv1k What browser is it failing for you? I confirmed Chrome is broken as well.

In all browsers on macOS 10.13.6:
Safari 11.1.2
Chrome 68.0.3440.106
Firefox 61.0.2

dmitrizzle:
@zarv1k I use 61.0.2 on macOS 10.12.6 and I can edit the text. Very strange.

zarv1k:
Some screen recordings:
[Safari](https://www.dropbox.com/s/0nr9riu402vdus7/ios-slate-react-0.18.4.mp4?dl=0) on iOS 11.4;
[FF 61.0.2](https://www.dropbox.com/s/tj5rxnpbg3f8eqi/FF-slate-react0.18.4.mov?dl=0) on macOS 10.12.6.

[...]

Issue #2143, opened 2018-08-31
Backspacing with no content in iOS throws an exception


ericedem:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Backspacing with no content in iOS throws an exception.
Try the rich text example: https://slatejs.org

1. Delete all the text.
2. Hit backspace
3. 💥 

![kapture 2018-08-31 at 9 26 28](https://user-images.githubusercontent.com/1060669/44924399-0331ca00-ad00-11e8-9f8f-bc827410a5eb.gif)

#### What's the expected behavior?

Shouldn't throw an error.

ericedem:
From errors I have seen in our app, it looks like the last block itself may be getting deleted, which throws the editor into chaos.

ericedem:
It looks like now the last block is able to get deleted. I was able to workaround this locally by adding the following to our schema:

```
document: {
  nodes: [
    {
      match: { object: 'block' },
      min: 1
    }
  ],
  normalize(change, error) {
    if (error.code === SlateSchemaViolations.CHILD_REQUIRED) {
      // We can't use change.insertBlock() because the first thing it will
      // do is try to find a `startBlock` which will throw an exception,
      // since there are no blocks.
      change.insertNodeByKey(
        error.node.key,
        0,
        { object: 'block', type: EditorConstants.NODE_TYPE_PARAGRAPH }
      );
    }
  }
}
```

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #2150, opened 2018-09-04
fix(content): attach the right `onNativeSelectionChange`.

ericedem:

Fixing a bug

This undoes a recent change https://github.com/ianstormtaylor/slate/pull/2131

Which hooked up `this.handlers.onNativeSelectionChange` to `'selectionchange'` as well as `this.handlers.onBeforeInput` to `'beforeinput'`.

It turns out that in order to fix the iOS issues, we only needed the change to `onBeforeInput`. The corect handler for `'selectionchange'` is the instance method `this.onNativeSelectionChange`. This broke a whole bunch of focus related issues.

I did double check that with this change, https://github.com/ianstormtaylor/slate/issues/2125 is still fixed.

Fixes: #2147

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -91,7 +91,7 @@ class Content extends React.Component {
 
     window.document.addEventListener(
       'selectionchange',
-      this.handlers.onNativeSelectionChange
+      this.onNativeSelectionChange
     )
 
     // COMPAT: Restrict scope of `beforeinput` to clients that support the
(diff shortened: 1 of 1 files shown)


ericedem:
It does look like this brings back: #2129, at least when the editor is empty.

ericedem:
Ahh nevermind, I was misremembering issues, #2129 will be fine. The issue I was remembering was https://github.com/ianstormtaylor/slate/issues/2143 which is still outstanding.

I think this PR is good to go as is.

ianstormtaylor:
Thanks @ericedem!

Issue #2139, opened 2018-08-30
Add comparator functions on schema


Calyhre:
#### Do you want to request a _feature_ or report a _bug_?

A feature

#### What's the current behaviour?

Right now, we can specify rules in the schema that will match against properties of the nodes. The comparison is often as simple as a `===` against two attributes.

Example for a rule like this:

```js
  // Merge adjacent text nodes.
  {
    match: { object: 'text' },
    next: [{ object: 'block' }, { object: 'inline' }],
    normalize: (change, error) => {
      // Normalize here...
    },
  },
```

It takes each node where `object === 'text'` and check if the next sibling node verifies `object === 'block'` or `object === 'inline'`. If not, it goes through the `normalize` method.

It's enough to cover lots of use case, but I think we can go a little further.

#### What's the expected behaviour?

Let's say we want to enforce no succeeding blocks of certain types. I'll take an `unordered-list` as example. Right now we would have to list each and every other type of block on the rule to enforce this.

```js
  {
    match: { type: 'unordered-list' },
    next: [{ type: 'ordered-list' }, { type: 'header-one' }, ...everyOtherTypes],
    normalize: () => { /**/ },
  },
```

Would be really inefficient and error prone, especially for published plugins who have no idea of every other block types.

One easy way to solve this would be to inverse the condition. But I actually have no idea how to properly implement this, so here is 3 other options instead 😅:

1. Allow function to be passed as rule, the function would return a falsey value, or the error code
   ```js
    {
      match: { type: 'unordered-list' },
      next: node => node.type === 'unordered-list' && 'next_sibling_invalid_type'
    }
   ```
2. Allow functions to be passed as matching attributes, same return as above. It would allow as to combine it with other simple conditions
   ```js
    {
      match: { type: 'unordered-list' },
      next: {
        type: type => type === 'unordered-list' && 'next_sibling_invalid_type'
      },
    }
   ```
3. Both :D

Would be as simple as adding a `typeof === 'function'` at a few places like here [src/models/schema.js#L533](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/models/schema.js#L533)

I could handle the implementation if that interests you :)

isubasti:
this could be useful to do some kind of negation check like not having void node as the last node rather than checking that the last node should be of type like the example schema for images. any thought @ianstormtaylor ?

PR #2151, opened 2018-09-04
Add support for validating functions inside schema's rules

Calyhre:

Adding a feature

#### What's the new behaviour?

With this change we can now use validating functions inside schema's matching properties, at least for a few keys.

A few example:
```js
{
  match: { /* something */ },
  object: kind => kind !== 'text',
  type: type => ['paragraph', 'image'].includes(type),
  data: data => data.has('thing'),
  marks: mark => mark.type === 'mention' && mark.data.has('userId'),
}
```

I also added documentation, specs, and default normalization on `next` and `previous` rules.

A more complex one now allowed by this change:

```js
{
  // Merge two consecutives lists
  match: { type: 'list' },
  next: {
    type: type => type !== 'list'
  },
  normalize: (change, error) => {
    if (error.code === 'next_sibling_invalid_type') {
      change.mergeNodeByKey(error.next);
    }
  },
}
```

Fixes: #2139

### DIFF
--- a/docs/reference/slate/schema.md
+++ b/docs/reference/slate/schema.md
@@ -154,6 +154,24 @@ Will determine whether the node is treated as a "void" node or not, making its c
 
 Will validate the last child node against a [`match`](#match).
 
+### `next`
+
+`Object|Array`
+
+```js
+{
+  next: { type: 'quote' },
+}
+```
+
+```js
+{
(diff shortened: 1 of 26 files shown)


ianstormtaylor:
Thank you @Calyhre, this is an absolutely perfect pull request!

PR #2155, opened 2018-09-05
Fix deprecated function calls in slate-react

skogsmaskin:

Fixing deprecated function calls in ``slate-react`` (get-event-range.js).


ericedem:
Nice catch @skogsmaskin, does this fix any outstanding issues that you know of?

skogsmaskin:
@ericedem, not that I know of, I discovered it because we have some code in our implementation that is using ``getEventRange`` from slate-react, and I'm currently getting the code up to date with all the recent changes in Slate.

skogsmaskin:
Actually found another in that file (``moveToEndOf``). I've amended it to the PR commit.

PR #2156, opened 2018-09-06
Docs: Add info on how to use schema object

FWirtz:
Extend the documentation to more obviously show where to reference the created schema object.

In light of a couple of questions on Slack recently [[1]](https://slate-js.slack.com/archives/C1RH7AXSS/p1535275266000100) [[2]](https://slate-js.slack.com/archives/C1RH7AXSS/p1534953134000100?thread_ts=1534952365.000100&cid=C1RH7AXSS) regarding how to tell the editor about the schema object one created, I propose an extension to the current docs.
I am by no means a Slate expert (actually pretty much the opposite) but I think that this extension would eliminate some confusion around schemas.
I am very open to changes/comments/whatever regarding this proposal.

---
Extends the documentation.

I have.

### DIFF
--- a/docs/guides/schemas.md
+++ b/docs/guides/schemas.md
@@ -8,6 +8,8 @@ That said, just because Slate is agnostic doesn't mean you aren't going to need 
 
 To that end, Slate provides a `Schema` model, which allows you to easily define validations for the structure of your documents, and to fix them if the document ever becomes invalid. This guide will show you how they work.
 
+> ❗️To tell Slate about your custom schema add it to the editor as a prop [like this](https://github.com/ianstormtaylor/slate/blob/405cef0225c314b4162d587c74cfce6b65a7b257/examples/forced-layout/index.js#L62).
+
 ## Basic Schemas
 
 Slate schemas are defined as Javascript objects, with properties that describe the document, block nodes, and inline nodes in your editor. Here's a simple schema:


ericedem:
Might also be worth mentioning that they can be piecemeal with plugins too. Though there aren't relevant examples I'm aware of.

ericedem:
Ah yea, the general idea is that most things you can add as props to the editor can also be added as props in a plugin. So things like event handlers / schemas etc. can also be done there.

I've been thinking there should be a bigger more complicated example, probably with schemas rich text and lots of plugins to show really advanced use cases. Still thinking about what that should look like.

FWirtz:
I'd love that! Am currently working on a slightly more complicated schema usecase, maybe that helps as inspiration:
- root nodes: `title` (exactly `1`) and `section` (`n` different)
- all other nodes (`img`, `p`, `code`, `a`, ...) must be inside `section`s
- `section` may only be a child of `document` (no nested `section`s)
- `title` may only contain text

That would be a little more restrictive and a larger example, therefore it'd be nice to split some schema logic up and move it to the plugins. I still have a few struggles with that though.

Will work on that on Monday, maybe you could give some input on it and we could then add it as a further example?

FWirtz:
@ericedem did a lot of reading ([GitbookIO/slate-edit-code](https://github.com/GitbookIO/slate-edit-code/blob/master/lib/validation/schema.js) seems to be making good use of that, the code is pretty complex though) but am stuck now with my setup.

In my simple usecase I don't want to allow any kinds of marks for the `title` element. I go at it as such:
```js
const schema = {
    blocks: {
        title: {
            marks: [],
        }
    }
}
```
It seems that Slate doesn't recognise this schema. Seems to get lost somewhere on the way `Plugin -> Editor`.
This is the path upwards:
```jsx
export default function Title(options) {
    return {
        // changes, helpers etc. here
        plugins: [
            // other funcs here
            schema,
        ],
    }
}
```
And in `Editor.jsx`:
```jsx
import Title from './plugins/title'

this.plugins = [
    ...Title().plugins,
    // other plugins
]

<SlateEditor plugins={this.plugins} />
```

I've been looking at my code and Gitbooks code now for quite a while and feel like I am blind. Can you have a quick look whether you spot the bug? 🙏 
(The project is opensource, [this is the commit](https://github.com/schul-cloud/edtrio/commit/cc7a5862ec05249c41cd31ea8ae325fdd91b68cb))

ericedem:
Everything looks ok to me except that schema itself isn't a plugin, so you would need the return value of your plugin to be:
```
return {
  plugins: [{
    schema
  }]
}
```

Though since you are just creating one plugin, I would simplify it. To just return the one plugin, which would change your code to be:

```
return {
  schema
}
```

And in the Editor it would be:

```
this.plugins = [
  Title()
]
```

P.S. The slack channel is a good place to ask questions like this, lots of helpful people there :)

FWirtz:
@ericedem wow thanks it actually was the `{}`...
Sorry, I know about the Slack, probably was to fried to think about that yesterday.
I'll delete my other message so that we don't mess up the PR here.

Thanks for your help!

Issue #2154, opened 2018-09-05
Examples/code-highlighting error


An0nymous0:
#### Do you want to request a _feature_ or report a _bug_?
examples bug

https://github.com/ianstormtaylor/slate/tree/master/examples/code-highlighting

#### What's the current behavior?
Enter 1 and press Enter to repeat this fast
![image](https://user-images.githubusercontent.com/6836030/45081344-9700fd00-b129-11e8-9309-13ea37e843d3.png)
![image](https://user-images.githubusercontent.com/6836030/45081384-abdd9080-b129-11e8-99f6-44cc757b10c5.png)

#### What's the expected behavior?
No error

ericedem:
@An0nymous0 Can you please add browser info and more detailed steps? I'm having trouble reproducing this issue.

uipoptart:
Looks like there is an issue with the way the decoration is made on line 228-236. The anchor and focus values aren't getting their paths assigned. So, when the paths are compared later down the line the value is null. I'll try to put together a pull request later this afternoon.

SmilinBrian:
#2319 has a more definitive fix.

brendancarney:
Closing in favor of #2319.

PR #2157, opened 2018-09-06
Pass path to decoration in code highlight example

uipoptart:
Decoration accepts a point for it's anchor and focus values. A point requires a key, offset and a path. Without the path this will crash when comparing the point's paths.
https://github.com/ianstormtaylor/slate/issues/2154

Fixing a bug outlined in the issue above.

Fixes: #2154
Reviewers: @An0nymous0

### DIFF
--- a/examples/code-highlighting/index.js
+++ b/examples/code-highlighting/index.js
@@ -226,18 +226,10 @@ class CodeHighlighting extends React.Component {
 
       if (typeof token != 'string') {
         const dec = {
-          anchor: {
-            key: startText.key,
-            offset: startOffset,
-          },
-          focus: {
-            key: endText.key,
-            offset: endOffset,
-          },
-          mark: {
-            type: token.type,
-          },
-        }
+            anchor: { key: startText.key, offset: startOffset, path: node.getPath(startText.key) },
+            focus: { key: endText.key, offset: endOffset, path: node.getPath(endText.key) },
+            mark: { type: token.type }
+          }
 
         decorations.push(dec)
       }


isubasti:
should probably fix code inside https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/text.js#L122 instead of the example?

ericedem:
This is a bit confusing to me, since it seems a `Point()` can have either a `key` or `path` or both. It seems like we need to do something like `startPath = start.path ||  document.assertPath(start.key).

matt-erhart:
@ericedem true. Each node in this.state.value.document.getTexts() has a key and not a path, but this.state.value.selection.toJS({preserveKeys: true}) has both, though I had to hunt for {preserveKeys: true}, and I haven't found an explanations for what exactly path is or how its different to key.

ericedem:
After asking on slack my understanding is that keys are a monotonically increasing identifier for a node, similar to a primary key in a database. These are great for keeping track of nodes, but are very non performant because finding a node by a given key involves doing a depth first search across the document.

Paths were introduced to allow fast lookups, as they are an array of child indices in the document tree. They are often interchangeable for the purposes of identifying a node, but not always.

This is why points can have both, since they key may identify the point for comparisons, but you might also provide a path as well.

With this in mind, I think it does make sense to just lookup the path based on the key if it doesn't exist.

P.S. There is an issue to convert most of the slate internal logic to use paths for performance reasons but having trouble finding it just now.

uipoptart:
I'm closing this as it sounds like we should fall back to looking up by key when the path doesn't exist.

Issue #1962, opened 2018-07-09
Setting focus and selection with toggleMark destroys mark


uipoptart:
#### Do you want to request a _feature_ or report a _bug_?
Bug
#### What's the current behavior?
Programmatically setting selection and/or focus at the same time as toggleMark results in the newly applied mark being lost.

http://g.recordit.co/wi96Bp20tk.gif
https://jsfiddle.net/321nvc4k/

Mac OS 10.13.5
Google Chrome 67.0.3396.87
Slate version: Using forked current JS fiddle to reproduce

#### What's the expected behavior?
The editor will take focus with the desired selection. Typing will result in text with desired mark applied. 

#### When this matters
If a user tries to select font family, size, etc before clicking in the editor the controls will appear as if they don't work.

Soreine:
In the JSFiddles, it does not actually update the selection at all.

PR #2158, opened 2018-09-07
Fix marks being cleared on selection events

Soreine:
This is a bug fix.

#### Bug example

https://jsfiddle.net/fj9dvhom/2040/

Here is a JSFiddle with a button that inserts some bold text, and then remove the bold from active marks. The bug is that the bold stays active:

marks

#### What's the new behavior?

The bug come from the fact that any `onSelect` event will clear the active marks from the selection. And in this case, an `onSelect` event is happening after the operation. Though the selection has not changed, this is something that can happen [as stated here](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/content.js#L290-L294). The fix happens in the `onSelect` handling by the `AfterPlugin`, that was currently ignoring active marks.

---

I have checked that:

Reviewers: @uipoptart because it could be related to https://github.com/ianstormtaylor/slate/issues/1962

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -677,6 +677,10 @@ function AfterPlugin() {
       if (next) range = range.moveFocusTo(next.key, 0)
     }
 
+    // Preserve active marks from the current selection.
+    // They will be cleared by `change.select` if the selection actually moved.
+    range = range.set('marks', value.selection.marks)
+
     let selection = document.createSelection(range)
     selection = selection.setIsFocused(true)
     change.select(selection)


ianstormtaylor:
Thanks @Soreine this is great!

bryanph:
@Soreine for me this commit breaks the examples

packages/slate-react/src/plugins/after.js
```diff
@@ -677,6 +677,10 @@ function AfterPlugin() {
       if (next) range = range.moveFocusTo(next.key, 0)
     }
 
+    // Preserve active marks from the current selection.
+    // They will be cleared by `change.select` if the selection actually moved.
+    range = range.set('marks', value.selection.marks)
+
     let selection = document.createSelection(range)
     selection = selection.setIsFocused(true)
```
Soreine: My bad, it should be `selection.set('marks', ...)`

Soreine:
I have fixed this in https://github.com/ianstormtaylor/slate/pull/2181
Thank you @bryanph

PR #2161, opened 2018-09-07
docs(change.md): correct `setOperationFlag()` signature

ericedem:

Fixing docs

`setOperationFlag()` actually accepts a `key` and a `value`.

https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/models/change.js#L161


ianstormtaylor:
Thanks @ericedem!

Issue #2162, opened 2018-09-08
Slate-react editor crashes when `window.getSelection()` returns null


eramdam:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

Versions:

```json
"react": "^16.4.1",
"react-dom": "^16.3.2",
"slate": "^0.40.2",
"slate-html-serializer": "^0.7.2",
"slate-plain-serializer": "^0.6.2",
"slate-react": "^0.18.3",
```

#### What's the current behavior?

On Firefox, `window.getSelection()` can return null [in some cases](https://bugzilla.mozilla.org/show_bug.cgi?id=827585), in that case, when it's called in an iframe that is visually hidden.

This causes slate-react's `Content` component to completely crash on the first mount because the code assumes that `window.getSelection()` always returns something

https://github.com/ianstormtaylor/slate/blob/3db717761e8c4e553c2c259dcff415e8f79ebe4d/packages/slate-react/src/components/content.js#L146-L147

I have a [JSFiddle snippet](http://jsfiddle.net/bmj6840o/1/) that reproduces the bug.

As far as I could tell, modifying this

```js
const native = window.getSelection()
const { rangeCount, anchorNode } = native
```

to

```js
const native = window.getSelection()

if (!native) return;

const { rangeCount, anchorNode } = native
```

fixes the issue. If that sounds good to you, [here is a PR](https://github.com/ianstormtaylor/slate/pull/2163)

#### What's the expected behavior?

Don't crash 😅

PR #2163, opened 2018-09-08
Avoid crash in updateSelection() in editor when getSelection() is null

eramdam:

Fixing a bug 

#### What's the new behavior?

Content in slate-react will not crash if on componentDidMount if `window.getSelection()` returns null

`window.getSelection()` can sometimes return null (see https://bugzilla.mozilla.org/show_bug.cgi?id=827585). This change checks the return value of `window.getSelection()` before trying to use it.

Fixes: #2162
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -144,6 +144,11 @@ class Content extends React.Component {
     const { isBackward } = selection
     const window = getWindow(this.element)
     const native = window.getSelection()
+
+    // .getSelection() can return null in some cases
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=827585
+    if (!native) return
+
     const { rangeCount, anchorNode } = native
 
     // If both selections are blurred, do nothing.


packages/slate-react/src/components/content.js
```diff
@@ -144,6 +144,11 @@ class Content extends React.Component {
     const { isBackward } = selection
     const window = getWindow(this.element)
     const native = window.getSelection()
+
+    // .getSelection() can return null in some cases
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=827585
+    if (!native) return
```
ericedem: Hmm, won't this prevent the DOM selection from updating to match the editor correctly? I wonder if maybe there should be some `null` checks further down to make sure the rest of the logic gets executed.

eramdam: @ericedem do you have any suggestion as to tests that I could perform to de-risk this PR?

ericedem: Nope, I think unfortunately `content.js` doesn't really have any tests yet. I'm not familiar enough with the codebase to give a recommendation.

ianstormtaylor:
Thanks @eramdam!

Issue #2152, opened 2018-09-04
Runaway recursion when trying to access `editor.value` in a plugin `onChange` event


ericedem:
#### Do you want to request a _feature_ or report a _bug_?

Report a bug

#### What's the current behavior?

Get a depth exception when trying to access `editor.value` in a plugin `onChange` event.

Just try to load this fiddle and it will throw errors: https://jsfiddle.net/uvkrn9ge/1/

![image](https://user-images.githubusercontent.com/1060669/45058352-487e3080-b04d-11e8-89cb-614a3b7072b7.png)

Chrome latest Mac OSX.

Note: I did narrow this down to a change in version 4.0.2 of `memoize-one`. I think this error may have always been happening, it was just covered up.

ianstormtaylor:
@ericedem what do you need to access `editor.value` for? Instead you might be wanting `change.value` which is the current value of the change with operations applied?

ericedem:
We do a comparison of `editor.value` vs `change.value` to determine if we should save the editor value to a database. Though maybe there is a better way to determine if there are changes 🤔

ericedem:
Basically just following this example: https://docs.slatejs.org/walkthroughs/saving-to-a-database

Comparing `change.value.document` against `editor.value.document` since in a plugin we don't have access to the state.

ericedem:
Ok so, the JSFiddle I linked before doesn't actually reproduce the issue we are seeing in our dev environment. I'm having trouble reproducing in a fiddle or examples, but still working on it.

While debugging, I noticed part of our infinite recursion is coming from this bit of code. When accessing `editor.value` this piece of code runs:

https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/editor.js#L392-L393

Which creates a change on the value and forces all of the plugin change events to run again. 
@zhujinxuan  Do you remember the motivation behind this bit of code? It seems like a strange side effect of trying to get the value to create an empty change and call a bunch of event handlers, when there wasn't actually a change.

ericedem:
Ok this will be reproducible with this fiddle once a new version of slate gets published with memoize-one@4.0.2. https://jsfiddle.net/uvkrn9ge/21/

The loop is:
1. onChange() in the plugin gets called initially with the value being set.
2. editor.value is dereferenced, causing the getter in this.resolveChange to get called (mentioned in my last comment)
3. The plugins are then fed an empty Change instance into their `onChange()` handlers.
4. GOTO: 2

[...]

PR #2165, opened 2018-09-10
Gives a warning when visiting editor.value during onChange

zhujinxuan:

perhaps a bug;  Editor run into a runaway recursion when visiting editor.value during onChange
https://github.com/ianstormtaylor/slate/issues/2152

#### What's the new behavior?

1. Allows user visit editor.value during onChange. 
2. Gives a warning when user do it.

Adds tmp.isResolvingValue in editor

Fixes: #2152
Reviewers: @ericedem

### DIFF
--- a/packages/slate-react/src/components/editor.js
+++ b/packages/slate-react/src/components/editor.js
@@ -82,6 +82,7 @@ class Editor extends React.Component {
   tmp = {
     change: null,
     isChanging: false,
+    isResolvingValue: false,
     operationsSize: null,
     plugins: null,
     resolves: 0,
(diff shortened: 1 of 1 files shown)


ericedem:
I'm not sure I understand. Is the API user supposed to be only calling editor.props.value? How could the user know that there is a change being resolved in order to call the right one?

It seems like the user shouldn't be calling `editor.value` at all?

zhujinxuan:
@ericedem You can call `editor.value` with onEvent like onKeyDown, but not with onChange

ericedem:
@zhujinxuan What benefit is there to calling `editor.value` in an event handler as opposed to `editor.props.value`? Isn't the value guaranteed to be normalized in handlers (except onChange)?

Issue #2069, opened 2018-08-10
Option + Arrow not working in image demo.


bebbi:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

Noticed this when evaluating slate:
In the [image demo](https://www.slatejs.org/#/images), trying to move the cursor with `Option + Left/Right Arrow` (macos)

For the text above the image, option + left arrow works as expected, but option + right arrow moves only 1 character per keypress.

For the text below the image, it's the opposite: option + right arrow works, while option + left arrow moves 1 char at a time.

Tried in macos 10.13, Chrome, FF, Safari

![slate_cursor_word_move](https://user-images.githubusercontent.com/2752391/43947452-702fffa6-9c88-11e8-93ba-adae3fb13dd7.gif)

#### What's the expected behavior?

The `option` key should modify the cursor behavior to move words in both directions.

ericedem:
Ok this is a little trickier than I first thought. The current problem is that if the next text is in a void node, when trying to jump right it will always only move one character. This is because we don't want to jump over the void node, which is the browser's native behavior (at least in chrome).

This is problematic, because we don't necessarily know exactly how the browser is tracking word boundaries. It seems like we have a couple options here to try and fix this:

1. We try to do a custom word boundary implementation for checking. This will be complicated, and there is no guarantee we will get it right across all browsers.
2. We just remove the void checking logic, and let the modifier key jump over void nodes. This isn't ideal, but would be less broken than what is currently happening.

I'd vote for 2 for a stop gap, and maybe someone can think up something cleverer than 1.

skogsmaskin:
I would like to add that the selection also is moved (with the same length) like this reported here:
https://github.com/ianstormtaylor/slate/issues/2224 when it should collapse.

PierBover:
Hey @ericedem why would option 2 be less ideal? Why is it problem if the cursor jumps over the void node?

I guess I'm missing something, but jumping over the non text node is what I would expect as a user. It's what Google docs does.

![2018-10-16 08_17_39](https://user-images.githubusercontent.com/226483/47019160-03552000-d11c-11e8-96a9-98a0a5550799.gif)

PierBover:
Ah I get it. Excuse my ignorance, I'm relatively new to Slate.

Personally I'd rather have solution 2 than no solution at all... Could this be solved temporarily by a quick and dirty plugin?

[...]

PR #2167, opened 2018-09-11
fix: moving around void nodes

ericedem:

Fixing a bug

#### What's the new behavior?

Now the behavior works as expected around void nodes. The cursor jumps forward to the next word when pressing option + forward arrow, and to the previous word when pressing option + back arrow.

The problem was that we were arbitrarily checking if the next node was a void, even though there was still text to traverse in the current node. Now we check to see if the current node contains text and if so just move as normal.

Fixes: #2069


ericedem:
I realized that this is going to break behavior when the text is empty, closing in favor of: #2169

Issue #1678, opened 2018-02-26
Suggestion: Render placeholders after regular node children for consistent cursor behavior.


lukasbuenger:
#### Do you want to request a _feature_ or report a _bug_?

Kind of a feature request, more like a suggestion 

#### What's the current behavior?

slate-react's Node component renders the placeholder from `Plugin.renderPlaceholder` before the regular children of the node.

https://github.com/ianstormtaylor/slate/blob/3339d088e14ad0bf04a674c14834179403ae369f/packages/slate-react/src/components/node.js#L158 

#### What's the expected behavior?

I'd expect Placeholder components to get rendered *after* the regular (and probably empty) children of a node in order to have consistent cursor positioning (at start and before the placeholder) even when the placeholder is not absolutely positioned:

`children = [...children, placeholder,]`

vilic:
@ianstormtaylor Any updates on this? I tried this in my own build previously, and it works in my case.

Related to #1344

PR #2168, opened 2018-09-11
improvement(slate-react): optimise cursors in inline placeholders

lukasbuenger:

This improves cursor positioning in empty nodes with an inline (not absolutely positioned) placeholder.

#### What's the new behavior?

Cursor is positioned *before* the placeholder instead of after. 
It's basically `[...children, placeholder]` instead of `[placeholder, ...children]` 

Out of the box. No adjustments to existing code needed. 

Fixes: #1678
Reviewers: @vilic @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/node.js
+++ b/packages/slate-react/src/components/node.js
@@ -174,7 +174,7 @@ class Node extends React.Component {
         key: `${node.key}-placeholder`,
       })
 
-      children = [...children, placeholder, ]
+      children = [...children, placeholder]
     }
 
     const element = stack.find('renderNode', {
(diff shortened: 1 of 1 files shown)


vilic:
It's failing ESLint. 😂

lukasbuenger:
Aw man, mucho embarrassing. I'm sorry guys!

ianstormtaylor:
I believe the reason it was positioned before instead of after was because in some browsers it would cause an issue where you couldn't click into the content. But I think https://github.com/ianstormtaylor/slate/issues/2257 gives us the best chance of making this kind of behavior customizable.

PR #2169, opened 2018-09-11
fix: use text utils to move forward and backward by word

ericedem:
This was causing a problem where when the current text was followed by a void node using a modifier key to move forward would force the selection to creep forward a character at a time.

bug

#### What's the new behavior?

With this change, now the modifier key will cause selection to move as expected, and not jump over void nodes. Moving forward and backward with and without modifier keys will now be fully controlled by slate, instead of trying to rely on a combination of browser behavior and preventing that behavior sometimes by slate.

Fixes: #2069 
Reviewers: @ianstormtaylor


ericedem:
@ianstormtaylor does this seem like a reasonable solution?

ericedem:
Yea ideally there would be some `moveWordForward()` method we could use to hop forward and stop on void nodes, but I think this will be a pain to implement (unless there is some unicode boundary checking library out there). Though maybe someone else has a better idea.

The change in this PR will cause word boundary movement to work (relying on the browser to detect word boundaries). The problem is that the browser doesn't know what a void node is, and that it should stop on an image or some other widget. While not ideal, I think this would probably be a more expected behavior by a user.

ianstormtaylor:
> Yea ideally there would be some moveWordForward() method we could use to hop forward and stop on void nodes, but I think this will be a pain to implement (unless there is some unicode boundary checking library out there). Though maybe someone else has a better idea.

This should already be implemented, as `moveWordForward` I think. We have unicode-checking logic in the `TextUtils` helper functions. 😄

ericedem:
> This should already be implemented, as moveWordForward I think.

Ok thanks, I'll need to look around more as I couldn't find it. Looking again I do see we have `Changes.deleteWordBackwardAtRange` which makes use of `TextUtils.getWordOffsetBackward` which should solve this problem. Will update the PR.

skogsmaskin:
Thanks for working on this @ericedem! Will this also fix issue https://github.com/ianstormtaylor/slate/issues/2224?

ericedem:
@ianstormtaylor @skogsmaskin I've updated the code to what we talked about, and I think it is a bit easier to understand now too. If you guys like this approach, I'll add some more tests like the one I added and update the docs for the new commands.

One other note: The only commands I think we need to check for is normal left / right and modified left / right to move backward / forward by word. This is just ctrl+arrow on windows (and linux I think?) and opt+arrow on mac. If you guys know any other hotkeys I can try that out too, but right now whatever those behaviors are should just result in the default browser behavior.

ianstormtaylor:
@ericedem this is an absolutely amazing pull request! Thank you!

I noticed, I think <kbd>cmd+left/right</kbd> is also the `moveLineBackward/Forward` on the Mac. I think we should handle that? And we'd also want `moveLineBackward/Forward` methods as well? 

(In non-browser contexts, we equate "line" to "block". But in the browser we can allow a browser-specific plugin to override "line" to actually calculate based on line breaks. But this can be something that core doesn't have the care about.)

ericedem:
@ianstormtaylor based on my testing it looks like using the browser defaults gets us the "line" behavior. Where cmd + left / right move to the beginning / end of a line based on display. I think we should be good to go without having to handle it manually.

There was a problem with move line before, where moving towards a void node would step one character at a time using `moveForward()` or `moveBackward()`. Now that we don't lump all the movement into one handler, we are free to let move line behavior do its own natural behavior.

ianstormtaylor:
Ah, that sounds good. Thank you very much @ericedem!

skogsmaskin:
Sorry didn't get the chance to say **great work @ericedem** before it was merged! You rock!

skogsmaskin:
I see there's a little snag (shift-left/right), but I'm sure that will get sorted out.  <3

Issue #2164, opened 2018-09-10
change.insertText(...).extend is not a function


victor2405:
after updating slate version got error: 
change.insertText(...).extend is not a function. The same error on examples here:https://www.slatejs.org/#/links. Any suggestion?

ericedem:
Looks like this one got missed, `extend()` has been renamed to `moveFocus()` in 0.37.x https://github.com/ianstormtaylor/slate/blob/master/packages/slate/Changelog.md#0370--august-3-2018

ericedem:
I don't have collaborator status to reopen, but there is a fix ☝️

maysam:
TypeError: change.insertText(...).moveFocus is not a function

PR #2175, opened 2018-09-14
fix(example/links): a few fixes for the links example

ericedem:
* Correctly use the selection
* check whether the user cancelled the prompt
* use `moveFocusBackward`

_bug_

#### What's the new behavior?

Make links example work again after API changes, and add cancel handling for the prompt.

Fixes: #2164


ianstormtaylor:
Thank you @ericedem, this is great!

Issue #1819, opened 2018-05-04
Clicking stepper arrows on number input outside Slate crashes in Firefox


bjoerge:
This seems to be related to #1744, but doesn't look like it's exactly the same issue.

![slate-bug](https://user-images.githubusercontent.com/876086/39630142-bebe2bc4-4fae-11e8-91db-c8aa746d1d2c.gif)

Here's a jsfiddle to reproduce: https://jsfiddle.net/cmu57eym/5

Happens in Firefox 59.0.2 on macOS 10.13.4 and possible in other firefoxes too.

bjoerge:
So did some more digging and boiled it down to the following repro case (which is, from what I understand from reading the source code, is equivalent with what happens in Slate):

```html
<input type="number" id="input">
<script>
document.getElementById('input').oninput = () => {
  window.getSelection().anchorNode.nodeType
}
</script>
```

This fails with the same error as shown above.

This looks very much like a bug with Firefox. What happens is that when the `input`-event is firing, the `window.getSelection().anchorNode` is pointing to the actual stepper button, which is restricted, causing attempts to access *any* of its properties to throw.

bjoerge:
Fixed in #2178

Issue #1744, opened 2018-03-30
Using <input> inside a contenteditable="false" is broken in Firefox


Zhouzi:
We noticed a bug in Firefox, rendering a node as follows is broken:

```html
<div contenteditable="false">
  <input type="text" />
</div>
```

There are two things to note:

1. Left/right arrow keys don't move the caret
2. Pressing a character key (or backspace) throws an error:

> Permission denied to access property "nodeType"

I created a [repository](https://github.com/Zhouzi/slate-firefox-bug) reproducing the error with Slate and a minimal setup:
http://gabinaureche.com/slate-firefox-bug/

The error comes from `isInEditor` which tries to read `window.getSelection().anchorNode.nodeType` but `anchorNode` is "Restricted".

This bug is reproducible without Slate, here's a pretty simple scenario:
https://jsfiddle.net/hsqebsk2/1/

I filed an issue to Firefox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1449936

Probably related to #843

CameronAckermanSEL:
This can be worked around using 
```
<div contenteditable>
    <div contenteditable="false">
        <input contenteditable="true" type="text" value="Some text content">
    </div>
</div>
```

But if this bug is reproducible outside of slate, I think this is an indication that this is not a slate bug.

CameronAckermanSEL:
This can be worked around by using 

```

<div contenteditable>
    <div contenteditable="false">
        <input type="text" value="Some text content">
    </div>
</div>
```

Or by using the componentDidThrow() react event handler to keep these errors from crashing the editor.

I also don't think this is a Slate bug.

Zhouzi:
Thanks for the suggestion @CameronAckermanSEL 👍 

I gave it a try and it does work outside of Slate: https://jsfiddle.net/hsqebsk2/4/

But doesn't in the context of Slate 😕

nkoder:
not sure whether it will help anyone, but I have similar issue, and here is my code sample to reproduce it on Firefox: https://codepen.io/nkoder/pen/pKLQWz

Iuriy-Budnikov:
I have the same issue +1

Soreine:
Either we try to make Slate work around the bug, or we find a way to pressure Firefox into conforming to other browser's behavior. We have **[filed this bug to Mozilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1449936)**, but it is not getting much attention...

[...]

Issue #843, opened 2017-05-25
Embeds example completely broken on Firefox


AlbertHilb:
Embeds example completely broken on Firefox 53 (Ubuntu 16.04). :cry:

1. Insert cursor in the first line of text above the video, let's say just after the word "this" and then press `down-arrow` key, the caret jumps to the `input` element below the video but pressing `backspace` or `delete` keys still removes text from the line above the video.

2. Clicking on the input element below the video, put the caret inside it. Then, try to move the caret using the mouse or pressing `left-arrow` or `right-arrow` keys, the caret doesn't move.

3. Look at console while editing `input`element content. A lot of errors of the type
    ```
    Error: Permission denied to access property "isContentEditable"
    ```
    are raised.

The main problem is this: in Firefox, when the selection is inside an `input` element,  `anchorNode` and `focusNode` are **private** text nodes living inside it and accessing their properties or methods is denied.

geakstr:
I see `Error: Permission denied to access property "window"` in Firefox (63) when clicking on input in embed youtube example. In my project as well, I have void blocks with inputs, can't find workaround for that.

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/1496368/49668143-b126e000-fa76-11e8-8900-095a35f5f85a.png">

brendancarney:
I can't recreate this error. It may be resolved. Should we close the issue?

brendancarney:
Unable to recreate, please comment on the issue if you can and would like it reopened.

PR #2178, opened 2018-09-17
Gracefully handle permission denied error in FF when checking for element.nodeType

bjoerge:
Fixing a bug

#### What's the new behavior?
Not crashing in Firefox when clicking stepper arrows on number inputs (even outside of Slate itself) (#1819) and not crashing in embeds example (#843), and possibly providing a workaround for #1744.

This is a workaround for an issue with Firefox that causes `Restricted` nodes to throw on property access (e.g. `.nodeType`). I've tried to narrow the test to identify the particular situation as much as possible to avoid accidentally catching other errors that may happen here.


ianstormtaylor:
Thank you @bjoerge, this is a great pull request!

PR #2179, opened 2018-09-17
Fix deprecated .move > .moveForward in the after plugin

skogsmaskin:
Fixing a bug where a deprecated call to ``move`` was made in the After Plugin (onDrop). Should be ``moveForward``

#### What's the new behavior?
No new behaviour.

Fixes: (could not find any)
Reviewers: @

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -319,7 +319,7 @@ function AfterPlugin() {
       selection.end.key == target.end.key &&
       selection.end.offset < target.end.offset
     ) {
-      target = target.move(
+      target = target.moveForward(
         selection.start.key == selection.end.key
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/plugins/after.js
```diff
@@ -319,7 +319,7 @@ function AfterPlugin() {
       selection.end.key == target.end.key &&
       selection.end.offset < target.end.offset
     ) {
-      target = target.move(
+      target = target.moveForward(
```
ericedem: Might be cleaner to use backward here:
```
target = target.moveBackward(
  selection.start.key == selection.end.key
    ? selection.end.offset + selection.start.offset
    : selection.end.offset
)
```

ianstormtaylor:
Thanks @skogsmaskin!

PR #2181, opened 2018-09-18
Actually fix marks being cleared on selection events

Soreine:
I improperly rebased my fix in https://github.com/ianstormtaylor/slate/pull/2158 and was setting `marks` on the old `Range` objects instead of the new `Selection` object.


ianstormtaylor:
Thank you @Soreine, no worries.

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/commit/018ada8b88b9c5c2be18823c0e71c60fb8ff7bb5.

PR #2187, opened 2018-09-19
Refactor Editor Memoize interface

zhujinxuan:

Feature

I think memoize is fine, but the problem is that we are using `memoize-one` as the only memoize method.  `memoize-one` is good for `plugins`, `stack` and `schema`.  But for `value` and `change`, the result depends on how the context/process to generate value or change, it cannot be easily processed by `memoizeOne(...props)`, but more easier to be implement by `memoizeOne(updates)`.

Here we isolate the `value` and `change` in a different memoize process.  I set a cache `editor.memoized` and clear/reset the memoize cache for each update (by resetting at props change, componentDidUpdate, componentDidMount, `editor.onChange`)

Fixes: #
Reviewers: @ianstormtaylor @ericedem

### DIFF
--- a/packages/slate-react/src/components/editor.js
+++ b/packages/slate-react/src/components/editor.js
@@ -119,8 +119,7 @@ class Editor extends React.Component {
     this.tmp.updates++
 
     const { autoFocus } = this.props
-    this.resolveMemoize()
-    const { change } = this.memoized
+    const change = this.resolveChange()
 
     if (autoFocus) {
       if (change) {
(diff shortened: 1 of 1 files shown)


ericedem:
I could be missing something, but this doesn't seem to be solve the core problem, which is that access in getters shouldn't cause mutations and change events. The state of the app shouldn't change just by looking at a value.

zhujinxuan:
@ericedem Hi, the access of getter does not cause mutation.  The access of getter will trigger resolveMemoize, and resolveMemoize will initial memoize once and only once in each react lifecycle.

zhujinxuan:
Indeed, this PR is just to bring back the `componentWillReceiveProps` with an alias as `resolveMemoize`.

zhujinxuan:
Reopen trying to activate codecov report after rebasing

zhujinxuan:
This PR is no longer needed when we plan to have editor controller.

Issue #2136, opened 2018-08-28
add normalizing based on operations/paths


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now, the `deferNormalization` helper has bad performance because it doesn't have enough information to make targeted normalizations, so it has to do a depth-first search of the tree for nodes to normalize.

#### What's the expected behavior?

Now that we have `path` information and lookup methods though, this can be improved. We can look at each `Operation` and determine exactly which nodes it invalidates, to create a targeted group of nodes to run validation on again.

Further, this can actually clean up some of the internal logic which relies on the `normalizeParentByKey` or `normalizeAncestorsByKey` change methods, which are error prone if you choose the wrong one. Instead, all normalization can be performed automatically based on the operations that have already occurred.

For example...

```jsx
<document>
  <a />
  <b />
  <c />
</document>
```

If we use `insert_node` `<d>` at `[3]`, then we know that `<d>` and `<document>` need to be normalized. Or if we do `move_node` from `[2]` to `[0]`, we can know that just `<document>` needs to be normalized.

This can be taken further to solve https://github.com/ianstormtaylor/slate/issues/2135, because even inside `normalize` functions, each new operation that is applied is accounted for, and automatically re-processed afterwards. So that there are never nodes that escape normalization because of the order of processing the tree.

This will also make https://github.com/ianstormtaylor/slate/issues/2134 easy to implement, since the remaining operations will just be normalized in `editor.onChange`.

I propose we add a single change method:

```js
change.normalizeOperations()
```

Which looks at the existing `change.operations` and normalizes any "new" operations that haven't previously been accounted for.

Issue #2135, opened 2018-08-28
fix broken parent-child schema normalization


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

Schema normalization is currently broken in cases where normalizing a parent node results in changing one of its children, because it doesn't reprocess the child. And since helpers like `previous/next/first/last` are based on parent nodes, this happens more than you'd like.

For example, say you have a schema with:

```jsx
document: {
  nodes: {
    match: { type: 'a' },
  },
  normalize: (change, error) => {
    change.setNodeByKey(error.node.key, 'a')
  }
},
blocks: {
  a: {
    text: t => t === 'a',
    normalize: (change, error) => {
      change.setTextByKey(error.node.key, 'a')
    }
  },
  b: {
    text: t => t === 'b',
    normalize: (change, error) => {
      change.setTextByKey(error.node.key, 'b')
    }
  }
}
```

This requires that `<a>` nodes have text of `'a'`, and `<b>` nodes have text of `'b'`. And it also requires that documents only contain `<a>` nodes. Now imagine a document like:

```jsx
<document>
  <b>b</b>
</document>
```

Normalizing is depth first, so it will validate `<b>` and see that it is valid. Then it will move up the tree to validate `<document>` and see that it contains an invalid `<b>` node, and set the node to `<a>`. However, it won't realize that it needs to now re-validate the node since it's type changed. Resulting in an invalid final document of:

```jsx
<document>
  <a>b</a>
</document>
```

#### What's the expected behavior?

Normalization needs to know to re-process nodes that have changes applied to them during the normalization process, so that the normalizations can cascade. Otherwise you end up with an intractable issue of every `normalize` function being responsible for satisfying every possible schema rule's logic.

This will require writing more performant normalization logic, using operations from https://github.com/ianstormtaylor/slate/issues/2136.

Issue #2134, opened 2018-08-28
fix leaky schema normalization


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

Right now, schema normalization is leaky because it can actually still result in non-normalized documents when you don't expect it.

Ending a change with `normalize: false` results in non-normalized values. The `normalize: false` option is useful, but right now if you end a change with `normalize: false` this will result in invalid values because there's no fallback that ensures the value is normalized before rendering again. For example...

```js
editor.change(change => {
  change.removeNodeByKey(key, { normalize: false })
})
```

This breaks the assumption that we want to be true everywhere else, that you are assured that your schema rules will always be enforced.

#### What's the expected behavior?

We should add normalization logic to `editor.onChange` to ensure that any changes that pass through the editor are always normalized, even if the last change method was called with `normalize: false`.

This will require writing more performant normalization logic, using operations from https://github.com/ianstormtaylor/slate/issues/2136.

Issue #2132, opened 2018-08-28
Remove the validation optimization in insertTextAtRange


zhujinxuan:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

We have a validation normalization in https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/changes/at-range.js#L872 .   But this optimization has two problems:
1. it is a source of bugs as like https://github.com/ianstormtaylor/slate/pull/1823 .  And the code gets more complex to fix it.
2. It is hard to compat with text content match validation/schema

#### What's the expected behavior?

Perhaps to remove the optimization is a good choice?

PR #2193, opened 2018-09-21
refactor normalization to be operations-based

ianstormtaylor:

Improvement.

#### What's the new behavior?

This changes the normalization logic to be operations (and `key`) based, instead of the current logic which is more haphazard, and which has bugs that lead to non-normalized documents in certain cases.

Now, every time a change method is called, after it applies its operations, those operations will be normalized. Based on each operation we can know exactly which nodes are "dirty" and need to be re-validated.

This change also makes it easy for the `withoutNormalizing` (previously `withoutNormalization`) helper to be much more performant, and only normalize the "dirty" nodes instead of being forced to handle the entire document.

To accommodate this new behavior, the old "operation flags" have been removed, replaced with a set of more consistent helpers:

- `withoutNormalizing`
- `withoutSaving`
- `withoutMerging`

All of them take functions that will be run with the desired behavior in scope, similar to how Immutable.js's own `withMutations` works. Previously this was done with a more complex set of flags, which could be set and unset in a confusing number of different ways, and it was generally not very well thought out. Hopefully this cleans it up, and makes it more approachable for people.

We also automatically use the `withoutNormalizing` helper function for all of the changes that occur as part of schema `normalize` functions. Previously people had to use `{ normalize: false }` everywhere in those functions which was error-prone.

With this new architecture, you sure almost never need to think about normalization. Except for cases where you explicitly want to move through an interim state that is invalid according to Slate's default schema or your own custom schema. In which case you'd use `withoutNormalizing` to allow the invalid interim state to be moved through.

Fixes: #1363
Fixes: #2134
Fixes: #2135
Fixes: #2136
Fixes: #1579
Fixes: #2132
Fixes: #1657


CameronAckermanSEL:
This is the best day

docs/reference/slate/change.md
```diff
@@ -53,40 +53,50 @@ function onSomeEvent(event, change) {
...
+This method normalizes the document with the value's schema. This should run automatically-you should not need to call this method unless you have manually disabled normalization (and you should rarely, if ever, need to manually disable normalization). The vast majority of changes, whether by the user or invoked programmatically, will run `normalize` by default to ensure the document is always in adherence to its schema.
 
 > 🤖 If you must use this method, use it sparingly and strategically. Calling this method can be very expensive as it will run normalization on all of the nodes in your document.
 
-### `withoutNormalization`
+### `withoutNormalizing`
 
-`withoutNormalization(customChange: Function) => Change`
+`withoutNormalizing(fn: Function) => Change`
```
ericedem: Sorry if I missed previous discussion on this, but I wonder if it would make sense for this method to pass change as an argument so that they user doesn't have to rely on creating a closure. For example:

```js
import myChangeFunction from './changeHelpers';
change.withoutSaving(myChangeFunction);
```

Same for `withoutSaving()` and `withoutMerging`

CameronAckermanSEL: Hey @ericedem , this is how it works already. Thanks!

ericedem: @CameronAckermanSEL Oh yea good call, thanks!

matt-erhart:
In "version": "0.18.10", I'm getting:
slate-react.es.js:3580 Uncaught TypeError: change.withoutSaving is not a function
    at Object.onChange (slate-react.es.js:3580)
    at Stack$$1.run (slate.es.js:7333)

ianstormtaylor:
@matt-erhart it sounds like it might be that one of your dependencies is pegged to an old version?

PR #2197, opened 2018-09-21
Merge decorations from multiple plugins

alanctkc:
Previously, `decorateNode` would search the plugin stack for the first `decorateNode` handler that returned a value. With this change, it will call the `decorateNode` handler on every plugin in the stack and merge their results together.

The end result is that multiple plugins can provide `decorateNode` decorations to the same node.

cc @ianstormtaylor @zhujinxuan


ianstormtaylor:
Thanks @alanctkc!

Issue #2124, opened 2018-08-24
Unable to copy/cut when only void inline node is selected


uipoptart:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?

- Insert void inline node 
- Click on a void inline node it’s state to selected (in this case it gets a black outline)
- Use the keyboard or dropdown menu to copy 
- Paste
- The copied node is not actually copied to the clipboard. The same behavior happens with cut.

![copy](https://user-images.githubusercontent.com/1156239/44598762-ec7ff600-a788-11e8-998d-fc0b7cedac7c.gif)

Didn't include a fiddle as this happens in the current emoji example.
https://www.slatejs.org/#/emojis
Mac OS 10.13.6
Google Chrome 67.0.3396.99

Some background:
In our use case, the void node contains underlying data that customized by the user. Like the emoji, some of it is displayed by the inline for the user to see. This scenario makes just inserting another instead of copy/pasting much more time consuming. 
![copy](https://user-images.githubusercontent.com/1156239/44598608-6fed1780-a788-11e8-8b2a-64d75bd2e3a4.gif)

#### What's the expected behavior?
The inline node is copied (or cut) to the clipboard

uipoptart:
Looks like the issue is in `cloneFragment` where we use `value.document.getClosestVoid` to find the startVoid. If only a node is highlighted `getClosestVoid` returns undefined so the next line down returns: 
`if (native.isCollapsed && !startVoid) return `

I've also noticed if an inline node is last or first thing in a selection the text will copy but the node is left out, which might be another issue but seems related.

ianstormtaylor:
Due to the tangential discussion in #1744 I've rolled this back. We need to find a solution that solves this without breaking the existing behaviors.

laobeijing0806:
I have the same issue here, any updates?

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

Issue #2072, opened 2018-08-11
consider introducing an `<Island>` component


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Feature.

#### What's the current behavior?

Right now there are a few different situations where you want to include components directly inside the `contenteditable` editor element—menus, placeholders, etc. Sometimes you can get away with putting them in portals, but 

But there are some issues when the components are inside the `contenteditable`:

- The components's HTML will be copied in `text/html` data, which isn't good because it messes up users copying HTML content out of your editor.
- Clicking on the component won't actually select the nearest spot in the editor, which isn't great for placeholders which should be click-through.

#### What's the expected behavior?

I wonder if it might be good to introduce an `<Island>` component that allows you to embed these small islands of potentially-ineditable, potentially-uncopyable, etc. content into the editor.

It might look like:

```jsx
<Island
  editable={Boolean}
  copyable={Boolean}
  clickable={Boolean}
>
  ...
</Island>
```

Where they all default to `false`, but could be toggled to `true` if needed...

- `editable === false` would make it `contenteditable="false"`.
- `copyable === false` would remove its HTML content from the editor on copies and cuts.
- `clickable === false` would make it so that clicking on it focuses the nearest editable element in the editor, like a placeholder behaves. Whereas `clickable === true` makes it so that it prevents the `click` event after you've handled it, so that it doesn't end up stealing the focus from the editor?

brendancarney:
So I know they aren't the same thing, but I wonder if this could end up solving the problem I mention here: https://github.com/ianstormtaylor/slate/issues/2538

Could an `Island` potentially also contain normal slate components? The use case is toggling off the ability to edit a specific node depending on some state. Or is this outside the scope of this feature?

josephluck:
This really would be amazing. For example, I have a use-case where I would like to embed another editor (Monaco) in to a Slate document. Having an `<Island />` component would be perfect for my use case.

RavenColEvol:
@ianstormtaylor is this still a open Issue or should be closed?

PR #2198, opened 2018-09-22
Normalize copying void inline nodes

uipoptart:
fixing a bug

#### What's the new behavior?
1. Click on a void inline node to select it
2. Copy (or cut), it will be placed on the clipboard.

The fix can be seen in the emoji demo where you can use command+c and x to copy an emoji after clicking on it. 
When a lone void inline node is selected copy and cut should behave consistently with what happens when a void inline node is selected along with other text. Prior to this commit, selecting a void node along with other text would cause it to be copied/cut. However, selecting a lone void inline would not permit copying. This change makes the behavior consistent.

`isInEditor` would return `false` for anything marked `contenteditable=false` causing void inline nodes to always return false. This change returns `true` when the `contenteditable=false` element has a parent that is a void inline node. The event handler looks at `isInEditor` to decide if it will let the onCopy event propagate. 

* [x ] The new code matches the existing patterns and styles.
* [x ] The linter passes with `yarn lint`. (Fix errors with `yarn prettier`.)
* [x ] The relevant examples still work. (Run examples with `yarn watch`.)

Fixes: #2124

Also consider #2072, which calls for some changes that would allow for void inline nodes that can enable or disable copy abilities. This fix unifies the current experience, but longer term it would be rad to have the ability to explicitly decide this.

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -284,8 +284,10 @@ class Content extends React.Component {
 
       throw err
     }
+
+    const allowEdit = el.isContentEditable || el.closest('[data-slate-void]')
     return (
-      el.isContentEditable &&
+      allowEdit &&
       (el === element || el.closest('[data-slate-editor]') === element)
     )
   }


ianstormtaylor:
Due to the tangential discussion in #1744 I've rolled this back. We need to find a solution that solves this without breaking the existing behaviors.

PR #2203, opened 2018-09-24
Add codecov badage

zhujinxuan:
Add codecov badge to Readme page


ianstormtaylor:
Hey @zhujinxuan, I'd like to limit the badges to only the ones that are really useful to have a quick glance at. I'm not really happy with Codecov so far, so I don't want to add it to the Readme.

Issue #2123, opened 2018-08-24
operations/invert should not rely on internal state


bryanph:
#### Do you want to request a _feature_ or report a _bug_?
Feature/bug-ish

#### What's the current behavior?
The undo/redo stack makes use of the invert function here https://github.com/ianstormtaylor/slate/blob/ecf48926cc4e4485a31bff9a2224a4341c47ceff/packages/slate/src/operations/invert.js#L22 to invert the applied operations and apply them. This function replies on internal state which is removed here: https://github.com/ianstormtaylor/slate/blob/ecf48926cc4e4485a31bff9a2224a4341c47ceff/packages/slate/src/models/operation.js#L239-L251
 As a result, reverting operations on the user-side requires custom code to invert the given operations after this internal state is gone (like after a refresh).

#### What's the expected behavior?
My suggestion is to make operation.invert() just be a function of the document state and the raw operation, not on any internal state so that the case of inverting an operation from within the editor and from outside the editor is the same. This allows you to instantiate the history stack with operations that have been synced to a server on refresh and would make implementing a "version" example quite straight forward.

ianstormtaylor:
I think there's a nice property of being able to invert an operation without needing the original `value/document` that created it. But I agree that right now not all operations are living up to that goal. 

I'd be down with making that more the case, by standardizing extra properties on the operations that can allow us to remove these "stateful" ones. For example, I think the only reason `selection` is kept in `set_selection` operations is to have access to the "previous properties" so it can invert them. But we could instead standardize a `propertiesProperties` property with that information instead?

bryanph:
I agree, it's a nice property. Another option might be just to include those values in the JSON as well on perhaps a properties object as you suggest. Then the user can decide whether to store the more detailed version or the minimal one. 

Maybe we should just aim to include an example which demonstrates versions with slate? So an example where there are a few version states (references to Change objects) between which you can switch (and which roll back or apply changes as needed). Then we can decide what we need to do to get an example like that functional.

ianstormtaylor:
@bryanph that sounds good to me! I think the example should ideally use the lower-level `Operation` objects, and not need to keep `Change` objects around.

bryanph:
[...]

PR #2205, opened 2018-09-25
Add a version example

bryanph:
feature

#### What's the new behavior?
This is a first try at a version example as discussed in #2123.

This approach syncs with the history stack. It collects all operations since the last version (in this case it just takes the undo stack and reverses it) and applies them to allow you to switch between versions. 

This is an append-only example, so no branching.

Fixes: #
Reviewers: @


examples/versions/index.js
```diff
@@ -0,0 +1,282 @@
...
+    this.onChange(change)
+  }
+
+  /**
+   * Sets a version as the active version
+   *
+   * @param {Number} index
+   */
+
+  setVersion = index => {
```
zhujinxuan: A small suggestion.  Would you like to use `this.setState( state => )` to avoid calling `this.state`?  Or use 
```
this.setState(({ value, versions, activeVersionIndex }) => {...})
```

ianstormtaylor:
Very cool, thanks @bryanph!

PR #2207, opened 2018-09-25
Update products to include Cake

ericedem:
Now that Cake is coming out of invite only mode, figured it was a good time to add it as a product using Slate. :)

Link to https://www.cake.co/

### DIFF
--- a/docs/general/resources.md
+++ b/docs/general/resources.md
@@ -19,6 +19,7 @@ These tools are helpful when developing with Slate:
 
 These products are built with Slate, and can give you an idea of what's possible:
 
+* [Cake](https://www.cake.co/)
 * [GitBook](https://www.gitbook.com/)
 * [Grafana](https://grafana.com/)
 * [Guru](https://www.getguru.com/)


ericedem:
@oyeanuj I had to built something custom because there wasn't really anything out there that was still maintained and fully featured.

We just launched this feature, but I have a strong desire to open source it. It's almost there, but needs some TLC like API cleanup and some strong documentation / examples.

ianstormtaylor:
Awesome, thank you @ericedem! Congrats on the launch!

oyeanuj:
@ericedem Open-sourcing it would be amazing, it is the one functionality that there aren't many options for! 

If that is going to be a while, it will super useful to get a sense of the broader contours of how you set it up - inlines vs marks, data flow, etc.

And congrats on the launch!

ericedem:
@oyeanuj Yea it will probably be a while, we are a small team so until we grow a bit it will be hard  to maintain an open source project full time.

When I get some time I'll write something up about the approach we took.

oyeanuj:
> When I get some time I'll write something up about the approach we took.

Thanks, looking forward to it!

PR #2211, opened 2018-09-27
Fix crash when normalizing documents with multiple siblings.

soulplant:
Fixing a bug.

#### What's the new behavior?
Currently normalization crashes in documents like this:

```
<document>
  <invalid_node />
  <valid_node />
</document>
```

This is because <invalid_node /> will get deleted when it is normalized, which means that the original path to valid_node, [1], changes to [0]. However the current algorithm for normalization tries to access it at its original path, causing a crash.

Without this change normalization effectively works by creating a list of the paths of all the nodes that need to be normalized and then calls Node.normalizePath() on each one in turn, which causes problems when nodes disappear during normalization as those paths can become invalid.

This change fixes the problem by reversing the order in which sibling nodes are visited during the normalization process. This way nodes that get removed during normalization will not invalidate the paths of nodes yet to be normalized. Of course it's still fairly easy to write custom normalizers that thwart this process and cause crashes, but this change is intended to handle this edge case without changing the overall approach (the motivations for which I am not familiar with).

Note that this will cause any tests that assume a certain order of normalization to break (as you can see in the updated test case).

### DIFF
--- a/packages/slate-hyperscript/test/fixtures/cursor-across-blocks-and-inlines.js
+++ b/packages/slate-hyperscript/test/fixtures/cursor-across-blocks-and-inlines.js
@@ -39,7 +39,7 @@ export const output = {
         nodes: [
           {
             object: 'text',
-            key: '11',
+            key: '13',
             leaves: [
               {
                 object: 'leaf',
@@ -69,7 +69,7 @@ export const output = {
           },
           {
             object: 'text',
-            key: '12',
+            key: '14',
             leaves: [
               {
                 object: 'leaf',
@@ -88,7 +88,7 @@ export const output = {
         nodes: [
           {
             object: 'text',
-            key: '13',
+            key: '11',
             leaves: [
               {
                 object: 'leaf',
@@ -118,7 +118,7 @@ export const output = {
           },
           {
             object: 'text',
-            key: '14',
+            key: '12',
             leaves: [
               {
                 object: 'leaf',
--- a/packages/slate/src/interfaces/node.js
+++ b/packages/slate/src/interfaces/node.js
@@ -73,33 +73,65 @@ class NodeInterface {
   }
 
   /**
+   * Call f with the node and path of each node in the tree rooted here.
+   *
+   * Deeper nodes are visited before shallower nodes, and later siblings are
+   * visited before earlier siblings. This is a useful order because deleting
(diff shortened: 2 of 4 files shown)


ericedem:
@soulplant Hey this is a pretty cool idea. Is there any chance you could provide a fiddle with the problem you found? Trying to wrap my head around your use case, and I'm wondering if this might be a symptom of a larger problem with normalization. Thanks!

ericedem:
Hmm good points. Ignoring performance for a minute, it seems like what we really should have is a guarantee that on any given `normalize()` call, the state should at least be *consistent* even if not invalid.

It seems like if one node gets normalized, we can't reliably say that what we know about the whole document is consistent. And at that point we should restart the entire run and make sure paths are correct, before attempting to do any more normalization. This way we wouldn't need to care about what order normalizations are performed in, as long as they are consistent.

Intuition tells me that this would in general be fairly performant, since it is unlikely that many nodes would be normalized at the same time (causing an O(n^2) kind of behavior). Though I could be missing some particular complex use case.

So the procedure would be:

1. Fetch data about the current state and start the run.
2. Search for a node to normalize.
3. Perform normalization on the first node we find -> The run is invalidated GOTO 1.
4. Didn't find a node to normalize, the document is now in a valid state.

soulplant:
Yep, I think your intuition is right - normalize() would then be O(n*m) where n is the number of nodes in the document and m is the number of changes that will ultimately be required to normalize it. The common case is for there to be no changes required, which is O(n).

The algorithm you describe has the same runtime complexity as simply calling normalize() repeatedly until no more changes are performed. Shall we land this change, in order to prevent the currently existing crash, and then consider changing normalize() semantics to guarantee (for non-pathological cases) that normalize() will cause a document to become valid? I think the latter may have other subtle concerns and tradeoffs, so it may be worth considering it as a separate change.

ianstormtaylor:
@soulplant thank you for taking the time to work on this! And @ericedem thanks for the discussion which helped me understand more. I'm sorry about this bug, it was an oversight by me. I thought we could avoid the path transformation logic, but we can't.

I've just added #2222 which changes the logic to transform any existing dirty paths against any new operations that come in, so it should solve it for all cases, not just the back-to-front. Let me know if you run into any edge cases with that one and I can take a look!

Issue #1879, opened 2018-06-06
Chinese input - slate crashes with "Unable to find a DOM node..." in certain cases


k7d:
How to repeat (on OS X 10.13.4, Chrome 66):

1) Open [Slate example editor](http://slatejs.org/#/rich-text)
2) Select Pinyin - Traditional input source
    ![image](https://user-images.githubusercontent.com/181645/41053816-d9c20ef6-6979-11e8-9c02-c8a2ed6c5e8c.png)
3) In a new line type A, A, 1, then select the 2 just inserted characters and type A, A, 1 again
4) Slate crashes with `Error: Unable to find a DOM node for "10". This is often because of forgetting to add props.attributes to a custom component.`. See the gif below:

![composition-crash](https://user-images.githubusercontent.com/181645/41053924-2bb5d936-697a-11e8-9671-931d6ba9ed22.gif)

kulukyo:
I've the same issue. It happens when a whole selected line is about to be replaced.

![20180613_141832](https://user-images.githubusercontent.com/38679950/41333651-26b6186a-6f15-11e8-99c3-5f90bccea1b2.gif)

zhujinxuan:
@k7d @kulukyo I was investigating that issue.  The strange thing is that the chinese input method does not trigger onInput or onBeforeInput, but triggers onChange.

I will work on that when I have time.  Perhaps four weeks later at least.

yunwuxin:
It's still like that.

@ianstormtaylor

PR #2218, opened 2018-09-30
Prevent crash when handling composition when selection is not collapsed

k7d:

Fix 

#### What's the new behavior?

Doesn't crash when compositions starts on expanded selection.

When composition starts and the current selection is not collapsed, the
second composition key-down would drop the text wrapping <spans> which
resulted on crash in content.updateSelection after composition ends
(because it cannot find <span> nodes in DOM). This is a workaround that
erases selection as soon as composition starts and preventing <spans>
to be dropped.

Note that composition is still not handled quite right in Slate - when composing, the caret should be at the end of composition rather that at start. I might look into that later.

Fixes: https://github.com/ianstormtaylor/slate/issues/1879

### DIFF
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -173,6 +173,20 @@ function BeforePlugin() {
       editor.setState({ isComposing: true })
     }
 
+    const { value } = change
+    const { selection } = value
+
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @k7d!

Issue #2214, opened 2018-09-27
idea to reduce confusion of "before/after" plugins


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Idea.

#### What's the current behavior?

Right now we use "before" and "after" plugins in `slate-react` because we need to both preempt the userland plugins and provide default fallback behaviors. But this is kind of confusing, and although we can get away with it, userland plugins don't have the ability to easily do the same thing... since they'd need to explain to the user how to initialize both copies and place them in the correct order.

#### What's the expected behavior?

Something similar to this has been solved nicely by Koa. Where you can use the `next` argument to continue through the stack, and then choose provide more logic after the rest of the stack has decided whether to act or not. This would be useful for the `slate-react` case, as well as providing more helpful warnings for "unhandled" events, etc.

For example it might look like...

```js
onKeyDown(event, change, next) {
  // do some stuff...

  const handled = next()
  if (handled) return

  // otherwise, do some other stuff since it was unhandled
}
```

ianstormtaylor:
I think the one issue here is with the `handled` state. How is that actually determined? For different handlers it might be different. It could be via a return value that is a convention, like the current `return true`. But that would require people to obey the convention.

Koa on the other hand doesn't have this concept. It doesn't allow for the kind of communication between plugins, which might be smarter. Instead, you'd look to `change.operations` or similar, depending on what your own plugin considered "handling".

Then the question would be whether it's still worth it, if you can't actually have this "handled" state. I'm leaning towards yet, but not sure.

ianstormtaylor:
I also realized that this will enable creating "wrapper" renderers for nodes. For example, you might decide that you want to wrap all `image`, `video` and `audio` nodes in a "figure" component that renders their caption. Using `next`, you could now do this at the `renderNode` layer, instead of having to manage it in the React components themselves. Equivalent in this case, but it allows for plugins to be able to wrap these things themselves for more complex behaviors.

Issue #2206, opened 2018-09-25
mismatch between "editor" and "value" concepts


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Discussion.

#### What's the current behavior?

Just opening this up for discussion, since it's something that I think Slate hasn't really solved well so far, and leads to it being hard to decide in either direction for the API...

Right now there are two concepts vying for the "top spot" in the architecture:

1. `value` — which holds the document, the selection, the history and the schema. This can easily be created server-side (or wherever), since it's not coupled to the view layer at all. (I'd consider `change` to be equivalent to `value` here, since they hold all the same data.)

2. `editor` — which holds the plugins, the current value, and the schema to an extent. This is less easy to use server-side, since the editor is inherently tied to the view layer. (Although this need not always be the case.)

So far, it's been unclear which is the "primary" concept. For example... 

- Event handlers are passed both a `change` and an `editor` to allow the user to decide which they want to use. You normally use the `change`, but sometimes, for example if you want to trigger an async change later, you need to use the `editor`.

- For the idea of "commands and queries" from #2066, it's unclear whether these commands should live inside the schema (or similar) and be exposed as `change.command(...)`. Or whether, since they are tied to the plugins, they should be exposed as `editor.command(...)`. Both have tradeoffs.

- You could make the argument that `history` shouldn't really be inside the `value`, and instead be controlled by the `editor` which is the long-lived entity—which would seem totally reasonable to me.

- Right now there's a confusion in that the editor's plugins create a final `schema` object that gets set on the `value`. But the value can also be initialized with its own `schema`, which will be overwritten... and it's not clear how they interact.

#### What's the expected behavior?

I'm not totally sure. This is something that I've wrestled with, but never found a nice mental model for being clear about the distinctions between the two. I'd love for anyone with ideas to comment! (Even if you aren't sure yourself, anything is helpful!)

I think it's helpful to have a few goals:

- It would be good for "commands", "history", etc. to all be able to be modeled in server-side environments where there is no concept of the "view" layer. This doesn't necessarily mean it has to be the `value` model, but it has to be something that isn't tied to the DOM or rendering.

- It would be best if the top-level concept can unify the plugins, schema, history, commands, etc. in a way that makes it clear where all of this stuff stems from.

zhujinxuan:
> but sometimes, for example if you want to trigger an async change later, you need to use the editor.

Let editor.change return a Promise to force editor.change be asynchronously?

[...]

Issue #2199, opened 2018-09-24
slate-hyperscript ignores some empty texts


Soreine:
#### Do you want to request a _feature_ or report a _bug_?

Reporting a bug

#### What's the current behavior?

I can't create document like this out of the box, to get this, I need to run Slate's normalization:

```jsx
<paragraph>
  <text />
  <link>Some inline link</link>
  <text />
</paragraph>
```

Here is a reproducible example, using hyperscript

```js
import Slate from 'slate'
import hyperscript from 'slate-hyperscript'

hyperscript(
  'block',
  { type:'paragraph' },
  [Slate.Text.create(''), Slate.Inline.create({ type: 'link' }), Slate.Text.create('')]
)

/*
Block { 
  "data": Map {}, 
  "key": "15", 
  "type": "paragraph",
  "nodes": List [
     Inline { "data": Map {}, "key": "12", "nodes": List [], "type": "link" },
     Text { "leaves": List [ Leaf { "marks": Set {}, "text": "" } ], "key": "13" } 
  ]
}
*/
```

#### What's the expected behavior?

The first `Text` is not kept. It should be present in the output.

#### Possible source of the bug

I have looked into `slate-hyperscript` source, but I could not fix it. I just identified that during the creation of the block, the `createChildren` function is called with `[Text, Inline, Text]` but the first `Text` is never pushed into the final children array.

ericedem:
Probably related to: https://github.com/ianstormtaylor/slate/issues/2176

Soreine:
This bug exists in 34.7 as well

ericedem:
I did some debugging, and this is the spot where text nodes are getting normalized away:

https://github.com/ianstormtaylor/slate/blob/master/packages/slate-hyperscript/src/index.js#L356-L363

The general behavior seems to be to collapse any child text nodes together, and then only add them as children if there is actually text. I'm not familiar with the use cases for `slate-hyperscript` so I'm not sure what the actual behavior is supposed to be. Is the output always supposed to match the input?

Regardless @Soreine here is a gist for a fixture that tests the first broken case you mentioned which does fail:

https://gist.github.com/ericedem/66248a1e5fe707004b426c11a65cd153

Issue #2176, opened 2018-09-14
update slate-hyperscript to latest API


ericedem:
#### Do you want to request a _feature_ or report a _bug_?

_bug_

#### What's the current behavior?

When searching around the codebase, I noticed that `slate-hyperscript` package is still using a lot of old APIs that were deprecated in [0.37.0](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/Changelog.md#0370--august-3-2018) of slate. It should be updated to that latest.

PR #2221, opened 2018-10-02
Add controller

ianstormtaylor:
This is a structural change to the architecture of some of the core parts of Slate. Specifically it replaces the old "pseudo-models", in the `Stack`, `Schema` and `History` with a new `Editor` controller. In doing so it upgrades the plugin middleware stack to be more flexible. All with the goal of simplifying the core architecture, making Slate easier to use server-side, in tests and in other view layers, and making more of the core behaviors of Slate more customizable by plugins.

---

###### NEW

**Introducing the `Editor` controller.** Previously there was a vague `editor` concept, that was the React component itself. This was helpful, but because it was tightly coupled to React and the browser, it didn't lend itself to non-browser use cases well. This meant that the line between "model" and "controller/view" was blurred, and some concepts lived in both places at once, in inconsistent ways.

A new `Editor` controller now makes this relationship clear. It borrows many of its behaviors from the React `<Editor>` component. And the component actually just instantiates its own plain JavaScript `Editor` under the covers to delegate the work to. 

This new concept powers a lot of the thinking in this new version, unlocking a lot of changes that bring a clearer separation of responsibilities to Slate. It allows us to create editors in any environment, which makes server-side use cases easier, brings parity to testing, and even opens us up to supporting other view layers like React Native or Vue.js in the future.

It has a familiar API, based on the existing `editor` concept:

```js
const editor = new Editor({ plugins, value, onChange })

editor.change(change => {
  ...
})
```

However it also introduces imperative methods to make testing easier:

```js
editor.run('renderNode', props)

editor.event('onKeyDown', event)

editor.command('addMark', 'bold')

editor.query('isVoid', node)
```

I'm very excited about it, so I hope you like it!

**The `<Editor>` can now choose to not normalize on mount.** A nice side effect of splitting out the `Editor` logic into a reusable place is that it's easier to implement customizable behaviors for normalization. You can now pass an `options={{ normalize: false }}` prop to the React `<Editor>` which will disable the default normalization that takes place when the editor is constructed. This is helpful in cases where you are guaranteed to have an already normalized value, and don't want to incur the performance cost of normalizing it again.

**Introducing the "commands" concept.** Previously, "change methods" were treated in a first-class way, but plugins had no easy way to add their own change methods that were reusable elsewhere. And they had no way to override the built-in logic for certain commands, for example `splitBlock` or `insertText`. However, now this is all customizable by plugins, with the core Slate plugin providing all of the previous default commands.

```js
const plugin = {
  commands: {
    wrapQuote(change) {
      change.wrapBlock('quote')
    }
  }
}
```

Those commands are then available directly on the `change` objects, which are now editor-specific:

```js
change.wrapQuote()
```

This allows you to define all of your commands in a single, easily-testable place. And then "behavioral" plugins can simply take command names as options, so that you have full control over the logic they trigger.

**Introducing the "queries" concept.** Similarly to the commands, queries allow plugins to define specific behaviors that the editor can be queried for in a reusable way, to be used when rendering buttons, or deciding on command behaviors, etc.

For example, you might define an `getActiveList` query:

```js
const plugin = {
  queries: {
    getActiveList(value) {
     
    }
  }
}
```

And then be able to re-use that logic easily in different places in your codebase, or pass in the query name to a plugin that can use your custom logic itself:

```js
const { value } = change
const list = change.getActiveList(value)

if (list) {
  ...
} else {
  ...
}
```

Taken together, commands and queries offer a better way for plugins to manage their inter-dependencies. They can take in command or query names as options to change their behaviors, or they can export new commands and queries that you can reuse in your codebase.

**The middleware stack is now deferrable.** With the introduction of the `Editor` controller, the middleware stack in Slate has also been upgraded. Each middleware now receives a `next` function (similar to Express or Koa) that allows you to choose whether to iterating the stack or not.

```js
// Previously, you'd return `undefined` to continue.
function onKeyDown(event, change, editor) {
  if (event.key !== 'Enter') return
  ...
}

// Now, you call `next()` to continue...
function onKeyDown(event, change, next) {
  if (event.key !== 'Enter') return next()
  ...
}
```

While that may seem inconvenient, it opens up an entire new behavior, which is deferring to the plugins later in the stack to see if they "handle" a specific case, and if not, handling it yourself:

```js
function onKeyDown(event, change, next) {
  if (event.key === 'Enter') {
    const handled = next()
    if (handled) return handled

    // Otherwise, handle `Enter` yourself...
  }
}
```

This is how all of the core logic in `slate-react` is now implemented, eliminating the need for a "before" and an "after" plugin that duplicate logic.

Under the covers, the `schema`, `commands` and `queries` concept are all implemented as plugins that attach varying middleware as well. For example, commands are processed using the `onCommand` middleware under the covers:

```js
const plugin = {
  onCommand(command, change, next) {
    ...
  }
}
```

This allows you to actually listen in to all commands, and override individual behaviors if you choose to do so, without having to override the command itself. This is a very advanced feature, which most people won't need, but it shows the flexibility provided by migrating all of the previously custom internal logic to be based on the new middleware stack.

**Plugins can now be defined in nested arrays.** This is a small addition, but it means that you no longer need to differentiate between individual plugins and multiple plugins in an array. This allows plugins to be more easily composed up from multiple other plugins themselves, without the end user having to change how they use them. Small, but encourages reuse just a little bit more.

###### BREAKING

**The `Value` object is no longer tied to changes.** Previously, you could create a new `Change` by calling `value.change()` and retrieve a new value. With the re-architecture to properly decouple the schema, commands, queries and plugins from the core Slate data models, this is no longer possible. Instead, changes are always created via an `Editor` instance, where those concepts live.

```js
// Instead of...
const { value } = this.state
const change = value.change()
...
this.onChange(change)

// You now would do...
this.editor.change(change => {
  const { value } = change
  ...
})
```

Sometimes this means you will need to store the React `ref` of the `editor` to be able to access its `editor.change` method in your React components.

**Remove the `Stack` "model", in favor of the `Editor`.** Previously there was a pseudo-model called the `Stack` that was very low level, and not really a model. This concept has now been rolled into the new `Editor` controller, which can be used in any environment because it's just plain JavaScript. There was almost no need to directly use a `Stack` instance previously, so this change shouldn't affect almost anyone.

**Remove the `Schema` "model", in favor of the `Editor`.** Previously there was another pseudo-model called the `Schema`, that was used to contain validation logic. All of the same validation features are still available, but the old `Schema` model is now rolled into the `Editor` controller as well, in the form of an internal `SchemaPlugin` that isn't exposed.

**Remove the `schema.isVoid` and `schema.isAtomic` in favor of queries.** Previously these two methods were used to query the schema about the behavior of a specific `node` or `decoration`. Now these same queries as possible using the "queries" concept, and are available directly on the `change` object:

```js
if (change.isVoid(node)) {
  ...
}
```

**The middleware stack must now be explicitly continued, using `next`.** Previously returning `undefined` from a middleware would (usually) continue the stack onto the next middleware. Now, with middleware taking a `next` function argument you must explicitly decide to continue the stack by call `next()` yourself.

**Remove the `History` model, in favor of commands.** Previously there was a `History` model that stored the undo/redo stacks, and managing saving new operations to those stacks. All of this logic has been folded into the new "commands" concept, and the undo/redo stacks now live in `value.data`. This has the benefit of allowing the history behavior to be completely overridable by userland plugins, which was not an easy feat to manage before.

**The `editor` object is no longer passed to event handlers.** Previously, the third argument to event handlers would be the React `editor` instance. However, now that `Change` objects contain a direct reference to the editor, you can access this on `change.editor` instead. 

```js
function onKeyDown(event, change, next) {
  const { editor } = change
  ...
}
```

In its place is the new `next` argument, which allows you to choose to defer to the plugins further on the stack before handling the event yourself.

**Values can no longer be normalized on creation.** With the decoupling of the data model and the plugin layer, the schema rules are no longer available inside the `Value` model. This means that you can no longer receive a "normalized" value without having access to the `Editor` and its plugins.

```js
// While previously you could attach a `schema` to a value...
const normalized = Value.create({ ..., schema })

// Now you'd need to do that with the `editor`...
const value = Value.create({ ... })
const editor = new Editor({ value, plugins: [{ schema }] })
const normalized = editor.value
```

While this seems inconvenient, it makes the boundaries in the API much more clear, and keeps the immutable and mutable concepts separated. This specific code sample gets longer, but the complexities elsewhere in the library are removed.

**The `Change` class is no longer exported.** Changes are now editor-specific, so exporting the `Change` class no longer makes sense. Instead, you can use the `editor.change()` API to receive a new change object with the commands and queries specific to your editor's plugins.

**`slate-hyperscript` no longer normalizes values.** This behavior was very problematic because it meant that you could not determine exactly what output you'd receive from any given hyperscript creation. The logic for creating child nodes was inconsistent, relying on the built-in normalization to help keep it "normal". While this is sometimes helpful, it makes writing tests for invalid states very tricky, if not impossible.

Now, `slate-hyperscript` does not do any normalization, meaning that you can create any document structure with it. For example, you can create a block node inside an inline node, even though a Slate editor wouldn't allow it. Or, if you don't create leaf text nodes, they won't exist in the output. 

For example these are no longer equivalent:

```jsx
<document>
  <paragraph>
    <link>word</link>
  </paragraph>
</document>
```
```jsx
<document>
  <paragraph>
    <text />
    <link>word</link>
    <text />
  </paragraph>
</document>
```

Similarly, these are no longer equivalent either:

```jsx
<document>
  <paragraph />
</document>
```
```jsx
<document>
  <paragraph>
    <text />
  </paragraph>
</document>
```

This allows you to much more easily test invalid states and transition states. However, it means that you need to be more explicit in the "normal" states than previously.

**The `<text>` and `<mark>` creators now return useful objects.** This is a related change that makes the library more useful. Previously you could expect to receive a `value` from the `<value>` creator, but the others were less consistent. For example, the `<text>` creator would actually return an array, instead of the `Text` node that you expect.

```js
// Previously you had to do...
const text = <text>word</text>[0]

// But now it's more obvious...
const text = <text>word</text>
```

Similarly, the `mark` creator used to return a `Text` node. Now it returns a list of `Leaf` objects, which can be passed directly as children to the `<text>` creator.

**The `findRange`, `findPoint`, `cloneFragment`, and `getEventRange` utils now take an `editor`.** Previously these utility functions took a `schema` argument, but this has been replaced with the new `editor` controller instead now that the `Schema` model has been removed.

**The `getClosestVoid`, `getDecorations` and `hasVoidParent` method now take an `editor`.**  Previously these `Node` methods took a `schema` argument, but this has been replaced with the new `editor` controller instead now that the `Schema` model has been removed. 

**The `slate-simulator` is deprecated.** Previously this was used as a pseudo-controller for testing purposes. However, now with the new `Editor` controller as a first-class concept, everything the simulator could do can now be done directly in the library. This should make testing in non-browser environments much easier to do.

---

Fixes: https://github.com/ianstormtaylor/slate/issues/2045
Fixes: https://github.com/ianstormtaylor/slate/issues/2152
Fixes: https://github.com/ianstormtaylor/slate/issues/2066
Fixes: https://github.com/ianstormtaylor/slate/issues/2214
Fixes: https://github.com/ianstormtaylor/slate/issues/837
Fixes: https://github.com/ianstormtaylor/slate/issues/2199
Fixes: https://github.com/ianstormtaylor/slate/issues/2176
Fixes: https://github.com/ianstormtaylor/slate/issues/2206

Addresses: https://github.com/ianstormtaylor/slate/issues/1013
Addresses: https://github.com/ianstormtaylor/slate/issues/1411
Addresses: https://github.com/ianstormtaylor/slate/issues/2098#issuecomment-424483484
Addresses: https://github.com/ianstormtaylor/slate/issues/1456
Addresses: https://github.com/ianstormtaylor/slate/issues/1730
Addresses: https://github.com/ianstormtaylor/slate/issues/1311
Addresses: https://github.com/ianstormtaylor/slate/issues/2190


packages/slate/src/controllers/editor.js
```diff
@@ -0,0 +1,312 @@
...
+
+    // If the change doesn't define any operations to apply, abort.
+    if (change.operations.size === 0) {
+      return
+    }
+
+    // Store a reference to the last `value` and `plugins` that were seen by the
+    // editor, so we can know whether to normalize a new unknown value if one
+    // is passed in via `this.props`.
+    this.tmp.lastValue = change.value
```
ericedem: Is `this.tmp.lastValue` always normalized? If so, could we call this `lastNormalizedValue`?

packages/slate/src/controllers/editor.js
```diff
@@ -0,0 +1,312 @@
...
+
+  setValue(value, options = {}) {
+    const { normalize = true } = options
+
+    // PERF: If the plugins and value haven't changed from the last seen one, we
+    // don't have to normalize it because we know it was already normalized.
+    if (
+      normalize === false ||
+      (this.plugins === this.tmp.lastPlugins &&
+        this.value === this.tmp.lastValue)
```
ericedem: Maybe I'm not understanding, but shouldn't this be checking against the value being passed in? Like: `value === this.tmp.lastValue`?

packages/slate-react/src/components/editor.js
```diff
@@ -153,260 +140,93 @@ class Editor extends React.Component {
...
     debug('render', this)
+    const { controller, tmp } = this
     const props = { ...this.props }
-    const tree = this.stack.render('renderEditor', props, this)
-    return tree
+
+    // Update the props on the controller before rendering.
+    const { options, readOnly } = props
+    const plugins = this.resolvePlugins(props.plugins, props.schema)
+    const value = tmp.change ? tmp.change.value : props.value
```
ericedem: The editor will only rerender if there is a props change, should we only render prop.value here instead? Or maybe `tmp.change` should be `state.change` if we want a render cycle to happen.

ianstormtaylor: @ericedem not sure I understand what you mean. I believe here `tmp.change` should only be present if we needed to normalize the `value` before the component was mounted and ready to receive changes. So it's just saying that if there's an interim value, use it, since it will soon be flushed via `onChange`.

ianstormtaylor: Op! Sorry, I just got what you mean now. You're right, since we're already rendering via `editor.value` we're already achieving this, and the `tmp.change` logic here isn't doing anything!

packages/slate/src/controllers/editor.js
ianstormtaylor: Yup, you're right! Good catch.

ianstormtaylor:
@ericedem thanks, that is super helpful!

bryanph:
Damn that's quick! Looking forward to a writeup

packages/slate-react/src/plugins/browser.js
```diff
@@ -23,29 +26,46 @@ import setEventTransfer from '../utils/set-event-transfer'
...
  * @return {Object}
  */
 
-function AfterPlugin() {
+function BrowserPlugin() {
+  let activeElement = null
+  let compositionCount = 0
+  let isComposing = false
+  let isCopying = false
+  let isDragging = false
```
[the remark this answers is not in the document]
ianstormtaylor: @CameronAckermanSEL could you give me an example of what you're thinking about?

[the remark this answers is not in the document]
CameronAckermanSEL: Thinking more on this, I believe the needs of that use case were met. I was having an issue suppressing drag events originating from very specific selections in the editor, and needed some way of determining whether this was an "internal" drag and drop as opposed to a drag and drop from an external source. At the time I had to suppress drop in general, but I think here I'd be able to get by with the plugin delegation via next() in order to suppress the event conditionally.

bryanph:
First of all, this is really cool! Here are some comments/questions:

> **Introducing the "queries" concept.** Similarly to the commands, queries allow plugins to define specific behaviors that the editor can be queried for in a reusable way, to be used when rendering buttons, or deciding on command behaviors, etc.
> 
> For example, you might define an `getActiveList` query:

How will we avoid name clashes on the change object?

> **The `Value` object is no longer tied to changes.** Previously, you could create a new `Change` by calling `value.change()` and retrieve a new value. With the re-architecture to properly decouple the schema, commands, queries and plugins from the core Slate data models, this is no longer possible. Instead, changes are always created via an `Editor` instance, where those concepts live.

Does this mean we can make the Change object immutable? As in, just apply the return value of the `editor.change` function?

> Sometimes this means you will need to store the React `ref` of the `editor` to be able to access its `editor.change` method in your React components.

This sounds a bit overly complicated because this is a very common use case I feel.

Could you update one of the examples so we can play around with it and get a feel for how the new API might work in practice?

skogsmaskin:
I like these changes, but I'm also a little worried by the need to have the editor controller object in order to get a new change object. Doesn't it mean that a change will have to be applied to the editor before we are able to build further on that change? I don't think that is always what you want.

We have this from before:
```
change.doSomething()
// Then later:
change.value.change().doSomethingElse
```
We know for sure that we can build further on the change.

Will this still be the case using the ``editor.change``? How do we know that something has not changed the value in the meantime?

CameronAckermanSEL:
I was initially also concerned about the need to have an editor in order to get a chance object (as opposed to getting one from a value), but in all cases I was calling `value.change()` it was for a toolbar where I already had the editor on hand. I'm not sure how other folks are using `value.change()` though. It does seem a little awkward but I think it's one of those things where one can get used to it.

I'm also really supportive of the idea of the change methods being pure (as in they emit a new change object). It always struck me as odd that the change methods functions in the previous implementation were mutators and they returned a new change object (I think the latter was to drive change method chaining).

ianstormtaylor:
@bryanph @skogsmaskin @CameronAckermanSEL thanks for the replies!! I'll go through all your questions/comments and try to give answers. Please feel free to do another round!

> How will we avoid name clashes on the change object?

Under the covers, the way that commands and queries are implemented in the editor is using the same middleware stack logic as other handlers. So there's an `onCommand` and `onQuery` hook that middleware can handle. Technically, there's no actually need for a command to be registered with the editor and used as a method on the change object for them to be fired. You could always do:

```js
editor.command('someCustomCommand', ...args)
```

And if an `onCommand` middleware happens to handle it, power to them. (A single middleware could handle all commands, or many commands, or just a single one.) The signature is `(command, change, next)` where `command` has `{ type, args }`.

However! Although you could do everything with `editor.command`, I see the `Change` object as a way to make the API easier to use. So to that end, you can "register" commands which creates the helper methods directly on the change object. All this does is essentially:

```js
Change.prototype[type] = (...args) => editor.command(type, ...args)
```

So in terms of the change object themselves, there's technically no clashes, since registering a command doesn't actually describe its behavior—it still just calls into the editor middleware to find one to handle the command. 

So the way precedence works is that the first plugin to have registered `commands` would end up handling them, and not proceeding to future ones in the stack. So to have priority, you need to be earlier in the plugin stack, like with the other handlers.

Does that make sense?

> Does this mean we can make the Change object immutable? As in, just apply the return value of the editor.change function?

> I'm also really supportive of the idea of the change methods being pure (as in they emit a new change object). It always struck me as odd that the change methods functions in the previous implementation were mutators and they returned a new change object (I think the latter was to drive change method chaining).

We actually used to have this. I think the thing that made me not as fond of it is that for changes, we're guaranteed to be in a position where we are applying multiple commands to be packaged up into a single change. And when you know you want to make a bunch of interim changes to something, you usually don't want to deal with the overhead of creating a bunch of immutable interim states.

It's kind of the same thing as Immutable.js's `withMutations`. The way I see it, `change` objects are essentially a "with mutations" but for the editor's value itself. So it feels weird to use immutability where we are forced to create a bunch of interim states, when really we only care about the final one being immutable.

Also, it prevents chaining. Which is less important, but still convenient I think.

Does that make sense? 

> I like these changes, but I'm also a little worried by the need to have the editor controller object in order to get a new change object. Doesn't it mean that a change will have to be applied to the editor before we are able to build further on that change? I don't think that is always what you want.
>
> We have this from before:
>
>     change.doSomething()
>     // Then later:
>     change.value.change().doSomethingElse()
>
> We know for sure that we can build further on the change / value. And we don't need to involve the editor.
> 
> Will this still be the case using the editor.change? How do we know that something has not changed the value in the meantime and the change returned from editor will be invalid to apply the planned changes on? I'm thinking async operations etc. Not sure if I'm totally making my point clear here.

@skogsmaskin When do you have that? If you have a reference to a change object, that object lives for the duration of the function, so you can continue adding changes to it. And if you want to make a new change object, that's when you need an `editor` reference. But you can still (just as now) continue making changes with the same `change` reference.

I'm not sure I understand your use case there at all. It sounds kind of convoluted, but I'm not sure what the patches are, or what the goal of that function is.

> I was initially also concerned about the need to have an editor in order to get a chance object (as opposed to getting one from a value), but in all cases I was calling value.change() it was for a toolbar where I already had the editor on hand. I'm not sure how other folks are using value.change() though. It does seem a little awkward but I think it's one of those things where one can get used to it.

@CameronAckermanSEL Nice, this is my understanding too. There are very few places right now that don't already have access to the `editor`.

> This sounds a bit overly complicated because this is a very common use case I feel.
>
> Could you update one of the examples so we can play around with it and get a feel for how the new API might work in practice?

@bryanph the example are updated with this. You can check out the `rich-text` one which uses the `ref` when its toolbar buttons are pressed.

ianstormtaylor:
@skogsmaskin thank you, that's super helpful! I think I understand now. I'll explain how I see it, and then how'd you might handle your situation.

There are a few different ways to manipulate a Slate `value`, from low-level to high-level:

1. You can call the super-low-level methods on the `Value`, like `value.insertText`. These are actually what the operations use under the hood. This is similar to how all of the models have mutating methods, like `text.splitLeaves` or whatever.

2. You can apply operations to the value. This is slightly higher level, but still very low. But since operations are schema-agnostic, it's easy to do.

3. You can use the `Change` object to apply semantic groups of operations (which used to be called "change methods", but are now called "commands"), which are defined in terms that developers can more easily understand, and that mimic user behaviors. These commands are schema-aware, because they keep the value "normalized".

The **1** approach is so low-level that it's rarely used. Although if you're applying operations by hand, it might be easier to just use the methods directly. Higher up, **2** is used when you want fine-grained control—for example in operational transform for collaborative editing. Neither of these are schema-aware.

However, **3** is problematic because it is schema-aware. (This was true before this PR, and it true after too.) Which means that you need to know all of the plugins to determine the proper schema. Which means you need more than just the "pure data" in the value object to properly perform a command.

The issue was that previously, the `<Editor>` would attach its own `schema` to the value. And then while commands were executed, they would look at `value.schema` to see how to normalize properly. But this only worked seamlessly on the client-side with React. On the server-side (or places without React), you'd have to instead initialize the value with the schema directly to get proper normalization. It was the same outcome, but handled in two different ways...

```js
// Previously, with React...
<Editor plugins={plugins} value={value} ... />

editor.change(change => {
  ...
})

const newValue = editor.value
```
```js
// Previously, without React...
const schema = Schema.create({ plugins })
const value = Value.create({ schema, ... })
const change = value.change()
...

const newValue = change.value()
```

But this difference in approach was convoluted, and required schema's to be initialized in multiple ways, and treated some of the core logic as more privileged than plugins, which made it hard to override—or even hard to find the right abstraction in the first place.

Instead, now those two concepts are handled very similarly:

```js
// Now, with React...
<Editor plugins={plugins} value={value} ... />

editor.change(change => {
  ...
})

const newValue = editor.value
```
```js
// Now, without React...
const editor = new Editor({ plugins, value })

editor.change(change => {
  ...
})

const newValue = editor.value
```

This is made possible by folding all of the previously duplicated, separated logic into a single `Editor` controller than handles it consistently.

---

For your use case, you can still choose to either use the operation-based approach from **2**, or now the editor-based approach from **3**. If you need your changes to be schema-aware, and to be properly normalized after making them, you have to use **3**. Otherwise, if you're dealing with low-level mutations you can use **2** (or even **1**).

Now there's no question that if you need to use "commands" which result in normalization, you need to initialize an `Editor` with your plugins. (You could still choose to use no plugins when initializing, which just gives you core behavior.)

Does that make sense?

I realize that it seems like the **3** got harder. But really it's just because we removed some of the magic that was happening only in the React world (which wasn't obvious, and would result in unnoticed bugs and incompatibilities) and streamlined it to make it the same everywhere now.

ericedem:
This is really great, looking forward to this!

On the logistics side, is there anything you'd like from the community in terms of testing before this gets rolled out? Given how major this change is, it seems likely there will be a lot of bugs in the semver this lands in (though this will should fix a bunch too).

ianstormtaylor:
@ericedem thanks! Nothing special compared to normal releases with breaking changes I think.

justinweiss:
Would it be possible or worthwhile to be able to generate a new `editor`, cloning the same properties as an old one?

`(new Editor(this.editor)).change(c => c.insertText()).value` wouldn't be much worse than `value.change().insertText().value`, if you just want to modify a value with some changes.

ianstormtaylor:
@justinweiss I'm not sure what you mean. Could you give me a code sample or use case?

justinweiss:
@ianstormtaylor Never mind! I thought I was using `value.change()` more than I actually was. Nothing looks like it would be too difficult to move to `Editor`.

amitm02:
Look awesome @ianstormtaylor!
On my current project, the Value is always stored in a Redux state container. Any change to the Value state in the store results in a setState() that re-renders the React editor.
The triggers for changing the Value can come from either the React Editor, or from other external sources. 
So, while some of the changes to the Value are made via the react editor, others are made outside of the react editor, via `value.change()` and the result is stored in the Redux directly.

What is the best approach to handle it with the new version now that `value.change()` is no longer available?
Do i need to create a second "mocked" editor for the externally triggered changes?

ianstormtaylor:
@amitm02 yup, I'd do it exactly as you said. Create another non-React `Editor` instance with the same plugins if you need to edit it outside of the React editor. (Or with only the core plugins if you don't care about mismatches.)

skogsmaskin:
@ianstormtaylor - Man! After having refactored our code during the weekend for the new release, I just want to say that this change is really awesome! There were a lot to refactor though (we relied on ``Value.change()`` a lot for instance), but everything is getting so much cleaner now. I especially love the fact that I can override built in stuff like `splitBlock` which was a real troublemaker for us (various hacks can now be removed). Also love the new query-interface which let me remove a lot of code around the codebase, and have in one central place. Great work 👍

ianstormtaylor:
@skogsmaskin thank you, that means a lot to me. I get a bit nervous sometimes to make breaking changes, but it’s really nice to hear that they laid off like we thought!

Issue #2215, opened 2018-09-28
The normalization map should be invalidated when a node normalizes


ericedem:
#### Do you want to request a _feature_ or report a _bug_?

_bug_

#### What's the current behavior?

After a node is normalized, it may have removed itself, siblings, or generally changed the state of the document to where the normalization "map" is no longer valid.

#### What's the expected behavior?

After a node normalizes the normalization should abort until a new map that is consistent with the current state of the document should be found.

I don't think we had this problem until we started using paths instead of keys, since if we couldn't find a node anymore we would just abort.

There is some more detail / discussion in this pr https://github.com/ianstormtaylor/slate/pull/2211

ericedem:
Something else we could do with this is we could keep track of which key a path is "supposed" to have.  If we can't find a node at the path we are supposed to, or if the node we found doesn't have a key matching what we expected, we can use the key to find the correct node again.

Issue #2194, opened 2018-09-21
use operational transform to track dirty nodes


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now we use a key-based approach to keep track of dirty nodes that need to be re-normalized when operations affect them. This works, but keys are always slower than using paths directly.

#### What's the expected behavior?
If we had an `Operations.transform` helper with proper operational transform support, we could use it instead for what should be a performance increase.

PR #2222, opened 2018-10-02
Fix normalizing dirty paths

ianstormtaylor:

Fix.

#### What's the new behavior?

The dirty paths in a change are now transformed against incoming operations, such that they don't get out of sync as normalizations occur. This is a rough pass to get correctness and the bug fixed, and we can later optimize lots of the little details for performance.

Fixes: #2211
Fixes: #2215
Fixes: #2194


ericedem:
@ianstormtaylor This is great! Thanks for fixing this.

soulplant:
@ianstormtaylor Thanks for the fix!

Could I suggest adding the test case from #2211 to the test suite to protect against similar bugs in the future?

Issue #2192, opened 2018-09-20
make `operation.newPath` consistent


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now there are two properties involved in a `move_node` operation: the `path` and the `newPath`. The `path` makes sense, since it refers to the node in the current document.

But the `newPath` is confusing, because it currently refers to the new location of the node, but in respect to the current (old) document. This is usually fine, but gets tricky in cases of siblings. For example, given a document of:

```jsx
<document>
  <paragraph key="a">
    one
  </paragraph>
  <quote key="b">
    <paragraph>two</paragraph>
  </quote>
</document>
```

To arrive at:

```js
<document>
  <quote>
    <paragraph>two</paragraph>
    <paragraph>one</paragraph>
  </quote>
</document>
```

You use an operation of:

```js
{
  type: 'move_node',
  path: [0],
  newPath: [1, 1],
}
```

Where `[1, 1]` refers to the location the node would be move to relative to the current document, **but without removing the node from its original location**. However, when you look at the end state, you'll notice that the node doesn't end up at `[1, 1]`, but at `[0, 1]` on account of removing it earlier in the document and shifting everything backward one space.

This seems confusing?

#### What's the expected behavior?

Change the `newPath` to refer to the location in the document **after** the original node has been removed, so that it always ends up where `newPath` says it will be?

**EDIT:** Based on the conversation below, we should go with "**before** the original node has been removed" always instead.

justinweiss:
I think this is true sometimes, but not all the time.

This might have changed, but when I last worked on this, moving a node under the same parent would act as if you had removed the node first, but other moves would act as if you hadn't. For example: `move_node` with `path: [2]` and `newPath: [4]`, the node would end up in position `[4]` in the new document, not position `[3]`. Moves to new parents worked the way you described.

I don't mind using either the new tree or the old tree, but I think all moves should be consistent. For what it's worth, I wrote a helper function to rewrite all the moves to use old-tree-positions, because I found them easier to think about.

ianstormtaylor:
@justinweiss thank you! That's helpful.

> For what it's worth, I wrote a helper function to rewrite all the moves to use old-tree-positions, because I found them easier to think about.

What was it that made them easier you think?

justinweiss:
You only have to keep one operation in your mind, instead of two. 

[...]

PR #2225, opened 2018-10-04
Make all ops invertible and remove `value` from ops

bryanph:
both kind of

#### What's the new behavior?
* remove `value` from ops
* make all `set_x` ops keep track of previous properties in `properties`

The reason for this change is that right now operations can't be inverted if you convert them to JSON and back. This change adds the required attributes to make all operations invertible even after serializing/unserializing them. It also removes `value`, which causes high memory usage.

Fixes: #2123
Fixes: https://github.com/ianstormtaylor/slate/issues/2272
Reviewers: @


bryanph:
I have another branch that puts these values on a separate invertProps property but I figured might as well serialize them and let the user decide if they want to store the full object or not. What do you think?

bryanph:
Maybe this should actually be put optionally on the operation because as it is right now it causes errors when you try to parse fromJSON an operation that is not invertible. Maybe we can introduce a "invertible" property on an operation that indicates whether we can invert it or not (if the right props were passed in fromJSON)? Or perhaps operations should be invertible by default?

ianstormtaylor:
@bryanph I agree that we should strive to be able to invert all of them. 

I think one issue is that for inversion we're currently storing things like `node`, `mark`, and `value`. But these can be big objects (well, `value` can). And it would instead be better to store it as `previousProperties`, so that we only store the minimum information we need to invert them. 

Just because I don't think we can realistically get away with serializing `value` to JSON for `set_value` operations. Especially since they only need the `value.data` and `value.decorations` properties. (And in the future I think I'd like to see decorations split out into `add_decoration`, `remove_decoration` and `set_decoration` anyways, so it would just be for `value.data`.)

bryanph:
@ianstormtaylor Cool, so serializing it to a previousProperties object it is. Although serializing is never actually done in slate right, just when the user decides he needs to store the operations on a server? 

* Should we have a previousProperties object internally as well, or just when serializing? 
* Shall we have an "invertible" property for Operation so we can determine whether it is invertible or not (the required previousProperties are set)?

ianstormtaylor:
@bryanph I think it would be good to use the `previousProperties` ourselves in Slate for the inverting logic. Otherwise we'll end up with multiple code paths to have to maintain. I'm pretty sure that's all we're using these existing object references for anyways.

And then we can eliminate the need for `invertible`.

bryanph:
@ianstormtaylor great, thanks! That makes a lot of sense. I'll work on it. So when constructing an Operation object, do we directly pass `previousProperties`? Right now all `Operation.create()` calls have a flat structure which includes (sometimes) the properties required to invert. It might have a pretty significant impact.

ianstormtaylor:
@bryanph not sure I totally understand, but yeah I think we'd pass in `previousProperties` each time when creating the operation.

ianstormtaylor:
@bryanph good questions. It does seem like we have some inconsistencies that we should clear up while fixing this. Given a few different things...

---

I think for the `remove_*` operations, we always need the previous value, whatever it is. So for example, right now with `remove_text` we include `text` I believe. Similarly, we need to include the `node` and `mark` for `remove_node` and `remove_mark`:

```js
{
  type: 'remove_text',
  path: Array,
  offset: Number,
  text: String,
}
```
```js
{
  type: 'remove_node',
  path: Array,
  node: Node,
}
```
```js
{
  type: 'remove_mark',
  path: Array,
  offset: Number,
  length: Number,
  mark: Mark,
}
```

We could opt to make these `previousProperties` instead, but I think it's okay to go with a full representation in these cases where it's a full removal.

---

With the `set_*` operations however, we don't want to go with full representations, because we'd be duplicating a lot of stuff (which we are currently doing). For example, `set_value` includes the entire value object which is obviously bad. But even `set_node` including the entire node is bad because `node.nodes` can be huge too.

So for these, I think it's best to use a properties-only approach.

```js
{
  type: 'set_node',
  path: Array,
  properties: Object,
  newProperties: Object,
}
```
```js
{
  type: 'set_mark',
  path: Array,
  offset: Number,
  length: Number,
  properties: Object,
  newProperties: Object,
}
```
```js
{
  type: 'set_value',
  properties: Object,
  newProperties: Object,
}
```

I've opted to put the **old** values in `properties`, and the **new** values in `newProperties` for two consistency reasons:

- In the `move_node` operation, we already use the pattern of `new*` with `newPath`.
- In the `merge_node` operation, we already use `properties` for the old properties.

The properties objects are the same objects that you'd use for commands like `setBlocks`. The top-level keys in the objects are the same keys that nodes/marks/values have.

justinweiss:
It's been a long time since I looked at this, but are we doing something similar with `set_selection` now? I remember having to go into the `selection` property to find the previous selection data when inverting it, and `selection` wasn't serialized. If not, it would be nice to see `previousProperties` on `set_selection` as part of this change.

ianstormtaylor:
@justinweiss you're absolutely right, I forgot `set_selection` as one of the `set*` operations. I'll update the previous comment with it now for completeness.

bryanph:
Thanks guys, that clarifies it a lot. 

@ianstormtaylor I am not very familiar with the way marks work. Marks are differentiated using the "type" property right? it seem to me that for set_mark we still need to pass the mark since any portion of text can contain multiple marks?

bryanph:
@ianstormtaylor @justinweiss I changed all ops except for set_mark because it wasn't clear to me how the merging should occur when there is no mark passed. Also, I tried partially storing the previous selection but this caused two tests to fail with regards to deleting a selection between blocks and undoing, so for now I am just passing the whole previous selection: see https://github.com/ianstormtaylor/slate/blob/995ed2905b52b8e723e7260269049bf95a907526/packages/slate/src/commands/on-selection.js#L586

ianstormtaylor:
@bryanph awesome!

That's a good point about marks. They are currently de-duped (or updated) based on matching the `properties` of two marks, so see if they are referring to the same one. So I guess in that case we will always need to have the `properties` (and `newProperties`) contain both the `type` and `data`, even if only one is being updated? That way we can always fully compare them to see which mark it is?

For the `toJSON`, I believe that is because Lodash doesn't work with the immutable objects? But I might not be reading it correctly, so let me know.

bryanph:
@ianstormtaylor yes that sounds reasonable. Right now methods like `Text.updateMark` and `Leaf.updateMark` take an actual mark instance as one of their arguments. By changing to properties and nextProperties nothing would really change under the hood but it would be nice for consistency I suppose. I'll give it a shot.

bryanph:
@ianstormtaylor I updated set_mark to take `properties` and `newProperties` as params and made sure the `properties` param is a full Mark instance. This makes it different from all other `set_x` operations where those other operations can partially include the previous operation, `set_mark` requires `properties` to include a full `Mark` instance (or an object with all the properties required to create one)

justinweiss:
Awesome! All else being equal, I actually prefer keeping around the entire previous selection properties, rather than merging them when you need them -- it's not much extra data, but it makes working with them easier.

ianstormtaylor:
@bryanph this is amazing, thank you! It all looks good to me now.

I'd like to merge this one at the same time as the other operation-level changes we want to make, so that we can make fewer breaking releases to the operations logic. Specifically I'm thinking about https://github.com/ianstormtaylor/slate/issues/2192

bryanph:
@ianstormtaylor do you think we can get this merged anytime soon?

brendancarney:
I thought I'd share some memory profile testing I've been looking at. Both of these examples are for ~1200 line plain text paste. This change makes some large improvements:

**Before:**
before

**After:**
after

Note that the paste is still dramatically slow, but these memory improvements are still a huge win here and in other areas.

bengotow:
Hey folks! Hope you're having a great weekend. Anything we could do to help get this merged? I think this will address the memory leaks in #2456 and it'd be great to ship it. 🙏

ianstormtaylor:
Absolutely perfect pull request @bryanph, thank you very much for doing it. I love the debt it clears away for us. I'm sorry it took me so long to merge.

Issue #2223, opened 2018-10-04
Strange issue in value.activeMarks getter


zarv1k:
#### Do you want to request a _feature_ or report a _bug_?

a bug

#### What's the current behavior?

I've prepared sandbox on [JSFiddle](https://jsfiddle.net/zarv1k/u3fbspm1/88/).
Short description: assume that there is 'bug' word in the editor.

Letter 'b' marked by bold, italic and undernline marks.
Letter 'u' marked by italic and undernline marks.
Letter 'g' marked by italic mark only.

Selection of the whole word leads to crash with the following error:
<details><summary>Stacktrace</summary>

```
Uncaught TypeError: Cannot read property '1' of undefined
    at immutable.js:1377
    at ArrayMapNode.iterate.HashCollisionNode.iterate (immutable.js:1726)
    at Map.__iterate (immutable.js:1375)
    at Set.__iterate (immutable.js:3905)
    at Set.forEach (immutable.js:4381)
    at immutable.js:3859
    at Set.Map.withMutations (immutable.js:1353)
    at Set.intersect (immutable.js:3858)
    at slate.js:1893
    at List.__iterate (immutable.js:2206)
(anonymous) @ immutable.js:1377
ArrayMapNode.iterate.HashCollisionNode.iterate @ immutable.js:1726
Map.__iterate @ immutable.js:1375
Set.__iterate @ immutable.js:3905
forEach @ immutable.js:4381
(anonymous) @ immutable.js:3859
Map.withMutations @ immutable.js:1353
Set.intersect @ immutable.js:3858
(anonymous) @ slate.js:1893
List.__iterate @ immutable.js:2206
forEach @ immutable.js:4381
(anonymous) @ slate.js:1892
Map.withMutations @ immutable.js:1353
getActiveMarks @ slate.js:1891
object.(anonymous function) @ slate.js:1485
getActiveMarksBetweenOffsets @ slate.js:1845
getActiveMarksAtRange @ slate.js:16250
get$$1 @ slate.js:14444
renderNode @ Inline Babel script:124
editorPlugin.(anonymous function) @ slate-react.js:7827
find @ slate.js:12450
render @ slate-react.js:4614
finishClassComponent @ react-dom.development.js:11320
updateClassComponent @ react-dom.development.js:11297
beginWork @ react-dom.development.js:11676
performUnitOfWork @ react-dom.development.js:13644
workLoop @ react-dom.development.js:13753
callCallback @ react-dom.development.js:1527
[...]

PR #2226, opened 2018-10-04
fix activeMarks for intersecting marks text

zarv1k:

Fixes bug

Immutable OrderedSet is used instead of Set

Fixes: #2223


ianstormtaylor:
Thanks @zarv1k, this is a great catch!

PR #2227, opened 2018-10-04
serialize path as well in Operation.toJSON

bryanph:
fixing a bug

#### What's the new behavior?
serializes the path for operations that have one instead of returning the immutable instance

Fixes: #
Reviewers: @


ianstormtaylor:
Good call, thanks @bryanph!

PR #2229, opened 2018-10-05
Fix an occasional crash in Firefox in isInEditor

k7d:

Fix 

#### What's the new behavior?

Firefox occasional crashes in `Content.isInEditor` because target node is comment that doesn't have `.closest`.

Adds check in `Content.isInEditor` if the target node is comment, and returns false if it is.


packages/slate-react/src/components/content.js
```diff
@@ -270,6 +270,11 @@ class Content extends React.Component {
     let el
 
     try {
+      if (target.nodeType === 8) {
+        // on Firefox, sometimes the node can be comment which doesn't have .closest and it crashes
```
ianstormtaylor: Hey @k7d, can you copy the `COMPAT:` style comments from elsewhere in the codebase?

k7d: @ianstormtaylor sure thing. done

PR #2233, opened 2018-10-05
feat: add a basic mention implementation example

ericedem:
Just a simple example to get people started with their own mentions implementation.

_example_

#### What's the new behavior?

kapture 2018-10-16 at 17 13 39

Fixes: #
Reviewers: @


ianstormtaylor:
Hey @ericedem this is awesome! Thank you! Would you be willing to update it to use a floating portal like the hover menu does? I think it would make it more compelling for people that way.

ericedem:
@ianstormtaylor My intent was to create something as simple as possible since it is already a fairly complicated topic. Though thinking about it again, a lot of the complexity in our app comes from making it accessible (keyboard) and containing it within a plugin. Here it might be possible to do it in a simple way, I'll look into it 🤔

ericedem:
@ianstormtaylor Ok I updated the example to use a floating anchored menu, and updated to 0.42. Cheers 🍻

ianstormtaylor:
Thank you @ericedem this is awesome!

Issue #2231, opened 2018-10-05
Zero-width spaces cause unwanted wrapping / empty newlines before inline blocks


k7d:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Slate inserts span with zero-width spaces in front of inline blocks
starting at the begging of a line. If the inline block is longer than the line and doesn't have any whitespace in it, it causes a permanent newline before the block. Another sideeffect is that you cannot use key down to move down - the caret is just stuck there. This can be repeated with [included links example](https://www.slatejs.org/#/links). Just type some text without whitespace and add a link to it:

![image](https://user-images.githubusercontent.com/181645/46558900-dac35f80-c8ab-11e8-93ff-95119fff54ac.png)

(see the empty line abobe "longlong..")

This is a common scenario when implementing automatic link wrapping (ie. if I paste a link I want it to get automatically converted to a clickable link).

For now my workaround is to just do this:

```css
span[data-slate-zero-width='z'] {display: none;}
```

It fixes the issue and seem to not break anything (but I'm not 100% sure).

#### What's the expected behavior?

There should be no line breaks/empty lines before inline block in the above scenario.

k7d:
Actually display: none on zero-width spans does causes issues. @ianstormtaylor what if we replace U+200B with U+FEFF (ZERO WIDTH NO-BREAK SPACE)?

It seems to both fix the linebreak issue and doesn't breaking existing behaviour (for example adding text before inline block at a beginning of a line).

ianstormtaylor:
@k7d good call, that sounds like a good solution to me!

k7d:
okay cool i'll pr

PR #2234, opened 2018-10-06
Replace "zero width whitspace" with "zero width no break" character

k7d:

Bugfix

#### What's the new behavior? How does this change work?

Regular "zero width whitspace" (200B) caused odd line-breaks/empty lines when it got inserted before an inline block that exceeded width of container (fe. long text without whitespaces). "Zero width no break whitespace" FEFF is similar 200B but doesn't trigger a line break.
 
See https://github.com/ianstormtaylor/slate/issues/2231 for more context.

Fixes: https://github.com/ianstormtaylor/slate/issues/2231


ianstormtaylor:
Thank you @k7d this is great!

Issue #2095, opened 2018-08-18
When switching focus between editors in Firefox, the initial selection does not match the click position


birtles:
#### Do you want to request a _feature_ or report a _bug_?

Report a bug.

#### What's the current behavior?

1. Open https://jsfiddle.net/birtles/s3fp8qrj/ in Firefox (I am testing with Firefox 63 on Windows 10)
2. Click somewhere in the middle of the second line of text
3. Notice that the cursor jumps to the start of the line

If you click again in the same line, it will update correctly. The first click after changing editor focus does not respect the mouse position.

![focus](https://user-images.githubusercontent.com/1232595/44298515-5c245a00-a31f-11e8-99c1-e288c166b856.gif)

#### What's the expected behavior?

The cursor is positioned somewhere in the vicinity of the mouse click. Chrome 68 and Edge show the expected behavior.

Note that Draft JS previously had a very similar issue: https://github.com/facebook/draft-js/issues/1055. In that case IE11 also experienced the same issue. I haven't been able to try these steps in IE11 however since I can't even get jsfiddle to load there.

birtles:
So I only had a fairly brief look at this but I suspect the problem is something like the following.

1. When we change editors we call `updateSelection` in the `Content` component.
2. That triggers a call to [`removeAllRanges`](https://github.com/ianstormtaylor/slate/blob/e9c3c39d69f0388b2ffc496dc2e40fa86ea13ae8/packages/slate-react/src/components/content.js#L154)
3. In Firefox that updates the selection more-or-less immediately (compare the behavior of [this jsbin](http://jsbin.com/sahanirolu/edit?html,js,console,output) in Firefox and Chrome)
4. As a result when we get the (debounced) call to `onNativeSelection` → `this.props.onSelect` → `Editor.onEvent` → ... `onSelect` (after.js) → `findRange` which calls `window.getSelection()` the selection will already be cleared in Firefox but not in Chrome.

The discussion in https://github.com/facebook/draft-js/issues/1055#issuecomment-320107175 suggests that Chrome is probably wrong here.

Locally I can confirm that simply commenting out the call to `removeAllRanges` fixes the problem but I suspect it's there for a reason so I need to work out how this is expected to work.

pascalw:
@birtles did you ever get anywhere with this? I'm running into the same issue on FF 64 on macOS.

ianstormtaylor:
[...]

PR #2236, opened 2018-10-08
fix(firefox): fixed the bug that happens when changing the focus from one field to another

lakatosandrei:

fixing

This bug:
https://drive.google.com/file/d/1uG3kKuGRLO8d-XHAlDH4cz9dkgbuvPvM/view

There's a problem with the `updateSelection` function for Firefox. When changing focus from one editor to another the previous element is focused again.

#### What's the new behavior?

When trying to change the focus from one slate editor to another slate editor the cursor is placed in the new editor where it should.

This is the new behaviour:
http://www.giphy.com/gifs/LVuGFQ0qlmKb7HBcLg

I removed the code that was not needed anymore and added a check for some code that was messing up things on Firefox. Previously some of the code was designed specifically for Firefox.

Don't think so


CameronAckermanSEL:
@lakatosandrei Hiya, would you mind clearing out this last eslint warning? I think that's the only barrier to getting this looked at.

Thanks :)

ererik:
At first glance, this seems to solve #2631 and #2095 .

CameronAckermanSEL:
This looks good to me, can we revert the updates to the package.json files?

lakatosandrei:
> This looks good to me, can we revert the updates to the package.json files?

Done

lakatosandrei:
Fixed the conflicts as well.

darkyshiny:
@ianstormtaylor Could you take a look at this so that it can hopefully be merged soon?

This fixes #2095 which has been around for a long time and is a pretty bad experience for firefox users

kaydelaney:
Us at Grafana would also love to see this in the next release - Slate is becoming an increasingly important part of Grafana, and I'm sure we'd be willing to share some maintenance responsibilities to see PRs like this one move along more smoothly.

@ianstormtaylor Any thoughts on having this PR merged sometime soon? (After conflicts are resolved, of course). Thanks!

ianstormtaylor:
Thanks @lakatosandrei!

lakatosandrei:
> Thanks @lakatosandrei!

My pleasure!

themithy:
Hi,

I looks like this commit causes a regression on firefox, you can see it on one of the examples on the home page.

1. Pressing arrow left or right does not move the selection.
2. Typing a character makes the selection to jump somewhere else in the editor.

This makes the experience pretty difficult.

Firefox 68.

Cheers.

clauderic:
I can confirm the issue @themithy is seeing, both with Firefox 66 and 68 (latest). The left and right arrows do not work to move the cursor. 

Typing initially inserts the first character in the right position, but then the cursor jumps to the nearest mark or the beginning of the current block if there are no marks in the current block.

This PR should probably be reverted, and the issue it aimed to solve should be fixed in a subsequent PR seeing as it causes a much larger issue which makes Slate unusable on Firefox

kaydelaney:
As for the issue itself, I spent some time looking into it today and it seems like the root cause is that on Firefox at least, `componentDidUpdate`, and consequently `updateSelection` is called before `onNativeSelectionChange`, but _after_ the native selection is actually updated, causing all ranges to be removed from the native selection.

Issue #2201, opened 2018-09-24
Delete->Undo will move cursor to the beginning of editor


zhujinxuan:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

See the gif below: ![](http://g.recordit.co/bHDsEj0zLV.gif)

Delete a char and then undo will place the cursor to a wrong position.

#### What's the expected behavior?

Cursor shall be where the edit happens.

zhujinxuan:
<img width="488" alt="screen shot 2018-09-24 at 12 45 16 pm" src="https://user-images.githubusercontent.com/746159/45965918-c0f66280-bff7-11e8-945a-fd21bff92136.png">
It seems update selection triggers an strange onSelect whose cursor position is not right... I think it is related to https://github.com/ianstormtaylor/slate/issues/1969

zhujinxuan:
<img width="488" alt="screen shot 2018-09-24 at 12 45 16 pm" src="https://user-images.githubusercontent.com/746159/45966770-10d62900-bffa-11e8-8208-905fddcc3cf8.png">

tmp.isUpdatingSelection is true, but stack.run('onSelect') is still activated...

zhujinxuan:
It seems onSelect propogates upward from Container to Content

ericedem:
May be related to this? https://github.com/ianstormtaylor/slate/issues/2098

zhujinxuan:
@ericedem I am unsure.  There are three types of problems about undo-redo:
1. normalize or browser native selection adjust after undo;  for example, cursor moved to the end of inline after undo, then browser may move the cursor to the beginning of the next span.
2. seleccion update like this question.
3. when cursor is inside the block, undo->redo will move cursor out of the block.

zhujinxuan:
In this case, I guess the problem is triggered by
1. select events propagate from Container to Content. 
2. onNativeSelection is not prevented

zhujinxuan:
@ericedem You are right.  The problem is about undo the selection...  I just found some strange behavior in logs.


<img width="487" alt="screen shot 2018-09-25 at 2 06 21 pm" src="https://user-images.githubusercontent.com/746159/46033511-5e22cb00-c0cc-11e8-9ead-a3b47f00f1d5.png">


It seems the the history does not batch operations in the right way.  The slate editor undos all operations until the start state of the document.

zhujinxuan:
Perhaps we can set history stack barrier in these events:
[...]

PR #2240, opened 2018-10-09
Fix delete undo

inkubux:

by using the options sent to the applyOperation method 

`change.applyOperation(
    {
      type: 'set_selection',
      value,
      properties: props,
      selection: selection.toJSON(),
    },
    snapshot ? **{ skip: false, merge: false }** : {}
  )`

`applyOperation(operation, options = {}) {
    const { operations } = this
    let { value } = this
    let { history } = value

    // Add in the current `value` in case the operation was serialized.
    if (isPlainObject(operation)) {
      operation = { ...operation, value }
    }

    operation = Operation.create(operation)

    // Default options to the change-level flags, this allows for setting
    // specific options for all of the operations of a given change.
    let { merge, save } = **{ ...this.tmp, ...options }**
`

* [ x] The new code matches the existing patterns and styles.
* [ x] The linter passes with `yarn lint`. (Fix errors with `yarn prettier`.)
* [ x] The relevant examples still work. (Run examples with `yarn watch`.)

Fixes: https://github.com/ianstormtaylor/slate/issues/2201
Reviewers: @zhujinxuan

### DIFF
--- a/packages/slate/src/changes/on-selection.js
+++ b/packages/slate/src/changes/on-selection.js
@@ -607,7 +607,7 @@ Changes.setStart = (change, ...args) => {
 
 Changes.snapshotSelection = change => {
   change.withoutMerging(c =>
-    change.select(change.value.selection, { snapshot: true })
+    c.select(change.value.selection, { snapshot: true })
   )
 }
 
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Beautiful PR @inkubux, thanks for updating it to use the `withoutMerging` helper! And thanks for the test case as well.

PR #2247, opened 2018-10-10
Take Blocks as word break to fix word counting in example

zhujinxuan:

FIx two Bugs in example.
1. Block break is not considered as word break in the old example
2. Continuous spaces are considered as words

Fixes: #
Reviewers: @


ianstormtaylor:
Thanks @zhujinxuan!

Issue #2042, opened 2018-08-06
consider adding range comparison methods to `Point`


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now if you want to compare a point to a range, you need to write the logic yourself by comparing the `path` and `offset` properties. But it might be nice to provide a few conveniences here, like:

- [ ] `Point.isAfterRange`
- [ ] `Point.isAtStartOfRange`
- [ ] `Point.isAtEndOfRange`
- [ ] `Point.isBeforeRange`
- [ ] `Point.isInRange`

PR #2248, opened 2018-10-10
Add range comparison methods to `Point`

Dundercover:

Improving.

#### What's the new behavior?

Adds range comparison methods to `Point` as suggested in #2042 

* Renamed the `compare` method in `path-utils.js` to `compareOfSameSize` and then added a new `compare` method that does a full comparison. The new `Point` methods uses the new `compare` method.
* Did not add any tests because I couldn't find any for neither `path-utils` nor `Point` (I could fix this if you give some pointers to how I should go about).

Fixes: #2042

### DIFF
--- a/packages/slate-react/src/components/text.js
+++ b/packages/slate-react/src/components/text.js
@@ -119,10 +119,10 @@ class Text extends React.Component {
 
       // If the node's path is before the start path, ignore it.
       const path = document.assertPath(key)
-      if (PathUtils.compareOfSameSize(path, start.path) === -1) return false
+      if (PathUtils.compare(path, start.path) === -1) return false
 
       // If the node's path is after the end path, ignore it.
-      if (PathUtils.compareOfSameSize(path, end.path) === 1) return false
+      if (PathUtils.compare(path, end.path) === 1) return false
 
       // Otherwise, include it.
       return true
--- a/packages/slate-react/src/components/text.js
+++ b/packages/slate-react/src/components/text.js
@@ -119,10 +119,10 @@ class Text extends React.Component {
 
       // If the node's path is before the start path, ignore it.
       const path = document.assertPath(key)
-      if (PathUtils.compare(path, start.path) === -1) return false
+      if (PathUtils.compareOfSameSize(path, start.path) === -1) return false
 
       // If the node's path is after the end path, ignore it.
-      if (PathUtils.compare(path, end.path) === 1) return false
+      if (PathUtils.compareOfSameSize(path, end.path) === 1) return false
 
       // Otherwise, include it.
       return true
--- a/packages/slate/src/interfaces/element.js
+++ b/packages/slate/src/interfaces/element.js
@@ -1547,7 +1547,7 @@ class ElementInterface {
     this.assertNode(newParentPath)
 
     const [p, np] = PathUtils.crop(path, newPath)
-    const position = PathUtils.compareOfSameSize(p, np)
+    const position = PathUtils.compare(p, np)
 
     // If the old path ends above and before a node in the new path, then
     // removing it will alter the target, so we need to adjust the new path.
--- a/packages/slate/src/interfaces/element.js
+++ b/packages/slate/src/interfaces/element.js
@@ -1547,7 +1547,7 @@ class ElementInterface {
     this.assertNode(newParentPath)
 
     const [p, np] = PathUtils.crop(path, newPath)
-    const position = PathUtils.compare(p, np)
+    const position = PathUtils.compareOfSameSize(p, np)
 
     // If the old path ends above and before a node in the new path, then
     // removing it will alter the target, so we need to adjust the new path.
--- a/packages/slate/src/models/point.js
+++ b/packages/slate/src/models/point.js
@@ -135,6 +135,19 @@ class Point extends Record(DEFAULTS) {
   }
 
   /**
+   * Check whether the point is at the start of a `range`.
+   *
+   * @return {Boolean}
+   */
+
+  isAtStartOfRange(range) {
+    if (this.isUnset) return false
+    const is =
+      this.key === range.start.key && this.offset === range.start.offset
+    return is
+  }
+
+  /**
    * Check whether the point is before a `range`.
    *
    * @return {Boolean}
--- a/packages/slate/src/models/point.js
+++ b/packages/slate/src/models/point.js
@@ -135,6 +135,20 @@ class Point extends Record(DEFAULTS) {
   }
 
   /**
+   * Check whether the point is before a `range`.
+   *
+   * @return {Boolean}
+   */
+
+  isBeforeRange(range) {
+    if (this.isUnset) return false
+    const is =
+      (this.key === range.start.key && this.offset < range.start.offset) ||
+      PathUtils.compare(this.path, range.start.path) === -1
+    return is
+  }
+
+  /**
    * Check whether the point is at the end of a `node`.
    *
    * @param {Node} node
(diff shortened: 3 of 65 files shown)


packages/slate/src/models/point.js
```diff
@@ -120,6 +120,79 @@ class Point extends Record(DEFAULTS) {
     return !this.isSet
   }
 
+  /**
+   * Check whether the point is after a `range`.
+   *
+   * @return {Boolean}
+   */
+
+  isAfterRange(range) {
```
ianstormtaylor: Could you refactor this to add `isAfterPoint(point)` which can then be used inside `isAfterRange(range)`? That way we get both options.

packages/slate/src/models/point.js
```diff
@@ -120,6 +120,79 @@ class Point extends Record(DEFAULTS) {
...
+  /**
+   * Check whether the point is inside a `range`.
+   *
+   * @return {Boolean}
+   */
+
+  isInRange(range) {
+    if (this.isUnset) return false
+
+    const afterOrAtStart =
```
ianstormtaylor: Could you refactor these to use the other methods defined in this pull request to make it a bit easier to follow and reduce code?

packages/slate/src/models/point.js
```diff
@@ -120,6 +120,79 @@ class Point extends Record(DEFAULTS) {
...
+    return is
+  }
+
+  /**
+   * Check whether the point is before a `range`.
+   *
+   * @return {Boolean}
+   */
+
+  isBeforeRange(range) {
```
ianstormtaylor: Could to do the same for `isBeforePoint(point)` too?

packages/slate/src/interfaces/element.js
```diff
@@ -1547,7 +1547,7 @@ class ElementInterface {
     this.assertNode(newParentPath)
 
     const [p, np] = PathUtils.crop(path, newPath)
-    const position = PathUtils.compare(p, np)
+    const position = PathUtils.compareOfSameSize(p, np)
```
ianstormtaylor: I think since we already used `crop` that this can also use the new `compare` instead?

packages/slate/src/utils/path-utils.js
```diff
@@ -102,7 +131,7 @@ function increment(path, n = 1, index = path.size - 1) {
 
 function isAbove(path, target) {
   const [p, t] = crop(path, target)
-  return path.size < target.size && compare(p, t) === 0
+  return path.size < target.size && compareOfSameSize(p, t) === 0
```
ianstormtaylor: I think we can use the new compare here?

packages/slate/src/utils/path-utils.js
```diff
@@ -115,7 +144,7 @@ function isAbove(path, target) {
 
 function isAfter(path, target) {
   const [p, t] = crop(path, target)
-  return compare(p, t) === 1
+  return compareOfSameSize(p, t) === 1
```
ianstormtaylor: And here too?

packages/slate/src/utils/path-utils.js
```diff
@@ -128,7 +157,7 @@ function isAfter(path, target) {
 
 function isBefore(path, target) {
   const [p, t] = crop(path, target)
-  return compare(p, t) === -1
+  return compareOfSameSize(p, t) === -1
```
ianstormtaylor: And here too?

packages/slate/src/models/point.js
Dundercover: Sounds good, I'll fix it.

packages/slate/src/models/point.js
Dundercover: Of course, I wasn't sure if I could use the other methods because it would lead to potentially more calls to `compare` but it would be easier to understand and maintain. I'll fix it.

packages/slate/src/models/point.js
Dundercover: :+1:

packages/slate/src/interfaces/element.js
Dundercover: It can! I'll fix it.

packages/slate/src/utils/path-utils.js
Dundercover: :+1:

packages/slate/src/utils/path-utils.js
Dundercover: :+1:

packages/slate/src/utils/path-utils.js
Dundercover: :+1:

packages/slate-react/src/components/text.js
```diff
@@ -119,10 +119,10 @@ class Text extends React.Component {
 
       // If the node's path is before the start path, ignore it.
       const path = document.assertPath(key)
-      if (PathUtils.compare(path, start.path) === -1) return false
+      if (PathUtils.compareOfSameSize(path, start.path) === -1) return false
```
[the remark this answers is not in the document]
ianstormtaylor: Ah yeah, I'm not worried about that `PERF` callout. I'd say lets get rid of it completely if we can!

packages/slate/src/utils/path-utils.js
```diff
@@ -1,14 +1,43 @@
...
+    const tv = target.get(i)
+
+    // If the path's value is ever less than the target's, it's before.
+    if (pv < tv) return -1
+
+    // If the target's value is ever less than the path's, it's after.
+    if (pv > tv) return 1
+  }
+
+  // Otherwise size decides.
```
ianstormtaylor: Why does the length decide here? Is there a specific reason that we need that to end up that way?

What does it mean for `[0]` to be "before" `[0, 1]`? I guess it kind of makes sense in the top-down tree thinking. But I'm wondering if it would be better to either return `null` or `0`? Just curious.

packages/slate-react/src/components/text.js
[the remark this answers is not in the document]
Dundercover: I'll remove it then.

packages/slate/src/utils/path-utils.js
Dundercover: Good call, I actually don't know why I decided that a node should "come before" a descendant. Personally I would probably prefer to return `null` so that it's more apparent that it is not a normal scenario and when I use the method I can handle it separately if needed. What do you think?

Dundercover:
I have updated the pull request with the changes made in response to the feedback. Let me know if there is anything else!

zhujinxuan:
Hi, would you like to add some tests in `slate/packages/slate/test/models` to fix codecov?

packages/slate/src/utils/path-utils.js
```diff
@@ -1,18 +1,17 @@
...
  * @param {List} path
- * @param {List} b
+ * @param {List} target
  * @return {Number|Null}
  */
 
 function compare(path, target) {
-  // PERF: if the paths are not the same size we can exit early.
-  if (path.size !== target.size) return null
+  const m = min(path, target)
```
zhujinxuan: A small suggestion.  I am feeling that perhaps it is a bit better to avoid `for`

```
let result = null;
path.zip(target).find( ([pv, tv] => {
   if (pv === tv) return false;
   result = pv > tv ? 1 : 0;
   return true
})

if (result !== null) return result;
return path.size === target.size ? 0 : null
```

Dundercover:
@zhujinxuan I'll give it a try

zhujinxuan:
@Dundercover It looks good to me~

ianstormtaylor:
@Dundercover this is an absolutely stellar pull request. Thank you for taking the time to submit it!

Dundercover:
I'm happy to help @ianstormtaylor

Issue #2209, opened 2018-09-26
Inserting text at a range instead inserts at the current selection in some situations


alanctkc:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

When applying `instertTextAtRange`, if the editor is focused on a different line (text node) than the node in the range passed to `insertTextAtRange`, the text will be inserted at the current focus (selection) instead of the range specified.

![peek 2018-09-26 07-22](https://user-images.githubusercontent.com/273453/46079578-20ba4e00-c15d-11e8-802e-910d71fc81c1.gif)

[Here](https://jsfiddle.net/fj9dvhom/2350/) is the fiddle for the above.

#### What's the expected behavior?

`insertTextAtRange` should always insert text at the specified range, regardless of the current editor selection.

[Here's a test case](https://github.com/alanctkc/slate/blob/ff7c422c0e3d2693c0cf1dbd5092d313c4d8aefd/packages/slate/test/changes/at-range/insert-text-at-range/non-matching-section-and-range.js) that fails on the bug:

```javascript
export default function(change) {
  const { key } = change.value.document.getFirstText()
  const range = new Range({
    anchor: new Point({ key, offset: 0 }),
    focus: new Point({ key, offset: 3 }),
  })
  change.insertTextAtRange(range, 'That')
}

export const input = (
  <value>
    <document>
      <line>The change will be here.</line>
      <line>
        The cursor is <cursor />over here.
      </line>
    </document>
  </value>
)

export const output = (
  <value>
    <document>
      <line>That change will be here.</line>
      <line>
        The cursor is <cursor />over here.
      </line>
    </document>
  </value>
)
```

#### The cause

These are the offending lines:

https://github.com/ianstormtaylor/slate/blob/6813ac41b5eba4211e87fdb998752c937279960f/packages/slate/src/changes/at-range.js#L809-L813

After the range is deleted, the start key is reset to the start of the current selection. This is presumably to handle key changes and what not after deleting multiple lines, but it's wrong to assume the current selection will reflect the original range. Maybe there's another way to achieve the same behavior after delete without using current selection?

ericedem:
Weird, I'm not sure why there is selection logic in there at all. It seems like we could just delete that block and have:

```
change.withoutNormalizing(() => {
  if (range.isExpanded) {
    change.deleteAtRange(range)
  }

  change.insertTextByKey(key, offset, text, marks)
})
```

alanctkc:
[...]

PR #2256, opened 2018-10-12
Fix insertTextAtRange selection

alanctkc:

Fixing a bug.

#### What's the new behavior?

The issue is describe in detail in #2209 and demonstrated by the test included here. But, basically, this makes `insertTextAtRange` insert text at the range specified instead of the current selection when the selection is elsewhere. Here was the original issue in action:

ouch

Instead of referencing the current selection to determine the key to use for text insertion, `insertTextAtRange` now preserves the path of the original block that may be deleted during the replacement of the range. If the block is deleted and keys change, the path should remain consistent. So, this change gets the key from the new block at that path, regardless of where the current selection is. (Previously, it tried to get the range start block from the current selection, which isn't always the same).

Fixes: #2209 
Reviewers: @ianstormtaylor, @ericedem, @czechdave, @echenley, @blakeembrey

### DIFF
--- a/packages/slate/src/commands/at-range.js
+++ b/packages/slate/src/commands/at-range.js
@@ -794,7 +794,8 @@ Commands.insertTextAtRange = (change, range, text, marks) => {
   const { document } = value
   const { start } = range
   let key = start.key
-  let offset = start.offset
+  const offset = start.offset
+  const path = start.path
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thank you for this @alanctkc!

Issue #1754, opened 2018-04-04
insertFragmentAtRange results in invalid selection


Soreine:
#### Do you want to request a _feature_ or report a _bug_?

A bug

#### What's the current behavior?

As exposed by this PR #1379 there was a test that was passing, but was revealed to be ill-formed. The test is the following, and it yields an invalid selection:

```jsx
// test/changes/at-current-range/insert-fragment/start-inline.js

export default function(change) {
  change.insertFragment(
    <document>
      <quote>fragment</quote>
    </document>
  )
}

export const input = (
  <value>
    <document>
      <paragraph>
        <link>
          <cursor />word
        </link>
      </paragraph>
    </document>
  </value>
)

export const output = (
  <value>
    <document>
      <paragraph>
        fragment<cursor />
        <link>word</link>
      </paragraph>
    </document>
  </value>
)

// The result has an invalid selection for now:
//
// "selection": {
//   "anchorOffset": 8
//   "anchorPath": [
//     0
//     1
//     0
//   ]
//   "focusOffset": 8
//   "focusPath": [
//     0
//     1
//     0
//   ]
```

#### What's the expected behavior?

The selection object should point to a valid place in the document:

```jsx
// "selection": {
//   "anchorOffset": 8
//   "anchorPath": [
//     0
//     0
//   ]
//   "focusOffset": 8
//   "focusPath": [
//     0
//     0
//   ]
```

zhujinxuan:
I think the problem is caused by the cursor restore in insertFragment, not in insertFragmentAtRange

Issue #1494, opened 2017-12-28
bug: if all the text under an inline node selected, and new text is typed, the new text does not keep the format 


amitm02:
In slatejs playground, [http://slatejs.org/#/links](http://slatejs.org/#/links):
If you select all the text of the link ("hyperlinks"), and type the keyboard, the first letter is in right "link" format, but the following letters are in not.

![qkxcux2lhd](https://user-images.githubusercontent.com/1888927/34417371-cd081dd2-ec00-11e7-84e4-16a24101ca8f.gif)
[gif](http://g.recordit.co/qkxcUx2LHD.gif)

compared to underline & bold:
![gnfp53cftr](https://user-images.githubusercontent.com/1888927/34417403-f6d6b916-ec00-11e7-8362-a1e8d7e80357.gif)
[gif](http://g.recordit.co/GnFP53cfTr.gif)

compared to a regular editor (e.g Word
![yw7lbocfl8](https://user-images.githubusercontent.com/1888927/34417386-e56213ce-ec00-11e7-8abc-73697013736c.gif)
[gif](http://g.recordit.co/Yw7lbOCFL8.gif)

ianstormtaylor:
Interestingly, both Medium and Dropbox Paper have the same failure as Slate currently. I'd be open to a pull request that fixes this by going the Google Docs route, where none of the text has a link when it is replaced. (Going the Microsoft Word route that maintains the link is going to be too complex.)

amitm02:
I have a use case in which it is important to be able to replace a text under an inline node in the manner described above. Any suggestions on how to go about it?

ianstormtaylor:
What's your use case? I'm not really sure. It would require investigation to figure out why the change happened, and whether it's possible to have the old behavior back in a cross-browser way.

amitm02:
I want to build a custom “field” node with a fieldID attribute (E.g “personName” ). I want  it to have the following behavior. 
1. When the cursor is inside the field node, select all the text in the node. 
2. When typing the keyboard text, the pervious text in the field is replaced with what is typed. 
3 I can access the field content via the fieldID. E.g get the “personName” field content from the document. 

I’m not suf

ianstormtaylor:
Is there a reason the `field` node needs to be inline? could it be a block instead?

amitm02:
[...]

PR #2271, opened 2018-10-16
Alter cursor placement in `insertFragment` tests to reflect reality

steobrien:
Previously, `start-inline` and `end-inline` included the cursor at the end of an inline, i.e.
```js
<paragraph>
  <link>example<cursor /></link>
</paragraph>
```

Using Slate in a browser, it doesn’t seem possible to me to actually achieve this state: i.e. whenever the cursor is on either side of an inline such as a link, typing will not add text to that inline. (This is consistent with other editors, for example Google Docs).

Now, these tests are updated to have a cursor just outside the inline. This leads to more sensible outputs.

(Finally, I removed a TODO comment indicating that `middle-inline-fragment-inline`’s cursor output was incorrect. I believe this is a sensible expectation, which matches behavior in editors including Google Docs.)

Fixes https://github.com/ianstormtaylor/slate/issues/1754

### DIFF
--- a/packages/slate/test/commands/at-current-range/insert-fragment/end-inline.js
+++ b/packages/slate/test/commands/at-current-range/insert-fragment/end-inline.js
@@ -14,9 +14,8 @@ export const input = (
   <value>
     <document>
       <paragraph>
-        <link>
-          word<cursor />
-        </link>
+        <link>word</link>
+        <cursor />
       </paragraph>
     </document>
   </value>
@@ -27,10 +26,7 @@ export const output = (
     <document>
       <paragraph>
         <link>word</link>
-        fragment
-        <link>
-          <cursor />
-        </link>
+        fragment<cursor />
       </paragraph>
     </document>
   </value>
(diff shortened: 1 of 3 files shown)


zhujinxuan:
I agree that it does not happen likely in browser by default.   However, it is achieved in model when using some plugins.  For example, if we 
   1. use sticky-inline plugin
   2. use a plugin overwrite the `onSelect` function that do not update selection if offset different is 0 (a walkaround to fix https://github.com/ianstormtaylor/slate/issues/1494)  
then the cursor can be placed in the inline end.

steobrien:
Thanks for that context @zhujinxuan.

It seems worthwhile to have tests here for the more core/common use case, but perhaps we should additionally have tests for cursors at inline edges. It’s worth pointing out that the current behavior in those cases is not great.

zhujinxuan:
@steobrien Yes, I agree.  I think these edge cases shall be tested separately.

ianstormtaylor:
Thanks @steobrien!

I agree with @zhujinxuan that we should have tests for both, but this can be handled in another PR to make those edge cases behave better.

Issue #2064, opened 2018-08-09
Slate data fragment merging leads to extra line breaks


Slapbox:
In particular, with lists, but no doubt this affects other types too. This makes envisioning a full solution difficult as I can't imagine all the different scenarios which a solution needs to cover.

There's not, as far as I can tell, 100% standard behavior across all editors on this, but we can definitely improve Slate's behavior to be more predictable for users anyway.

**Here's how Google Docs handles an example for lists:**
http://recordit.co/SEzxiNiodn

**Here's how Slate handles it:**
http://recordit.co/Gr9dTOaKHi

The fragments Slate decodes look like this:

![image](https://user-images.githubusercontent.com/6835891/43913177-5515c2ac-9bd2-11e8-8679-a03cc23cb585.png)

### Relevant files
**The line where a fragment is encoded into the HTML on copy:** https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/clone-fragment.js#L33

**First line where an encoded fragment is decoded (if it exists) on paste:** https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/utils/get-event-transfer.js#L64 << This is probably the file we need to modify

It occurs to me we can check the types of the enclosing block for the fragment and for the paste location and avoid duplicating the block type, in this case `numbered-list`, if they match, but I'm not sure if this covers all use cases.

Can anyone think of a possible case this wouldn't work in? For me, tables come to mind, but if you try to paste table rows inside a table, you're gonna have a bad time. This seems beyond what Slate itself needs to handle, and an acceptable solution in my view should simply not interfere with someone else who wants to extend the behavior to handle tables.

cybojenix:
This has become a pretty big problem for us.
It seems that when copying, all the blocks leading up to the block you've copied are cloned.

As an example, we use lists that can be nested. When copying text from a nested list, the full structure is copied with it.

```yaml
- ul_list:
  - list_item:
    - paragraph: "some text"
    - ul_list:
      - list_item:
        - paragraph: "This is to be copied"
```
The expectation would be either ```"This is to be copied"``` would get copied as text, or the single `list_item` gets copied.
Instead what we get is this in the copied fragment:
```yaml
- ul_list:
  - list_item:
    - ul_list:
      - list_item:
        - paragraph: "This is to be copied"
```

Currently we are stripping much of the structure away to give us `list_items`, however this method is very fragile, and doesn't take into account copying of other types.
[...]

PR #2273, opened 2018-10-17
[proposal] Meld fragment into document in `Commands.insertFragmentAtRange`

steobrien:

Fixing a bug. This is a proposed solution for https://github.com/ianstormtaylor/slate/issues/2064 (which also aims to maintain the existing _single nested block_ behavior added in https://github.com/ianstormtaylor/slate/pull/1366).

#### What's the new behavior?

Today, we always insert all nodes of the fragment at the range.

After this change, if the fragment is wrapped by/nested in blocks, and the nesting pattern already exists at the range, we meld the fragment contents into the document’s existing nesting pattern as much as possible.

Put another way, we can skip adding the fragment’s wrapping nodes if equivalent nodes already wrap the range.

Here’s a quick sketch that may better demonstrate the intention:
image

Here, both the document and fragment wrap their content in two paragraphs. Therefore, we can add the fragment contents within the existing pair of nested paragraphs.

A new function is added, `findInsertionNode`. Given a `fragment`, `document`, and `startKey`, it returns a node from the fragment to be inserted into the document. By default, the node returned will be the `fragment`, but if a common nesting pattern is detected, a sub-node of the fragment will be returned.

The detection algorithm is as follows:
1. Ensure the fragment has a single child node, call it `fragmentInner`
2. Find the furthest document ancestor of the `startKey` matching `fragmentInner.type`. Call this `documentInner`
3. Drill into both `documentInner` and `fragmentInner` as long as each has a single child node and their types match. The stopping point for the fragment is the sub-node to insert into the document.

`Commands.insertFragmentAtRange` remains mostly unchanged. One important modification is that the _single nested block_ (again, see https://github.com/ianstormtaylor/slate/pull/1366) only happens if the found insertion node is the fragment. For anything more inner, we meld.

Fixes: https://github.com/ianstormtaylor/slate/issues/2064

### DIFF
--- a/packages/slate/test/commands/at-current-range/insert-fragment/merge-deep-nested.js
+++ b/packages/slate/test/commands/at-current-range/insert-fragment/merge-deep-nested.js
@@ -1,0 +1,49 @@
+/** @jsx h */
+
+import h from '../../../helpers/h'
+
+export default function(change) {
+  change.insertFragment(
+    <document>
+      <code>
+        <quote>
+          <paragraph>2</paragraph>
+          <paragraph>3</paragraph>
+        </quote>
+      </code>
+    </document>
+  )
+}
+
+export const input = (
+  <value>
+    <document>
+      <paragraph>
+        <code>
+          <quote>
+            <paragraph>
+              1<cursor />
+            </paragraph>
+          </quote>
+        </code>
+      </paragraph>
+    </document>
+  </value>
+)
+
+export const output = (
+  <value>
+    <document>
+      <paragraph>
+        <code>
+          <quote>
+            <paragraph>12</paragraph>
+            <paragraph>
+              3<cursor />
+            </paragraph>
+          </quote>
+        </code>
+      </paragraph>
+    </document>
+  </value>
+)
--- a/packages/slate/test/commands/at-current-range/insert-fragment/merge-lists.js
+++ b/packages/slate/test/commands/at-current-range/insert-fragment/merge-lists.js
@@ -5,10 +5,10 @@ import h from '../../../helpers/h'
 export default function(change) {
   change.insertFragment(
     <document>
-      <quote>
-        <quote>one</quote>
-        <quote>two</quote>
-      </quote>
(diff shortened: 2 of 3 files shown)


packages/slate/test/commands/at-current-range/insert-fragment/merge-lists.js
```diff
@@ -0,0 +1,41 @@
...
+      <list>
+        <item>1</item>
+        <item>23</item>
+        <item>
+          4<cursor />
+        </item>
+      </list>
+    </document>
+  </value>
+)
```
steobrien: on master, `output` would be:
```js
<value>
  <document>
    <list>
      <item>1</item>
      <item>2</item>
      <list>
        <item>3</item>
        <item>
          4<cursor />
        </item>
      </list>
    </list>
  </document>
</value>
```

packages/slate/test/commands/at-current-range/insert-fragment/merge-deep-nested.js
```diff
@@ -0,0 +1,49 @@
...
+            <paragraph>12</paragraph>
+            <paragraph>
+              3<cursor />
+            </paragraph>
+          </quote>
+        </code>
+      </paragraph>
+    </document>
+  </value>
+)
```
steobrien: on master, `output` would be:
```js
<value>
  <document>
    <paragraph>
      <code>
        <quote>
          <paragraph>1</paragraph>
          <code>
            <quote>
              <paragraph>2</paragraph>
              <paragraph>
                3<cursor />
              </paragraph>
            </quote>
          </code>
        </quote>
      </code>
    </paragraph>
  </document>
</value>
```

steobrien:
Hey @ianstormtaylor, I’d really appreciate some guidance on how to proceed here. Would you be able to review this PR some time soon?

ianstormtaylor:
Hey @steobrien, I think I like this change. Thanks for taking the time to write it up really well so that it was easier to understand!

PR #2277, opened 2018-10-17
Fix bolded text in input-tester initial value

SmilinBrian:
Initial value in the input-tester example renders the bolded items as "bold".

This fixes the initial value so it reads correctly.

Current initial value:
screenshot

Fixed:
screenshot


ianstormtaylor:
😄 thank you @SmilinBrian!

PR #2278, opened 2018-10-17
Fix exception in Input Tester example

SmilinBrian:
If an Input event with inputType `"insertReplacementText"` is received, the example crashes because `event.dataTransfer.get is not a function`

The correct API is `getData()`, not `get()`. Ref: https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/getData

You can cause the exception by mis-spelling a word in the Input Tester with Safari, then using control-click (or right-click on a 2-button mouse) to correct the spelling.

### DIFF
--- a/examples/input-tester/index.js
+++ b/examples/input-tester/index.js
@@ -176,7 +176,7 @@ const Event = ({ event, targetRange, selection }) => {
       </td>
       <td>
         <StringCell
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thanks @SmilinBrian!

PR #2283, opened 2018-10-19
add favicon to the examples

ericedem:
Was tired of seeing the error so I stoled this from the gitbook documentation.

I figured this was better than nothing, but might be worth having a slightly different color scheme or something to differentiate browser / examples tabs.


ianstormtaylor:
Beautiful, thanks @ericedem!

Issue #2279, opened 2018-10-18
Select all + Cut crashes the editor


linvain:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

1. `Ctrl + A`
2. `Ctrl + X`
3. Press any button
4. Error

![gif](https://user-images.githubusercontent.com/42490556/47141160-b8c4d680-d2c8-11e8-9968-20213049c351.gif)

javascriptlove:
Just noticed that too, very annoying

ericedem:
It looks like all the blocks are getting deleted in addition to all the text. Something is goin haywire with the delete in https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/after.js#L216

ericedem:
It looks like there are two problems happening:

1. When you cut, the first cut happens and deletes the content. Then for some reason another cut happens after the first one when there is no content left.
2. Calling delete (cut in this case) with no content deletes the paragraph (oops!).

This test illustrates 2: https://gist.github.com/ericedem/9e51a181689769ae9510fb02428fc1d8

ericedem:
@linvain due to the 0.43 release that just went out. 0.42 should be working.

skogsmaskin:
Fixed again in https://github.com/ianstormtaylor/slate/pull/2343

PR #2298, opened 2018-10-22
Fix cloneFragment conditional handling of clipboardData.setData

kalley:

Fixing a bug

#### What's the new behavior?

Fixes the `cloneFragment` function to the intended purpose

According to the comments in the file and the diff, the intention was for `requestAnimationFrame` to only be used when the Clipboard API's `setData` method was either not available or lackluster (IE). 

The issue was introduced accidentally here (which is how I figured out what needed to be done): https://github.com/ianstormtaylor/slate/commit/dc95ad66a522e9d5c1626e65844f160834224696#diff-dfb9fbad6106fa548b8335a644d03e49L109

This was especially problematic in the `onCut` command, since the callback was being called twice in browsers with the necessary support.

Fixes: #2279
Reviewers: @zhujinxuan (since you made the change above)

### DIFF
--- a/packages/slate-react/src/utils/clone-fragment.js
+++ b/packages/slate-react/src/utils/clone-fragment.js
@@ -114,6 +114,7 @@ function cloneFragment(event, editor, callback = () => undefined) {
     event.clipboardData.setData(FRAGMENT, encoded)
     event.clipboardData.setData(HTML, div.innerHTML)
     callback()
+    return
   }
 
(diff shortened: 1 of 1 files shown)


zhujinxuan:
My fault.  I shall add `return` there.

Thank you very much.

ianstormtaylor:
Thanks @kalley, good catch!

Issue #2275, opened 2018-10-17
Empty blocks display no selection state


steobrien:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

When selecting, for empty blocks, no selection state is visible:
![image](https://user-images.githubusercontent.com/1694410/47055547-e5240a00-d185-11e8-8974-7a58ba04de6e.png)

#### What's the expected behavior?

The empty blocks should have a minimal selection state. This improves usability when making selections with keyboard or mouse.

Some examples from other contexts:

- Plain textarea:
![image](https://user-images.githubusercontent.com/1694410/47055580-0be24080-d186-11e8-8e87-88de50fc6d2f.png)
- Draft.js:
![image](https://user-images.githubusercontent.com/1694410/47055594-2e745980-d186-11e8-8ebf-0e892595092c.png)
- Quill:
![image](https://user-images.githubusercontent.com/1694410/47055630-611e5200-d186-11e8-8b05-aaabdee12939.png)
- Trix:
![image](https://user-images.githubusercontent.com/1694410/47055640-709d9b00-d186-11e8-810b-28f80c4b5943.png)
- ProseMirror:
![image](https://user-images.githubusercontent.com/1694410/47055649-7c895d00-d186-11e8-94e9-db20af559517.png)

ericedem:
Changing the character rendered in empty blocks from a zero width char to a normal space fixes this issue. And copy pasting still works as far as I can tell:

https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/leaf.js#L145

Anyone see any drawbacks to this?

steobrien:
@ericedem I think the issue then will be that clicking on an empty block can momentarily show the cursor in the wrong position:
<img src="https://user-images.githubusercontent.com/1694410/47218132-c89bf380-d378-11e8-81f3-e4b8db3bef3b.gif" width="200" />

bryanph:
What browser do you use? In my case (firefox, linux), the selection is never highlighted like you mentioned for empty blocks.

ericedem:
@bryanph I think @steobrien is just demonstrating what happens when you render a normal space character like I suggested instead of the normal zero width character.

As I was experimenting with some other styling solutions, I noticed that Chrome and Firefox (on mac) don't show the selection for empty blocks. Safari is very different, and expands the selection across the entire component.

[...]

PR #2299, opened 2018-10-22
Use <br /> inside empty block so those appear selected

kalley:

Fixing more or less. 

#### What's the new behavior?

Selected empty blocks now look selected as well
image

Instead of rendering a zero-width no-break space in an empty block, render a `<br />` instead

This is what other editors do (I checked quill and prosemirror specifically). It didn't appear to break any expectations, but I did only test in Chrome. I tried the old zero-width whitespace, as well as no-break space and neither would appear selected. I went through old PRs etc to make sure I had the back story, and tested copying the block and pasting it back in and that worked as well.

Fixes: #2275
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/leaf.js
+++ b/packages/slate-react/src/components/leaf.js
@@ -142,7 +142,7 @@ class Leaf extends React.Component {
       parent.text === '' &&
       parent.nodes.last() === node
     ) {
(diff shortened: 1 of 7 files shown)


kalley:
Looks like I missed something at first. Closing this while I try to figure it out.

Issue #2307, opened 2018-10-24
Shift+left and Shift+right hotkeys do not work in slate-react 0.19.4


dannsam:
#### What's the current behavior?

Using Shift+Left and Shift+Right hotkeys stopped working in slate-react@0.19.4
It works find in 0.19.3

It can be easily replicated on all the examples https://www.slatejs.org.

#### What's the expected behavior?

Shift+Left and Shift+Right hotkeys should extend selection backward or forward.

It seems that it was caused by https://github.com/ianstormtaylor/slate/pull/2169
cc author of that PR @ericedem

Slapbox:
I concur regarding #2169. Before I finished reading your comment I remembered seeing a commit about arrow key movement, then came back to paste it and saw you beat me to the punch.

Slapbox:
Confirmed that checking out the commit just prior restores proper selection behavior, but I can't for the life of me see anything wrong with the code or the tests. Very weird. Perhaps @ericedem will have more luck since he might have a better idea of where to look.

ericedem:
Oh nice catch guys, thanks! I'll take a look.

ericedem:
@dannsam @Slapbox Where can I find standard behavior on how this works for feature parity? I tried doing it on mac / windows and the sites I tried behave the same as slate in 0.42.

Was this a custom feature for slate?

ericedem:
Ah sorry nevermind, I must have had old cold locally. Sorry about that 😅 .

Slapbox:
Programatically I'm not sure how it's handled, but the behavior should be the same as in any text editor. It should just select the character to the left or right like this: http://recordit.co/t68e41QNED

It would be awesome to have a simple way to test previous versions, but right now the best thing I'm aware of is to clone the repo and checkout the earlier commit you want to try.

PR #2314, opened 2018-10-24
fix: extend forward / backward

ericedem:

_bug_

#### What's the new behavior?

Fix behavior for <kbd>shift + left</kbd> and <kbd>shift + right</kbd>. Now they will correctly move forward / backward.

turns out that the hotkey `'left'` would also pick up with a modifier key `'shift+left'` so checking for `Hotkeys.isExtendForward()` needs to come before `Hotkeys.isMoveForward()` and it needs to short circuit with `return true`. Likewise for backward.

Fixes: #2307 
Reviewers: @ianstormtaylor

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "html-webpack-template": "^6.1.0",
     "image-extensions": "^1.1.0",
     "immutable": "^3.8.1",
-    "is-hotkey": "^0.1.1",
+    "is-hotkey": "^0.1.4",
     "is-url": "^1.2.2",
     "jsdom": "^11.5.1",
     "lerna": "^2.7.1",
--- a/packages/slate-hotkeys/package.json
+++ b/packages/slate-hotkeys/package.json
@@ -13,7 +13,7 @@
     "lib/"
   ],
   "dependencies": {
-    "is-hotkey": "^0.1.3",
+    "is-hotkey": "0.1.4",
     "slate-dev-environment": "^0.2.0"
   },
   "scripts": {
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -526,6 +526,33 @@ function AfterPlugin(options = {}) {
       return change.moveFocusToEndOfBlock()
     }
 
+    // Note, these need to come before `isMoveForward()` and `isMoveBackward()`
+    if (Hotkeys.isExtendBackward(event)) {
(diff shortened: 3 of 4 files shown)


ericedem:
It does seem strange that `'shift+left'` would get picked up by the hotkey looking for `'left'`. I would think that should only happen if it were `'shift?+left'` but I could be missing something here.

skogsmaskin:
I think it the order as you have done here in this PR is the right one (most complex combination tested first). Isn't it just that a test for 'left' would also be true when pressing shift-left?

But I'm curious about why you had to return true and not just the change like before?

ericedem:
@skogsmaskin The problem is that if it doesn't match the void / empty check, then no change will be returned and the onKeyDown handler will continue checking hotkeys. Eventually it reaches the `moveForward()` and `moveBackward()` handlers which still match the <kbd>shift</kbd> modified arrow key.

I think the root cause is that the unmodified hotkey should not match the <kbd>shift</kbd> modified hotkey. Filed an issue with `is-hotkey` here: https://github.com/ianstormtaylor/is-hotkey/issues/11

skogsmaskin:
Oh, so it's the other way around? Shift-left evals true for *only* left too? That must be a bug.
Maybe something [here](https://github.com/ianstormtaylor/is-hotkey/blob/master/src/index.js#L151) that doesn't take non-character keys into account. I suggest we fix the root-cause (is-hotkey) first, then solve this without the workaround.

ianstormtaylor:
Thanks for tracking that down @ericedem and @skogsmaskin! With `0.1.4` of `is-hotkey` I believe this should be fixed, so maybe we can just bump instead of reordering now?

ericedem:
> maybe we can just bump instead of reordering now?

Works for me! Looking at upgrading, we seem to have two different dependencies on `is-hotkey`, one as a dev dependency at the root level 0.1.1, and one in slate-hotkeys at 0.1.3. Not super familiar with yarn, trying to sort through that now.

skogsmaskin:
Just bump both, I think the package.json one is for the examples.

ericedem:
I manually modified both package.json files to `^0.1.4` and ran `yarn install`. Fixed the bug for me locally.

ianstormtaylor:
Thanks @ericedem and @skogsmaskin, this is great!

dannsam:
Awesome work @ericedem @skogsmaskin!

Issue #2312, opened 2018-10-24
operations need to normalize all ancestors


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

Right now when an operation is normalized, we [determine which paths need to be normalized](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/controllers/change.js#L319-L379) based on which type of operation it is. 

https://github.com/ianstormtaylor/slate/blob/ea6c4dcb08734a2480f4c8faaad1986bbaa0e7da/packages/slate/src/controllers/change.js#L319-L379

However, we are currently missing out on the ancestors of the target nodes in many situations. This leads to `marks` validation not working properly, and things like `first/last` not working properly in certain scenarios.

#### What's the expected behavior?

We need to always include all of the ancestor paths of the dirty paths (instead of just the target node, or the target and its parent), so that the normalization can continue up the document tree.

Slapbox:
@ianstormtaylor do you have a fix in mind for this/any rough estimate of when we might see a fix if so?

I updated to `0.42` and I'd like to stay on it and try to help discover/clean up loose ends, but this bug is pretty brutal and I'm torn between staying the course and rolling back my commits to when I was using `0.41`

ianstormtaylor:
@Slapbox I'm working on a few other things right this moment, but I'd get to this at some point soon-ish otherwise. The fix is somewhat straightforward. As an example, in the case of an `insert_node` right now at `[0,0,1]`...

https://github.com/ianstormtaylor/slate/blob/ea6c4dcb08734a2480f4c8faaad1986bbaa0e7da/packages/slate/src/controllers/change.js#L332-L337

We'd end up marking a few paths as dirty:

- The parent's path: `[0,0]`
- The node's path: `[0,0,1]`
- And all of the node's children's paths: `[0,0,1,...]`

But to be fully correct, we need to make sure the dirty paths include all of the node's ancestors, and not just its direct parent, including:

- The grandparent's path: `[0]`
- The document's path: `[]`

To do so, we can actually simplify some of the existing logic, by introducing a `PathUtils.getAncestors` (we can find a better name later) utility:

```js
PathUtils.getAncestors(path).toArray()
// [
//   [],
//   [0],
//   [0,0],
// ]
```

Which we can then use to easily add all of the higher-up paths to the dirty paths for each of the different operations:

```
return [...ancestors, path, ...children]
```

[...]

PR #2315, opened 2018-10-24
Fixes the worst schema failures in 0.42.x

Slapbox:

Bug fix

#### What's the new behavior?

Parent nodes are normalized on `set_node` operations

`set_node` operations normalize using the same logic as `insert_node` operations. This fixes `block` nesting issues, ~~include the `first` and `last` rules~~ _(It seems to fix some of these, but not all)_. This does not fix issues with `marks` (or with `insert_text`, though I'm not aware of any bugs related to that)

(Partially) Fixes: #2312
Reviewers: @ianstormtaylor @zhujinxuan @thesunny 

**To be clear, things are still pretty broken, but this is a definite improvement in my testing**

### DIFF
--- a/packages/slate/src/controllers/change.js
+++ b/packages/slate/src/controllers/change.js
@@ -327,7 +327,7 @@ function getDirtyPaths(operation) {
     case 'set_mark': {
       return [path]
     }
-    
+
     case 'set_node':
     case 'insert_node': {
       const table = node.getKeysToPathsTable()
--- a/packages/slate/src/controllers/change.js
+++ b/packages/slate/src/controllers/change.js
@@ -324,11 +324,11 @@ function getDirtyPaths(operation) {
     case 'insert_text':
     case 'remove_mark':
     case 'remove_text':
-    case 'set_mark':
-    case 'set_node': {
+    case 'set_mark': {
       return [path]
     }
-
+    
+    case 'set_node':
     case 'insert_node': {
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @Slapbox thank you for this, but I'd rather have the fix be complete. Because as currently written this will cause lots of extra unnecessary text nodes to be normalized whenever `set_node` occurs I believe.

Slapbox:
I have no doubt it does, and it's very hacky. I don't really think this PR should be integrated, but currently Slate `0.42` is unusable if you rely on schema rules that rely on `types` as opposed to `objects`.

I'm desperate to see some improvement in this so I don't need to do some crazy branching and re-merging later in my own project. I also _really_ love the concept of the Editor controller and I never want to go back.

Is there any ETA on a fix, or anything you can explain briefly that would help me to provide a proper solution worthy of integration? I'm guessing this is beyond my pay grade, but I'd love to move this along.

ianstormtaylor:
@Slapbox just added some more information here: https://github.com/ianstormtaylor/slate/issues/2312#issuecomment-432837627

bryanph:
I've been trying to do some performance optimizations in my app and came upon this PR. I'm having a hard time understanding why all ancestors have to be marked as dirty. For example on every `insert_text` operation all parent nodes currently get marked as dirty which can be quite a lot of nodes if the particular text node is nested a few levels. Could you explain?

PR #2316, opened 2018-10-24
fix: use ancestors for dirt paths

ericedem:

_bug_

#### What's the new behavior?

Now on normalization, anywhere we were marking parent paths as dirty, we will now also mark all ancestors as dirty using the new `PathUtils.getAncestors()` function.

Fixes: #2312
Reviewers: @ianstormtaylor @Slapbox

### DIFF
--- a/packages/slate/src/controllers/change.js
+++ b/packages/slate/src/controllers/change.js
@@ -332,20 +332,20 @@ function getDirtyPaths(operation) {
     case 'insert_node': {
       const table = node.getKeysToPathsTable()
       const paths = Object.values(table).map(p => path.concat(p))
-      const parentPath = PathUtils.lift(path)
-      return [parentPath, path, ...paths]
+      const ancestors = PathUtils.getAncestors(path).toArray()
+      return [...ancestors, path, ...paths]
     }
 
     case 'split_node': {
-      const parentPath = PathUtils.lift(path)
+      const ancestors = PathUtils.getAncestors(path).toArray()
       const nextPath = PathUtils.increment(path)
-      return [parentPath, path, nextPath]
+      return [...ancestors, path, nextPath]
     }
 
     case 'merge_node': {
-      const parentPath = PathUtils.lift(path)
+      const ancestors = PathUtils.getAncestors(path).toArray()
       const previousPath = PathUtils.decrement(path)
-      return [parentPath, previousPath]
+      return [...ancestors, previousPath]
     }
 
     case 'move_node': {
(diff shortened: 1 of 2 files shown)


ericedem:
@Slapbox Did you have any use cases I could use to test this?

ericedem:
Also feel free to steel part or all of this code if you have a better idea 😄 .

Slapbox:
@ericedem I don't have a ready-made test case, though I plan to make one in the future.

For now, I'm building Slate with your change and dropping it into my project. I'll let you know how it goes momentarily. I'll be working on trying to patch up some things in Slate for at least the next hour. If you want to communicate more quickly feel free to ping me on Slack.

Slapbox:
The code you provided works, but it only fixes one problem in my schema oddly. I can no longer nest codeblocks inside of codeblocks, but I can still nest everything else inside of anything. I'm not sure what to make of that yet.

Slapbox:
@ericedem, your change works pretty well when combined with https://github.com/ianstormtaylor/slate/pull/2315

@ianstormtaylor - I'm not grasping why setting a node type doesn't require parents to be revalidated. It seems like if you have a situation where you're not allowed to nest a certain node within another that it would need to revalidate the parent, no?

ericedem:
@Slapbox nice, that fixes the issue I was seeing too!

Slapbox:
@ericedem unfortunately even combined it doesn't fix this example I made. Copy this over your ``Rich Text Example` `index.js` https://gist.github.com/Slapbox/cd5eb0809ef854f25bc3e74fd6135cc4

It should disallow marks in paragraphs, and it does on start, but not afterwards. If you add a bold mark, then change the block to a blockquote, and then back again, it will renormalize it properly, but not before you do all that.

packages/slate/src/controllers/change.js
```diff
@@ -332,20 +332,20 @@ function getDirtyPaths(operation) {
     case 'insert_node': {
```
ianstormtaylor: I think we need to use `...ancestors` for the first switch block too, to allow for `marks` rules to trigger on higher-up blocks.

ericedem: Ahh good call, that works for me locally 👍

ericedem:
@Slapbox I tested the schema code you provided and it seems to work now with the change @ianstormtaylor recommended.

Slapbox:
@ianstormtaylor @ericedem this looks great so far! Testing it in my full setup now.

packages/slate/src/controllers/change.js
Slapbox: This is probably good. I'm seeing two bugs, but it's more likely than not that this is caused by changes to my schema rules from before I realized the problem wasn't with my rules.

grsmvg:
This solved my bug also, @ericedem . Thanks for the quick responses (also @ianstormtaylor). I'm not yet comfortable enough with the source code to do these PR's myself but I'll keep reading along until I do. 🎊

Issue #2291, opened 2018-10-21
Delete all will fail when there are three blocks or more


skogsmaskin:
**EDIT**: previously this was only an issue when the last block ended in a inline void node. Now it is *always* an issue.

#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?

If you have three blocks or more, where the last block ends with an inline void node, and you select everything and hit backspace (delete), it will fail with error:
```
Uncaught Error: `Node.assertNode` could not find node with path or key: List [ 1 ]
```

Reproduction:

1) Goto the [emoji-example](https://www.slatejs.org/#/emojis)
2) Add an emoji at the end of the last block. 
3) Select everything (command/ctrl-A)
3) Hit backspace

#### What's the expected behavior?
That everything will be deleted without error.

I have noticed that when deleting all (and you have three or more blocks) the change will include an operation with type ``move_node`` where the path property is a block path (i.ie ``[2]`` with the same value as the newPath prop ``[2]``. I think Slate will regenerate keys for this operation and it may lead to issues.

Slapbox:
Seeing the same thing. Strong suspicion it's related to https://github.com/ianstormtaylor/slate/issues/2279

Slapbox:
I've narrowed down where this error occurs to at least here: https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/commands/at-range.js#L238-L242

The error itself seems like to be in mergeNodeByKey

Slapbox:
More specifically it fails on a second pass through here:

https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/commands/by-path.js#L170-L183

Error is: ``Node.assertNode` could not find node with path or key: List [ 1 ]`

skogsmaskin:
Here are some findings, all seems related to the algorithm inside [``.deleteAtRange``](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/commands/at-range.js#L66)

* If we remove [``change.withoutNormalizing``](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/commands/at-range.js#L101) this case will work as intended but breaks other tests.
* If we remove [``moveNodeByKey``](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/commands/at-range.js#L229) this case will work as intended but breaks other tests.

[...]

PR #2322, opened 2018-10-25
Exempts remove_text operation from traversing ancestors

Slapbox:
**_This is not necessarily good to go_ - I've submitted it to make reviewing the change easier - It may be worth merging or it may not**

Bug fix

#### What's the new behavior?

Exempt

Exempts remove_text operation from traversing ancestors

(Partially) Fixes: #2291 - Restores functionality prior to merging of #2316
Reviewers: @ianstormtaylor @ericedem @skogsmaskin

### DIFF
--- a/packages/slate/src/controllers/change.js
+++ b/packages/slate/src/controllers/change.js
@@ -320,10 +320,13 @@ function getDirtyPaths(operation) {
   const { type, node, path, newPath } = operation
 
   switch (type) {
+    case 'remove_text': {
+      return [path]
+    }
+
     case 'add_mark':
     case 'insert_text':
     case 'remove_mark':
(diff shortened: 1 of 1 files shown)


ericedem:
I think we do need to traverse text ancestors, because you could have a schema rule that demands a text node as a child of another node.

Slapbox:
@ericedem I agree. I just figured this should help pinpoint the cause. Prior to this I figured it was related to removing, merging, or moving nodes, but it seems like it's related to the text operation.

ianstormtaylor:
I think this isn’t what we need since like @ericedem says we need to be traversing ancestors there. But thank you @Slapbox!

PR #2324, opened 2018-10-25
Fix invalid transform path in move_node operation

skogsmaskin:

Fixing a bug. 

Moving a node from ``path`` to an equal ``newPath`` would increment the returned path when it shouldn't.

Fixes: #2291
Reviewers: @ericedem @ianstormtaylor @Slapbox


Slapbox:
This looks great and works (for inlines as the last node too!)

skogsmaskin:
Special thanks to @ericedem for pinpointing the buggy code location!

ianstormtaylor:
Thank you very much @skogsmaskin for taking the time to do this!

Issue #2334, opened 2018-10-26
simplify the "Change vs. Editor" question


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Idea.

#### What's the current behavior?

This is something I was discussing with @alanctkc in Slack. Right now, there are two "main" instances that you care about when developing plugins: the `Editor` and the `Change`.

This issue is that they are used semi-interchangeably, but they're different. For example commands receive a `change` object. Whereas queries receive an `editor` object. And event handlers receive a `change` object. But components should receive an `editor`.

That alone isn't a huge deal, but it does complicate things.

But things get worse when you consider that there is confusion between `editor.value` and `change.value`—people might not even realize they are different. But `editor.value` is the currently rendered value of the editor, and it only changes after new props are passed in. Whereas `change.value` is the most-up-to-date value object that has the newest operations in the change applied to it.

This distinction is a bit complex. It means that you can't just access `editor.value` in your queries and have everything work as expected.

#### What's the expected behavior?

A fairly simple solution that @alanctkc came up with is to change queries to actually be passed a `change` object, which will default to the currently active change. This seems strange, but it's actually kind of ingenious because it means that queries can always access `change.value` as the most-up-to-date.

This could work.

---

I was wondering though, if there might be a slightly different solution that we could use, by getting rid of the current `Change` object altogether. Sounds drastic, but it might actually result in a simpler mental model.

It would entail...

- [ ] **Registering commands and queries directly on editor.** Right now if you want to invoke a command on the editor you do `editor.command('insertText', 'word')`. Whereas if you want to invoke it on a change you do `change.insertText('word')`. Under the new architecture, the commands and the queries would be available as top-level methods on the editor itself:

```js
editor.insertText('word').splitBlock()
```
```js
if (editor.isQuoteAllowed()) {
  editor.setQuoteBlock()
}
```

- [ ] **Tweak changes would be emitted asynchronously.** To make this possible, we'd tweak the current logic to emit changes via `onChange` in an asynchronous way, more similarly to how `setState` works. (The current synchronous architecture might [already be causing problems](https://github.com/ianstormtaylor/slate/issues/2282).) Any command would queue a change (just like React state does) and then the queue is flushed on the next tick. That way you can continue chaining commands synchronously just like you can now.

[...]

Issue #2282, opened 2018-10-19
react error "Cannot update during an existing state transition"...


Ognian:
Hello,
using "slate-react": "^0.19.3",
I've got the following react warning ( which is actually an error)
```
index.js:1452 Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.
```
![image](https://user-images.githubusercontent.com/1231174/47207542-bd9e8e80-d38b-11e8-9fae-1d8a337cddc2.png)

Digging deeper I found that indeed the render method of [editor.js](https://github.com/ianstormtaylor/slate/blob/53fd59bc62b42e6f57d425da11d492da6e3cb05d/packages/slate-react/src/components/editor.js#L155) is calling setValue which in turn is is calling the onChange chain leading in the top most component to a setState...

To be able to reproduce the error you need first a short text (value in Editor) and then a much longer text triggered from outside (change value of Editor from short to quite long text), otherwise somehow it does not reproduce...

Any ideas on this
Thanks
Ognian

PR #2337, opened 2018-10-26
remove change, fold into editor

ianstormtaylor:

Improvement / debt.

#### What's the new behavior?

This pull request removes the `Change` object as we know it, and folds all of its behaviors into the new `Editor` controller instead, simplifying a lot of the confusion around what is a "change vs. editor" and when to use which. It makes the standard API a **lot** nicer to use I think.

---

###### NEW

**The `editor.command` and `editor.query` methods can take functions.** Previously they only accepted a `type` string and would look up the command or query by type. Now, they also accept a custom function. This is helpful for plugin authors, who want to accept a "command option", since it gives users more flexibility to write one-off commands or queries. For example a plugin could be passed either:

```js
Hotkey({
  hotkey: 'cmd+b',
  command: 'addBoldMark',
})
```

Or a custom command function:

```js
Hotkey({
  hotkey: 'cmd+b',
  command: editor => editor.addBoldMark().moveToEnd()
})
```

###### BREAKING

**The `Change` object has been removed.** The `Change` object as we know it previously has been removed, and all of its behaviors have been folded into the `Editor` controller. This includes the top-level commands and queries methods, as well as methods like `applyOperation` and `normalize`. _All places that used to receive `change` now receive `editor`, which is API equivalent._

**Changes are now flushed to `onChange` asynchronously.** Previously this was done synchronously, which resulted in some strange race conditions in React environments. Now they will always be flushed asynchronously, just like `setState`.

**The `render*` and `decorate*` middleware signatures have changed!** Previously the `render*` and `decorate*` middleware was passed `(props, next)`. However now, for consistency with the other middleware they are all passed `(props, editor, next)`. This way, all middleware always receive `editor` and `next` as their final two arguments.

**The `normalize*` and `validate*` middleware signatures have changed!** Previously the `normalize*` and `validate*` middleware was passed `(node, next)`. However now, for consistency with the other middleware they are all passed `(node, editor, next)`. This way, all middleware always receive `editor` and `next` as their final two arguments.

**The `editor.event` method has been removed.** Previously this is what you'd use when writing tests to simulate events being fired—which were slightly different to other running other middleware. With the simplification to the editor and to the newly-consistent middleware signatures, you can now use `editor.run` directly to simulate events:

```js
editor.run('onKeyDown', { key: 'Tab', ... })
```

###### DEPRECATED

**The `editor.change` method is deprecated.** With the removal of the `Change` object, there's no need anymore to create the small closures with `editor.change()`. Instead you can directly invoke commands on the editor in series, and all of the changes will be emitted asynchronously on the next tick.

```js
editor
  .insertText('word')
  .moveFocusForward(10)
  .addMark('bold')
```

**The `applyOperations` method is deprecated.** Instead you can loop a set of operations and apply each one using `applyOperation`. This is to reduce the number of methods exposed on the `Editor` to keep it simpler.

**The `change.call` method is deprecated.** Previously this was used to call a one-off function as a change method. Now this behavior is equivalent to calling `editor.command(fn)` instead.

---

Fixes: https://github.com/ianstormtaylor/slate/issues/2334
Fixes: https://github.com/ianstormtaylor/slate/issues/2282


skogsmaskin:
@ianstormtaylor - do you have an rough estimate on when this will be ready to be merged?

Issue #2319, opened 2018-10-25
code-highlighting  example fix


An0nymous0:
1.https://www.slatejs.org/#/code-highlighting

2.Two carriage returns

![image](http://g.recordit.co/SWCGHLgb2k.gif)

ericedem:
I think I found a bit of a better solution for this. As best I can tell, decorations themselves never get normalized. This is why we end up with `null` values on the points even though they should be getting normalized away.

There are two entry points for creating decorations `setDecorations()` and `decorateNode()`. The approach I'm thinking is that `Decoration` should have a normalization method that looks like:

```js
normalize(node) {
  return this.merge({
    anchor: this.anchor.normalize(node),
    focus: this.focus.normalize(node),
    mark: this.mark.normalize(node),
  })
}
```

And then we should normalize when we have access to a node. So in [`getDecorations()`](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/interfaces/element.js#L452), we normalize before they get used anywhere:

```js
return decorations.map(dec => dec.normalize(this))
```

We need to do it here, because `getDecorations()` gets called when the document is being rendered already, and we can't wait for a top level normalization. This seems to work pretty well locally but I havne't done extensive testing. I'm most worried about performance.

We also need to normalize at the top level in case someone passes decorations in directly through something like `setDecorations()`. I think this needs to happen in `Editor#normalize()` along with the document normalization, but still wrapping my head around how this would work. Probably needs to happen after the document finishes normalizing, and just another map to normalization like the example above.

@ianstormtaylor What do you think?

tommoor:
This is by far the most common error coming out of our Slate editor… @ericedem still thinking of taking on this fix?

ericedem:
@tommoor Ahh sorry, this one fell off my radar. This is the branch I was working out of [is here](https://github.com/ericedem/slate/tree/fix-code-highlighting-example) if you want to take a look at the code for your own PR. I haven't been working on Slate much recently but will probably be digging back in in the next few weeks and might be able to tackle this.

thesunny:
Fixed it here https://github.com/ianstormtaylor/slate/pull/2698

PR #2338, opened 2018-10-26
fix: make to have a path when comparing decoration paths.

ericedem:
If no path, use a key to get a path.

_bug_

This example is specifying decorations with keys rather than paths (which is valid). When rendering, the Text nodes were trying to reference and compare the paths of the nodes which did not exist.

This change makes it so if there is no path when trying to compare paths, get the path from the key.

Fixes: #2319
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/text.js
+++ b/packages/slate-react/src/components/text.js
@@ -119,10 +119,12 @@ class Text extends React.Component {
 
       // If the node's path is before the start path, ignore it.
       const path = document.assertPath(key)
-      if (PathUtils.compare(path, start.path) === -1) return false
+      const startPath = start.path || document.assertPath(start.key)
+      if (PathUtils.compare(path, startPath) === -1) return false
 
       // If the node's path is after the end path, ignore it.
-      if (PathUtils.compare(path, end.path) === 1) return false
+      const endPath = end.path || document.assertPath(end.key)
+      if (PathUtils.compare(path, endPath) === 1) return false
 
       // Otherwise, include it.
       return true


ianstormtaylor:
Hey @ericedem, the goal here makes sense. But I think it would be better for us to ensure that the paths are correct at the time the decorations are created (as in pass each one through `document.createPoint` or whatever), and update the decorations themselves. That way we aren't keeping unset decorations around for other places to get tripped up.

ericedem:
@ianstormtaylor Ok I ended up doing some more digging, and I found something pretty confusing. I did find comments that it looks like the points are supposed to normalize and pick up a path. But it looks like when Points are created with a "key" and not a "path" in the current codebase, the key gets nuked before the path can normalize.

I wrote a small test to illustrate

```js
it('should work', () => {
  const point = new Point({ key: '12', offset: 42, path: null })

  const result = point.toJS()
  assert.equal(result.key, '12')
  assert.equal(result.offset, 42)
  assert.equal(result.path, null)
})
```

Which fails:

```sh
  1 failing

  1) slate should work:
     AssertionError [ERR_ASSERTION]: undefined == '12'
      at Context.<anonymous> (packages/slate/test/index.js:150:12)
```

As best I can tell, `Slate.Point` just inherits the constructor from `Immutable.Record`. Stepping through there also seems like the right thing is happening. This issue seems like it should be breaking a whole lot more things in slate, but maybe we have reliably moved to paths in most places so it isn't a problem.

ericedem:
Ugh, it looks like this is a bug in Immutable maybe? Using normal getters work but serializating `toJS()` or `toJSON()` is broken. This works:

```js
assert.equal(point.get('key'), '12')
assert.equal(point.get('offset'), 42)
assert.equal(point.get('path'), null)
```

ianstormtaylor:
@ericedem for `toJS` make sure to remember `toJS({ preserveKeys: true })`, otherwise they won't be serialized by default.

ericedem:
Ok I'm going to close this in favor of having some more discussion with a different approach in #2319

PR #2344, opened 2018-10-28
add clarification for paste HTML example

### LINKED ISSUE #1392: Paste HTML example is broken.

The Paste HTML example doesn't work, and neither does the code in my own project. Pasting a HTML string simply renders that string as-is into the editor. 

To reproduce: 

- visit http://slatejs.org/#/paste-html
- select all the text content, and paste `<p>Test</p>`
- It will display, verbatim: `<p>Test</p>`

Pasting HTML should convert that HTML into proper rendered markup. Pasting the above should display "Test", which should be correctly wrapped in paragraph tags in the DOM.

wildlifehexagon:
fixes #1392

This just adds clarification to the paste HTML example that confused myself and others (as evidenced by #1392).

#### What's the new behavior?
N/A

This only changes one line in the JSON example.

Fixes: #
Reviewers: @

### DIFF
--- a/examples/paste-html/value.json
+++ b/examples/paste-html/value.json
@@ -25,7 +25,7 @@
             "leaves": [
               {
                 "text":
-                  "Try it out for yourself! Copy and paste some rendered HTML content (not the actual syntax) from another site into this editor."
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @wildlifehexagon this is great!

Issue #2360, opened 2018-11-01
Bug on Mentioning example -- `.withoutSaving` is not a function


zwallacedev:
Reporting a bug

#### What's the current behavior?
Currently, entering text into the mentions example requires two keystrokes into the editor of the same key for it to respond.

Here is a recordit of the bug: http://g.recordit.co/FGQkKoXQor.gif
- Current OS is macOS Sierra
- Current browser is Chrome (latest version)
- Slate is the most current version because it's on the website example page

#### What's the expected behavior?
The expected behavior is that the editor should `insertText` on the first keystroke. This will most likely be fixed by following the trail left by the console. I will enter a PR later today for a potential fix.

As a side note -- I am working on an independent plugin that reflects the example provided, and I will hopefully release this as an open-sourced plugin. Thanks again for all the hardwork that went into producing this product.

ericedem:
Oh thanks for filing. Looks like this didn't get refactored in the latest 0.43 release to use editor directly instead of working through changes.

zwallacedev:
@ericedem no problem. I was going to take a look at the issue mentioned above after work, but if you know the solution, I'll let you take it 😄

ericedem:
@zwallacedev Thanks! I wrote this example so it's a pretty quick fix.

> I am working on an independent plugin that reflects the example provided, and I will hopefully release this as an open-sourced plugin. Thanks again for all the hardwork that went into producing this product.

That's awesome! Please feel free to ping me on [slack](https://github.com/ianstormtaylor/slate/blob/master/Contributing.md#asking-questions) if you have any questions. This example is based off the custom implementation I did on https://www.cake.co. I did it as a react component here to keep things simple, but if you do something open source you will need to make it a plugin which will come with some additional complexity. Take a look at the doc comments in the example if you haven't already for an idea of what will need to be done for something production ready.

zwallacedev:
@ericedem that's awesome to hear that this will be a pretty quick fix, and thank you for the example. The example has helped me out tremendously!

I will most likely ping you at some point for some assistance. I am working on a plugin implementation as an example for our pilot program into slate for my company. We just began moving over towards slate, but we still aren't 100% sold. So this first plugin will be beneficial, not only to our team but also toward a strive for us to move some of our product towards open source as well. I have definitely run into some complexity with creating the plugin, but I am also in the midst of learning the slate editor as well, so it's been a decent learning curve for sure 😄.

[...]

PR #2363, opened 2018-11-01
fix: update mentions example to slate 0.43

ericedem:
It also works in Safari now :)

_bug_

Convert to using the editor directly instead of a change object.

Fixes: #2360
Reviewers: @ianstormtaylor 

#### Caveat

I'm not super excited about how I had to fix the `onChange()` handling, by wrapping it in a `setState()` callback. It seems to work ok, but it feels a bit off. If anyone has a similar issue or recommendations please let me know!


ianstormtaylor:
Thanks @ericedem!

Issue #2257, opened 2018-10-13
consider implementing "placeholders" as decorations


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Idea.

#### What's the current behavior?

Right now we have the special `renderPlaceholder` logic, but it gets a bit opinionated about where to render the placeholders, and it's a bit more confusing to understand how it works because it adds yet another concept to the rendering logic.

#### What's the expected behavior?

I wonder if we could replace it purely with decorations instead. The default case would be a decoration applied to the document only when it has no text. And then maybe it can be styled with `::before` or `::after`? Or with `contenteditable="false"` children? Or with portals? Not exactly sure what will result in proper browser support.

bryanph:
Perhaps we can define this in the schema and pass a prop in renderNode that indicates this node needs to be rendered as a placeholder and let the user be responsible for rendering the node as a placeholder?

ianstormtaylor:
@bryanph I think it should remain a rendering-only concern though, and not be tied to the document structure though.

ianstormtaylor:
@bryanph although maybe I’m misunderstanding! Perhaps if we had `renderNode` called on the `document` node then we’d already have the ability to do these things. Is that what you mean? That could be cool.

bryanph:
@ianstormtaylor not really but that sounds like it might be better idea :smile: 

The `renderPlaceholder` does two things right?

* Determine if the block being rendered is considered "empty"
* Actually rendering the placeholder

So I thought maybe that first part can be done in the schema somehow or just in slate internally because a block being empty just means that 

1. it is a block
2. it has no children

right? and then we can just call renderNode with a isEmpty prop or something like that so that we can render a placeholder instead.

Issue #2067, opened 2018-08-09
don't re-render editor during compositions


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Bug,

#### What's the current behavior?

Right now [we re-render the editor to show/hide the placeholder](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/before.js#L144-L170) during `compositionstart` and `compositionend` events. But this results in the composition being aborted/broken in some IMEs.

#### What's the expected behavior?

We need to change this logic to no re-render the editor, but just re-render the placeholder itself I think. This could be accomplished with some sort of "connected" `<Placeholder>` component I believe, which can be toggled from the plugin code.

This is related to #2062.

zhujinxuan:
Hi, I am not sure why we put isComposing in state.  `this.setState` is not immediate, and can cause re-render.

Is there a special reason of putting isComposing in state?

ianstormtaylor:
@zhujinxuan check out the comments in the source linked above.

PR #2370, opened 2018-11-02
Remove `renderPlaceholder`

ianstormtaylor:

Improvement.

#### What's the new behavior?

This pull request changes the placeholder logic to use decorations instead of the one-off `renderPlaceholder` middleware, which makes it more flexible. It also fixes some issues related to selection tracking that created placeholder-related bugs.

The placeholder logic is packaged as a separate `slate-react-placeholder` plugin, which can be used for custom placeholders. (Or for super custom use cases you can write your own.)

The `<Editor>` component will now automatically add an instance of the `slate-react-placeholder` component to its stack when it receives a `placeholder=` prop, to best emulate regular DOM editors. But by omitting this prop you can handle the behaviors completely on your own instead.

Fixes: https://github.com/ianstormtaylor/slate/issues/2067
Fixes: https://github.com/ianstormtaylor/slate/pull/2168
Fixes: https://github.com/ianstormtaylor/slate/issues/2257
Fixes: https://github.com/ianstormtaylor/slate/issues/1678
Fixes: https://github.com/ianstormtaylor/slate/issues/1344


garth:
I don't see any documentation on how to get placeholders working after this change. Simply having a placeholder prop on editor does not work, adding the new placeholder package as a plugin causes render mark to be called with a type of 'placeholder', but if render mark returns a value then it becomes editable (not a placeholder)?

ianstormtaylor:
@garth make sure you have the latest two versions of `slate` and `slate-react`. There was a bug fix in `slate` that is required, and I forgot to bump the peer deps until `.1` of this new release.

garth:
@ianstormtaylor yes, I have the latest version of all slate packages, but still can't get the placeholder working. I've tried a few different ways to get it working, but not had any luck so far. Is there an example of how to use the new version?

ianstormtaylor:
@garth most of the existing examples use a `placeholder=` prop and they seem to work.

garth:
ok, I found the problem. The signature for `renderMark` and `renderNode` has been changed. So next() was never called.

garth:
BTW: the typescript typings have not been updated, which is why I didn't pick this up earlier.

```
[ts] Type '(props: RenderNodeProps, editor: Editor, next: () => any) => any'
  is not assignable to type '(props: RenderNodeProps, next: () => void) => any'.
```

cudr:
Hello, @ianstormtaylor! Now it looks, like we can use only one placeholder text multiple times, because first placeholder plugin render text each time.  Maybe we can change it to multiple different placeholders? I mean, maybe we can we pass placeholder text to decorator for that?

svenadlung:
Looking for the same as @cudr. Changing the placeholder dynamically would be awesome.

cudr:
Make PR for that (#2416 )

Issue #2366, opened 2018-11-02
fix `wrapInlineAtRange` logic


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

Right now in `wrapInlineAtRange`:

https://github.com/ianstormtaylor/slate/blob/ea3443eb425fdceb7cddaf77dedd5fe13fb996cf/packages/slate/src/commands/at-range.js#L1246-L1255

But `getClosestInline` isn't guaranteed to return a node.

#### What's the expected behavior?

Needs to guard against that.

Dundercover:
Is this not solved by just adding

```
if (!inlineParent) {
  return
}
```

plus a couple of tests or am I missing something?

ianstormtaylor:
@Dundercover yup, that sounds right to me. I was just noting this while working on some adjacent parts of the code.

PR #2374, opened 2018-11-04
Check if inline ancestor exists before using it in wrapInlineAtRange

Dundercover:

Fixing a bug

Check if inline ancestor exists before using it in wrapInlineAtRange when selection is collapsed.

Fixes: #2366


ianstormtaylor:
Thank you @Dundercover!

PR #2376, opened 2018-11-05
Changing the selection when input by ime sometimes will break the ime.

yunwuxin:
Changing the selection when input by ime sometimes will break the ime.


ianstormtaylor:
Hey @yunwuxin the `state.isComposing` flag isn't toggled anymore in the latest versions of `slate-react` so we'll need to find another way to do this.

yunwuxin:
@ianstormtaylor expect

PR #2380, opened 2018-11-05
fix: support mobile devices in examples

ericedem:

_bug_

#### What's the new behavior?

Go from this:

image

To this:

image

This change adds the standard meta tag for supporting mobile devices. Now when viewing on mobile the viewport width will adapt to the device.

https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag#Viewport_basics

Fixes: #
Reviewers: @ianstormtaylor

### DIFF
--- a/support/webpack/config.js
+++ b/support/webpack/config.js
@@ -68,6 +68,16 @@ const config = {
       title: 'Slate',
       template: HtmlWebpackTemplate,
       inject: false,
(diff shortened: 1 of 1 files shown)


ericedem:
Another improvement we can make is a responsive menu that collapses the examples into a dropdown menu for small screens, since the list of examples is getting longer.

ianstormtaylor:
Thank you for solving this @ericedem!

PR #2388, opened 2018-11-06
Improve schema validation for nodes and min/max

aiwenar:

Feature/enhancement

#### What's the new behavior?

Consider the following schemas

```
nodes: [
    // ...
    { match: [set A], min: M, max: N },
    { match: [set B], min: P },
    // ...
]
```
and
```
nodes: [
    // ...
    { match: [set C], min: K, max: L },
]
```

There's a total of 7 ways in which it can be violated

1. A repeats less than M times and is followed by B
2. A repeats more than N times and is followed by B
3. B repeats less than P times
4. There is a node neither in A nor in B between those two groups
5. C repeats less than K times
6. C repeats more than L times
7. There is a node not in C following it

Currently slate reports those violations using just three codes: `child_type_invalid` for cases 1, 2, 3, and 4, `child_unknown` for cases 6 and 7, and `child_required` for case 5.

While the small number of codes compared to the number of cases is not a problem, the way they are reported is:

- In 1. the first node of B is reported as violating rules for A, but it's not; rules for A are violated by _lack of nodes_.

- In 2. the first additional node of A is reported as violating rules for B. While this is technically correct, it's a much less useful information than that there is an additional node in A, especially if those two cases are to be normalized differently.

- The same type of violation, less nodes present than required, is being reported in two different ways: as `child_type_invalid` when followed by other nodes, and as `child_required` when not.

This makes it at times necessary to reimplement schema rules in custom normalizations, as the `error` object does not provide enough information.

Instead I propose that we deprecate `child_required`, and introduce two new codes:

- `child_required_underflow`, reported when there are fewer matching nodes than required; and

- `child_required_overflow`, reported when there are more matching nodes than required.

Both codes would include the `rule` and `node` properties, common to all codes, plus the following

- `index: Number`, index of the last matching node;
- `count: Number`, number of matching nodes;
- `limit: Number`, number of nodes required to satisfy `rule`, minimum for `child_required_underflow` and maximum for `child_required_overflow`.

Thus `limit - count` is the number of nodes missing for `child_required_underflow`, and `count - limit` is the number of additional nodes for `child_required_overflow`.

With this, cases 1, 3, and 5. would be reported as `child_required_underflow`, cases 2. and 6. as `child_required_overflow`, and cases 4. and 7. as they are now, or as `child_type_invalid` and `child_unknown` respectively.

- Function `validateNode()` was rewritten to function as suggested. 
- Core schema, documentation, and relevant examples and tests, were updated to match the new behaviour.
- Two new tests were added.

* [x] The new code matches the existing patterns and styles. _(I think so)_

Fixes: 
Reviewers:


docs/reference/slate/schema.md
```diff
@@ -303,17 +303,34 @@ When supplying your own `normalize` property for a schema rule, it will be calle
 
 Raised when the `object` property of a child node is invalid.
 
-### `'child_required'`
+### `child_required_underflow`
```
ianstormtaylor: Could we change these to be `child_min_invalid` and `child_max_invalid` instead? That way they maintain the parallel with their definitions.

docs/reference/slate/schema.md
```diff
@@ -303,17 +303,34 @@ When supplying your own `normalize` property for a schema rule, it will be calle
...
+
+### `child_required_overflow`
+
+```js
+{
+  index: Number,
+  count: Number,
+  limit: Number,
+  node: Node,
+  rule: Object,
```
ianstormtaylor: In the case of overflow, do we have access to `child` so that we can include it? Just as a useful extra piece of information for people. (It would simplify the default rule already.)

aiwenar: The reason I didn't include it is that I'm not sure which of the overflowing nodes should it be, and that it would break symmetry with `child_min_invalid`, but we could add it.

As for the default rule, perhaps it would be better to just remove all overflowing nodes at once?
```
return node.nodes.slice(index - count + limit, index + 1).forEach(child => {
    editor.removeNodeByKey(child.key)
})
```

ianstormtaylor:
Thank you @aiwenar this is an amazing pull request!

valorize:
Is it still possible to use the schema node min / max numbers to specify how many nodes need to be present?
Or is it mandatory to use `child_type_invalid` and `child_min_invalid` as in https://github.com/ianstormtaylor/slate/blame/0c94ccebce18bb05eafb4466737adc6a576d1db9/examples/forced-layout/index.js#L25 ?

PR #2392, opened 2018-11-07
[UPDATE] docs on how to install and run the example packages

luuduchieu:
Help new guy clone the repo and run the example packages. As current, we need to install all the peer dependency but i haven't found it on the example document

Fixes: #
Reviewers: @

### DIFF
--- a/examples/Readme.md
+++ b/examples/Readme.md
@@ -27,6 +27,17 @@ yarn install
 yarn build
 ```
 
+For those who have problem when start, you must install all the peer dependencies first on `package.json`
+- for npm (automatically) or install manually:
+```
+npm install -g npm-install-peers
+npm-install-peers
+```
+- for yarn:
+```
+yarn install --peer
+``` 
+
 Then start the watcher and examples server:
 
 ```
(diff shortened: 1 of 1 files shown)


examples/Readme.md
```diff
@@ -27,6 +27,21 @@ yarn install
 yarn build
 ```
 
+For those who have problem when run start, you must install all the peer dependencies first on `package.json`
+
```
CameronAckermanSEL: To my knowledge, the general advice is to not use npm because it's not compatible with the yarn.lock file and we won't get the same versions of dependencies on install.

luuduchieu: Yes, i mean for the npm user, if he wants to try slate out then he must install yarn, and it is not a good experience.

CameronAckermanSEL: Yes, but providing instructions to an NPM user will imply that this tool can be correctly built with NPM. It cannot be built with NPM because it does not have the NPM lock file, it only has the yarn lock file. Therefore building with NPM is not guaranteed to pull down the same versions of each dependency.

brendancarney:
I agree with @CameronAckermanSEL: It's recommended to use `yarn` when developing slate, and not using it could cause issues that are not supported.

Issue #2390, opened 2018-11-07
onMouseDown and onMouseUp listeners being ignored


blambo:
#### Do you want to request a _feature_ or report a _bug_?

Report a bug

#### What's the current behavior?

Passing in `onMouseDown` and `onMouseUp` event listeners to the `Editor` component are being ignored. It _seems_ like adding them into EVENT_HANDLERS in https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/constants/event-handlers.js is enough to get the events to trigger (not tried on the latest version of slate-react).

Example JSFiddle: https://jsfiddle.net/qbpc6z0e/

OS: Mac OSX 10.14
Browser: Chrome 70.0.3538.77
Slate version: 39.3 (also doesn't work in 43.7)

#### What's the expected behavior?

`onMouseDown` and `onMouseUp` are called when their respective events trigger.

Issue #2352, opened 2018-10-30
Focus is lost when tabbing into a Slate editor


steobrien:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

Pressing tab from either a plain input or another Slate editor, to focus a Slate editor, results in a cursor very briefly showing (~100ms) in the editor, which then disappears and focus is lost.

Example gif from https://www.slatejs.org/#/syncing-operations:

![gifable-01df4f01-365f-406a-988a-ea451425ceee](https://user-images.githubusercontent.com/1694410/47733990-d6d2f500-dc3f-11e8-921e-8d079cbdb07e.gif)

(This is a little hard to make out: I'm clicking into the first editor and pressing tab on my keyboard. You can see a flash of cursor in the second editor a few times.)

#### What's the expected behavior?

A tabbed-to editor should obtain and maintain focus

<hr />

Some background in a related issue: https://github.com/ianstormtaylor/slate/issues/2147#issuecomment-434334329

steobrien:
For now, we’re working around this using the approach outlined in https://github.com/ianstormtaylor/slate/issues/2147#issuecomment-417965381.

Slapbox:
This is likely related to changes becoming async in `slate@0.43` - I've also had to implement some workarounds.

steobrien:
@Slapbox I think the problem may predate `0.43` – we saw this issue appear in our app when upgrading from `0.13` to `0.41`.

Slapbox:
Wow that's one big upgrade, and you're right. Here's a link where you can use Slate as it was with PR #2100 
https://deploy-preview-2100--slatejs.netlify.com/#/syncing-operations. #2100 wasn't committed, but that corresponds roughly with Slate `0.37`:

By #2112 the issue appears: https://deploy-preview-2112--slatejs.netlify.com/#/syncing-operations, which corresponds roughly with Slate `0.38` or `0.39`

Some change in that time period seems to be responsible.

You can view the times that all versions were published with the command `npm info slate time`

Issue #2210, opened 2018-09-26
Editor cursor doesn't always display when tabbing to Editor


javery014:
#### Do you want to request a _feature_ or report a _bug_?
Report a bug

#### What's the current behavior?
When tabbing through a document to a Slate editor, the cursor in the Slate editor doesn't always appear. It seems that the cursor will only appear if the previous cursor location in the Slate editor was anywhere else but the very start of the editor. 

![editor-tabbing-issue](https://user-images.githubusercontent.com/10987301/46107566-4e28eb00-c1a1-11e8-9325-6268bcb6bf72.gif)
https://jsfiddle.net/fj9dvhom/1/

#### What's the expected behavior?
I would expect whenever I tab to the Slate editor, the cursor would be visible at the start of the editor (or the end, looking at #1964).

everdimension:
I'm also bothered by this bug, but for the time being I'm using this for the fix (using `slate-react`):

```jsx
<Editor
  onFocus={(event, change) => { change.focus(); }} // <-- the fix
  value={this.state.value}
  onChange={this.handleChange}
/>
```

Issue #2137, opened 2018-08-29
Clicking on an editor doesn't always give focus


quentez:
#### Do you want to request a _feature_ or report a _bug_?

I'm reporting a bug: when using multiple editors on the same page, moving the focus from one editor to the next with the pointer is unreliable.

#### What's the current behavior?

With more than one editor on the same page, clicking from one editor to another does not always give it the focus. The behavior is non-deterministic (as a far as I can tell), and it sometimes works.

This can be reproduced on the `Syncing Operations` example on the [SlateJS website](https://www.slatejs.org/#/syncing-operations).
(For this reason, I'm not including a Fiddle).

![issue gif](https://user-images.githubusercontent.com/504055/44814561-fc854480-ab91-11e8-8bb1-fa7358a1a388.gif)

My setup:
- MacOS `10.14`
- Chrome `68.0.3440.106`
- Slate `0.40.2`

#### What's the expected behavior?

The focus should reliably move to an editor when clicking on it.

bryanph:
Probably related to #2097

Mythli:
I can confirm this. Having two editor components with just the "getting started" editor renders one editor unusable.

ericedem:
@quentez Do you still see this behavior? I can't reproduce and I think it may have been fixed in https://github.com/ianstormtaylor/slate/pull/2150

Saebyuckbaan:
I still seeing this issue on 
slate: "0.44.9"
slate-react: '0.21.15"

jrm2k6:
I am still seeing the same issue with:
```
slate@0.44.10
slate-react@0.21.16
```

FezVrasta:
![Kapture 2020-04-06 at 16 39 16](https://user-images.githubusercontent.com/5382443/78575711-0110e500-782c-11ea-9df9-826daead1011.gif)

I think I'm reproducing the same issue, see the above gif.

FezVrasta:
@CameronAckermanSEL may this issue get reopened?

CameronAckermanSEL:
Hey @FezVrasta , this original issue was related to a long deprecated version of slate. I think it’d make more sense to open something new with some fresh information. Happy to reopen it though if you feel strongly that the same conditions are true. Also, see this https://github.com/ianstormtaylor/slate/issues/3380#issuecomment-612432934, it might be related?

PR #2396, opened 2018-11-08
fix focusing across multiple editors

ianstormtaylor:

Bug.

#### What's the new behavior?

Fixes focus events and selection updating logic across multiple editors.

There was an issue in resolving two separate cases...

- When focusing for the first time, we need to call `editor.focus()` to ensure Slate's internal model knows it should be focused, otherwise it will blur itself in the DOM to keep its internal model "correct".

- When re-focusing an editor that used to have focus, Slate's internal model will still show it having a selection range, and it will try to update to that range immediately. Doing so would clobber the incoming `selectionchange` event, causing the editor to re-focus with its old selection instead. So we need to check if its a focus derived from mouse events, and deselect while focusing.

Fixes: #2352 
Fixes: #2390 
Fixes: #2038
Fixes: #2137
Fixes: #2210

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -146,103 +146,120 @@ class Content extends React.Component {
     const { isBackward } = selection
     const window = getWindow(this.element)
     const native = window.getSelection()
+    const { activeElement } = window.document
 
-    // .getSelection() can return null in some cases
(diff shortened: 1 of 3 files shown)


Saebyuckbaan:
I am still seeing issue mentioned in https://github.com/ianstormtaylor/slate/issues/2137

Issue #2351, opened 2018-10-30
add `node.getAllAncestorsAtRange` method


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Feature.

#### What's the current behavior?

Right now you have to manually iterate the `blocks` and look up each of their ancestors, and then make sure you convert it to a set to remove duplicates.

#### What's the expected behavior?

It would be nice to have a `getAllAncestorsAtRange` method that returned all of the ancestors in the range exactly once, so you can more easily find wrapping parents of certain types.

zhujinxuan:
Hi, may I know how this is different with
```
this.getAncestors(path).filter?
```
?

ianstormtaylor:
@zhujinxuan good question, actually the name I wrote in the issue is confusing. I think it would be better to have:

```js
const ancestors = document.getAllAncestorsAtRange(range)
```

Which would return an de-duped list of ancestors of all of the text nodes in a range. That way you could then do your own `.filter` on them.

```js
const quotes = document.getAllAncestorsAtRange(range).filter(a => a.type === 'quote')
```

ianstormtaylor:
@Dundercover very good questions, I don't have solid answers. It probably means we should re-evaluate all of those types of methods at once to stay consistent. Some of them are poorly named now too.

Right now we have `getBlocksAtRange` which returns the leaf-most blocks, and causes confusion for people since this is not obvious. I think that's why I tended towards `All*`, but it's probably better for us to qualify the other confusing methods names instead. So for blocks we might end up with:

```js
getBlocksAtRange() // returns *all* of the blocks
getLeafBlocksAtRange() // returns the leaf-most blocks in the range
getRootBlocksAtRange() // returns the root-most blocks in the range
```

I was thinking of `Leaf` and `Root` as qualifiers, but `Closest` and `Furthest` could work better, they're just a bit longer and more likely to misspell. Then again, `Leaf` is already a concept we have, so maybe we stick with `Closest/Furthest`. All of these would apply to inlines too.

```js
getNodesAtRange()

getTextsAtRange()

getBlocksAtRange() // changed to return *all*
getClosestBlocksAtRange()
getFurthestBlocksAtRange()

getInlinesAtRange() // changed to return *all*
getClosestInlinesAtRange()
getFurthestInlinesAtRange()
```

[...]

PR #2398, opened 2018-11-08
Add `node.getNodesAtRange`

Dundercover:

Feature

#### What's the new behavior?

Adds method `node.getNodesAtRange` which returns all text nodes inside the range plus all ancestors of those text nodes up to (but not including) the calling node.

Added a method which does a depth-first stack-based search for all relevant nodes using paths.

Fixes: #2351 

#### Thoughts / discussion

- **Big caveat**: I'm not sure if paths are always based on the document root or if they are based relative to the callling node. I went with the calling node so let me know if I'm mistaken.
- I went for a custom and more performant implementation instead of using existing methods. This might make it more difficult to maintain so let me know if you want something else.
- I only made one method `getNodesAtRange` which returns a [`List`](https://facebook.github.io/immutable-js/docs/#/List) which is memoized, is that ok?

### DIFF
--- a/docs/reference/slate/node.md
+++ b/docs/reference/slate/node.md
@@ -253,6 +253,12 @@ Get the next [`Text`](./text.md) node after a descendant by `path` or `key`.
 
 Get a node in the tree by `path` or `key`.
 
+### `getNodesAtRange`
+
+`getNodesAtRange(range: Range) => List`
+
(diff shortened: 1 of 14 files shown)


ianstormtaylor:
Absolutely beautiful PR @Dundercover, thank you! Those caveats are exactly right.

Issue #2103, opened 2018-08-21
add `Node.isNodeInRange` helper method


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Feature.

#### What's the current behavior?

Right now it's harder than it should be to answer the question of if a given node is in a given range. I think having this method would help.

#### What's the expected behavior?

```js
node.isNodeInRange(key, range) => Boolean
node.isNodeInRange(path, range) => Boolean
```

zhujinxuan:
Shall it be `isNodeInRange` or `isDescendantInRange`?

ianstormtaylor:
I think `isNodeInRange` would be most flexible, using `getNode/getPath`. We have the `*Descendant` and `*Child` alternatives for access, but these are more edge cases I think.

PR #2401, opened 2018-11-08
Add `node.isNodeInRange`

Dundercover:

Adding a feature

#### What's the new behavior?

Adds `node.isNodeInRange` method which can be used to check whether a descendant node is inside a range. This will return true for all text nodes inside the range and all ancestors of those text nodes up to this node.

Using `PathUtils.compare` method against the start and end paths of the range.

Fixes: #2103


ianstormtaylor:
@Dundercover this is awesome, thank you! And thank you very much for the great tests too.

Issue #2397, opened 2018-11-08
Placeholader was accidentally added


yunwuxin:
https://github.com/ianstormtaylor/slate/blob/f2f97e502e5fd648666960e823872d829be1ef99/packages/slate-react/src/plugins/react.js#L116-L124

```
{
  "document": {
    "nodes": [
      {
        "object": "block",
        "type": "paragraph",
        "nodes": [
          {
            "object": "inline",
            "type":"image",
            "data":{...}
            ]
          }
        ]
      }
    ]
  }
}
```

when `value` is like this ,  `placeholader` should not be added, but actually added.

ianstormtaylor:
@yunwuxin wouldn't that get normalized to have text nodes on either side of the inline?

yunwuxin:
@ianstormtaylor The place judged above is not

ianstormtaylor:
@yunwuxin ah! I see what you mean, it doesn't take that into account. Sounds like we might need to tweak the check to only happen when there is a single text node too? `node.getTexts().size === 1`?

yunwuxin:
It should be ok

PR #2403, opened 2018-11-09
Do not show placeholder when document contains multiple text nodes

Dundercover:

Fixing a bug

#### What's the new behavior?

Placeholder are not shown when document contains multiple text nodes.

We check for `node.getTexts().size === 1`.

Fixes: #2397


ianstormtaylor:
Beautiful, thank you @Dundercover!

Issue #2391, opened 2018-11-07
Merging blocks from inside Inline node deletes extra character


Malgalad:
Reproducible with [links example](https://www.slatejs.org/#/links).

**Steps**:
1. Place selection inside hyperlink, for example after "hype".
2. Press <kbd>Enter</kbd>.
3. Press <kbd>Backspace</kbd>.

**Expected**
- I get visually the same content as what I had at step one, that is hyperlink with "hyperlinks" text.

**Actual**
- Previous character is removed after merging blocks, e.g. "hyprlinks".

This does **not** happen if you move the selection outside of Inline block, for example by pressing <kbd>←</kbd> once after pressing <kbd>Enter</kbd>.

Edit:
![ice_video_20181107-145712](https://user-images.githubusercontent.com/1475852/48133043-2c417e80-e29e-11e8-9f0b-7c98b09ddbd8.gif)

Malgalad:
I can try my hand at fixing it, but I need help with starting point. I couldn't even find where are the default event handlers, if there is such concept.

CameronAckermanSEL:
I'd guess that this is a logical error in https://docs.slatejs.org/slate-core/commands#mergenodebykey-path

PR #2408, opened 2018-11-11
Fix deleteBackwardAtRange removing extra character when called at start of Inline

Malgalad:

Fixing a bug

#### What's the new behavior?
ice_video_20181111-021728

The bug was that `deleteBackwardAtRange` grabbed the previous Text node, but it was in the same Block as Inline and merging Blocks was not triggered correctly; then it tried to remove previous character and did so in the previous block.

* [ ] The new code matches the existing patterns and styles. _(not checking because who am I to judge?)_

Fixes: #2391
Reviewers: @ianstormtaylor


ianstormtaylor:
Good call, thanks @Malgalad!

PR #2410, opened 2018-11-11
[Fix] Copy-paste error "object.nodes is undefined"

jvdsande:
Fixing a bug

#### What's the new behavior?
I simply added a "undefined" check in "findInsertionNode", which seems to fix the bug where it was not possible to Paste content in an empty block.
Honestly I'm not sure if this is a fix or simply hiding the real problem. It does however fixes the behavior in my project: pasting now works as expected. Plus there was already part of the check in place so I guess this is OK.

Just checking if "object.nodes" is not undefined before getting "object.nodes.size"

No


ianstormtaylor:

Issue #2065, opened 2018-08-09
Slate data fragments from one editor should not be accepted by another


Slapbox:
Slate will continue to become more popular and the issue will become that Slate will accept encoded fragments from any source and attempt to merge them into Slate without checking whether they belong as Slate fragments or regular HTML in this particular editor.

This leads to unpredictable paste behavior when pasting data between Slate editors, and could potentially insert malicious data into a node's data structure as well. Primarily though, my concern is formatting. Malicious data seems a very low probability risk.

**I think the best solution for this is including a hash of the schema in the slate data fragment appended to HTML pastes. This would provide a reasonably simple way to ensure the pasted data comes from an editor with the same schema specifications.**

This would also provide `nonce`-like security against potentially malicious data being added to a block's data section. The security provided would be very weak and easily bypassed by someone targeting a particular implementation, but it would be better than trusting all fragments. 

All the same security could be provided via schema rules, but this leaves the onus on Slate users. Like I said though, I see this is a very low risk as is and I'm really more interested in avoiding formatting troubles.

ianstormtaylor:
Sounds good to me, good writeup @Slapbox!

ianstormtaylor:
Closing in favor of https://github.com/ianstormtaylor/slate/issues/1024

PR #2411, opened 2018-11-11
Only insert fragment of same schema on paste and drop

Dundercover:

Improving a feature

#### What's the new behavior?

When user pastes or drops something into an editor it will only insert an attached fragment if it has the same schema (hash code) as the target editor.

Computes a schema hash code of the schema rules inside the `SchemaPlugin`. `SchemaPlugin` then exposes this hash code with a query that `cloneFragment`, `onPaste`  and `onDrop` uses.

Fixes: #2065

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -352,7 +352,13 @@ function AfterPlugin(options = {}) {
       }
     }
 
-    if (type == 'fragment') {
+    const schemaHashCodeChecksOut =
+      fragment &&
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Hey @Dundercover, thank you so much for taking the time to work on this. I'm still very interested in this potentially with a hash of the editor itself now in `0.50`. But this PR is now out of date since so much has changed. I'm going to close it out, but if you are down to resubmit it that would be amazing! Thank you for understanding.

Issue #2407, opened 2018-11-10
Rich text sample seems to be broken


ufef:
I think rich text sample has got broken after recent changes. 
Try the following scenario:

https://www.slatejs.org/#/rich-text

Open the page, click outside of the editor area so focus gets lost, and then click again inside the editor.
It throws this exception:

`
TypeError: Cannot read property 'key' of undefined
    at a.n.renderBlockButton (https://www.slatejs.org/main-21d3cb05a6e56c129e06.js:42:22776)
...
`

The error is due to `value.blocks.first()` being null in this line of `renderBlockButton` method:
`
      const parent = value.document.getParent(value.blocks.first().key);
`

I am using Chrome 70.

DouglasLivingstone:
This is also broken on iOS, so not browser specific.

PR #2412, opened 2018-11-11
Fix `renderBlockButton` of rich text example

Dundercover:

Fixing a bug.

#### What's the new behavior?

When refocusing (click outside editor -> click inside editor) the example will not crash.

refocusingeditordoesnotbork

Checking if there are any blocks in the selection before doing stuff with the first block in the selection.

Fixes: #2407


ianstormtaylor:
Thanks @Dundercover!

PR #2415, opened 2018-11-12
Changing the selection when input by ime sometimes will break the ime.

yunwuxin:

_bug_

#### What's the new behavior?
Changing the selection when input by ime sometimes will break the ime.

Fixes: #2376

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -151,7 +151,7 @@ class Content extends React.Component {
 
     // COMPAT: In Firefox, there's a but where `getSelection` can return `null`.
     // https://bugzilla.mozilla.org/show_bug.cgi?id=827585 (2018/11/07)
-    if (!native) {
+    if (!native || editor.isComposing()) {
       return
     }
 
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -32,7 +32,6 @@ function BeforePlugin() {
   let isCopying = false
   let isDragging = false
 
-
   /**
    * The before plugin queries.
    *
--- a/packages/slate-react/src/plugins/before.js
+++ b/packages/slate-react/src/plugins/before.js
@@ -32,6 +32,17 @@ function BeforePlugin() {
   let isCopying = false
   let isDragging = false
 
+
+  /**
+   * The before plugin queries.
+   *
+   * @type {Object}
+   */
+
+  const queries = {
+    isComposing: () => isComposing,
+  }
+
   /**
    * On before input.
    *
@@ -480,6 +491,7 @@ function BeforePlugin() {
     onKeyDown,
     onPaste,
     onSelect,
+    queries,
   }
 }
 


packages/slate-react/src/plugins/before.js
```diff
@@ -32,6 +32,17 @@ function BeforePlugin() {
...
 
+
+  /**
+   * The before plugin queries.
+   *
+   * @type {Object}
+   */
+
+  const queries = {
+    isComposing: () => isComposing,
```
ianstormtaylor: This is genius @yunwuxin! I never thought about exposing this information via queries like this, cool idea.

packages/slate-react/src/components/content.js
```diff
@@ -151,7 +151,7 @@ class Content extends React.Component {
 
     // COMPAT: In Firefox, there's a but where `getSelection` can return `null`.
     // https://bugzilla.mozilla.org/show_bug.cgi?id=827585 (2018/11/07)
-    if (!native) {
+    if (!native || editor.isComposing()) {
```
ianstormtaylor: I think instead of not reflecting the update here, we might want to instead not handle `onSelect` in the first place? Such that the there are no `set_selection` operations that are queued from composition events?

yunwuxin: Because `componentDidUpdate' will also trigger `updateSelection`. When I input a word using the ime, and then enter the second word quickly, `componentDidUpdate` will trigger during the process of entering the second word, and then move the cursor, causing the ime to exit.

yunwuxin:
@ianstormtaylor any other problems?

packages/slate-react/src/components/content.js
ianstormtaylor: @yunwuxin but shouldn't when entering the second word it is back in an `isComposing` state?

ianstormtaylor:
@yunwuxin I think we still need to address that one comment. I'm still confusing about whether it makes senses to me that the selection itself will be updated to states that haven't been added to the `Document` yet? Or am I missing something there?

Can you provide a GIF of what the IME editing looks like, along with an explanation of when the selection/document is updated? What I don't want to do is end up with a situation where the selection is updated to a state that isn't reflected in the document itself?

(Imagine it from a collaborative editing point of view.)

yunwuxin:
aa
@ianstormtaylor  you can see this  

I I keep typing `g`-`e`-`space` , Sometimes the first letter  `g` loses.  Especially when the input is fast.

giddens9527:
Also bothered by this IME issue and thank you @yunwuxin for this solution! I tried patching this diff on my window 10 machine with Microsoft Chinese IME and it worked well.

zhengjunxin:
@yunwuxin thanks, I'm bothered by the same issue too

giddens9527:
@yunwuxin Just found one minor problem, when I tested this on the website, it worked perfect. However, when I patched the diff in my electron app, the editor.isComposing() always returned false. Not sure why but just wonder if you have encountered the same issue.

ejwill04:
Any updates on this PR?  I forked Slate and add this PR and #2454 and they both helped with selection issues with IME.

hecor:
Thanks @yunwuxin , help me a lot a lot a lot ...

PR #2416, opened 2018-11-13
update placeholder

cudr:

bug

#### What's the new behavior?

After (#2370) we can use only one placeholder text multiple times, because first placeholder plugin render self placeholder each time.

Now we pass placeholder to mark data and get it from that.

Fixes: #
Reviewers: @


valoricDe:
Has this been obsoleted?

Issue #1257, opened 2017-10-18
consider adding `defaultValue/State` prop to <Editor>


ianstormtaylor:
To behave more like an `<input>/<textarea>`.

ianstormtaylor:
Actually this would be nice in the examples, because most of them don't actually need to be using a controlled component, they could just use the `defaultValue` instead.

PR #2418, opened 2018-11-13
Add `defaultValue` prop to `Editor` component

Dundercover:

Adding a feature

#### What's the new behavior?

The `Editor` component now takes a `defaultValue` prop which makes it behave like an uncontrolled input component.

Adds a state inside the `Editor` component which keeps track of a `Value` if it is initially set using `defaultValue`. Examples have also been updated to use this prop.

Fixes: #1257


ianstormtaylor:
Thanks @Dundercover, this is awesome!

Issue #2111, opened 2018-08-22
change `.getClosest` to include the node itself


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

Right now the `.getClosest` method doesn't return the node itself if it matches the iterator. This is confusing because it's different than the behavior of the DOM's native [`el.closest`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) API.

#### What's the expected behavior?

It should return the node itself if it matches too.

Dundercover:
Are you saying that these methods also need to change?

- `getClosestBlock`
- `getClosestInline`
- `getClosestVoid`

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/3093.

PR #2436, opened 2018-11-16
Include current node in iterator of `node.getClosest`

Dundercover:

Improving a feature

#### What's the new behavior?

Node methods `getClosest`, `getClosestBlock`, `getClosestInline` and `getClosestVoid` will now return the current node if it is a match (so it matches the behaviour of the DOM's native [Element.closest()](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest).

Add the current node to the list of nodes that will be iterated over in `node.getClosest`.

Also added some tests for `node.getClosest`.

Fixes: #2111

### DIFF
--- a/packages/slate/test/models/node/get-closest/matching-current-using-path.js
+++ b/packages/slate/test/models/node/get-closest/matching-current-using-path.js
@@ -1,0 +1,37 @@
+/** @jsx h */
+
+import h from '../../../helpers/h'
+
+export const input = (
+  <value>
+    <document>
+      <quote key="a" {...{ className: 'slate-node' }}>
+        <paragraph key="b" {...{ className: 'slate-node' }}>
+          <text key="c">one</text>
+        </paragraph>
+      </quote>
+      <paragraph key="d" {...{ className: 'slate-node' }}>
+        <text key="e">two</text>
+      </paragraph>
+      <paragraph key="f" {...{ className: 'slate-node' }}>
+        <inline type="link" key="g" {...{ className: 'slate-node' }}>
+          <text key="h">three</text>
+        </inline>
+        <text key="i">four</text>
+      </paragraph>
+    </document>
+  </value>
+)
+
+export default function({ document }) {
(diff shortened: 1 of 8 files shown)


Dundercover:
By the way! Shouldn't `getFurthest` methods now also include current node?

packages/slate/test/models/node/get-closest/matching-current-using-path.js
```diff
@@ -0,0 +1,37 @@
+/** @jsx h */
+
+import h from '../../../helpers/h'
+
+export const input = (
+  <value>
+    <document>
+      <quote key="a" {...{ className: 'slate-node' }}>
```
ianstormtaylor: Why write these like this with the object spreading? Instead of just writing `className=`? Even better might be to use `thing="value"` or something more abstract just to avoid confusion with DOM stuff.

packages/slate/test/models/node/get-closest/matching-current-using-path.js
```diff
@@ -0,0 +1,37 @@
...
+        <paragraph key="b" {...{ className: 'slate-node' }}>
+          <text key="c">one</text>
+        </paragraph>
+      </quote>
+      <paragraph key="d" {...{ className: 'slate-node' }}>
```
ianstormtaylor: Also no need to include `key=` on all elements here. We only need it on the specific node we're targeting in the test itself. That way we reduce maintenance burden.

Dundercover: Sounds good!

Dundercover:
@ianstormtaylor I'm actually not aware of a `furthest` method in DOM land, do you know what you would use for that?

ianstormtaylor:
Ah, I didn't realize there wasn't one. I'd use the same logic—including ancestors and the node itself. And then just search top-down instead of bottom-up.

ianstormtaylor:

Issue #1536, opened 2018-01-16
slate-hyperscript can conflict with common babel configs (create-react-app)


jasonphillips:
This doesn't necessarily classify as a bug, since it depends upon interaction with babel configurations, but it happens on the default `create-react-app` configuration, which means it potentially affects quite a few users who are setting up small apps.

If your babel configuration includes certain widely used babel transforms that add special metadata to JSX function calls for debugging purposes (the main offender is `babel-plugin-transform-react-jsx-self`, which adds `__self` to each tag's attributes, and which is **turned on by default in create-react-app**), the slate hyperscript integration will preserve these elements in the generated Value / Document objects -- which is invisible and inconsequential at first, yet has the unfortunate consequence of rendering the resulting Value object (and all Value objects derived from it via operations) incapable of JSON serialization, throwing a "circular JSON" error even when correctly invoking the .toJSON() or .toJS() functions prior to serialization.

Once I finally discovered this root problem that was making my editor's state unserializable, I implemented a simple workaround for my app which averts the issue by pointing the pragma to a wrapper function:

```js
/** @jsx stripMeta */

import { createHyperscript } from 'slate-hyperscript'

const h = createHyperscript({
  blocks: {
    p: 'paragraph',
  },
  marks: {
    b: 'bold',
    i: 'italic',
  },
});

// delete potentially circular attributes added by babel, prior to passing to h()
const stripMeta = (tagName, attributes, ...children) => {
  delete attributes.__self;
  delete attributes.__source;
  return h(tagName, attributes, ...children);
}

const value = (
  <value>
    <document>
         /* etc */
```

A workaround like that one is feasible and I hope others are able to find it here if they start running into circular JSON errors that are way downstream from this problem with the starting content object. 

One way this could be averted for everyone, however, is by simply excluding all double-underscore attributes from slate-hyperscript by default, which seems relatively safe since it is doubtful someone would use those for meaningful attributes -- is that a change that could be agreed on, or is the workaround (or manually reconfiguring babel, where possible) preferable?

alanctkc:
I'm actually running into this myself as well. I use `slate-hyperscript` for initial editor values, so I see junk data in _user_'s data that exposes local source code filesystem paths. Pretty ugly and unexpected. It also causes failing test assertions when I match on entire `slate-hyperscript` values.

I'm able to mitigate it in my own app using @jasonphillips's (clever) workaround above, but overall, it's a rough default for CRA apps.

[...]

PR #2437, opened 2018-11-16
Filter out double-underscore keys (babel junk) from hyperscipt values

alanctkc:

Fixing a quasi-bug (unexpected behavior), #1536.

#### What's the new behavior?

All hyperscript attributes prefixed with double underscores (`__foo`) will be filtered out by convention. This gets rid of the unexpected `__source` and `__self` attributes added by create-react-app's default babel configuration.

I basically copied @jasonphillips's workaround from #1536. Iterate all the attribute keys and delete the prefixed ones from `attributes`.

I do have a small concern that mutating `attributes` like this will have some adverse effects, maybe with something babel-related. However, when I tried copying/filtering keys into a new object, other things were somehow going awry. `attributes` must be a more rich object than I realized.

**^ All examples were throwing an error in local for me both before and after the change. Will try rebasing against latest to see if I can get things clean.**

Fixes: #1536
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-hyperscript/src/index.js
+++ b/packages/slate-hyperscript/src/index.js
@@ -77,6 +77,13 @@ function createHyperscript(options = {}) {
       .filter(child => Boolean(child))
       .reduce((memo, child) => memo.concat(child), [])
 
+    // by convention, strip attributes prefixed with double underscores
+    for (const key in attributes) {
+      if (key.lastIndexOf('__', 0) === 0) {
+        delete attributes[key]
+      }
+    }
+
     const ret = creator(tagName, attributes, children)
     return ret
   }


alanctkc:
Just rebased against latest master locally, and all examples are good!

packages/slate-hyperscript/src/index.js
```diff
@@ -77,6 +77,13 @@ function createHyperscript(options = {}) {
       .filter(child => Boolean(child))
       .reduce((memo, child) => memo.concat(child), [])
 
+    // by convention, strip attributes prefixed with double underscores
+    for (const key in attributes) {
+      if (key.lastIndexOf('__', 0) === 0) {
```
ericedem: This might be a dumb question, but could we just use `key.startsWith('__')` instead here for clarity? I think the only difference is that IE doesn't support it but I think it would get transpiled on build.

ericedem: Oh hmm, I guess it would be slightly different in case something has more underscores, or ends with underscores.

alanctkc: Actually, I just wasn't sure what polyfills were in there or what browsers Slate targets, but if we're good to use `startsWith`, I prefer it too! It should work just fine here (that thing is ugly).

ericedem: The [browserlist has IE11](https://github.com/ianstormtaylor/slate/blob/master/browserslist#L5). So I think babel will handle it properly? 🤷‍♂️

alanctkc: I'm not _totally_ sure it's not in the repo, but searching around now I'm not seeing babel-polyfill itself required, which I think is how babel does all polyfilling to keep it out of core. Pretty beefy to bring the whole thing in. It'd be possible probably to just bring the `startsWith` polyfill from core-js, but then it's still overkill for one call. :)

@ianstormtaylor Does Slate have any polyfills or existing pattern for it?

ianstormtaylor: @alanctkc Slate uses stuff that's fine for newest browsers and expects people to polyfill things like `startsWith` or `closest` I think.

ianstormtaylor:
Hey @alanctkc, thank you so much for taking the time to work on this. I'm still down for this. But this PR is now out of date since so much has changed. I'm going to close it out, but if you are down to resubmit it that would be amazing! Thank you for understanding.

Issue #2444, opened 2018-11-20
"versions" example is broken on Chrome


fralonra:
#### Do you want to request a _feature_ or report a _bug_?
Report a bug

#### What's the current behavior?
The example of [`versions`](https://www.slatejs.org/#/versions) is broken.
It threw the following error on **Chrome 67.0**, but not on Firefox 62:
```
An error was thrown by one of the example's React components!

TypeError: e.toArray(...).reverse(...).map(...).flat is not a function
    at a.n.saveVersion (https://www.slatejs.org/main-9a612105532ed99867a4.js:34:53507)
    at a.value (https://www.slatejs.org/main-9a612105532ed99867a4.js:34:53727)
    at Yo (https://www.slatejs.org/main-9a612105532ed99867a4.js:22:84444)
    at Zo (https://www.slatejs.org/main-9a612105532ed99867a4.js:22:80997)
    at Vo (https://www.slatejs.org/main-9a612105532ed99867a4.js:22:80496)
    at Wo (https://www.slatejs.org/main-9a612105532ed99867a4.js:22:80322)
    at ts (https://www.slatejs.org/main-9a612105532ed99867a4.js:22:85921)
    at Tt (https://www.slatejs.org/main-9a612105532ed99867a4.js:22:31232)

    in a
    in a
    in a
    in div
    in Styled(div)
    in div
    in a
    in a
    in a
    in a
```
The related code was added in https://github.com/ianstormtaylor/slate/pull/2379.

#### What's the expected behavior?
To show the example correctly.

Dundercover:
Ah, I missed that [`Array.prototype.flat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) is an [experimental technology](https://developer.mozilla.org/en-US/docs/MDN/Contribute/Guidelines/Conventions_definitions#Experimental) (see [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#Browser_compatibility)), I'll jump on this next.

Thanks for reporting it!

PR #2446, opened 2018-11-20
Fix: Do not use experimental array method `flat` in `Versions` example

Dundercover:

Fixing a bug.

#### What's the new behavior?

The `Versions` example will not break on (versions of) browsers that does not support the experimental array method `flat`.

Since we are flattening a nested array that are exactly two levels, we can just use `reduce` instead.

Fixes: #2444

### DIFF
--- a/examples/versions/index.js
+++ b/examples/versions/index.js
@@ -129,24 +129,13 @@ class Versions extends React.Component {
   setVersion = n => {
     const { editor } = this
     const { versions, v } = this.state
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @Dundercover!

PrimeObjects:
slate-hyperscript still use many flat. It crashed IE

Dundercover:
@PrimeObjects Please submit a new issue for that problem

PR #2448, opened 2018-11-21
Improve/fix index property of child_min_invalid validation error

aiwenar:

Unsure, but seems more like a bug to me (would've been part of #2388 had I noticed it before submitting that PR).

#### What's the new behavior?

My intent in #2388 for the `index` property of `child_min_invalid` was that it pointed at the last matching node. Unfortunately there were two problems with implementation:

1. when `min` was violated in the last group, `index` would point one past the last matching node
2. when the group was empty, `index` would point at the previous index. This, when the first group required exactly one element (such as the _Forced Layout_ example), made it impossible to distinguish violation of `min` for the first and second group, as both would report `index` as 0.

This PR changes `index` to always point at the last matching node, unless there are none (`count == 0`), in which case it will point at where the first matching node would be had it existed. It also updates documentation to include description for properties of `child_min_invalid` and `child_max_invalid`.

- Adjust `index` where relevant
- Add tests for this and other properties of `child_min_invalid` and `child_max_invalid` (as I should have in the original PR)

Fixes:
Reviewers:

### DIFF
--- a/docs/reference/slate/schema.md
+++ b/docs/reference/slate/schema.md
@@ -317,10 +317,15 @@ Raised when the `object` property of a child node is invalid.
 
 Raised when a child node repeats less than required by a rule's `min` property.
 
+* `index`: index of the last matching node, or where the first matching would be, if there are no matching nodes
+* `count`: number of matching nodes
+* `limit`: value of rule's `min` property
+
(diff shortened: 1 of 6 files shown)


ianstormtaylor:

Issue #2419, opened 2018-11-14
Hide main menu for the examples site on mobile


Dundercover:
#### Do you want to request a _feature_ or report a _bug_?

Improvement (suggestion)

#### What's the current behavior?

When browsing the examples site on my phone, the menu with all the examples takes up my entire screen. It hides the main focus of the site (the editor) and does not look so clean imho.

#### What's the expected behavior?

Maybe put the main menu to the left and make it togglable on smaller screens (hamburger menu).

PR #2450, opened 2018-11-21
feat(Examples): improve example navigation

ericedem:

_feature_

#### What's the new behavior?

For small screens, don't show tabs by default. Instead, show a hamburger menu. When clicking the hamburger menu, expand the examples list.

I arbitrarily picked 600px threshold which seemed reasonable to me, open to other suggestions!

kapture 2018-11-21 at 14 36 40

Fixes: #2419
Reviewers: @ianstormtaylor @Dundercover


Dundercover:
Thank you @ericedem! I can try it on a couple of different browsers and screens if you want some QA on it. 

My very subjective opinions:

- I think it would be better with a typical left menu with one item per row instead of a second top bar with a centered menu. It's difficult to get an overview when you have multiple rows with different number of items in it.
- I think it would be cleaner with the hamburger inside the top bar. It takes up a lot of space for just one button.

ericedem:
> I think it would be better with a typical left menu with one item per row instead of a second top bar with a centered menu. It's difficult to get an overview when you have multiple rows with different number of items in it.

Hmm that's a good point. The reason I had decided to just leave it with multiple items per row was because the menu was going to overflow way below the content (the editor) of the page. I agree it isn't ideal this way either. 🤔 

> I think it would be cleaner with the hamburger inside the top bar. It takes up a lot of space for just one button.

Yea it is kind of in an awkward position. It didn't feel like it belonged in the top bar since that is nav links. Thinking about it again, it could probably live next to the "Slate Examples" logo. Maybe as a carrot instead of a hamburger.

ericedem:
Another quick iteration on improving the navigation based on your suggestions @Dundercover 

kapture 2018-11-26 at 8 40 23

Playing with it I think I like it a lot better. The one thing I was trying to decide was whether to try and close the menu when the user clicks a nav item. It would be convenient, but might be a pretty confusing / unexpected first user experience.

Dundercover:
Thank you @ericedem! it looks much cleaner now imho :) I do think we should close the menu when clicked since it's more convenient and an established kind of behaviour (GitHub does it for example).

Minor tweaks if you're game:

- I think it would be good if the top bar is static, it's convenient to always have it (the buttons) accessible and that it's not affected by whether or not a scrollbar is present.
- It might be nice to gray out the body (editor) when the left menu is visible on smaller screens so it does not confuse users that it is something you could interact with.

Great work!

ericedem:
Ah good points! Yea I think those would be good improvements.

I wonder if it might be better for desktop to also have the same collapsable nav menu instead of only for mobile. I think this new mobile menu is a bit better than what is there currently.

ericedem:
Ok I added those changes with the exception of static nav bar. The problem is that the examples list is so long that it ends up going off the page itself. We would need the menu to float with the nav bar, which means we need a separate scroll area for the menu, which starts getting pretty wonky. Especially on mobile devices.

One other thing I noticed is that you can't see the name of the example anymore without opening the menu to check. I feel like we should probably put it somewhere on the page. I don't think we can put it to the right of "slate examples" because on small screens the nav bar content will be too wide. I'm thinking it might be best to move the hamburger and the example name to the next line.

ericedem:
Here is what the underlay looks like: 

image

ericedem:
Latest iteration:

kapture 2018-11-26 at 19 06 33

ianstormtaylor:
Very cool @ericedem!

Only change is could you make the hamburger menu next to the title of the example, instead of in the toolbar above it? That way I think it makes more sense for people to find where to change the current example.

Also, could you add a little padding above the white editor box again?

Thank you!

ericedem:
> Also, could you add a little padding above the white editor box again?

Good catch! I thought something looked a little off but couldn't put my finger on it 😄 

I made those changes, it's looking pretty good now. Latest:

kapture 2018-12-14 at 8 08 33

thesunny:
Okay, I've tested this and it looks pretty good.

I'm going to merge this soon unless there are any objections.

thesunny:
Merged and thanks @ericedem

ianstormtaylor:
Thank you @ericedem and @thesunny for merging!

PR #2454, opened 2018-11-26
fix(IME): cursor is placed at start of uncommitted string while compo…

akumatus:

fixing a bug

#### What's the new behavior?

The cursor is placed at the end of the composition string.

Ignore updating the native selection from `editor.value.selection` while composing. Because `onSelect` returns early in before plugin, without updating editor selection from native.

Fixes: #
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -69,6 +69,7 @@ class Content extends React.Component {
 
   tmp = {
     isUpdatingSelection: false,
+    isComposing: false,
   }
 
   /**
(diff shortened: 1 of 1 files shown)


SmilinBrian:
Bravo! I have been using a very similar fix on our "several versions back private fork" with success and haven't had time allocated to try to bring it up to date with current version for a PR.

freeplant:
👍 This PR should fix the problem.

yunwuxin:
It's not completely solved. You can see this.

#2415

ejwill04:
Any updates on this PR? I forked Slate and add this PR and #2415 and they both helped with selection issues with IME.

SmilinBrian:
> Is it possible to make this tweak a plugin, and use the plugin to solve the problem?

I don't think so, it's too "deep" in the event handler.

PR #2455, opened 2018-11-26
fix(IME): can not input when use IME in Safari

akumatus:

fixing a bug

#### What's the new behavior?

Can input when use IME in Safari.

In input events level 2, when a part or the entire composition string is to be committed to the DOM, a beforeinput event with the inputType `insertFromComposition` is dispatched with the data attribute set to the part of the final composition string that is to be committed to the DOM.

Thus I suppose `insertFromComposition` should be handled in after plugin.

Fixes: #
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -115,6 +115,7 @@ function AfterPlugin(options = {}) {
       }
 
       case 'insertFromYank':
+      case 'insertFromComposition':
       case 'insertReplacementText':
       case 'insertText': {
         // COMPAT: `data` should have the text for the `insertText` input type


FnTm:
Hi @akumatus. I checkout out your branch, ran the examples, and it just doesn't seem to work like it should. 
I can now get the accented chars to input _sometimes_, but not predictably, and nowhere near always. + I can get some funk behaviour, like this:
- Hit apostrophee 5 times
- Enter a single letter a
- All five apostrophes disappear, and _maybe_ an `ā` appears (for Latvian language that is)

akumatus:
Hi @FnTm. I suppose there are two problem in Safari when using IME tool to input, another one is [cursor](https://github.com/ianstormtaylor/slate/pull/2454). Try to merge that pr also then find out if something is different. By the way, I test those pr with Chinese IME tool, may be there are differences between different languages?

sethyuan:
I can confirm that it fixes Chinese IME issue with Safari. Without this PR, you simply cannot type Chinese in iPad, which makes Slate useless for Chinese users.

FnTm:
@akumatus curious - why are you closing this issue? Is it solved in master?

akumatus:
Hi @FnTm I suppose this problem haven't resolved yet. I close this pr because it didn't solve the problem completely. I encounter some Safari cursor problem in collaborative edit mode also.

PR #2462, opened 2018-11-30
undo and redo operations are now wrapped with withoutNormalizing call…

CameronAckermanSEL:
… to prevent normalizing occuring in between application or reversal of batched operations

Bug

#### What's the new behavior?

The body of the undo and redo history commands are now wrapped in a withoutNormalizing call which prevents normalization from happening until the entirety of the operation batch has been applied.

Fixes: #
Reviewers: @ianstormtaylor


ianstormtaylor:
Thanks @CameronAckermanSEL!

PR #2469, opened 2018-12-03
fix InputEvent global check

rexxars:

Fixing a bug

#### What's the new behavior?

When importing slate in a JSDom environment, it won't crash

We are importing some code from node.js that we run through JSDOM. It in turn imports slate, which thinks it's a browser (this is fine, since that's the reason for using JSDOM in the first place). There is a check in the  `slate-dev-environment` module that _checks_ for `window.InputEvent`, but tries to instantiate an event from the *global* `InputEvent`, which in this case is the node environment, not the browser.

This is a simple fix that prevents the crash from happening.

Reviewers: @skogsmaskin

### DIFF
--- a/packages/slate-dev-environment/src/index.js
+++ b/packages/slate-dev-environment/src/index.js
@@ -65,7 +65,7 @@ const FEATURE_RULES = [
   [
     'inputeventslevel1',
     window => {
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Good catch, thanks @rexxars!

PR #2471, opened 2018-12-03
Use weakmap to prevent memory leak and remove indirect __cache__key

zhujinxuan:

Fix potential bug and Refactor code

#### What's the new behavior?

Instead of this.__cache and __cache_key; the cache will be stored in WeakMap.  

Then we can:
1. We have some one-time range in memorized `getXXXAtRange(range)`. After this PR, the memoized one-time range results will be removed.
2. clear cache will be localized in one place, rather than `__cache_key` logic scattering everywhere.
3. clean the object stucture.  We will not see `this.__cache` in logging out the object.  Perhaps it makes moving to immer with cache easier.
4. Use symbol instead of object for some readability.

For normal arguments, the cache tree will visit directly as a property of dictionary.  If an object is passed, it will visit a WeakMap store to ensure no memory leak.

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate/src/utils/memoize.js
+++ b/packages/slate/src/utils/memoize.js
@@ -15,7 +15,7 @@ let ENABLED = true
  * @type {Object}
  */
 
-const LEAF = {}
+const LEAF = Symbol('LEAF')
 
 /**
  * A value to represent a memoized undefined value. Allows efficient value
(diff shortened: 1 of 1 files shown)


zhujinxuan:
It is ready for review now.

ianstormtaylor:
Very cool, thank you @zhujinxuan! This is awesome.

Issue #2475, opened 2018-12-03
Multiple Nodes Deleted when no text and only void inlines in block.


ldavidpace:
#### Do you want to request a _feature_ or report a _bug_?
_bug_

#### What's the current behavior?
When multiple void inlines are placed in the same block with no text. Pressing backspace on the line while the inlines are not selected deletes all of the inlines in the block.

[`jsfiddle`](https://jsfiddle.net/paceld/j8weLoym/)

![inlinedeletebug](https://user-images.githubusercontent.com/25068382/49406902-18116580-f714-11e8-95fd-e50b757de417.gif)

#### What's the expected behavior?

Only the inline next to the cursor should be deleted. If at the start of the line the inlines should be merged into the line before.

PR #2476, opened 2018-12-04
Checking to make sure there is only text in the node before removing the whole node on deleteBackward

ldavidpace:
Fixing a Bug

#### What's the new behavior?

Old Behavior 
inlinedeletebug2

New Behavior
inlinedeletebugfixed

  Checking to see that there is only one node in the block before using the PERF case of removing the whole node. to see if there are other nodes besides text.

   I am getting a lot of unable to resolve path to module 'slate'. No other errors. Not sure what I did to invalidate the linter setup.

Fixes: #2475


ianstormtaylor:
Thank you @ldavidpace, great pull request!

JinJieTan:
But I still run into this problem and can't remove it when I insert only one inline

please help me ! 

#: https://github.com/ianstormtaylor/slate/issues/2962

PR #2482, opened 2018-12-04
Adding Guilded to Products list

rfeltis:
We just added tons of features to our text editor by switching to Slate and would love to rep it.


ianstormtaylor:
Thanks @rfeltis!

Issue #2456, opened 2018-11-26
Undo stack implementation retains unreachable `Value` objects, leaks memory


bengotow:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?

Open the https://www.slatejs.org/#/history slate example, open the Developer Tools and start a Memory > Allocation Instrumentation on Timeline profile. Then perform the following:

1. Type 4-5 characters
2. Undo the typed characters so the undo stack length is 0
3. Type a character to clear the redo stack
4. Type another character to remove any remnants of previous props from React's memoziedProps.

Observe that memory allocated when the original 4-5 characters were typed is still around, even though it is now impossible to reach those operations via undo/redo:

<img width="707" alt="screen shot 2018-11-26 at 1 06 18 pm" src="https://user-images.githubusercontent.com/1037212/49042304-a1f18980-f17c-11e8-8cff-9ca0be3593ab.png">

* Note: I've verified that disabling undo/redo entirely and it also fixes these leaked allocations, they're definitely undo/redo related.

#### What's the expected behavior?

Slate should not retain memory related to undo/redo operations which can no longer be reached via undo/redo.

#### Fix

I think this is caused by the fact that the undo redo stack is stored in `value.data`. Since each undo operation has a copy of `value`, and `value` contains an undo stack, Slate creates a recursive structure of Value's with undo stacks that reference older values, and keeping a handle to a single undo/redo entry is enough to retain the entire stack forever. Clipping to 100 undo entries is nice, but since entry 0's `operation.value.data.get('undos')` retains 100 more entries further into the history, it's not effective at limiting the memory footprint.

After applyOperation applies an undo or redo, the code restores the current undo/redo stack which means that it actually doesn't need the undo/redo stacks of each saved `Value` at all. I patched `Commands.save` to remove these and it resolves the issue:

```
Commands.save = (editor, operation) => {
  const { operations, value } = editor

  // Remove the undo/redo history from the operation's Value
  operation = operation.withMutations(op => {
    const v2 = op.value.withMutations(v => {
      let d2 = op.value.data
      d2 = d2.remove('undos')
      d2 = d2.remove('redos')
      v.set('data', d2)
    })
    op.set('value', v2)
  })
  
  ...
```

[...]

PR #2488, opened 2018-12-08
Fix memory leak in undo stack implementation (#2456)

bengotow:

Fixing a bug

#### What's the new behavior?

This change strips Operation's values of their undo/redo stacks before placing them on the undo/redo stack. This avoids recursively retaining previous undo stacks containing previous values which are more than 100 elements deep. This had significant memory use implications documented in #2456.

Discussed in #2456!

Fixes: #2456
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate/src/commands/on-history.js
+++ b/packages/slate/src/commands/on-history.js
@@ -22,6 +22,17 @@ Commands.save = (editor, operation) => {
   let { save, merge } = editor.tmp
   if (save === false) return
 
+  // Remove the undo/redo history from the operation's Value. This prevents
+  // us from recursively retaining the entire history of all Value objects.
+  operation = operation.withMutations(op => {
+    op.set(
+      'value',
+      op.value.withMutations(v => {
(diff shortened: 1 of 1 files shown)


zhujinxuan:
A small question: why not 
```
   operation = operation.withMutations(op => { 
    op.deleteIn(['value', 'data', 'undos'];
    op.deleteIn(['value', 'data', 'redos'];
})
```

zhujinxuan:
I am feeling perhaps it is better to use:
```
operation = operation.deleteIn([value, data, undos]).deleteIn([value, data, redos])
```

I think it is not a frequently visited code, so withMutation optimization is perhaps not so necessary.

bengotow:
Hey folks! I'm upgrading Mailspring to Slate 0.45.x (hey, I'll catch up to latest eventually) and I'm happy to report that the need for this is totally resolved by https://github.com/ianstormtaylor/slate/pull/2225. Going to close this PR. Thanks to @bryanph for his work on that PR!

PR #2491, opened 2018-12-08
Update Placeholder to be compatible with older versions of React

davegw:
`React.Fragment` is only available in react >= 16.2, which does not meet the dependency requirements specified by the package (react >= 0.14.0), so unfortunately needs to be replaced in order to maintain compatibility with older versions of react.

Updates from `React.Fragment` -> `<span>` to provide coverage for older versions of react.

Fixing a dependency bug

#### What's the new behavior?
`Placeholder` component that is compatible with older versions of react

Replaces `React.Fragment` with `<span>` wrapper because `React.Fragment` is not available in react versions before 16.2.

Fixes: #
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react-placeholder/src/index.js
+++ b/packages/slate-react-placeholder/src/index.js
@@ -70,12 +70,12 @@ function SlateReactPlaceholder(options = {}) {
       }
 
       return (
-        <React.Fragment>
+        <span>
           <span contentEditable={false} style={style}>
             {placeholder}
           </span>
           {children}
-        </React.Fragment>
+        </span>
       )
     }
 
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
I'm not sure if it has adverse effects on `contenteditable` behaviors. I guess it probably won't. Although part of me thinks we should just have `slate-react-placeholder` take React 16+ instead?

davegw:
Yeah, I wasn't entirely sure about the effects on `contenteditable` either, but thought it would be okay. 

IMO, it seems like a shame to require React 16+ for `slate-react-placeholder` (and consequently `slate-react`) just for the one use case here, but it's your call. If there were more React 16 features being used that made it difficult to support < 16, I think it would be a stronger argument for bumping the dependency.

ianstormtaylor:
@davegw very fair, I'll merge and we can decide to limit to React 16+ later if we find the need. Thank you!

PR #2493, opened 2018-12-10
fix(schema): previous and child variable may be exactly the same

akumatus:

 fixing a _bug_

#### What's the new behavior?

In `validateNodes` function, the `index` variable may decrease by code `index -= 1` in some specific occasion.

Thus the old code `previous = child` and `child = children.get(index)` will make child and previous variable exactly the same, and eventually lead to error of `previous_sibling_type_invalid`.

The new way can make sure the `previous` variable is the real previous node of child.

Fixes: #
Reviewers: @ianstormtaylor


ianstormtaylor:
Great call, thank you @akumatus!

Issue #2364, opened 2018-11-01
IE11:  Doesn't allow cursor to focus or edit text on editor input 


lranches-sfdo:
#### Do you want to request a _feature_ or report a _bug_?

Bug 🐛 

#### What's the current behavior?'
![ie-rteditor](https://user-images.githubusercontent.com/43382367/47881688-176b7380-dde4-11e8-8fc3-bf72f6d53cb6.gif)

Using Internet Explorer 11, if the editor input already contains text. It won't allow on click to focus on the Editor Input and edit. Also upgraded to the latest Slate versions and still has the same issue but posted the versions being used below. It also is working perfectly fine in Chrome.

Any help would be much appreciated! 

```    
    "slate": "^0.33.4",
    "slate-react": "^0.12.4",
```

#### What's the expected behavior?

Working on Chrome!

![chrome-rteditor](https://user-images.githubusercontent.com/43382367/47881702-1f2b1800-dde4-11e8-8d86-36f2dc0def76.gif)

isubasti:
@lranches-sfdo 	do you have polyfill for IE, I think it might be due to IE 11 doesn't have `Element.closest` and slate relies on that and it's not included hence you need to include your own.

lranches-sfdo:
@isubasti Thank you for your response. We actually have the closest polyfill for IE and still no luck :(. Any other ideas?

lranches-sfdo:
Any updates? :)

dannyiacono:
The same issue applies to your provided examples, for instance with your link demo: https://www.slatejs.org/#/links

afrankel-sfdo:
I can't even get the site to load in IE11 on Windows 10 or 8.1...

![screen shot 2018-11-21 at 12 02 00 pm](https://user-images.githubusercontent.com/31106469/48863110-81dd5700-ed85-11e8-9973-21261e997332.png)

ericedem:
@afrankel-sfdo Tracked here: https://github.com/ianstormtaylor/slate/issues/2385.

afrankel-sfdo:
@eramdam thanks, that was just a sidebar while attempting to reproduce the editor issue in IE11 above.

isubasti:
[...]

PR #2496, opened 2018-12-12
Use `babel-polyfill` in place of `polyfill.io` CDN

### LINKED ISSUE #2385: IE11 slatejs.org breaks due to `includes()`

_bug_

Looks like a use of includes in [`slate-dev-environment`](https://github.com/ianstormtaylor/slate/blob/fa343c07f4d9bb8840c5a85ecd8efc617dbabde1/packages/slate-dev-environment/src/index.js#L112) is not getting transpiled for IE11. And is blowing up on https://www.slatejs.org/.

ie11_includes_broken

It looks like this module is missing in transpilation. IE11 is in the browserlist, so maybe some other configuration is interfering?

davegw:
Update provides support for running the examples app on IE 11.

Fixes #2385 and makes investigation of #2364 easier.

There appeared to be a number of issues with using `polyfill.io` with slate. While trying to run the example app on IE 11, I ran into each of the following:
https://github.com/Financial-Times/polyfill-service/issues/792
https://github.com/Financial-Times/polyfill-service/issues/854
https://github.com/Financial-Times/polyfill-service/issues/1341

`babel-polyfill` provides more robust support for polyfilling and will allow for better targeting of supported browser envs in the future (if desired).

Also includes `element-closest` polyfill, which is necessary for using with IE 11 😞 

Bug in examples app

#### What's the new behavior?
Allows the examples app to be loaded on IE 11

Uses required polyfills to support loading the examples app in IE 11

Fixes: #2385 
Reviewers: @ianstormtaylor


ianstormtaylor:
Sounds good to me, thank you for investigating @davegw!

PR #2506, opened 2018-12-21
Add `isInverted` flag on inverted operations

Calyhre:

Adding a feature

#### What's the new behavior?

This PR adds a `isInverted` flags on operations' record.

Every operations has a `isInverted` defaulted to `false`, and set to `true` when operation has been inverted.

### DIFF
--- a/docs/reference/slate/operation.md
+++ b/docs/reference/slate/operation.md
@@ -10,7 +10,7 @@ There are a handful of Slate operation types. The goal is to have the fewest pos
 
 See each operation separately below to see what they consist of.
 
-Note that all operations have a `data` property which can be used to attach arbitrary data to the operation, just like the [`Node`](./node.md) models ([`Block`](./block.md), [`Inline`](./inline.md), etc).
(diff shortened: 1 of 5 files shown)


Calyhre:
:up: While some operation can be inverted quite easily, having this flag helps a lot to understand user intention when processing undone operations

Calyhre:
Hey @ianstormtaylor, thank you for reviewing this! 

Currently, operations like `moveNode` or `removeNode` are missing the said `node` and can't be inverted outside of the editor by using only the `operation`. I think we should be able to maintain a remote source of data by using operations only and not rely on any given `value`. But until operations are compatible with this, those inverted operations make it impossible to implement this second source of data. 

We encountered a hard wall while implementing OT for example. By wiring an adapter on the `onChange` and going through each operations, without the nodes our only way to get changes is by using the included `value`, whose not synced when operation is being inverted.

I ended to the conclusion that the only options here are:
- The work being made on #2225. I think it might work except during `moveNode` operations
- Keeping the `value`, but inverting it alongside the inversion of an operation's properties
- This `isInverted` flag, and let the dev's implementation handle it

I'd definitely go for #2225, with the inclusion of the `node` property on `moveNode`. The flag introduced by this PR isn't a real long-term solution.

ianstormtaylor:
@Calyhre I totally agree. I'm in favor of #2225 and the related other PRs that we want to bundle with it to make the breaking changes less difficult.

PR #2512, opened 2018-12-27
fix unwrapBlockAtRange disorder

akumatus:

fixing a _bug_

#### What's the new behavior?

Blocks behave right  after unwrap.

As `document` variable is assigned before `forEach` and there are serval operations in each loop, it's not equal to `editor.value.document`. This causes `index` variable inaccurate and blocks disorder after unwrap.

Fixes: #
Reviewers: @ianstormtaylor


CameronAckermanSEL:
LGTM good find

thesunny:
Thanks @akumatus and for vetting @CameronAckermanSEL

PR #2514, opened 2018-12-28
Use strict equality except for null/undefined for improved performance and behavior

DomAmato:

Strict equality checking can help prevent unintended behavior and improves performance since no type conversion takes place when evaluating. For a more in depth discussion around the subject see this stackoverflow discussion: https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

#### What's the new behavior?

there is no new behavior

See the above stack overflow discussion among other resources regarding strict equality if there is any confusion

No


ianstormtaylor:
I'm down with this, but could you also add the ESLint rule so that we maintain this style going forward?

DomAmato:
Will do 👍

DomAmato:
Ok added the rule and made sure tests pass :)

ianstormtaylor:
Beautiful, thank you very much @DomAmato!

PR #2522, opened 2019-01-03
Refactor to utils

thesunny:

Refactored methods to utils in preparation for adding Android mobile support.

#### What's the new behavior?

Apart from refactoring to make the code more reusable, I changed the set selection code after a spelling correction (in `after.js` on the `onInput` method). It used to move the cursor to the end of the change. I've changed it so that it sets the cursor position to be the same as in the DOM. In other words, the browser decides the behavior of where the cursor should be which makes Slate behave the way the browser natively behaves.

This same logic will be used when I implement the Android composition handling.

Mostly, this just creates two new methods in `utils` called `set-selection-from-dom` and `set-text-from-dom-node`.

The `onInput` method calls `set-text-from-dom-node` then `set-selection-from-dom`. These are mostly just refactoring of existing methods. Also `set-selection-from-dom` is called from the `onFocus` method.

Does not fix, but is in preparation for https://github.com/ianstormtaylor/slate/issues/2062 as well as almost all the other issues related to Android and mobile.


CameronAckermanSEL:
This looks pretty good to me!

thesunny:
Thanks @CameronAckermanSEL 

Merged it.

PR #2524, opened 2019-01-04
Remove wasted space in header for small devices by enabling horizonta…

thesunny:

Improving a feature

#### What's the new behavior?

The slate examples list of editors is getting long. When viewing the demo on small devices, most of the screen is taken up by the list of editor types. This makes the list scroll horizontally on mobile devices so it doesn't take all the screen space from the editor itself.

Added a media-query for max-width: 768px and below and added overflow-x: scroll to it along with a few other things to make it work.

### DIFF
--- a/examples/app.js
+++ b/examples/app.js
@@ -106,6 +106,13 @@ const TabList = styled('div')`
   & > * + * {
     margin-left: 0.5em;
   }
+
+  @media only screen and (max-width: 768px) {
+    box-sizing: border-box;
+    width: 100%;
+    overflow-x: scroll;
+    white-space: nowrap;
+  }
 `
 
 const MaskedRouterLink = ({ active, ...props }) => <RouterLink {...props} />


thesunny:
If anybody really hates this, I'll take it out but otherwise I'll probably merge it pretty soon. It just makes the example editor easier to see on a mobile device.

CameronAckermanSEL:
@thesunny  https://github.com/ianstormtaylor/slate/pull/2450

thesunny:
Ooh, thanks, I'm glad you pointed that out. No, it's unrelated. Didn't know that PR existed.

I'll hold off on this unless people end up preferring this.

Pros:
- less code
- easy to scroll in mobile

Cons:
- Less standard than the menu button

thesunny:
This is what it looks like

image

ericedem:
This is a cool idea! I think unfortunately though the list has gotten a big long for a horizontal scroll, since it will be several pages wide on small devices. I'd recommend checking out #2450 and let me know if you had suggestions on how to improve!

thesunny:
Sounds reasonable. I left it here mostly in case the other PR needs more time until it gets merged. I think this is just 6 lines of CSS and improves the UI compared to the old version for mobile dev.

I'll close it but if it looks like your PR needs more time to get through, consider merging this and then removing it when you merge yours. Removing would be easy since it's just CSS.

Issue #2517, opened 2019-01-01
Weird scrolling issue on .focus()


raymestalez:
#### Do you want to request a _feature_ or report a _bug_?

Report a _bug_.

#### What's the current behavior?
I have a weird scrolling issue in slate. Look at the `scrollTo()` function here:
https://jsfiddle.net/495oha18/3/

My goal is to scroll to the Line 30 in the example, position it in the middle of the visible area, and move cursor to it (so that user could click a button that automatically takes them to the part of the document I want, where he can start typing right away).

It works as expected until I do `editor.focus()`, which breaks the scroll position, even though it really shouldn't. In the example, I do `setTimeout` for one second to illustrate the issue. When you click the button scrolling works as expected, then after one second `.focus()` is triggered and scrolling jumps into the wrong position.

If you click the button again, issue disappears and everything works as I want it to. To see the issue again you gotta refresh the page.

On [gif](https://i.imgur.com/lwNNwt1.gifv) you can see that once I click on outline in my editor, it triggers a smooth scroll to the part of the document I want. Then, once the smooth scroll is finished, I run `editor.moveToRangeOfNode(node).moveToEndOfBlock()` to move cursor in correct position and then `editor.focus()` focus the editor, and that's when selection snaps into the wrong position.

I'm running Google Chrome on Ubuntu 18.04, Slate 0.44.0.

PR #2526, opened 2019-01-06
Prevent default browser scroll on content focus

brendancarney:
Fixing a bug. 

#### What's the new behavior?

Without this, when calling `this.element.focus`, the browser will scroll to the top of the container if the selection is not already set. Since we set the selection later in this function, we don't need the browser to handle this. 

Fixes: #2517
Reviewers: @


ianstormtaylor:
Interesting, I didn't even realize that option existed. Thank you @brendancarney!

PR #2529, opened 2019-01-09
Refactor duplicate code in mixin

zhujinxuan:

Refactor

#### What's the new behavior?
Eliminate some duplication of `mixin(create())`

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate/src/interfaces/object.js
+++ b/packages/slate/src/interfaces/object.js
@@ -48,8 +48,6 @@ function create(type) {
  * Mix in the object interfaces.
  */
 
-debugger
-
 Object.entries({
   Block,
   Change,
--- a/packages/slate/src/interfaces/object.js
+++ b/packages/slate/src/interfaces/object.js
@@ -48,18 +48,22 @@ function create(type) {
  * Mix in the object interfaces.
  */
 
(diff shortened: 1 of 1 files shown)


thesunny:
Looks good to me.

ianstormtaylor:
Hey guys, what is the reason for doing this? It doesn't strike me as necessary at all, and it just makes this harder to read...

This seems unnecessary...

zhujinxuan:
@ianstormtaylor I feel like repeatedly typing 
```
mixnin(create(objName), [obj])
```
is a bit duplicate, and easy to make mistake (like typo) when refactor code.

I think it is perhaps better to make the same patterns combined together.  Feel free to revert this PR if you think it is not necessary.

ianstormtaylor:
@zhujinxuan I understand what you're saying, but in practice that kind of logic leads to refactoring things that don't need to be refactored, and in ways that make the code harder to read. 

The pull requested code here is harder to understand than the original. The duplication in this case wasn't actually causing any harm. I'll revert it when I get the time.

In the future, please don't merge PR's like this.

thesunny:
@ianstormtaylor Sorry, that's my bad especially since I agree with you.

PR #2530, opened 2019-01-09
[Refactor] Use Symbol instead of ad-hoc strings for isObject

zhujinxuan:

Refactor

#### What's the new behavior?

Previously, we used a `a["@@__SLATE_....@@"]` for describing the type.  I think it is perhaps nicer to use Symbol to remove some ad-hoc `@@` chars.

Use symbol and let the mixin supports symbols.

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate/src/utils/is-object.js
+++ b/packages/slate/src/utils/is-object.js
@@ -4,22 +4,25 @@
  * @type {Object}
  */
 
-export const TYPES = {
-  block: '@@__SLATE_BLOCK__@@',
-  change: '@@__SLATE_CHANGE__@@',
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Does `Symbol` work in all the browsers we support?

zhujinxuan:
If we determine browsers as the browser list
```
last 2 Chrome versions
last 2 Firefox versions
last 2 Safari versions
last 2 Edge versions
IE 11
>1%
```

I am not sure about `>1%`.  Symbol is supported natively in Chrome, Firefox, Safari and Edge.  It is not supported natively in IE, but the babel has poly-filled it.

Issue #2487, opened 2018-12-06
Examples stop working in MS Edge when changing focus


n1ghtmare:
#### Do you want to request a _feature_ or report a _bug_?
_bug_

#### What's the current behavior?
Changing focus breaks the examples in MS Edge.

![slatejs-bug](https://user-images.githubusercontent.com/3255810/49604135-92591880-f9a6-11e8-892d-ff3035b6f1be.gif)

Version: Microsoft Edge **42.17134.1.0**

#### What's the expected behavior?
Should be able to refocus.

Let me know if I can provide any additional info.

steida:
Thank god Microsoft Edge will die soon.

pjshy:
I have the same problem **Version: Microsoft Edge 42.17134.1.0**

alizeaiter:
Still having the same problem, editor focus 30% of the time on Edge

alizeaiter:
A temporary fix would be focusing the editor manually on click events.
```
setTimeout(() => {
	editor.focus();
}, 0);
```

alizeaiter:
Scratch that, just found out that even selecting doesn't work on edge.

pjshy:
find a issue in react-dom about this, but i am not sure https://github.com/facebook/react/issues/13212

davegw:
Believe this was fixed in #2535. However, while focus can be given back to the editor now, the initial position of the cursor on refocus goes back to the previous position before the editor was blurred.

brendancarney:
It does look like this is fixed in latest. @davegw if there's not an issue for what you mentioned, we should create that.

PR #2535, opened 2019-01-14
Fix mouse refocus in IE 11

statianzo:
Fixing a **bug** 🐛 

#### What's the new behavior?

In IE11, refocusing by click would cause the editor to immediately lose focus because of `editor.deselect().focus()`. The original behavior was added by #2396.

Now when refocusing, IE no longer does a deselect.

I've updated to skip the `deselect` in IE browsers as suggested by following @epotockly's [comment](https://github.com/ianstormtaylor/slate/issues/2364#issuecomment-443717248)

>Fixed this by running editor.focus() without deselect() inside onFocus event in MS Edge, problem definitely are in new range selection, but i don't know how to fix it without removing function

* [ ] The linter passes with `yarn lint`. (Fix errors with `yarn prettier`.) *Unrelated errors*

Fixes: #2364 
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -3,7 +3,7 @@ import Debug from 'debug'
 import Hotkeys from 'slate-hotkeys'
 import Plain from 'slate-plain-serializer'
 import getWindow from 'get-window'
-import { IS_IOS } from 'slate-dev-environment'
+import { IS_IOS, IS_IE } from 'slate-dev-environment'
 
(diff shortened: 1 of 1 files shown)


davegw:
I believe there is also a focus issue in Edge (#2487). Not sure if the same fix can be applied or if it's a different issue.

flynnham:
Thanks for this PR. It's working wonders

ianstormtaylor:
Thanks @statianzo!

Issue #2528, opened 2019-01-08
bug: `normalizeNode` is called on non-existent path and throws an error.


bebbi:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

When merging nodes, the `getDirtyPaths` function collects an illegal path, leading `normalizeNodeByPath` function to throw on its `assertNode` call.

##### Repro

1. Open https://jsfiddle.net/y5q9120p/2/
2. Open the dev console
3. Click into the first line, drag down to anywhere in the middle of the last line to create a selection.
4. Press Delete
5. See "Uncaught Error: `Node.assertNode` could not find node with path or key: List [ 0, 0, 0, 11 ]"

The path appears to refer to the right side adjacent text node that gets merged with its neighbor [here](https://github.com/ianstormtaylor/slate/blob/09db2c93b6f69b2c1c733bca8fd1bcf71d7688a3/packages/slate/src/plugins/core.js#L165)

#### What's the expected behavior?

This should just work without an error.

bebbi:
Possibly related, while above works for a simple node: https://jsfiddle.net/x96ja1fv/ it fails with an `assertNode` error on undo.
Should that be a separate issue?

bebbi:
@CameronAckermanSEL I've tried it and no, the issue is independent of https://github.com/ianstormtaylor/slate/pull/2505.

While in that PR dirty handling appears key based instead of path based, the dirty calculation code path seems to be the same.

The reason that PR appears to work is because its early exit on missing nodes in `normalizeNodeByPath` instead of the original assertNode(path).

While I am not familiar with the details of dirty path calculation, I did notice that the offending path is added to the `editor.tmp.dirty` array before the call to [merging adjacent text nodes](https://github.com/ianstormtaylor/slate/blob/09db2c93b6f69b2c1c733bca8fd1bcf71d7688a3/packages/slate/src/plugins/core.js#L165).

During that merge, the (offending) node gets deleted but the path never gets removed from the dirty paths array. It looks like that is the actual issue to me.

Example:

before merging the adjacent nodes:

tree structure:

```
14 (document)
| 15 container
| | 16 paragraph
| | | 17 text "This is abo" 
| | | 28 text "e content" 
| | 21 container
...
```

and

[...]

PR #2536, opened 2019-01-14
Fix dirty path normalization on move_node

Calyhre:

Fixing a bug

#### What's the new behavior?

`PathUtils.transform` now returns correct updated path after a `move_node` operation

In this particular example:
```xml
<document>
  <block key="A" path=[0] />
  <block key="B" path=[1] />
  <block key="C" path=[2] />
</document>
```

If I move the node `B` to the path `[2, 0]` so it becomes this:
```xml
<document>
  <block key="A" path=[0] />
  <block key="C" path=[1]>
    <block key="B" path=[1, 0] />
  </block>
</document>
```

If in the same `withoutNormalization` batch, I have a `tmp.dirty` that contains `[1, 0]`, the current `PathUtils.tranform` will transform our dirty path according to the `move_node` operation's `newPath`. Here it will be `[2,0]`. The new dirty path will eventually become `[2, 0, 0]` refering to a non-existing block, that will fail during the next normalization.

A very simple test to show the issue in real life:

0. Open the `console`
1. Go to [Slatejs.org `rich-text` example](https://www.slatejs.org/#/rich-text)
2. Insert a list after the very first paragraph
3. Create a selection from somewhere in the paragraph, and the list item below, like this:  <details>
    <summary>Screenshot of a selection example</summary>
    screen shot 2019-01-14 at 19 05 53
  </details>

4. Press backspace then undo
6. You should now have an error in your console:
  ```
  element.js:1991 Uncaught Error: `Node.assertNode` could not find node with path or key: List [ 2, 0, 0 ]
    at a.ct.(/anonymous function) [as assertNode] (https://www.slatejs.org/main-cf7fba3ac17904d8ae2c.js:1:144119)
    at it (editor.js:597)
    at editor.js:580
    at e.value (editor.js:398)
    at rt (editor.js:578)
    at e.value (editor.js:400)
    at a.value (editor.js:275)
    at on-history.js:116
    at La.withoutSaving (on-history.js:168)
    at e.value (editor.js:148)
  ```

Fixes: #2528 
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate/src/utils/path-utils.js
+++ b/packages/slate/src/utils/path-utils.js
@@ -351,9 +351,17 @@ function transform(path, operation) {
     const npAbove = isAbove(np, path)
 
     if (pAbove) {
-      path = np.concat(path.slice(p.size))
+      if (isAfter(np, p)) {
+        path = decrement(np, 1, min(np, p) - 1).concat(path.slice(p.size))
+      } else {
+        path = np.concat(path.slice(p.size))
+      }
     } else if (pEqual) {
-      path = np
+      if (isAfter(np, p)) {
+        path = decrement(np, 1, min(np, p) - 1)
+      } else {
+        path = np
+      }
     } else {
       if (pYounger) {
         path = decrement(path, 1, pIndex)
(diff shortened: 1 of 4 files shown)


zhujinxuan:
@ianstormtaylor Hi, just wondering if there is a specific reason of using dirty path rather than using normalizeChildren + cached getFirstInvalidDescendant?  Personally I think the later one is a bit easier to implement.

ianstormtaylor:
Thank you @Calyhre this is awesome!

PR #2537, opened 2019-01-15
Add path to Range Points

valoricDe:
As far as I know Points consist out of key, offset and path.

The Slate-react -> Text Component assumes that paths are set for anchor and focus points.
Reference: https://github.com/ianstormtaylor/slate/blob/5d0f53da3ffd6016b779551c065fdc874ea4baf7/packages/slate-react/src/components/text.js#L122

Therefore I resolve the path with document.getPath.


valoricDe:
Hey @ianstormtaylor it's three months anniversary of this pull request. It would be great if you could merge this 2 lines of code

TimSusa:
@ianstormtaylor Would you be ok by declaring yourself ready to comply with any comment on this?

ianstormtaylor:
Thanks @valoricDe!

Issue #2544, opened 2019-01-17
Paste on line with void an no text deletes void node


ldavidpace:
#### Do you want to request a _feature_ or report a _bug_?
_bug_

#### What's the current behavior?

When Pasting on a line that only contains a void with no additional text it deletes the void. 

[JSFIDDLE](https://jsfiddle.net/paceld/74c8dvep/)
![inlinedeleteonpaste](https://user-images.githubusercontent.com/25068382/51335931-8437f680-1a40-11e9-9e25-edbe0a07c855.gif)
Mac Os High Sierra,  Chrome: 71.0.3578.98  slate version: ^0.44.8

#### What's the expected behavior?

It should not paste the text next to the void.

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #2545, opened 2019-01-17
Pasting into a Block with only a Void Inline

ldavidpace:
 BUG  🐛 

#### What's the new behavior?

Nodes will count extra nodes as content during insertFragmentAtRange. So that void nodes can be counted as content.

inlinedeleteonpastefixed

Added an isEmpty method on the node interface. It seems that to check if a node is empty there is a check for text being empty. Feels like there should be an explicit check for if a node is empty. I am now using it in the InsertFragmentAtRange method in order to allow for pasting onto the same line as a void Inline without deleting it.

* [ x] The new code matches the existing patterns and styles.
* [ x] The tests pass with `yarn test`.
* [ x] The linter passes with `yarn lint`. (Fix errors with `yarn prettier`.)
* [ x] The relevant examples still work. (Run examples with `yarn watch`.)

Fixes: [2544](#https://github.com/ianstormtaylor/slate/issues/2544)

### DIFF
--- a/docs/reference/slate/node.md
+++ b/docs/reference/slate/node.md
@@ -372,7 +372,6 @@ Check whether a node exists in the tree by `path` or `key`.
 
 Check whether a node is inside a `range`. This will return true for all [`Text`](./text.md) nodes inside the range and all ancestors of those [`Text`](./text.md) nodes up to this node.
 
-
 ### `isEmpty`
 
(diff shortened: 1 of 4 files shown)


ianstormtaylor:
Hey @ldavidpace can you fix this bug without introducing the new method? Thanks!

Issue #2368, opened 2018-11-02
fix editing with IMEs


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

Right now Slate doesn't work well with IME's, for different reasons. There are a few different issues we are tracking that we hope will make this better.

- [x] ~~https://github.com/ianstormtaylor/slate/issues/2067 — we added a hack to re-render the placeholder when a composition starts, but it's re-rendering the entire editor, which we think has an issue with IME's. This would be the first thing to fix.~~

- [ ] https://github.com/ianstormtaylor/slate/issues/2062 — second, @thesunny has been awesome enough to do a bunch of research to document the exact order of composition events on Android devices with IME-based keyboards. Although fixing this issue would likely guarantee IME support, it's also possible that there's a smaller amount of work that can get IME support without full Android support first. (Just implementing desktop Chrome's behaviors.)

There might be other things we're not aware of. But until these are solved, IME's will have issues. If you are experience with IME's and want to help, that would be amazing! I'm going to be referring all IME-related bugs here in the meantime, to focus efforts.

ericedem:
I was doing some research on handling IMEs, and I found some helpful articles:

Background in what IMEs are and how they work:
https://developer.mozilla.org/en-US/docs/Mozilla/IME_handling_guide

Some potential solutions for handling IME events:
https://www.stum.de/2016/06/24/handling-ime-events-in-javascript/

CodePlayer:
This issue is also bothering me, I hope to have a better solution as soon as possible.

thesunny:
FYI, Android support published to Slate @ 0.44.12 four days ago.

There are still some bugs but a huge improvement over existing support.

froatsnook:
@thesunny Just wanted to report that I updated to the latest slate and it's working SO much better in Android.  Thank you for fixing this!

thesunny:
I'm working on fixing the remaining Android bugs now.

In order to be more efficient, please submit bugs and help with testing over in this issue:

**Fix all showstopper bugs in Slate on Android (Help us by reporting and testing bugs)**
https://github.com/ianstormtaylor/slate/issues/2726

Akashdeep-Sarin:
Is there any plan for fixing the Japanese Language bugs...??
The issue seems to exist till now and is replicable in this [example](https://www.slatejs.org/#/rich-text)

Akashdeep-Sarin:
[...]

PR #2553, opened 2019-01-21
Android 8.0, 8.1 and 9.0 Support

thesunny:

Adding a feature. Support for Android devices.

#### What's the new behavior?

The new behavior is that Android will work on versions 8.0, 8.1 and 9.0.

Live example here: https://thesunny.github.io/slate/#/rich-text

This is a complicated PR and encompasses more than a month of full-time work and has been in progress for over a year. Some of the details can be found in this issue:

- https://github.com/ianstormtaylor/slate/issues/2062

Many issues and how they were overcome are described in the PR itself in a markdown file as well as comments in the plugin itself:

- packages/slate-react/src/plugins/ANDROID.md
- packages/slate-react/src/plugins/android.js

This PR avoids impacting any other browsers by existing in an `android` plugin that only gets added to Android browsers. The reason is that the logic in Android is so different than the other browsers that code reuse didn't make sense.

Note: There are some known bugs but I don't think they crash the editor. They do, however, sometimes end up with the wrong text which you can fix and these are edge cases.

I'm PR'ing this because this is a significant improvement over the current state of the editor on Android. It is also before I attempt to create a `Combo` object that will try to fix the issues in a different way and which may introduce bugs of its own. The `Combo` object will simplify most of the complex objects and will also be required to fix Android API 24 and 25 (Android 7.00, 7.1.1) and probably needed for Android 6.1 if we want to attempt that in the future.

Fixes:
https://github.com/ianstormtaylor/slate/issues/2062
https://github.com/ianstormtaylor/slate/issues/2368

Reviewers:
@ianstormtaylor


packages/slate-react/src/plugins/ANDROID.md
```diff
@@ -0,0 +1,219 @@
...
+# Debugging
+
+```
+slate:android,slate:before,slate:after,slate:update,slate:reconcile
+```
+
+# API 25
+
+### Backspace Handling
+
```
DamareYoh: "There appears to be no way to discern that backspace was placed in the events."

packages/slate-react/package.json
```diff
@@ -27,6 +27,7 @@
     "slate-plain-serializer": "^0.6.33",
     "slate-prop-types": "^0.5.24",
     "slate-react-placeholder": "^0.1.12",
+    "snapback": "^0.8.0",
```
DamareYoh: are we using this dependency anywhere?

packages/slate-react/src/plugins/ANDROID.md
thesunny: Thanks. I'll update that text. Should have said something like: There is no way to tell from the event that there was a backspace in API 25. I've got a working DOM analysis tool that will help me determine when a node has been obliterated which indicates that a backspace occurred over a critical threshold like deleting a block or a span that includes a mark. In other cases like a regular character deletion, we the reconcile code handles it properly.

packages/slate-react/package.json
thesunny: Thanks again! No we are not! That was a library that helped revert the DOM but it turned out to not be sufficient so I wrote the custom Snapshot library.

jtadmor:
Is the main issue just that this android plugin needs to come before the existing `slate-react` `BeforePlugin()`?

Would a mechanism that allows users to import `BeforePlugin` and manually provide a truly custom plugin list be sufficient?

thesunny:
@jtadmor Not entirely sure what you are asking but this plugin will just work. It adds the correct behavior through a plugin to Android browsers only.

I'm just asking for review in case I missed something.

thesunny:
So, my apologies, it feels like my PR wording is confusing.

I have pointed out that there are known bugs in this PR but my plan is to merge even with the known bugs. This is because even with the bugs, this is a marked improvement over the existing version which is not usable on Android.

I'm asking for review in case there are other issues which should prevent this merge from happening like, for example, breaking behavior on other browsers.

My plan is to fix the known issues using a `Combo` library but felt like this PR makes Android usable and hence we should merge at this point.

jtadmor:
@thesunny 

I guess what I'm asking is how much of this should be put into `slate` itself vs. just maintained as a separate plugin. If `slate` / `slate-react` allowed you to put this plugin before the `BeforePlugin`, would that make less code be required to go into the core lib?

thesunny:
@jtadmor 

In my opinion this goes into Slate proper for these reasons:

- All other browsers are supported in Slate proper and not as plugins including iOS support which has lower market share than Android.
- It's a reliable way to keep Android support up-to-date with current Slate releases including when Slate has breaking changes.
- We modify the `content` module in Slate which can't be done in a plugin
- We add the `android` plugin before Slate's `before` plugin. This can't be done by a regular plugin that you add to the editor.
- Android uses modules inside of Slate that aren't exposed as an export
- There are modules created for Android support which would benefit other parts of Slate

Some of the technical limitations could be overcome; however, as a whole, there is more benefit to keeping it in Slate proper. The primary benefit to separating it is a smaller deliverable if you don't need Android support; however, the additional size is small compared to the benefits IMO.

While it was a hard problem to solve, the amount of code isn't that big and implementing the Combo module later will make it smaller still. Just did a quick check and minified, I estimated it is about 10k (Slate is 175k) and using Slate's average gzip compression ratio, it should be 2k of deliverable (Slate is 40k).

packages/slate-react/src/utils/get-selection-from-dom.js
```diff
@@ -0,0 +1,77 @@
+import findRange from './find-range'
+
+export default function getSelectionFromDOM(window, editor, domSelection) {
```
ianstormtaylor: Can you rename this to `getDomSelection`?

packages/slate-react/src/utils/is-input-data-enter.js
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Note that at this point it is too late to prevent the event from affecting
+ * the DOM so we use other methods to clean the DOM up after we have detected
+ * the input.
+ *
+ * @param  {String} data
+ * @return {Boolean}
+ */
+
+export default function isInputDataEnter(data) {
```
ianstormtaylor: Can you rename this to be Android specific it it's only used there. Or better yet inline this utility functions if it's so custom?

packages/slate-react/src/utils/set-selection-from-dom.js
```diff
@@ -1,77 +1,14 @@
-import findRange from './find-range'
+import getSelectionFromDOM from './get-selection-from-dom'
```
ianstormtaylor: Can you rename this to `setDomSelection`?

packages/slate-react/src/utils/set-text-from-dom-node.js
```diff
@@ -1,5 +1,19 @@
 import findPoint from './find-point'
```
ianstormtaylor: Can you rename this to `setDomText`?

packages/slate-react/src/utils/slate-snapshot.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  /**
+   * Apply the snapshot to the DOM and set the selection in the Editor.
+   *
+   * @param {Editor} editor
+   */
+
+  apply(editor) {
+    if (editor == null) throw new Error('editor is required')
```
ianstormtaylor: Why would editor ever be null?

packages/slate-react/src/utils/slate-snapshot.js
```diff
@@ -0,0 +1,52 @@
+import closest from './closest'
```
ianstormtaylor: Could you rename this to `DomSnapshot` instead? as that seems more clear.

packages/slate-react/src/plugins/android.js
```diff
@@ -0,0 +1,624 @@
+import Debug from 'debug'
```
ianstormtaylor: Why is this implemented as a separate plugin? It seems like it's going to make it really hard to maintain, since most fixes/changes won't be kept in sync? And it also means that we don't get any of the IME fixes that Android made in other platforms?

thesunny:
Sure @ianstormtaylor, I'll do my best to answer your questions.

packages/slate-react/src/utils/slate-snapshot.js
steida: TypeScript knows. 😇

packages/slate-react/src/utils/get-selection-from-dom.js
thesunny: Wouldn't the rename be confusing? It seems to imply that the return value would be a DOM selection. I named it this way because most methods in Slate appear written from the point of view of Slate. So in this instance, we are getting the Slate (implied) selection from the DOM selection. I can see my name was not clear (since it confused you) but perhaps something like `getSlateSelectionFromDomSelection`? I'm not passionate about it though so if you are still happy with `getDomSelection` I will change.

packages/slate-react/src/utils/is-input-data-enter.js
thesunny: Yes, actually, I think I can remove this as there is another function now is-input-data-last-char that does the same thing and more.

packages/slate-react/src/utils/set-selection-from-dom.js
thesunny: Please see comment on `getSelectionFromDOM` and let me know if you'd still like to rename it.

packages/slate-react/src/utils/slate-snapshot.js
thesunny: Sure.

packages/slate-react/src/utils/slate-snapshot.js
thesunny: Nope, it shouldn't be. It was a guard during development because the method didn't originally take an `editor` object and I saw a few bugs when I forgot to add the editor. I can remove it.

packages/slate-react/src/plugins/android.js
thesunny: So last year when I took a stab at this, I tried implementing Android as part of the regular `before` and `after` plugins; however, I found it difficult to implement because the changes were almost always to fix Android specific issues and they almost always had side effects that affected other browsers. I ended up having very low confidence that I didn't break things in other editors, and a very high rate of Android specific code mixed in and making it difficult to reason about the code for other browsers (lots of "if Android do this else do that statements").

Moving Android into a separate plugin is one reason development went so much faster this time. I didn't have to worry about regressions in other browsers.

However, that said, my hope is that we will see what's common between the different editors and the way they handle compositions and we could perhaps merge them together... possibly. But I can't say that with high confidence because most of the problems seem specific to Android. I feel a higher confidence level in abstracting compositions into a pluggable composition library that we could customize for each browser.

An example of Android having specific changes that don't affect other browsers is that in the normal Slate `before` plugin, there is some code in there that says something like if a composition ends but a new composition starts again then we should treat it as if the composition hasn't ended yet. That logic doesn't work in Android so I created an `isStrictComposing` in addition to the `isComposing` flag. So now we had two separate but similar concepts of `isComposing`. But later, we have to split code on whether to use isComposing or isStrictComposing depending on the browsers. Also as I dove deeper this time around, I realized that Android will actually do a `compositionEnd` without having a corresponding `compositionStart` which was a kind of implied composition. There is no corresponding logic in other browsers that I'm aware of. These situations had to be handled differently. Furthermore, there is just a lot of really convoluted logic that figures out intent through the combination of events and not a single event. This is in my code which I was planning to abstract into a Combo object.

I can see how looking at my code one might think that there must be an easier way. There is no such complexity in any of the other browsers. That was my feeling at first as well. I will continue to simplify and abstract the code to make it more understandable. I'm fairly confident though that at this point in time, a separate Android plugin makes the most sense and is, in fact, easier rather than harder to keep in sync by virtue of the fact that its easier to understand the code by (IMO) an order of magnitude.

You asked about IME fixes in other platforms as well. I do feel like the IME fixes for Android can be implemented into the rest of Slate but perhaps in another step; for example, there are already a lot of nice abstracted pieces that we can move into the regular handling that wasn't abstracted before. My first step was to get something working for Android though.

packages/slate-react/src/utils/slate-snapshot.js
thesunny: Done

packages/slate-react/src/utils/slate-snapshot.js
thesunny: Done.

packages/slate-react/src/plugins/debug.js
```diff
@@ -0,0 +1,64 @@
...
+    'onKeyDown',
+    'onPaste',
+    'onSelect',
+  ]
+
+  const plugin = {}
+
+  for (const eventName of events) {
+    plugin[eventName] = function(event, editor, next) {
+      debug(eventName, { event })
```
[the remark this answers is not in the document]
thesunny: This plugin should be removed from the default plugins in `plugins/dom.js`.

One question I have is whether we might want to leave the library in the code as it is valuable during debugging. I relied on this when developing the Android plugin. The reason is it allowed me to see when an event passed through a plugin. For example, I can see if the android plugin allowed or disallowed an event from passing through to the `before` plugin. The debug statements in the `before` plugin don't tell us this definitively as sometimes they return before the `debug` message gets called.

Also, sometimes I would place this debug plugin after the `before` plugin to see what made it through there as well.

If I'm working on Android again, I'd probably add it back in for the duration of my work and then remove it when I'm done. It would be simpler if it wasn't in a separate NPM package which I have to add to `package.json` and remove.

I've removed it from `plugins/dom` just now (Done). Please let me know what you'd prefer about keeping or removing the library as a whole. Thanks.

packages/slate-react/src/components/content.js
```diff
@@ -133,6 +147,14 @@ class Content extends React.Component {
...
 
   componentDidUpdate() {
+    debug.update('componentDidUpdate')
+    // NOTE:
+    // Don't disable `updateSelection` on Android. Clicking a word and a
+    // suggestion breaks on API27. It does fix the crazy jumping cursor loop
+    // when doing an auto-suggest on a fully enclosed text with bold though.
+    // Most likely it still needs other fix issues though.
+    //
+    // if (IS_ANDROID) return
```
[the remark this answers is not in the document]
thesunny: Code cleanup that I missed. Sorry about that.

Removal. Done.

thesunny:
@ianstormtaylor One question, should I resubmit the PR request with this branch? Not sure of the workflow here.

packages/slate-react/src/components/content.js
```diff
@@ -4,7 +4,12 @@ import Types from 'prop-types'
 import getWindow from 'get-window'
 import warning from 'tiny-warning'
 import throttle from 'lodash/throttle'
-import { IS_FIREFOX, HAS_INPUT_EVENTS_LEVEL_2 } from 'slate-dev-environment'
+import {
+  IS_ANDROID,
+  IS_FIREFOX,
+  HAS_INPUT_EVENTS_LEVEL_2,
+} from 'slate-dev-environment'
+import ANDROID_API_VERSION from '../utils/android-api-version'
```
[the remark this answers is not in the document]
ianstormtaylor: @thesunny that sounds fine to me. We can always split it apart later if it needs to be, but keeping all of the environment logic in one place is a good way to start.

packages/slate-react/src/plugins/debug.js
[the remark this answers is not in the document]
ianstormtaylor: @thesunny that makes sense, I'm okay with keeping it in the source here. And potentially we can improve it over time to make it more useful UX wise.

packages/slate-react/src/utils/get-selection-from-dom.js
ianstormtaylor: Very fair. Okay let's keep the current naming and I can figure out a rename in the future after it's in master.

ianstormtaylor:
@thesunny thanks! Yup, if you can resubmit the PR since I needed to revert it, that would be perfect. And we can discuss the plugin vs. core in that new PR I think.

packages/slate-react/src/utils/set-text-from-dom-node.js
thesunny: Going to leave it as per your other comment. Thanks.

thesunny:
@ianstormtaylor I think I've addressed all the addressable issues save the plugin issue now. Let me know if there's anything else.

packages/slate-react/src/plugins/debug.js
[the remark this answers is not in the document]
thesunny: Acknowledged. Left it in.

Issue #2328, opened 2018-10-26
MoveNodeByPath broken?


skogsmaskin:
#### Do you want to request a _feature_ or report a _bug_?
I was trying to add some tests for the ``by-path`` commands, and tried adding a simple test for just switching positions of two blocks. I can't get it to work, it will just replace one of the blocks.

I.e. ``change.moveNodeByPath([0], [1], 1)``

#### What's the current behavior?
It will replace one of the blocks leaving only one of the two blocks behind.

Relevant fiddle: https://jsfiddle.net/skogsmaskin/t54xc1ap/1/

#### What's the expected behavior?

As I understand moveNodeByPath it should not replace blocks, but just move it? I want to move the first block below the second one.

Could be I misunderstand something here, some input would be appriciated.

ianstormtaylor:
@skogsmaskin might it have something to do with the inconsistent behavior mentioned in https://github.com/ianstormtaylor/slate/issues/2192?

skogsmaskin:
@ianstormtaylor Yeah, that's probably part of the problem (the path annotation for the newPath param is very confusing). But I consider it a bug that it will actually delete/replace nodes like in the above fiddle? It should maybe throw an error instead until ``moveNodeByPath`` is refactored like you propose in #2192.

jtadmor:
@skogsmaskin @ianstormtaylor I believe that the particular fiddle was not showing an actual break in `moveNodeByPath`.

I was able to get the operation to work as expected by changing the code to call the underlying `applyOperation` directly: https://jsfiddle.net/La0bsjm1/3/

I think the reason we were seeing that behavior was because of the `CorePlugin` rule preventing mixing `block` and `text` nodes: https://github.com/ianstormtaylor/slate/blob/7a6de8725c21343f265f95b5ab916c586040dec6/packages/slate/src/plugins/core.js#L66

The jsfiddle is actually moving a `paragraph` node to be the child of a `paragraph` that already has an existing `text` child node, so (I think) the new node was just being removed right after the `move_node` operation went through.

ianstormtaylor:
Is this fixed now?

nz-chris:
Doesn't seem to be fixed. I ended up using `editor.applyOperation` with type as `move_node` instead.

ianstormtaylor:
[...]

PR #2555, opened 2019-01-22
move_node tests and code clean up

jtadmor:

Improvement (mostly adding tests, some refactoring).

Mostly, this work / investigation was done to satisfy: https://github.com/ianstormtaylor/slate/issues/2192 (which appears to no longer be necessary, unless I misunderstood)

That issue was cited as a blocker for: https://github.com/ianstormtaylor/slate/pull/2225

This should also clear up: https://github.com/ianstormtaylor/slate/issues/2328

#### What's the new behavior?

n/a

- `invert.move_node` was substantially simplified using `PathUtils`, bringing the code more in line with `element.moveNode`
- `Commands.moveNodeByPath` function arguments were renamed (internal) to clarify their purpose.
- tests added

Fixes: #2192, #2328
Reviewers: @

### DIFF
--- a/packages/slate/src/interfaces/element.js
+++ b/packages/slate/src/interfaces/element.js
@@ -1763,13 +1763,12 @@ class ElementInterface {
     const newParentPath = PathUtils.lift(newPath)
     this.assertNode(newParentPath)
 
-    const [p, np] = PathUtils.crop(path, newPath)
-    const position = PathUtils.compare(p, np)
+    const position = PathUtils.compare(path, newPath)
 
     // If the old path ends above and before a node in the new path, then
     // removing it will alter the target, so we need to adjust the new path.
     if (path.size < newPath.size && position === -1) {
-      newPath = PathUtils.decrement(newPath, 1, p.size - 1)
+      newPath = PathUtils.decrement(newPath, 1, path.size - 1)
     }
 
     let ret = this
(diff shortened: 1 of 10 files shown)


packages/slate/src/interfaces/element.js
```diff
@@ -1763,13 +1763,12 @@ class ElementInterface {
     const newParentPath = PathUtils.lift(newPath)
     this.assertNode(newParentPath)
 
-    const [p, np] = PathUtils.crop(path, newPath)
-    const position = PathUtils.compare(p, np)
```
jtadmor: Minor issue, but we don't really need `p` and `np` since `compare` only checks the smaller size anyway. This avoids the two calls to `.slice` that `crop` makes. We just use `path.size` in the conditional below because we know it is smaller than `newPath` in order to hit that line

packages/slate/src/operations/invert.js
```diff
@@ -45,36 +45,17 @@ function invertOperation(op) {
...
-        newPath
-          .slice(0, newPathLast)
-          .every((e, i) => e == inverseNewPath.get(i)) &&
-        newPath.last() <= inverseNewPath.get(newPathLast)
-      ) {
-        inverseNewPath = inverseNewPath
-          .slice(0, newPathLast)
-          .concat(inverseNewPath.get(newPathLast) + 1)
-          .concat(inverseNewPath.slice(newPathLast + 1, inverseNewPath.size))
+      if (path.size < newPath.size && position === -1) {
```
jtadmor: Use the same style as the check in `element`. Some new tests were added to make sure this all still works.

Issue #2000, opened 2018-07-27
convert remaining `Node.get*` methods to use paths


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

In #1997 many (almost all) of the `Node.get*` models were converted to use paths, which are often much more performant. But a few commonly used ones remain key-based. This is unfortunate since the more often key-based getters are used, the more often the expensive to calculate `key -> path` mapping has to be created.

#### What's the expected behavior?

If we can get Slate's core internal usage of keys to be almost all gone, performance should increase a good amount.

The remaining methods that overtly take `key` arguments are:

- [ ] `Node.getNextBlock`
- [ ] `Node.getPreviousBlock`
- [ ] `Node.getOffset`

But then there are also the `*AtRange` methods which are currently key-based under the covers as well:

- [ ] `Node.getActiveMarksAtRange`
- [ ] `Node.getCharactersAtRange`
- [x] `Node.getFragmentAtRange`
- [ ] `Node.getInlinesAtRangeAsArray`
- [ ] `Node.getInsertMarksAtRange`
- [ ] `Node.getOrderedMarksAtRange`

And then some of the less standardized ones which we can either convert to using paths or try to remove entirely:

- [ ] `Node.getFirstInvalidDescendant`
- [ ] `Node.getSelectionIndexes`
- [ ] `Node.getOrderedMarksBetweenPositions`
- [ ] `Node.getMarksAtPosition`
- [ ] `Node.getTextsBetweenPositionsAsArray`

ianstormtaylor:
I'd go further to say that this is probably the biggest place for performance gains in the codebase, at least as far as the data model is concerned. (There may be other ones at the DOM level in `slate-react`.) So any help is much appreciated!

For an example of places where the `key -> path` map is used in ways that degrade performance, check out this comment: https://github.com/ianstormtaylor/slate/pull/1993#issuecomment-408546270

Issue #2413, opened 2018-11-11
introduce a "refs" concept


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Idea.

#### What's the current behavior?

Right now the `Point` model in Slate has "paths" and "keys" (in addition to "offsets") to reference a location in the document. Paths are generally more performant, since you can use them to traverse directly to a node. Whereas keys need to be looked up by searching through the entire tree. (We cache keys to try to alleviate this issue, but the caches are invalidated often.)

Ideally we'd use paths in many more places, but the problem with paths is that you can't guarantee they aren't stale after an operation is applied because it may have been invalidated by `insert_node/remove_node/etc.` operations. The same goes for offsets as well as paths, since `insert_text/remove_text` operations could invalidate an offset as well.

But we might be able to create editor-specific objects to alleviate this...

#### What's the expected behavior?

We could create a mutable `Pointer` model that referenced a point in the editor's document. When you create it, you pass in the `Point` you're wanting to keep an up-to-date reference to.

```js
const pointer = new Pointer(point)
```

Then, whenever a new operation is applied to the editor, it updates all of the pointers it's keeping track of, so that they aren't stale. And at any time in the future, you can access `pointer.path` or `pointer.offset` and it will be guaranteed not to be stale.

```js
const { path, offset } = pointer
```

This could alleviate a lot of the current uses of keys. Even `renderNode` would be able to receive a `Pointer` instance since its mutable and referentially equal. (It can't receive the `path` right now because every time the path changed it would force a re-render.)

However, this could also help for asynchronous commands, like inserting images, where you need to perform some asynchronous work (eg. uploading the image) before you insert it, and you need to make sure that the insertion point stays updated as more editing happens.

---

Also if anyone has a better name than "pointer" I'm all ears!

CameronAckermanSEL:
Would index be an inappropriate name? That might create confusion with array indexes.

Dundercover:
What is your reasoning in keeping it mutable instead of immutable and closer to the editor? like `editor.createPointer` and `editor.getPointerValue`?

Instead of `Index` I would much rather see `Pointer` or something more verbose like `PointReference` or `PointRef`. Will there ever be higher abstractions of this like `NodeRef` or `RangeRef`?

ianstormtaylor:
[...]

PR #2557, opened 2019-01-23
Convert remaining `Node.get*` methods to use paths

jtadmor:

Improvement.

Opening this PR right now to get discussion on the basic approach (see below).

#### What's the new behavior?

See #2000 for background.

The difficulty using `path` rather than `key` is that there is no `node.path` equivalent for `node.key`.

There are at least 4 potential approaches (some or all could be combined):
1. Store some sort of pointer / reference to `path` on the node itself, and update all of them whenever a change is made that might affect some paths. This would make the rest of the changes easier, just switch to use `node.path` where we use `node.key`.

2. Store some sort of pointer / reference to `path` as a prop when rendering in the editor to make `getSelectionIndexes` easier to calculate at `render` time. 

(See https://github.com/ianstormtaylor/slate/issues/2413 for related discussion).

3. Create a `getPathForKey(key)` that is more efficient than `getKeysToPathsTable` for when we just need the path for a particular node (we can bail early). This would then allow us to call `getPathForKey(node.key)` where we use `node.key`. We could cache this to some extent, but probably calling it more than once or twice is just as expensive as `getKeysToPathsTable`.

4. Create methods that are similar to current `getFirstText`, `getNextBlock` etc, but which return the node and path. This allows some of the more complex methods such as `getActiveMarksAtRange` to iterate internally and still continue to use paths to call various getters.

This PR has three parts:
1. Adds tests
2. Some work towards (4) (see e.g. `getNextTextAndPath`, which can be used to improve the perf of `getActiveMarksAtRange`)
3. A promising approach for (2)

Opening now so we can discuss the approaches / tradeoffs.

If we don't like them, I can just add the tests in a separate PR (fwiw, some of them are failing against `master`, but I'm pretty sure I've captured the expected / desired behavior, and so those might represent legitimate bugs).

Fixes: #2000
Reviewers: @ianstormtaylor @alanctkc

### DIFF
--- a/packages/slate/src/interfaces/element.js
+++ b/packages/slate/src/interfaces/element.js
@@ -219,7 +219,7 @@ class ElementInterface {
         return false
       }
 
-      if (child.object != 'text') {
+      if (child.object !== 'text') {
         ret = child.forEachDescendantWithPath(iterator, childPath, findLast)
         return ret
       }
--- a/packages/slate/src/interfaces/element.js
+++ b/packages/slate/src/interfaces/element.js
@@ -452,7 +452,6 @@ class ElementInterface {
    */
 
   getClosestBlock(path) {
-    path = this.resolvePath(path)
     const closest = this.getClosest(path, n => n.object === 'block')
     return closest
   }
@@ -465,7 +464,6 @@ class ElementInterface {
    */
 
   getClosestInline(path) {
-    path = this.resolvePath(path)
     const closest = this.getClosest(path, n => n.object === 'inline')
     return closest
   }
@@ -547,7 +545,7 @@ class ElementInterface {
   /**
    * Get a descendant node.
    *
-   * @param {List|string} path
+   * @param {List|String} path
    * @return {Node|Null}
    */
 
@@ -556,12 +554,11 @@ class ElementInterface {
     if (!path || !path.size) return null
 
     let node = this
-    let i = 0
 
-    while (i < path.size && node) {
-      node = node.getIn(['nodes', path.get(i)])
-      i++
-    }
+    path.forEach(index => {
+      node = node.getIn(['nodes', index])
+      return !!node
+    })
 
     return node
   }
@@ -630,7 +627,7 @@ class ElementInterface {
   /**
    * Get the furthest ancestor of a node.
    *
-   * @param {Path|string} path
+   * @param {List|String} path
    * @return {Node|Null}
    */
 
@@ -649,7 +646,6 @@ class ElementInterface {
    */
 
   getFurthestBlock(path) {
-    path = this.resolvePath(path)
     const furthest = this.getFurthest(path, n => n.object === 'block')
     return furthest
   }
@@ -662,7 +658,6 @@ class ElementInterface {
    */
 
   getFurthestInline(path) {
-    path = this.resolvePath(path)
     const furthest = this.getFurthest(path, n => n.object === 'inline')
     return furthest
   }
@@ -670,7 +665,7 @@ class ElementInterface {
   /**
    * Get the furthest ancestor of a node, where all ancestors to that point only have one child.
    *
-   * @param {Path|string} path
+   * @param {Path} path
    * @return {Node|Null}
    */
 
--- a/packages/slate/src/interfaces/element.js
+++ b/packages/slate/src/interfaces/element.js
@@ -112,7 +112,7 @@ class ElementInterface {
   }
 
   /**
-   * Recursively find all descendant nodes by `iterator`.
+   * Recursively find a descendant node by `iterator`.
    *
    * @param {Function} iterator
    * @return {Node|Null}
@@ -129,6 +129,45 @@ class ElementInterface {
     })
 
     return found
+  }
+
+  /**
+   * Recursively find a descendant node and its path by `iterator`.
+   *
+   * @param {Function} iterator
+   * @return {Null|[Node, List]}
+   */
+
+  findDescendantAndPath(
+    iterator,
+    pathToThisNode = PathUtils.create([]),
+    findLast = false
+  ) {
+    let found
+    let foundPath
+
+    this.forEachDescendantWithPath(
+      (node, path, nodes) => {
+        if (iterator(node, path, nodes)) {
+          found = node
+          foundPath = path
+          return false
+        }
+      },
+      pathToThisNode,
+      findLast
+    )
+
+    return found ? [found, foundPath] : null
+  }
+
+  // Easy helpers to avoid needing to pass findLast boolean
+  findFirstDescendantAndPath(iterator, pathToThisNode) {
+    return this.findDescendantAndPath(iterator, pathToThisNode, false)
+  }
+
+  findLastDescendantAndPath(iterator, pathToThisNode) {
+    return this.findDescendantAndPath(iterator, pathToThisNode, true)
   }
 
   /**
@@ -157,6 +196,39 @@ class ElementInterface {
   }
 
   /**
+   * Recursively iterate over all descendant nodes with `iterator`. If the
+   * iterator returns false it will break the loop.
+   * Calls iterator with node and path.
+   *
+   * @param {Function} iterator
+   * @param {List} path
+   * @param {Boolean} findLast - whether to iterate in reverse order
+   */
+
+  forEachDescendantWithPath(iterator, path = PathUtils.create([]), findLast) {
+    let nodes = this.nodes
+    let ret
+
+    if (findLast) nodes = nodes.reverse()
+
+    nodes.forEach((child, i) => {
+      const childPath = path.concat(i)
+
+      if (iterator(child, childPath, nodes) === false) {
+        ret = false
+        return false
+      }
+
+      if (child.object != 'text') {
+        ret = child.forEachDescendantWithPath(iterator, childPath, findLast)
+        return ret
+      }
+    })
+
+    return ret
+  }
+
+  /**
    * Get a set of the active marks in a `range`.
    *
    * @param {Range} range
@@ -169,31 +241,33 @@ class ElementInterface {
 
     if (range.isCollapsed) {
       const { start } = range
-      return this.getMarksAtPosition(start.key, start.offset).toSet()
+      return this.getMarksAtPosition(start.path, start.offset).toSet()
     }
 
     const { start, end } = range
-    let startKey = start.key
+    let startPath = start.path
     let startOffset = start.offset
-    let endKey = end.key
+    let endPath = end.path
     let endOffset = end.offset
-    let startText = this.getDescendant(startKey)
+    let startText = this.getDescendant(startPath)
+    let endText = this.getDescendant(endPath)
 
-    if (startKey !== endKey) {
-      while (startKey !== endKey && endOffset === 0) {
-        const endText = this.getPreviousText(endKey)
-        endKey = endText.key
+    if (!PathUtils.isEqual(startPath, endPath)) {
+      while (!PathUtils.isEqual(startPath, endPath) && endOffset === 0) {
+        ;[endText, endPath] = this.getPreviousTextAndPath(endPath)
         endOffset = endText.text.length
       }
 
-      while (startKey !== endKey && startOffset === startText.text.length) {
-        startText = this.getNextText(startKey)
-        startKey = startText.key
+      while (
+        !PathUtils.isEqual(startPath, endPath) &&
+        startOffset === startText.text.length
+      ) {
+        ;[startText, startPath] = this.getNextTextAndPath(startPath)
         startOffset = 0
       }
     }
 
-    if (startKey === endKey) {
+    if (PathUtils.isEqual(startPath, endPath)) {
       return startText.getActiveMarksBetweenOffsets(startOffset, endOffset)
     }
 
@@ -202,21 +276,23 @@ class ElementInterface {
       startText.text.length
     )
     if (startMarks.size === 0) return Set()
-    const endText = this.getDescendant(endKey)
     const endMarks = endText.getActiveMarksBetweenOffsets(0, endOffset)
     let marks = startMarks.intersect(endMarks)
+
     // If marks is already empty, the active marks is empty
-    if (marks.size === 0) return marks
+    if (marks.size === 0) {
+      return marks
+    }
 
-    let text = this.getNextText(startKey)
+    ;[startText, startPath] = this.getNextTextAndPath(startPath)
 
-    while (text.key !== endKey) {
-      if (text.text.length !== 0) {
-        marks = marks.intersect(text.getActiveMarks())
+    while (!PathUtils.isEqual(startPath, endPath)) {
+      if (startText.text.length !== 0) {
+        marks = marks.intersect(startText.getActiveMarks())
         if (marks.size === 0) return Set()
       }
 
-      text = this.getNextText(text.key)
+      ;[startText, startPath] = this.getNextTextAndPath(startPath)
     }
     return marks
   }
@@ -342,8 +418,8 @@ class ElementInterface {
 
   getChild(path) {
     path = this.resolvePath(path)
-    if (!path) return null
-    const child = path.size === 1 ? this.nodes.get(path.first()) : null
+    if (!path || path.size > 1) return null
+    const child = this.nodes.get(path.first())
     return child
   }
 
@@ -376,6 +452,7 @@ class ElementInterface {
    */
 
   getClosestBlock(path) {
+    path = this.resolvePath(path)
     const closest = this.getClosest(path, n => n.object === 'block')
     return closest
   }
@@ -388,6 +465,7 @@ class ElementInterface {
    */
 
   getClosestInline(path) {
+    path = this.resolvePath(path)
     const closest = this.getClosest(path, n => n.object === 'inline')
     return closest
   }
@@ -469,17 +547,23 @@ class ElementInterface {
   /**
    * Get a descendant node.
    *
-   * @param {List|String} path
+   * @param {List|string} path
    * @return {Node|Null}
    */
 
   getDescendant(path) {
     path = this.resolvePath(path)
-    if (!path) return null
+    if (!path || !path.size) return null
 
-    const deep = path.flatMap(x => ['nodes', x])
-    const ret = this.getIn(deep)
-    return ret
+    let node = this
+    let i = 0
+
+    while (i < path.size && node) {
+      node = node.getIn(['nodes', path.get(i)])
+      i++
+    }
+
+    return node
   }
 
   /**
@@ -546,14 +630,14 @@ class ElementInterface {
   /**
    * Get the furthest ancestor of a node.
    *
-   * @param {Path} path
+   * @param {Path|string} path
    * @return {Node|Null}
    */
 
   getFurthestAncestor(path) {
     path = this.resolvePath(path)
-    if (!path) return null
-    const furthest = path.size ? this.nodes.get(path.first()) : null
+    if (!path || !path.size) return null
+    const furthest = this.nodes.get(path.first())
     return furthest
   }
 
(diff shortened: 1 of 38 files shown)


jtadmor:
(Unfortunately it seems that many of the `changes` and `models` benchmarks no longer work post `0.42.0` release that changes `value.change` -> `editor.change`.

I will try to separately take some time to refactor those so we can use them going forward.

jtadmor:
So here is what we have so far:

```
benchmarks
    changes
      ✔ add-mark: 61.24 -> 76.74 ops/sec (25.30% faster)
      ★ delete-backward:
            user: 88.62 -> 117.55 ops/sec (32.64% faster)
            real: 99.63 -> 138.03 ops/sec (38.54% faster)
      ★ delete-forward:
            user: 94.62 -> 129.18 ops/sec (36.53% faster)
            real: 106.99 -> 147.76 ops/sec (38.10% faster)
      ✖ insert-node-by-key:
            user: NaN -> NaN ops/sec (NaN% slower)
            real: NaN -> NaN ops/sec (NaN% slower)
      ✔ insert-text-by-key: 212.15 -> 247.90 ops/sec (16.85% faster)
      ★ insert-text:
            user: 71.20 -> 95.65 ops/sec (34.34% faster)
            real: 79.38 -> 113.80 ops/sec (43.35% faster)
      ★ normalize:
            user: 88.52 -> 134.62 ops/sec (52.08% faster)
            real: 97.56 -> 160.76 ops/sec (64.78% faster)
      ✔ remove-node-by-key: 89.72 -> 107.05 ops/sec (19.31% faster)
      ★ set-blocks:
            user: 143.84 -> 192.14 ops/sec (33.57% faster)
            real: 155.11 -> 212.33 ops/sec (36.90% faster)
      ★ split-block:
            user: 42.77 -> 56.03 ops/sec (30.99% faster)
            real: 47.44 -> 64.22 ops/sec (35.36% faster)
    models
      ✔ from-json-big: 13.20 -> 13.22 ops/sec (0.15% faster)
      ✔ from-json: 673.32 -> 624.56 ops/sec (8.00% slower)
      ★ get-active-marks-at-range:
            user: 2425.19 -> 7441.32 ops/sec (206.83% faster 🙌 )
            real: 2882.27 -> 11225.81 ops/sec (289.48% faster 🙌 )
      ★ get-ancestors:
            user: 2207.39 -> 3887.20 ops/sec (76.10% faster)
            real: 2483.08 -> 4408.68 ops/sec (77.55% faster)
      ✔ get-blocks: 12054.74 -> 10864.26 ops/sec (11.00% slower)
      ✔ get-common-ancestor: 5002.48 -> 4502.59 ops/sec (11.00% slower)
      ★ get-descendant:
            user: 3169.75 -> 4791.45 ops/sec (51.16% faster)
            real: 3583.80 -> 5295.84 ops/sec (47.77% faster)
      ✔ get-furthest-ancestor: 5130.20 -> 4837.64 ops/sec (6.00% slower)
      ✔ get-inlines: 13828.54 -> 11670.79 ops/sec (18.00% slower)
      ★ get-leaf-blocks-at-range:
            user: 723.44 -> 1938.30 ops/sec (167.93% faster 🙌 )
            real: 989.10 -> 3374.90 ops/sec (241.21% faster 🙌 )
      ★ get-leaf-inlines-at-range:
            user: 664.17 -> 1024.20 ops/sec (54.21% faster)
            real: 797.17 -> 1314.38 ops/sec (64.88% faster)
      ✔ get-leaves: 1588404.21 -> 1323132.37 ops/sec (20.00% slower)
      ✔ get-marks-at-range: 816.86 -> 984.49 ops/sec (20.52% faster)
      ✔ get-marks: 3263.45 -> 3045.23 ops/sec (7.00% slower)
      ★ get-parent:
            user: 3054.65 -> 4315.50 ops/sec (41.28% faster)
            real: 3347.41 -> 4733.00 ops/sec (41.39% faster)
      ✔ get-path: 4770.06 -> 4701.75 ops/sec (1.00% slower)
      ★ get-texts-at-range:
            user: 1903.06 -> 4531.84 ops/sec (138.13% faster 🙌 )
            real: 2155.01 -> 5494.71 ops/sec (154.97% faster 🙌 )
      ✔ get-texts: 12039.79 -> 11066.59 ops/sec (9.00% slower)
      ★ has-node-multiple:
            user: 2176.28 -> 3666.96 ops/sec (68.50% faster)
            real: 2410.70 -> 4024.98 ops/sec (66.96% faster)
      ★ has-node:
            user: 3087.33 -> 4416.95 ops/sec (43.07% faster)
            real: 3392.67 -> 4788.70 ops/sec (41.15% faster)
      ✔ to-json: 2920.13 -> 2807.39 ops/sec (4.00% slower)
    html-serializer
      ✔ deserialize: 153.02 -> 158.60 ops/sec (3.65% faster)
      ✔ serialize: 692.19 -> 694.58 ops/sec (0.35% faster)
    plain-serializer
      ✔ deserialize: 3221.63 -> 3915.51 ops/sec (21.54% faster)
      ✔ serialize: 3908.12 -> 4062.15 ops/sec (3.94% faster)
    rendering
      ✔ decoration: 55.34 -> 62.83 ops/sec (13.53% faster)
      ✔ large-with-selection: 34.23 -> 39.28 ops/sec (14.73% faster)
      ✔ normal: 67.86 -> 73.46 ops/sec (8.25% faster)
```

jtadmor:
Some notes / thoughts:

- Low level benchmarks seem much improved. This is good. However, I'm not convinced this necessarily represents dramatic improvements in the places where perf is hurting (normalization, big changes). I may work on a separate PR to try to get some better / more complex benchmarks in, and encourage others to do likewise while this PR is discussed.
- Passing `path` instead of `key` to a lot of these methods will likely break a lot of the current memoization. We are using `WeakMap` but unless the same exact `List` is passed in, object equality will not hold and it will fail. We should consider changing memoization to coerce any `List` args with a `.join(.)`. This will slow down memoization / all method calls a bit.
- Should we try to get some commands to move over to using `path` instead of `key` internally?

packages/slate/src/interfaces/node.js
```diff
@@ -112,9 +112,20 @@ class NodeInterface {
 
   getNode(path) {
     path = this.resolvePath(path)
+    return this.getNodeByPath(path)
```
ianstormtaylor: Why the need to separate this out into another method here?

packages/slate/src/models/point.js
```diff
@@ -355,7 +355,23 @@ class Point extends Record(DEFAULTS) {
     }
 
     const { key, offset, path } = this
-    const target = node.getNode(key || path)
```
ianstormtaylor: Can you switch this section back, but maybe use `path || key` instead if that's the ideal case? I'd rather separate out performance improvements into another PR from this one, instead of bundling them in with the changes.

packages/slate/src/utils/path-utils.js
```diff
@@ -212,14 +212,26 @@ function isYounger(path, target) {
  * Lift a `path` to refer to its parent.
  *
  * @param {List} path
- * @return {Array}
+ * @return {List}
```
ianstormtaylor: ❤️

packages/slate/src/interfaces/element.js
```diff
@@ -367,6 +447,34 @@ class ElementInterface {
...
 
+  /**
+   * Get closest parent of node that matches an `iterator`.
+   *
+   * @param {List} path
+   * @param {Function} iterator
+   * @return {Node|Null}
+   */
+
+  getClosestByPath(path, iterator) {
```
ianstormtaylor: Why the need for this method? This seems like another optimization that I'd rather have in a separate PR, since the original `getClosest` already works by path.

That way, instead of creating these extra methods we can just use the optimized versions when we drop key support which is coming soon.

packages/slate/src/interfaces/element.js
```diff
@@ -474,11 +584,28 @@ class ElementInterface {
...
-    const ret = this.getIn(deep)
-    return ret
+  /**
+   * Get a descendant node by its path.
+   *
+   * @param {List} path
+   * @return {Node|Null}
+   */
+
+  getDescendantByPath(path) {
```
ianstormtaylor: Same here, I want to avoid these `*ByPath` methods. We used to have all of our methods separated by path and key and I'd like to move us towards only supporting paths at all.

jtadmor:
Thanks for the review, Ian.

I'll try and split this PR into a few different ones, so we can get something in that lets all the `Node.get` methods accept paths.

I'm fine with dropping the `byPath` methods from the interface itself. I would still think it might be a good idea to have some utility methods that:
(a) bypass the `resolvePath` lookup when we know we have a path
(b) return a `[node, path]` type signature. Even if we are accepting paths in our methods, if we turn around and look everything up by key, we aren't really utilizing any of the advantages. This could definitely be rolled into a different PR though.

ianstormtaylor:
@jtadmor for the (a) case there, it won't matter as soon as we remove keys, which we're hoping to do soon, since we won't need the `resolvePath` calls I think? For (b) I agree that sometimes returning `[node, path]` is a nice touch, and I'm not against that at all, I like that solution you came up with.

Thanks!

packages/slate/src/models/point.js
jtadmor: Hi Ian,

I tried reverting the changes to the `point` model, as well as just using `path || key`, but I was running into failing tests. See my more complete comments on the changes I made.

jtadmor:
It seems like the one area that I had to overstep to make some changes was the `point` model.

The issue I needed to address is a situation where a range is passed into `normalize` that contains both a `path` and `key`, but where the two are not in alignment. This was passing through tests unnoticed before because we were not using the `path` as much for look-ups, but not that we are, tests were failing.

I think there are a couple approaches we could take. One potentially less intrusive one is to replace the following code by always calling `node.getPath(target.key)`, thus ensuring the path is always valid.

```
    const point = this.merge({
      key: target.key,
      path: path == null ? node.getPath(target.key) : path,
      offset: offset == null ? 0 : Math.min(offset, target.text.length),
    })
})
```

My approach should be roughly equivalent to your suggestion of calling `getNode(path || key)`, except that it also fixes the case where `path` exists but no longer points to a valid node, in which case all we need to do is look it up by `key` and adjust the path. If we don't add that code, we end up hitting the following code:

```
    if (!target) {
      warning(false, "A point's `path` or `key` invalid and was reset!")
      const text = node.getFirstText()
      if (!text) return Point.create()
      const point = this.merge({
        key: text.key,
        offset: 0,
        path: node.getPath(text.key),
      })
      return point
    }
}
```

We don't want to hit this just because the `path` was wrong.

jtadmor:
Perf (vs master) with current changes to `point` model:

```
  benchmarks
    changes
      ✔ add-mark: 80.81 -> 90.05 ops/sec (11.44% faster)
      ✔ delete-backward: 101.99 -> 104.70 ops/sec (2.66% faster)
      ✔ delete-forward: 102.61 -> 116.56 ops/sec (13.60% faster)
      ✔ insert-fragment-deep-with-problems: 87.06 -> 100.77 ops/sec (15.74% faster)
      ✔ insert-fragment-large: 4.80 -> 4.68 ops/sec (3.00% slower)
      ✔ insert-fragment: 88.58 -> 92.93 ops/sec (4.91% faster)
      ✔ insert-node-by-key: 162.67 -> 190.95 ops/sec (17.38% faster)
      ✔ insert-text-by-key: 199.31 -> 188.30 ops/sec (6.00% slower)
      ✔ insert-text: 84.82 -> 92.63 ops/sec (9.20% faster)
      ✔ normalize-with-problems: 12.37 -> 12.68 ops/sec (2.47% faster)
      ★ normalize:
            user: 100.92 -> 134.35 ops/sec (33.13% faster)
            real: 150.49 -> 210.20 ops/sec (39.67% faster)
      ✔ remove-node-by-key: 106.96 -> 113.18 ops/sec (5.82% faster)
      ✔ set-blocks: 128.69 -> 140.35 ops/sec (9.06% faster)
      ★ split-block:
            user: 52.44 -> 57.76 ops/sec (10.16% faster)
            real: 71.26 -> 96.05 ops/sec (34.80% faster)
    models
      ✔ from-json-big: 14.33 -> 14.53 ops/sec (1.43% faster)
      ✔ from-json: 776.59 -> 912.43 ops/sec (17.49% faster)
      ★ get-active-marks-at-range:
            user: 791.31 -> 8067.54 ops/sec (919.52% faster 🙌 )
            real: 1233.13 -> 16281.77 ops/sec (1220.36% faster 😱 )
      ✔ get-ancestors: 849.76 -> 1109.30 ops/sec (30.54% faster)
      ✔ get-blocks: 13428.36 -> 13943.85 ops/sec (3.84% faster)
      ✔ get-common-ancestor: 1128.40 -> 1136.00 ops/sec (0.67% faster)
      ★ get-descendant:
            user: 922.33 -> 1136.45 ops/sec (23.21% faster)
            real: 1356.25 -> 1808.83 ops/sec (33.37% faster)
      ✔ get-furthest-ancestor: 1707.33 -> 1718.72 ops/sec (0.67% faster)
      ✔ get-inlines: 20340.92 -> 16256.84 ops/sec (25.00% slower)
      ★ get-leaf-blocks-at-range:
            user: 774.97 -> 1730.92 ops/sec (123.35% faster 🙌 )
            real: 1519.80 -> 3207.27 ops/sec (111.03% faster 🙌 )
      ★ get-leaf-inlines-at-range:
            user: 688.52 -> 1025.94 ops/sec (49.01% faster)
            real: 1157.08 -> 1617.37 ops/sec (39.78% faster)
      ✔ get-leaves: 711462.68 -> 671281.78 ops/sec (6.00% slower)
      ✔ get-marks-at-range: 868.16 -> 947.32 ops/sec (9.12% faster)
      ✔ get-marks: 3777.36 -> 3557.86 ops/sec (6.00% slower)
      ★ get-parent:
            user: 3104.61 -> 4775.34 ops/sec (53.81% faster)
            real: 4949.73 -> 5977.02 ops/sec (20.75% faster)
      ✔ get-path: 5547.45 -> 5301.61 ops/sec (5.00% slower)
      ★ get-texts-at-range:
            user: 1889.71 -> 4717.25 ops/sec (149.63% faster 🙌 )
            real: 2469.09 -> 7300.67 ops/sec (195.68% faster 🙌 )
      ✔ get-texts: 18000.77 -> 16905.00 ops/sec (6.00% slower)
      ★ has-node-multiple:
            user: 2522.42 -> 3876.44 ops/sec (53.68% faster)
            real: 3474.16 -> 5067.83 ops/sec (45.87% faster)
      ★ has-node:
            user: 3291.46 -> 4973.10 ops/sec (51.09% faster)
            real: 4571.57 -> 6294.98 ops/sec (37.70% faster)
      ✔ to-json: 3797.19 -> 4113.14 ops/sec (8.32% faster)
    html-serializer
      ✔ deserialize: 186.33 -> 179.87 ops/sec (4.00% slower)
      ✔ serialize: 750.78 -> 622.86 ops/sec (21.00% slower)
    plain-serializer
      ★ deserialize:
            user: 3261.39 -> 4002.36 ops/sec (22.72% faster)
            real: 6153.33 -> 8429.92 ops/sec (37.00% faster)
      ✔ serialize: 7439.56 -> 6969.68 ops/sec (7.00% slower)
    rendering
      ✔ decoration: 66.64 -> 68.55 ops/sec (2.87% faster)
      ✔ large-with-selection: 45.59 -> 46.60 ops/sec (2.22% faster)
      ✔ normal: 77.87 -> 79.86 ops/sec (2.56% faster)
```


Perf with my other proposed change:

```
benchmarks
    changes
      ✔ add-mark: 80.81 -> 86.93 ops/sec (7.57% faster)
      ✔ delete-backward: 101.99 -> 108.76 ops/sec (6.65% faster)
      ★ delete-forward:
            user: 102.61 -> 117.44 ops/sec (14.45% faster)
            real: 153.12 -> 206.45 ops/sec (34.83% faster)
      ✔ insert-fragment-deep-with-problems: 87.06 -> 111.72 ops/sec (28.32% faster)
      ✔ insert-fragment-large: 4.80 -> 4.70 ops/sec (2.00% slower)
      ✔ insert-fragment: 88.58 -> 90.45 ops/sec (2.11% faster)
      ✔ insert-node-by-key: 162.67 -> 166.62 ops/sec (2.42% faster)
      ✔ insert-text-by-key: 199.31 -> 200.73 ops/sec (0.71% faster)
      ✔ insert-text: 84.82 -> 88.31 ops/sec (4.11% faster)
      ✔ normalize-with-problems: 12.37 -> 12.38 ops/sec (0.05% faster)
      ★ normalize:
            user: 100.92 -> 131.60 ops/sec (30.41% faster)
            real: 150.49 -> 207.06 ops/sec (37.59% faster)
      ✔ remove-node-by-key: 106.96 -> 100.75 ops/sec (6.00% slower)
      ✔ set-blocks: 128.69 -> 136.21 ops/sec (5.84% faster)
      ✔ split-block: 52.44 -> 54.12 ops/sec (3.21% faster)
    models
      ✔ from-json-big: 14.33 -> 15.00 ops/sec (4.66% faster)
      ✔ from-json: 776.59 -> 819.72 ops/sec (5.55% faster)
      ✔ get-active-marks-at-range: 791.31 -> 836.93 ops/sec (5.77% faster)
      ✔ get-ancestors: 849.76 -> 1080.67 ops/sec (27.17% faster)
      ✔ get-blocks: 13428.36 -> 14503.10 ops/sec (8.00% faster)
      ✔ get-common-ancestor: 1128.40 -> 1107.27 ops/sec (2.00% slower)
      ✔ get-descendant: 922.33 -> 1088.12 ops/sec (17.97% faster)
      ✔ get-furthest-ancestor: 1707.33 -> 1707.62 ops/sec (0.02% faster)
      ✔ get-inlines: 20340.92 -> 17841.61 ops/sec (14.00% slower)
      ✔ get-leaf-blocks-at-range: 774.97 -> 976.56 ops/sec (26.01% faster)
      ★ get-leaf-inlines-at-range:
            user: 688.52 -> 858.79 ops/sec (24.73% faster)
            real: 1157.08 -> 1691.14 ops/sec (46.16% faster)
      ✔ get-leaves: 711462.68 -> 742397.55 ops/sec (4.35% faster)
      ✔ get-marks-at-range: 868.16 -> 671.51 ops/sec (29.00% slower)
      ✔ get-marks: 3777.36 -> 3012.14 ops/sec (25.00% slower)
      ★ get-parent:
            user: 3104.61 -> 4977.22 ops/sec (60.32% faster)
            real: 4949.73 -> 5811.46 ops/sec (17.41% faster)
      ✔ get-path: 5547.45 -> 5282.14 ops/sec (5.00% slower)
      ✔ get-texts-at-range: 1889.71 -> 2349.50 ops/sec (24.33% faster)
      ✔ get-texts: 18000.77 -> 17436.63 ops/sec (3.00% slower)
      ★ has-node-multiple:
            user: 2522.42 -> 4095.73 ops/sec (62.37% faster)
            real: 3474.16 -> 4900.46 ops/sec (41.05% faster)
      ★ has-node:
            user: 3291.46 -> 4982.40 ops/sec (51.37% faster)
            real: 4571.57 -> 5967.59 ops/sec (30.54% faster)
      ✔ to-json: 3797.19 -> 4451.78 ops/sec (17.24% faster)
    html-serializer
      ✔ deserialize: 186.33 -> 165.02 ops/sec (13.00% slower)
      ✔ serialize: 750.78 -> 658.09 ops/sec (14.00% slower)
    plain-serializer
      ✔ deserialize: 3261.39 -> 3810.37 ops/sec (16.83% faster)
      ✔ serialize: 7439.56 -> 7438.26 ops/sec (0.00% slower)
    rendering
      ✔ decoration: 66.64 -> 64.56 ops/sec (3.00% slower)
      ✔ large-with-selection: 45.59 -> 45.96 ops/sec (0.82% faster)
      ✔ normal: 77.87 -> 77.32 ops/sec (1.00% slower)
```

ianstormtaylor:
Hey @jtadmor, that makes perfect sense. It seems like those cases where key/path diverge are spots where our internal logic is broken still, since it's still relying on keys. And maybe our next step is to fail loudly on those cases and see why our assumptions are wrong. But trying to fix all of those in one pull request is probably too mind-bending to do.

I like everything you've laid out here, including those new helper iteration functions. Thank you very much for this pull request, it is really high quality!

PR #1661, opened 2018-02-21
Perhaps faster normalization: Never visit a descendant twice in normalizaton

zhujinxuan:
I see @czechdave  say that normalization potentially slows down the Value.fromJSON; And I think perhaps it is a result from `normalizeKeys.includes`, a O(childNodes^2) comparison.  

This PR take advantage of cache and validate to ensure speed up normalization.

I am not confident about the benchmark running on my laptop.  Could anyone help me rerun the benchmark in their own computer?
```
benchmarks
    html-serializer
      deserialize
        8.98 → 11.02 ops/sec
      serialize
        1057.71 → 1185.42 ops/sec
    plain-serializer
      deserialize
        553.52 → 536.36 ops/sec
      serialize
        9107.18 → 9525.40 ops/sec
    rendering
      normal
        111.93 → 126.17 ops/sec
    changes
      delete-backward
        105548.40 → 144896.70 ops/sec (37% faster)
      delete-forward
        19318.61 → 30794.16 ops/sec (59% faster)
      insert-text-by-key-multiple
        85.57 → 97.89 ops/sec
      insert-text-by-key
        1045.61 → 1251.19 ops/sec
      insert-text
        1071.37 → 1150.59 ops/sec
      normalize
        1435.06 → 215908.18 ops/sec (14945% faster) 😱
      split-block
        22.43 → 83.04 ops/sec (270% faster) 🙌
    models
      from-json
        126.48 → 221.79 ops/sec (75% faster)
      get-blocks-at-range
        515597.57 → 691750.84 ops/sec (34% faster)
      get-blocks
        2389340.86 → 2876387.07 ops/sec
      get-characters-at-range
        675896.57 → 780662.25 ops/sec
      get-characters
        2906877.93 → 3160090.06 ops/sec
      get-inlines-at-range
        715855.63 → 848482.59 ops/sec
      get-inlines
        2821376.68 → 3106357.04 ops/sec
      get-leaves
        1591780.52 → 1837992.19 ops/sec
      get-marks-at-range
        800173.93 → 890650.77 ops/sec
      get-marks
        3167696.93 → 3177520.59 ops/sec
      get-path
        975600.69 → 910776.69 ops/sec
      get-texts-at-range
        937712.33 → 924806.06 ops/sec
      get-texts
        3334330.56 → 3305250.49 ops/sec
      has-node-multiple
        128830.81 → 123509.47 ops/sec
      has-node
        985303.14 → 930603.71 ops/sec
      to-json
[...]

PR #2559, opened 2019-01-28
Mutate newDirtyPaths instead of creating a new array every iteration

brendancarney:
Performance. 

#### What's the new behavior?

Mutate `newDirtyPaths` on each iteration instead of copying it. For my unscientific paste test, this brings a 4s paste to about 2.5s. 

### DIFF
--- a/packages/slate/src/controllers/editor.js
+++ b/packages/slate/src/controllers/editor.js
@@ -100,14 +100,13 @@ class Editor {
 
     // Get the paths of the affected nodes, and mark them as dirty.
     const newDirtyPaths = getDirtyPaths(operation)
-    const dirty = this.tmp.dirty.reduce((memo, path) => {
-      path = PathUtils.create(path)
-      const transformed = PathUtils.transform(path, operation)
-      Array.prototype.push.apply(memo, transformed.toArray())
-      return memo
-    }, newDirtyPaths)
 
-    this.tmp.dirty = dirty
+    const dirty = this.tmp.dirty.map(path => {
+      path = PathUtils.create(path)
+      return PathUtils.transform(path, operation).toArray()
+    })
+
+    this.tmp.dirty = Array.prototype.concat.apply(newDirtyPaths, dirty)
 
     // If we're not already, queue the flushing process on the next tick.
     if (!this.tmp.flushing) {
--- a/packages/slate/src/controllers/editor.js
+++ b/packages/slate/src/controllers/editor.js
@@ -103,7 +103,7 @@ class Editor {
     const dirty = this.tmp.dirty.reduce((memo, path) => {
       path = PathUtils.create(path)
       const transformed = PathUtils.transform(path, operation)
-      memo = memo.concat(transformed.toArray())
+      Array.prototype.push.apply(memo, transformed.toArray())
       return memo
     }, newDirtyPaths)
 
--- a/packages/slate/src/controllers/editor.js
+++ b/packages/slate/src/controllers/editor.js
@@ -103,7 +103,8 @@ class Editor {
 
     const dirty = this.tmp.dirty.map(path => {
       path = PathUtils.create(path)
-      return PathUtils.transform(path, operation).toArray()
+      const transformed = PathUtils.transform(path, operation)
+      return transformed.toArray()
     })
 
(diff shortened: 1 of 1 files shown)


zhujinxuan:
Hi, can I know how do you test it?

I remember both push.apply and concat have similar speed. (https://jsperf.com/array-concat-vs-push-2/17)  

But perhaps there is another optimization like this:
```js
    const dirty = Array.prototype.concat.apply(
        newDirtyPath, 
        this.tmp.dirty.map(path => {      
           path = PathUtils.create(path)
          return PathUtils.transform(path, operation).toArray()
     })
   );
```

brendancarney:
Unfortunately, I think I spoke too soon. The test is slower in my actual slate example, so building the large array is likely slow. I'm concerned that different cases will have different results, so I'm going to keep experimenting and try to find a more reliable optimization here.

zhujinxuan:
@brendancarney Can I have a loook at the benchmark result by:

1. in master branch, `yarn build && yarn benchmark:save`
2. in your branch, `yarn benchmark`

brendancarney:
Sorry for the delay, I was out of town for a week. Unfortunately, I'm not sure there's a benchmark accurately reflecting my use case. Running the benchmarks show no meaningful improvements in my branch. 

However, I took a recent production example of a large document and tried pasting it into an editor when running the slate examples. Over several attempts with both branches, here are the rough averages:
master: ~60s
this PR: ~18s

Still really slow in both cases unfortunately. I understand that we'll probably want a benchmark that shows this improvement, so I'll try to get that added soon.

brendancarney:
Ok so I sort of have a benchmark ready, but still trying to figure out the best way to introduce it. Currently, we run each benchmark a minimum of 100 times. This is a little problematic for this specific case because it is least performant with larger documents, and it's really really slow. So running it one hundred times is probably not great. Should we make these options configurable by benchmark? 

Here are the results from inserting a paragraph block 1,000 times, wrapped in `editor.withoutNormalizing`:

screen shot 2019-02-18 at 3 31 05 pm

```
★ insert-block:
     user: 0.02 -> 0.07 ops/sec (217.23% faster 🙌 )
     real: 0.02 -> 0.08 ops/sec (230.25% faster 🙌 )
```
As you can see, still really really slow, but there is an improvement.

zhujinxuan:
@brendancarney The benchmark is very naive at this moment.  But I think the 200% is very convincing.

If you are interested, you can temporarily add a large document insert benchmark inside `benchmark/slate/changes` and then run it with only `yarn benchmark:save --grep pattern` and  `yarn benchmark --grep pattern` (where pattern shall be a RegExp).  

BTW, can you have a try on the `Array.prototype.concat.apply` version in the paste large text example? (Perhaps not faster because map is slow...)

I am a bit surprising that the paste took 18 seconds... Last year I did same test after #1661, it was taking 5 seconds (though still slow).  I think I may need to investigate why when I have time.

brendancarney:
Hi @zhujinxuan. It turns out they are about the same. I think your suggestion is slightly more readable. Thoughts?

Also, I did not commit a benchmark because it would be ridiculously slow with the current setup. 

To test the speed difference, you can cut and paste the large document example in master, and then in here.

zhujinxuan:
@brendancarney As there is significant difference between `concat <<< map` and push, I would suggest keep the push version.

packages/slate/src/controllers/editor.js
```diff
@@ -100,14 +100,12 @@ class Editor {
...
-      path = PathUtils.create(path)
-      const transformed = PathUtils.transform(path, operation)
-      memo = memo.concat(transformed.toArray())
-      return memo
-    }, newDirtyPaths)
-
-    this.tmp.dirty = dirty
+
+    const dirty = this.tmp.dirty.map(path =>
+      PathUtils.transform(PathUtils.create(path), operation).toArray()
```
ianstormtaylor: Is there any reason for collapsing this into a one-liner? Could you re-expand it to be multiple lines like it was previously? Having one-liners like this makes it much hard to debug I've found, since setting breakpoints and checking values is much harder.

brendancarney: For sure!

ianstormtaylor:
Thanks @brendancarney!

PR #2565, opened 2019-01-30
Android 8 and 9 Support

thesunny:

Adding a feature

#### What's the new behavior?

Add Android 8 and 9 support for Slate.

Please see https://github.com/ianstormtaylor/slate/pull/2553

Reviewers: @ianstormtaylor


thesunny:
**For reference, here's the comment from the original pull request with respect to Android as a plugin:**

So last year when I took a stab at this, I tried implementing Android as part of the regular `before` and `after` plugins; however, I found it difficult to implement because the changes were almost always to fix Android specific issues and they almost always had side effects that affected other browsers. I ended up having very low confidence that I didn't break things in other editors, and a very high rate of Android specific code mixed in and making it difficult to reason about the code for other browsers (lots of "if Android do this else do that statements").

Moving Android into a separate plugin is one reason development went so much faster this time. I didn't have to worry about regressions in other browsers.

However, that said, my hope is that we will see what's common between the different editors and the way they handle compositions and we could perhaps merge them together... possibly. But I can't say that with high confidence because most of the problems seem specific to Android. I feel a higher confidence level in abstracting compositions into a pluggable composition library that we could customize for each browser.

An example of Android having specific changes that don't affect other browsers is that in the normal Slate `before` plugin, there is some code in there that says something like if a composition ends but a new composition starts again then we should treat it as if the composition hasn't ended yet. That logic doesn't work in Android so I created an `isStrictComposing` in addition to the `isComposing` flag. So now we had two separate but similar concepts of `isComposing`. But later, we have to split code on whether to use isComposing or isStrictComposing depending on the browsers. Also as I dove deeper this time around, I realized that Android will actually do a `compositionEnd` without having a corresponding `compositionStart` which was a kind of implied composition. There is no corresponding logic in other browsers that I'm aware of. These situations had to be handled differently. Furthermore, there is just a lot of really convoluted logic that figures out intent through the combination of events and not a single event. This is in my code which I was planning to abstract into a Combo object.

I can see how looking at my code one might think that there must be an easier way. There is no such complexity in any of the other browsers. That was my feeling at first as well. I will continue to simplify and abstract the code to make it more understandable. I'm fairly confident though that at this point in time, a separate Android plugin makes the most sense and is, in fact, easier rather than harder to keep in sync by virtue of the fact that its easier to understand the code by (IMO) an order of magnitude.

You asked about IME fixes in other platforms as well. I do feel like the IME fixes for Android can be implemented into the rest of Slate but perhaps in another step; for example, there are already a lot of nice abstracted pieces that we can move into the regular handling that wasn't abstracted before. My first step was to get something working for Android though.

thesunny:
**Further comments:**

I feel like I need to talk about the difference between compositions in general and the rest of the event handlers (I'll call them atomic event handlers) and why it might be worthwhile to separate compositions into a different plugin.

- In atomic event handlers, we only need to look at that event in isolation. Was a key pressed? Great. Handle it and reflect that in the `Value`. In compositions, at the very least, we are looking at `compositionStart` and `compositionEnd`.
- In atomic event handlers, we can prevent the default event from being handled by the browser. In compositions, in many cases we should not (because it breaks the composition and the editor behaves differently) or cannot (because preventDefault doesn't actually prevent the action from happening).
- In atomic event handlers, we only need to look at the individual event to determine what we should do. The logic is isolated to the event handler itself. In compositions, we sometimes need to look at what happened immediately before (within the same tick) to find out what we should do.
- In atomic event handlers, we handle the event at the time that the event occurs. In compositions, we sometimes have to defer the handling until a series of events have executed and then do what we need to do after.
- In atomic event handlers, it makes sense to allow the events to flow through the plugin chain and let developers hijack (in a sense) the events to provide their own handling. In compositions, this doesn't work because hooking into a composition related event might work in one browser but break in another so it becomes an unreliable way to customize. I think there are reliable ways to do this though (perhaps through the commands @ianstormtaylor mentioned).

These differences in my opinion make it worthwhile for us to consider splitting all the composition code into a separate plugin. The atomic code feels fairly consistent browser to browser but compositions feel like they are quite different.

With respect to the concern about keeping the code in sync, my view of the code base is that there is code that is common among all compositions like the `setTextFromDomNode` and `setSelectionFromDom` (which may be renamed but you get the idea). This code can and should be used across all compositions (once we get to it).

There are two areas where compositions seem to differ between browsers:

- When to reconcile
- Fixing broken dom

In both these cases, there would be little code overlap between browsers (at least Android and other browsers) so we wouldn't get much benefit by putting all the code in one place. But I think this isn't an issue for the benefits of code sharing anyways because the when/fixing code would call out to the code that would be shared between browsers. Basically, what will happen is if we change Slate away from Immutable, for example, to pure JSON objects and Immer, we just need to fix the code in `setTextFromDomNode` and friends and the "when" and "broken DOM" code would not need to be changed. We can survive major breaking change Slate version upgrades without having to touch the composition plugin code.

Note that the `setTextFromDomNode` code is already shared with the `after` plugin so on a breaking code change, this would be included in the fix to the after `input` handler. If the `input` worked, the compositions should work as well.

I also feel like, and have been thinking of, separating composition logic even further. One plugin for each browser. Or at least set of browser with different composition behavior (for example, all webkit might be able to go together). This should improve performance and make the code more readable. Right now when debugging Android API 28 code, you have to sift through the rest of the code and see what applies which slows working on the code. From the performance perspective, there is a lot of if/then/elseif/switch/case in the code at runtime for every event call. If each platform had a different plugin, there would only be one switch statement once at initialization time.

Slapbox:
Anything a lowly contributor such as myself can do to help this PR along? I know it's a substantial PR and needs proper review, but mobile support is so damn exciting. 

@thesunny - would you be open to merging your PR into the `master` branch of your Slate repo? Yarn can't handle references to branches on Github when the package is a monorepo, but I'd love to give this some testing in my app.

thesunny:
@Slapbox 

I was considering releasing a version on NPM. Maybe under `slate-for-android` until it gets merged.

Slapbox:
image

thesunny:
@slapbox @steida @pradel @gracicot

Just a heads up that I published under these two NPM packages:

- slate-react-android
- slate-dev-environment-android

Unfortunately, because of `slate-dev-environment-android`, you'll have to alias the package in webpack, babel or whatever you are using.

- For babel, this might work https://www.npmjs.com/package/babel-plugin-module-resolver
- For webpack, here's some documentation https://webpack.js.org/configuration/resolve/

Let me know how it goes!

Slapbox:
Remaining issues I've found so far:

1. [Backspace timing issue](https://imgur.com/a/aQbj7Ko) - If you don't wait until the keyboard notices it's at a word boundary and loads its suggestions before continuing to `Backspace` into a new word it will insert a fragment of a word at the cursor's location
2. Selecting a word and using `Backspace` only removes the last character
3. Copying a series of words only copies the first word
4. Cutting a word or series of words does neither copies the words into clipboard nor removes the words from the editor

**All the above results from Pixel 2 running Android 9 | Chrome 72 | GBoard**

The weird thing is that I don't recall these being issues in January (At the very least I don't recalls Points 1 and 4 functioning that way.)

So the question is, is it changes to Slate, changes to @thesunny's method of supporting Android, changes to Chrome, or an oversight by us when testing previously? I'm leaning heavily against this being an oversight since I tried hard to take a wrecking ball to @thesunny's work.

gracicot:
I don't know if this is my fault or the patch, but I have an error when using it on a desktop browser:

error screenshot

The error appears each time I type a character.

Slapbox:
@gracicot you definitely have `slate-dev-environment` resolving to `slate-dev-environment-android`? When I didn't have this I was seeing `null` for `focus.path` and `anchor.path`

gracicot:
@Slapbox I did some research and it still happen with the master branch. I'll try to find out why. Thanks.

ianstormtaylor:
Hey @thesunny, sorry for the delay on this. And thank you again for putting in the work to make this happen, this is really awesome.

Summary: It looks good to me, feel free to merge!

---

Longer version: it's hard for me—with no Android experience—to fully "approve" this pull request. But I don't think that's necessary for now, and am happy to have it merged in to live alongside the codebase. The only caveat is that I can't always guarantee that changes I make while Slate is in beta will happen to be fully working across Android. And because of that I'll need you (and others's) help to keep it up to date. I hope that makes sense.

But I'm happy to have this merged in. And over time, I think we can work to make it more integrated with the composition handling across other environments, so that we move away from having a separate code path just for Android. Which will hopefully make it easier to maintain over time for everyone. (Although I know this won't be easy, since we'll have to come up with new abstractions to stay sane while working on it 😄)

Anyways, feel free to merge! And thank you very much.

thesunny:
Thanks Ian. I've merged master into the PR and there were some unexpected conflicts. I don't think that I missed any weird issues but could a few people please do a quick check to make sure things are working as expected:

https://thesunny.github.io/slate/#/composition/split-join

Note that the "Special" sub-example at this URL may fail but the others should work. I've included them so that I won't forget them the next time I revisit Android which will be some time after I come back from vacation (leaving on March 13 and back on March 20).

AndyBitz:
@thesunny I've tested the examples on my Pixel 3, "Split/Join" works, as does "Insertition". First example under "Special" with "it is" works too, the others do not.

Also, selecting multiple words and then hitting the "bold" or "italic" button will apply the changes only to one word. I guess this is related to @Slapbox findings.

On a side note, I've been playing around with contenteditable as well and I came up with this pattern which uses `MutationObserver`, I don't know if this can be applied here or if it would be useful in this case, but feel free to take a look: https://gist.github.com/AndyBitz/aac539f17f81dfe976904fce060d6de7.

And thank you for all your work, this is pretty amazing  🙏

thesunny:
@AndyBitz Thank so much for checking the PR.

I took a look at the gist you linked to but I have to admit I'm not sure what you're trying to show me.

I did try to use Mutation observer to revert the DOM to a known safe state but unfortunately it didn't work. At least not all the time. The solution that worked was to save references to the original DOM nodes (before the changes) and re-arrange and re-populate them to their original positions and values.

thesunny:
Yay! Android support is merged!

AndyBitz:
@thesunny No worries, I thought the composition event was the main reason why it didn't work on Android and tried to make it work this way, but I guess I oversimplified the solution.

Anyway, thanks again for making this possible!

PR #2568, opened 2019-01-30
Add @convertkit/slate-code to docs

brendancarney:
Adding plugin to docs

`slate-code` adds code block editing support to an editor. It is maintained by ConvertKit, and is relied upon for their upcoming new email editor release, and can be expected to be maintained. We understand potential future incoming changes to Slate, and intend to update and support this plugin through them. We hope to be adding more soon 😃


ianstormtaylor:
Thank you @brendancarney, that is awesome to hear!

PR #2569, opened 2019-01-30
Fix benchmarks that use value.change

jtadmor:

Fixing a _bug_ (with the benchmarks)

#### What's the new behavior?

All existing benchmarks except for `changes/insert-node-by-key` will actually run now.

The benchmarks were using the old `value.change()` syntax, simply import `Editor` from `slate` and use that to make changes.

Considered making that part of the benchmark runner itself, but this was pretty easy and is more straightforward for new people coming into the codebase.

Fixes: #
Reviewers: @


benchmark/slate-react/index.js
```diff
@@ -29,7 +29,7 @@ module.exports.run = function(include) {
       const dir = resolve(benchmarkDir, benchmark)
       const module = require(dir)
       const fn = module.default
-      bench.input(() => module.input)
```
jtadmor: Minor change here just to make the benchmarks in `slate-react` line up better with `slate`.

The `bench.input` code already accepts either a function or a plain value, so this line was just breaking when any benchmark had `module.exports.input = () => { ... }`

CameronAckermanSEL:
LGTM

benchmark/slate/changes/normalize.js
```diff
@@ -35,5 +36,5 @@ const value = (
 )
 
 module.exports.input = function() {
-  return value.change()
+  return new Editor({ value }, { normalize: false })
```
CameronAckermanSEL: what's the lifetime of this normalize: false flag? does it expire when the value is set to the editor?

jtadmor: It is only used in the initial `setValue` call. I figured it might defeat the purpose of the benchmark if we were already per-normalizing the value before the initial run.

https://github.com/ianstormtaylor/slate/blob/f677f7655928ee6a848f997e4f3361d4fb1b5705/packages/slate/src/controllers/editor.js#L67

ianstormtaylor:
Looks great, thank you @jtadmor!

PR #2587, opened 2019-02-07
Fix insert-node-by-key benchmark, add normalize and insert-fragment benchmarks

jtadmor:

Improving _benchmarks_

#### What's the new behavior?

(See PR title)

`insert-node-by-key` wasn't working because the index passed into all but the first couple operations were invalid.

`normalize-with-problems` should check how long it takes to normalize a Slate tree that has some problems (`block` with no `text` children, `inline` not surrounded by `text`). This could potentially be improved by adding other types of things that violate the core schema, but those are two issues that we've noticed causing a lot of issues in our own app.

Fixes: #
Reviewers: @

### DIFF
--- a/benchmark/helpers/h.js
+++ b/benchmark/helpers/h.js
@@ -16,6 +16,12 @@ const h = createHyperscript({
     list: 'list',
     item: 'item',
     image: 'image',
+    table: 'table',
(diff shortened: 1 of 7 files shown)


ianstormtaylor:
Thanks @jtadmor!

Issue #2575, opened 2019-02-01
On simple inline copy-then-paste, `insertFragmentAtRange` inserts as a nested block instead of inline.


bebbi:
#### Do you want to request a _feature_ or report a _bug_?

bug
#### What's the current behavior?

A simple inline copy-then-paste gets pasted as separate blocks.

![copy_paste](https://user-images.githubusercontent.com/2752391/52136941-996c7200-2649-11e9-8aa0-f843b0d85a57.gif)

1. open fiddle: https://jsfiddle.net/kj2vspaq/
(1a. for verifying, a more nested fiddle: https://jsfiddle.net/mat306x4/1/)
2. in the second paragraph, select any letter in the middle, press copy (Cmd-C), then paste (Cmd-V)
3. See the paragraph fall apart into multiple blocks.

Two issues:
- The fragment isn't inserted inline
- The selection jumps to the end.

#### What's the expected behavior?

Step 3 should result in no visual difference from before, including the selection endpoint.

bebbi:
So this is related to the functionality introduced in https://github.com/ianstormtaylor/slate/pull/2273.

@steobrien thanks for the great explanation over there. I'd like to relax your check

    hasSingleNode(documentInner)

[here](https://github.com/ianstormtaylor/slate/blob/8e659dbd28cf88ca7659460417b31e0e558c2488/packages/slate/src/commands/at-range.js#L827)

because I think, instead of allowing the melding only in the subset of cases where `documentInner` has a single node, when we have multiple nodes there we can instead check if we *find a path* that matches the fragment's path and if found, return that as the insertion node.

I'm going to see if I can PR this but would love your quick feedback.

steobrien:
Hey @bebbi, that sounds like a good idea. Good luck with the PR!

bebbi:
#### On this PR

I'm going to propose a PR that changes the `findInsertionNode` logic (only).
Originally, I just wanted to allow melding pastes when the document has a non-trivial structure per above. Then I understood that this doesn't address many deeply nested paste issues, so I ended up with a larger change proposal to start addressing nested pastes more generally. Here's the reasoning behind the change:

In relatively 'flat' lists of blocks, the current paste approach works fine, but in deeply nested scenarios (nodes are mostly part of other nodes) and very different block depths throughout the document, it works less well.

[...]

PR #2589, opened 2019-02-09
Address fragment pasting in nested documents in a more general manner.

bebbi:

This addresses #2575 .

#### What's the new behavior?

Same behaviour for examples and tests.

The proposed change address fragment pasting in nested documents in a more general manner.

Here is an explanation to visualize this change:
Think of the leaves of the document as the document's surface, and think of the paste fragment as a wedge. The objective is to drive the wedge as deeply into the surface as reasonable to maximize smoothness of the resulting document surface.

The constraints from the previous implementation still apply: we can only go so deep that the ancestor node type at each ancestor level of the wedge matches the document's ancestor node type at some corresponding depth level.

To achieve this, we make our "wedge depth" calculation relative to document surface instead of document root.

Unlike the current document root based calculation, this ensures we won't be trying to match our wedge against nodes too deep below the surface level when we paste a shallow wedge into a deep part of the doc.

The new approach solves the issue's fiddle; it generally gives copying from shallow trees into deep trees a reasonable default behaviour that I think should be more accurate than the previous.

#### An example:

In our use-cases, we have deeply nested trees that have `container` type nodes which in turn contain blocks of leaves, and more `container`s, similar to what's in the fiddle.
But the example will looks similar if we have `table`s nested inside `table`s or other type of repeating structures.

Now if our document selection has the following ancestor structure:

    container
      blockWithTextLeaf
      container
        blockWithTextLeaf
          <cursor>

and we're trying to copy-paste a word from a paste fragment that looks like:

    container
      blockWithTextLeaf
        <word to paste>

Then the new approach lets us match the fragment container with the leaf-most container as opposed to the root-most container which wouldn't be helpful.

See https://github.com/ianstormtaylor/slate/issues/2575#issuecomment-462027027

Fixes: #2575
Reviewers: @ianstormtaylor


jtadmor:
We've had to implement something similar like this for lists / tables, I can try to see if this code would prevent the need for custom code to handle those use cases.

steobrien:
I haven't got time to closely examine and think think through the implications here, but from a quick read this looks great.

However, it would be great to see some new fixtures added to demonstrate – and lock in – the improvement. In my experience for problems like these, the fixtures are more long-term valuable than the code itself.

Do you think you could isolate some examples, ideally based on the real-world problems you encountered?

bebbi:
@steobrien yes I was thinking of getting some feedback and then add a test based on the solved fiddle, agree that this is important.

packages/slate/src/commands/at-range.js
```diff
@@ -807,36 +807,86 @@ Commands.insertFragmentAtRange = (editor, range, fragment) => {
...
+ * @param {string} documentKey
+ * @param {Document} fragment
+ * @return {Node}
+ */
 
-  if (hasSingleNode(fragment)) {
-    let fragmentInner = firstNode(fragment)
+const findInsertionNode = (fragment, document, documentKey) => {
+  // Find the deepest block in a doc with no siblings.
+  const deepestSingleBlock = doc => {
```
jtadmor: Note that this will return `doc` even if doc has multiple child nodes. Is that intended?

packages/slate/src/commands/at-range.js
```diff
@@ -807,36 +807,86 @@ Commands.insertFragmentAtRange = (editor, range, fragment) => {
...
+    )
+  }
 
-      if (fragmentInner.type === documentInner.type) {
-        node = fragmentInner
-      } else {
-        break
+  // Given two reverse lists of ancestors, check if all fragment ancestor types
+  // match the doc ancestors at some position.
+  const findMatchIndex = (documentAncestors, fragmentAncestors) => {
```
jtadmor: Since this returns a block + children, maybe call it `findMatchAncestor` / `findMatchFragment` or some such?

packages/slate/src/commands/at-range.js
```diff
@@ -807,36 +807,86 @@ Commands.insertFragmentAtRange = (editor, range, fragment) => {
...
+    for (let fragIdx = 0; fragIdx < fragmentAncestors.size; fragIdx++) {
+      // The docIdx loop relaxes our check in that we can still match if there
+      // are node type differences leaf-side.
+      // This is important for example if our fragment inserts multiple siblings
+      // or inserts another type while the tree structure remains the same.
+      for (
+        let docIdx = 0;
+        docIdx <= documentAncestors.size - fragmentAncestors.size;
```
jtadmor: What about the case of `fragmentAncestors.size` > `docmentAncestors.size`? If we are never getting a match in this case (which appears to be the case), we can at least return quickly.

jtadmor:
Made a couple comments about potential edge cases that may need to be addressed via code and / or tests. Would generally agree (per your issue) that it would be nice if a lot of this logic could be extracted into internal utils or even some exported utils that plugin-developers could rely on.

bebbi:
thanks for review @jtadmor !
Actually my thinking was that this PR can serve as a pretty generic implementation until someone comes up with a breaking example. It certainly covers the existing and new tests more generally than before (and at all).

By its nature and like in the previous code, it can't nicely solve some cases of pasting from deep to shallow because if the fragment is deeper than the doc, we may lack some user-land information on deciding about how planar the resulting document surface should be.
I think I have an approach to this, as well as for the general case that has no repeating ancestor types; we could leave this issue open beyond accepting the PR to discuss that second part or create a new one.

packages/slate/src/commands/at-range.js
jtadmor: Agreed, just wanted to make sure that was the functionality you had in mind.

Also, since the loop will never find anything, we can probably do a check at the top, before upper `for` loop, so that we can return fast if `fragmentAncestors.size > documentAncestors.size`.

jtadmor:
I like the idea of getting this in sooner if tests look good and then working on more complex solutions / utils in a later PR.

packages/slate/src/commands/at-range.js
bebbi: ah, got you

ianstormtaylor:
Hey @bebbi, thanks this sounds awesome!

I agree with @steobrien that us having good test fixtures is the way that this kind of behavior is going to be solved long term. If I were to merge this, I'd like to be able to easily see some of the common use cases for this kind of pasting in our test fixtures, otherwise it's hard for me to tell what is changed.

Off the top of my head, for a list of "structures" that we should try to support out of the box with reasonable behavior if possible:

- Individual words/phrases
- Flat paragraphs
- Nested quotes
- Nested lists
- Nested tables
- Nested "complex" blocks

We probably already have fixtures for some of these. But I think to merge this I'd like to see someone write up a quick summary of what we have now, what we'd need to be at a comfortable level, and then have the PR that we make include those extra fixtures.

bebbi:
Rebasing this branch onto newest master.
This allows us to add in more pasting tests in this branch that would throw errors on previous base,
and to close #2604 and #2605.

Still todo: review the test cases above, do we have all these cases covered.

jtadmor:
Hi @ianstormtaylor @bebbi @steobrien,

I re-arranged the existing `insert-fragment` into multiple folders. I renamed some of the existing tests, removed duplicate ones, and added a new one. I think this should better indicate the intent of the various tests and give us a decent picture of the current behavior.

I created `tables` and `lists` folders although strictly speaking those are just examples of `nested-block` pastes. They do give some clearer picture of how some existing behavior may not be ideal. I've added "TODO" comments to indicate where I personally think it is not.

One thing I would note is that Slate may not be able to handle all scenarios correctly. For example, unwrapping a `quote` block and merging the text into a `paragraph` may be reasonable, but doing the same behavior into a table would likely not be.

I believe that suggests that while this PR is certainly an improvement, the final state will almost certainly require that Slate either provide a boolean param for `insertFragment`, provide additional commands, or at least provide some utils for decomposing a fragment and separately inserting / merging leaf nodes,

steobrien:
Nice work! The diff is hard to read here, maybe this needs a rebase/merge?

jtadmor:
@bebbi @steobrien @ianstormtaylor Force pushed a rebased version of this, diff should be cleaner now

bebbi:
Thanks, this looks like a solid improvement to me.

steobrien:
This is looking good – great to see some new test cases! Are you planning to solve the TODOs before merging or is your intention for them to live in the codebase?

jtadmor:
I wouldn't necessarily feel right "solving" them as part of this PR. That is how functionality works today for certain basic pastes, but I would want some more input if others similarly feel strongly that some of it is wrong.

Each one can then be solved one at a time, if the community is cool with it.

ianstormtaylor:
Hey @bebbi and @jtadmor, thank you for this!

I've been hesitating to merge it, because of the huge changes coming in , but I don't want to lose this work. Especially appreciative of the new tests! So I'm going to merge this now, but I can't guarantee the behavior stays exactly as indicated in this PR now for the next few iterations of Slate. We might need to revisit this at some point, and really come to a good decision about how fragment insertion should work by looking at other editors. I might have to break fragment insertion just to get over the barrier of shipping some of these bigger changes. I hope that makes sense. I really appreciate the work you two have put in here.

PR #2592, opened 2019-02-11
Fix error when pasting an image into Slate image example.

eroepken:
This pull request is to fix a bug in the Slate image example code.

The following error occurs on https://www.slatejs.org/#/images when attempting to paste an image URL and the image isn't actually added due to the function not returning properly:
`Uncaught TypeError: String.prototype.endsWith called on null or undefined`

I'm aware that there is a separate image pasting plugin, however. Feel free to kill this PR if it makes more sense to use that instead.

#### What's the new behavior?
This patch re-enables the pasting of images into the editor without using the previously mentioned plugin.

It strips query tags from the end of the URL, then grabs the URL extension and compares it to the given image extension list.

I couldn't find a bug that was reported for this.
Reviewers: @ianstormtaylor

### DIFF
--- a/examples/images/index.js
+++ b/examples/images/index.js
@@ -29,7 +29,7 @@ const Image = styled('img')`
   box-shadow: ${props => (props.selected ? '0 0 0 2px blue;' : 'none')};
 `
 
-/*
+/**
  * A function to determine whether a URL has an image extension.
  *
  * @param {String} url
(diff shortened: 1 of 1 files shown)


examples/images/index.js
```diff
@@ -37,7 +37,14 @@ const Image = styled('img')`
  */
 
 function isImage(url) {
-  return !!imageExtensions.find(url.endsWith)
+  return !!imageExtensions.find(
+    element =>
+      element ===
```
brendancarney: Hi @eroepken, what do you think about wrapping this in a function, like `getExtension(url)`?

ianstormtaylor:
Good call here. Could you use the `URL` API instead to ensure that we're doing the parsing/splitting correctly? Using it to get the proper path of the URL and then just checking that instead?

examples/images/index.js
eroepken: I can definitely do that.

eroepken:
> Good call here. Could you use the `URL` API instead to ensure that we're doing the parsing/splitting correctly? Using it to get the proper path of the URL and then just checking that instead?

I will def look into that.

Thank you all for the feedback! I will update this ASAP.

eroepken:
Updated. Please let me know if anything else here should be changed. Thanks again everyone!

brendancarney:
Looks great @eroepken, thanks for this contribution!

Issue #2521, opened 2019-01-03
bug: `normalizeNode` is not called on all changed nodes.


bebbi:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

The `normalizeNode` function does not check all relevant nodes on editor changes in trees, and the editor state degrades.

##### Editor structure to normalize:
This is a simplified example of our use-case which requires the `normalizeNode` function.
The allowed editor structure is a tree: a `container` node (a node of `{ type: 'container' }`) which contains a `paragraph` as a first child node, and optionally more `container`s as subsequent children nodes.

This structure gets broken by edit operations because `normalizeNode` is not called on all changed nodes.

##### Reproduction:

(macos 10.14.2/Chrome 71/0.44.9)

The repro annotates the editor value with keys in `node.data` for easier reading.

1. Open https://jsfiddle.net/gLk4dh29/1/
2. Open the dev console
3. Put the cursor at the start of the second line that ends with "para 2"
4. Hit Backspace and see the first 2 lines merge.
5. Move the cursor to the last line and enter any character.
6. See the console output from the `normalizeNode` function which found an illegal structure but is called too late.

#### What's the expected behavior?

Right at step 4 of the repro, `normalizeNode` should be called with the node `container 2` because the structure of that node has changed.

CameronAckermanSEL:
I think the root cause of this issue is that the `getDirtyPaths` function that is called in `applyOperation` doesn't grab the correct path for where the node used to be when it is merged out of a parent node.

https://github.com/ianstormtaylor/slate/blob/76a88a649a9ec7d27d207416d94b3de1de0828d6/packages/slate/src/controllers/editor.js#L520

At least I think so, I'm having some trouble groking this (haven't had coffee yet ;))

CameronAckermanSEL:
I think we might be missing a few dirty nodes here:

1. Like in move node, we probably need to get the previous set of ancestors in conjunction with the new ancestors.
2. To address the previous rule, we need to also get the sibling of the previous path.

Issue #2574, opened 2019-02-01
bug: `normalizeNodeByPath` throws on delete.


bebbi:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

If deleting nodes, slate `normalizeNodeByPath` encounters an illegal path and throws.
This is similar to #2528 and using the same fiddle, but looks like not a regression of #2536.

1. Open https://jsfiddle.net/y5q9120p/2/
2. Open the dev console
3. Select all (e.g. Cmd+A)
4. Press Delete
5. See "Uncaught Error: Node.assertNode could not find node with path or key: List [ 0, 0, 0, 0 ]"

I checked whether it's a regression [here](https://jsfiddle.net/dx54Lcve/2/), this is 0.44.9 but fails too:
cc @Calyhre 

I additionally get that error a lot on undo after a working delete (also like in #2528), but that may be the same issue.

#### What's the expected behavior?

It should just work fine.

isubasti:
i believe this is fixed on #2593

bebbi:
fixed by #2593

PR #2593, opened 2019-02-12
Fix move_node dirtyPath and PathUtils.transform calculations

jtadmor:

Fixing a _bug_

#### What's the new behavior?

`getDirtyPaths` does a better job of handling some `move_node` operation edge cases.

`PathUtils.transform` also does a better job of handling a `move_node` operation edge case where the `path` we are transforming is the path that was moved.

Lastly, the `element.moveNode` itself also uses the same logic to determine where to actually insert the node, ensuring that it is being put in the correct place in the hierarchy.

Previously there was code to handle certain cases, prefaced by a `// HACK` comment.

This generalizes the previous approach by running all ancestors of `path` and `newPath` through `PathUtils.transform` to get the true new path that results after the move operation, rather than potentially non-existent (or incorrect) paths.

A test is added that confirms the correct behavior for the edge case discovered (where `newPath` is the parent of `path`).

While doing research, I also discovered that `PathUtils.transform` did not use the correct check for `move_node` in a few specific cases.

Critically, the only real edge case is when the path being transformed is either the `operation.path` or a descendant of it. In that case, the only circumstance where we aren't simply moving the path to its new location is when the `operation.path` is a left-sibling of an ancestor of the `operation.newPath`.

Thus, a new check is added for `isYounger(path, newPath) && path.size < newPath.size`, not `isBefore(path, newPath)`.

The rest of the logic for that method is simplified by simply running it through the two actions that make a move: the removal of the node at one path and the insertion of that node at the new path.

Tests were added to confirm this behavior is correct, you can see the sample paths I used in those tests (I had to draw out some trees on paper).

Fixes: #2521, #2574
Reviewers: @


jtadmor:
Not 100% sure the last test I added (`delete/all-with-nested-blocks`) is correct, but it does not error out.

ianstormtaylor:
Hey @jtadmor thanks for tackling this.

It's hard for me to tell exactly what these changes are doing, and whether or not they introduce any new bugs. What exact cases are broken? And why the need to switch the way the logic works in places to do more convoluted things with operations?

bebbi:
@ianstormtaylor 

> What exact cases are broken?

#2521, #2574, #2604 with #2605.

It would be helpful to understand what to contribute beyond those.

> It's hard for me to tell exactly what these changes are doing

Would more inline comments and/or more high level context in the pull solve this?

> whether or not they introduce any new bugs

I'll try to qualify this PR from our experience: When working with highly nested trees, slate was unusable to our client due to lots of `Uncaught Error: Node.assertNode could not find node with path or key: List [...]` occurring on both undo and paste. Yes, they suspended our project because of this.

Since merging this they haven't encountered a single error so far. It's one of the most important stability fixes I have seen in slate.
Formally, I'd think addressing all current tests plus the new issues' tests appears to answer this. Do you feel there's a gap of tests around core functionality here that needs to be addressed?

Side-note: the reason #2605 is a separate PR with only tests in it is that while it belongs to another pull #2589, I couldn't add those tests to #2589 because before this fix here they would just error out with the same error above - so #2605 tests depends on both this fix and #2589.

jtadmor:
Hi ian,

Thanks for taking a look.

I know some of the tests are a bit hard to read and immediately understand what was broken. It's also a bit harder to understand because several different things were broken.

1. `getDirtyPaths` not picking up all paths on `move_node` operations, depending on how `path` and `newPath` related.
- Test:  Jump to `with-plugin-normalization.js`)
- Code: Jump to `editor/getDirtyPaths`

2. For multi-op commands including `move_node`, previous dirty paths were not being transformed correctly. During normalization, this would throw errors.
- Tests: User facing bug: `delete/all-with-nested-blocks`.
- Code: `PathUtils.transform`.

3. When inverting a `move_node` op, the inverted paths were not always correctly calculated. This would throw errors during undo.
- Tests: `undo/insert-fragment` for the user facing bug.
- Code: `invert/invertOperation`.

4. `element.moveNode` was not always accurate, and would sometimes move a node to the wrong place. This was also breaking when trying to undo certain operations.
- Tests: `undo/insert-fragment` for user facing bug.
- Code: `element/moveNode`


What I determined reading through the code is that there were many different places where we were trying to deal with various cases in `move_node`. By re-using `PathUtils.transform` (which was written after a lot of the other code), we can handle these cases more consistently. I also added a variety of tests (and fixes) to `PathUtils.transform` to ensure we were capturing the correct behavior for some of the trickier cases (see `from-before-not-younger`, `from-before-younger`, `siblings-backward`, `siblings-forward`).

The alternative is to copy all the logic from `PathUtils.transform` into `getDirtyPaths`, `moveNode`, `invertOperation`. My view was that this made the code harder to read and understand, and harder to maintain going forward. It seems that simply deferring to `PathUtils.transform` is more self-documenting: what we want is a reference to where a node in the tree is based on what its path was before a `move_node` op.

ianstormtaylor:
@jtadmor thanks, that's very helpful. I agree with you now about having it more self-documenting by reusing `PathUtils.transform`. Thank you!

@bebbi that makes total sense. Once I merge this you can merge those two PRs into one it sounds like. Thank you!

Issue #2609, opened 2019-02-21
discussion: add keyboard control to the mentions example and / or make it a plugin


ericedem:
#### _idea_

When I implemented the mentions example, the goal was to highlight how to interact with Slate around capturing mentions, inserting void nodes, and displaying the suggestions as a floating menu. An additional goal was to try and do this in as simple a fashion as possible, which means putting all the logic in the parent component instead of using a Slate plugin.

This is not how we have it implemented in production at [cake.co](https://www.cake.co). Because there is so much additional logic, in large part due to keyboard control, we have it implemented as a plugin. Implementing as a plugin also adds quite a bit of complexity and conceptual overhead, which is something I wanted to avoid in the slate plugin.

I have received some requests to either add support for keyboard support in the plugin, which will add quite a bit of complexity, or to open source the plugin itself. I'd love to open source it, but I don't know if I have the time or energy to maintain such a project right now, so I'm worried that it will become defunct like the plethora of other plugins (this is why I added it as an example in the first place).

I'm open to updating the example to be a bit more fully featured, and converting the example to a plugin. Here are the pros and cons I see:

### Pros
1. **Drop in solution** This is probably the most compelling solution. While it would be hard to understand, it would be very easy to just copy paste and use as is, even in an already complicated application.
2. **It's already pretty complex** It's already a pretty complicated solution, so maybe it would be more worthwhile just to go all the way so the people have the right framework to work from.

### Cons
1. **Ease of understanding** Right now the example is already fairly complicated, but it is easy enough to understand. If we make it a plugin, now users will need to understand what a plugin is (which are already somewhat complex) before they can even begin to understand how to implement mentions.
2. **Maintainability** There are a lot of edge cases to consider with plugins, and I'm worried that it will become pretty difficult to maintain by the community. This is probably my biggest concern.

Anyways, this will take a little bit of work for me to do, so I'd like some feedback from the community about whether this is the right thing to do before jumping into it.

oyeanuj:
@ericedem Firstly, thank you for contributing that mentions example, and for starting this discussion. I think mentions is one functionality that there isn't a solid plugin developed in user-space - hence the pressure on you :)

I imagine outside of fleshing out the examples to include the plugin logic, a couple of other options that might work are:

1. Create a plugin but put a notice for co-maintainers who can keep it running so that you don't have to commit all of your energy.

[...]

PR #2610, opened 2019-02-22
examples(mentions): convert to plugin and handle keyboard events

ericedem:

Improving an _example_.

#### What's the new behavior?

Update the mentions example to be a plugin and to handle keyboard events.

Here is an illustration of the new keyboard control behavior for the suggestions:

kapture 2019-02-22 at 12 56 21

Here is an overview of the new components added to the example:

```
- MentionPlugin is the core plugin that defines how mentions should be handled
  in slate. It is designed in such a way that you can have multiple instances of
  it in the same editor, like for user mentions and hashtags.

- UserMentionPlugin is the higher level plugin that defines the business logic
  for the mentions implementation like schema restrictions and how the mentions
  themselves should be rendered. It also provides the `<UserMentionSuggestions>`
  component.

- <UserMentionSuggestions> encapsulates business logic around searching for
  suggestions and for handling selection, both with mouse and keybaord.
```

Fixes: #2609

cc: @oyeanuj


oyeanuj:
@ericedem Apologies that I didn't see this earlier, something with Github notification settings. 

I did a quick parse-thru and this looks great! Thank you for documenting as you went along. I think this is ready to merge for an example.

examples/mentions/UserMentionSuggestions.js
```diff
@@ -0,0 +1,242 @@
...
+              {user.username}
+            </Suggestion>
+          )
+        })}
+      </SuggestionList>,
+      root
+    )
+  }
+
+  renderSuggestions() {}
```
realknack: It looks like this serves no purpose.

almaron:
There's a serious issue calling `setState` from `componentDidUpdate`

colinmcd94:
@ericedem amazing work, thanks for doing this. I'd love to see this get properly merged into master

ianstormtaylor:
Hey @ericedem thank you very much for all of your efforts with Slate. I really appreciate it.

That said, my goal for the examples is for each one to be very self-contained and simple. I don't think it makes sense for us to make the mentions example any more complex (I actually would like to simplify it further). If anyone wants to maintain a full-fledged mentions plugin or example they are more than welcome to. But I can't justify adding the maintenance burden to this repository for example purposes. I hope you understand. I'm sorry I didn't get to this earlier so that I could have saved you some effort.

ericedem:
@ianstormtaylor no worries, I've been on the fence about adding them myself. It wasn't very much work, a lot of this was just copy 🍝 .

alainib:
hello. i don't found MentionPlugin  in doc. 
how to use it please ?

PR #2615, opened 2019-02-25
Add @convertkit/slate-lists to plugins page

brendancarney:
Adding plugin to docs

`slate-lists` adds list editing support to an editor. It is maintained by ConvertKit. We rely on it for our new email editor, as well as our existing landing page editor, both are used/will be used by thousands of users.


ianstormtaylor:
Awesome, thanks @brendancarney!

PR #2618, opened 2019-02-27
Add API version and Input Events Level 2 for Android

thesunny:

Adding a feature and fixing a bug. The feature is the export of `ANDROID_API_VERSION` from `slate-dev-environment`. The bug is that `HAS_INPUT_EVENTS_LEVEL_2` returns false for Android 9 but should return true.

#### What's the new behavior?

Returns true for `HAS_INPUT_EVENTS_LEVEL_2` which results in Slate using the real `beforeinput` event instead of React's synthetic version.

Queries the browser's `window.navigator.userAgent` to derive the API version. When setting `HAS_INPUT_EVENTS_LEVEL_2` sets it to true if the `ANDROID_API_VERSION` is 28 even though it otherwise is not identified as supporting the feature.

### DIFF
--- a/packages/slate-dev-environment/src/index.js
+++ b/packages/slate-dev-environment/src/index.js
@@ -92,6 +92,43 @@ if (isBrowser) {
 }
 
 /**
+ * Array of regular expression matchers and their API version
+ *
+ * @type {Array}
+ */
+
+const ANDROID_API_VERSIONS = [
+  [/^9([.]0|)/, 28],
+  [/^8[.]1/, 27],
+  [/^8([.]0|)/, 26],
+  [/^7[.]1/, 25],
+  [/^7([.]0|)/, 24],
+  [/^6([.]0|)/, 23],
+  [/^5[.]1/, 22],
+  [/^5([.]0|)/, 21],
+  [/^4[.]4/, 20],
+]
+
+/**
+ * get the Android API version from the userAgent
+ *
+ * @return {number} version
+ */
+
+function getAndroidApiVersion() {
+  if (os !== 'android') return null
+  const { userAgent } = window.navigator
+  const matchData = userAgent.match(/Android\s([0-9\.]+)/)
+  if (matchData == null) return null
+  const versionString = matchData[1]
+
+  for (const [regex, version] of ANDROID_API_VERSIONS) {
+    if (versionString.match(regex)) return version
+  }
+  return null
+}
+
+/**
  * Export.
  *
  * @type {Boolean}
@@ -109,5 +146,8 @@ export const IS_IOS = os === 'ios'
 export const IS_MAC = os === 'macos'
 export const IS_WINDOWS = os === 'windows'
 
+export const ANDROID_API_VERSION = getAndroidApiVersion()
+
 export const HAS_INPUT_EVENTS_LEVEL_1 = features.includes('inputeventslevel1')
-export const HAS_INPUT_EVENTS_LEVEL_2 = features.includes('inputeventslevel2')
+export const HAS_INPUT_EVENTS_LEVEL_2 =
+  features.includes('inputeventslevel2') || ANDROID_API_VERSION === 28
--- a/packages/slate-dev-environment/src/index.js
+++ b/packages/slate-dev-environment/src/index.js
@@ -150,4 +150,5 @@ export const ANDROID_API_VERSION = getAndroidApiVersion()
 
 export const HAS_INPUT_EVENTS_LEVEL_1 = features.includes('inputeventslevel1')
 export const HAS_INPUT_EVENTS_LEVEL_2 =
-  features.includes('inputeventslevel2') || ANDROID_API_VERSION === 28
+  features.includes('inputeventslevel2') ||
+  (IS_ANDROID && (ANDROID_API_VERSION === 28 || ANDROID_API_VERSION === null))


thesunny:
I've broken this code out of the Android PR https://github.com/ianstormtaylor/slate/pull/2565 related to this issue https://github.com/ianstormtaylor/slate/issues/2062 for a few reasons:

1. It fixes a bug in Slate currently where Android doesn't get real `beforeinput` events, only the React synthetic ones so it is a standalone improvement to Slate
2. It will allow developers that want to start using/testing the Slate Android code which I've published to NPM to do so without having to do special configuration in their webpack or babel configs which will lower the barrier to entry for people to help figure out issues that only appear when integrating Android into their own editors.

packages/slate-dev-environment/src/index.js
```diff
@@ -91,6 +91,43 @@ if (isBrowser) {
...
+ * get the Android API version from the userAgent
+ *
+ * @return {number} version
+ */
+
+function getAndroidApiVersion() {
+  if (os !== 'android') return null
+  const { userAgent } = window.navigator
+  const matchData = userAgent.match(/Android\s([0-9\.]+)/)
+  if (matchData == null) return null
```
wmertens: How about, if the useragent matches Android but not the table, just assume the highest API? That way it's somewhat compatible with newer versions without sites having to upgrade their code. Older versions than Android 4 are probably not that likely to visit JS heavy sites.

CameronAckermanSEL: This is probably a good idea, but there is a trade off at play. As implemented, if there's a new android major that comes out, the plugin won't detect that it should apply, which will most certainly break functionality. 

If we default to the highest version as suggested, we have at least some chance of hiding incompatibility problems, but there will probably be some insidious issues in play. 

Partially working is probably better than not working at all.

thesunny: I agree with the concerns. That said, I recommend we keep it this way for now.

A major version number update for Android has previously always broken Slate so we don't gain much from future proofing major version numbers.

I'm tempted to lean the other way and break Android loudly for undetected version numbers by throwing an error that this version of Android isn't supported. Although I think that's not exactly straightforward  because we don't want to disable Android in the case where somebody is using Slate on Android only to view content. We don't want to throw errors in that case because displaying content through Slate on Android will work fine, it's just the editing that breaks.

Let me know if you think it's worth holding off a merge at this point. Otherwise, I'll merge in a few days.

thesunny: Forgot to say thank you for doing a code review on this PR. Very much appreciated.

CameronAckermanSEL: I'm personally comfortable with merging as is, with this review we at least have some historical documentation about the decision to loudly break if there's a major android version increment.

I know ian's busy but did we get his opinion on that?

thesunny: I’ll send him a message. He is the one that suggested moving this code into slate-dev-environment as a comment in the Android PR. He didn’t specifically approve of putting into a separate PR.

wmertens: Actually, I'm thinking of the many websites that will be semi abandoned in a few years and they will all have broken Slates, whereas maybe the future android version might have real input level 2 support.

I'd rather warn on newer versions, saying something like "Slate was not yet tested with this Android version, it might not work". You can just parse the version and see if it's higher than the current highest…

thesunny: @wmertens I'll update this PR to add INPUT_EVENTS_LEVEL_2 for non-matching Android API versions. I'm not going to do the Android warning here right now since this isn't the PR for Android support. That's a separate PR.

thesunny: Added INPUT_EVENTS_LEVEL_2 support for non-matching Android.

thesunny: @CameronAckermanSEL I've reached out to @ianstormtaylor about this PR but haven't gotten a reply yet. I'm assuming he's busy.

I asked him some questions with respect to contributing as I've noticed that I'm the only one merging PRs other than him and whether he wanted me to continue with that as well as offering to write up some documentation on rules for accepting PRs.

Since he hasn't replied yet and hasn't asked me to stop in the past despite having rejected two of my PRs (he's allowed the rest of them) I'm going to assume it's okay. He also has the option to revert them and he controls publishing to NPM so none of these updates will get into the wild without his publishing them explicitly. I'm also aware of why he rejected the last two PRs so I'm making sure that he doesn't have to say the same thing twice.

I'll let this PR sit for another day or so with the revision. If anybody sees anything wrong with the update, please let me know. I haven't tested this code against Android as I don't have Android Studio setup at my home computer (where I'm working today) but the change is pretty straightforward.

packages/slate-dev-environment/src/index.js
```diff
@@ -91,6 +91,43 @@ if (isBrowser) {
...
+
+/**
+ * get the Android API version from the userAgent
+ *
+ * @return {number} version
+ */
+
+function getAndroidApiVersion() {
+  if (os !== 'android') return null
+  const { userAgent } = window.navigator
```
wmertens: I just realized this runs on module load, and I'm wondering if this code will ever be run during SSR. In that case the use of `window` has to be guarded.

thesunny: @wmertens Nice catch! Didn't even think about that.

thesunny: @wmertens Nice catch; however, `window.navigator` will never get called because `os !== 'android'` will be true if we aren't in a browser so I think we're still okay.

wmertens: Ah yes of course - unless we start running render farms on cheap android tv sticks ;)

ianstormtaylor:
Thanks @thesunny this looks good to me!

Issue #2603, opened 2019-02-19
Nested block data is not transformed to Map (immutable.js)..


delijah:
..which gives us troubles altering the data later on.

The following example says more than a hundred words :)

```
const block = Block.create({
    // ...
    data: {
        src: 'https://...',
        image: {
            src: 'https://...',
        },
    },
    // ...
});

const newBlockData = block.data.mergeDeep({
    image: {
        src: 'new url',
    },
});

console.log('top-level src is missing now:', newBlockData.src, newBlockData);
```

A possible solution would be using `fromJS(data)` instead of `Map(data)` on the following line: https://github.com/ianstormtaylor/slate/blob/8e659dbd28cf88ca7659460417b31e0e558c2488/packages/slate/src/models/block.js#L91

Related stackoverflow:
https://stackoverflow.com/questions/34605528/losing-properties-when-using-mergedeep-of-immutable-js

ericedem:
Hmm that's a good point. Seems like the right solution to me. 👍

ianstormtaylor:
I think the issue is that we don't really want to enforce that `data` has to be immutable objects? I think this might be something we avoid, since we want to remove immutable objects anyways.

benjycui:
Ref: https://github.com/ianstormtaylor/slate/commit/b8c372df93ac94b3d0449e0c496cdf8bec9af0fc

Agree with @ianstormtaylor , enforce `data` to be immutable is not a good idea.

thesunny:
FYI, I'm suggesting we revert it as per my comments in https://github.com/ianstormtaylor/slate/issues/2684#issuecomment-482434319.

I wouldn't have merged it if I had read Ian's comment and I can't explain why I didn't see those comments before I merged.

PR #2619, opened 2019-02-27
Use fromJS instead of Map to convert data into block data

delijah:
Fixes: #2603

Switched `Map` to `fromJS` in order to convert the whole js object (also nested data structures) to immutable.js when creating a block.


thesunny:
FYI

https://github.com/ianstormtaylor/slate/issues/2684

PR #2622, opened 2019-02-27
Fix Edge positioning with newlines

statianzo:

Fixing a Bug 🐛 

Edge has a [bug][1] where `Range.prototype.toString()` will convert `\n`
into `\r\n`. The bug causes a loop when slate-react attempts to
reposition its cursor to match the native position.

Example of broken behavior
ienewlinebug

Live example:
https://codesandbox.io/s/6mpnnqkpr
Open in Edge and attempt to click between `b` and `c`. Notice that the cursor will reposition to the end.

[1]: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/10291116/

#### What's the new behavior?

This change avoids calling `Range.prototype.toString()` by cloning the
content nodes and measuring `textContent` on those instead.

`textContent` doesn't add `\r` to every `\n`, so the calculated offset will be correct

* [ ] The linter passes with `yarn lint`. (Unrelated failures)

Fixes: #
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/utils/find-point.js
+++ b/packages/slate-react/src/utils/find-point.js
@@ -51,7 +51,7 @@ function findPoint(nativeNode, nativeOffset, editor) {
     range.setStart(textNode, 0)
     range.setEnd(nearestNode, nearestOffset)
     node = textNode
-    offset = range.toString().length
+    offset = range.cloneContents().textContent.length
   } else {
     // For void nodes, the element with the offset key will be a cousin, not an
     // ancestor, so find it by going down from the nearest void parent.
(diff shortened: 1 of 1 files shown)


trentdm:
👍

ianstormtaylor:
Thanks @statianzo this look good. Can you add a `COMPAT:` comment (check the others in the codebase) above this line so that we don't forget why we do this in the future.

statianzo:
Added `COMPAT:`. Thanks for the review.

ianstormtaylor:
Beautiful, thank you @statianzo! Great PR.

PR #2624, opened 2019-02-27
fix insert fragment with void children

jtadmor:

Fixing a _bug_.

When inserting a fragment into a range where `startBlock` has void children but no text, `startBlock` would be removed.

#### What's the new behavior?

`startBlock` is not removed.

Check for void children before removing a node.

Fixes: #2545
Reviewers: @ianstormtaylor @ldavidpace


ianstormtaylor:
Good call, thanks @jtadmor!

PR #2627, opened 2019-03-04
render visible nodes only

harsh376:

Improving a _feature_.

- This used `react-virtualized` to render only the visible nodes (and a few buffer nodes)

Fixes: #
Reviewers: @

### DIFF
--- a/examples/huge-document/index.js
+++ b/examples/huge-document/index.js
@@ -64,6 +64,7 @@ class HugeDocument extends React.Component {
         defaultValue={initialValue}
         renderNode={this.renderNode}
         renderMark={this.renderMark}
+        style={{ height: 500 }}
       />
     )
   }
--- a/packages/slate-react/package.json
+++ b/packages/slate-react/package.json
@@ -20,6 +20,7 @@
     "memoize-one": "^4.0.0",
     "prop-types": "^15.5.8",
     "react-immutable-proptypes": "^2.1.0",
+    "react-virtualized": "^9.21.0",
     "selection-is-backward": "^1.0.0",
     "slate-base64-serializer": "^0.2.95",
     "slate-dev-environment": "^0.2.1",
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -13,6 +13,7 @@ import findRange from '../utils/find-range'
 import getChildrenDecorations from '../utils/get-children-decorations'
 import scrollToSelection from '../utils/scroll-to-selection'
 import removeAllRanges from '../utils/remove-all-ranges'
+import List from './list'
 
 const FIREFOX_NODE_TYPE_ACCESS_ERROR = /Permission denied to access property "nodeType"/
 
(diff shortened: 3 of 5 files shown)


alanctkc:
Thanks so much for taking this further! :smile:  Will have to pull down the code and see how the positioning plays with selection in my app.

The two main things that need to be explored, in my opinion:

1. At least in my app, a [WindowScroller](https://github.com/bvaughn/react-virtualized/blob/master/docs/WindowScroller.md) of some sort is more appropriate. Is there a way to use it or something similar to make sure `height` is not required? Dynamic heights will be a pretty common use case.
2. Selection beyond the viewport—Is there a way to make sure "select all" behaves as expected, or that the initial part of a selection is not lost if a user starts selecting and drags the selection to scroll the viewport?

I also wonder, if integration with slate ends up being simple enough, if react-virtualized can eventually be removed as a dependency and just relevant bits brought into slate.

Awesome contribution, and hoping it can all come together in a general-purpose way with free performance benefits for all.

jtadmor:
My 2 cents: I am against bringing `react-virtualized` into the `slate-react` package itself. I think the code here looks great and I'm sure a lot of people would be able to make use of it. I think having this package published as an open source extension would be great, and I think it would be great if `slate-react/component` could allow a custom render prop fn, which it seems like is all it would take to allow this behavior to happen.

Despite liking this idea, I am against bringing this into `slate-react` for 3 reasons:

1. Virtualization immediately breaks a lot of browser behavior, in particular the ability to `ctrl+f` search things in the page. We don't want to opt users into that. Scroll behavior also changes in ways that aren't ideal (jumpy, the scroll bar shrinks as you scroll down and doesn't reflect actual progress down the page).

2. `react-virtualized` is just one virtualization solution. Some people might be able to use native browser behavior (or soon-to-be-developed browser behavior) such as scroll anchoring (https://developers.google.com/web/updates/2016/04/scroll-anchoring) or be able to take advantage of React features like Suspense to prioritize certain rendering. Forcing everyone to use one particular solution is outside the scope of an editor, IMO.

3. Even if everyone wanted to use `react-virtualized`, my experience with that library is that there's a ton of fine tuning around cell measure caching, cell height, how many rows to render, etc. I think we're basically going to need to allow a pass through for every conceivable option that the library supports, in which case I don't see what the advantage is over just having users implement it themselves and not needing to look through `react-slate` code to understand the interaction.

alanctkc:
Agreed that this ultimately makes more sense as a plugin with the right hooks exposed in slate-react.

harsh376:
@jtadmor definitely agree, `react-virtualized`as a dependency in the core lib doesn't seem like the best solution. and i like the idea about the custom render prop, allows users to handle virtualization based on their needs

alanctkc:
Thinking about @jtadmor's earlier point, I wonder if shooting for solid "select all" behavior along with virtualized rendering is a pipe dream, especially considering Ctrl+F and other side effects I hadn't considered. Maybe you can't have your cake and eat it too.

An implementation that just does virtualization with dynamic height, ignoring "select all" and ctrl+f may be a good first pass. Maybe in the future, apps could opt into a halfway approach that renders the text in plain divs, but leaves out any other slate processing. This might only be useful in a use case like mine, where I rely heavily on decorations. Not sure if there's any other costly render-time slate operation where just dropping into divs would help. :man_shrugging:

jtadmor:
Yes native browser selections are another thing that wouldn't work with virtualization, I suspect it would require some interception of copy events.

ianstormtaylor:
Thanks for opening this @harsh376! I'm going to close it since it sounds like you three arrived at a conclusion of trying to integrate this as a plugin instead, which sounds great to me. Feel free to discuss and open another PR if anything needs to change in `slate-react` to make this possible!

PR #2628, opened 2019-03-04
Remove return true on key down function

matepaiva:
Correct me if I am wrong, but I believe that this `return true` is a no-op. I removed it and everything kept working as usual. And, when I was following the tutorial, that particular line created an interrogation mark in my brain. So, my suggestion is to remove that line to avoid creating entropy for no reason.

No.

#### What's the new behavior?
Just updating the docs.

Correct me if I am wrong, but I believe that this `return true` is a no-op. I removed it and everything kept working as usual. And, when I was following the tutorial, that particular line created an interrogation mark in my brain. So, my suggestion is to remove that line to avoid creating entropy for no reason.

Nope.

### DIFF
--- a/docs/walkthroughs/adding-event-handlers.md
+++ b/docs/walkthroughs/adding-event-handlers.md
@@ -83,7 +83,6 @@ class App extends React.Component {
 
     // Change the value by inserting 'and' at the cursor's position.
     editor.insertText('and')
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
You're right, thanks @matepaiva!

PR #2629, opened 2019-03-04
Fix missing break in last example

matepaiva:
In the last chunk of code, it was missing the `break` (or `return`). So, if used as is, the bold key will toggle the `code` feature as well.

A bug in the implementation example in the documentation

#### What's the new behavior?
The `b` key should only toggle the `bold` behavior. That was fixed.

Nope :)


ianstormtaylor:
Ah good catch, thanks @matepaiva!

PR #2638, opened 2019-03-08
Fix examples/tables variable "previous" is null when select text and …

dongyuanxin:
Fix a bug in `examples/tables/index.js`

#### What's the new behavior?
- When delete the choosen content in editor, the console won't print error: `Uncaught TypeError: Cannot read property 'key' of null`. (e.g.,  press `Ctrl+a` to select all and press `Backspace` to delete)
- User can key down `Ctrl + z` to recover deleted content, and no error in console

When delete all content, the varible "previous" is `null`. So I added check for null values, and returns the default behavior when it's null.


ianstormtaylor:
Thanks @dongyuanxin!

PR #2656, opened 2019-03-20
Use polyfill for Object.entries

almaron:

Google Search bot uses Chrome 41 that doesn't support
Object.entries. So we use polyfill to remedy that.

Fixes: #
Reviewers: @


ianstormtaylor:
Hey @almaron thanks for your time here. But please check out the other threads about polyfills, we don't want to add them directly in Slate's core. They should be used in userland instead, so that everyone can make their own decisions about what they need.

Issue #2474, opened 2018-12-03
Blur and refocus breaks cursor position after new line


emaciel10:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?

If the current block is an ordered list or a numbered list then blurring the editor (focusing on another window) and returning to the editor causes the cursor position to be incorrect after pressing 'Enter'

This issue can be reproduced on the rich text example at https://www.slatejs.org/#/rich-text

I am on macOS Mojave v10.14, Chrome Version 70.0.3538.110

Here is a gif of the issue:

![slate-blur-bug](https://user-images.githubusercontent.com/11020939/49404320-21530000-f71d-11e8-88d9-ba58c8daf1c0.gif)

#### What's the expected behavior?

The expected behavior is that after pressing enter, the cursor is moved to the next line along with the new block element that has been inserted.

alexfi:
Any progress on this one?

gdehmlow:
I've encountered the same issue with programmatically inserting text after blurring / focusing. You can also reproduce it on the home page by just blurring then focusing then hitting enter on any kind of text.

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #2657, opened 2019-03-21
Fix tab switch -> Enter keyDown

jtadmor:

Fixing a _bug_

See #2474 

#### What's the new behavior?

When switching tabs back to an active editor and hitting `Enter`, the cursor moves to the new line as expected.

I spent some time following the `debug` statements and found the following flow.

1. `Commands.insertBlock` gets properly called on the `Enter`. At the end of that command (including after the `moveToEnd` bit), `value.selection` is correct.

2. A `select` command is triggered with the correct `selection`. This happens after the `Enter` regardless of the tab focus.

Those two steps always happen, regardless of whether it is after a tab focus or not. But if just coming from a new tab, there is the following split:

3. **Normal Flow:** `render` cycle is triggered.

3. **From other tab:** `react-select/content` to hit `onEvent` -> `onSelect`. This hits the `After` plugin `onSelect` -> `setSelectionFromDOM` -> `editor.select`. At this last step, the `selection` is wrong (still points to original focus point).

My naive fix for this is to introduce a new `visibilitychange` event listener in the `content` component, and setting an internal flag to ignore `onSelect` events when we've just come back from a different tab.

This fixes the problem at hand, but:
- May end up having unintended consequences, I don't think we have test coverage for all scenarios
- Seems to add a fair amount of complexity (although nothing truly, we already do similar) for a pretty minor bug.

It's also possible that this bug points to a more serious misalignment of Slate / browser selection that needs to be remedied more holistically. Perhaps some other action needs to be taken on the `visibilitychange` event to get them more aligned, and it may also be a broader issue with blur / focus cycle.

Open to discussion.

Fixes: #2474
Reviewers: @

### DIFF
--- a/docs/guides/data-model.md
+++ b/docs/guides/data-model.md
@@ -83,7 +83,7 @@ Unlike the DOM, Slate offers some constraints to prevent "impossible" situations
 
 * **Documents must have block nodes as direct children.** This constraint mirrors how rich-text editors work. The top-most elements are blocks that may be split when pressing <kbd>Enter</kbd>.
 
-* **Blocks may contain either other block nodes, or inlines and text nodes.** This constraint helps you avoid boilerplate `if` statements. You can trust blocks either wrap other blocks, or contain actual content.
+* **Blocks may contain either only block nodes, or a combination of inline and text nodes.** This constraint helps you avoid boilerplate `if` statements. You can trust blocks either wrap (a) exclusively blocks, or (b) a combination of non-block nodes made up of inline and/or text nodes.
 
 * **Inlines can only contain inline or text nodes.** This constraint helps you avoid boilerplate code. When working within the context of an inline you can trust the contents do not contain blocks.
 


mydigitalself:
Hi folks. Is there any update on this PR? It's a bug we experience at Slite and would love for it to get fixed. Is there anything we can help with?

elm2178:
Is there a reason for not merging this one in? It's been open for 3 months now and I'm just wondering what the blockers are to see if there is any way I can help getting this issue resolved?

ianstormtaylor:
I'm down to merge, if @jtadmor can you fix the conflict?

My only issue is that I haven't tested it, or thought about any of the potential other things it impacts. But others can submit an issue if it breaks something for them.

jtadmor:
Hey folks. I totally botched this up, force pushed master onto it :(

Does anyone have a copy of the branch elsewhere?

Calyhre:
> Does anyone have a copy of the branch elsewhere?

@jtadmor Github kept the commit, it shows previous HEAD to be https://github.com/ianstormtaylor/slate/commit/185bacfcfea4bd3cf6104eb2e31f4ef124dec53e

Issue #2567, opened 2019-01-30
Add an `editor` prop to the `Editor` react component. 


brendancarney:
#### Do you want to request a _feature_ or report a _bug_?
Feature

#### What's the expected behavior?

Currently, you can only get the `editor` controller instance by using `ref`s. I think being able to instantiate your `Editor` controller above your component in the tree, and pass it as a prop to the editor would improve the ergonomics a bit when you need to interact with your editor outside of plugins. This can also sort of be accomplished with `renderEditor`, but at the expense of a bit of clarity. 

Example:
```
function App() {
  const editor = new SlateEditor({plugins})
 
  return (
    <Editor
         editor={editor} // could be `controller` maybe?
    />
  )
}
```

If this were implemented, we could remove ref support and cleanup the editor component code a bit. You could still technically do the same thing, you'd just have grab the editor from the props of your ref vs acting on your ref directly. 

Thoughts?

alanctkc:
Couldn't agree more. I've had multiple instances where I wished I didn't have to re-instantiate a separate editor controller outside the context of my React component hierarchy. This would be really useful for lots of scenarios (including tests).

ianstormtaylor:
For testing, you can use the non-React editor controller that is exported from Slate core. 

I’m not sure this buys us anything other than not having to use refs for sibling components. But I’m not sure that is a big enough gain. Seeing as you’d have to use refs already in normal React to do similar.

alanctkc:
@ianstormtaylor I do exactly that for testing, but it's still awkward to have to instantiate two editors (or wrap with a React component to get the same instance) if you need to pass the editor to another component or use it in an external context.

As an example, here's a test where I have a `Toolbar` that takes the editor controller as a prop. I need to render the editor into the DOM and check that on mouse click when the first line has the selection (in editor value), the popper target for the Toolbar is the first line. I was able to hack around it by passing the editor `value` (instead of the controller itself) into my Slate editor so at least the keys matched, but they aren't technically the same editor instance:

```jsx
test('sets the target to the DOM node of the first selected block', () => {
  const div = document.createElement('div');
  document.body.appendChild(div);

  const editor = makeEditor({ value: multiselected });  // instantiates an editor controller with lines selected
  const editorElement = renderEditor({ value: editor.value });  // renders a Slate Editor with the controller's value
  const wrapper = mount((
    <div>
      <Toolbar editor={editor} />
      {editorElement}
    </div>
  ), { attachTo: div });

[...]

PR #2660, opened 2019-03-23
Only call setValue when the value changes

### LINKED ISSUE #2549: setState() will block editor.insertText() before it.
A bug.

Code:
`
 class App extends React.Component {
  state = {
    value: initialValue,
	currentState:'0',
  }
  onChange = ({ value }) => {
    this.setState({ value })
  }
  onKeyDown = (event, editor, next) => {
//Insert text.
editor.insertText('11')
//Update state.
	  this.setState({currentState: '0'})

  }
}
`

It's expected to insert text "11" when I press any key. But in fact above code only show one character that I press. After remove setState, everything is OK.

brendancarney:

Fixes #2549 

#### What's the new behavior?

Re-rendering a component that holds a slate `Editor` won't trigger `setValue` unless the value changes. Previously, if you called `setState` in a parent component after calling any functions on the `editor` controller, the changes would not be reflected in the editor because the re-render sets the value back to what it was. 

I think this editor code could use a little more work, and could benefit strongly from being refactored to hooks where some of these code paths would be much more clear. I also think treating the editor controller as a prop as suggested in #2567 could make some of this code simpler and less prone to bugs. 

Fixes: #2549

### DIFF
--- a/packages/slate-react/src/components/editor.js
+++ b/packages/slate-react/src/components/editor.js
@@ -93,6 +93,19 @@ class Editor extends React.Component {
     updates: 0,
   }
 
+  constructor(props) {
+    super(props)
+
+    const { commands, placeholder, plugins, queries, schema } = props
+    this.resolveController(plugins, schema, commands, queries, placeholder)
+
+    // Set the current props on the controller.
+    const { options, value: valueFromProps } = props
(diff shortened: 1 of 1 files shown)


ianstormtaylor:

PR #2665, opened 2019-03-26
Clarify semantics of return from deserialize

tslocke:
This can trip folks up who are using compile-to-JS languages (ClojureScript in my case), where return semantics may be different.

Fixes: #
Reviewers: @


ianstormtaylor:
Thanks @tslocke! 

(I agree this is not the least confusing API. Maybe we’ll find a better solution in the future.)

Issue #2470, opened 2018-12-03
[Slate-React]: Allow customisation of placeholder style


sgreav:
#### Do you want to request a _feature_ or report a _bug_?
Feature

#### What's the current behavior?
Currently, the placeholder text for a slate editor automatically has [certain styles applied to it](https://github.com/ianstormtaylor/slate/blob/a09c3c2fdfe87605988a13e968643f81b4936978/packages/slate-react-placeholder/src/index.js#L62) inline. This can be quite tricky to override, particularly opacity, where even passing your own custom placeholder child component will inherit this attribute.

#### What's the expected behavior?
Rather than having these styles inline, it would be nice if they were associated with a class, so that they could be easily targeted and overridden by the end user, e.g. by adding
`.slatePlaceholder { opacity: 1 }` to their own stylesheet

brendancarney:
Maybe we could support a `style` or `className` option here: https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react-placeholder/src/index.js#L17-L24

brendancarney:
If @ianstormtaylor approves of this, it would be a good PR for someone new to contribute.

sgreav:
Happy to work on the PR if approved : )

brendancarney:
Hi @sgreav, I'm confident that we'd be ok to go ahead and add a `style` option that we merge with the default styles if you want to create a PR for that and I can review/merge it. Thanks!

PR #2667, opened 2019-03-26
Allow custom placeholder styles

sgreav:
_Improving a feature_

#### What's the new behavior?
_You can now specify and fully customise the style of your placeholder_
E.g. 
image

_Users can provide a_ `style` _object when creating their_ `PlaceholderPlugin` _which is applied to the wrapper div of the placeholder on render_

Fixes: #2470
Reviewers: @brendancarney


packages/slate-react-placeholder/src/index.js
```diff
@@ -21,7 +21,7 @@ function SlateReactPlaceholder(options = {}) {
     data: { key: instanceId },
   }
 
-  const { placeholder, when } = options
+  const { placeholder, when, style } = options
```
ianstormtaylor: Can you add `style = {}` here?

examples/rich-text/index.js
```diff
@@ -76,6 +77,19 @@ class RichTextExample extends React.Component {
     this.editor = editor
   }
 
+  plugins = [
+    {
+      queries: {
+        isEmpty: editor => editor.value.document.text === '',
+      },
+    },
+    PlaceholderPlugin({
```
ianstormtaylor: Can you create a new example with the custom placeholder instead of using the `rich-text` example? I'd like to keep them all single-focused.

sgreav: You bet, I’ll get on this ASAP :)

sgreav:
I think those issues should be fixed now, @ianstormtaylor  :)

ianstormtaylor:
Beautiful pull request, thank you @sgreav!

PR #2674, opened 2019-03-30
Improve syntax highlighting code, fix bugs

CodePlayer:

```bug```

#### What's the new behavior?

Copying html content into the code blocks will cause a crash.

Improve syntax highlighting code: fix bugs in pasted content, fix path bug, and improve syntax highlighting styles, support arbitrary syntax token marks.

Fixes: #2520 
Reviewers: None

### CODE REVIEW

File: examples/code-highlighting/index.js [hunk 1]
```diff
@@ -122,34 +154,15 @@ class CodeHighlighting extends React.Component {
...
-        )
-      case 'punctuation':
-        return (
-          <span {...attributes} style={{ opacity: '0.75' }}>
-            {children}
-          </span>
-        )
-      default:
-        return next()
+    if (CODE_SYNTAX_MARKS[mark.type]) {
```
euvl: This would make example much harder to understand

CodePlayer: The previous code is not suitable for others to use. In addition, the code is obvious, I also believe that most people have considerable technical level. If you still think it's not easy to understand, it is recommended to add a few lines of comments.

examples/code-highlighting/index.js
```diff
@@ -239,6 +263,55 @@ class CodeHighlighting extends React.Component {
...
+
+  /**
+   * On paste inside code blocks, insert code lines . Fix the original error of pasting html content.
+   * Note that: paste any unknown html, fragment content should not be supported in code blocks
+   *
+   * @param {Event} event
+   * @param {Editor} editor
+   * @param {Function} next
+   */
+
```
CameronAckermanSEL: I'm unsure whether we need to have a paste handler in this example. These examples are not meant to be fully featured.

examples/code-highlighting/index.js
CameronAckermanSEL: I agree that this makes it harder to understand.

examples/code-highlighting/index.js
CodePlayer: @CameronAckermanSEL 
I agree with you, the examples are not meant to be fully featured. However, it will cause a crash when pasting. Official examples may lack some features, but should not report any fatal errors in any way while the user is performing normal operations.
I still hope that the official example code is robust and worthy of reference.

examples/code-highlighting/index.js
CodePlayer: How should I put it?
I need to apply the feature of this example to my own project, but the original code is misleading. Using `switch-case-default` here is not the good idea.

examples/code-highlighting/index.js
thesunny: @CameronAckermanSEL @CodePlayer A paste handler in the example seems reasonable since (a) it's minimum viable feature if it crashes and (b) It is separate from the syntax highlight part of the code, so it doesn't add complexity to understanding it. If unneeded, easy to ignore.

examples/code-highlighting/index.js
thesunny: I agree with both. The code in its previous form is easier to understand and to customize. The code in its new form is out-of-the-box more useful as a drop in case but at the cost of making understanding what's going on a little more difficult.

I wonder if it's appropriate to split this into two examples:

- The way it is currently
- A new way called "Syntax Highlighting with Prism Themes"

Feels like more examples would be useful in this case. Projects like `next.js` provide a lot of examples and variations and I'm always glad for it.

thesunny: First of all, thanks for submitting this PR @CodePlayer 

It adds complexity which I don't like for a basic example but I like the idea of having an example using Prism Themes. For me personally, I'd appreciate having a separate "Syntax Highlighting with Prism Themes" example.

That said, I feel like there's some unnecessary complexity in this Prism example as well. I haven't gone through the example with a fine tooth comb but it's hard to parse on first viewing.

ianstormtaylor:

Issue #2570, opened 2019-01-31
Make use of clientX and clientY in getEventRange


amytych:
Is there a reason why `x` and `y` are being used here:

https://github.com/ianstormtaylor/slate/blob/7a71de387cebd17dd5814409b16c994bbb0e53a1/packages/slate-react/src/utils/get-event-range.js#L26-L27

These are experimental properties and they are aliases for `clientX` and `clientY` and might be missing in some browsers: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/x

It causes a problem for us when trying to test a drag&drop functionality in our app using Cypress. It omits `x` and `y` from the programatically created event, but `clientX` and `clientY` are present. These properties missing cause `getEventRange` to always return `null`.

Would you be open to a PR switching to use `clientX` and `clientY` or are there drawbacks to it that I'm missing?

brendancarney:
I don't think there are any drawbacks. Feel free to create the PR

PR #2675, opened 2019-04-01
Make use of clientX and clientY in getEventRange

amytych:
#### What's the new behavior?

Instead of using `x` and `y` from `event` object, which are [experimental properties](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/x) and missing in some cases, it uses `clientX` and `clientY` which are widely supported.

It doesn't change anything

Fixes: #2570
Reviewers: @


ianstormtaylor:
Great simple pull request, thank you @amytych!

PR #2676, opened 2019-04-01
Use <s> for strikethrough example

DominikSerafin:
`<strike>` is obsolete and discouraged https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strike

`<s>` is better alternative https://developer.mozilla.org/en-US/docs/Web/HTML/Element/s


ianstormtaylor:
Thanks @DominikSerafin!

PR #2680, opened 2019-04-07
Unicode/emoji offset on remove and move ops

### LINKED ISSUE #2632: Deleting Plaintext Emoji as First Character Throws Error
_bug_

Inserting an emoji in plain text -- either by using `insertText()`, copy pasting, or using Mac's emoji picker -- as the first character in the editor, then hitting the left arrow once and hitting backspace results in an error: `slate.js:6451 Uncaught TypeError: Cannot read property 'text' of null`
Sandbox example: 

https://codesandbox.io/s/8zyqw2jy38

This also happens in the Slate.js example editor:

Allow the backspace, resulting in the replacement character (�). I'd be happy to open a PR if this is indeed a bug. 

Thanks!

### LINKED ISSUE #2635: Backspace does not properly delete complicated emojis
Bug Report: Backspace does not handle complicated emojis

For an emoji that is represented with multiple characters (example: male zombie 🧟‍♂️), attempting to delete this emoji with the backspace key does not work as expected. 

The user must hit the backspace key `n` times where n is the number of characters in the emoji representation.

Pressing backspace once should delete the emoji.

#### Reproduction

You can reproduce this on the Slate examples page: https://www.slatejs.org/#/plain-text.
I am using Chrome on Mac 10.14.2

complicated

adjourn:
#### TL;DR 
Side-by-side comparison with default: https://jsfiddle.net/swjq5h8k/
All different cases that were tested: https://jsfiddle.net/gxnhcbo6/
 
Both.

### What's the new behavior?

Plain text emojis are broken because:
- it's possible to move selection "into" the emoji (between surrogate pairs) and break it into � (broken) characters if new characters are inserted or `Backspace`/`Delete` is pressed while "in" emoji
- have to press `arrow keys` multiple times to move selection over the emoji "for real" (it can range from 2 to >10 depending on emoji) without any visual feedback (caret is not accurate, terrible UX)
- have to press `Backspace` and `Delete` multiple times when emoji is other than simple surrogate pair

I think this is something that Slate should be able to handle.

If you currently want to fix emojis, you'd have to cover all cases emojis could be inserted and convert them into `inlines`. I think that it makes our editors overly complex and decreases performance.

Proposed function calculates the offset required to move selection over whole (emoji) sequence. It covers `arrow keys`, `Delete` and `Backspace`, other cases like trying to put caret in the middle of sequence where it shouldn't be, is already handled by browser. 

Compare with current Slate: https://jsfiddle.net/swjq5h8k/ (PR implemented as plugin).

#### Performance
It's quite performant, Im seeing max `0.5ms` with cold compiler (usually `~0.2ms`) and `<0.05ms` otherwise. Even 6x CPU slowdown is less than `1ms` with cold compiler and my CPU is at least 5 years old.

#### Handled cases
Unicode rabbit hole is deep, there are zero width joiners, variation selectors, surrogates, etc I didn't even know existed few days ago. You can see all my test cases here: https://jsfiddle.net/gxnhcbo6/. 

#### Graceful fallback
If you manage to break it, **please** let me know and you'll get a virtual cookie. It should never throw an error, edge cases that might slip through, have the same experience as with all emojis at current Slate state.

### How does this change work?
It aims to mimic browser behavior as much as possible without dictionaries or 10kB regex.

I changed the way `getCharOffset` works which is used by `Delete` and `Backspace` already and made forward and backward events (arrow keys) in `slate-react` use it aswell.

`getCharOffset` always receives a string from (about to be) selection until the end of block. `while` loop iterates over string's code points until it has determined that current code point doesn't form a sequence with previous (and next in some cases). E.g complex emojis take 3-7 iterations, single characters 1.

#### Each iteration generally checks the following things for current code point:
Order matters, if any IF is true, we'll either `break` the loop if it determines that previous and current code points don't form a sequence or `continue` to next code point after we've updated offset and other meta. 

- IF surrogate
- IF ZWJ (zero width joiner)
- IF variation selector
- BREAK (most "regular" characters fall under here, 1 iteration)

#### Few examples (numbers are .length): 
`😀😀` `[emoji 2][emoji 2]` 
Selection should move over whole emoji (not end up in the middle of surrogate pair) and be able to move between the two, **offset should be 2**.

`😀‍😀` `[emoji 2][zero width joiner 1][emoji 2]` 
ZWJ (zero width joiner) combines the two, selection should now move over whole thing, **offset should be 5**. You should also notice that you can't copy these emojis one-by-one anymore, it's a single unit.

`👮🏻` `[emoji 2][modifier 2]`
Modifier "fuses" with previous unicode/surrogate pair, **offset should be 4**. The trick here is to know if it's modifier because it has same characteristics as any other surrogate pair and we've established in previous cases that selection should be able to move between the two without ZWJ. This is exception.

These even aren't all the cases and you can already see that rules contradict and complexity grows. It gets a lot worse because it has to work in backward direction too, in which case many forward rules would fail.

Fixes #2635, fixes #2632


SmilinBrian:
Looks good and seems very helpful. Side-by-side comparison is compelling.

Is there any chance similar work needs to be done to support kanji or other multi with modifiers? Sorry, no example at hand, but maybe someone more familiar with asian or arabic languages knows of a case.

adjourn:
@SmilinBrian I have no idea, I have exactly zero experience with these languages. Im open to feedback and situations where it might fail, in order to cover all cases. 

I've also explored alternatives because lets be honest, it's a brute force solution, but none worked out.

adjourn:
@SmilinBrian I have exactly zero experience with these languages. I'm open to feedback and situations where it might fail, in order to cover all cases. That's why it's labeled WIP.

I've explored alternatives because lets be honest, it's a brute force solution, but none worked out.

@ianstormtaylor What's your take on this? I see no other realistic alternatives right now. 

I did experiment with letting browser handle it (only prevent default if browser doesn't know what to do, e.g void related) and syncing Slate value. After all, browsers do the right thing with unicodes.
Selection should be easy but `Delete` and `Backspace` require DOM diffing and rethinking core concepts which is not desirable, especially when Slate is already in the middle of big change.

adjourn:
Should I handle combining characters as well in this PR? 

For example `é` can be written using single unicode (length 1) or `e` + ` ́` which has length of 2.

packages/slate-react/src/plugins/after.js
```diff
@@ -511,7 +512,13 @@ function AfterPlugin(options = {}) {
         return editor.moveToStart()
       }
 
-      return editor.moveBackward()
+      const { start } = selection
```
ianstormtaylor: I think instead of inlining the length calculating logic here, we should instead make it handle a single "character" by default when no `n` is provided? (Instead of defaulting to `1`?)

rikkit:
(This more of an observation than a opinion, great work on the PR @adjourn !)

> ZWJ (zero width joiner) combines the two, selection should now move over whole thing, offset should be 5. You should also notice that you can't copy these emojis one-by-one anymore, it's a single unit.

Should ZWJ glue together all emojis, or just those that are presented together? Only pointing this out cos you noted

> It aims to mimic browser behavior without dictionaries or 10kB regex.

Firefox and Edge (pre Chromium version) both select a single emoji

image
image

Whereas Chrome selects both (as per the logic in the PR)
image

You could argue that either implementation is correct, I have absolutely no idea how Firefox and Edge have done this though. Like you say it'd be ideal to just let the browser handle it but there's the issues with delete and backspace you mentioned

adjourn:
Sorry about the delay, I'm on vacation but I'll return to this in few days. If someone REALLY needs this feature, please see side-by-side comparison, it's implemented as a plugin in there.

danieltorrer:
This feature is really helpful! Just as a heads up: I used your jsfiddle code as plugin and noticed that with this plugin activated you can no longer select text with `shift + arrow key` or `cmd + shift + arrow key`.

adjourn:
@danieltorrer Whoa, was worried for a second! But fortunately this is due to the fact that `shift` logic is not handled in plugin `onKeyDown` handler, everything works as it should when it's part of Slate.

adjourn:
@rikkit Thank you for pointing that out. 

I think that current implementation is the way to go because:

* conceptually ZWJ is meant to connect two characters
* it would add **so many** edge cases
* nobody (don't quote me on that) adds ZWJ-s manually
* so many users use Chromium based browsers

But if someone has objections, I'm all ears.

packages/slate-react/src/plugins/after.js
adjourn: @ianstormtaylor To be clear, are you suggesting to move offset calculation into [moveForward/Backward](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/commands/on-selection.js#L358)? Like it's already in [deleteCharForward/Backward/AtRange](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/commands/at-range.js#L427)?

adjourn: @ianstormtaylor To be totally clear, are you suggesting to move offset calculation into [moveForward/Backward](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/commands/on-selection.js#L358)? Like [deleteCharForward/Backward/AtRange](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/commands/at-range.js#L427) already functions with the exception that only calculate it when no `length args` is provided.

ianstormtaylor: @adjourn Yup, moving it in there. But actually, maybe it's better to treat the `n` argument as "characters" rather than "length"? So that it's what you'd expect for all cases.

adjourn:
That was my worst conflict resolution to date, glad it's over.

@ianstormtaylor It's ready for another review.

Notable changes (see [this commit](https://github.com/ianstormtaylor/slate/pull/2680/commits/9c41cfab8c2e0e274412a23d9bfc366c260725ad), last one is nitpicking and `git` lying to me): 
* moved offset logic from `slate-react` plugin into `move*` commands (`moveForward/moveBackward`)
* changed the concept/`move*` argument from `length` to `character` which means that e.g if there's 3 emojis each with length 5, `moveForward(2)` doesn't move selection in the middle of first one, it moves selection over 2 emojis, i.e "characters".

See it in action [here](https://jsfiddle.net/kmn6j8ht/) (read the paragraph starting with **PS**).

adjourn:
I think something is still off, Im going to investigate this, if someone has pointer, I'd be glad to hear them.

adjourn:
Possible features for another day, another PR:
* characters like `é` can be written using single unicode (length 1) or `e` + ` ́` which has length of 2
* support negative `chars` argument which would result in calling the opposite command, e.g `moveForward(-2)`calls `moveBackward(2)` under the hood and vice versa
* useful unicode symbol categories like math symbols, etc
* explore unicode edge cases, exotic languages, etc (I don't think I ever get here)

packages/slate/src/utils/text-utils.js
```diff
@@ -81,168 +138,144 @@ function getCharLength(char) {
...
+ * Get the offset to the end of character(s) before an `offset` in `text`.
  *
  * @param {String} text
  * @param {Number} offset
+ * @param {Number} chars
  * @return {Number}
  */
 
-function getCharOffsetBackward(text, offset) {
+function getCharsOffsetBackward(text, offset, chars) {
```
adjourn: @ianstormtaylor I think it makes sense to also add default `chars = 1` here (and forward function)? Should I retain original `getCharOffsetBackward` and add `getCharsOffsetBackward` as separate function? Even though this API is not documented, it would be a breaking change.

ianstormtaylor: Having that argument makes sense to me, good call. Let's just keep the function name non-plural for now? And we can fix it later with a deprecation when needed.

isubasti:
@adjourn just found this package that is actually used by `eslint-plugin-jsx-a11y` as its dependency for parsing emoji: https://www.npmjs.com/package/emoji-regex, maybe slate can rely on that package instead?

slotterbackW:
Is there any update on the progress of this PR? Will it be merged soon?

ianstormtaylor:
Hey @adjourn I'm super interested in this. For some reason the diff shows a bunch of changed stuff that is from earlier commits to master. Is there any way to fix that? It's hard to see what exactly his changing here. Thanks!

slotterbackW:
@ianstormtaylor from adjourn's comment above here's an easy to read view of commits:

> Notable changes (see [clean view of all my commits](https://github.com/ianstormtaylor/slate/pull/2680/files/5bef253855787c72e589aab0a804b3ef14a5b197..89f0e26663138ca5382cf71e7d8edc821b1143b3)):
> 

Excited to see development on this! Thanks both of you!

kwokhuen:
@ianstormtaylor I know how to fix the commits. I also found some bugs while deleting and moving forward and backward. I'm wondering if I could continue to work on this PR. If yes, how would I go about it? Should I just checkout @adjourn branch, do my thing and force push?

ianstormtaylor:

PR #2686, opened 2019-04-12
Fix selection being reset incorrectly after combination of events

sgurenkov:
If mouseDown happens inside editor container and mouseUp outside it,
then programmatically calling editor.focus() after that will mess up
the selection

fixing a _bug_

#### What's the new behavior?
to reproduce the bug do the following:
1. mouseDown inside editor, move to select content and release mouse button outside editor
2. blur the editor (click outside the editor component)
3. programmatically call editor.focus() (maybe clicking a toolbar formatting button for example)
4. see selection being collapsed instead of resetting it to the selected fragment

the problem was is that when on mouseDown we set a closure variable isMouseDown to true and reset it on mouseUp event, but since mouseUp may happen outside the editor it was not resetting it.
The solutin is to reset it inside onSelect handler. We only need that variable in onFocus handler. 
The order of events is following:
* mouseDown
* focus (isMouseDown being used here to distinguish events created by mouse)
* select
* mouseUp

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/plugins/after.js
+++ b/packages/slate-react/src/plugins/after.js
@@ -634,9 +634,11 @@ function AfterPlugin(options = {}) {
     const window = getWindow(event.target)
     const selection = window.getSelection()
     setSelectionFromDom(window, editor, selection)
-    // we need to reset it here in case mouseUp happens outside editor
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @sgurenkov!

Issue #2495, opened 2018-12-11
switch to plain JSON models


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

This is an issue summarizing the discussion in https://github.com/ianstormtaylor/slate/issues/2345. I'm opening it so that we can track some of the dependent issues that need to be solved before we can fully migrate, since there are some things left to be figured out...

###### `slate`
- [x] Change memoization to not use instances. (#2471, thanks @zhujinxuan!)
- [ ] Change `slate` core to use paths or path refs, not keys. (#2864)
- [ ] Remove extraneous instance methods. (#2711)
- [ ] Remove the `*ByKey` commands. (#2858)
- [ ] Deprecate instance methods in favor of static methods.

###### `slate-react`
- [x] ~~Pass a `pathRef/pathQuery` object to components, not keys. (#2413)~~
- [x] Change `findDOMNode/etc.` to use refs instead of searching by `key`. (#2716)
- [x] Change decorations to use relative paths, and always set them. (#2745)
- [x] Change handlers to use `findPath` instead of `findNode().key`. (#2717)
- [ ] Augment the React plugin to set keys on nodes, and when operations are applied.

After all of those are landed, we can then have a single breaking change where we change everything to use plain JSON objects instead Immutable.js instances, since none of the internals are relying on the instances anymore.

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/3093.

PR #2687, opened 2019-04-13
Revert "Use fromJS instead of Map to convert data into block data"

thesunny:
Reverts ianstormtaylor/slate#2619


czechdave:
Is the direction forward going to be not enforcing Immutable.js downstream on users of Slate? This latest change causes a lot of issues in places we use Slate's data on nodes.

CameronAckermanSEL:
The direction going forward is to remove immutable JS https://github.com/ianstormtaylor/slate/issues/2495

Issue #2679, opened 2019-04-05
Uncaught TypeError: previous.getMarksAtIndex is not a function


lapico:
https://github.com/ianstormtaylor/slate/blob/bb5d6beffa393176c55392b07ccbb0de63bd2138/packages/slate/src/interfaces/element.js#L970-L977

Maybe replace `return previous` by `return previousText` ?

jtadmor:
My apologies, looks like I broke that. Agreed that the fix should just be calling `return previousText`, seems like we also don't have any tests for that method so I'll take a look at those.

PR #2688, opened 2019-04-13
fix get marks from previous text on method getMarksAtPosition

arrizalamin:
Fixing a bug when calling `getMarksAtPosition` with offset 0 throws an error.

#### What's the new behavior?
Nothing

There is a typo calling `getMarksAtIndex` method on `previous` which is a tuple, it should be `previousText`.

* [ ] The relevant examples still work. (Unable to connect to `http://localhost:8080/`)

Fixes: #2679
Reviewers: @ianstormtaylor @jtadmor


jtadmor:
Thanks for pulling this together and adding all these tests!

ianstormtaylor:
Beautiful pull request, thanks @arrizalamin!

PR #2689, opened 2019-04-13
removed ie/edge condition that causes "node of undefined" TypeError i…

robinrutten:

Fixing a bug in Edge. When empty editor is focused, it throws "node of undefined" type error.

#### What's the new behavior?

Only removes browser (edge/ie) specific code. As far as I can see code is not doing anything at all.

Cannot reproduce bug described in PR #1574. (Tested in Edge/18.18262)

Removes unused edge/ie code.

Reverts code added in PR #1574. I was not able to reproduce issue in #1574 using Edge/18.18262.

### DIFF
--- a/packages/slate-react/src/utils/find-range.js
+++ b/packages/slate-react/src/utils/find-range.js
@@ -1,10 +1,8 @@
 import getWindow from 'get-window'
 import invariant from 'tiny-invariant'
-import { IS_IE, IS_EDGE } from 'slate-dev-environment'
 import { Value } from 'slate'
 
 import findPoint from './find-point'
-import findDOMPoint from './find-dom-point'
 
 /**
  * Find a Slate range from a DOM `native` selection.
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/utils/find-range.js
```diff
@@ -51,21 +49,6 @@ function findRange(native, editor) {
   const focus = isCollapsed ? anchor : findPoint(focusNode, focusOffset, editor)
   if (!anchor || !focus) return null
```
CameronAckermanSEL: would it be prudent to leave in the IS_IE check?

robinrutten: I tested the issue described in PR #1574 in IE11. Seems to work fine.

ie11

robinrutten: Tested #1574 in IE11, but I can't find any issues. It think the issue might have been solved by refactored code since Jan 18?

ie11 2019-04-18 13_31_44

ianstormtaylor:
Thanks @robinrutten!

Issue #2616, opened 2019-02-26
Type Enter at end of first line in Syntax Highlighting Demo Throws Error


thesunny:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

If you hit ENTER at the end of the first code line which is a comment in the demo here:

https://www.slatejs.org/#/code-highlighting

The result is this error:

```
TypeError: Cannot read property 'size' of null
    at We (https://www.slatejs.org/main-e8e0520e642ae32de49e.js:1:26219)
    at Object.Le [as compare] (https://www.slatejs.org/main-e8e0520e642ae32de49e.js:1:25121)
    at https://www.slatejs.org/main-e8e0520e642ae32de49e.js:1:173417
    at https://www.slatejs.org/main-e8e0520e642ae32de49e.js:1:242097
    at pa.__iterate (https://www.slatejs.org/main-e8e0520e642ae32de49e.js:1:235496)
    at r.Ia.r.__iterateUncached (https://www.slatejs.org/main-e8e0520e642ae32de49e.js:1:242066)
    at ue (https://www.slatejs.org/main-e8e0520e642ae32de49e.js:1:215563)
    at r.U.__iterate (https://www.slatejs.org/main-e8e0520e642ae32de49e.js:1:214394)
    at r.forEach (https://www.slatejs.org/main-e8e0520e642ae32de49e.js:1:259439)
    at https://www.slatejs.org/main-e8e0520e642ae32de49e.js:1:233254
```

It only happens on the first line.

#### What's the expected behavior?

Should not throw an error.

brendancarney:
I believe this is a duplicate of #2319

thesunny:
It may be a duplicate, but it seems unclear because the bugs present themselves differently.

- One is hitting enter twice at the second-last position of any line
- This one is hitting enter once at the last position of the first line

I suggest we keep this open until someone can confirm that they are the same.

Issue #2696, opened 2019-04-18
Add support for but warn when a decoration doesn't have a path (use key instead)


thesunny:
#### Do you want to request a _feature_ or report a _bug_?

An improvement that stops editor from crashing when paths are not provided for a decoration. This behavior was supported in the past.

#### What's the current behavior?

If a decoration is created that doesn't have a `start.path` or `end.path` even though it has a `start.key` or `end.key` it throws an error and breaks the editor.

For example, this is causing the editor to break at times in the syntax highlighting example.

https://github.com/ianstormtaylor/slate/issues/2616

#### What's the expected behavior?

For the time being, I suggest we derive the `path` from the `key` and use that and display a deprecation warning.

I'm putting a PR together right now.

PR #2697, opened 2019-04-18
Support decorations without path

thesunny:

Fixing a bug.

Helps with
- https://github.com/ianstormtaylor/slate/issues/2319
- https://github.com/ianstormtaylor/slate/issues/2616

#### What's the new behavior?

Instead of throwing on decorations without a path but a key, use the key to generate the path but display a warning. The syntax highlighting example is causing this issue and I had to deep dive to figure out the cause. This warning will let everything still work while giving us the opportunity to figure out the source of the path not being on the decoration.

My guess is that the example used to work but broke at some point when the `path` suddenly was required and the `key` wasn't enough.

If a path is missing on a decoration's `start` or `end` it will provide a warning but will use `document.assertPath` to deriver the path and the editor won't crash. 

Fixes: https://github.com/ianstormtaylor/slate/issues/2616 https://github.com/ianstormtaylor/slate/issues/2696 https://github.com/ianstormtaylor/slate/issues/2319
Reviewers: @ianstormtaylor 

Note that I don't believe it 100% fixes 2616 or 2319 because those should probably have `path` in them but (a) it stops the crashing and it kinda works again and (b) it helps us provide a safe path to getting syntax highlighting fixed without having crashing editors.

### DIFF
--- a/packages/slate-react/src/components/text.js
+++ b/packages/slate-react/src/components/text.js
@@ -125,10 +125,12 @@ class Text extends React.Component {
         start.path != null,
         'No path for the decoration start. Not providing one is deprecated.'
       )
+
       warning(
         end.path != null,
         'No path for the decoration end. Not providing one is deprecated.'
       )
+
       const startPath = start.path || document.assertPath(start.key)
       const endPath = end.path || document.assertPath(end.key)
 
--- a/packages/slate-react/src/components/text.js
+++ b/packages/slate-react/src/components/text.js
@@ -118,19 +118,6 @@ class Text extends React.Component {
       if (start.key === end.key) return false
 
       const path = document.assertPath(key)
-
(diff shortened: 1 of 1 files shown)


thesunny:
@ianstormtaylor I think I'm going to have to leave this one to you.

I've fixed syntax highlighting properly; however, it was a bit of work to track down. If anybody else creates a decoration without a path but with a key, this will (a) make it work and (b) give a helpful warning. Feels like it would be helpful but not sure if you want to add more code to show the warning/make it work.

Another option is that I could just show throw a helpful error instead of a warning instead.

ianstormtaylor:
Hey @thesunny, thank you for this, this is a great call. I'm going to remove the warnings for now, and just set the path by default, so we don't have to bump. Later keys will be removed entirely so people will need to change anyways.

PR #2698, opened 2019-04-18
Fix syntax highlighting

thesunny:

Fix bug in https://github.com/ianstormtaylor/slate/issues/2616 and https://github.com/ianstormtaylor/slate/issues/2319

#### What's the new behavior?

Stops the editor from crashing in syntax highlighting when hitting enter in certain places

Generates a path for each decoration. Previously, we were only passing through the key that seems to work sometimes but not others. This is also referenced in a separate PR I made here https://github.com/ianstormtaylor/slate/pull/2697. Both these PRs improve Slate. This one fixes this bug. The other PR fixes all similar issues and provides a nice warning.

Fixes: #2616 #2319 
Reviewers: @

### DIFF
--- a/examples/code-highlighting/index.js
+++ b/examples/code-highlighting/index.js
@@ -1,7 +1,9 @@
 import { Editor } from 'slate-react'
+import { Value } from 'slate'
+
 import Prism from 'prismjs'
 import React from 'react'
-import { Value } from 'slate'
+
 import initialValueAsJson from './value.json'
 
 /**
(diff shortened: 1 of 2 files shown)


thesunny:
Over aggressive VS code plugin is re-arranging the imports unnecessarily. Will make an update before merging.

unshift:
@thesunny Thank you so much, I was struggling with this issue!

thesunny:
@unshift My pleasure and you're welcome. Thank you for sending the appreciation.

PR #2703, opened 2019-04-23
Add value inspector to composition example

thesunny:

Improving examples

#### What's the new behavior?

Adds a text box below the editor in the compositions example that shows the text inside the internal state of Slate. This is important because there can be bugs in compositions where the DOM correctly displays some content but that text has not been reconciled to Slate's internal `Value` yet.

This lets us see what's inside Slate as well as what's in the DOM.

Converts Slate's Value to an array of Strings (one string per top level block) then converts each item into a `div` displayed below the Editor.

Fixes: #
Reviewers: @

### DIFF
--- a/examples/composition/index.js
+++ b/examples/composition/index.js
@@ -9,6 +9,7 @@ import insert from './insert.js'
 import special from './special.js'
 import { isKeyHotkey } from 'is-hotkey'
 import { Button, Icon, Toolbar } from '../components'
+import { ANDROID_API_VERSION } from 'slate-dev-environment'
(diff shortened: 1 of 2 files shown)


thesunny:
FYI, the purpose of this PR is primarily to help with Android development.

1. Before this PR, it's difficult to see what changes have been committed to Slate's internal `value`
2. I have multiple API versions of Android running and this allows me to see easily which version each emulator is running

thesunny:
Thanks @CameronAckermanSEL

PR #2710, opened 2019-04-26
Improve performance on large documents with many nodes

yonahforst:
Fixing a bug

#### What's the new behavior?
Some debug payloads contain expressions that need to be evaluated, which can be expensive. (for example calling `value.toJSON()` on a large document with many nodes)

This PR makes sure that those payloads are only evaluated if debugging is actually enabled, thereby improving performance in production.

We simply call `if (debug.enabled)` before those debug calls. This way, if debugging is indeed disabled, the contained expressions are not needlessly evaluated.

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -162,6 +162,7 @@ class Content extends React.Component {
     const window = getWindow(this.element)
     const native = window.getSelection()
     const { activeElement } = window.document
+
     if (debug.enabled) {
       debug.update('updateSelection', { selection: selection.toJSON() })
     }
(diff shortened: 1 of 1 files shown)


CameronAckermanSEL:
This looks good to me. Do you have any benchmarks you can share though?

ianstormtaylor:
Great call, thank you @yonahforst!

Slapbox:
I'd love to see benchmarks too if available, but work saved is work saved and I'm super grateful for this find! Thank you @yonahforst!

yonahforst:
@Slapbox @CameronAckermanSEL - I just ran the benchmarks but saw no meaningful difference. I think this is because the `slate-react`benchmarks are setup to profile the initial rendering, where the effect of this PR is only visible on changes.

Slapbox:
@yonahforst - I spent a day or more trying to figure out how to fork a monorepo for use with npm, but I have no luck at all. Most people who responded to my inquiry told me they were pretty sure it wasn't currently possible, and my research found nothing to the contrary.

That was about 5 months ago. I'd love to find out that that's changed.

ianstormtaylor:
@yonahforst published!

Issue #2481, opened 2018-12-04
Active mark is missing when typing


zhujinxuan:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

See the gif below, when we select a bold word and then type, and the bold mark does not exist for the newly insert text:

![](http://g.recordit.co/2og8AEHaUS.gif)

#### What's the expected behavior?

Bold mark shall exists, just like the old slate did (at least I remember the mark is kept in v0.40)

I will submit a PR on that this weekend if it is confirmed as a bug.

Browser: Firefox 63.0.3 (64-bit)

Issue #2705, opened 2019-04-23
Value.activeMarks returns incorrect set of marks for overlapping marks in selection


ferdaber:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
The site's example rich text editor reproduces this behavior: https://www.slatejs.org/#/rich-text.

![image](https://media.giphy.com/media/chyZTg3JBJhnRhZQQw/giphy.gif)

In the gif above, I overlapped an underline mark and a bold mark. Selecting the text from the start of the bold mark produces correct behavior of having the bold mark button be rendered as active, but as the selection grows to also encompass the underlined text, the bold mark button is still rendered as active. This appears to be behaving as expected when there are no overlapping marks (just bold text where the selection grows larger than the bolded text range)

#### What's the expected behavior?

The `Value.activeMarks` function only returns the subset of marks where *every* character in the selection possesses that mark. In the above case, I expected the bold mark button to be deactivated when the selection grows beyond the bolded text.

I am more than happy to help fix this, if someone could point me in the direction of how `Value.activeMarks` is actually populated (seems to be a getter in `Selection.marks` but that's where I got lost).

ferdaber:
From some debugging it looks like it's failing at `Text.getActiveMarksBetweenOffsets`, where the `forEach` loop is breaking early on the first intersection it finds: https://github.com/ianstormtaylor/slate/blob/bb5d6beffa393176c55392b07ccbb0de63bd2138/packages/slate/src/models/text.js#L314

Removing that last `return false` line makes it work as expected. I can create a PR if the maintainers agree that this is the solution.

isubasti:
@ferdaber check the doc https://docs.slatejs.org/slate-core/value#activemarks, https://docs.slatejs.org/slate-core/value#marks. what you want is `value.marks` not `value.activeMarks`

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #2730, opened 2019-05-02
add iterables

ianstormtaylor:

Improvement.

#### What's the new behavior?

This adds ["iterables"](http://2ality.com/2015/02/es6-iteration.html) to Slate's core data model, with several goals:

1. To make working with nodes and paths easier, since the iterables can return "entries"—like the native `Map` and `Set` do—comprised of `[node, path]`.

2. To avoid having to traverse the entire tree or compile an entire list for times when you just need the first `n` of a list, or when you need to check items against a predicate function.

3. To make it easier to maintain Slate's iteration logic, and prevent it from being defined differently all over the place, as it was before.

This introduces several iteratable-producing methods on the `Element` interface, which `Document`, `Block` and `Inline` all implement.

There are iterables for traversing the entire tree:

```js
element.blocks(options)
element.descendants(options)
element.inlines(options)
element.texts(options)

element.ancestors(path, options)
element.siblings(path, options)
```

You can use them just like the native JavaScript iterables. For example, you can through the next text nodes after a specific node:

```js
for (const next of document.texts({ path: start.path })) {
  const [node, path] = next
  // do something with the text node or its path
}
```

Or you can traverse all of the "leaf" blocks:

```js
for (const [block] of document.blocks({ onlyLeaves: true })) {
  // ...
}
```

And because these iterations use native `for/of` loops, you can easily `break` or `return` out of the loops directly—a much nicer DX than remembering to `return false`.

Fixes: https://github.com/ianstormtaylor/slate/issues/2740
Fixes: https://github.com/ianstormtaylor/slate/issues/2481

Maybe: https://github.com/ianstormtaylor/slate/issues/2705


ianstormtaylor:
Nevermind, added to https://github.com/ianstormtaylor/slate/pull/2747 to avoid having to deal with a weird merge.

Issue #2582, opened 2019-02-05
insertFragment does not reposition cursor correctly for fragments with an inline before a text node


gdehmlow:
#### Do you want to request a _feature_ or report a _bug_?

Bug!

#### What's the current behavior?

When a fragment is inserted that has a block with an inline splitting up some text, the cursor ends up at the end of the text of the inline rather than at the end of the fragment text.

GIF: 
![insertfragment text position](https://user-images.githubusercontent.com/850025/52255008-7d750480-28c5-11e9-9007-7f792d18c167.gif)

JSFiddle (just hit the enter key to insert the fragment):
https://jsfiddle.net/Lysnd9o7/3/

#### What's the expected behavior?

When a fragment is inserted, the cursor should end up at the very end of the fragment.

I think the issue is here:

https://github.com/ianstormtaylor/slate/blob/6e18719f3d705a403905526ec9b10eb5a9518c5c/packages/slate/src/commands/with-intent.js#L281

If you start with a blank text editor, `newTexts` looks like this:

```js
[
  { leaves: [{ marks: [], text: "Here is an example, " }], key: "75" },
  { leaves: [{ marks: [], text: "cursor ends up here :" }], key: "76" },
  { leaves: [{ marks: [], text: ", when it should end here:" }], key: "78" }
];
```

So the last new text is `", when it should end here:"`, and we move the cursor to the front of that. But then we call `editor.moveForward` with the _total_ length of the text inserted (and maybe more) with `lastBlock.text.length`.. which is all of those texts combined! And `editor.moveForward` seems to just not do anything if the value exceeds the length of the current text.

So it seems that our selection positioning logic assumes that the last block of the fragment is one continuous text.. which doesn't work for this use case.

I feel like there are a few edge cases here that we have to consider in order to get this working properly.. will think about a good way to solve this.

ererik:
The issue can also be observed by copying and pasting text containing links in https://www.slatejs.org/#/links

Entkenntnis:
We experience the same bug in our project and applied above patch.

PR #2738, opened 2019-05-06
fix selection after pasting inline nodes

Entkenntnis:

Fix

#### What's the new behavior?

After pasting a fragment with inline nodes, the selection is moved to the correct postition (after the inserted fragment). See the [Link example](https://www.slatejs.org/#/links).

Skip inline nodes before summing up the length of all text nodes, 

* [x ] The new code matches the existing patterns and styles.
* [x ] The linter passes with `yarn lint`. (Fix errors with `yarn prettier`.)
* [x ] The relevant examples still work. (Run examples with `yarn watch`.)

Fixes: #2582
Reviewers: @

### DIFF
--- a/packages/slate/src/commands/with-intent.js
+++ b/packages/slate/src/commands/with-intent.js
@@ -283,13 +283,14 @@ Commands.insertFragment = (editor, fragment) => {
     // these have to be skipped.
     const { nodes } = lastBlock
     const lastIndex = nodes.findLastIndex(
-      node => node && node.object == 'inline'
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Hey @Entkenntnis thanks for taking the time to work on this! Can you add a test case that we can use to make sure we don't regress in the future? Thanks!

Entkenntnis:
I'm happy to help, added a test case

ianstormtaylor:
Awesome, thank you @Entkenntnis!

Issue #2745, opened 2019-05-08
split up marks, decorations, and annotations


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now there are three ways that mark-level formatting can be applied in Slate:

1. Via `mark` objects that are embedded in the document itself.

2. Via decoration ranges (which contain a mark object) that are added to a specific node by the `decorateNode` middleware. These never actually live in the data model, and are a render-time-only construct.

3. Via `value.decorations` which contains a group of decorations that live in the data model, but not necessarily inside the individual text nodes. They are rendered similarly to the other decorations.

Each one of these has their own uses cases. Marks are for classic formatting like bold, italic, etc. Render-time decorations are for code highlighting, markdown highlighting, etc. And top-level decorations are for comments, suggestions, collaborative cursors, etc.

But there are also some confusing overlaps between the three, which leads to sloppier code and less performance. 

For example, decorations are usually rendered at the block level, but their ranges are relative to the entire document, which makes isolating them while rendering less performant than it could be. And top-level decorations are stored in an immutable list, when many use cases require referencing them by an identifier, which requires slowly traversing the entire list.

#### What's the expected behavior?

I think we need to split them apart.

###### MARKS

The current marks use case is the most solid and widely used, and it will stay as is.

###### DECORATIONS

The render-time decorations use case (which was the first use of "decorations") will also stay very similar. 

Except, the paths on a decoration will change to be scoped to the specific node that they are decorating, instead of relative to the document itself. This makes it easy to use paths only (and drop keys support in the future). It also means that they can be compared more quickly while rendering the tree of nodes.

In addition, instead of using a nested `decoration.mark` property, and overloading the `renderMark` middleware, they contain that formatting information as top-level properties. So a decoration will contain:

```js
{
  type: String,
  data: Map,
  anchor: Point,
  focus: Point,
}
```

And it will be rendered with a new `renderDecoration` middleware.

###### ANNOTATIONS

The top-level `value.decorations` use case is the most problematic. It solves the "search highlighting" use case fine, but it isn't well suited to other cases like attaching comments and suggestions, or rendering collaborative cursors.

The first change will be to introduce a new name called "annotations" to separate the concept from the render-time "decorations". They are really two separate concepts, and although they share some similarities, it makes more sense to give them first-class treatment.

[...]

Issue #2717, opened 2019-04-27
add a `findPath` DOM helper


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now `findNode` is used, and then since it's easiest we end up using `node.key` for subsequent lookups.

#### What's the expected behavior?

If there was a `findPath` DOM helper instead, all of the subsequent lookups could use paths also instead of keys.

Issue #2716, opened 2019-04-27
change `findDOMNode` to use a React ref tree


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Debt.

#### What's the current behavior?

Right now `findDOMNode` uses the `data-key` attribute of elements. 

#### What's the expected behavior?

But it should be changed to use React refs instead. Each component can keep the refs to its children, and then the tree can be traversed to find a given element by `path` instead.

Issue #2520, opened 2019-01-03
throw error when pressing the Enter key or the Backspace key


CodePlayer:
I want to develop based on the official demo, the link is [https://www.slatejs.org/#/code-highlighting](https://www.slatejs.org/#/code-highlighting) 

However, when the mouse cursor was at the end of the first line, I pressed the `Enter` key, it thrown error.
Anyone can reproduce the result in the demo.

Similarly, in my local demo, when I deleted the whole second line of the `code` by pressing the `Backspace` key, it also reported an error.

The `value.json` file I am using is as follows:
```
{
    "document": {
        "nodes": [
            {
                "object": "block",
                "type": "paragraph",
                "nodes": [
                    {
                        "object": "text",
                        "leaves": [
                            {
                                "text": "There are certain behaviors that require rendering dynamic marks on string of text, like rendering code highlighting. For example:"
                            }
                        ]
                    }
                ]
            },
            {
                "object": "block",
                "type": "code",
                "data": {
                    "language": "js"
                },
                "nodes": [
                    {
                        "object": "block",
                        "type": "code_line",
                        "nodes": [
                            {
                                "object": "text",
                                "leaves": [
                                    {
                                        "text": "function mainx(){ "
                                    }
                                ]
                            }
                        ]
                    },
                    {
                        "object": "block",
                        "type": "code_line",
                        "nodes": [
                            {
                                "object": "text",
                                "leaves": [
                                    {
                                        "text": "     "
                                    }
                                ]
                            }
                        ]
                    },
                    {
                        "object": "block",
                        "type": "code_line",
                        "nodes": [
                            {
                                "object": "text",
                                "leaves": [
                                    {
                                        "text": "}"
                                    }
                                ]
                            }
                        ]
                    }
                ]
            }
        ]
    }
}
```

The error is as follows:
![local error](https://user-images.githubusercontent.com/3793502/50640295-6332c880-0f9f-11e9-8bca-212e541af10a.png)

bebbi:
Looks like the issue is caused by this chain:

`examples/code-highlighting/index.js:220`: Decorations `{ anchor, focus, mark }` are assigned with a `key` but no `path` property.

`slate/src/models/decoration.js:118`: The `{ anchor, focus, mark }` all do not have a path object

`slate/src/models/point.js:97`: The point is created with a `null` path.

`slate-react/src/components/text.js:122`: The `start.path` is null which is a bad input for `PathUtils.compare`.


[...]

PR #2747, opened 2019-05-09
Introduce annotations

ianstormtaylor:

Improvement.

#### What's the new behavior?

This pull request introduces the `Annotation` model, as well as separates marks, decorations and annotations. In the process, it upgrades a lot of the rendering logic in `slate-react` to be more consistent going forward.

###### NEW

**Introducing the `Annotation` model.** This is very similar to what used to be stored in `value.decorations`, except they also contain a unique "key" to be identified by. They can be used for things like comments, suggestions, collaborative cursors, etc.

```js
{
  object: 'annotation',
  key: String,
  type: String,
  data: Map,
  anchor: Point,
  focus: Point,
}
```

**There are three new `*_annotation` operations.** The set of operations now includes `add_annotation`, `remove_annotation` and `set_annotation`. They are similar to the existing `*_mark` operations.

###### BREAKING

**The `value.decorations` property is now `value.annotations`.** Following with the split of decorations into annotations, this property was also renamed. They must now contain unique `key` properties, as they are stored as a `Map` instead of a `List`. This allows for much more performant updates.

**The `Decoration` model no longer has a nested `mark` property.** Previously a real `Mark` object was used as a property on decorations, but now the `type` and `data` properties are first class properties instead.

```js
{
  object: 'decoration',
  type: String,
  data: Map,
  anchor: Point,
  focus: Point,
}
```

**The `anchor.path` and `focus.path` of decorations must be relative!** Previously they were optional, and they were calculated based on the path in the top-level document. Now they are paths relative to the node being decorated! See the code highlighting example for how this is achieved, it ends up being easier than before.

**The `renderNode` middleware has been split into `renderBlock` and `renderInline`.** Previously the single middleware handled both node types. This change makes it easier to define the most common cases, and paves the way for similar serializers in the future. There is also a new `renderDocument` middleware, but most people don't need to concern themselves with it.

**The `renderMark` middleware is no longer used for decorations!** Previously `renderMark` would be used for rendering both types of decorations. Now there are separate `renderAnnotation` and `renderDecoration` middleware functions to use instead.

**You must now assign `attributes.ref` to a DOM node.** This new attribute that is passed to the rendering functions must be passed into a native DOM component (using `forwardRef` if necessary). This is required to eliminate our dependence on keys. If you are using libraries that don't implement `forwardRef` you may need to use `innerRef` or similar for this.

**The `slate-hyperscript` package now uses the "annotations" name.** All of the existing APIs that previously used the word "decorations" in `slate-hyperscript` have been updated.

###### DEPRECATED

**The `find*` and `findDOM*` helpers are deprecated.** Previously you'd use these helpers to find a Slate object from a DOM object or vice versa. Now these helpers are all exposed as queries on the editor itself.

```js
// Before...
findDOMNode(key, window)
findNode(element, editor)

// After...
editor.findDOMNode(path)
editor.findNode(element)
```

Fixes: https://github.com/ianstormtaylor/slate/issues/2716
Fixes: https://github.com/ianstormtaylor/slate/issues/2745
Fixes: https://github.com/ianstormtaylor/slate/issues/2717
Fixes: https://github.com/ianstormtaylor/slate/issues/2696
Fixes: https://github.com/ianstormtaylor/slate/issues/2520

Maybe: https://github.com/ianstormtaylor/slate/issues/2636


rememberlenny:
@ianstormtaylor Can you point to an example of how to use this?

PR #2758, opened 2019-05-10
Text documentation update

quicksnap:

Docs

#### What's the new behavior?

Update Text for latest changes. `text` property is no longer computed. Add `marks` field.


ianstormtaylor:
Thank you @quicksnap this is great!

Issue #2761, opened 2019-05-11
Backwards keyboard selection throws error


arimah:
**Do you want to request a _feature_ or report a _bug_?**

Bug. 🐞

**What's the current behavior?**

https://codesandbox.io/s/mq1j15o7jx
https://streamable.com/nolnq

When a block contains multiple text nodes, backwards keyboard selection with Shift+LeftArrow causes an error to be thrown, as a path gets computed incorrectly. The mark is used here to force multiple text nodes.

The selection is updated and everything seems to work despite this error.

**What's the expected behaviour?**

No error should be thrown.

---

I *think* the issue here: https://github.com/ianstormtaylor/slate/blob/a5a25f97dde1de06ef0746e6f5285c04e503351c/packages/slate-react/src/plugins/dom/after.js#L555-L558

Compare with forward keyboard selection, where the return value of `document.texts()` receives array destructuring: https://github.com/ianstormtaylor/slate/blob/a5a25f97dde1de06ef0746e6f5285c04e503351c/packages/slate-react/src/plugins/dom/after.js#L575

I have not yet had time to build Slate locally to try and fix it myself.

PR #2764, opened 2019-05-12
Destructure texts iterable causing an error on backward shift+arrow selection

adjourn:

Bug

#### What's the new behavior?

As it was intended, error free.

It's pretty easy to understand the fix if you compare the following two (`document.texts`):

https://github.com/ianstormtaylor/slate/blob/a5a25f97dde1de06ef0746e6f5285c04e503351c/packages/slate-react/src/plugins/dom/after.js#L553-L571

https://github.com/ianstormtaylor/slate/blob/a5a25f97dde1de06ef0746e6f5285c04e503351c/packages/slate-react/src/plugins/dom/after.js#L573-L587

Fixes: #2761
Reviewers: @ianstormtaylor


ianstormtaylor:
Thank you @adjourn!

PR #2766, opened 2019-05-12
Fix leaf memoization bug

adjourn:

Bug

#### What's the new behavior?

This fix is necessary to ensure leaf re-rendering when only content changes. Good example of that would be Slate search highlighting example. First try the [old version](https://www.slatejs.org/#/search-highlighting) and then netlify deployment below. 

This happens because all the other conditions still pass, only the text content changes. If you search for `h`, previous leaf content was `This is editable text that you can search. As you search, it looks for matching strings of text, and adds "annotation" marks to them in realtime.`, new one is `T`. Everything else is the same, including parent, index `0`, no annotations, etc.

Also compare text content when deciding if to re-render or not.

Fixes: #2767

### DIFF
--- a/packages/slate-react/src/components/leaf.js
+++ b/packages/slate-react/src/components/leaf.js
@@ -201,15 +201,13 @@ Leaf.propTypes = {
 
 const MemoizedLeaf = React.memo(Leaf, (prev, next) => {
   return (
+    next.block === prev.block &&
     next.index === prev.index &&
     next.marks === prev.marks &&
     next.parent === prev.parent &&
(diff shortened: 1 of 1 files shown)


adjourn:
Does anyone else experience input lag when clearing text input or deleting last character?

pbackup12345:
Yes. There is a noticable lag here, too. Otherwise this fix was life saving. Thanks a lot. Hope it will be committed soon.

ianstormtaylor:
Thanks @adjourn!

Issue #2771, opened 2019-05-13
Reversed order after pasting list and paragraph


Entkenntnis:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

<Img src=http://g.recordit.co/FwPI8Fx2K5.gif height="200">

1. Go to the [slate rich text](https://www.slatejs.org/#/rich-text) demo.
2. Create a list and two paragraphs
3. Copy the list and one paragraph
4. Put the cursor in front of the second paragraph
5. Paste the list and the paragraph: The order of the fragment is reversed

#### What's the expected behavior?

The fragment should not be reversed

PR #2772, opened 2019-05-13
fixing order of blocks if inserted at start of node

Entkenntnis:

Bugfix

#### What's the new behavior?

The order of blocks is correct (and not reversed) after pasting in some edge cases

insertBlockAtRange() determines if a block should be added before/behind a block or the block should be splitted. The PR refactors this logic into a separate function `getInsertionMode`, which can be called inside insertFragmentAtRange() to find out whether the blocks should be added in normal or reversed order (to end up in the correct order)

Fixes: #2771 
Reviewers: @


ianstormtaylor:
Thanks @Entkenntnis!

PR #2774, opened 2019-05-13
Gracefully handle null content ref in slate-react findDOMNode query

cdunn:

Bug

#### What's the new behavior?

Return null instead of raising exception if missing content ref.  I started seeing these intermittently after upgrading to slate-react 0.22.0

Fixes: #
Reviewers: @


rg4real:
Thank you, great man! I think it will fix [my issue](https://github.com/ianstormtaylor/slate/issues/).

ianstormtaylor:
Thanks @cdunn!

PR #2780, opened 2019-05-15
Fix: prop name

jportella93:
Nodes weren't properly rendering because the Editor was trying to call the function prop renderBlock but instead another called renderNode was passed.


ianstormtaylor:
Thanks @jportella93!

PR #2782, opened 2019-05-15
Restore dom

thesunny:

Adding a feature which will help in fixing Android

#### What's the new behavior?

Add a command to the editor called `editor.restoreDOM()`. This will blow away the current DOM and replace it with a new one built from scratch using the data in the Editor's `value`.

```
editor.restoreDOM()
```

To show this working, the PR includes an example where we change the state in the example without changing the `document` in Slate. In the example, I change the background color of the blocks when you click a button in the toolbar. Normally, this means that we won't see a change since the document hasn't changed (i.e. the state is outside of Slate).

GIF

As you can see in the example, the colors do change in all blocks. This is because after changing the state, I call `editor.restoreDOM()` which forces the editor to blow away the DOM and start again.

This serves two purposes:

- It allows the entire DOM to re-render if you have state outside of Slate that has changed and you are okay taking a brute force approach to updating the DOM.
- The real reason though is to get Android working. Android messes with the DOM a lot and it is often not preventable. This means that React crashes the Editor often. By calling `restoreDOM`, we can ensure that the `render` will not crash the Editor. DraftJS uses this technique in order to prevent Android from crashing their editor as well.

It works internally by adding a `key` to the `content`. We increment the `key` by 1 when `restoreDOM` is called.

Doesn't fix but is part of the fix for https://github.com/ianstormtaylor/slate/issues/2062

Reviewers: @ianstormtaylor

### DIFF
--- a/examples/app.js
+++ b/examples/app.js
@@ -27,6 +27,7 @@ import PlainText from './plain-text'
 import Plugins from './plugins'
 import RTL from './rtl'
 import ReadOnly from './read-only'
+import RestoreDOM from './restore-dom'
 import RichText from './rich-text'
 import SearchHighlighting from './search-highlighting'
 import Composition from './composition'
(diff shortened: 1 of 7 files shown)


ianstormtaylor:
Hey @thesunny thanks for looking into this issue and your work to get things working on Android. The `editor.restoreDOM()` seems like it could be a necessary piece. 

That said, can you remove this example? Restoring the DOM like this in an implementation detail and there’s no need for us to have a prime example of it for users. Especially when we factor in the maintenance burden. We need to be simplifying existing examples instead of adding new ones right now.

thesunny:
@ianstormtaylor Thanks for the input. Sure, I'll remove it when I get a chance.

PR #2786, opened 2019-05-16
fix: passing Boolean to non-Boolean attribute

jportella93:
This fixes a console warning caused by passing a Boolean to a non-Boolean attribute `'active'` from the DOM element `button`. The attribute expects a string containing `'true'` or `'false'`;

```
Warning: Received `false` for a non-boolean attribute `active`.

If you want to write it to the DOM, pass a string instead: active="false" or active={value.toString()}.

If you used to conditionally omit it with active={condition && value}, pass active={condition ? value : undefined} instead.
```

### DIFF
--- a/examples/rich-text/index.js
+++ b/examples/rich-text/index.js
@@ -124,7 +124,7 @@ class RichTextExample extends React.Component {
 
     return (
       <Button
-        active={isActive}
+        active={isActive.toString()}
         onMouseDown={event => this.onClickMark(event, type)}
       >
         <Icon>{icon}</Icon>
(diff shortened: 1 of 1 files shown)


thesunny:
@jportella93 

I'm a little confused about what's going on here and maybe you can help clarify. It appears that `active` is being passed to the `Button` component (a function) in `/slate/examples/components.js` which can be viewed here: https://github.com/ianstormtaylor/slate/blob/c9bbb64e785c0a2297d658f7e94c972037477e40/examples/components.js

In the component, the `active` prop is not being rendered to the DOM.

I might be missing something (it's happened before!) but I also wonder if this is addressing a bug that was fixed more recently?

Okay, I just browsed through the history of the page, I think this refactor about a week ago may have fixed the bug: https://github.com/ianstormtaylor/slate/commit/a5a25f97dde1de06ef0746e6f5285c04e503351c#diff-eb0c43e3f5842d30f5ba4e01ea254cd6

Let me know if you think that isn't the case.

YannisMarios:
Hello,

This issue still exists on latest version: 

image


image

PR #2790, opened 2019-05-16
Extract Instruction component into components for examples

thesunny:

Refactor an Instruction component so that it can be re-used by examples

#### What's the new behavior?

Same as before with a slight visual improvement. The main difference is that the `Instruction` component can now be re-used in the examples.

Moved code and made it fit style of other shared components.

Fixes: #
Reviewers: @


thesunny:
Also fixes a warning to do with the removal of leaves.

PR #2792, opened 2019-05-16
Add error boundary to save Slate from a crash on corrupt DOM

thesunny:

Feature / Improvement

#### What's the new behavior?

Usually, if React encounters a corrupted DOM, Slate crashes and is unrecoverable. This feature/improvement catches the error and re-renders the DOM. It appears to fix all corrupted DOM issues which is particularly important for Android.

Make no mistakes, this is a HUGE improvement. This means that small issues with DOM don't mean irrecoverable errors.

We add a React Error Boundary which catches the error and tries to recreate the component tree from scratch.

Fixes: #
Reviewers: @

### DIFF
--- a/examples/app.js
+++ b/examples/app.js
@@ -27,6 +27,7 @@ import PlainText from './plain-text'
 import Plugins from './plugins'
 import RTL from './rtl'
 import ReadOnly from './read-only'
+import RestoreDOM from './restore-dom'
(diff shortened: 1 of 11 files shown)


thesunny:
As discussed in this issue https://github.com/ianstormtaylor/slate/issues/2062#issuecomment-492320466

Issue #2788, opened 2019-05-16
Drag and drop broken (throwing errors) after leaf removal


cdunn:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Example code breaks as well: [https://www.slatejs.org/#/images](https://www.slatejs.org/#/images)
![May-16-2019 09-37-11](https://user-images.githubusercontent.com/20292/57871500-4f556980-77be-11e9-98fc-dde5271b56ac.gif)

#### What's the expected behavior?

`event.target` in https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/dom/before.js#L272 targeting leaf nodes that cannot be found with `findNode`.  Appears to be inter-related to other issues after leaf removal.

PR #2794, opened 2019-05-16
fallback findPath query to find closest element with a data-key

cdunn:
#### What's the new behavior?

May-16-2019 16-16-59

Moves to `findEventRange` query and adds a fallback to `findPath` that looks for the closest element with a `data-key` so it does not throw errors if dragging and element over a `data-slate-string` (amonst others).

Also, `document[iterable](path)` using `nextTexts` was undefined in `onDrop` so moved to `getNextText`

Fixes: #2788
Reviewers: @


ianstormtaylor:
Thanks for this @cdunn!

PR #2797, opened 2019-05-17
Fix findPoint with annotations

czechdave:
Fixing a bug. Current selection behavior with annotations present:

What the `findPoint` code does in this scenario is roughly this:
```js
const range = window.document.createRange()
range.setStart(textNode, 0)
range.setEnd(nearestNode, nearestOffset)
const contents = range.cloneContents()
offset = contents.textContent.length
```
now, the `contents` contain annotation's DOM nodes and any text rendered inside them is included in the `offset`. But this offset is being calculated for Slate's internal offset on a text node inside `value.document` which has no knowledge of annotations. So the offset is wrong and fallbacks to select the end of the block.

#### What's the new behavior?

If we remove the annotations DOM nodes before the `contents.textContent.length` runs. The offset is calculated correctly and selection works as expected.

Doesn't consider annotations DOM nodes when calculating native selection offset.

To elaborate more on the ugly selector:
```js
${SELECTORS.ANNOTATION} *:not(${SELECTORS.STRING})
```
this is necessary because of how annotations render. They render **around** `document` nodes. We need to have those to calculate the correct `offset` but we need to get rid of everything else that came out of `renderAnnotation`. This selector does exactly that. Select everything inside an annotation DOM node that is not coming from `document`. There might probably be some cleaner way to make this distinction?

Fixes an issue where Slate would fail to reconcile the native selection with internal selection because of annotation DOM nodes.

Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/constants/selectors.js
+++ b/packages/slate-react/src/constants/selectors.js
@@ -17,4 +17,5 @@ export default {
   TEXT: `[${DATA_ATTRS.OBJECT}="text"]`,
   VOID: `[${DATA_ATTRS.VOID}]`,
   ZERO_WIDTH: `[${DATA_ATTRS.ZERO_WIDTH}]`,
+  ANNOTATION: `[${DATA_ATTRS.OBJECT}="annotation"]`,
 }
--- a/packages/slate-react/src/plugins/react/queries.js
+++ b/packages/slate-react/src/plugins/react/queries.js
@@ -309,9 +309,12 @@ function QueriesPlugin() {
       range.setStart(textNode, 0)
       range.setEnd(nearestNode, nearestOffset)
       const contents = range.cloneContents()
-      const zeroWidths = contents.querySelectorAll(SELECTORS.ZERO_WIDTH)
+      const utilityNodes = contents.querySelectorAll([
+        SELECTORS.ZERO_WIDTH,
+        SELECTORS.ANNOTATION,
+      ])
 
-      Array.from(zeroWidths).forEach(el => {
+      Array.from(utilityNodes).forEach(el => {
         el.parentNode.removeChild(el)
       })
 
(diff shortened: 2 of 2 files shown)


ianstormtaylor:
Hey @czechdave can you explain this change?

czechdave:
Hey @ianstormtaylor, of course, here's a gif of the issue. If there's a better way to address this I'm all years.

Behavior on current `master` :
annotations selection bug

Behavior with this PR:
annotations work

ianstormtaylor:
@czechdave thanks that GIF is helpful. But can you explain why the code changes in this PR cause the changed behavior? What is it about the selector change that is necessary? What does the DOM look like that was different from the prior code's expectations?

czechdave:
OK, let me try to describe this in different words.

What the `findPoint` code does in this scenario is roughly this:
```js
const range = window.document.createRange()
range.setStart(textNode, 0)
range.setEnd(nearestNode, nearestOffset)
const contents = range.cloneContents()
offset = contents.textContent.length
```
now, the `contents` contain annotation's DOM nodes and any text rendered inside them is included in the `offset`. But this offset is being calculated for Slate's internal offset on a text node inside `value.document` which has no knowledge of annotations. So the offset is wrong and fallbacks to select the end of the block.

If we remove the annotations DOM nodes before the `contents.textContent.length` runs. The `offset` is calculated correctly and selection works as expected.

@ianstormtaylor let me know if this makes more sense.

I wonder if there might be another way to solve this inside the `normalizeNodeAndOffset` at the top of the `findPoint` function?
```js
    const { node: nearestNode, offset: nearestOffset } = normalizeNodeAndOffset(
      nativeNode,
      nativeOffset
    )
```

This change solved my problem so I haven't looked any deeper, feel free to guide me towards a better way to solve this if there is one.

ianstormtaylor:
@czechdave thank you! That makes sense. I like your "masking" approach there. 

Assuming in this case that you're talking about having rendered extra content inside an annotation (presumably with `contenteditable="false"`), here's a question then... with this logic we're only singling out annotations. Should it instead be updated to handle all `decoration/mark/annotation/text` cases? 

In other words, could we instead change the selector to something like:

```js
`${SELECTORS.STRING}`
```

And then just loop over only the actual string content to get the text content? Would that successfully handle any of the rendering cases where stuff is put adjacent to Slate string content?

ianstormtaylor:

Issue #2769, opened 2019-05-13
Moving cursor into/out of an inline sometimes causes Node.assertNode to raise an error during selectionchange


Jarred-Sumner:
#### Do you want to request a _feature_ or report a _bug_?
:bug: Bug

#### What's the current behavior?

On `selectionchange`, when the cursor moves focus on the start or end of an `inline`, _most_ of the time, the below exception occurs. It seems to happen more often when the inline is at the end of a block and the cursor moves to the next line, but it doesn't happen 100% of the time.

The editor seems to continue to work, but gets slower and slower.

Gif (notice the exception appearing when the cursor moves from the end of the link to the next block)
![gif](https://user-images.githubusercontent.com/709451/57599648-48073500-750c-11e9-964b-49bb1cf6c3af.gif)

Exception:
```
slate.es.js:14890 Uncaught Error: `Node.assertNode` could not find node with path or key: List [ 1, 1, 0 ]
    at Block.ElementInterface.<computed> [as assertNode] (slate.es.js:14890)
    at Block.createIterable (slate.es.js:12063)
    at Block.descendants (slate.es.js:12270)
    at Block.texts (slate.es.js:14225)
    at findSelection (slate-react.es.js:2245)
    at onQuery (slate.es.js:8560)
    at next (slate.es.js:10500)
    at onQuery (slate.es.js:8559)
    at next (slate.es.js:10500)
    at onQuery (slate.es.js:8559)
    at next (slate.es.js:10500)
    at onQuery (slate.es.js:8559)
    at next (slate.es.js:10500)
    at Editor.run (slate.es.js:10529)
    at Editor.query (slate.es.js:10402)
    at Editor$$1.method [as findSelection] (slate.es.js:10461)
    at onSelect (slate-react.es.js:4805)
    at next (slate.es.js:10500)
    at onSelect (slate-react.es.js:5291)
    at next (slate.es.js:10500)
    at Editor.run (slate.es.js:10529)
    at Editor$$1.run (slate-react.es.js:5654)
    at Object.onEvent (slate-react.es.js:5544)
    at HTMLDocument.<anonymous> (slate-react.es.js:1418)
    at invokeFunc (debounce.js:97)
    at HTMLDocument.debounced (debounce.js:177)
```

With debugger on:

![image](https://user-images.githubusercontent.com/709451/57599403-67519280-750b-11e9-923c-f38ccc5cb094.png)

#### What's the expected behavior?

[...]

Issue #2779, opened 2019-05-15
Inline selection is broken (slate@0.47)


vashchukmaksim:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Run **examples/links** and try to select "hyperlinks" link in the editor area. It throws an error:

```js
element.js:2583 Uncaught Error: `Node.assertNode` could not find node with path or key: List [ 0, 1, 0 ]
    at Block.ElementInterface.<computed> [as assertNode] (element.js:2583)
    at Block.createIterable (element.js:165)
    at Block.descendants (element.js:365)
    at Block.texts (element.js:1875)
    at findSelection (queries.js:486)
    at onQuery (queries.js:36)
    at next (editor.js:304)
    at onQuery (queries.js:35)
    at next (editor.js:304)
    at onQuery (queries.js:35)
```

![Screen Recording 2019-05-15 at 12 43 07](https://user-images.githubusercontent.com/26834615/57766153-86207680-770f-11e9-9e01-63f13710a9f4.gif)

I catch this error at the same situation in my own app, then checked online demo and then cloned slate monorepo and ran example locally. The error is presented everywhere.

#### What's the expected behavior?

No errors

adjourn:
Im pretty sure it was fixed by #2764 (not released yet). [Test it here](https://deploy-preview-2764--slatejs.netlify.com/#/links).

vashchukmaksim:
@adjourn Unfortunately it is still there

adjourn:
Ah yes, managed to replicate it with double click. Doesn't seem to be related to #2764.

vashchukmaksim:
Yeah, now only double-click throws this error but part selection seems to be fixed with #2764

audunhalland:
Duplicate of #2769

adjourn:
Is it fixed and can be closed now? #2799 "fixes" was not correctly used.

vashchukmaksim:
Yes, it seems that it's working properly with those fixes. I'm closing this issue then.

PR #2799, opened 2019-05-17
Use relative path to find a node's text nodes

justinweiss:

Fixing a bug.

#### What's the new behavior?

We no longer see errors when the selection endpoints are at the beginning or end of inlines.

When `block.texts()` is passed a path, it treats that path as relative to `block`. `findSelection` passes it the selection path, which is relative to `document`.

In order to find the correct text nodes, we must convert the selection path to a path relative to `block`. Then, when a new path is returned, we need to convert that new block-relative path back to a document-relative path.

Fixes: #2769 #2779

### DIFF
--- a/packages/slate-react/src/plugins/react/queries.js
+++ b/packages/slate-react/src/plugins/react/queries.js
@@ -483,11 +483,14 @@ function QueriesPlugin() {
       anchor.offset === anchorText.text.length
     ) {
       const block = document.getClosestBlock(anchor.path)
-      const [next] = block.texts({ path: anchor.path })
+      const depth = document.getDepth(block.key)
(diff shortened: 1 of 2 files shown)


audunhalland:
I'm not a contributor to slate, but a user of the library, and affected by this bug. Would it be possible to add a test, since none of the existing tests were able to expose this bug?

justinweiss:
I would love to, but I haven't been able to find any examples of how to test this code.

audunhalland:
Still, eagerly awaiting this fix and a new bugfix release :)

ianstormtaylor:
Good catch, thanks @justinweiss!

Issue #2783, opened 2019-05-15
Executing `editor.moveToRangeOfDocument().delete()` twice throws an error


olabala:
#### Do you want to request a _feature_ or report a _bug_?
BUG

#### What's the current behavior?
After executing `editor.moveToRangeOfDocument().delete()` twice, the editor will stop work!!!
Here is a example https://codesandbox.io/s/slarereactbutwithselection-j6u6x, as u can see, after u click `enter` key twice, then focus editor, and u'll get an error in the console panel when u try to input something.
```
slate.js:3179 Uncaught TypeError: Cannot read property 'equals' of null
    at Point.moveTo (VM6529 slate.js:3179)
    at eval (VM6529 slate.js:14979)
    at Selection.updatePoints (VM6529 slate.js:15172)
    at Selection.moveTo (VM6529 slate.js:14978)
    at deleteExpandedAtRange (VM6529 slate.js:6106)
    at Commands$1.insertTextAtRange (VM6529 slate.js:6970)
    at Editor.command (VM6529 slate.js:11290)
    at Editor.command (VM6683 slate-react.js:5387)
    at onCommand (VM6529 slate.js:5855)
    at next (VM6529 slate.js:11491)
    at Editor.run (VM6529 slate.js:11525)
    at Editor.command (VM6529 slate.js:11297)
    at Editor.method [as insertTextAtRange] (VM6529 slate.js:11418)
    at Commands$6.insertText (VM6529 slate.js:10735)
    at Editor.command (VM6529 slate.js:11290)
    at Editor.command (VM6683 slate-react.js:5387)
    at onCommand (VM6529 slate.js:5855)
    at next (VM6529 slate.js:11491)
    at Editor.run (VM6529 slate.js:11525)
    at Editor.command (VM6529 slate.js:11297)
    at Editor.method [as insertText] (VM6529 slate.js:11418)
    at onBeforeInput (VM6683 slate-react.js:2258)
    at next (VM6529 slate.js:11491)
    at onBeforeInput (VM6683 slate-react.js:2928)
    at next (VM6529 slate.js:11491)
    at Editor.run (VM6529 slate.js:11525)
    at Editor.run (VM6683 slate-react.js:5436)
    at Object.onEvent (VM6683 slate-react.js:5107)
    at Content.onEvent (VM6683 slate-react.js:4931)
    at obj.<computed> (VM6683 slate-react.js:4601)
[...]

PR #2800, opened 2019-05-17
Allow deleteAtRange with a zero-length range on the first character

justinweiss:

fixing a bug.

#### What's the new behavior?

Calling `deleteAtRange` with a collapsed selection at the beginning of a text node will both null out the selection and delete the entire text node.

deleteAtRange will consider a zero-length range on the first character of a text node as a hanging selection, which is incorrect. This should not be considered hanging.

Note: It's still possible to hit the [`startKey === endKey && isHanging`](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/commands/at-range.js#L166) conditional if endKey is in a void node, since we will bump the selection to the previous node and update endKey (but then endOffset is no longer 0, so it's not _really_ hanging anymore). We have an existing test for that, and it still passes after this change. 

I thought this was cleaner than recalculating `isHanging` after bumping `endKey` / `endOffset`, since we'd still want to delete the entire text node in that case.

Fixes: #2783

### DIFF
--- a/packages/slate/src/commands/at-range.js
+++ b/packages/slate/src/commands/at-range.js
@@ -114,7 +114,8 @@ Commands.deleteAtRange = (editor, range) => {
     endOffset === 0 &&
     isStartVoid === false &&
     startKey === startBlock.getFirstText().key &&
-    endKey === endBlock.getFirstText().key
+    endKey === endBlock.getFirstText().key &&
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Good point there, thanks @justinweiss!

Issue #2746, opened 2019-05-09
treat "hanging" ranges in a consistent way


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Bug / improvement.

#### What's the current behavior?

Rich text editors depend on a special behavior when a range spans multiple blocks and the end is right at the beginning of a block. For example:

```jsx
<document>
  <paragraph><anchor />Some text...</paragraph>
  <paragraph><focus />More text.</paragraph>
</document>
```

We often need to do special behaviors, like ignore the extra "hanging" block when iterating over the blocks. This happens with inlines too, like:

```jsx
<document>
  <paragraph><anchor />one <link><focus />two</link> three</paragraph>
</document>
```

#### What's the expected behavior?

We should add first-class support for "hanging" selections like this. The iterables like `node.blocks()` and `node.inlines()` should have a `includeHanging` option to toggle this, defaulting to `false`. Other parts of the codebase will then get this behavior automatically.

For the cases where people do want the extra "hanging" nodes, they can opt-in by passing `includeHanging: true` and get them, but I think it would be the exception.

justinweiss:
I'm not sure if it's the same issue (or would be fixed by the fix for this), but I've run into hanging problems especially when triple-clicking. Here's an example:

![Screen Recording](https://user-images.githubusercontent.com/1020/64574208-65fe6e00-d323-11e9-80cb-c805a078ec23.gif)

It's hard to know what the right thing to do is here. Imagine the "Human" cells are empty. Is a selection hanging into the other node, or is the person trying to select two nodes? I don't think there's a way to tell the difference. I think some editors solve this with dummy nodes at the end of each block to capture a hanging selection and report it as "hanging," to do the right thing. Considering how much code deals with zero-width spaces for void inlines, I don't know if we want to pile another of that kind of hack on top 😄 

Let me know if you'd rather this be a separate issue, this seemed like the right place to start bringing it up.

ianstormtaylor:
@justinweiss seems like as good a place as any! Would be nice to have all of the different edge cases related to hanging selections in one spot.

I'm thinking that "the person was trying to select a hanging selection" might be a case that can be thrown out. As long as hanging selections are handled in a known way, it might be okay to handle them that way even when the person tried to select it. For example, if it was nudged back to be non-hanging there, that seems reasonable?

justinweiss:
When the hanging selection spans multiple table cells, I think nudging it back is the right thing to do.

[...]

Issue #2711, opened 2019-04-26
remove model-level mutation methods


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Debt.

#### What's the current behavior?

Right now when an operation is applied, it calls down into model-level mutation methods like:

```js
Value.insertText
  -> Element.insertText
      -> Text.insertText
```

The idea here is that we continue to provide useful low-level mutator methods for folks when they need them. But it also causes a lot of indirection which makes it hard to reason about what happens in individual operations. 

I *think* it might also be costing us a bit of performance, as certain bits of logic end up getting repeated in different places, when it could all be handled in one swoop.

As we transition off Immutable (https://github.com/ianstormtaylor/slate/issues/2495), I think it will be more important to do all of the mutation logic in centralized places, with `immer`. And it will already help clear things up now too I think. 

My guess is that folks aren't really using any of these methods—except maybe in unit tests which should be more integration-level—and we can pretty safely remove them.

---

Specifically, this would remove the following methods:

```
Value:
  addMark
  insertNode
  insertText
  mergeNode
  moveNode
  removeMark
  removeNode
  removeText
  setNode
  setMark
  splitNode

Element:
  addMark
  insertNode
  insertText
  mergeNode
  moveNode
  removeMark
  removeNode
  removeText
  setNode
  setMark
  splitNode

Text:
  addMark
  insertText
  removeMark
  removeText
  setMark
```

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/3093.

Issue #2321, opened 2018-10-25
add `element.hasProperties` method


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Feature.

#### What's the current behavior?

It's common to want to check whether an element (inline/block) matches a specific set of properties. The naive way is to just check `element.type === type`, but that doesn't account for things like headings where you might have a  `data.level === 2` to check as well.

#### What's the expected behavior?

We should add a method that makes this easy to check. It should take `Object|String`, and use `Block.createProperties` to normalize them. And then check to see if the element has the properties in question.

```js
block.hasProperties('quote')
inline.hasProperties({ type: 'inline', data: { url: 'https://example.com' }})
```

I think we should also add a `getProperties` that returns just the settable properties of the element (ie. `type` and `data`), omitting the other things that are present when fully serializing.

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/3093.

Issue #2002, opened 2018-07-27
expose `path` to schema normalize functions


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the expected behavior?

Following from https://github.com/ianstormtaylor/slate/pull/1993#issuecomment-408546270 it would be good if normalization could expose the `path` of the node to `rule.normalize` functions, allowing them to use the faster `*ByPath` change methods.

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/3093.

PR #2809, opened 2019-05-21
remove more core `key` usage

ianstormtaylor:
This removes more of the `key` usage in Slate's core. We'll need to remove all of it in order to get to #2495, so this is one step in that direction.

---

###### BREAKING

**Slate's handler signatures have changed.** Previous there were inconsistencies in the ordering of arguments to Slate handlers, which were separated into "commands", "queries" and "middleware". These inconsistencies added complexity, and the distinctions between handlers wasn't actually necessary. Now there is only a single handlers concept.

```js
// Before, for middleware...
onKeyDown: (...args, editor, next) => {}

// Before, for commands and queries...
insertImage: (editor, ...args) => {}
isActive: (editor, ...args) => {}

// After, for all handlers...
handler: (fn, editor) => (...args) => {}
```

Not only has the signature changed, but the "middleware stack" has been removed too, in favor of function composition. So the `fn` argument to handlers is actually the old function that existed that the handler is overriding. If you want to pass through to it, you call it just like a normal function:

```js
// Before, deferring to the stack...
middleware: (...args, next) => {
  next()
}

// After, deferring to the wrapped function...
handler: (fn, editor) => (...args) => {
  return fn(...args)
}
```

**Slate commands and queries are no longer chainable.** As part of streamlining the handlers, we need to be able to return values from them. And so we have to remove the old logic that always returned `this` for chaining.

**Calling unknown handlers on the editor will now throw.** Previously these would just do nothing silently, but now they will throw an error saying that the handler was not registered by any of the plugins.

---

###### NEW

**The new `editor.exec` method replaces old handler invoking.** This is the new method for calling handlers dynamically using a `(name, ...args)` signature. All handlers are still exposed directly as methods on the editor as well.

```js
editor.exec('insertText', ...)
```

---

###### DEPRECATED

**The `editor.run`, `editor.query` and `editor.command` methods are deprecated.** All of these have been replaced by the new `editor.exec` which can return values, make changes, and respond to events.

**All of the getters on `Value` objects are deprecated.** All of the getter shorthands on `Value` objects are now deprecated. The full list includes:

**All of the manipulation methods on `Value` objects are deprecated.** Previously you could call these methods and receive a new immutable value object as a return value. The full list includes:

```
value.addAnnotation()
value.addMark()
value.insertNode()
value.insertText()
value.mergeNode()
value.moveNode()
value.removeAnnotation()
value.removeMark()
value.removeNode()
value.removeText()
value.setAnnotation()
value.setMark()
value.setNode()
value.setProperties()
value.setSelection()
value.splitNode()
```

To achieve this now you'll have to use `editor.applyOperation` instead. This ensures that all changes to a value go through the operations pipeline and can be normalized by a schema. This was an advanced feature and this change won't affect most people.

---

Fixes: https://github.com/ianstormtaylor/slate/issues/1883
Fixes: https://github.com/ianstormtaylor/slate/issues/2321
Fixes: https://github.com/ianstormtaylor/slate/issues/2812
Fixes: https://github.com/ianstormtaylor/slate/issues/2746
Fixes: https://github.com/ianstormtaylor/slate/issues/2711
Fixes: https://github.com/ianstormtaylor/slate/issues/2985
Fixes: https://github.com/ianstormtaylor/slate/issues/2858
Fixes: https://github.com/ianstormtaylor/slate/issues/2002


BrentFarese:
@ianstormtaylor do you have any idea when this PR will be merged into master?  We are working on a complex Slate editor implementation right now and it seems like a number of breaking changes are in here.  Wondering when they will hit...

Thanks.

BrentFarese:
@ianstormtaylor thank you!  If that branch is public or referenced in an issue somewhere, it'd be great if you can let me know here and I can take a look.

ianstormtaylor:
@BrentFarese it's not public nope. We've talked a little about it in the Slack contributing channel but that's it.

ianstormtaylor:
Closing in favor of https://github.com/ianstormtaylor/slate/pull/3093.

PR #2814, opened 2019-05-22
Add noop plugin

thesunny:

Adding a feature

#### What's the new behavior?

There is know way to know how the browser behaves natively within the context of Slate's DOM.

The `noop` plugin which is enabled by adding debug to localStorage (like all other debug) with a value of `slate:noop`.

All of the debug plugins will run before `noop` Plugin but after that, all the events aren't passed through the rest of Slate's stack. This means Slate's `value` is never updated and the Editor is never re-rendered.

The plugin simple doesn't call `next` on any of the `EVENT_HANDLERS` defined for Slate.

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/plugins/debug/noop.js
+++ b/packages/slate-react/src/plugins/debug/noop.js
@@ -1,0 +1,41 @@
+import EVENT_HANDLERS from '../../constants/event-handlers'
+
+/**
+ * A plugin immediately after any Debug plugins that prevents events from
+ * going any further.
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
@thesunny sorry, but this does not seem like something I want in Slate. The way of enabling it by adding `slate:noop` to Local Storage is very counterintuitive. Whatever the goal this is trying to solve needs to be done in another way.

PR #2815, opened 2019-05-22
Debug mutations

thesunny:

Add a feature

#### What's the new behavior?

Add ability to debug mutations in a Slate Editor.

Adds a MutationObserver at instantiation

Fixes: #
Reviewers: @


ianstormtaylor:
@thesunny similarly, this is extremely hacky code and it doesn't seem like something I'd like in Slate as is. In the future can you pull request these but then wait for a review before merging? 

I understand the idea of wanting to debug DOM mutations, which sounds like a helpful thing for moving towards using `MutationObserver`, but there are many things here that are done in a janky way that we need to find a better way to solve.

packages/slate-react/src/plugins/debug/debug-mutations.js
```diff
@@ -0,0 +1,86 @@
...
+
+const debug = Debug('slate:mutations')
+
+/**
+ * Options for MutationObserver
+ *
+ * @type {Object}
+ */
+
+const observeOptions = {
```
ianstormtaylor: These are only used in one place, can you inline it instead?

packages/slate-react/src/plugins/debug/debug-mutations.js
```diff
@@ -0,0 +1,86 @@
...
+  characterDataOldValue: true,
+}
+
+/**
+ * Properties on a MutationRecord
+ *
+ * @type {Object}
+ */
+
+const keys = [
```
ianstormtaylor: Can you name this as constant case with a name like `MUTATION_PROPERTIES`?

packages/slate-react/src/plugins/debug/debug-mutations.js
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Takes a mutation record and turns it into an Object that is easy to log to
+ * the console. Notably, it removes `null` and `undefined` values as well as
+ * values that are a `NodeList` that has no entries.
+ *
+ * @param {MutationRecord} mutationRecord
+ */
+
+function normalize(mutationRecord) {
```
ianstormtaylor: This is also only used in one place, can you inline it instead?

packages/slate-react/src/plugins/debug/debug-mutations.js
```diff
@@ -0,0 +1,86 @@
...
+ * <https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver>
+ * <https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord>
+ *
+ * @param {Object} options
+ */
+
+function DebugMutationsPlugin({ __editor__ }) {
+  const observer = new window.MutationObserver(flush)
+
+  function flush(mutations) {
```
ianstormtaylor: This is also only used in one place I think, can you inline it?

packages/slate-react/src/plugins/debug/debug-mutations.js
```diff
@@ -0,0 +1,86 @@
...
+    const array = Array.from(mutations).map(normalize)
+    debug(`${array.length} MutationRecord`)
+    // Used `console` because `debug` was not easy to view succinctly
+    window.console.log(...array)
```
ianstormtaylor: Doesn't `debug()` use `console.log` under the covers anyways? Can you post a screenshot of the difference. I think we should consolidate to only have one way to log debug output. Without checking I'd assuming that we could just put the entire array in and let people expand it in the console.

packages/slate-react/src/plugins/debug/debug-mutations.js
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * More information about mutations here:
+ *
+ * <https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver>
+ * <https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord>
+ *
+ * @param {Object} options
+ */
+
+function DebugMutationsPlugin({ __editor__ }) {
```
ianstormtaylor: Can you rename this to `editor`?

Issue #2736, opened 2019-05-05
Disabling mark with no selection doesn't cause it to stop being active


glebtv:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?
https://www.slatejs.org/#/rich-text
Enable "bold"
Type some text
Disable bold (don't select anything), button in toolbar shows as not active
Type some text, button in toolbar becomes active again
Second part of the text (after disabling bold) is bold, and bold button is re-highlighted in toolbar

#### What's the expected behavior?
Second part of the text (after disabling bold) is *not* bold

Possibly, ```removeMark``` should now do a ```split_node``` if the mark was active and the selection is collapsed?

Note: [gitbook slate rich text demo](https://gitbookio.github.io/slate/#/rich-text) doesn't have this problem

rebirthtobi:
I have the same problem, its like slate is broken at some point

rebirthtobi:
@glebtv Gitbook might be using a different version

isubasti:
this bug is actually different then what #1113 try to fix. it's a bug introduced due to seemingly incomplete implementation of removing leaves https://github.com/ianstormtaylor/slate/commit/a220cd5ae10c7e5f7df6d018deae0670526dbb73. @ianstormtaylor it looks like on `remove/adding` mark, `Text` node is not created/removed properly

mwegman:
I ended up downgrading to 0.45 to avoid this bug so it was definitely something introduced in 0.46.

rebirthtobi:
do you also downgrade slate react also?

mwegman:
> do you also downgrade slate-react also?
> 
> > I ended up downgrading to 0.45 to avoid this bug so it was definitely something introduced in 0.46.

I didn't. It seems to work fine without downgrading slate-react

rebirthtobi:
I'm still experiencing the same issue probably because my editor lose focus as i click on the button

> > do you also downgrade slate-react also?
> > > I ended up downgrading to 0.45 to avoid this bug so it was definitely something introduced in 0.46.
> 
> I didn't. It seems to work fine without downgrading slate-react

rebirthtobi:
I've resolve this using `onMouseDown` instead of `onClick`, but thwn i have a colour picker, font and font size picker also. I'm still trying to make this work

brynshanahan:
[...]

PR #2818, opened 2019-05-22
Fix mark toggle bugs

adjourn:

Bug fix. Most mark related commands are broken (e.g `toggleMark`, `insertText*`, etc) because `insertTextByPath` fails to split node and remove marks in most situations.

#### What's the new behavior?

When text is inserted into node and they have different marks (current node marks _vs_ inserted text marks), node is correctly split and marks are applied to new node.

```jsx
// Starting state
<paragraph>
  <text marks=["bold", "italic"]>
    Hello Worl<selection marks=["bold", "italic"] />
  </text>
</paragraph> 

// Bold and Italic are disabled and character "d" is insterted

// Current Slate
<paragraph>
  <text marks=["bold", "italic"]>
    Hello World<selection marks=[] />
  </text>
</paragraph> 

// After this PR
<paragraph>
  <text marks=["bold", "italic"]>
    Hello Worl
  </text>
  <text marks=[]>
    d<selection marks=[] />
  </text>
</paragraph> 
```

Currently `insertTextByPath` invokes `addMarksByPath` only when marks are provided.

If no marks are provided (optional argument) and text is inserted into node with marks, `addMarksByPath` isn't invoked, node doesn't get split and characters are inserted into same node. Additionally command fails in many other scenarios if you combine more than one type of marks.

Even worse, it's usually unclear which marks to add or remove because `marks` doesn't carry that information. Fortunately I found a way without changing command or marks signature.

---

Compare current Text node marks and incoming marks argument:

* if equal, don't do anything, marks stay the same
* if not equal
  * if incoming marks is empty Set, remove all marks (a way to enforce mark-less text)
  * if incoming marks have more entries than node marks, add difference marks
  * if node marks have more entries than incoming marks, remove difference marks
  * if same amount of marks but different types, add incoming marks

Fixes: #2736

### DIFF
--- a/docs/reference/slate/commands.md
+++ b/docs/reference/slate/commands.md
@@ -426,7 +426,7 @@ Insert a [`Fragment`](./fragment.md) at `index` inside a parent [`Node`](./node.
 `insertTextByKey(key: String, offset: Number, text: String, [marks: Set]) => Editor`
 `insertTextByPath(path: List, offset: Number, text: String, [marks: Set]) => Editor`
 
-Insert `text` at an `offset` in a [`Text Node`](./text.md) by its `key` with optional `marks`.
+Insert `text` at an `offset` in a [`Text Node`](./text.md) by its `key` with optional `marks`. [`Text Node`](./text.md) marks apply to inserted `text`, you can avoid that by passing empty `Set` as `marks`.
 
 ### `mergeNodeByKey/Path`
 
--- a/packages/slate/src/commands/by-path.js
+++ b/packages/slate/src/commands/by-path.js
@@ -158,6 +158,7 @@ Commands.insertTextByPath = (editor, path, offset, text, marks) => {
       if (node.marks.size > markSet.size) {
         const remove = node.marks.subtract(markSet)
         editor.removeMarksByPath(path, offset, text.length, remove)
+        return
       }
 
       // If equal sizes but not equal marks, marks were added.
--- a/packages/slate/src/commands/by-path.js
+++ b/packages/slate/src/commands/by-path.js
@@ -106,8 +106,6 @@ Commands.insertNodeByPath = (editor, path, index, node) => {
  */
 
 Commands.insertTextByPath = (editor, path, offset, text, marks) => {
-  // `marks` and `markSet` are different variables because we
-  // allow empty `marks` Set in order to insert text without any marks.
   const markSet = Mark.createSet(marks)
   const { value } = editor
   const { annotations, document } = value
@@ -141,25 +139,29 @@ Commands.insertTextByPath = (editor, path, offset, text, marks) => {
       text,
     })
 
-    if (!node.marks.equals(markSet)) {
-      if (offset + text.length < node.text.length) {
-        editor.splitNodeByPath(path, offset + text.length)
+    if (marks && !node.marks.equals(markSet)) {
+      // If marks argument is empty Set, remove all marks.
+      if (markSet.size === 0) {
+        const remove = node.marks.union(markSet)
+        editor.removeMarksByPath(path, offset, text.length, remove)
+        return
       }
 
-      if (offset > 0) {
-        editor.splitNodeByPath(path, offset)
-        path = PathUtils.increment(path)
+      // If more incoming marks, marks were added.
+      if (markSet.size > node.marks.size) {
+        const add = markSet.subtract(node.marks)
+        editor.addMarksByPath(path, offset, text.length, add)
+        return
       }
 
-      if (marks) {
-        // TODO: diff old and new marks and only remove/add based on that.
-        // Not sure if >performant, hints: Set.subtract(), Set.intersect().
-        editor.removeAllMarksByPath(path)
-
-        if (markSet.size > 0) {
-          editor.addMarksByPath(path, 0, text.length, markSet)
-        }
+      // If more node marks, marks were removed.
+      if (node.marks.size > markSet.size) {
+        const remove = node.marks.subtract(markSet)
+        editor.removeMarksByPath(path, offset, text.length, remove)
       }
+
+      // If equal sizes but not equal marks, marks were added.
+      editor.addMarksByPath(path, offset, text.length, markSet)
     }
   })
 }
--- a/packages/slate/src/commands/by-path.js
+++ b/packages/slate/src/commands/by-path.js
@@ -110,7 +110,7 @@ Commands.insertTextByPath = (editor, path, offset, text, marks) => {
   marks = Mark.createSet(marks)
   const { value } = editor
   const { annotations, document } = value
-  document.assertNode(path)
+  const node = document.assertNode(path)
 
   editor.withoutNormalizing(() => {
     for (const annotation of annotations.values()) {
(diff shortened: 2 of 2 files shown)


adjourn:
Alternative solution: if splitting node wouldn't transfer marks to new node, none of this would be needed.

erquhart:
@adjourn great work on this! Just a heads up that I had also been working on a solution and it differs enough from this one that I'll likely open a separate PR. But no failing tests! I'll link here when it's open, would love it if you could help test.

adjourn:
@erquhart superb, this fixes the issue but it also changes command behavior quite a bit.

erquhart:
Here it is: https://github.com/ianstormtaylor/slate/pull/2822

adjourn:
@erquhart tried all your tests.

This fails, mark is not removed. But is this a bug or feature? It only happens when said Text node is first and only in block and it feels natural (you can remove it with button). If there's previous block and `Backspace` is pressed one more time, mark is removed with block. I think it's similar to the question if marks should carry over to next block when `Enter` is pressed.

- [only-character-with-mark.js](https://github.com/ianstormtaylor/slate/blob/effba387613e4b376b24a10cfd7c1987b18e8f76/packages/slate/test/commands/at-current-range/delete-backward/only-character-with-mark.js)

These two fail for the same reason. Value is correct but selection moves to next Text node according to tests. This is a bit hazy because same situations work as expected in examples and I'm not sure if [this](https://github.com/ianstormtaylor/slate/commit/b3175ba565e8c307d35854bff80913e00ee78799#diff-dc737afa91245a08a36c966a9da47690R417) could have anything to do with this (i.e point can exist at the end of a text node, or at the start of the following one).

- [remove-middle-collapsed-selection.js](https://github.com/ianstormtaylor/slate/blob/effba387613e4b376b24a10cfd7c1987b18e8f76/packages/slate/test/commands/at-current-range/toggle-mark/remove-middle-collapsed-selection.js)

- [remove-some-collapsed-selection.js](https://github.com/ianstormtaylor/slate/blob/effba387613e4b376b24a10cfd7c1987b18e8f76/packages/slate/test/commands/at-current-range/toggle-mark/remove-some-collapsed-selection.js)

Example:

```diff
 "selection": {
   "anchor": {
     "object": "point"
-    "offset": 0
+    "offset": 1
     "path": [
       0
-      2
+      1
     ]
   }
   "focus": {
     "object": "point"
-    "offset": 0
+    "offset": 1
     "path": [
       0
-      2
+      1
     ]
   }
```

erquhart:
This is the closest we've gotten for sure 👍 As you mentioned in Slate, there are edge cases. Until the core issue is resolved, the best we can do is very nice hacks.

To reiterate the conversation from Slate for anyone reading here, the core issue here is the unclear intended role of the `marks` parameter for `insertTextByPath/Key` and other commands like it. Either the `marks` replace existing marks or they add to them, but they can't do both. And if we go with adding, we need to determine how to go about inserting text when a mark is disabled on the selection but not in the surrounding text node (eg. when selection is collapsed and `toggleMark` is called).

@ianstormtaylor any thoughts?

FerrielMelarpis:
Hi all!
I've just started using slate API and I'm not quite familiar with it yet but I'd like to know if there is a chain of command I can use to work around this `toggleMark` bug.
Any help is appreciated.
FWIW, I think Canner has solved this issue. Their [sample](https://canner.github.io/canner-slate-editor/?selectedKind=canner-slate-editor&selectedStory=Simple%20demo&full=0&addons=1&stories=1&panelRight=0) works fine.

isubasti:
@FerrielMelarpis I believe canner is still on much older version of slate. you can use slate<0.46.0 while waiting for this bug to be fixed

ianstormtaylor:
@erquhart good point, I think the `marks` argument to `insertText` should be strictly defining the entire set of new marks, not merging with the existing ones. In case that helps.

@adjourn is this one good to go? I haven't been keeping track of the changes.

erquhart:
The challenge is folks using `insertText` may not know what their inserting into, and will often want surrounding marks to apply, and Slate is sort of set up to work that way atm. So we'll at least need to provide a migration path if we're going to change how that method works, a second method, or some simplistic way to get a hold of the marks that would naturally be used.

Maybe we can leave `insertText` as is and add something like `insertTextAndSetMarks`?

adjourn:
@ianstormtaylor it still has one bug left which I'm not able to fix without breaking changes/changing APIs (would still be a hack): toggling multiple marks before other commands is buggy. This PR is a hacky fix to make current situation a lot better until we find a better way (didn't know it's so difficult to fix when I started).

It has 3 outcomes:
* empty Set of marks - insert text and don't merge with existing marks
* no Set of marks -  merge with existing marks
* Set of marks - add Set of marks and merge with existing ones

It's ready to go if you think it's a good idea.

I'd love to work on alternative PR to fix it properly but I just don't have the time right now. I suggest @erquhart to give it a shot if this PR doesn't work out. As he implied, merging is the root of this issue.

erquhart:
The problem is that this is a breaking change for anyone passing an empty Set to `insertTextByPath`, but the current release is already broken to the point that, if I understand correctly, folks are avoiding upgrading to it.

Unfortunately I completed migration of a very complex project to the latest version, including serializing to and from the new leaf-less model, _after_ discovering this bug 😅 So I'm all in.

@ianstormtaylor I agree with @adjourn that this PR is as good as it gets without an addition to the API. I think the best move is to merge this and note the breaking change to `insertTextByPath` and any methods that use it in regards to empty Sets removing all marks. Beyond that release a proper solution can be provided, likely through one or more new API methods.

sgurenkov:
@adjourn @erquhart I agree with @ianstormtaylor that it should not be a merge, it should set marks exactly how they were passed to the `insertTextByPath` method, that being said why not just:

`
if (!node.marks.equals(marks)) {
  editor
      .removeMarksByPath(path, offset, text.length, node.marks)
      .addMarksByPath(path, offset, text.length, marks)
}
`

sgurenkov:
@ianstormtaylor @erquhart @adjourn I've created a new PR #2936 to fix this issue, would be awesome if you can help testing it and getting this bug resolved soon as possible

mpkelly:
Hi, sorry to hijack this thread. I came here via #2822 which seems to describe a problem I am facing using the latest NPM build of slate and react-slate. I am implementing some basic markdown substitutions using regular expressions like so:

```JavaScript
const Bold = /(\*\*|__)(.*?)\1/;

const onSpace = (event, editor, next) => {
  const { value } = editor;
  const { selection } = value;
  if (selection.isExpanded) return next();
  const { startBlock, startText } = value;
  const match = startText.text.match(Bold);
  if (match) {
    // match = ["**text**", "**", "text", index: 44, input: "A line of text in a paragraph. this is some **text**", groups: undefined]
    editor
      .deleteBackward(match[0].length)
      .insertTextByKey(startText.key, match.index, match[2], [
        Mark.create({ type: "bold" })
      ])
      .removeMark("bold");
    return next();
  }
...
```
this substitution for bold works fine if the marked-up text is somewhere in the middle of the line but the bold mark remains if the substitution happens at the end of a line i.e. the last word goes bold as expected but the bold mark stays on when I continue typing. 

Can anyone confirm if this is a bug or if I am simply not using the API correctly? I've read through the docs and tried a bunch of things but nothing is working. 

Thanks
Mike

sgurenkov:
@mpkelly try my PR #2936 it might fix your problem, but weird that It works in the middle of the line, should be failing in both cases

mpkelly:
@sgurenkov thank you for your comment and for your PR. I can confirm that this resolves the issue I had (finally!). 

Regarding updates to the middle of a line, this might have worked for me because my markdown substitution is triggered after hitting the space key rather that using the toolbar.

bateau:
@ianstormtaylor Can you please take a look at this PR or #2936 to fix marks during insertText.

samuelcolvin:
is this now fixed via #2936?

erquhart:
Yeah, #2936 superseded this.

thesunny:
Closing this PR since #2936 resolves it and has been merged.

Let me know if this was in error.

Issue #2760, opened 2019-05-10
Editor component's placeholder prop is broken


franjohn21:
#### Do you want to request a _feature_ or report a _bug_?
🐛 

#### What's the current behavior?
When the `Editor` component from `slate-react` has the `placeholder` prop specified, the expected behavior is that the placeholder text will be rendered when there is no text entered. The current behavior is that the placeholder text is not being rendered in the latest slate-react.

This bug is verifiable in Slate's own examples, e.g. https://www.slatejs.org/#/rich-text where [the source code](https://github.com/ianstormtaylor/slate/blob/master/examples/rich-text/index.js#L102) specifies a `placeholder` prop, the empty state for the editor does not render the placeholder.

![Screen Shot 2019-05-10 at 2 18 44 PM](https://user-images.githubusercontent.com/7452924/57557833-5ef13000-7330-11e9-8e86-87c284b0d986.png)

#### What's the expected behavior?
The expected behavior is that the placeholder text is visible in the empty state. I am completely new to slate though so it's possible I'm misunderstanding what this is supposed to do.

franjohn21:
👋 @ianstormtaylor think we could get some eyes on https://github.com/ianstormtaylor/slate/pull/2765 when you gets a chance? IMO the `placeholder` is fairly critical to maintain feature parity with native text inputs

ianstormtaylor:
Thanks @franjohn21, I totally missed that one!

PR #2819, opened 2019-05-22
Fix placeholder not rendering

adjourn:

Fix bug.

#### What's the new behavior?

If decoration/annotation has both start and end offsets `0` and leaf content length is `0`, decoration/annotation is still added to leaf. 

This makes possible to add decoration/annotation that is content in itself, versus decorating the content that's in leaf. This fixes a bug where placeholder was not added to leaf and was not rendered.

Fixes: #2760 
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react-placeholder/src/index.js
+++ b/packages/slate-react-placeholder/src/index.js
@@ -70,10 +70,10 @@ function SlateReactPlaceholder(options = {}) {
    * @return {Element}
    */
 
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thanks for digging into this @adjourn!

zarv1k:
@ianstormtaylor could you please publish `slate-react-placeholder` version with this fix?

Issue #2548, opened 2019-01-19
Proposal to attach `window` to `editor`


thesunny:
#### Do you want to request a _feature_ or report a _bug_?

Proposal for an improvement.

#### What's the current behavior?

There are 8 results for `get-window` in the Slate repository. In 6 of those cases, the `editor` object is in scope. In the Android support I'm working on, there are another 10 or so cases where we need to use `get-window` and an `editor` is in scope.

#### What's the expected behavior?

My proposal is to add the `window` object to the `editor`. I am assuming that once the `editor` is created, the `window` would never change.

The benefits are:

- Improve performance
- Reduce complexity
- Reduce the need to pass the `window` object separately

@ianstormtaylor I can make a PR but I wanted to make sure there wasn't a reason for not doing this already. For example, maybe it is possible for the editor to change the `window` its attached to.

thesunny:
Fair enough about the performance improvement.

I like the idea of adding the element to the editor as `editor.el`. Makes sense since you can get the `window` from the `el` but not the other way around.

PR #2820, opened 2019-05-22
Add div to editor as editor.el

thesunny:

Adding a feature

#### What's the new behavior?

As per https://github.com/ianstormtaylor/slate/issues/2548 this PR adds the editor's DOM element to the `editor` object.

`content.js` uses a `ref` prop with `this.ref` which was a `React.createRef()` in order for other methods like `updateSelection` to access the DOM element within `content.js` itself.

I switched the `ref` prop to `this.setRef`.

Within `this.setRef`, I set both `this.ref.current` and `editor.el` to the DOM element.

This maintains backwards compatibility but adds the `editor.el` property.

Fixes: #2548
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -112,6 +112,17 @@ class Content extends React.Component {
   ref = React.createRef()
 
   /**
+   * Set both `this.ref` and `editor.el`
+   *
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @thesunny!

PR #2822, opened 2019-05-22
Fix mark removal at the edges of text nodes

erquhart:

Bugfix

#### What's the new behavior?
Toggling a mark off at the end of a text node and then continuing typing currently results in the mark becoming active again. This can also result in an empty editor with active marks, which can only be removed by selecting marked text and then toggling marks off.

This fix restores the original behavior.

- Previously a set of marks could be passed to `insertTextByPath` for inclusion, but there was no way to ensure that _no_ marks were used, as an empty set was ignored. With this change, an empty set passed as the `marks` arg ensures that no marks are applied to the inserted text.
- `removeAllMarksByPath`, a currently unused and undocumented command, now has an extended signature to support the new behavior in `insertTextByPath`, optionally accepting an offset and text length so that marks can be removed at certain points within a node rather than from the entire node.
- `insertText`, as a higher level abstraction, will override this new behavior in `insertTextByPath` by giving preference to the marks collected from the selection, if any, as is this is the generally expected outcome as evidenced by Slate's tests.

Fixes: #2736

### DIFF
--- a/packages/slate/src/commands/by-path.js
+++ b/packages/slate/src/commands/by-path.js
@@ -106,10 +106,10 @@ Commands.insertNodeByPath = (editor, path, index, node) => {
  * @param {Set<Mark>} marks (optional)
  */
 
-Commands.insertTextByPath = (editor, path, offset, text, marks) => {
-  marks = Mark.createSet(marks)
+Commands.insertTextByPath = (editor, path, offset, text, marksArg) => {
   const { value } = editor
   const { annotations, document } = value
+  const marks = Mark.createSet(marksArg)
   document.assertNode(path)
 
   editor.withoutNormalizing(() => {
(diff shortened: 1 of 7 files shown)


adjourn:
Found one failing scenario:

* select a word
* apply multiple marks
* drop collapsed selection in the middle of word
* if you disable one mark and type, same bug appears
* only works if you disable all active marks

erquhart:
Good catch, adding a failing tests for that and taking a look now 👍

erquhart:
Update: bug discovered by @adjourn revealed the real issue here - text insertion commands provide a way to _add_ marks, but no way to _remove_ them. Both his and my PR's attempted to re-interpret an empty set of marks to mean insert text with no marks, which works only because an empty set previously had no impact. But removing one of multiple marks can't be accomplished through the text insertion commands using the current API.

I'm spinning up the latest version of Slate that doesn't have this bug to determine how this used to work before Slate migrated away from leaves, hopefully that will shed some light on a path forward.

adjourn:
@erquhart I managed to crack it, all test are passing. Do you mind if I borrow your commited tests?

erquhart:
Go for it!

adjourn:
@erquhart I managed to crack it, all test are passing. Do you mind if I borrow your commited tests?

I'm going to push latest working version in 15-30 minutes.

erquhart:
Go for it!

erquhart:
Pushing up some failing tests that I have locally as well.

erquhart:
Closing in favor of #2818.

PR #2824, opened 2019-05-23
Add noop plugin as detached plugin

thesunny:

Feature

#### What's the new behavior?

Added a NoopPlugin which eats events so that Slate can't react to them. This helps us see DOM mutations and events while ignoring the noise that comes from what Slate itself is doing to DOM mutations and adding events.

Using the list of EVENT_HANDLERS, it creates functions that do nothing. In particular, they don't call the `next` method. Unlike the previous rejected version, the Plugin is detached. It must be hard-coded into the plugin chain for it to work.

For those following, this approach was approved in a Slack chat by @ianstormtaylor 

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/plugins/debug/noop.js
+++ b/packages/slate-react/src/plugins/debug/noop.js
@@ -1,0 +1,45 @@
+import EVENT_HANDLERS from '../../constants/event-handlers'
+
+/**
+ * This plugin prevents events from going any further and is useful in dev.
+ *
+ * The purpose is to see how the editor events and mutations behave without
+ * the noise of the editor also adding its own events and mutations.
+ *
+ * IMPORTANT:
+ *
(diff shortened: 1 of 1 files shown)


icarusysuper:
@thesunny Hi thesunny, you should wait for a review before merging into master branch.

ianstormtaylor:
@icarusysuper Sunny and I discussed this one and a few others, so he’s good to go on these things.

icarusysuper:
@ianstormtaylor @thesunny Nice work and thanks a lot. 👍  I love slate and pay attention to new commits every day.

thesunny:
@icarusysuper thanks!

I realized that this merge may look like I was just remerging a rejected PR which is why I added the comment about the approach being approved by Ian. I probably should have said the **merge** was approved to be more clear.

Issue #2754, opened 2019-05-10
Update docs to remove leaves


rg4real:
As I got they are deprecated now so might get rid of them.

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #2831, opened 2019-05-25
Remove leaves from docs

knpwrs:
As a new user not being able to go through the walkthrough docs was somewhat harder than it need to be. Something else that could be improved is that the error message in the console told me what was wrong but not how to fix it.

This is a partial fix for #2754. I'd remove other references to leaves in the docs but I'm not yet up to speed on the project as a whole to understand the changes that need to be made.


ianstormtaylor:
Thanks you @knpwrs!

Issue #2832, opened 2019-05-25
Add `componentDidMount` and `componentDidUpdate` event to plugins


thesunny:
#### Do you want to request a _feature_ or report a _bug_?

Request a feature.

Note: Also looking for input on this feature request in case there is a better way to handle this without having to add this feature or to implement a feature that can provide the same benefits in a better way.

#### What's the current behavior?

There is no way to know when the Editor component has mounted or updated.

I have these scenarios where this is important in the Android implementation:

1. I need to set up a `MutationObserver` on the Editor's root element. Currently, in the plugin, I use a `requestAnimationFrame` during the plugin's initialization to add the mutation observer to the Editor's root element; however, this is hacky because (a) changes in implementation may mean that the root element might not exist after the `requestAnimationFrame` (possibly created later) (b) there may be a period of time between the component mounting and the `requestAnimationFrame` during which a mutation could come in and be missed.

2. When I call `editor.restoreDOM`, the root element of the Editor is discarded and rebuilt from scratch. Because it is built from scratch, the existing `MutationObserver` would be on an `Element` that is no longer in the DOM because it has been replaced by a newly created one. We could use `componentDidUpdate` to check then add the `MutationObserver` on the new object.

3. I need to ignore mutations that React is performing while paying attention to mutations that the user is doing. I can do this by setting a local variable `updating = true` before I call Editor commands (which modify the React DOM). All mutations that come in while `updating === true` are ignored so we don't react to React mutations. On the `componentDidUpdate` method, we set `updating = false` and we can pay attention to mutations again.

Note: The Android implementation is sensitive to timing issues and we need to make everything synchronous with respect to updates.

#### What's the expected behavior?

Create two or more additional events that hook into React's component life-cycle.

```js
const myPlugin = {
  componentDidMount(event, editor, next) {
    console.log(`Hey, the Editor's "Content" just mounted!`)
    next()
  },
  componentDidUpdate(event, editor, next) {
    console.log(`Hey, the component just updated!`)
    next()
  }
}
```

These could be valuable generally for when the developer needs to add some initialization to the Editor's root element. This would provide an easy way to hook into them synchronously during a mount or update. It could be argued that these are too low level for a common use case in which case we could remain them `unstable_componentDidMount` and `unstable_componentDidUpdate` in the style of React methods that are available but generally shouldn't be relied on.

thesunny:
@ianstormtaylor this provide more details as per the message I sent to you through Slack.

[...]

PR #2833, opened 2019-05-27
Add lifecycle event handlers

thesunny:

Adding a feature

#### What's the new behavior?

Adds Plugin events for the following React lifecycle events:

- componentDidMount
- componentDidUpdate
- componentWillUnmount

As per the issue discussed here: https://github.com/ianstormtaylor/slate/issues/2832

in `content.js` where the Editor contents are rendered, I added calls like `this.props.onEvent('onComponentDidMount', {target: this.ref.current})` for each lifecycle callback.

Note that the `event` object passed through has a `target` property similar to what one would see in an actual event. This makes it easy to get the DOM element that represents the Editor element. Since this is likely the primary use case for these events, it felt useful to add them.

### Development Notes

For the time being, I added these lifecycle events as `componentDidMount` as opposed to @brendancarney preference to go with the `unstable_componentDidMount` names but will be happy to rename if @ianstormtaylor agrees this is preferable. I defaulted to keeping it `componentDidMount` which is a guess based on a previous preference by @ianstormtaylor to keep the names clean.

Also, @brendancarney I'm fine changing this to another method to achieve the same goals if we find a better solution so I'm not disregarding your desire for a better option. I can take this out no problem if/when we find a better solution.

Fixes: #https://github.com/ianstormtaylor/slate/issues/2832
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -158,6 +158,8 @@ class Content extends React.Component {
     }
 
     this.updateSelection()
+
+    this.props.onEvent('onComponentDidMount', { target: this.ref.current })
   }
 
   /**
@@ -180,6 +182,8 @@ class Content extends React.Component {
         this.handlers.onBeforeInput
       )
     }
+
+    this.props.onEvent('onComponentWillUnmount', { target: this.ref.current })
   }
 
   /**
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Having the second argument be a pseudo-event doesn’t make sense to me. If we have any arguments on those handlers they should be the same arguments that are received by the handlers natively. Can you not retrieve the element in another way?

thesunny:
@ianstormtaylor I will remove the pseudo-event.

I can retrieve them through the `editor` which I can get through the plugin now and retrieve the dom element at path `[]`.

PR #2836, opened 2019-05-28
Corrected placeholder alignement in chrome

gracicot:
Fixing a bug

#### What's the new behavior?

The cursor will be positionned correctly under chrome.

Before, the cursor would have a slightly different position when the mouse is pressed under chrome.

before

This doesn't look significant until you apply some styling:

before

It works by adding a vertical alignement to the placeholder element so it's aligned to the text-top position.

No matter the size, the placeholder element is aligned with the text top, keeping the cursor position at the right place in chrome.

Here's a gif that shows the behavior in chrome after the fix:

after
 
Firefox don't seem to be affected by this problem or the fix since it's algorithm to place the cursor seem to behave differently.

### DIFF
--- a/packages/slate-react-placeholder/src/index.js
+++ b/packages/slate-react-placeholder/src/index.js
@@ -81,6 +81,7 @@ function SlateReactPlaceholder(options = {}) {
         maxWidth: '100%',
         whiteSpace: 'nowrap',
         opacity: '0.333',
+        verticalAlign: 'text-top',
         ...style,
       }
 


gracicot:
Another solution I had was to set the `<span data-slate-zero-width="n" data-slate-length="0">﻿` element to be `display: inline-block`. Tell me if that is better or not.

ianstormtaylor:
This seems smart to me, thanks @gracicot!

PR #2847, opened 2019-05-31
Fix several errors in the deprecated `getEventRange`

justinweiss:

Fixing a bug.

#### What's the new behavior?

`getEventRange` no longer crashes when called.

This version of the method has drifted from `editor.findEventRange`, and now crashes when it's being used. This commit fixes an import and changes some code to match the new code in `editor.findEventRange`.


ianstormtaylor:
Thanks @justinweiss!

Issue #2621, opened 2019-02-27
Deleting Inline at beginning of editor causes error.


bnowlin:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?

1. Go to https://www.slatejs.org/#/links
2. Highlight the beginning text "In"
3. Enter any text for link
4. Move cursor to end of "In"
5. Press backspace - deletes n
6. Press backspace - deletes I
7. Press backspace - Causes error

#### What's the expected behavior?

The inline (void false) Object is deleted on step 6. 

Edit: My use case involves small pieces of text so I see this more often than an edge case. Triggering this bug in my React app causes the UI to freeze ~15.

dons20:
Is this the error you're getting as well?
![image](https://user-images.githubusercontent.com/3958349/53690825-e1adab80-3d3f-11e9-8909-97adf20b424f.png)

bnowlin:
Yes.

ShanSenanayake:
Is there any plan to merge https://github.com/ianstormtaylor/slate/pull/2850? 

We are having this issue in our platform as well. Would be nice to have it merged for next release.

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #2850, opened 2019-06-03
Fix deleting backwards when there are only inlines

North101:
This is a fix for #2621.

The bug happens when there are no text nodes between the cursor and the start of the document.

Fixing a bug

#### What's the new behavior?

Deletes an inline without erroring.

If the previous node is an Inline, the focus offset is 0 and the text value of the inline is empty, then the inline is deleted.

* [YES] The new code matches the existing patterns and styles.
* [YES] The tests pass with `yarn test`.
* [YES] The linter passes with `yarn lint`. (Fix errors with `yarn prettier`.)
* [YES] The relevant examples still work. (Run examples with `yarn watch`.)

Fixes: #2621

### DIFF
--- a/packages/slate/src/commands/at-range.js
+++ b/packages/slate/src/commands/at-range.js
@@ -377,9 +377,10 @@ Commands.deleteBackwardAtRange = (editor, range, n = 1) => {
   let node = text
   let offset = 0
   let traversed = focus.offset
+  let prevNode = document.getPreviousText(node.key)
 
-  while (n > traversed) {
-    node = document.getPreviousText(node.key)
+  while (prevNode && n > traversed) {
+    node = prevNode
     const next = traversed + node.text.length
 
     if (n <= next) {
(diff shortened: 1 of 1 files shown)


packages/slate/src/commands/at-range.js
```diff
@@ -375,23 +375,38 @@ Commands.deleteBackwardAtRange = (editor, range, n = 1) => {
...
   let traversed = focus.offset
+  const prevSibling = document.getPreviousSibling(node.key)
 
-  while (n > traversed) {
-    node = document.getPreviousText(node.key)
-    const next = traversed + node.text.length
+  // If the next sibling is an inline, has no text and the focus offset is 0,
+  // then delete the inline
+  if (
+    traversed === 0 &&
```
Tetramputechture: would you want to refactor this compound conditional to its own method? something like 'shouldDeleteInline'

ianstormtaylor:

PR #2853, opened 2019-06-04
Add Android 8 and 9 compatibility using Mutations

thesunny:

Improving a feature

#### What's the new behavior?

It's Android 8 and 9 support that works better than the existing version.

It replaces the event based Android 8 and 9 support and replaces it with `MutationObserver` based support.

Fixes: #
Reviewers: @


ianstormtaylor:
Hey @thesunny, glad to hear it's working so well for Android!

Feel free to merge whenever you are ready. It's hard to me to review because I haven't read up on mutation observers or any of the other pieces, but since it's kept separate for now it sounds good to continue going with it. There are things that seem strange to me like the 20ms timeout for `setTimeout`, but since I don't have the context I can't really turn them down. And we can just hopefully clean things up over time.

If there was a way to use mutation observers for all of Slate's updating logic I'd love that!

packages/slate-react/src/components/content.js
```diff
@@ -512,7 +529,10 @@ class Content extends React.Component {
       ...props.style,
     }
 
+    // console.log('rerender content', this.tmp.contentKey, document.text)
```
Tetramputechture: We should be able to remove this here and the blank line below it

packages/slate-react/src/plugins/android/composition-manager.md
```diff
@@ -0,0 +1,32 @@
...
+The `CompositionManager` looks at mutations using `MutationObserver` and
+bypasses all other event processing from the `dom` plugin.
+
+It uses mutations to determine how to modify the Editor `value` where `dom`
+uses events to determine how to modify the Editor `value`.
+
+## How It Works
+
+We try to avoid an Editor `render` in the middle of a composition. At the same
+time we want to make sure all mutations are eventuallyreflected in the Editor
```
Tetramputechture:
```suggestion
time we want to make sure all mutations are eventually reflected in the Editor
```

asdillon:
@thesunny Thank you very much for this work. We just ran into this issue.

scottluptowski:
This is great, thank you so much!

When do you plan to release a new tagged version of Slate with this fix?

sharpshark28:
Looking forward to this, thanks for all the hard work!

goggalore:
Works great! Thank you for the hard work. Been working off of master and this addresses many of the big problems with mobile keyboards such as cursor positioning, backspacing, entering, etc. Even fixed an issue I was having where file inputs wouldn't open the file browser.

However, is anyone else noticing a huge amount of `duplicate key` warnings being console logged in dev mode with this update? Maybe this is due to working off master, though, and not this PR specifically.

Also, it seems like the `renderMark` property passed to editor isn't executed when a mark is toggled (only on mobile). This behavior can be seen in Chrome mobile on the rich text editor example. This is more of an additional fix request though, as this PR already addresses so, so many issues.

Great work though! Excited to see it in the next official release.

dovydaskukalis:
Not sure it's only me or not but `onChange` event is never called unless I set `autoFocus` to `true` or call editor methods like `toggleMark`. Tested on Android 9.

Issue #733, opened 2017-04-19
add `data-gramm="false"` to disable Grammarly


ianstormtaylor:
Grammarly works by editing the HTML out from under the editors in weird ways that it's not expecting, which results in glitchy selections and content changes. But apparently this attribute disables it. Grammarly's team has already disabled it internally for all Draft.js editors, but we can do it on our side to avoid the wait.

imrameez:
How to stop Grammarly extension on my website using javascript

Issue #449, opened 2016-11-23
Integration with Grammarly


oyeanuj:
Hey @ianstormtaylor! Playing with Grammarly plugin alongside Slate, I am seeing issues where Grammarly extension is unable to update text in the editor. It is easily reproducible on the examples.

I haven't had the chance to investigate much further yet, but I wanted to create an issue incase anyone else here already knew what was going on..

ianstormtaylor:
I _think_ Grammarly tries to manually edit the DOM state? That's probably going to cause issues with React if so, but I haven't investigated it at all personally. But we can see if others have experience and can chime in!

oyeanuj:
Yep, seems like it. Did a little bit of digging around and found that Draft also has issues with Grammarly: https://github.com/facebook/draft-js/issues/616 cc: @kigorw (Grammarly dev)

kigorw:
Thanks, we are thinking how to fit those issues into our roadmap.

oyeanuj:
Thanks for the update @kigorw! 

In the meantime, I'll try to get to integrate @Soreine & @SamyPesse's [Rousseau](https://github.com/GitbookIO/rousseau) library with Slate as an alternative. 

@Soreine @SamyPesse, if you have already done it or plan to, do drop a note here since you both know those libraries the best!

ianstormtaylor:
Going to close since there's nothing actionable here for now for Slate, but feel free to keep discussing grammar checkers!

isubasti:
https://github.com/facebook/draft-js/issues/616, hopefully grammarly release a version that use `document.execCommand` pretty soon. then we can start improving onInput that still seems to be a problem especially for IME. on the other note, I notice that tools such as https://languagetool.org/ and https://www.scribens.com/# breaks slate as well and no obvious way on how to turn this off all together programmatically

oyeanuj:
@kigorw Any updates on using Grammarly playing well with Slate?

oyeanuj:
@kigorw @blacktaxi Any updates on using Grammarly playing well with Slate?

blacktaxi:
Hi @oyeanuj,

Slate is definitely on our radar, but is not in our immediate plans to support. Nevertheless, I've just ran a quick test of the new Grammarly, and while underlining works fine, the text modification through `document.execCommand` doesn't. It looks like the Slate editor might be misinterpreting input from `execCommand`, as the changes made by this call are reverted back once you type something with your keyboard.

[...]

PR #2855, opened 2019-06-06
Add prop to optionally enable grammarly

nadeemshaik:
Feature(i guess): This enables the user to optionally pass a prop to enable Grammarly which was previously disabled in [#733](https://github.com/ianstormtaylor/slate/issues/733).

#### What's the new behavior?
slate-grammarly

You can also check it here [sandbox](https://49zhg.codesandbox.io/). You will, of course, need to install the Grammarly extension. The above demo was tested on version 14.915.2321

Since Grammarly has moved to an API based content modification, it would be really helpful for users to be able to enable Grammarly if it suits them. So I have added a prop to enable the same. The only non-trivial part is that the extension doesn't seem to work as long as the prop `data-gramm` exists irrespective of its value.

Fixes: #733
Reviewers:

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -4,6 +4,7 @@ import Types from 'prop-types'
 import getWindow from 'get-window'
 import warning from 'tiny-warning'
 import throttle from 'lodash/throttle'
+import omit from 'lodash/omit'
 import { List } from 'immutable'
 import {
   IS_ANDROID,
@@ -541,8 +542,11 @@ class Content extends React.Component {
       [DATA_ATTRS.KEY]: document.key,
     }
 
+    const domProps = omit(this.props, Object.keys(Content.propTypes))
+
     return (
       <Container
+        {...domProps}
         key={this.tmp.contentKey}
         {...handlers}
         {...data}
@@ -559,7 +563,9 @@ class Content extends React.Component {
         // COMPAT: The Grammarly Chrome extension works by changing the DOM out
         // from under `contenteditable` elements, which leads to weird behaviors
         // so we have to disable it like this. (2017/04/24)
-        data-gramm={false}
+
+        // just the existence of the flag is disabling the extension irrespective of its value
+        data-gramm={domProps['data-gramm'] ? undefined : false}
       >
         <Node
           annotations={value.annotations}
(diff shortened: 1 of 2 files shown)


brendancarney:
Thanks for the PR @nadeemshaik. I don't think we should have a prop specific to a third party service. Maybe there's a solution that allows passing of generic data attributes to the DOM node?

nadeemshaik:
Makes sense @brendancarney. There doesn't seem to be existing support for data/dom attributes, but I can see that data attributes are being used locally [here](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/content.js#L519) Maybe extend this through props?

brendancarney:
My opinion is that any prop passed to `Editor` that is not slate specific should be passed through to the dom. So you would do:
```
<Editor
   {...editorProp}
   data-gram="false"
   data-whatever="true"
/>
```
and it would work. 

This one is probably @ianstormtaylor's call though, so let's see what he thinks.

ianstormtaylor:
Passthrough for all seems reasonable to me! I agree that it would be good to find a solution that doesn't special-case one prop over others.

nadeemshaik:
@ianstormtaylor @brendancarney how does this look, couldn't think of a better way to omit the editor/content props. Any ideas?

nadeemshaik:
Hey @ianstormtaylor, did you get a chance to take a look at this?

ianstormtaylor:
Thanks @nadeemshaik!

Jacfem:
Hi @nadeemshaik + @ianstormtaylor! Do you happen to know what slate release included this change? I searched through some recent releases but still see `data-gramm` set to `false`. I know Grammarly is [a challenge to integrate with many editors](https://github.com/ianstormtaylor/slate/issues/449), and I'm just trying to understand the state of things in regards to Slate <> Grammarly as it seems to be working on the [Slate Demo](https://www.slatejs.org/examples/richtext). Some context is that my team is on an older version of slate (0.34.7) and while investigating what a major upgrade will entail, we'd like to understand if the Grammarly integration might be fixed in later versions. Thanks for any help!

nadeemshaik:
Hey @Jacfem , I am sure it is present in version 0.44.9. It might have been published just a little earlier than that.
After the fill and till v0.47 , u can do data-gramm={false} and it should work. But after v0.5x major updates were made to the code. With those versions I guess u will need to do something like [data-gramm={undefined}](https://github.com/ianstormtaylor/slate/pull/2855/files#diff-021b2a72cf52cd5e8214d0a035b3ebc9R567) explicitly to disable it.

The idea was always to just have the ability to pass any attributes/props from the implementation level and have it override the props of the editor.

PR #2872, opened 2019-06-11
Fix bugs in getRelativeRange.

themithy:
Hi,

There are problems with annotations that span across multiple blocks or other nodes after they are recomputed with `getRelativeRange`. 

The first problem is with `setStart` and `setEnd` called subsequently, because they do a check against the "backwardness" to set the anchor or focus relatively. But the check is done after the first point was already set, which might change the order, and lead to the same point being set twice. 

The second problem is that after the annotation is split, the keys do not get set, which leads to this annotation being omited, because the anchor or focus point is treated as unset (both path and key must be set).

Thanks in advance for reviewing it.


ianstormtaylor:
Beautiful, thanks @themithy!

Issue #2826, opened 2019-05-23
Mentions example doesn't work since upgrade.


frother:
Since the breaking change that replaces decorations with annotations I cannot get the mentions example to work. I am not sure how I upgrade my own mentions plugin to use the annotations instead.

guljarpd:
@frother Same issue with me. 

@thesunny @ianstormtaylor  guys please help.

PR #2877, opened 2019-06-12
Fixed mentions example

nadeemshaik:
Fixing a bug

I have made some minor changes to get the mentions example working.

Fixes: #2826 
Reviewers: @ianstormtaylor

### DIFF
--- a/examples/mentions/index.js
+++ b/examples/mentions/index.js
@@ -41,12 +41,24 @@ import Suggestions from './Suggestions'
 const USER_MENTION_NODE_TYPE = 'userMention'
 
 /**
- * The annotation mark type that the menu will position itself against. The
+ * The annotation type that the menu will position itself against. The
  * "context" is just the current text after the @ symbol.
  * @type {String}
  */
 
-const CONTEXT_MARK_TYPE = 'mentionContext'
+const CONTEXT_ANNOTATION_TYPE = 'mentionContext'
+
+/**
+ * Get a unique key for the search highlight annotations.
+ *
+ * @return {String}
+ */
+
+let n = 0
+
+function getMentionKey() {
+  return `highlight_${n++}`
+}
 
 const schema = {
   inlines: {
@@ -119,7 +131,8 @@ class MentionsExample extends React.Component {
           onChange={this.onChange}
           ref={this.editorRef}
           renderInline={this.renderInline}
-          renderMark={this.renderMark}
+          renderBlock={this.renderBlock}
+          renderAnnotation={this.renderAnnotation}
           schema={schema}
         />
         <Suggestions
@@ -131,8 +144,8 @@ class MentionsExample extends React.Component {
     )
   }
 
-  renderMark(props, editor, next) {
-    if (props.mark.type === CONTEXT_MARK_TYPE) {
+  renderAnnotation(props, editor, next) {
+    if (props.annotation.type === CONTEXT_ANNOTATION_TYPE) {
       return (
         // Adding the className here is important so that the `Suggestions`
         // component can find an anchor.
@@ -140,6 +153,16 @@ class MentionsExample extends React.Component {
           {props.children}
         </span>
       )
+    }
+
+    return next()
+  }
+
+  renderBlock(props, editor, next) {
+    const { attributes, node } = props
+
+    if (node.type === 'paragraph') {
+      return <p {...attributes}>{props.children}</p>
     }
 
     return next()
@@ -216,11 +239,13 @@ class MentionsExample extends React.Component {
       const { selection } = change.value
 
       let annotations = change.value.annotations.filter(
-        value => value.mark.type !== CONTEXT_MARK_TYPE
+        annotation => annotation.type !== CONTEXT_ANNOTATION_TYPE
       )
 
       if (inputValue && hasValidAncestors(change.value)) {
-        annotations = annotations.push({
+        const key = getMentionKey()
+
+        annotations = annotations.set(key, {
           anchor: {
             key: selection.start.key,
             offset: selection.start.offset - inputValue.length,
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Thanks @nadeemshaik!

MatthewHerbst:
I'm finding that `this.editor.current.setAnnotations(annotations);` doesn't work for me. When I click into my editor after adding the mentions example's code, I get this:

```
_this3.editor.current.setAnnotations is not a function
```

Is there any reason why the editor might not be available on its ref?

I use `this.editor = React.createRef();` to create the ref in my `constructor`, and and pass it `ref={this.editor}` to the `Editor`. Did I miss something?

MatthewHerbst:
Upon closer look, `setAnnotations` does not appear to be a method on the editor, only `setAnnotation` is. Is that correct?

nadeemshaik:
@MatthewHerbst That was the issue (missing `setAnnotations` method). I have added the method and a few other things in this PR.

MatthewHerbst:
Ah, got it. THANK YOU @nadeemshaik !

@ianstormtaylor any chance we could please get this in a patch release before the next minor/major release?

AntonAL:
I have installed "slate": "^0.47.4" and "slate-react": "^0.22.4", but the issue is not resolved for me.
I'm still getting `editor.setAnnotations is not a function` error…
Is a patch is released?

lupos:
Following. Also need this fix.

anshudwibhashi:
Same ^^ I just updated slate and slate-react to the latest versions listed on the Github readme, and I still have the issue.

scottluptowski:
I am also experiencing the `editor.setAnnotations is not a function` issue

scottluptowski:
> editor.setAnnotations is not a function

I tried getting around this error by adding `setAnnotations` to a `commands` object in a plugin and copy+pasting the definition used in this PR.

Instead I see this error:
> Annotation.createList is not a function

Which seems a bit unusual due to the swap from `createList` to `createMap`

anshudwibhashi:
Another thing I tried is just looping through the annotations map and adding each annotation manually because `addAnnotation` is a method that I know for a fact exists, but when I do that it seems like it has no effect and the `renderAnnotation` method is never called. The `Suggestions` component seems to update just fine with the correct suggestions. It just displays them at (-1000, -1000), the default position, because it can't find the anchor.

anshudwibhashi:
Update: The `addAnnotation` workaround actually seems to (almost) do the trick. The annotations are actually being added, but the `renderAnnotation` method seems to never be called. This seems to be a known issue: https://github.com/ianstormtaylor/slate/issues/2909

lupos:
incase it's unclear to anyone im pretty sure this is fixed but they just need to release a new version. if you can't wait you can always download the source and include it manually until they update NPM.

commit in question is [here](https://github.com/ianstormtaylor/slate/commit/5471343ae84d24ee05173de48a0d5a5c47a5d7d4)

JoeCase:
@ianstormtaylor Any news on the updated package release? Tried every workaround above to no avail. Just switched from draft JS to Slate. Loving how much easier it is to navigate and build on, but the mentions is a crucial functionality I need. Actually switched specifically for those. 

If anyone has found a viable workaround would love to hear about it. Tried the looping, the source import, tried putting in the setAnnotations from the source manually but got the Annotation.createList is not a function error as well.

JoeCase:
Finally got a workaround working if anyone comes across this. I put the source folders for both slate and slate-react packages from the dowloaded repo into my project. Then in my component:
`import { Editor } from [path to local slate folder]
import { Value } from [path to local slate-react folder]`
THEN had to redirect the editor import in the local slate-react editor component to point to the local slate folder. So in local slate-react/components/editor.js... 
change
`import { Editor as Controller } from 'slate'`
to
`import { Editor as Controller } from '[path to local slate folder]'`

Then it all works fine. 

Absolutely loving slate after switching from draft-js. So much easier to navigate and build on.

lupos:
@JoeCase Thanks for this. I was trying to do this as well the other day and almost got it sorted but I am attempting to use "yarn/npm link". May be worth looking into for you.

JoeCase:
Credit to @lupos for suggesting to download the source. I never would have thought to do it. I tried all the previous methods to no avail. Hopefully the new package is live on npm soon. Hoping to break this functionality out into a plugin with key commands.

JoeCase:
Adding in case anyone still needs to hack this. It's a crucial element to my app so been patching away:

The local source folder imports worked fine in dev but crashed in production because in production `Object.keys(Editor.proptypes)`  in `local-slate-react/components/editor.js` and `Object.keys(Content.propTypes)` in `slate-react/components/editor.js` were crashing things because the propTypes were being stripped out because of [this rule.](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-foreign-prop-types.md)

Tried all day to get npm link working but couldn't so just went in and assigned those manually to a variable:

`const editorProptypes = ["autoCorrect", "autoFocus", "className", "defaultValue", "id", "onChange", "options", "placeholder", "plugins", "readOnly", "role", "schema", "spellCheck", "style", "tabIndex", "value", "onBeforeInput", "onBlur", "onClick", "onContextMenu", "onCompositionEnd", "onCompositionStart", "onCopy", "onCut", "onDragEnd", "onDragEnter", "onDragExit", "onDragLeave", "onDragOver", "onDragStart", "onDrop", "onInput", "onFocus", "onKeyDown", "onKeyUp", "onMouseDown", "onMouseUp", "onPaste", "onSelect"]
const domProps = omit(this.props, editorProptypes)`
 and
`const contentProptypes = ["autoCorrect", "className", "contentKey", "editor", "id", "readOnly", "role", "spellCheck", "style", "tabIndex", "tagName"]
const domProps = omit(this.props, contentProptypes)`

I'm on a hard deadline so doing whatever I can to get this to work and wanted to share but it might not be the most elegant. If anyone has a better solution or could get the npm link to work would love to clean it up. 

Really hoping the new version is released soon.

PR #2883, opened 2019-06-14
Improve mentions example by including scroll calculation

MatthewHerbst:

Improving an example (not technically a bug fix, but a UI/UX designer would call it a bug fix)

#### What's the new behavior?

Currently, if you type `@` in the mentions example, it initiates the creation of an annotation wrapper. That wrapper is the target of the suggestions list dropdown.

Because of the use of `React.createPortal` to keep the suggestion list outside the editor's HTML, if you scroll before typing anything after `@` then the positions of the list will be incorrect relative to the cursor. (Note: you need to make the example contain enough text such that the page scrolls to replicate.)

By including the current scroll position in the absolute positions of the list, we ensure that it is always in the correct spot relative to the cursor regardless of the scroll position of the page.

NOTE: this fix will not fix the issue of if the user type `@` and then changes the window size before typing anything further. That would require a listening HOC which would ensure the component re-renders on every window size change.

N/A

### DIFF
--- a/examples/mentions/Suggestions.js
+++ b/examples/mentions/Suggestions.js
@@ -100,8 +100,8 @@ class Suggestions extends React.PureComponent {
     const anchorRect = anchor.getBoundingClientRect()
 
     this.setState({
-      top: anchorRect.bottom,
-      left: anchorRect.left,
+      top: anchorRect.bottom + window.pageYOffset,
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @MatthewHerbst!

Issue #2870, opened 2019-06-10
Undoing adding/removing marks no longer takes the length of the mark into account


jas7457:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?
Inverse operations are not added correctly for (presumably) remove_mark

Codesandbox: https://codesandbox.io/s/eager-architecture-feupo
Recordit: http://recordit.co/AdqywQRyo3
#### What's the expected behavior?
Toggling a mark and clicking Undo would correctly put the document in the correct state.

*Note:* This seems to have been broken at the 0.46.0 release. Prior to that upgrade, this worked, but something (presumably the refactor for removing leaves) broke this behavior in that release.

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #2892, opened 2019-06-20
No longer tries removing marks that aren't there - fixes erroneous en…

jas7457:
…tries in the undo history

Bug

#### What's the new behavior?
Marks are no longer "removed" if they don't exist to begin with. This fixes an erroneous entry that gets added to the undo history which would add that mark which did not exist to begin with.

This simply checks that a node (Text) has a mark before it tries removing it. This problem was introduced by removing leaves and putting Marks on Text.

When a Mark is added, the node is split. In that logic, it moves the Focus to the beginning of the next node instead of staying on the end of the Mark/Text that was created. Then, if you go to toggle that Mark off, it tries removing the Mark both from the Text that has the Mark, and the Text directly after it. Since the code allows removing marks from nodes that don't have that mark, an entry in the history is created to add the mark to the text that never had it to begin with.

Perhaps a more comprehensive solution is later needed, but regardless, calling removeMark on Text that does not have that mark should do nothing.

#### How to reproduce in the current code
Go to the [example of rich text](https://www.slatejs.org/#/rich-text), select some text, click bold twice (once to add, one to remove), and then Ctrl+z to undo the removal of the bold.

* [ ] The linter passes with `yarn lint`. (Fix errors with `yarn prettier`.) - Note: I'm running on Windows with my git config to autocrlf (so it should pick up the current file line ending), but when running `yarn lint` I get a lot of errors ("error  Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style"). It may be helpful to add a .gitattributes file, but I will leave this up to you: https://help.github.com/en/articles/dealing-with-line-endings#per-repository-settings

Fixes: #2870
Reviewers: @ianstormtaylor

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "lint:eslint": "eslint benchmark packages/*/src packages/*/test examples/*/*.js examples/dev/*/*.js",
     "lint:prettier": "prettier --list-different '**/*.{md,json,css}'",
     "open": "open http://localhost:8080",
-    "prettier": "prettier --write '**/*.{js,jsx,md,json,css}'",
+    "prettier": "prettier --write \"**/*.{js,jsx,md,json,css}\"",
     "release": "yarn build:production && yarn test && yarn lint && lerna publish && yarn gh-pages",
     "server": "webpack-dev-server --config ./support/webpack/config.js",
     "start": "npm-run-all --parallel --print-label watch server",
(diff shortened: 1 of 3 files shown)


jas7457:
@ianstormtaylor have you had a chance to look at this? To me, it's a pretty bad bug that toggling a mark and then undoing it causes the mark to be added to text that it wasn't part of before.

ianstormtaylor:
Hey @jas7457, great catch and thank you for taking the time to help with this. I've just merged https://github.com/ianstormtaylor/slate/commit/b6a99ce7526c7c74bc82b886093167586fc37ed7 which I think might solve this issue as well? Let me know!

jas7457:
@ianstormtaylor it seems like the issue no longer occurs. Thanks! I can close out this PR

PR #2905, opened 2019-06-30
Add a mark in the search highlight example

bryanph:
I just wanted to add this to start a discussion on how we can make the decorations and marks work together in a better way. I thought it could give us some more chances to find a solution for this case.

The problem I see with marks is that right now they inherently modify the structure of the document by splitting the `text` nodes which invalidates any `range` that decorations reference. This makes it quite difficult to apply decorations consistently when marks are added/removed as you need to apply multiple of them for every text node and hence set of marks. Also, decorations don't seem to work across text nodes.

### DIFF
--- a/examples/search-highlighting/index.js
+++ b/examples/search-highlighting/index.js
@@ -117,6 +117,7 @@ class SearchHighlighting extends React.Component {
           defaultValue={initialValue}
           schema={this.schema}
           renderAnnotation={this.renderAnnotation}
+          renderMark={this.renderMark}
           spellCheck
         />
       </div>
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thanks @bryanph!

Slapbox:
@bryanph @ianstormtaylor do you have any thoughts on the best approach to fixing decorations not applying across marks? 

I can take a crack at that if anyone can give me some clue about where to begin with a solution. I figure my time would be better spent adding this back to Slate rather than creating a solution that only works for my use case.

PR #2914, opened 2019-07-08
Add point normalization to addAnnotation.

themithy:
Hi,

It looks like the annotation should be normalized when it is inserted and the anchor's and focus's keys should be populated to allow subsequent operations in `mapRanges`.


ianstormtaylor:
Thank you @themithy, great catch!

PR #2915, opened 2019-07-08
Add annotation-related ops to path.transform. (#2)

themithy:
Hi,

Annotation-related operations are not listed in path transform which leads to a crash, when we try to transform any path by such an operation. The result should be the path itself, as annotation-related op do not interfere with the tree.

Thanks in advance.


ianstormtaylor:
Good call, thank you @themithy! I really appreciate all your help.

PR #2916, opened 2019-07-08
Deduplicate dirty paths for normalization.

themithy:
Hi,

Paths accumulated for normalization are not deduplicated which leads to poor performance in case we wait with flushing for some reasons.

Thanks.


kierson:
I would like to add that without this deduplication, applying a lot of operations at once is ultra slow.

rikkit:
I think this code would read better by using lodash - already a dependency

```js
this.tmp.dirty = _.uniqBy(dirty, path => path.join(","));
```

Calyhre:
This deduplication fix so much of our performance issues too. Applying a fair amount of operations is non-linear on time spent normalizing dirty paths.

Calyhre:
Combine it with #2959 for even faster operations

ianstormtaylor:
Thank you @themithy! This is great.

PR #2918, opened 2019-07-09
Fix rendering of collapsed annotation at offset zero.

themithy:
Hi,

`renderAnnotation` is not called for *collapsed* annotation at offset *zero*. This patch fixes it.

Thanks.


ianstormtaylor:
Hey @themithy I think this one is failing Prettier, can you double check it? Happy to merge once it's green. Thanks!

themithy:
@ianstormtaylor it's green now. Thanks.

ianstormtaylor:
Thanks @themithy!

PR #2925, opened 2019-07-12
fix: Unknown event handler property `onEvent`

linonetwo:
#### What's the new behavior?

```
index.js:1 Warning: Unknown event handler property `onEvent`. It will be ignored.
	in div (created by Content)
    in Content (created by Editor)
```

Will gone.

Simply add a line of proptype.

Fixes: I'm using a local copy of the latest master and found this bug.
Reviewers:


ianstormtaylor:
Thanks @linonetwo!

PR #2928, opened 2019-07-15
Fix getRelative range on text level.

themithy:
Hi,

It looks like there is an inconstistency in `getRelativeRange` on text level. The path on this level should be empty in my opinion, as we cannot go to a lower level (we are already on the lowest node - namely text node).

I do not know whether this causes any bugs, but it looks like inconsistency in the function.

Cheers.


ianstormtaylor:

PR #2929, opened 2019-07-15
Fix typo in marks

johnisaac:
This PR just changes a type from `an` to `can` in the marks section of the documentation


brendancarney:
Thanks @johnisaac!

PR #2934, opened 2019-07-18
Fix renderBlock() in paste-html example

nareshbhatia:
It appears that render block is missing a case for rendering `paragraph`. Without this paragraphs are being rendered as `<div>`s.

It is fixing a bug in the example

#### What's the new behavior?

paragraphs are now rendered as `<p>` tags

No

Fixes: #
Reviewers: @


ianstormtaylor:
Thanks @nareshbhatia!

PR #2938, opened 2019-07-19
Fix splitNode range update when selection is backward

Calyhre:

This fixes a bug

#### What's the new behavior?

| Before | After |
|---|---|
|  |  |

When splitting a node, [every ranges inside the value are re-normalized](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/models/value.js#L799-L816). This includes selection and annotations for example. But this is done using `range.start` and `range.end`. 

- Due to the order in which those are done, a very special case could move start **after** the end, before the end is normalized
- This is then passed to [`range.setEnd`](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/interfaces/range.js#L530-L533) which check if the [`range.isBackward`](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/interfaces/range.js#L50-L52) which returns nulls, because points are [`isUnset`](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/models/point.js#L110) (path is null at this point).
- This result of `point.start` to be set instead of `point.end`

To avoid aving to check `isBackward` and having the wring order, a very simple fix is to normalize `anchor` then `focus` as the order does not matter here.

Fixes: 🤷‍♂️
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate/src/models/value.js
+++ b/packages/slate/src/models/value.js
@@ -798,16 +798,16 @@ class Value extends Record(DEFAULTS) {
 
     value = value.mapRanges(range => {
       const next = newDocument.getNextText(node.key)
-      const { start, end } = range
+      const { anchor, focus } = range
 
-      // If the start was after the split, move it to the next node.
-      if (node.key === start.key && position <= start.offset) {
-        range = range.moveStartTo(next.key, start.offset - position)
+      // If the anchor was after the split, move it to the next node.
+      if (node.key === anchor.key && position <= anchor.offset) {
+        range = range.moveAnchorTo(next.key, anchor.offset - position)
       }
 
-      // If the end was after the split, move it to the next node.
-      if (node.key === end.key && position <= end.offset) {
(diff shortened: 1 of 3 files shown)


mgdawber:
Nice one!

packages/slate/src/models/value.js
```diff
@@ -798,16 +798,16 @@ class Value extends Record(DEFAULTS) {
 
     value = value.mapRanges(range => {
       const next = newDocument.getNextText(node.key)
-      const { start, end } = range
+      const { anchor, focus } = range
 
       // If the start was after the split, move it to the next node.
```
akumatus: I think if the annotation also changed from 'start' to 'anchor' will be better to understand.

akumatus:
Good job! I have found the same problem and fix it exactly like you are.

ianstormtaylor:
Thank you @Calyhre!

PR #2941, opened 2019-07-23
Fix/tables example

NihilusZero:
When the last cell of the table is empty AND the cursor is positioned at the start of the text following the table, hitting backspace will remove the cell and replace it with the <div> tag the text is wrapped in. The result will be a table with a visibly missing last cell.

This fixes a bug.

#### What's the new behavior?
Old Behaviour
old
New behaviour
new

After Slate replaces the `table-cell` with the `<div>` that wraps the next paragraph, the added code converts the `<div>` to a `table-cell`.

I don't know if this issue has been reported before. I found it myself.

### DIFF
--- a/examples/tables/index.js
+++ b/examples/tables/index.js
@@ -173,7 +173,10 @@ class Tables extends React.Component {
       if (prevBlock.type === 'table-cell') {
         if (['Backspace', 'Delete', 'Enter'].includes(event.key)) {
           event.preventDefault()
-          if(value.startBlock.type != 'table-cell' && previous.text.length == 0)
+          if (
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @NihilusZero, thanks for taking the time to look at this. I think this is something that should be solved with schema normalizations, instead of trying to normalize in event handlers.

ianstormtaylor:
Hey @NihilusZero , thank you so much for taking the time to work on this. I'm still interested in this behavior in `0.50`. But this PR is now out of date since so much has changed. I'm going to close it out, but if you are down to resubmit it that would be amazing! Thank you for understanding.

Issue #2926, opened 2019-07-14
marks removed cannot be undone


cvlmtg:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

if you remove some marks and then press undo, they are reapplied in the wrong way. e.g. go to https://www.slatejs.org/#/rich-text, select some text, add bold, remove bold, then press ctrl-z to undo. the marks are reapplied on the wrong characters

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/6401008/61184914-3862b480-a653-11e9-99a8-fa841e1f533b.gif)

info:

* macOS 10.14.5
* Chrome 75
* slate 0.47.4
* slate-react 0.22.4

#### What's the expected behavior?

slate should correctly undo the operation. looks like slate is trying to create the reverse of the last operation applied instead of simply restoring the previous immutable state

justinweiss:
I think this is happening because when the ["split node before", "split node after", "remove mark"] is being undone, it happens as ["add mark", "merge node after", "merge node before"].

Usually, if you merge a node into another node, you'd want it to take on the same marks as the node you're merging it into. Like if you're merging a paragraph into the previous paragraph, and the previous paragraph is bold, you want the other paragraph to be bold, too. During this kind of undo, that doesn't make sense.

I wonder how we could detect those two situations and do something differently?

PR #2946, opened 2019-07-26
Fix undo a removeMark re-add mark to wrong nodes

Calyhre:

This fixes a bug

#### What's the new behavior?

| Before | After |
|---|---|
| Screen Recording 2019-07-26 at 15 46 06 2019-07-26 15_49_15 | Screen Recording 2019-07-26 at 15 46 41 2019-07-26 15_50_45 |

When adding a mark or selecting text around one, and then removing the mark, it results in an unwanted `remove_mark` operation on the next following text node. Undoing it would then result in an improperly reapplied mark.

A very naive fix is to check if the marks we are trying to remove are in the node beforehand.

<details>
<summary>👉 Click for a more detailed step by step 👈</summary>

```xml
<!-- Initial step -->
<paragraph>
  One <anchor />two<focus /> three
</paragraph>

<!-- After addMark -->
<paragraph>
  One <b><anchor />two</b><focus /> three
</paragraph>
<!-- 👆 The focus position here is very important -->
```

At this moment, the paragraph has three nodes:
- "One "
- "two", with the anchor and the bold mark
- " three", with the focus

Removing the mark at this very moment with result in a bunch of operations:
1. `remove_mark` on second node
2. `merge_node` on second node
3. `remove_mark` on third node <-- 🚨 That's where the bug is 
4. `merge_node` on third node

Undoing this will then generate this:
1. `split_node` around "two"
2. `add_mark` on second node
4. `add_mark` on third node <-- 🚨 There was no mark here, but the operation is correctly inverted

</details>

Fixes: #2926
Reviewers: @ianstormtaylor @justinweiss

### DIFF
--- a/packages/slate/src/commands/by-path.js
+++ b/packages/slate/src/commands/by-path.js
@@ -237,6 +237,10 @@ Commands.removeMarksByPath = (editor, path, offset, length, marks) => {
   const { document } = value
   const node = document.assertNode(path)
 
+  if (marks.intersect(node.marks).isEmpty()) {
+    return
+  }
+
   editor.withoutNormalizing(() => {
     // If it ends before the end of the node, we'll need to split to create a new
     // text with different marks.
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thank you @Calyhre!

Issue #2891, opened 2019-06-20
Redos are lost when focus is changed


evasteingrims:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

A. Redos are lost when trying to undo past the history stack using keyboard shortcut
B. Redos are lost when focus is changed

This is problem in the history example: https://www.slatejs.org/#/history

**To replicate (A):**
1. Type some text
2. Undo past the history stack using keyboard shortcut
3. Click on editor to give it focus and try to redo using keyboard shortcut. It won't work

|Redos are lost when trying to undo past the history stack using keyboard shortcut |
| -- |
| ![slate-issue](https://user-images.githubusercontent.com/11168195/59844738-21c68780-934b-11e9-909d-251cd0792ce7.gif)|
**NOTE**:  This works if you undo/redo using the buttons in the toolbar

**To replicate (B):**
1. Type some text
2. Change focus in the editor by clicking somewhere else
3. Try to redo using keyboard shortcut. It won't work

|Redos are lost when focus is changed |
| -- |
| ![slate-issue2](https://user-images.githubusercontent.com/11168195/59844554-b7ade280-934a-11e9-94ac-7aa9ac594cce.gif)|

Chrome Version 74.0.3729.169
Safari Version 12.0.1 (14606.2.104.1.1)
MacOs 110.14.1 (18B75)

#### What's the expected behavior?

Undo/Redo history should be retained.

riosn-n:
Similar problem occurs if -in the [rich text editor](https://www.slatejs.org/#/rich-text) example- you select a piece of text, mark it as bold, then hit Ctrl+Z twice, Ctrl+Y doesn't work anymore.

bryanph:
It is because setting the selection is seen as an "undoable action" (an operation). We might want to consider separating setting the selection from the operation mechanism and instead infer it from the operations that occur. This could also clear the way to handling multiple selections (such as with OT) in the future?

Slapbox:
If anybody wants an (undocumented) starting point, I'm pretty sure my PR resolves that, but it's outdated and incomplete: https://github.com/ianstormtaylor/slate/pull/2347

julienR2:
thanks @bryanph and @Slapbox for giving some feedbacks and hint to dig the issue.
One thing I still don't get, is why would we wipe the redos ? It doesn't seems the common behaviour on any other editing tools/input/text area

[...]

Issue #2729, opened 2019-05-01
Applying changes without editor focus clobbers undo history 


quicksnap:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Replicated on https://www.slatejs.org/#/rich-text with latest Chrome browser.

When applying changes to a selection when the editor doesn't have focus, the undo history is clobbered.

To replicate: 
1. Select some text.
2. Click outside the editor
3. Click on "Bold" toolbar mark
4. Click on editor and try to undo. It won't work.

![2019-05-01 16 37 00](https://user-images.githubusercontent.com/1199742/57050042-7b55e400-6c2f-11e9-9ee0-9f4174b2c9b7.gif)

Further, if you make changes and then try to undo beyond this invalid undo point, the user may no longer redo. This may be a separate issue.

#### What's the expected behavior?

Undo history should be retained.

quicksnap:
In my local environment I was unsure of how to debug this. Through trial and error I figured out this was the root issue. 

Any tips on inspecting why operations fail to undo/redo? I tried logging `slate:operation*`, but there is a lot of noise, and console-logged Immutable objects are tough to inspect =(

Slapbox:
@quicksnap this PR fixes _some_ of the issues with history, but probably introduces more issues. It's also outdated now, slightly. You'd need to integrate the new changes to the `on-history.js` file (very minimal.)

If you want to work on improving this, I'd be happy to help. The trouble is that I got it nearly working just in time to step away for a few days, and when I came back, nothing made sense to me anymore. So it moves us closer, but good luck figuring out how =/

quicksnap:
@Slapbox which PR?

Slapbox:
Ah crap, sorry about that. Here you go: https://github.com/ianstormtaylor/slate/pull/2347

Slapbox:
By the way, there's a related issue here which might be of interest to you, though it has no solutions itself: https://github.com/ianstormtaylor/slate/issues/2098

quicksnap:
Once the problem becomes more of a priority for me, I would be interested in helping. For now, I believe one workaround would be to prevent toolbars from modifying the document without a focused selection. So, one could destroy the selection `editor.onBlur` and also cause a `editor.focus` event before any toolbar actions occur.

This is a problem, though, and I would like to know how better to debug what's going wrong. There's a lot of events happening at any given moment, so it's difficult to diagnose what's going wrong with the history.

[...]

PR #2948, opened 2019-07-26
Fix several undo/redo issues related to selection operation

sgurenkov:
_bug_

Should fix the following issues: #2891, #2729

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -11,6 +11,7 @@ import {
   IS_FIREFOX,
   HAS_INPUT_EVENTS_LEVEL_2,
 } from 'slate-dev-environment'
+import Hotkeys from 'slate-hotkeys'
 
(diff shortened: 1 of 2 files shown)


laufeyrut:
@ianstormtaylor Do you think you could review this PR 🙏

ianstormtaylor:
Thanks @sgurenkov!

bryanph:
I'm wondering if we can get by without storing set_selection operations? In what cases would you like to "undo" a selection? Or in what cases does a selection influence the behavior of an undo? Can't we just make sure that all operations are undone whilst setting the selection to some sane place?

ianstormtaylor:
@bryanph I don't think Slate currently does it properly, but ideally the selection would be undone to where it was at the time of the operation. I think Google Docs does this properly.

evasteingrims:
This does not totally fix #2891. Redos are still lost on selection change

see: 
not-fixed

PR #2964, opened 2019-08-09
Update glossary.md with definition of `match`

devinmorgan:
Documentation improvement

#### What's the new behavior?
`match` get's used a number of times within the slate/schema.md but it does not appear to be explicitly defined anywhere. This term seems like a great candidate for the Glossary.

The change is was trivial and only to the documentation

* [N/A ] The new code matches the existing patterns and styles.
* [N/A ] The tests pass with `yarn test`.
* [ N/A] The linter passes with `yarn lint`. (Fix errors with `yarn prettier`.)
* [ N/A] The relevant examples still work. (Run examples with `yarn watch`.)

No

Fixes: #
Reviewers: @


ianstormtaylor:
Beautiful @devinmorgan, thank you!

Issue #2965, opened 2019-08-11
deleteBackwardAtRange fails when number of characters to delete equals range offset


kaydelaney:
#### What's the current behavior?
As a result of what I assume is an off-by-one error, using `editor.deleteBackwardAtRange` when `n` is equal to the range's offset doesn't result in anything being deleted.

#### What's the expected behavior?
Supposing I have an editor populated with the value `He|llo World`, (where | indicates the caret) and I invoked `editor.deleteBackwardAtRange(editor.value.selection, 2)`, I would expect `He` to be deleted.

PR #2966, opened 2019-08-11
deleteBackwardAtRange now works when offset is equal to n

kaydelaney:
#### What's the new behavior?
If `editor.deleteBackwardAtRange` is invoked and the range's offset is equal to `n`, `n` characters will be deleted.

Changes the check that `n < focus.offset` to `n <= focus.offset`.

Fixes: #2965


ianstormtaylor:
Hey @kaydelaney can you add a test for this to make sure we don't regress in the future? Thanks!

kaydelaney:
@ianstormtaylor Sure, no problem

ianstormtaylor:
Thanks @kaydelaney!

PR #2970, opened 2019-08-20
Fix extracting relative ranges from decorations.

themithy:
Hi,

It looks like all decorations, those passed from `decorateNode` for the current node, and those received from the parent, should go through relative range mapping.

The bug is evident when decorating text with existing marks as then decorations do not comply well.

Thanks.


ianstormtaylor:
Ah, great catch, thanks @themithy!

PR #2971, opened 2019-08-20
Fix range normalization for node removal and split.

themithy:
Hi,

In 'splitNode' and 'removeNode' range normalization start/end should be changed to anchor/focus, as those might change during normalization and the order might be reversed - leading to a point being set twice and the other not being set at all.

Thanks.


ianstormtaylor:
Thanks @themithy!

Issue #2969, opened 2019-08-19
React does not recognize the 'renderMark' prop on a DOM element


larrywu13535:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?
slate: v0.47.6
slate-react: v0.22.6

[https://imgur.com/a/POO5Rnc](url)

#### What's the expected behavior?
Error did not occur in slate: v0.47.4 and slate-react: v0.22.4

scientihark:
same here and renderInline are also buged

wroughtec:
Also complains about renderBlock

eshavrov:
Need to make changes to the file `/packages/slate-react/src/components/editor.js`

```
  static propTypes = {
    autoCorrect: Types.bool,
    autoFocus: Types.bool,
    className: Types.string,
    decorateNode: Types.func,
    defaultValue: SlateTypes.value,
    id: Types.string,
    onChange: Types.func,
    options: Types.object,
    placeholder: Types.any,
    plugins: Types.array,
    readOnly: Types.bool,
    renderAnnotation: Types.func,
    renderBlock: Types.func,
    renderDecoration: Types.func,
    renderDocument: Types.func,
    renderEditor: Types.func,
    renderInline: Types.func,
    renderMark: Types.func,
    role: Types.string,
    schema: Types.object,
    spellCheck: Types.bool,
    style: Types.object,
    tabIndex: Types.number,
    value: SlateTypes.value,
    ...EVENT_HANDLERS.reduce((obj, handler) => {
      obj[handler] = Types.func
      return obj
    }, {}),
  }
```

jonathannen:
I believe this is related to #2951 as well.

Along with this bug, Firefox (Desktop) editing is buggy. After entering the second letter the cursor will jump back to the start of the block and/or the last mark.

I get this behavior right now at https://www.slatejs.org/#/rich-text when using Firefox.

mariuslundgard:
I still get this warning. Perhaps the fix was not published to npm?

davea38:
Plus one for this not working - I am getting the exact same error (v.0.58.4)

paramsinghvc:
I'm still getting this annoying warning all the time for setting requiredMark to false. Also, in production it still shows the mark!!


![Screenshot 2020-12-11 at 15 03 53](https://user-images.githubusercontent.com/4329912/101919160-27a8cc00-3bc2-11eb-8788-4bae0dcde704.png)

PR #2974, opened 2019-08-21
fix React warnings on renderMark, renderBlock, renderDecoration

yue-l:

Bug fix

#### What's the new behavior?

Error console log stops showing up.

Screen Shot 2019-08-21 at 12 20 17 PM

I 

Fixes: #2969
Reviewers: @


packages/slate-react/src/components/editor.js
```diff
@@ -44,6 +44,13 @@ class Editor extends React.Component {
     placeholder: Types.any,
     plugins: Types.array,
     readOnly: Types.bool,
+    renderAnnotation: Types.func,
+    renderBlock: Types.func,
+    renderDecoration: Types.func,
+    renderDocument: Types.func,
+    renderEditor: Types.func,
+    renderInline: Types.func,
+    renderMark: Types.func,
```
eshavrov: Need to add another property `decorateNode: Types.func,`.
And I would advise moving the list to constants `/packages/slate-react/src/constants/...`

mgdawber:
Nice :fire:

packages/slate-react/src/components/editor.js
yue-l: I put the fix as part of constants/event-handlers.js since they all share "Types.func". Or we create a different file for it?

ianstormtaylor:
Hey @yue-l can you move this out of the `event-handlers` constants, since these other ones technically aren't event handlers. I'd be okay with another similar file called `other-handlers` for now, or even just inlining the constant array in the component file itself. Thanks!

ianstormtaylor:
Thanks @yue-l!

PR #2976, opened 2019-08-21
Fix improper selection after insertFragment

Calyhre:

Fixing a bug

#### What's the new behavior?

| Before | After |
| --- | --- |
| before | after |

When inserting a fragment, the new selection is calculated by looking at new texts in the document. This is done by comparing previous keys to new ones. On certain cases, specially when the text node where the selection is, the current `startText` node will be split, creating a new key for the remaining part of the node. This node will later be considered as new text, and a candidate to put the updated selection to.

This is not reliable when the fragment ends with un-mergeable text nodes like inlines or marks (like the before GIF above).

This PR completely rewrite the re-placement of the selection inside the `insertFragment` method. We do not rely anymore on potential new keys, but put the selection relative to fragment length, and `selection.start` before the insertion. 

Fixes: #
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate/src/commands/at-range.js
+++ b/packages/slate/src/commands/at-range.js
@@ -735,7 +735,7 @@ Commands.insertFragmentAtRange = (editor, range, fragment) => {
 
     // Regenerate the keys for all of the fragments nodes, so that they're
     // guaranteed not to collide with the existing keys in the document. Otherwise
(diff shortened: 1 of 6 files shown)


eshavrov:
Great job!

ianstormtaylor:
Wow, absolutely amazing! Thank you @Calyhre!

PR #2982, opened 2019-08-27
FIX findPoint offset may be include zero width string

ibone:
This fixes a bug.

findPoint return offset is wrong
findPoint created a new range, but forget it may be include some zero width string, so i removed
sorry, my English is poor

 I found it myself.


ianstormtaylor:
Thanks @ibone!

Issue #2978, opened 2019-08-23
Firefox regressions


mehdikazi:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
This was initially reported on this [PR](https://github.com/ianstormtaylor/slate/pull/2236#issuecomment-523004658), but thought it might be helpful to make a new issue for it. 

You can see the following behavior on all the slate examples using Firefox:
1. Pressing arrow left or right does not move the selection.
2. Typing a character makes the selection jump somewhere else in the editor.

Video from PR - https://drive.google.com/file/d/1YcPlKy8PCiHnWIrpKx-3ttxmdGdGw2X-/view

jonathannen:
Confirm this is occurring for me too (GIF below). My workaround at the moment is to use slate-react `0.22.4` which is using slate `0.47.4`.

![Lc6UyRAuP5](https://user-images.githubusercontent.com/49863/63630991-7399cf80-c664-11e9-9fef-b1d7f93dd928.gif)

macrozone:
can also confirm this happens on @react-page: https://github.com/react-page/react-page/issues/720

devxom:
I can also confirm encountered a similar problem. 

System info: MacOS 10.14.6 Firefox 69.0b14

1. The problem appeared after updating:
- **slate-react** from `0.22.7` to `0.22.8` 
- **slate** from `0.47.7` to `0.47.8`

2. The symbolic deletion also broke when selecting a word. After the delete operation, a jump to the beginning of the section occurs. Also, a carriage change occurs when entering characters.

![2019-08-30 12 50 31](https://user-images.githubusercontent.com/2714216/64011773-0cc55c00-cb25-11e9-9822-34e2f36dddec.gif)

![2019-08-30 12 58 03](https://user-images.githubusercontent.com/2714216/64012308-17342580-cb26-11e9-9c6c-ff7a0330533c.gif)

kaydelaney:
Made a PR reverting the changes that introduced the regressions, and (hopefully) also fixing the original issue without causing other problems. Would love some reviews if anyone has the time!
https://github.com/ianstormtaylor/slate/pull/2995
@ianstormtaylor

projectLewis:
Can confirm what jonathannen said in that 0.22.4 which is using slate 0.47.4 are the last working versions in Firefox for me. slate-react - 0.22.7 and slate -0.47.7 as someone else mentioned did not work for me.

[...]

PR #2995, opened 2019-09-03
Bugfix: Fixes issue where focusing on editor always placed caret at offset 0

kaydelaney:
Fixing a bug.

#### What's the new behavior?
Focusing on editor places caret next to cursor as expected.

Previously, in Firefox, focusing on an editor would result in `updateSelection` being called before
`onNativeSelectionChange` is triggered, but after the native selection has been changed, resulting in `removeAllRanges` being called on the native selection. To prevent this, this PR introduces a member `nativeSelection` which stores the native selection's state when `onNativeSelectionChange` is triggered, so that when `updateSelection` is called, we can check if the native selection has been updated before `onNativeSelectionChange` has been triggered, and if so, prevent `removeAllRanges` from being called.

Fixes: #2978

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -288,39 +288,43 @@ class Content extends React.Component {
       // Otherwise, set the `isUpdatingSelection` flag and update the selection.
       updated = true
       this.tmp.isUpdatingSelection = true
+      removeAllRanges(native)
 
-      if (!IS_FIREFOX) {
-        removeAllRanges(native)
-
-        // COMPAT: IE 11 does not support `setBaseAndExtent`. (2018/11/07)
-        if (native.setBaseAndExtent) {
-          // COMPAT: Since the DOM range has no concept of backwards/forwards
-          // we need to check and do the right thing here.
-          if (isBackward) {
-            native.setBaseAndExtent(
-              range.endContainer,
-              range.endOffset,
-              range.startContainer,
-              range.startOffset
-            )
-          } else {
-            native.setBaseAndExtent(
-              range.startContainer,
-              range.startOffset,
-              range.endContainer,
-              range.endOffset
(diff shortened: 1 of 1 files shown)


joechesney:
lgtm

packages/slate-react/src/components/content.js
```diff
@@ -78,6 +98,8 @@ class Content extends React.Component {
     tagName: 'div',
   }
 
+  nativeSelection = {}
```
ianstormtaylor: Can you add a comment here explaining what this is for?

packages/slate-react/src/components/content.js
```diff
@@ -237,7 +259,12 @@ class Content extends React.Component {
 
     // If the Slate selection is unset, but the DOM selection has a range
     // selected in the editor, we need to remove the range.
-    if (selection.isUnset && rangeCount && this.isInEditor(anchorNode)) {
+    if (
+      selection.isUnset &&
+      rangeCount &&
+      this.isInEditor(anchorNode) &&
+      selectionsEqual(this.nativeSelection, native)
```
ianstormtaylor: Since this is only used once, can you inline this logic?

ianstormtaylor: Also, can you add a comment to explain what is going on there by comparing the current native selection to a cached value?

packages/slate-react/src/components/content.js
ianstormtaylor: Also, it seems like this might be better placed in `this.tmp.selection`?

Ridd0:
@ianstormtaylor Is there any chance to merge it in this week? Thanks!

ianstormtaylor:
Ah sorry missed the review request. Looks good to me, thanks so much @kaydelaney!

PR #2996, opened 2019-09-04
Update range.md

BrentFarese:

This is a documentation fix.  It appears that references to point.normalize(document) should be range.normalize(document).  See changes.

No.


ianstormtaylor:
Thanks @BrentFarese!

Issue #3000, opened 2019-09-05
Dragging text that spans multiple nodes into the same node as the end node will place the text in the wrong position


justinweiss:
#### Do you want to request a _feature_ or report a _bug_?

Report a bug.

#### What's the current behavior?

![Screen Recording](https://user-images.githubusercontent.com/1020/64387482-f6ba0e80-cff0-11e9-871d-7b2c7f113508.gif)

#### What's the expected behavior?

The text should be placed where the cursor is dropped.

This is happening because we try to recalculate the drop offset based on the number of removed characters. But when the drag contains multiple nodes, we can't calculate it correctly.

PR #3001, opened 2019-09-06
Drop dragged nodes into the correct place

justinweiss:

Fixing a bug.

#### What's the new behavior?

Screen Recording

When drag-and-dropping nodes within the same editor, the removal of
the dragged nodes can cause paths and offsets to change. This is
really hard to compensate for.

Instead of compensating for it, rearranging the order things happen
mean we are always working with a document in a reasonable state.

Now, we:

1. Fire a MouseUp event on `event.target` (because we haven't done
   anything, it's guaranteed to exist)
2. Save the range we're dragging (for later)
3. Select the target range (because we haven't deleted anything, this
   is still valid)
4. Delete the range we're dragging (this will automatically adjust the
   editor's selection)
5. Insert the dragged fragment at the current range (the adjusted target range)

No matter where we're dragging from or to, these should all be
pointing at places that both exist, and haven't changed.

Fixes: #3000 
Reviewers: @

### DIFF
--- a/packages/slate-react/src/plugins/dom/after.js
+++ b/packages/slate-react/src/plugins/dom/after.js
@@ -299,7 +299,7 @@ function AfterPlugin(options = {}) {
     const { value } = editor
     const { document, selection } = value
     const window = getWindow(event.target)
-    let target = editor.findEventRange(event)
+    const target = editor.findEventRange(event)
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @justinweiss!

PR #3005, opened 2019-09-06
Tighten mark addition logic against zero-length marks.

themithy:
Hi,

When adding marks to multi-line range it might happen that a mark is added to zero-length text node. Although such situation is corrected then by normalization, extra operations are generated unnecessarily.

This fix tightens the logic.

Thanks.

### DIFF
--- a/packages/slate/src/commands/at-range.js
+++ b/packages/slate/src/commands/at-range.js
@@ -60,10 +60,12 @@ Commands.addMarkAtRange = (editor, range, mark) => {
       let index = 0
       let length = node.text.length
 
-      if (key === start.key) index = start.offset
       if (key === end.key) length = end.offset
-      if (key === start.key && key === end.key)
-        length = end.offset - start.offset
+
+      if (key === start.key) {
+        index = start.offset
+        length -= start.offset
+      }
 
       editor.addMarkByKey(key, index, length, mark)
     })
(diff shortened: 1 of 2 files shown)


packages/slate/src/commands/at-range.js
```diff
@@ -60,10 +60,12 @@ Commands.addMarkAtRange = (editor, range, mark) => {
       let index = 0
       let length = node.text.length
 
-      if (key === start.key) index = start.offset
       if (key === end.key) length = end.offset
-      if (key === start.key && key === end.key)
-        length = end.offset - start.offset
+
```
CameronAckermanSEL: is this change significant for the fix or is this just trying to tighten up logic?

CameronAckermanSEL: This logic is slightly different now specifically in the case where `(key === start.key && key !== end.key)`

Before:
`length = node.text.length`
Now:
`length = node.text.length - start.offset`

themithy: Sure this is needed, if you move the offset you need to truncate the length accordingly - in both cases when key === and  !== end.key

ianstormtaylor:
Hey @themithy this looks good to me. Can you add a test to ensure this logic? Might be hard with normalization, not sure.

ianstormtaylor:

Issue #2504, opened 2018-12-20
Focus is stolen from input and textarea elements in void nodes


rfeltis:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?
Focus from ``input`` and ``textarea`` controls have their focus stolen when clicked, even though they are rendered as void nodes. If the controls are tabbed into, the focus remains.

May be a regression of this issue: https://github.com/ianstormtaylor/slate/issues/1744

JSFiddle: https://jsfiddle.net/rfeltis/fj9dvhom/2742/

![image](http://g.recordit.co/9qq7fFzVHv.gif)

OS: macOS Mojave 10.41.1
Browser: Chrome 71.0.3578.98 and Firefox 64.0

#### What's the expected behavior?

The focus stays inside of the ``input`` and ``textarea`` fields.

zhujinxuan:
I believe the problem is in the onMouseDown branch here:
https://github.com/ianstormtaylor/slate/blob/13107b0c437fe0e84231c87f6e7887f6defc08c8/packages/slate-react/src/plugins/after.js#L390-L398

nicoxx:
Hello, I have the same issue, @zhujinxuan @rfeltis did you found a workaround ?

nicoxx:
By capturing `onBlur` event on my `<input>`, I see that the code responsible for un-focusing is the one below.
Basically, there is a inconsistency between `activeElement` (which is the `input`) that is not the editor itself, and the editor is considered to be focused (`selection.isFocused=true`)

https://github.com/ianstormtaylor/slate/blob/13107b0c437fe0e84231c87f6e7887f6defc08c8/packages/slate-react/src/components/content.js#L175-L180

rfeltis:
> Hello, I have the same issue, @zhujinxuan @rfeltis did you found a workaround ?

@nicoxx As a workaround, we added ``preventDefault`` to the click handler. Fixes it for the time being.

jmoyers:
As an aside, I spent a few hours tracking down similar behavior in my app today. The short version is, if you have ANY `onBlur` handler attached to the Editor, `preventDefault/stopPropagation` workaround stops working and you can no longer focus inputs within a void node.

[...]

PR #3014, opened 2019-09-11
fix cursor jumping to end in embeds example

Entkenntnis:

Fixing a bug

#### What's the new behavior?

Click on the video embed and change the url _in the middle_. The cursor should stay where you expect it. (instead of jumping to the end of the input)

Controlled inputs can't be backed by the editor state, but need a local state instead, so the solution is to wrap the input in a functional component with state and pass the onChange-handler on.

Fixes: #2504
Reviewers: @


ianstormtaylor:
Thanks @Entkenntnis!

PR #3015, opened 2019-09-12
Move selection to previous node when it's at the start of a non-void inline node

akumatus:
feature

#### What's the new behavior?
屏幕快照 2019-09-12 上午11 21 41

When user move selection backward to the start of a non-void inline node and start to input words. At current we got a result like this:

屏幕快照 2019-09-12 上午11 21 11

The character became a part of the link node, and it seems confused to the user. 

Moreover, the editor will show a different result if the user moves selection forward to the start of the link. It is ambiguous because the two kinds of selection are visually equal. 

The new behavior:

屏幕快照 2019-09-12 上午11 21 27

Add COMPAT logic in `findSelection` function.

Fixes: #
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/plugins/react/queries.js
+++ b/packages/slate-react/src/plugins/react/queries.js
@@ -526,6 +526,35 @@ function QueriesPlugin() {
       }
     }
 
+    // COMPAT: If the selection is at the start of a non-void inline node, and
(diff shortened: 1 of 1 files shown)


ianstormtaylor:

PR #3016, opened 2019-09-12
Update getTexts() documentation for Node Model

mmdonaldson:

Fixing a bug in documentation - (Small fix - caught me out)

#### What's the new behavior?

Removed incorrect function parameter for `getTexts()`

Updated documentation

Fixes: #
Reviewers: @


ianstormtaylor:
Good catch, thanks @mmdonaldson!

Issue #2967, opened 2019-08-12
Programmatic changes in the document causes unwanted scrolling


gracicot:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

When modifying the document programmatically, either with commands, operations or a set_value, the scrolling position of the document is centered on the caret.

[Go to this code sandbox](https://codesandbox.io/s/slatejs-playground-xmokw). This is the [Syncing Operations](https://www.slatejs.org/#/syncing-operations) example modified to expose the problem. The bottom editor is entering text each seconds, so the top editor is syncing with it (through operations).

To reproduce, simply scroll down in the top editor. The editor will keep jumping back up.

This is caused by the selection changes when the text is inserted. For the cursor to stay at the same position, it has to account the offset of the changed text. Slate catches a selection change and reset the scroll position, as if the selection truely changed place. But in reality, the selection hasn't truely moved from the user perspective.

The implication of this bug is that scrolling in a document with collaborative editing is impossible, since the scrolling will keep getting reset to the cursor.

Also, this bug don't always show up. It will only happen when the position of the selection is changed. If there's a mark between the programmatic text insertion and the cursor, no scrolling will happen since the selection don't need to be reajusted. [Here's a demo where scrolling reset don't happen with a mark in the way](https://codesandbox.io/s/slatejs-playground-li9d6).

I found that bug when programming an app were styling information was streamed from a server. As the styling is streamed, it is applied and the scrolling position keep getting reset to the cursor position. Since applying styles in the text is almost always a path change, the scrolling keep getting reset, making the editor unusable while applying styles.

#### What's the expected behavior?

The expected behaviour is to scroll only on user operations. Programmatic modifications to the document should not implicitly change the scrolling position.

PR #3019, opened 2019-09-13
Prevent scrolling when no user actions are performed

gracicot:
Fixing a bug

#### What's the new behavior?
By default, the editor will not scroll on changes on the selection. Instead it will only scroll to make the cursor caret visible when a user action is performed

Some events on the `BeforePlugin` has been marked as *user action*. The plugin will set a local variable to `true` when a user event is triggered. Then, the `updateSelection` will check if a user action was done and will use a query to do that. A new command is added to put it back to `false` when the update is finished.

I'm not perfectly used to the codebase yet. If there is concerns or design changes to be done, I'll be happy to learn and make those changes!

Fixes:  #2967
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -198,6 +198,7 @@ class Content extends React.Component {
     debug.update('componentDidUpdate')
 
     this.updateSelection()
+    this.props.editor.clearUserActionPerformed()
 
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Thanks @gracicot!

PR #3023, opened 2019-09-17
[Android] Update the selection value whenever possible

gracicot:
Bug

[Video showing the bug](https://i.imgur.com/EE3SRTu.mp4)
Notice the selection not being updated in the slate's value.

#### What's the new behavior?
The selection is now committed to the value when a selection event is triggered.

When the user moves the cursor either with a swipe or using the handle, the slate value should reflect the new, updated value.

[Video of the new behavior](https://i.imgur.com/0EDnSHG.mp4)
Notice the selection state being updated. Also notice the toolbar being updated.

Inside the `onSelect` function in the android plugin, we check if we can commit the selection and call `editor.select` with the range.

We commit the selection when all of those are true:
 - The range extracted from the DOM is set (`range.isSet`)
 - The diff from the composition is empty (`last.diff === null`)
 - The range is different than the current selection from the value (to reduce superfluous calls)

I also noticed that ranges were sent to the `editor.select` function. I will make a followup pull request to fix that too.

### DIFF
--- a/packages/slate-react/src/plugins/android/composition-manager.js
+++ b/packages/slate-react/src/plugins/android/composition-manager.js
@@ -583,15 +583,7 @@ function CompositionManager(editor) {
       }
 
       const isPointsEqual = (point1, point2) => {
-        if (point1.path.size !== point2.path.size) {
-          return false
-        }
-
-        if (point1.offset !== point2.offset) {
-          return false
-        }
-
-        return point1.key === point2.key
+        return point1.offset === point2.offset && point1.key === point2.key
       }
 
       const isRangesEqual = (range1, range2) => {
--- a/packages/slate-react/src/plugins/android/composition-manager.js
+++ b/packages/slate-react/src/plugins/android/composition-manager.js
@@ -30,6 +30,17 @@ function renderSync(editor, fn) {
   })
 }
 
+function isPointsEqual(point1, point2) {
+  return point1.offset === point2.offset && point1.key === point2.key
+}
+
+function isRangesEqual(range1, range2) {
+  return (
+    isPointsEqual(range1.anchor, range2.anchor) &&
+    isPointsEqual(range1.focus, range2.focus)
+  )
+}
+
 /**
  * Takes text from a dom node and an offset within that text and returns an
  * object with fixed text and fixed offset which removes zero width spaces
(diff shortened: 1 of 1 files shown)


gracicot:
By playing around with the editor with this patch applied, I found two regressions.

The first issue need soft breaks:

```js
function SoftBreak(options = {}) {
  return {
    onCommand(command, editor, next) {
      const { type, args } = command

      if (type === 'splitBlock') {
        const [n] = args

        if (n === undefined) {
          return editor.insertText('\n')
        }
      }

      return next()
    },
  }
}
```
To reproduce, type in an empty editor:
- "AA↵"
- Then type rapidly
A composition may end prematurely and cause the editor to repeat the few first insertions.

---

To reproduce the second issue, simply simply type a letter in an empty editor, the cursor will stay at the beginning of the text. Strangely, it only happen before the first `restoreDOM` call.

Here's what I discovered about the problem:
- There seem to be a selection event that tell the editor to select the cursor at position 1, but the text has not been entered yet in the value.
 - Slate sees the past the end selection and correct the position at 0.
 - Since the cursor position has changed, it is applied to the DOM, so the cursor is explicitly set to the beginning of the text.
 - Since a new render is done, the android plugin clear the current diff.
 - However, there is no need to re-render the whole DOM since the document value hasn't changed, only the selection, so the letter still appear in the DOM, but not in the value.

thesunny:
That's a good breakdown of the problem and now that you bring it up, I recall I ran across this issue as well, specifically creating a Range object that reflected a position in a Slate Value that doesn't yet exist.

I can't remember if I implemented that method but I get the feeling that I might have just worked around it in the mean time.

This, to me, is another argument for doing https://github.com/ianstormtaylor/slate/issues/3065 as the Slate Document and Slate Selection get updated at the same time in sync with each other.

gracicot:
I realized that setting the selection position in the composition end was redundant since a selection event was triggered right after the composition end event. Also, it seem like `window.getSelection().getRangeAt(0)` didn't report the correct offset all the time inside the composition end event. Taking the selection position inside the selection event yielded to more accurate results.

[Before last commit](https://i.imgur.com/kn3qCzK.mp4)
[After last commit](https://i.imgur.com/pC4IaJD.mp4)

ianstormtaylor:

Thank you for understanding.

PR #3024, opened 2019-09-18
Send selection to editor.select on android

gracicot:
This pull request depends on #3023 

Bug

#### What's the new behavior?
Selections are sent to the `editor.select` command instead of ranges

Instead of sending ranges, we take the existing selection from the editor and apply the range's anchor and focus.

Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/plugins/android/composition-manager.js
+++ b/packages/slate-react/src/plugins/android/composition-manager.js
@@ -31,6 +31,20 @@ function renderSync(editor, fn) {
 }
 
 /**
+ * Takes a the range and the editor. It will take the selection from the editor
+ * and apply the range's anchor and focus on it.
+ *
+ * @param {Editor} editor
+ * @param {Range} range
+ */
+
+function selectionFromRange(editor, range) {
+  return editor.value.selection
+    .moveAnchorTo(range.anchor)
+    .moveFocusTo(range.focus)
+}
+
+/**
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/plugins/android/composition-manager.js
```diff
@@ -582,6 +596,28 @@ function CompositionManager(editor) {
...
+      // It's important that we don't run the select middleware or commit the
+      // selection to the value when the diff has a value. When the diff is
+      // non-null, the text structure in the value is different compared to the DOM.
+      if (
+        range.isSet &&
+        last.diff === null &&
+        !isRangesEqual(editor.value.selection, range)
+      ) {
+        editor.select(selectionFromRange(editor, range))
+      }
```
gracicot: These changes are mostly from #3023. These should disappear when merging it.

gracicot:
I saw there is also `onDrop` that send a range into the `editor.select` command. Is the API allows this? If simple ranges are permitted in `editor.select`, then I'll simply close the pull request.

It also seem that partial object are usually allowed in commands. With #2495 a range is simply a partial selection.

If we want to move away from partial object, I can widen this pull request. But that's really up to you.

gracicot:
Ranges can be sent to `editor.select` and are handled correctly. I will close this for now.

PR #3025, opened 2019-09-18
Recover from exception when flushing inside android plugin

gracicot:
Fixing a bug

#### What's the new behavior?
The plugin will recover its state when an exception is thrown while flushing

Before, the `startAction` function would simply fush everything and then cleanup the state.

When an exception was thrown, the plugin would not recover correctly. When using error boundaries and then recovering, the editor would simply not flush anymore since the variable `isFlushing` would always be true. Adding the finally can make the editor recover from a throwing plugin or an error in the flush


gracicot:
I don't know what exception safety Slate usually ensure. I don't have a strong opinion on this one, but I think be able to recover the error here can be useful.

thesunny:
Ooh nice catch. Happy to see people making improvements on the Android code.

When I get a chance, I'll test on my devices but it looks good from here. Thanks!

ianstormtaylor:

Thank you for understanding.

PR #3035, opened 2019-09-27
Correct the document mentioned by #3029

### LINKED ISSUE #3029: There is a document mistake about the type of the marks
It says the property of a `Text` named `marks` in [document](https://docs.slatejs.org/slate-core/text) is a `Immutable.List<Mark>,`, while it's actually a `Immutable.Set<Mark>,`.

It cause an error and takes me half an hour to find it.  T^T

Would you please correct it?

winrey:
Fix #3029 

#### What's the new behavior?
Only update the document.

Only update the document.

NO NEED but still done  : )

Fixes: #3029 

### DIFF
--- a/docs/reference/slate/text.md
+++ b/docs/reference/slate/text.md
@@ -30,7 +30,7 @@ The text contents of this node.
 
 ### `marks`
 
-`Immutable.List<Mark>,`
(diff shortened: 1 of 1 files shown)


thesunny:
Looks good. Thank you!

Issue #2781, opened 2019-05-15
Placing accents over characters does not work


TheLastBug:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
When typing '´' + 'a' to get 'á', the accent is ignored and only 'a' is written (same result is seen when attempting to write a á à ã ă â e é è ê i í ì ĩ o ó ò õ ô ơ u ú ù ũ ư y ý ỳ).
http://recordit.co/WrUR6GiCt4

**Note**: When typing from the middle or the text, both the accent and the letter are ignored, and neither are written.

#### What's the expected behavior?
http://recordit.co/yqayENkyIu

**Note**: Pasting these composite characters directly works.

You may reproduce it here: https://www.slatejs.org/#/rich-text
Browser: Safari 12.0.1
OS: macOS Mojave 10.14.1

alanhogan:
I can reproduce this, also on Safari + Mojave

alanhogan:
(It's fine in Chrome + Mojave)

philn:
on macOS to get 'à' I long-press on 'a' and a popup shows up offering me the variants for that character. So IMHO this issue in invalid on macOS, although it can be valid on other platforms like Linux.

rikkit:
<kbd>\`</kbd> + <kbd>a</kbd> = `à` works for me as expected on Firefox on Windows with United Kingdom Extended keyboard

ksimons:
If I type <kbd>´</kbd> with a Norwegian keyboard, the cursor moves backwards to the left of the accent, whereas on all other editors, it correctly stays to the right.

knubie:
@philn There are two ways to input accented characters on macOS. The way that you described, which is the default behavior (**method A**), and the way that @TheLastBug describes (**method B**). The behaviour can be toggled with a defaults write command.

That said, there exists bugs in each method.

#### In Safari
- method A works as expected
- method B - The accent is ignored and the standard character is written

#### In Chrome
- method A - Both the standard character and the accent character are inserted
- method B  works as expected

I have no tested Firefox yet

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

Issue #2979, opened 2019-08-23
Programmatic changes remove the composition visual cue on macos


gracicot:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?

A programmatic change to the editor remove the visual cue we usually see on macos when starting a composition.

For example, when typing 'ê', you start with '^' then 'e'. When composition starts, you see the '^' character underlined then the composed letter appear when typing 'e'.

Here's a gif demonstrating the problem:

![http://g.recordit.co/3RxpemJZj6.gif](gif)

[Code Sandbox](https://codesandbox.io/s/slatejs-playground-xmokw)

The problem happens with both firefox and chrome.

#### What's the expected behavior?
The composition visual cue is not really a character. It should not be removed before composition ends. It should only be removed when the selection effectively move or when the composition has finished.

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #2743, opened 2019-05-07
Fix composition input on Safari (native InputEvents)

beldar:
#### Is this adding or improving a _feature_ or fixing a _bug_?

Bug fix

#### What's the new behavior?

Composition events are now handled on native InputEvents

#### How does this change work?

It would allow accented characters (i.e à, ò, ï, â, etc)

#### Have you checked that...?

* [x] The new code matches the existing patterns and styles.
* [x] The tests pass with `yarn test`.
* [x] The linter passes with `yarn lint`. (Fix errors with `yarn prettier`.)
* [x] The relevant examples still work. (Run examples with `yarn watch`.)

#### Does this fix any issues or need any specific reviewers?

Fixes: #2499

Issue #2499, opened 2018-12-14
Accentuated characters with ^ or ¨ are broken on Safari 


nicoxx:
On Safari 12.0.2, french keyboard
"slate-react": "0.19.3"
"slate": "0.42.2"

When I type ^ or ¨ before a letter it is not taken into account, I expect to generate a î or ï (useful for french words)

It works fine in firefox or chrome.
The problem appears on slatejs official exemple (cf. gif below) as in my setup (copy of rich text example)
I have not error in safari dev console

![slatejs_bug_circonflex_accent](https://user-images.githubusercontent.com/6695280/50001048-e615c400-ff9b-11e8-89d3-fed7aee0eb74.gif)

It seems different from #1319  since it doesn't raise any error

Thanks for your help and slate.js project, I worked on it for a few days and it's great !

jordiponsllaurado:
For me it's also happening with ` and ´.
I'm using version 0.44.7

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #3040, opened 2019-09-28
Fix composition editing in Safari

knubie:

Fixing a bug

#### What's the new behavior?

In Safari, composition input was broken in at least two ways:

1. Pressing option + an accent character, e.g. opt+n to insert a ~, then pressing n again to compose that into an ñ.
2. Using an IME to input something like Japanese or Korean.

For method 1., the old behavior was that a normal 'n' character would be inserted instead of an 'ñ'. The new behavior is that the `ñ' gets inserted.

For method 2. this was just completely broken and behaving all kinds of unexpected ways.

The the main issue is that immediately after `onCompositionStart` is fired, an `onSelect` event is fired which updates the selection and triggers an `onCompositionEnd` to fire, thus interrupting the composition before the user is able to complete it.

In addition to the above, specific to method 1. Safari would fire a `insertFromComposition` when the `n` was pressed a second time, which was not being handled.

Specific to 2. when an IME composition is selected with 'enter' it was completing the composition (firing `onCompositionEnd`), and *also* inserting a new line. This is because the `onCompositionEnd` is handled *before* `onKeydown` (or whichever event triggers the new line). By ensuring that `isComposing` is set to false *after* the `onKeydown` is handled, we can prevent the default behavior of inserting a new line.

Instead of blocking `updateSelection` outright in Safari, we could store the composition state in `this.tmp.isComposing` and conditionally block `updateSelection` based on that. I decided to block it outright to simplify the code, and following the Android exception.

Fixes: #2781 #2499 #2743
Doesn't fix:  ~~#2979~~
Reviewers: @ianstormtaylor @thesunny

### DIFF
--- a/packages/slate-react/src/components/content.js
+++ b/packages/slate-react/src/components/content.js
@@ -9,6 +9,7 @@ import { List } from 'immutable'
 import {
   IS_ANDROID,
   IS_FIREFOX,
+  IS_SAFARI,
   HAS_INPUT_EVENTS_LEVEL_2,
 } from 'slate-dev-environment'
 import Hotkeys from 'slate-hotkeys'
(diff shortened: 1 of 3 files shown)


knubie:
It may be more appropriate to check for `HAS_INPUT_EVENTS_LEVEL_2` instead of checking for `IS_SAFARI` here.

ianstormtaylor:

Issue #982, opened 2017-08-14
[IME] FF/macOS: entering accented letters from the Accent Menu inserts double character


danburzo:
#### Do you want to request a *feature* or report a *bug*?

There is a __bug__ in Firefox/macOS with entering an accented letter from the [Accent Menu](https://support.apple.com/kb/PH25643?locale=en_US) (by holding down a key and pressing the keys _1_ to _N_). 

#### What's the current behavior?

At the subsequent key press (e.g. moving the caret with left/right arrow key), the non-accented character appears along with the accented one. [See video here](https://www.dropbox.com/s/fwpsfwuuqi1lbgw/accent-menu.mov?dl=0)

#### What's the expected behavior?

A single, accented letter is inserted. Safari/Chrome work as expected.

#### Other RTEs

* _Draft.js_ exhibits the same problem
* _ProseMirror_ and _Quill_ don't seem to have it

danburzo:
Seems the Accent Menu does not fire `compositionstart` / `compositionend` events on Chrome & Safari, and Firefox fires both `compositionstart` and `compositionend` after the character has been composed.

There's also an [open, albeit deserted, issue](https://github.com/w3c/editing/issues/132) raised for the W3C Editing WG.

danburzo:
This works as expected with `isNative = false` but that does not make any sense to me, since when entering the character `à`, there are two `beforeinput`/`input` events, one with `a` and one with `à` 🤔

ianstormtaylor:
I believe this is fixed in the latest version of Slate.

javierfernandes:
This is still happening for me in `"slate": "^0.44.12",`
But I'm also able to reproduce this in the current demo editor from https://www.slatejs.org/#/rich-text
![slate-accent-menu-bug](https://user-images.githubusercontent.com/4428120/63605831-1c80f580-c5a5-11e9-8e23-18369dc9037b.gif)

Should this be reopened @ianstormtaylor  ?

Thanks

knubie:
This occurs in Chrome as well.

achou11:
There's a PR for this that's looks ready to be merged: https://github.com/ianstormtaylor/slate/pull/3041

Will probably require a slate upgrade

jens-lilt:
I can still see this occur in Slate 0.77.2 if I use the arrow keys to select the accent instead of typing the number shown underneath

PR #3041, opened 2019-09-28
Fix MacOS long-press accented char bug

knubie:

_bug_

#### What's the new behavior?

See #982 for details on the bug.

Screen Shot 2019-10-06 at 1 03 11 PM

The change works by trying to guess when an insertion from the popover is occurring. It does this by comparing the last `keyDown` value with the value of `onBeforeInput` (which is triggered after a selection is made). If the values are different, and the last `keyDown` value was a number between 1-9, and the text for `onBeforeInput` falls within the unicode range of an accented character, it will insert the character at the specified range. The range is specified as the current selection, with the focus moved back one character, thereby replacing the non-accented character that was inserted when the long press was initiated.

Fixes: #982
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-hotkeys/src/index.js
+++ b/packages/slate-hotkeys/src/index.js
@@ -38,6 +38,7 @@ const APPLE_HOTKEYS = {
   extendLineForward: 'opt+shift+down',
   redo: 'cmd+shift+z',
   transposeCharacter: 'ctrl+t',
+  longPressSelect: ['1', '2', '3', '4', '5', '6', '7', '8', '9'],
 }
 
 const WINDOWS_HOTKEYS = {
--- a/packages/slate-react/src/plugins/dom/after.js
+++ b/packages/slate-react/src/plugins/dom/after.js
@@ -3,7 +3,13 @@ import Debug from 'debug'
 import Hotkeys from 'slate-hotkeys'
 import Plain from 'slate-plain-serializer'
 import getWindow from 'get-window'
-import { IS_IOS, IS_IE, IS_EDGE } from 'slate-dev-environment'
+import {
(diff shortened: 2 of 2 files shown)


ianstormtaylor:
Hey @knubie thanks for the writeup here, helps to understand!

Do the native browser events that are being fired not ever expose selection-related information? I would have assumed that there might be a `composition*` or `beforeinput` event that had a `Range` associated with it telling us where the delete should occur?

knubie:
Thanks for taking a look @ianstormtaylor 

I wasn't able to find any useful range info for Chrome or Firefox, but it's possible that I could have overlooked something (or just don't know where to look). Here's a more detailed breakdown of my findings:

#### Chrome
Fires `onBeforeEvent` (synthetic, `nativeEvent` is `TextEvent`)
Couldn't find any useful range info here.

#### Firefox
Fires `onBeforeEvent` (synthetic, `nativeEvent` is [`compositionend`](https://developer.mozilla.org/en-US/docs/Web/API/CompositionEvent))
Couldn't find any useful range info here.

Fires `onCompositionEnd` (synthetic, `nativeEvent` is [`compositionend`](https://developer.mozilla.org/en-US/docs/Web/API/CompositionEvent))
Couldn't find any useful range info here

#### Safari
Fires `onBeforeEvent` (native)
`event.getTargetRanges() //=> {startOffset: 0, endOffset: 1, ...}`

knubie:
Also worth noting that insertion was completely broken on non-MacOS environments before my last commit (b12ae54), but it doesn't look like the CI tests were able to catch that.

ianstormtaylor:

PR #3044, opened 2019-09-30
Fix unicode offset

### LINKED ISSUE #2632: Deleting Plaintext Emoji as First Character Throws Error

### LINKED ISSUE #2635: Backspace does not properly delete complicated emojis

davidhoeller:
This is basically PR #2680 in the current version of slate (Fixes #2635, fixes #2632)
Thanks @adjourn for starting this.
Plus, fixes a bug where the cursor could not be moved beyond block with arrow keys.

#### What's the new behavior?
As explained in #2680 .
The emoji example has been extended to insert emojis as plain text (vs inline nodes) to demonstrate the new behaviour.

See #2680
Updated emoji example, see here:
Slate PR-3044 Fix unicode offset Example

Fixes: #2635, #2632
Reviewers: @adjourn @ianstormtaylor @kwokhuen

### DIFF
--- a/examples/emojis/index.js
+++ b/examples/emojis/index.js
@@ -37,7 +37,8 @@ const EMOJIS = [
   '👻',
   '🍔',
   '🍑',
-  '🔑',
+  '👩‍❤️‍👩',
+  '👨‍👩‍👦',
 ]
 
 /**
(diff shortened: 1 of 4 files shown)


ianstormtaylor:
Thanks @davidhoeller for update this! And huge thank you to @adjourn for taking the time to fix this in the first place.

scottluptowski:
Thank you so much!

PR #3049, opened 2019-10-04
Fix soft breaks on android using the diff's cursor

gracicot:
Fixing a bug

[Here's a video before the fix](https://i.imgur.com/6DuAfFq.mp4)
The soft break is done before the letter.

#### What's the new behavior?

We use the cursor position from the diff before applying the break.

[Here's a video after the fix](https://i.imgur.com/Vx5SVFe.mp4)

Previously, the editor would save a selection with the wrong offset, and apply that selection before doing the soft break. That would usually apply the soft break before the letter or the word that was entered.

To fix the issue, we accumulate previous adjacent text nodes offset and we also use the diff's cursor to apply the selection before splitting blocks.

Fixes: #

### DIFF
--- a/packages/slate-react/src/plugins/android/composition-manager.js
+++ b/packages/slate-react/src/plugins/android/composition-manager.js
@@ -43,10 +43,19 @@ function renderSync(editor, fn) {
  * @param {Boolean} isLastNode
  */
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:

Thank you for understanding.

Issue #2988, opened 2019-08-29
HTML content being pasted twice


moinism:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?
![twice](https://user-images.githubusercontent.com/2242528/63936985-cd423580-ca7a-11e9-9c1d-e4faaeabaac9.gif)

When HTML content is pasted, it is pasting it twice.

1. Copy some content from a webpage, even a line.
2. https://www.slatejs.org/#/paste-html
3. Paste.

#### What's the expected behavior?

Paste only once.

CameronAckermanSEL:
As far as I can tell, the defect is that the onPaste handler is firing a second time with a plain text paste as if `return next()` was being called to resume the plugin stack after the html is deserialized and inserted
https://jsfiddle.net/1nzto9vd/1/

larrywu13535:
Pasting also doesn't respect line breaks

CameronAckermanSEL:
That's probably related to how insert fragment works

eshavrov:
Hello
Need to change the `onPaste`
```js
  onPaste = (event, editor, next) => {
    event.preventDefault()  // <---- add me

    const transfer = getEventTransfer(event)
    if (transfer.type !== 'html') return next()
    const { document } = serializer.deserialize(transfer.html)
    editor.insertFragment(document)
  }
```

[link](https://github.com/ianstormtaylor/slate/issues/2972)

PR #3070, opened 2019-10-16
fixing paste html example

Entkenntnis:

Bug fix

#### What's the new behavior?

Paste HTML is working and not pasting content twice

Prevent default handler solves the issue.

Fixes: #2988
Reviewers: @


ianstormtaylor:
Thanks @Entkenntnis, good catch!

PR #3072, opened 2019-10-17
Fix when clear search text, all text split by blank

whyour:
In search-highlighting example, when search text is empty, no need to split text and addAnnotation.
Note dom changes in the right

hjhj
jjjjhhh


ianstormtaylor:
Thanks @whyour!

Issue #2061, opened 2018-08-08
fix spellcheck blinking as you type


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

One of the current issues with spellcheck is that when a node in the DOM is re-rendered, the browser seemingly takes a bit of time to update the spellcheck underlines, which cause a slight blink. So if you type a series of characters, with each character typed you'll see the red spellcheck underlines blinking.

#### What's the expected behavior?

To solve this for the 100% case is impossible, since it would mean never re-rendering. For example, selecting some text and add any mark to it will make the spellcheck underlines blink, but this transformation also isn't something the browser can do by itself (often).

However, solving for the 99% case is attainable, because it would just mean that inserting individual characters (which is the majority of Slate changes in any real world use case) should update the DOM directly, instead of re-rendering via React.

I think this could be solved by adding an `operation.isNative` property, perhaps with a convenient `change.asNative(c => { ... })` helper. This would allow us to create the same `insert_text` operations that we currently do, but flagged as "native" (eg. that they are already represented in the DOM) so that we can know that an individual `<Leaf>` does not need to re-render.

We'd need to also enforce that the `<Editor>` gets passed not only a `value=` property but also the previous `operations=` from a `Change` object if one exists. This way the current state has access to the previous operations to make the decision about whether to re-render.

juliankrispel:
I'm sure you've already seen this but this is how [draft js handles this problem](https://github.com/facebook/draft-js/blob/master/src/component/handlers/edit/editOnBeforeInput.js#L74)

steobrien:
Hey @ianstormtaylor, this is something we’d love to see fixed in Slate, and the pragmatic solution you hint at sounds exactly right. Any tips on where to get started with the implementation?

ianstormtaylor:
For more information on how best to solve this, if someone wants to help tackle it, check out https://github.com/ianstormtaylor/slate/issues/2051

tommoor:
@ianstormtaylor would you consider this to be the same bug, or something different? It really makes typing in any Slate input kind of a distracting experience as the spellcheck is constantly flashing under partially typed words…

![Kapture 2019-05-26 at 15 17 19](https://user-images.githubusercontent.com/380914/58387771-6586d500-7fc9-11e9-85ba-3ab87956ad07.gif)

ryanmitts:
[...]

PR #3076, opened 2019-10-22
Use native rendering for character insertion when possible to avoid spellcheck flicker.

ryanmitts:

Fixing a bug.

#### What's the new behavior?

See the original bug ticket for original behavior.

Before:

slate-bug-before

After:

slate-bug

Slate will re-render after each slate editor document operation. The most
noticeable and visually jarring behavior is when typing single characters
into the editor. Each character will cause the browser spellcheck underline
to re-render the block the insertion is in. This re-rendering also causes
the spellcheck to re-evaluate each misspelled word as letters are added
to it, instead of the usual behavior of waiting for typing to stop for a
period of time.

This change looks for when a single character is input in the `onBeforeInput`
method and when certain conditions are met, such as:

1. The selection is collapsed (only want to handle the simplest case right now)
2. The selection is not at the start of a text node (Chrome natively handles this inconsistently)

The editor command for this insertion is wrapped in a new `asNativeOperation`
function which will queue the operation generated by any command in the function
until flushed. The event is not prevented to allow the default behavior.

This queue is flushed in the `onInput` event, which happens right after the dom
is natively updated with the operation.

The leaf text nodes are updated to keep a reference to their DOM elements and
not render if their `textContent` matches what it will try to render. Since the
shadow DOM state is different than the actual DOM state for text nodes, we must
force it to update by changing the `key` on the node after each update. Otherwise
React thinks it doesn't need to change anything.

Fixes: #2061 
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/leaf.js
+++ b/packages/slate-react/src/components/leaf.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { createRef } from 'react'
 import Types from 'prop-types'
 import SlateTypes from 'slate-prop-types'
 import ImmutableTypes from 'react-immutable-proptypes'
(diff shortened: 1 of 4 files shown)


ajpalkovic:
This is a pretty cool change, we're probably to use it at Discord.  One thing I found is if you make one more change to this you actually fix a _ton_ of IME issues.  Native operations aren't flushed on composition end, and the onInput event is skipped if you are in the middle of a composition, so if you add `editor.controller.flushQueuedNativeOperations()` to onCompositionEnd in the before plugin, a lot of things start working better, like japanese input!

ryanmitts:
Seems the checks failed because it was being run around the same time that Node 13 was released and had problems downloading it.

ianstormtaylor:
Hey @ryanmitts, thank you so much for taking the time to work on this. I'm still very interested in this behavior to fix spellchecking. I think we can do it with a special property on the `Operation` objects now that they are just plain JSON in `0.50`. But this PR is now out of date since so much has changed. I'm going to close it out, but if you are down to resubmit it that would be amazing! Thank you for understanding.

PR #3077, opened 2019-10-23
Add a way to completely disable the default placeholder plugin

ajpalkovic:
I'd like to use slate-react, along with a custom placeholder.
The problem I'm running into though is that slate-react _always_ adds the PlaceholderPlugin pretty much.  Even setting the placeholder to an empty string will still actually add a magic div to the dom, and that's what breaking for me.

So, I figured adding support for passing _exactly_ null in to disable the plugin would work well.  Folks that don't pass in anything will still get the "default" of an "empty placeholder" so hopefully nothing breaks.

Improving a feature

#### What's the new behavior?
The placeholder plugin will not be added to the editor if the placeholder prop is exactly null.

The placeholder plugin will not be added to the editor if the placeholder prop is exactly null.

Fixes: #
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/plugins/react/index.js
+++ b/packages/slate-react/src/plugins/react/index.js
@@ -40,7 +40,7 @@ function ReactPlugin(options = {}) {
   // Disable placeholder for Android because it messes with reconciliation
   // and doesn't disappear until composition is complete.
   // e.g. In empty, type "h" and autocomplete on Android 9 and deletes all text.
-  const placeholderPlugin = IS_ANDROID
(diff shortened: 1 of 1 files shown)


ajpalkovic:
fwiw, this is a screenshot of the html of the existing placeholder example.  Notice that even though it specifies a custom PlaceholderPlugin, there's still an empty placeholder span added from the default instance of the plugin:
Screen Shot 2019-10-23 at 11 45 05 AM

ianstormtaylor:

PR #3079, opened 2019-10-27
Simplify the mentions example

0ctothorp:

Improving a feature.

#### What's the new behavior?

Simplified mentions example. Instead of displaying the suggestions dropdown, the `window.propmt` is displayed (as suggested in #2808)

When the user types in the `@` sign and a word immediately after it, the code makes a search for the character which name begins with the typed text. If there is only one result, the mention is placed in the editor. If there are 5 or less results, the user is presented a prompt with available names listed and the default text of the prompt is set to the text that the user typed after the `@`. After the user submits the prompt input, the next search is made and everything runs as described before.

Besides that, the `Suggestions` component is removed and `value.json` is moved inside `index.js`.  `users.json` is left in the separate file for the obvious reasons. Actually, this could be shortened to just a few names and also kept inside `index.js` - what do you think?

I'm open for the suggestions if this can be done better/easier.

Fixes: #2808


ianstormtaylor:
You are beautiful, thank you @0ctothorp!

PR #3083, opened 2019-10-28
Change renderNode docs to renderBlock and renderInline

0ctothorp:

Fixing docs.

#### What's the new behavior?

Change rendering docs to describe `renderBlock` and `renderInline` instead of deprecated `renderNode`.


ianstormtaylor:
Thanks @0ctothorp, this is great!

PR #3088, opened 2019-11-01
Fix android call next

gracicot:

Fixing a bug

#### What's the new behavior?

When writing such plugin:
```js
function MyPlugin() {
    return {
        onCompositionStart:() => {
            console.log("hello");
        }
    };
}
```

When starting a composition, `"hello"` is printed in the console. This works today on desktop.

However, this is not the case for Android. On Android, the Noop plugin was activated, and no hooks would call next.

As stated in `noop.js`:

> IMPORTANT:
>
> This plugin is detached (i.e. there is no way to turn it on in Slate).
> You must hard code it into `plugins/react/index`.

However, it was always activated in `dom/index.js`:
```js
  const androidPlugins = IS_ANDROID
    ? [AndroidPlugin(options), NoopPlugin(options)]
```

I added a switch to turn on or off the noop plugin and added calls to `next()` in the hooks of the Android plugin

Fixes: #
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/plugins/android/index.js
+++ b/packages/slate-react/src/plugins/android/index.js
@@ -44,16 +44,18 @@ function AndroidPlugin({ editor }) {
    * handle `onCompositionStart`
    */
 
-  function onCompositionStart() {
+  function onCompositionStart(event, e, next) {
     observer.onCompositionStart()
+    return next()
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ianstormtaylor:

Thank you for understanding.

Issue #3030, opened 2019-09-23
allow annotations to be serialized/deserialized


mrwnmonm:
#### Do you want to request a _feature_ or report a _bug_?
Report a bug.

#### What's the current behavior?
The serializer doesn't recognize annotations.

#### What's the expected behavior?
The serializer recognizes annotations and includes its markup in the output.

While annotations don't exist in the docs, it does exist in the [examples](https://github.com/ianstormtaylor/slate/blob/master/examples/search-highlighting/index.js#L188), but unfortunately, it is not recognized by the serializer or the deserializer.

CameronAckermanSEL:
I'm _pretty sure_ this is a deliberate design decision. Annotations are meant to be throw away, temporary mark up on the document and as a result it's not something that gets exported when the value is converted to JSON, and by extension when the value gets serialized to a different format. Marks would be more appropriate for permanency.

What is your use case for using the serializer?

mrwnmonm:
We convert marks to SSML, we want to apply some marks automatically at the beginning, and for some reason that doesn't work with marks, slate can't mark many words in the block, it just marks the first one.  
But it works with annotations, but annotations are not recognized by the serializer. Maybe we could apply annotations, then we go convert annotations to marks. Will try that.

ianstormtaylor:
I'd be down for the serializers to support annotations. They'd probably end up being serialized similarly to marks. And for deserializing they'd need to be de-duped by `key` to result in a single range. (Probably cheapest way is to just always grow the annotation based on the new `anchor/focus` points.)

@CameronAckermanSEL you're right, but that's for "decorations". Annotations now live in the JSON as part of the data model. (I realize this is a little confusing since they're so close in name! 😄)

justinweiss:
I think it's going to be a little more involved. In order to create an annotation, you need the full path / offset that annotation refers to. But it looks like deserialize doesn't have that context when building up nodes. You'd somehow need to pass partial paths down as you build the tree, or flag all of the annotated nodes somehow and build up the annotations later, right?

justinweiss:
It turns out flagging / fixing works well. I have a PR here: https://github.com/ianstormtaylor/slate/pull/3092

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

PR #3092, opened 2019-11-06
Support serialization / deserialization of annotations

justinweiss:

Adding a feature

#### What's the new behavior?

Annotations can now be serialized and deserialized in the same way marks can.

For serializing, we do the same thing as with marks -- we figure out which annotations apply to a node and pass them to a serializer.

Deserializing is harder, because we don't have paths to the nodes at the time we deserialize them. Instead, when we deserialize an annotation, we add the annotation object to a property of the serializer, and we keep a node-to-annotations lookup table. When we're done deserializing, we walk the value's nodes. If we find a node that has an annotation, we update the path of that annotation to include the node. If an annotation spans several nodes, we expand anchor / focus to span the entire range of those nodes.

Fixes: #3030 
Reviewers: @

### DIFF
--- a/packages/slate-html-serializer/src/index.js
+++ b/packages/slate-html-serializer/src/index.js
@@ -1,8 +1,8 @@
 import React from 'react'
 import { renderToStaticMarkup } from 'react-dom/server'
 import typeOf from 'type-of'
-import { Node, Value } from 'slate'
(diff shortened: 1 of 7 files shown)


ianstormtaylor:

Issue #3090, opened 2019-11-02
Support React Concurrent Blocking Mode


newtack:
We want to move as soon as possible to React Concurrent mode and are using Slate. 

As per React's documentation there is migration approach that gives most of the benefits without going all the way, see

https://reactjs.org/docs/concurrent-mode-adoption.html#migration-step-blocking-mode

It'd be great if Slate could support this soon.

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/3093.

Issue #3087, opened 2019-11-01
Mac's Text Replacement Doesn't Work


benawad:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

![slateg](https://user-images.githubusercontent.com/7872329/68037821-8baa5280-fc96-11e9-90d4-4fde1177c13b.gif)

#### What's the expected behavior?

`omw` should get deleted

ryanmitts:
Maybe this comment will be helpful in the future. I tried this in my CR that has native browser manipulation on simple insertions (#3076) and it also breaks. I'm new to all the InputEvent stuff, but seems like the beforeinput event is the synthetic React one. It looks the same as an insertText event, so I don't know a way to know if it's replacing text without looking at the DOM change after allowing the native manipulation to proceed. Problem is, Chrome messes up and deletes a lot of the nested elements for the text Nodes, so it breaks the reconciliation.

I tried what happens in Chrome outside of React using the native beforeInput event which should support InputEvents level two, but it also reports as inputType of insertText. This doesn't make sense as the spec makes it seem like insertReplacementText is more appropriate. This is also how Safari reports the event. I raised a Chrome bug for for this in the meantime ([1021460](https://bugs.chromium.org/p/chromium/issues/detail?id=1021460#c4))

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/3093.

Issue #3075, opened 2019-10-21
remove `isFocused` and `marks` from `Selection`?


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Idea / debt.

#### What's the current behavior?

Right now the `isFocused` and `marks` properties of a `Selection` are stored in the data model. But these things are tracking editor-specific state that doesn't apply across editor instance. This leads to leakiness:

It doesn't make sense to send a "blurring" selection operation to another editor while collaborating. It doesn't make sense for the history to be affected by blurring/focusing operations, or for undo to leave the editor blurred. And it doesn't make sense for selection marks to preserve across undos either—or at least Google Docs doesn't. All of these operations have to get ignored right now in these scenarios.

#### What's the expected behavior?

I think it would be better to have these properties be local-only. And closured to an instance of the editor (implemented in the React/DOM plugins instead).

This also means that `Selection` becomes just a plain `Range`, which is convenient.

justinweiss:
I agree on both those counts. I have code to massage those cases around undo and focus / blur, and it would be nice to get rid of them. Anecdotally, selection vs range has been tough for reviewers to wrap their minds around. I think this is a great idea.

thesunny:
So, my instinct is that moving these variables into local wouldn't affect this particular need but it feels like it's worth bringing up just in case.

One area where Slate currently can't handle compositions is when there are marks applied to a collapsed selection. Since the marks aren't reflected in the DOM when the user starts a composition, the text does not show the marks as it is being composed (e.g. bolded text will not be bold). Once the composition is completed, the DOM can be updated and the text shows as bold.

The solution is that when the user bolds text on a collapsed selection, a DOM node is inserted for the bold with no text inside and the selection is placed inside that node. Then when the user starts composing, it should be within that node.

So I guess I'd want to confirm that this adjustment would be compatible with this idea.

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/3093.

Issue #3061, opened 2019-10-11
slate-react assumes PlaceholderPlugin exists


marianoguerra:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

slate-react 0.22.9 fails to load because it assumes PlaceholderPlugin is loaded

https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugins/react/index.js#L4

Reproduction: https://marianoguerra.github.io/bugs/slate-placeholder-plugin/index.html 

Following instructions from https://github.com/ianstormtaylor/slate/blob/1aa1783c9c4f7d56a4836c3060194df9e6ac895f/docs/walkthroughs/using-the-bundled-source.md

Trying to load slate-react-placeholder fails too because in the compiled version it exports SlateReactPlaceholder and not PlaceholderPlugin: https://unpkg.com/slate-react-placeholder@0.2.8/dist/slate-react-placeholder.js

Fails on Chrome 77.0.3865.120 on Linux and Firefox Nightly on Linux but it's not a browser issue.

Versions:

* slate 0.47.8
* slate-react 0.22.9

#### What's the expected behavior?

The example loads and renders the editor, instead if fails with an error "TypeError: PlaceholderPlugin is not a function"

marianoguerra:
Forgot to mention, the example works if slate-react is 0.22.8 instead of 0.22.9

here's the same example that loads with 0.22.8: https://marianoguerra.github.io/bugs/slate-placeholder-plugin/works.html

ianstormtaylor:
Fixed by https://github.com/ianstormtaylor/slate/pull/3093.

PR #3093, opened 2019-11-06
Next

ianstormtaylor:
This work is ongoing and not ready yet.

---

### Differences (incomplete)

There are lots of changes in this branch. But here's an overview of the _big_ differences from an architectural point of view.

_More coming, still in progress…_

#### JSON

The data model is now comprised of simple JSON objects. Previously, it used [Immutable.js](https://immutable-js.github.io/immutable-js/) data structures. This is a huge change, and one that unlocks many other things. Hopefully it will also increase the average performance when using Slate. It also makes it much easier to get started for newcomers. This will be a large change to migrate from, but it will be worth it.

#### Interfaces & Namespaces

The data model is interface-based. Previously each model was an instance of a class. Now, not only is the data plain objects, but Slate only expects that the objects implement an interface. So custom properties that used to live in `node.data` can now live at the top-level of the nodes. Helpers are exposed as a collection of helper functions on a namespace. For example, `Node.get(root, path)` or `Range.isCollapsed(range)`. This ends up making code much clearer because you can always quickly see what interface you're working with.

#### TypeScript

The codebase now uses TypeScript. Working with pure JSON as a data model, and using an interface-based API are two things that have been made easier by migrating to TypeScript. You don't need to use it yourself, but if you do you'll get a lot more security when using the APIs. (And if you use VS Code you'll get nice autocompletion regardless!)

#### Fewer Concepts

The number of interfaces and commands has been reduced. Previously `Selection`, `Annotation`, `Decoration` used to all be separate classes. Now they are simply objects that implement the `Range` interface. Previously `Block` and `Inline` were separate, now they are objects that implement the `Element` interface. Previously there was a `Document` and `Value`, but now the top-level `Editor` contains the children nodes of the document itself.

The number of commands has been reduced too. Previously we had commands for every type of input, like `insertText`, `insertTextAtRange`, `insertTextAtPath`. These have been merged into a smaller set of more customizable commands, eg. `insertText` which can take `at: Path | Range | Point`.

#### Fewer Packages

In attempt to decrease the maintenance burden, and because the new abstraction and APIs in Slate's core packages make things much easier, the total number of packages has been reduced. Things like `slate-plain-serializer`, `slate-base64-serializer`, etc. have been removed and can be implemented in userland easily if needed. Even the `slate-html-deserializer` can now be implemented in userland (in ~10 LOC leveraging `slate-hyperscript`). And internal packages like `slate-dev-environment`, `slate-dev-test-utils`, etc. are no longer exposed because they are implementation details.

#### Plugins

Plugins are now plain functions that augment the `Editor` object they receive and return it again. For example can augment the command execution by composing the `editor.exec` function. Or listen to operations by composing `editor.apply`. Previously they relied on a custom middleware stack, and they were just bags of handlers that got merged onto an editor. Now we're using plain old function composition (aka wrapping) instead.

#### Elements

Block-ness and inline-ness is now a runtime choice. Previously it was baked into the data model with the `object: 'block'` or `object: 'inline'` attributes. Now, it checks whether an "element" is inline or not at runtime. For example, you might check to see that `element.type === 'link'` and treat it as inline.

#### `beforeinput`

We now use the `beforeinput` event almost exclusively. Instead of having relying on a series of shims and the quirks of React synthetic events, we're now using the standardized `beforeinput` event as our baseline. It is fully supported in Safari and Chrome, will soon be supported in the new Chromium-based Edge, and is currently being worked on in Firefox. In the meantime there are a few patches to make Firefox work. Internet Explorer is no longer supported in core out of the box.

#### More React-ish

Rendering and event-handling is no longer a plugin's concern. Previously plugins had full control over the rendering logic, and event-handling logic in the editor. This creates a bad incentive to start putting **all** rendering logic in plugins, which puts Slate in a position of being a wrapper around all of React, which is very hard to do well. Instead, the new architecture has plugins focused purely on the rich-text aspects, and leaves the rendering and event handling aspects to React.

#### Context

Previously the `<Editor>` component was doing double duty as a sort of "controller" object and also the `contenteditable` DOM element. This led to a lot of awkwardness in how other components worked with Slate. In the new version, there is a new `<Slate>` context provider and a simpler `<Editable>` `contenteditable`-like component. By putting the `<Slate>` provider higher up in your component tree, you can share the editor directly with toolbars, buttons, etc. using the `useSlate` hook.

#### Hooks

In addition to the `useSlate` hook, there are a handful of other hooks. For example the `useSelected` and `useFocused` hooks help with knowing when to render selected states (often for void nodes). And since the use React's Content API they will automatically re-render when their state changes.

---

### Reductions

One of the goals was to dramatically simplify a lot of the logic in Slate to make it easier to maintain and iterate on. This was done by refactoring to better base abstractions that can be built on, by leveraging modern DOM APIs, and by migrating to simpler React patterns.

To give you a sense for the change in total lines of code:

```
slate                       8,436  ->  4,038  (48%)
slate-react                 3,905  ->    715  (18%)

slate-base64-serializer        38  ->      0
slate-dev-benchmark           340  ->      0
slate-dev-environment         102  ->      0
slate-dev-test-utils           44  ->      0
slate-history                   0  ->    201
slate-hotkeys                  62  ->      0
slate-html-serializer         253  ->      0
slate-hyperscript             447  ->    410
slate-plain-serializer         56  ->      0
slate-prop-types               62  ->      0
slate-react-placeholder        62  ->      0
slate-schema                    0  ->    504

total                      13,807  ->  5,868  (43%)
```

It's quite a big difference, although it's not done so the final sizes will likely grow a bit before it's ready. But that doesn't even include the dependencies that were shed in the process too.

---

### Fixes

This is an estimate of which issues are fixed by this pull request. There are a lot of them, because it changes a lot of things. There might be a few incorrectly "fixed" ones here, so if one of them is your issue and you don't think it's fixed feel free to reopen a new issue.

Fixes https://github.com/ianstormtaylor/slate/issues/3087
Fixes https://github.com/ianstormtaylor/slate/issues/3056
Fixes https://github.com/ianstormtaylor/slate/issues/3090
Fixes https://github.com/ianstormtaylor/slate/issues/3075
Fixes https://github.com/ianstormtaylor/slate/issues/2890
Fixes https://github.com/ianstormtaylor/slate/issues/2325
Fixes https://github.com/ianstormtaylor/slate/issues/3007
Fixes https://github.com/ianstormtaylor/slate/issues/3061
Fixes https://github.com/ianstormtaylor/slate/issues/560
Fixes https://github.com/ianstormtaylor/slate/issues/2869
Fixes https://github.com/ianstormtaylor/slate/issues/3028
Fixes https://github.com/ianstormtaylor/slate/issues/3027
Fixes https://github.com/ianstormtaylor/slate/issues/1762
Fixes https://github.com/ianstormtaylor/slate/issues/1022
Fixes https://github.com/ianstormtaylor/slate/issues/3020
Fixes https://github.com/ianstormtaylor/slate/issues/2746
Fixes https://github.com/ianstormtaylor/slate/issues/2991
Fixes https://github.com/ianstormtaylor/slate/issues/2333
Fixes https://github.com/ianstormtaylor/slate/issues/2711
Fixes https://github.com/ianstormtaylor/slate/issues/2413
Fixes https://github.com/ianstormtaylor/slate/issues/2345
Fixes https://github.com/ianstormtaylor/slate/issues/2812
Fixes https://github.com/ianstormtaylor/slate/issues/2859
Fixes https://github.com/ianstormtaylor/slate/issues/2862
Fixes https://github.com/ianstormtaylor/slate/issues/2863
Fixes https://github.com/ianstormtaylor/slate/issues/2495
Fixes https://github.com/ianstormtaylor/slate/issues/2864
Fixes https://github.com/ianstormtaylor/slate/issues/2860
Fixes https://github.com/ianstormtaylor/slate/issues/2878
Fixes https://github.com/ianstormtaylor/slate/issues/2858
Fixes https://github.com/ianstormtaylor/slate/issues/2861
Fixes https://github.com/ianstormtaylor/slate/issues/2985 
Fixes https://github.com/ianstormtaylor/slate/issues/2987
Fixes https://github.com/ianstormtaylor/slate/issues/1883
Fixes https://github.com/ianstormtaylor/slate/issues/2002
Fixes https://github.com/ianstormtaylor/slate/issues/2022
Fixes https://github.com/ianstormtaylor/slate/issues/2989
Fixes https://github.com/ianstormtaylor/slate/issues/2990
Fixes https://github.com/ianstormtaylor/slate/issues/2876
Fixes https://github.com/ianstormtaylor/slate/issues/2983
Fixes https://github.com/ianstormtaylor/slate/issues/2945
Fixes https://github.com/ianstormtaylor/slate/issues/2986
Fixes https://github.com/ianstormtaylor/slate/issues/2939
Fixes https://github.com/ianstormtaylor/slate/issues/2968
Fixes https://github.com/ianstormtaylor/slate/issues/2895
Fixes https://github.com/ianstormtaylor/slate/issues/2900
Fixes https://github.com/ianstormtaylor/slate/issues/2873
Fixes https://github.com/ianstormtaylor/slate/issues/2567
Fixes https://github.com/ianstormtaylor/slate/issues/2868
Fixes https://github.com/ianstormtaylor/slate/issues/2867
Fixes https://github.com/ianstormtaylor/slate/issues/2668
Fixes https://github.com/ianstormtaylor/slate/issues/2029
Fixes https://github.com/ianstormtaylor/slate/issues/2759
Fixes https://github.com/ianstormtaylor/slate/issues/2701
Fixes https://github.com/ianstormtaylor/slate/issues/1884
Fixes https://github.com/ianstormtaylor/slate/issues/2503
Fixes https://github.com/ianstormtaylor/slate/issues/2620
Fixes https://github.com/ianstormtaylor/slate/issues/2420
Fixes https://github.com/ianstormtaylor/slate/issues/2708
Fixes https://github.com/ianstormtaylor/slate/issues/2538
Fixes https://github.com/ianstormtaylor/slate/issues/1247
Fixes https://github.com/ianstormtaylor/slate/issues/2466
Fixes https://github.com/ianstormtaylor/slate/issues/2111
Fixes https://github.com/ianstormtaylor/slate/issues/2297
Fixes https://github.com/ianstormtaylor/slate/issues/2321
Fixes https://github.com/ianstormtaylor/slate/issues/2329
Fixes https://github.com/ianstormtaylor/slate/issues/2060
Fixes https://github.com/ianstormtaylor/slate/issues/2274
Fixes https://github.com/ianstormtaylor/slate/issues/2108
Fixes https://github.com/ianstormtaylor/slate/issues/1466
Fixes https://github.com/ianstormtaylor/slate/issues/1759
Fixes https://github.com/ianstormtaylor/slate/issues/2043
Fixes https://github.com/ianstormtaylor/slate/issues/1128
Fixes https://github.com/ianstormtaylor/slate/issues/1464
Fixes https://github.com/ianstormtaylor/slate/issues/721
Fixes https://github.com/ianstormtaylor/slate/issues/674
Fixes https://github.com/ianstormtaylor/slate/issues/2977
Fixes https://github.com/ianstormtaylor/slate/issues/2336
Fixes https://github.com/ianstormtaylor/slate/issues/2361
Fixes https://github.com/ianstormtaylor/slate/issues/2093
Fixes https://github.com/ianstormtaylor/slate/issues/802
Fixes https://github.com/ianstormtaylor/slate/pull/2871
Fixes https://github.com/ianstormtaylor/slate/pull/3113
Fixes https://github.com/ianstormtaylor/slate/pull/2734
Fixes https://github.com/ianstormtaylor/slate/pull/3077
Fixes https://github.com/ianstormtaylor/slate/pull/3040
Fixes https://github.com/ianstormtaylor/slate/pull/2144
Fixes https://github.com/ianstormtaylor/slate/pull/3041
Fixes https://github.com/ianstormtaylor/slate/pull/2743
Fixes https://github.com/ianstormtaylor/slate/pull/2680
Fixes https://github.com/ianstormtaylor/slate/pull/2937
Fixes https://github.com/ianstormtaylor/slate/pull/2907
Fixes https://github.com/ianstormtaylor/slate/pull/2931
Fixes https://github.com/ianstormtaylor/slate/pull/2871 
Fixes https://github.com/ianstormtaylor/slate/pull/2850
Fixes https://github.com/ianstormtaylor/slate/pull/2436
Fixes https://github.com/ianstormtaylor/slate/pull/1823
Fixes https://github.com/ianstormtaylor/slate/pull/2073
Fixes https://github.com/ianstormtaylor/slate/pull/1576
Fixes https://github.com/ianstormtaylor/slate/pull/1490
Fixes https://github.com/ianstormtaylor/slate/pull/1202
Fixes https://github.com/ianstormtaylor/slate/pull/1027

---

### Eliminates

In addition to fixing a lot of things, the significant changes to the architecture mean that a lot of open issues are no longer relevant because they are talking about concepts  that no longer exist.

https://github.com/ianstormtaylor/slate/issues/3022
https://github.com/ianstormtaylor/slate/issues/3050
https://github.com/ianstormtaylor/slate/issues/3061
https://github.com/ianstormtaylor/slate/issues/3013
https://github.com/ianstormtaylor/slate/issues/3012
https://github.com/ianstormtaylor/slate/issues/3006
https://github.com/ianstormtaylor/slate/issues/2886
https://github.com/ianstormtaylor/slate/issues/2932
https://github.com/ianstormtaylor/slate/issues/2829
https://github.com/ianstormtaylor/slate/issues/2328
https://github.com/ianstormtaylor/slate/issues/2856
https://github.com/ianstormtaylor/slate/issues/2634
https://github.com/ianstormtaylor/slate/issues/2472
https://github.com/ianstormtaylor/slate/issues/2742
https://github.com/ianstormtaylor/slate/issues/2744
https://github.com/ianstormtaylor/slate/issues/2549
https://github.com/ianstormtaylor/slate/issues/2672
https://github.com/ianstormtaylor/slate/issues/2673
https://github.com/ianstormtaylor/slate/issues/2640
https://github.com/ianstormtaylor/slate/issues/2296
https://github.com/ianstormtaylor/slate/issues/826
https://github.com/ianstormtaylor/slate/issues/2330
https://github.com/ianstormtaylor/slate/issues/1915
https://github.com/ianstormtaylor/slate/issues/2331
https://github.com/ianstormtaylor/slate/issues/1749
https://github.com/ianstormtaylor/slate/issues/2289
https://github.com/ianstormtaylor/slate/issues/2082
https://github.com/ianstormtaylor/slate/issues/2078
https://github.com/ianstormtaylor/slate/issues/1843
https://github.com/ianstormtaylor/slate/issues/1703
https://github.com/ianstormtaylor/slate/issues/1794
https://github.com/ianstormtaylor/slate/issues/1821
https://github.com/ianstormtaylor/slate/issues/1652
https://github.com/ianstormtaylor/slate/issues/1464
https://github.com/ianstormtaylor/slate/issues/803
https://github.com/ianstormtaylor/slate/issues/1367
https://github.com/ianstormtaylor/slate/issues/651
https://github.com/ianstormtaylor/slate/issues/873
https://github.com/ianstormtaylor/slate/pull/1023
https://github.com/ianstormtaylor/slate/pull/784
https://github.com/ianstormtaylor/slate/pull/1756
https://github.com/ianstormtaylor/slate/pull/1626
https://github.com/ianstormtaylor/slate/pull/1827
https://github.com/ianstormtaylor/slate/pull/2280
https://github.com/ianstormtaylor/slate/pull/2410
https://github.com/ianstormtaylor/slate/pull/2448
https://github.com/ianstormtaylor/slate/pull/2660
https://github.com/ianstormtaylor/slate/pull/2602
https://github.com/ianstormtaylor/slate/pull/2505
https://github.com/ianstormtaylor/slate/pull/2674
https://github.com/ianstormtaylor/slate/pull/2928
https://github.com/ianstormtaylor/slate/pull/3003
https://github.com/ianstormtaylor/slate/pull/3092


steida:
Congrats! This is quite a lot of hard work! A big step for Slate.

steida:
Very good work, indeed! Congrats to huge refactoring toward even better Slate.

ianstormtaylor:
@steida thanks! Yeah, I'm open to them. I can't promise anything of course, but if you have any thoughts I'd be curious to hear them.

steida:
Great. Ideally, I would love to not create another contentEditable editor unless necessary. 

The first question, why Range is text only? Native contentEditable allows stepping with caret before and after images, which makes sense to me. In my editor experiment, I just translate what DOM Selection (I see you adopt the same workaround for name clashes as I did) gives us to the path. The idea is to have the exact one to one match with DOM. Another related question is, why text nodes need a span wrappers? 

As for code itself, I see classes and this's. Do you plan more changes toward functional architecture?

ianstormtaylor:
> The first question, why Range is text only? Native contentEditable allows stepping with caret before and after images, which makes sense to me. In my editor experiment, I just translate what DOM Selection (I see you adopt the same workaround for name clashes as I did) gives us to the path. The idea is to have the exact one to one match with DOM.

I'm open to this, but the reason it is the way it is right now is because it seems to dramatically simplify logic. When you have a `Point` you are safe to assume it refers to a text node, and that `offset` is an index inside that text node. I've worked with the DOM API and it is pretty awkward to be constantly figuring out which "offset" you are working with.

But there would be positives, in that we could allow void nodes to be completely empty.

> Another related question is, why text nodes need a span wrappers?

There are currently a handful of wrapper. I don't think all of them are necessary, and ideally we could do with just one per leaf potentially. I remember that some of them are necessary because browsers do weird things in the DOM if it's just DOM text nodes. But if you have DOM element node wrappers with attributes on them then they do less weird things.

> As for code itself, I see classes and this's. Do you plan more changes toward functional architecture?

Only if it has actual benefits. I'm not a functional purist for purity's sake.

I don't love the classes, but I haven't yet found a better way to model the system where we need (a) overridability and (b) statefulness at the same time. Classes are purpose built for this use case.

steida:
> I've worked with the DOM API and it is pretty awkward to be constantly figuring out which "offset" you are working with.

I know. But I figured it. For dynamic non-model related inline / block detection, we have to "materialize" selection/range anyway.

> But if you have DOM element node wrappers with attributes on them then they do less weird things.

True. But if a parent element itself has any attribute, it works. I made a rule to enforce any attribute because almost nobody will use bare tags I suppose. It can be handled without wrappers, as I experimented with. 

> I'm not a functional purist for purity's sake.

Nor me, I am highly pragmatic and the only reason why I adopted fp-ts (mature, battle-tested) was it actually helped me to write better code. I started only with pipe and Predicate interface, then Eq, Ord attracted my attention..., then I was caught. It makes code more readable and helps us with its structure. For example, this is how we can compare Value https://github.com/evolu-io/evolu/blob/master/packages/evolu/models/value.ts#L21 it composes El, and Selection, which composes paths etc. Very handy.

> I don't love the classes, but I haven't yet found a better way to model the system where we need (a) overridability and (b) statefulness at the same time. Classes are purpose-built for this use case.

I understand. But the state is hard to reason about etc. etc. If you describe the usage you would like to have, we can perhaps design maybe better API / architecture together. Now or never :-)

ianstormtaylor:
> For dynamic non-model related inline / block detection, we have to "materialize" selection/range anyway.

What do you mean by that?

> But if a parent element itself has any attribute, it works. I made a rule to enforce any attribute because almost nobody will use bare tags I suppose. It can be handled without wrappers, as I experimented with.

If that's the case then that sounds great. It sounds like a non-architectural issue so it's something that can easily be PR'd after this lands.

> Nor me, I am highly pragmatic and the only reason why I adopted fp-ts (mature, battle-tested) was it actually helped me to write better code. I started only with pipe and Predicate interface, then Eq, Ord attracted my attention..., then I was caught. It makes code more readable and helps us with its structure. For example, this is how we can compare Value https://github.com/evolu-io/evolu/blob/master/packages/evolu/models/value.ts#L21 it composes El, and Selection, which composes paths etc. Very handy.

Fair enough, but I don't really want to have this argument here. This is the same age old argument that every functional purist has. That code doesn't look like an improvement to me. I'm more concerned with actual architectural changes and improvements you want to suggest for the current codebase.

> I understand. But the state is hard to reason about etc. etc. If you describe the usage you would like to have, we can perhaps design maybe better API / architecture together. Now or never :-)

The usage is what's already built with Slate… the examples in this repository, the plugins, etc. We need the ability to (a) define custom commands/queries, (b) override the behaviors for built-in commands/queries, (c) perform all of these against the editor's current state. If you have specific improvements to the architecture I'd love to hear them.

steida:
As for inline / block detection, maybe I am overlooking something, but why the model should define it? Blocks can be detected by computed styles. Am I correct?

As for the usage, we can have classes with a state, or functional composition like
 pipe(value, insertText('foo'), selectWord()) etc. There is a reason why React slowly propagates pure functions everywhere and no 'this', but it's hard to explain why shortly and judging such API only at first sight is not enough. I will design API on my experimental editor because code is a better argument than words, and then we can continue.

Anyway, thank you for responses.

ianstormtaylor:
> As for inline / block detection, maybe I am overlooking something, but why the model should define it? Blocks can be detected by computed styles. Am I correct?

That sounds tightly coupled to the DOM? Which won't work when you want to run commands in non-DOM contexts.

> As for the usage, we can have classes with a state, or functional composition like
pipe(value, insertText('foo'), selectWord()) etc. There is a reason why React slowly propagates pure functions everywhere and no 'this', but it's hard to explain why shortly and judging such API only at first sight is not enough. I will design API on my experimental editor because code is a better argument than words, and then we can continue.

Sounds good!

As for React, I agree. But to get to get good ergonomics while moving more functional they are forced to re-implement things approximating algebraic effects (eg. Hooks, Suspense). The tradeoff is that the framework itself has to take on a **lot** of complexity. And every bit of logic has to be wrapped in a "context" that the framework has tight control over that is able to be arbitrarily resumed—for React this is the `render` loop.

If "effects" were built in to JavaScript natively I'm fairly sure we'd be using them for overriding commands—they would be a near perfect fit. But without them, we'd need to re-implement a similar kind of thing that React Suspense is doing. 

I'm not opposed to it. It's just a lot of work and for now I'm still unsure on whether the benefits are worth the costs.

steida:
> That sounds tightly coupled to the DOM? Which won't work when you want to run commands in non-DOM contexts.

I see. You want to modify Value in Node.js for example. Probably because syncing or something. Good point.

> take on a lot of complexity. 

They wrote somewhere the whole implementation of Hooks added 1kb gzipped. But I see your point.

> which is super hard to read/debug 

It's not like RxJS, which is much harder to debug, agreed. fp-ts pipe is just plain function, nothing fancy. The problem with various Ramda pipes etc. is, imho, it's not type-safe and never will be, because of how Ramda and the others auto curry everything. fp-ts doesn't do that. As for readability, check this https://vimeo.com/97344498#t=3m42s (give it five minutes) But I agree with you, because of real side effects, not just some mappable state, people would have to pass something additional to those mappers. Whether it's worth it or not, let's see. This week I will experiment with it.

TheSpyder:
This PR is definitely heading in a good direction. For context I'm on a team building around just the Slate engine with pure DOM, no React at all. Some of the things we've been doing have followed stated directions in other github issues, such as avoiding keys, and others line up with changes in this PR. I hope to have some more detailed feedback soon.


> If you describe the usage you would like to have, we can perhaps design maybe better API / architecture together. Now or never :-)

I like FP a lot but conflating major changes is always a bad idea imo. Let one settle before you start pushing for another 😂 

I agree it would be nice if Slate's architecture and API could be expressed in a compositional form instead of with classes - it might lead to a smaller bundle size - but Slate has been doing just fine the way it is.


> `pipe`, `reduce`, `fold`, etc. which is super hard to read/debug and hasn't really caught on anywhere in JavaScript

I've been using them in JS since about 2013; they're not that bad once you get used to them. They can be clunky but they do help to express logic succinctly when used properly.

rikkit:
Hi @ianstormtaylor, can I ask for a little bit more context on this section: 

> We now use the beforeinput event almost exclusively. Instead of having relying on a series of shims and the quirks of React synthetic events, we're now using the standardized beforeinput event as our baseline. It is fully supported in Safari and Chrome, will soon be supported in the new Chromium-based Edge, and is currently being worked on in Firefox. Internet Explorer is no longer supported in core out of the box.

- What's the reason for this change - does it mean Slate will work better with IMEs or is that something different? Or is it to just simplify the code?
- When you write IE is not supported out of the box - will it be possible to support it for those who need to? Is this change polyfillable for older browsers (not just IE11, but Edge, iOS Safari, etc)

Cheers

ianstormtaylor:
Hey @rikkit! Good questions...

> What's the reason for this change - does it mean Slate will work better with IMEs or is that something different? Or is it to just simplify the code?

My hope is that it will definitely make it easier to integrate with IMEs. In Safari right now (which is the closest to matching the [Input Events Level 2 spec](https://w3c.github.io/input-events/#dom-inputevent-inputtype)) I believe IME input is fully supported. Chrome has a few bugs in their IME implementation (eg. [1022204](https://bugs.chromium.org/p/chromium/issues/detail?id=1022204)), but I think they're fairly close and easy to adjust for.

It's also absolutely about simplifying the code paths for editing. The previous system of relying on React's event polyfills was very hard to keep of track because the polyfills themselves aren't well documented. And since it was using non-standard events and lots of `preventDefault` it made falling through to the "correct" behaviors on newer browser a lot more difficult than it should be.

I'd like for the new system to be "correct by default", with extra non-standard behaviors layered over the top afterwards—ideally even in separate packages.

> When you write IE is not supported out of the box - will it be possible to support it for those who need to? Is this change polyfillable for older browsers (not just IE11, but Edge, iOS Safari, etc)

My hope is exactly that, that people will be able to polyfill for IE support if they need it. It's just that I won't be the one maintaining all of that code, it will require the community to work together or someone to champion it.

Right now as far as I can tell this branch's support looks like:

- [x] Safari
- [x] Chrome (known issue: [1022204](https://bugs.chromium.org/p/chromium/issues/detail?id=1022204))
- [x] Edge Beta (the [next version to be released](https://www.microsoftedgeinsider.com/en-us/download/))
- [ ] Firefox (known issues: [970802](https://bugzilla.mozilla.org/show_bug.cgi?id=970802), [1305387](https://bugzilla.mozilla.org/show_bug.cgi?id=1305387), [1449831](https://bugzilla.mozilla.org/show_bug.cgi?id=1449831))
- [x] iOS
- [ ] Android (unknown support?)
- [ ] ~~IE (not supported)~~

Feel free to ask more!

steida:
@ianstormtaylor It seems you are not using react-hooks eslint plugin. I highly recommend it for exhaustive checks.

ianstormtaylor:
@steida actually I think all the ESLint plugins are not working for some reason on this branch! 😄 But that sounds good. Getting everything to work across Babel, TypeScript, etc. is a huge pain, so I probably screwed up something in there somewhere.

steida:
@ianstormtaylor It is. Maybe you can take something from https://github.com/evolu-io/evolu

ianstormtaylor:
@ryanmitts great calls! I believe the caret jumping is due to the lack of keys, since the node gets re-rendered and React doesn't make an attempt to preserve the selection. I've got keys being preserved locally, hoping to push it up soon.

As for the selection error. I think that's due to the use of `NODE_TO_ELEMENT`, which fails when the node changes. With the keys being preserved, I've changed it to `KEY_TO_ELEMENT` locally and I think that solves that issue—hopefully. 

The "resolving things backwards from the render tree" is one of the most confusing parts of logic, maybe there's a smarter way to be doing those things.

I really appreciate you checking these things out! Let me know if you see anything else that looks off.

packages/slate-react/src/plugin/dom-helpers.ts
```diff
@@ -0,0 +1,349 @@
+import { Element, Node, Point, Range, Value } from 'slate'
+
```
ryanmitts: Looks like this file and commands are moved into the main plugin now?

packages/slate-react/src/components/editor.tsx
```diff
@@ -0,0 +1,488 @@
...
+            // out from under `contenteditable` elements, which leads to weird
+            // behaviors so we have to disable it like editor. (2017/04/24)
+            data-gramm={false}
+            role={readOnly ? undefined : props.role || 'textbox'}
+            {...attributes}
+            data-slate-editor
+            data-slate-node="value"
+            contentEditable={readOnly ? undefined : true}
+            suppressContentEditableWarning
+            ref={ref}
```
ryanmitts: I was thinking, and maybe this can be a feature to worry about as a future item, that you could wrap the editor in [`forwardRef`](https://reactjs.org/docs/forwarding-refs.html). Then you can combine that ref with the ref used internally to allow users to pass in their own ref to attach to the editor.

packages/slate-react/src/components/editor.tsx
```diff
@@ -0,0 +1,488 @@
...
+              // Prevent the default outline styles.
+              outline: 'none',
+              // Preserve adjacent whitespace and new lines.
+              whiteSpace: 'pre-wrap',
+              // Allow words to break if they are too long.
+              wordWrap: 'break-word',
+              // Allow for passed-in styles to override anything.
+              ...style,
+            }}
+            onBlur={useCallback((event: React.FocusEvent) => {
```
ryanmitts: How would you attach an `onFocus` and `onBlur` hook now?

ryanmitts:
> @ryanmitts great calls! I believe the caret jumping is due to the lack of keys, since the node gets re-rendered and React doesn't make an attempt to preserve the selection. I've got keys being preserved locally, hoping to push it up soon.
> 
> As for the selection error. I think that's due to the use of `NODE_TO_ELEMENT`, which fails when the node changes. With the keys being preserved, I've changed it to `KEY_TO_ELEMENT` locally and I think that solves that issue—hopefully.
> 
> The "resolving things backwards from the render tree" is one of the most confusing parts of logic, maybe there's a smarter way to be doing those things.
> 
> I really appreciate you checking these things out! Let me know if you see anything else that looks off.

They look to be all fixed now! Thanks!

packages/slate-react/src/plugin/dom-helpers.ts
ianstormtaylor: Ah yup. I forgot to delete the previously split out ones. Had to combine them to get TypeScript to play nice with extending classes and `this`.

packages/slate-react/src/components/editor.tsx
ianstormtaylor: I was thinking, but haven’t gotten around to implementing, that maybe we could call any extra `on*` handlers passed in that clash first. And then maybe check to see if propagation is stopped and that would let you completely override the defaults if needed? Dunno.

packages/slate-react/src/components/editor.tsx
ianstormtaylor: Sounds good to me.

packages/slate-react/src/components/editor.tsx
ryanmitts: That sounds good, I think it's important to be be able to stop the default behavior.

packages/slate-react/src/components/editor.tsx
```diff
@@ -0,0 +1,510 @@
...
+
+                  if (isDOMElement(relatedTarget)) {
+                    // COMPAT: The event should be ignored if the focus is moving from
+                    // the editor to inside a void node's spacer element.
+                    if (relatedTarget.hasAttribute('data-slate-spacer')) return
+
+                    // COMPAT: The event should be ignored if the focus is moving to a
+                    // non- editable section of an element that isn't a void node (eg.
+                    // a list item of the check list example).
+                    const node = editor.toSlateNode(relatedTarget)
```
ryanmitts: This seems to cause a problem when clicking from an editor into another focusable element on the page. The `toSlateNode` function throws an error if the input is not in the editor. Should it just return null in that case (I believe mainline does this now for `findNode`)? If not, then` hasDomNode` seems to be needed to be checked first. This can be seen in the search highlight example.

ianstormtaylor: I was trying to see how far we can get with throwing, since it leads to less weirdness I think. Checking if it has it first sounds good to me.

packages/slate/src/classes/commands/mark.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  if (Range.isRange(at)) {
+    if (!hanging) {
+      at = editor.unhangRange(at)
+    }
+
+    const rangeRef = editor.createRangeRef(at, { affinity: 'inward' })
+    const [start, end] = Range.edges(at)
+    editor.splitNodes({ at: end, match: 'text' })
```
ryanmitts: Do one/both of these `splitNodes` need `always: true`. This function does not seem to be able to apply a mark to a collapsed selection.

packages/slate/src/classes/commands/node.ts
```diff
@@ -0,0 +1,773 @@
...
+          }
+        }
+      } else {
+        // Merge adjacent text nodes that are empty or have matching marks.
+        if (prev != null && Text.isText(prev)) {
+          if (Text.matches(child, prev)) {
+            this.mergeNodes({ at: at.concat(n) })
+            n--
+            continue
+          } else if (prev.text === '') {
```
ryanmitts: Should these two `else if` conditions check to make sure one of the nodes does not have different marks than the other before merging? If I create a new empty text node and add a mark, then this merges it with the previous text node.

packages/slate/src/classes/commands/mark.ts
ianstormtaylor: When you say to a collapsed selection, what are you expecting for the output?

For the case of "future marks", where you hit <kbd>cmd+b</kbd> and then type some text in bold, I'm thinking having those being a `slate-react`-only concern that is stored as local state. That way we don't need operations. Similar to how `editor.isFocused()` is now a DOM-only concern instead of in core.

packages/slate/src/classes/commands/node.ts
ianstormtaylor: Good question, I think it should you're right.

ianstormtaylor: Although, I'm not sure extra empty text nodes should be allowed actually. So maybe it shouldn't.

ianstormtaylor:
@steida I was thinking about the functional approach some more. One example of something I'm unsure how to model with good UX without classes is the idea of "void" nodes. If you have a purely function command like `insertText(value, 'string')` how do you pass information about which nodes to treat as void nodes?

steida:
@ianstormtaylor Check Twitter DM, please.

alexandcote:
@steida Should those conversations stay public? Maybe in a slack channel/thread. Other people may also have idea 😃.

steida:
The idea is to expose side effects via React imperative hande: https://github.com/evolu-io/evolu/commit/8859bb3105c522f9cf064ca6f1703f6f0f1ef1a3 

Then, when some transformation needs ad-hoc non-deterministic computations that can cause side effects, it will just take what it needs and then use the pipe. I especially ask fp-ts people, whether it is an idiomatic functional approach, and it seems it is. Just function composition with IO. IO 'splits" configuration and calling, so it's referential transparent.

https://github.com/gcanti/fp-ts/issues/1005

steida:
@alexandcote No, just to answer my stupid question about Slate details. I gave up the understanding of previous Slate details since the architecture of Evolu emerged to be completely different.

ianstormtaylor:
So… in other words `isVoid` and `isInline` are calculated by querying the DOM?

packages/slate/src/classes/commands/mark.ts
ryanmitts: It's the future marks case I'm talking about like you describe (cmd+b at the end of a node to make future text bold). Whatever works I think is fine, in the future it should just make its way to the rich-text editor example which had that as part of it (to keep it functionally equivalent to the current version).

steida:
I don't know what void in Slate does exactly, but for block/inline detection based on computed styles, yes. The editor needs side effects like any other software, the point is to isolate them via IO. It's not easy to understand why, it took me some time, but now I finally understand. 

Inject behavior instead of injecting the state. As Mark Seemans explained it in one of his talks.

steida:
Maybe I am answering for something else, not sure, that's why I asked for Twitter DM or similar.

steida:
If you want to make Slate more functional, start with removing all nulls, undefineds, and throws. Replace them with Option / Either. Then leverage simple interfaces like Predicate, Endomorphism, Refinement. Then leverage Eq, Ord, pipe... It's not as magic as it looks like. It's just bunch of little helpers here and there. You can use it without understanding monoids, monads, functors, etc. Actually, I understood theory after real examples 😅

alexandcote:
@steida This is a nice discussion. I'm wondering how easy other contributors will learn/understand those new concepts. I can understand where you come from and why you propose that. I personally love learning new concepts like that and forcing my brain to think differently but I'm not 100% convinced this refactor should go that far. If we want to go that far why not using ReasonML or a language similar? 

I'm also unsure if the consumer will have to use `fp-ts`? Think about people/business which build a big infrastructure around this editor, I think we should try to make the migration from previous versions "easy". I have the feeling that if we choose to introduce this kind of new concept, the migration may be harder for those people.

What do you think?

steida:
> I'm wondering how easy other contributors will learn/understand those new concepts. 

I have the same concerns. Six months ago, functional programming was for me: pure functions, immutability, and some totally blurred magic ✨I call it the "glass ceiling". Every time I started trying to do it seriously, monads and functors jargon fucked me up. The reason why I finally succeeded was 'fp-ts'. Stable, battle-tested, the whole ecosystem, tree shakeable.

> If we want to go that far why not using ReasonML or a language similar?

I would love, but I can't. It's a too big change for me because I am doing JS and TS consulting.
The good news is, we don't have to. Pattern matching, sound types, and perfect type inference is nice, but I suppose it's not the core of the functional programming. Those type classes (just a JSON with functions and some shape via TS interface) is. As I see it, it helps a lot while perfectly supported by TypeScript. It's good enough step for me. It's incremental progress. There are only a few concepts revolving all the time. Pipe, chain, map, fold... etc. I am pretty sure it's relatively easy to understand when people see real examples. And it is a good investment instead of learning some custom homemade approach for library architecture from scratch all the time.

> I'm also unsure if the consumer will have to use fp-ts? Think about people/business who builds a big infrastructure around this editor, I think we should try to make the migration from previous versions "easy". I have the feeling that if we choose to introduce this kind of new concepts, the migration may be harder for those people.

My idea for Evolu was to expose helpers around functional primitives, but then I realized I would just both duplicate and cripple existing API. Sure we can have helpers here and there, but, what if someone needs, say to compare paths? Without fp-ts, they will just duplicate code and naming. In functional programming, we have nice helpers for the concept of the order no matter what is ordered. gt, lt, geq (greater or equal), etc. I realized there are a lot of little gems like absurd functions (for exhaustive checks on switch imitating pattern matching in TS), or equality, etc. and I don't see any reason why this should be hidden before developers. fp-ts is tree shakeable useful things. Don't fight with it, embrace it. That's how I see it.

ryanmitts:
> If you want to make Slate more functional, start with removing all nulls, undefineds, and throws. Replace them with Option / Either. Then leverage simple interfaces like Predicate, Endomorphism, Refinement. Then leverage Eq, Ord, pipe... It's not as magic as it looks like. It's just bunch of little helpers here and there. You can use it without understanding monoids, monads, functors, etc. Actually, I understood theory after real examples 😅

What problem is this solving for Slate?

ianstormtaylor:
Thanks all. I definitely do not want to have a discussion about the merits of functional programming here. One of Slate's principles is to use familiar JavaScript concepts and patterns to solve its problems. And that principle is much more important to us than functional purity.

I was mainly curious about how the domain-specific queries might be handled in a more functional approach. That's the sticking point that prevents me from being able to get to something like `Commands.insertText(value, 'string')`, because one way or another we need to pass in an encapsulation of that query logic. 

So far it seems like the most JavaScript-y approach is to use a class with methods that can be overridden to do that. If we're talking about the different between:

```js
editor.insertText('string')
```

And:

```js
insertText(io, value, 'string')
```

It doesn't seem like the second case is a clear win.

steida:
@ianstormtaylor Sure, I will never answer anything here anymore. You obviously don't like it. Sorry, I made you uncomfortable.

thesunny:
Hey @ianstormtaylor , what is the plan with respect to Android compatibility? It looks like you deleted the Android plugin and its support files.

clauderic:
>  I definitely do not want to have a discussion about the merits of functional programming here. One of Slate's principles is to use familiar JavaScript concepts and patterns to solve its problems. And that principle is much more important to us than functional purity.

I strongly agree with this, let's try to keep the discussion centered around the merits and drawbacks of the changes that are being proposed here, not around different programming styles.

@ianstormtaylor curious if you've given thought on what the migration path will look like for current users of Slate. My understanding is it will likely require quite a lot of work for users to migrate; are you planning on writing documentation around this to ease that process?

> Hey @ianstormtaylor , what is the plan with respect to Android compatibility? It looks like you deleted the Android plugin and its support files.

I'm also curious about this. What is your vision around this? Is this something you're planning on supporting out of the box? My expectation would be that Slate would support the latest version of most major browsers out of the box (Chrome (Edge), Safari, Firefox, iOS, Android), and perhaps make the rest possible via polyfills from the community (though not really certain what the latter looks like)

Finally, I'd also be curious to hear more about the migration from keys to path. Currently, many commands rely on passing a node's key to update it. How straightforward is it for consumers to migrate away from key-based commands to path-based commands? Will each node hold a reference to it's path?

alexandcote:
@ianstormtaylor can you give us more context why we need to pass queries here: 
```ts
insertText(queries, value, 'string')
``` 

What I like about this functional approach is we don't have to add commands/queries from plugins to the editor object. In our project, we found it hard to get strong typing with those custom commands/queries. 

AFAIK, invoking the command from the editor directly is not valid with typescript because plugins or dynamic and can be enabled/disabled at runtime. 

```ts
editor.yourCustomCommand(...)
```

Invoking your command like this is valid but it comes with another issue where the params of your custom command are not strongly typed. 
```ts
editor.command('yourCustomCommand', ...)
```

By exposing commands/queries in a "functional" way, we solve this issue but I'm not aware of all the implications this can have.

clauderic:
For consumers using TypeScript, the most impactful change to custom queries / commands would be to find an elegant solution to strongly type them. 

I think what @alexandcote is getting at is that the functional approach to queries / commands has the advantage of being strongly typed, but I'd be open to hearing other suggestions that achieve the same strongly typed queries/commands for consumers.

TheSpyder:
Looks like he's deleted his comments? That's a shame. I saw them all via email though.

I don't want to derail the PR again, I'm on slack and happy to have a detailed conversation, but I think it's perfectly fine to progressively migrate to FP. ReasonML can help in unexpected ways - the DOM wrapper/layer for slate I'm working on is written in it - but one step at a time.

> I was mainly curious about how the domain-specific queries might be handled in a more functional approach.

It doesn't have to be fancy; it could simply be `insertText(editor, value, 'string')` as a starting point. Separating the logic (insert text) from the state (editor) and using composition instead of classes and inheritance can help a lot with separation of concerns. Major changes like this are certainly a good time to evaluate a better API, but the readme does make it clear that until 1.0 breaking changes are expected. It can be improved later.

rikkit:
@ianstormtaylor  thanks for the reply!

> > When you write IE is not supported out of the box - will it be possible to support it for those who need to? Is this change polyfillable for older browsers (not just IE11, but Edge, iOS Safari, etc)
> 
> My hope is exactly that, that people will be able to polyfill for IE support if they need it. It's just that I won't be the one maintaining all of that code, it will require the community to work together or someone to champion it.

That's fair enough. I was wondering if you were aware of any polyfills that exist already (which could just be loaded before Slate), or if you think there'll need to be an IE11 support plugin. I think for my use cases, it'd actually be OK to drop support for IE11 (providing readonly mode works), but just trying to gauge the kind of work that would be needed.


--- 

On the TypeScript issue @alexandcote @clauderic ....

> AFAIK, invoking the command from the editor directly is not valid with typescript because plugins are dynamics and can be enabled/disabled at runtime.
> 
> editor.yourCustomCommand(...)

This is actually relatively easy to make strongly typed - you just need to add your own typing and use declaration merging:

```ts
// slate-custom.d.ts

import { Editor } from "slate-react";

declare module "slate-react" {
  export interface Editor {
    customCommand: (arg: string) => Editor;
  }
}
```

image

> Invoking your command like this is valid but it comes with another issue where the params of your custom command are not strongly typed.
>
> editor.command('yourCustomCommand', ...)

AFAIK this style is not typeable - you can't add override signatures with declaration merging. 

image

> I think what @alexandcote is getting at is that the functional approach to queries / commands has the advantage of being strongly typed, but I'd be open to hearing other suggestions that achieve the same strongly typed queries/commands for consumers.

As demonstrated above it's possible to strongly type a customised class based API using declaration merging. It is a small amount of setup and maintenance of the extra file - compared to a fully functional API where all the types would be inferrable - but it's not the worst thing in the world.

alexandcote:
>As demonstrated above it's possible to strongly type a customized class-based API using-declaration merging. It is a small amount of setup and maintenance of the extra file - compared to a fully functional API where all the types would be inferrable - but it's not the worst thing in the world.

💯 Agree, I still prefer inferrable types tho but you're right, we can define our own types. Thanks for the tip 👍

ianstormtaylor:
> @ianstormtaylor curious if you've given thought on what the migration path will look like for current users of Slate. My understanding is it will likely require quite a lot of work for users to migrate; are you planning on writing documentation around this to ease that process?

It'll be a lot of work, yup. I'd like to update the docs for the newest version, and include an overview of the changes on one page to help migrating. But I won't have the bandwidth to make a hyper-detailed changelog probably because so much has changed. But yeah, existing code will _function_ fairly similarly, but the _syntax_ for most everything has changed.

---

> @alexandcote: What I like about this "functional" approach is we don't have to add commands/queries from plugins to the editor object. In our project, we found it hard to get strong typing with those custom commands/queries.

> @clauderic: I think what @alexandcote is getting at is that the functional approach to queries / commands has the advantage of being strongly typed, but I'd be open to hearing other suggestions that achieve the same strongly typed queries/commands for consumers.

The current class-based design with "mixins" (instead of "plugins") was a way to solve this. If you check out the `./site/examples` folder on this branch you can see how it works. By using the `with*()` mixins you actually get to keep strong typings on your editor for all of the methods that each mixin defines. 

It's right on the edge of what's possible with TypeScript I think. In the future if TypeScript is able to model `compose` functions properly it might get even better too.

---

> @TheSpyder: It doesn't have to be fancy; it could simply be `insertText(editor, value, 'string')` as a starting point. Separating the logic (insert text) from the state (editor) and using composition instead of classes and inheritance can help a lot with separation of concerns. Major changes like this are certainly a good time to evaluate a better API, but the readme does make it clear that until 1.0 breaking changes are expected. It can be improved later.

I agree that could be something useful and that we can revisit it later. I think the biggest issue that will need to be solved in the architecture is "overridability". Once that is solved it would be easier to transition to a functional approach. 

Honestly, if the `beforeinput` spec was more nailed down and it had React support out of the box, we could probably even push the overriding all into userland with an optional `onBeforeInput=` handler. Because DOM events have had to solve this via `preventDefault()` and `stopPropagation()` already. 

I'm just a little hesitant to do it right now while React has such bad support for it and while browsers are in flux on implementing `beforeinput`. We kind of still need the control to be able to fix some things up.

---

> @rikkit: That's fair enough. I was wondering if you were aware of any polyfills that exist already (which could just be loaded before Slate), or if you think there'll need to be an IE11 support plugin. I think for my use cases, it'd actually be OK to drop support for IE11 (providing readonly mode works), but just trying to gauge the kind of work that would be needed.

Nope, I'm not aware of any. I think it would definitely need to be a mixin as well as polyfills, since it will require attaching to specific events and customizing the behaviors I think.

ianstormtaylor:
> @clauderic I'm also curious about this. What is your vision around this? Is this something you're planning on supporting out of the box? My expectation would be that Slate would support the latest version of most major browsers out of the box (Chrome, Chromium-based Edge, Safari, Firefox, iOS, Android), and perhaps make the rest possible via polyfills from the community (though not really certain what the latter looks like in practice)

Yeah, that's the goal I think. Latest version of each browser for now. Firefox and Android are the current two unknowns. Edge will be supported as soon as they launch the Chromium-based one in January. I'm probably going to push ahead with releasing this `next` branch before that though and people can always stay on the older `0.47.x` versions if they need broader support.

---

> @clauderic Finally, I'd also be curious to hear more about the migration from keys to path. Currently, many commands rely on passing a node's key to update it. How straightforward is it for consumers to migrate away from key-based commands to path-based commands? Will each node hold a reference to it's path?

Fairly straightforward in many cases. You can call `editor.findPath(node)` (might be renamed in a bit when I clean up some inconsistencies) to get the node's path in a rendering context. Other contexts you'll likely already have the path.

TheSpyder:
> In the future if TypeScript is able to model `compose` functions properly it might get even better too.

This is what a `compose` function looks like right now, when TypeScript added tuple types it went a long way towards retaining the type information.
https://github.com/tinymce/tinymce/blob/master/modules/katamari/src/main/ts/ephox/katamari/api/Fun.ts#L7

ianstormtaylor:
Ah cool! Well maybe once I merge it someone with more TypeScript skills than me can figure out a way to get us a `compose` function for mixins too!

package.json
```diff
@@ -1,114 +1,88 @@
...
+    "fix": "yarn fix:prettier",
+    "fix:eslint": "yarn lint:eslint --fix",
+    "fix:prettier": "yarn lint:prettier --write",
+    "lint": "yarn lint:eslint && yarn lint:prettier",
+    "lint:eslint": "eslint '**/*.{js,jsx,ts,tsx}'",
+    "lint:prettier": "prettier --list-different '**/*.{css,md,js,jsx,json,ts,tsx}'",
+    "open": "open http://localhost:3000",
+    "release": "yarn build:production && yarn test && yarn lint && lerna publish",
+    "serve": "cd ./site && next",
+    "start": "npm-run-all --parallel --print-label watch server",
```
ryanmitts: `start` is broken after `server` was changed to `serve`

package.json
```diff
@@ -1,114 +1,88 @@
...
+    "test": "cross-env BABEL_ENV=test FORBID_WARNINGS=true mocha --require ./config/babel/register ./packages/*/test/index.js",
+    "test:debug": "cross-env BABEL_ENV=test FORBID_WARNINGS=true mocha debug --require ./config/babel/register ./packages/*/test/index.js",
+    "watch": "yarn build --watch"
```
ryanmitts: I don't think `--watch` is a valid argument to `build` anymore now that the next build happens there.

packages/slate-react/src/components/editor.tsx
```diff
@@ -0,0 +1,518 @@
...
+      }
+    }
+  }, [])
+
+  // Whenever the editor updates, make sure the DOM selection state is in sync.
+  useLayoutEffect(() => {
+    const { selection } = value
+    const domSelection = window.getSelection()
+
+    if (state.isComposing || !domSelection || !editor.isFocused()) {
```
ryanmitts: I believe you also need to check if `domSelect.type` is [`None`](https://developer.mozilla.org/en-US/docs/Web/API/Selection/type), otherwise, there is no range to select below.

packages/slate/src/interfaces/text.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * `Text` objects represent the nodes that contain the actual text content of a
+ * Slate document along with any formatting marks. They are always leaf nodes in
+ * the document tree as they cannot contain any children.
+ */
+
+interface Text {
+  text: string
+  marks: Mark[]
```
TheSpyder: I'm a little concerned by the performance implications of swapping this from `Immutable.Set` to an Array. In particular the cost of comparing adjacent text nodes to see if the marks are equal is now `O(n^2)` or worse, as seen in the `matches` implementation below:
```
for (const m of existing) {
  if (!Mark.exists(m, marks)) {
    return false
  }
}

for (const m of marks) {
  if (!Mark.exists(m, existing)) {
    return false
  }
}
```

I'm not sure what representation `Immutable.Set` uses but I'm pretty sure `set1.equals(set2)` is not quadratic.

ianstormtaylor: @TheSpyder That's a good concern. I have a feeling we might be able to make it not matter, because I think in 90% of cases either:

a. The text nodes will both have no marks.
b. One text node will have marks and the other won't.

So if we do a check for mismatching lengths, or both zero lengths we can short-circuit early in the common cases and not invoke the iterators at all?

TheSpyder: Certainly those two changes will help a lot! Since we can't easily sort the Marks array, the other way to do this without checking both directions is clone the first array then remove every element found in the second. Further reading:

https://stackoverflow.com/q/245509/7546
https://cs.stackexchange.com/q/69064

ianstormtaylor:
Okay, well I think this PR broke GitHub. Even though it doesn't look like it, it did successfully get merged. Just didn't get marked as merged somehow by GitHub's UI.

But it's merged! 🎉

mxstbr:
@ianstormtaylor this is awesome, I am so happy to see these changes! :fire: :fire: :fire: :fire:

lapico:
Absolutely awesome ! Thanks for this great piece of software.

TepyThai:
I just noticed in the updated documents that we can use Hooks now which is really awesome. Thanks for all the hard work 😄 👏

gvpmahesh:
Why would anyone merge these many changes in a single PR? it's impossible for others to review and difficult to track what's going on

wmertens:
@gvpmahesh it's an almost complete rewrite

enzoferey:
What an amazing effort. Congrats to everybody that worked on this 🔥

julienR2:
Woua !.. Great job on those changes !!
We'll be hell of a migration, but I can't wait to feel the performance kick, the improved Android support, and so much more !

Thanks again for this amazing work 👍

seanlaff:
Awesome! Everything sounds great- and stoked to see it in typescript 👍

sherwinyu:
Will there be a migration guide (or any other documentation on how to upgrade) coming? 

If not, I can try to contribute to one , if that would be helpful for others (will probably take detailed notes when I attempt this upgrade).

Also, this is amazing. I'm so excited for so many of these changes! Thank you!

sherwinyu:
Oh, just kidding there's a guide here! https://docs.slatejs.org/concepts/xx-migrating (in case other people read this comment and are looking).

wmertens:
Note for posterity: this was merged in https://github.com/ianstormtaylor/slate/commit/4ff6972096953d5774b7c82571d702e304c3f55f but GitHub somehow missed that

packages/slate/src/interfaces/range.ts
```diff
@@ -0,0 +1,269 @@
...
+  },
+
+  /**
+   * Check if a range includes a path, a point or part of another range.
+   */
+
+  includes(range: Range, target: Path | Point | Range): boolean {
+    if (Range.isRange(target)) {
+      if (
+        Range.includes(range, target.anchor) ||
```
ulion, line 83: Shouldn't this be '&&'? Or the next isBefore and isAfter should reverse parameters' order?

TheSpyder: I don't think so. First it checks whether either anchor or focus is inside `range` to cover overlap cases. If that isn't the case, it checks whether `target` is wholly inside `range`.

ulion: then, it does not cover the range is wholly inside the target case. which shouldn't be also the "part of another range" case?

TheSpyder: Hmm you're right it probably needs both `range includes target` and `target includes range` checks. There aren't any tests for this either, only [the path and point scenarios](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/test/interfaces/Range/includes). Would you mind logging a new ticket about it?

PR #3095, opened 2019-11-08
Fix delete on Android with a deep document

gracicot:

Fixing a bug

#### What's the new behavior?

Deleting correctly when operating on a document with a deep structure.

Consider this document:

```js
  document: {
    object: "document",
    data: {},
    nodes: [
      {
        object: "block",
        type: "depth1",
        data: {},
        nodes: [
          {
            object: "block",
            type: "depth2",
            data: {},
            nodes: [
              {
                object: "block",
                type: "depth3",
                data: {},
                nodes: [
                  {
                    object: "text",
                    text: "T",
                    marks: []
                  },
                  {
                    object: "text",
                    text: " ",
                    marks: [
                      {
                        object: "mark",
                        type: "style1",
                        data: {}
                      }
                    ]
                  }
                ]
              }
            ]
          }
        ]
      }
    ]
  }
```
[Video: Crash when deleting a parent node with swipe delete](https://i.imgur.com/YIe116R.mp4)
[Video: Crash when pressing backspace](https://i.imgur.com/oLEApVD.mp4)

[Video: Editor with the patches](https://i.imgur.com/E5H1ooo.mp4)

When pressing a single backspace, the node with the style gets deleted. This is a single mutation that is removing the node. The composition manager catches that and match on single mutation that removes nodes.

When pressing backspace on the last character, the node get deleted, the browser also adds a `<br />` node. This screws the matching, since it's two mutations instead of one.

The solution for that was to filter out `<br />` additions.

There was another problem with the swipe delete. Swipe delete removed the parent span containing the text node.

The problem with that is that `reconcileDOMNode` expect to alway find the span node where the focused text node is, but none is there, instead a nearby block is found:

```js
function reconcileDOMNode(editor, domNode) {
  // finds a block, there is no span anymore!
  const domElement = domNode.parentElement.closest('[data-key]') 
  const node = editor.findNode(domElement)
  // ...
}
``` 

My solution was to take the block and call `reconcileNode` on each on its text nodes. I'm unsure if this is necessary or if only calling `reconcileNode` on the first one is enough.

Reviewers: @ianstormtaylor @thesunny

### DIFF
--- a/packages/slate-react/src/plugins/android/composition-manager.js
+++ b/packages/slate-react/src/plugins/android/composition-manager.js
@@ -355,11 +355,22 @@ function CompositionManager(editor) {
 
     const firstMutation = mutations[0]
 
+    // Here we want to filter BR nodes out.
+    // When the content editable choose to remove a node instead of
+    // a 'characterData' mutation, it is usually the only mutation, unless
+    // it's the last node. When it's the last node, the browser also adds a BR.
+    // We want to filter it out so we match the mutation correctly and call `removeNode`
+    const filterBr = () => {
+      return mutations.filter(m => {
+        return !(m.addedNodes[0] && m.addedNodes[0].nodeName === 'BR')
+      })
+    }
+
     if (firstMutation.type === 'characterData') {
       resolveDOMNode(firstMutation.target.parentNode)
     } else if (firstMutation.type === 'childList') {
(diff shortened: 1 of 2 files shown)


ianstormtaylor:

Thank you for understanding.

PR #3113, opened 2019-11-13
Fix long pressed accents on Chrome+macOS

Calyhre:

Fixing a bug

#### What's the new behavior?

On macOS, long press triggers an IME to insert an accent. Browsers that implements `beforeinput` event uses the `insertReplacementText` and a proper range. Chrome is instead selecting the character to replace, **right when the input is triggered**.

| **Before** | **After** |
| -- | -- |
|![][before]|![][after]|

[before]: https://user-images.githubusercontent.com/3692274/68779241-9b318000-0634-11ea-9c41-be9be33972e4.gif
[after]: https://user-images.githubusercontent.com/3692274/68779245-9cfb4380-0634-11ea-896f-83272a0a0763.gif

If we detect macOS && Chrome, the we proceed to check if the DOM selection has been updated before the input event and a detection of a selection change. The check is very naive but enough for this particular case. This can't happen when previous editor selection was expanded I guess.

I had to expose through a query the composition state of the before plugin, this is used to discard completely this fix in case of a user using another IME.

This issue was also fixed by #3041 but the fix is more trustworthy in my opinion, as we are not relying on any previous input.

I couldn't figured out how to test this with the current toolchain.

Fixes: #982 
Reviewers: @ianstormtaylor @knubie

### DIFF
--- a/packages/slate-react/src/plugins/dom/after.js
+++ b/packages/slate-react/src/plugins/dom/after.js
@@ -3,7 +3,13 @@ import Debug from 'debug'
 import Hotkeys from 'slate-hotkeys'
 import Plain from 'slate-plain-serializer'
 import getWindow from 'get-window'
-import { IS_IOS, IS_IE, IS_EDGE } from 'slate-dev-environment'
+import {
+  IS_IOS,
+  IS_IE,
+  IS_EDGE,
+  IS_CHROME,
+  IS_MAC,
+} from 'slate-dev-environment'
 
 import cloneFragment from '../../utils/clone-fragment'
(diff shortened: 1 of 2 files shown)


Calyhre:
⚠️ This does not work with composition 🙇‍♂️ Trying to find a better fix

ianstormtaylor:

jens-lilt:
If I select the character in the menu with the arrow keys I can still reproduce this behavior in Slate 0.77.2
Typing the number under the letter works as expected

BitPhinix:
Applying them using arrow keys indeed still does seem to cause issues on the latest version. If you raise an issue I'll take a look at it next week

Issue #2457, opened 2018-11-27
Can not input chinese in safari


boyrest:
The issue is caused in Safari.
When I choose Chinese, I can't enter.
For example, I enter "好的(haode)"
  
![issue1](https://user-images.githubusercontent.com/11408703/49056561-2b659200-f237-11e8-9a11-b9f041675f77.gif)

miterfrants:
OMG! I got same problem.

hangzz:
![Uploading image.png…]()
@zjfjiayou

Yixian15:
I've been working on these Chinese input issues for a few days. There're actually two issues when it comes to typing Chinese in safari browser (Mac OS 10.12.6, Safari 10.1.2).

The first issue: editor doesn't insert Chinese character after word selection, i.e. it doesn't insert anything when composition is over. For browsers like Chrome without support for input events level 2, editor would simply call insertText() in onBeforeInput function as an after plugin. But for safari browser, editor decides what to do next depending on event.inputType, it seems the case 'insertFromComposition' is missed at this place. Therefore @hangzz 's solution would do the trick, which makes editor insert composition result.

The second issue: one key stroke inserts two characters when composing Chinese at end of block, which is the issue mentioned above. Seems it has something to do with onSelect after plugin, commenting out the second last line in that function

`editor.select(selection) // slate-react/src/plugins/after, onSelect function`

makes word composing work, but it invalidates select action of course. I'm stuck at this place to make both select and word composition work. If anyone has any idea to fix this issue, it would be greatly appreciated.

Chinese translation: 

Safari 浏览器输入中文两个问题：
- 选词之后无法插入文字，解决办法，在 onBeforeInput after plugin 里面补充 insertFromComposition 的 case，就是楼上 @hangzz 的方法。
- 在段落最后一次按键输入两个字母，after plugin onSelect 去掉最后的 editor.select(selection) 之后可以正常输入，但是那样会丢失选择的位置，还在研究解决方案。

ChalkleyZhou:
Same problem.
MacOS 10.14.1
Safari 12.0.1

Already insert the code 

> case "insertFromComposition ":

but still got the problem

sunshinelee:
Insert the code

> case  "insertFromComposition ":

but the problem cann't be solved completely while some Chinese shows and others don't

hecor:
Still got the same problem, can't input chinese on ios safari.
Hope to get help.

hudk114:
[...]

PR #3126, opened 2019-11-21
Fix/ime input bugs

uihoh0:
fixing some bugs with IME

#### What's the new behavior?
1, IME input failed in Safari. E.G. option + `+ a should input an à, but Safari input nothing.
2, The cursor is placed at the end of the composition string.
Can find that in  https://github.com/ianstormtaylor/slate/issues/2457

1, cherry-picks some code from this PR: https://github.com/ianstormtaylor/slate/pull/2454/, to fixed the cursor problem.
2, About the IME typing always failed, it's because Safari will do `insertCompositionText ` while input `a =>`deleteCompositionText ` to delete the `a before the à is converted =>`insertFromComposition` to input à, `deleteCompositionText` will delete the whole IME spell string, so, adding the input type `insertFromComposition` into the afterPlugin can fix it.
3, IME selection range: When typing with IME, the native inputEvent.getTargetRanges() will always return [](seems like a bug in Chrome, Safari is working well). And `updateSelection` in `onSelect` event will cause range-conflict in this situation, so just ignore the `updateSelection`.

Fixes: https://github.com/ianstormtaylor/slate/issues/2457
Reviewers: @ianstormtaylor


ianstormtaylor:

PR #3133, opened 2019-11-28
Docs fixes for 0.50.0

mattmazzola:
Fixing minor bugs in docs. Typos, grammar, suggestions, etc

I was taking a look at the new 0.50.0 release and doing the walkthroughs from the docs and thought I would record what I found. 
I know this whole thing was work in progress so maybe the bugs are already fixed on another branch but at least they can get some visibility here. Contributors can take whichever pieces of this PR are important. Don't have to merge the whole thing.

I also saw errors with typescript that I will ~open separate issues~ add comments here for.

### DIFF
--- a/docs/walkthroughs/01-installing-slate.md
+++ b/docs/walkthroughs/01-installing-slate.md
@@ -49,7 +49,7 @@ Of course we haven't rendered anything, so you won't see any changes.
 
 Next up is to render a `<Slate>` context provider.
 
-The provider component leeps track of your Slate editor, it's plugins, it's default value, and any changes that occur. It **must** be rendered above any `<Editable>` components. But it can also provide the editor state to other components like toolbars, menus, etc. using the `useSlate` hook.
+The provider component keeps track of your Slate editor, it's plugins, it's default value, and any changes that occur. It **must** be rendered above any `<Editable>` components. But it can also provide the editor state to other components like toolbars, menus, etc. using the `useSlate` hook.
 
 ```jsx
 const App = () => {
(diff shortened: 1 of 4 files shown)


mattmazzola:
About the `renderMark` callback changes.

It seems the docs were leaving off the array declaring dependencies.
I saw this error: **Expected 2 arguments, but got 1.**

image

Also, the docs show the code as this:
```js
const renderMark = useCallback((props: CustomMarkProps) => {
  switch (props.mark.type) {
    case 'bold': {
      return <BoldMark {...props} />
    }
  }
}, [])
```
Which doesn't allow to be assigned to the `Editor`
You see:
image

Based on the types the function must return a `JSX.Element` so the `switch` statement needs a default.

I got around it by adding a `<span>` such as:
```typescript
const renderMark = React.useCallback((props: CustomMarkProps) => {
  switch (props.mark.type) {
    case 'bold': {
      return <BoldMark {...props} />
    }
    default:
      return <span {...props} />
  }
}, [])
```

Also, this might be related. In the very first tutorial it shows code like this:

```typescript
<Slate editor={editor} defaultValue={defaultValue}>
  <Editable />
</Slate>
```
However this also gives an error saying `onDOMBeforeInput` must be defined.
image

I worked around by changing to this:
```typescript
<Editable
  onDOMBeforeInput={() => { }}
/>
```

Since one of the big things about 0.50.0 was it's claim to support TypeScript it would be nice to see the tutorials written with full types.
I would say again even though I'm picking out types bugs I would say these are minor overall and expected given the scope of changes.  Hope you take this as me trying to help rather thing giving negative feedback immediately.

ianstormtaylor:
This is amazing, thank you @mattmazzola! 

I agree, I think we should probably make the examples use TypeScript. Or at least a few of them. 

But I wasn’t sure I wanted the walkthrough to include it because it would scare off people who don’t understand types at all, because when you don’t know them they are pretty hard to parse.

PR #3138, opened 2019-11-28
Editable: onDOMBeforeInput should be optional

zbeyens:
Typescript error:

```
Property 'onDOMBeforeInput' is missing in type '{}' but required in type ...
```

Fixing a bug

#### What's the new behavior?

The following works when using typescript
```
<Slate editor={editor} defaultValue={defaultValue}>
  <Editable />
</Slate>
```

Make onDOMBeforeInput optional


ianstormtaylor:
Thank you @zbeyens, good catch.

PR #3142, opened 2019-11-28
Add Slite.com as example of usage

Calyhre:
I'd like to propose Slite.com as an example of usage. Our editor is extensively using Slate and collaboration with it.


ianstormtaylor:
Beautiful, thanks @Calyhre!

PR #3153, opened 2019-11-28
Remove controlled value state from docs

martonlanga:

Fixing misleading docs


ianstormtaylor:
Thanks @martonlanga!

PR #3156, opened 2019-11-28
Fix variable name in docs

martonlanga:

Fix


ianstormtaylor:
Thanks @martonlanga!

PR #3158, opened 2019-11-28
improve prop typings for Editable

Nathan-Fenner:

This change improves `<Editable />` prop typings by removing a few uses of `any`.

This change is not strictly backwards-compatible, but any violations are *probably* bugs.

#### What's the new behavior?

- `React.HTMLAttributes<HTMLDivElement>` replaces `{ [key: string]: any }`
- `React.CSSProperties` replaces `Record<string, any>`

Additional type annotations are needed throughout due to poor default interface. Omitting the type parameter from each event causes it to default to (e.g.) `FocusEvent<Element>` which is slightly less-specific than the best available type `FocusEvent<HTMLDivElement>`.

This slight change in types also requires the `isEventHandled` to need slightly different types to accept them all.

Fixes: _
Reviewers: _

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -36,18 +36,19 @@ import {
  * Editable.
  */
 
-export const Editable = (props: {
-  decorate?: (entry: NodeEntry) => Range[]
-  onDOMBeforeInput?: (event: Event) => void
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
@Nathan-Fenner this is absolutely amazing. Thank you!

PR #3169, opened 2019-11-29
Update Adding Event Handlers walkthrough

sherwinyu:
The previous example no longer uses `onChange`, so don't reference it in this walkthrough.

No

#### What's the new behavior?
Documentation change

No


ianstormtaylor:
Great catch, thanks @sherwinyu!

Issue #3174, opened 2019-11-30
change `NodeMatch/MarkMatch` to allow arrays for "or" logic


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now we have the `NodeMatch` and `MarkMatch` types which allow for writing quick matching logic against nodes or marks. For example you could do:

```js
match: { type: 'code' }
match: 'block'
match: ([node, path]) => node.type.endsWith('list')
...
```

There are great for simple matches.

#### What's the expected behavior?

I think we should just extend the type to accept an array of the matchers too, for "OR" logic. For example:

```js
match: [{ type: 'code' }, { type: 'quote' }]
```

Would match either a code block or a quote block.

This is already supported in `slate-schema`, so it would just allow people to use the same notation everywhere. For example in #3173 the `Editor.removeMarks` transform would take a `MarkMatch` which can be an array of these matchers. So removing bold or italics would be:

```js
Editor.removeMarks(editor, [{ type: 'bold' }, { type: 'italic' }])
```

PR #3190, opened 2019-12-01
fix: type

zbeyens:
forced-layout example is not using an array for `match`

bug

#### What's the new behavior?

allow non-array match

Fixes: #
Reviewers: @


ianstormtaylor:
Thanks @zbeyens! Good catch.

We'll improve this further with #3174 but this is a good interim for now.

PR #3192, opened 2019-12-02
Fix missing callback dependencies in slate-react <Editable />

calebdelnay:

Bug

When doing `<Editable onBlur={handleBlur} />`, if the identity of `handleBlur` changes because its dependencies have changed, the `Editable` component calls the original function because the component's internal `onBlur` is memonized without specifying any dependencies.

#### What's the new behavior?

When using the `<Editable />` React component, if the identity of event handler functions change, then the internal handlers for `onBlur` will be updated to call the new functions.

Specify `attributes.onXYZ` as a dependency for the relevant `useCallback()` callbacks.

#### Notes

This looks like a much larger change than it actually is due to lint formatting. Functional change is https://github.com/ianstormtaylor/slate/commit/1b2a89f2f396e52584d580de65359b58ba7c224b

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -498,7 +498,7 @@ export const Editable = (
           }
 
           IS_FOCUSED.delete(editor)
-        }, [])}
+        }, [attributes.onBlur])}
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Wow good catch, thanks @calebdelnay!

PR #3202, opened 2019-12-03
onCompositionEnd not need to insert text in firfox

Jokcy:

fixing bug

Dec-03-2019 13-54-05

#### What's the new behavior?

Firefox composition input work fine

Slate use React `onBeforeUpdate` event to insert text, compositionend also trigger `onBeforeUpdate` so no need to insert text in compositionend


ianstormtaylor:
Thanks @Jokcy!

Issue #3162, opened 2019-11-29
decorate not work well with ime input method


Jokcy:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

![Nov-27-2019 16-47-50](https://user-images.githubusercontent.com/2088642/69848292-a4695080-12b4-11ea-9573-1dd5b5f6b087.gif)

The problem which I found i 4x version still exist in 50 version. This time I drived into source code to find the reason. 

The main reason is when I input the cursor locate at some `decoration` for example `*123*`, after I finish IME input, first of all the text will be placed in the `span` where `*123*` in. Then slate update the text tree, rerender, but based on text tree in slate, this `decoration` won't need to be update, because it's range never change. In slate the text is "*123*这", after decorate, `*123*` is still there. So it never rerender, but in the real dom, "这" has been added.

Jokcy:
There are two ways to solve this problem:
1. Prevent browser from insert text to the decorated node, which I guess is very difficult to do.
2. Let the decorated node rerender.

Jokcy:
Example here https://codesandbox.io/s/trusting-gagarin-ij22y

ianstormtaylor:
What browser and operating system is this?

Jokcy:
@ianstormtaylor macos and chrome, all latest.

Jokcy:
Any idea about this problem? I tyied to solve this problem, it seems no other way except render the block every time since we can't prevent from editing and there is no better way to detect the decoration dom node changed.

Jokcy:
@ianstormtaylor How about add this code to `TextString` component?
```js
useEffect(() => {
    // console.log(ref.current)
    if (ref.current && ref.current.innerText !== text) {
      ref.current.innerText = text
    }
  })
```
This seems fix this problem, except it moves the cursor forward to the start of the `TextString`

Jokcy:
@ianstormtaylor I want to mark a PR about this, but when I run test, I got a lot of error:
```
TypeError: children.filter(...).flat is not a function
      at jsx (../src/hyperscript.ts:99:5)
      at Object.<anonymous> (packages/slate/test/transforms/wrapNodes/split-inline/inline.js:10:7)
      at Module._compile (internal/modules/cjs/loader.js:688:30)
      at Module._compile (node_modules/pirates/lib/index.js:99:24)
      at Module._extensions..js (internal/modules/cjs/loader.js:699:10)
[...]

PR #3205, opened 2019-12-03
fix ime input on decoration node error

Jokcy:

fix this https://github.com/ianstormtaylor/slate/issues/3162
#### What's the new behavior?

reset leaf node when slate text not equal dom text

do reset when `TextString` rerender

Fixes: #3162


Jokcy:
This PR just want to talk about whether this solution may work fine.

Jokcy:
@ianstormtaylor How about this PR now? Can I do anything to take it to the next step?

ianstormtaylor:
Can you explain what changes it’s making and why it’s doing that? I don’t understand it.

Jokcy:
@ianstormtaylor Sure!
The main point is, React will diff vdom <-> realdom to judge whether the real dom should be update. When in composition events, the contenteditable is not controlled by React. So wehn a compositionend event fired, the text has already been inserted into the real dom.

Most of case will work fine in slate, if we have this dom tree:
```js
<span data-slate-node="text">
  <span data-slate-leaf="true">
    <span data-slate-string="true">
      some text
    </span>
  </span>
</span>
```

when composition event I input *这* first of all, the real dom text change to `some text这`, then compositionend fired, slate tree update and rerender the **string** node, every thing work fine.

But in decroation case, for example:
```html
<span data-slate-node="text">
  <span data-slate-leaf="true">
    <span data-slate-string="true">
      some 
    </span>
  </span>
  <strong data-slate-leaf="true">
    <span data-slate-string="true">
      **text**
    </span>
  </strong>
</span>
```
when I put cursor at the end of `**text**`, the cursor actually in `<span data-slate-string="true">**text**</span>` dom node. when compositionend fired, the `这` added into this node. Then slate update, the text in slate change to `some **text**这`, then do decroate. We will find that the decroation node of `**text**` not change in slate tree, so the decroation node will not update the real dom node, the dom tree looks like this:
```html
<span data-slate-node="text">
  <span data-slate-leaf="true">
    <span data-slate-string="true">
      some 
    </span>
  </span>
  <strong data-slate-leaf="true">
    <span data-slate-string="true">
      **text**这
    </span>
  </strong>
  <span data-slate-leaf="true">
    <span data-slate-string="true">
      这
    </span>
  </span>
</span>
```

And this PR just want to fix this, just like React controlled input

ianstormtaylor:
Thanks @Jokcy that explanation is very helpful.

Does it only apply in Chrome?

And for the fix, why is it using a selection operation? Why not using something like React's `key=` property to force a re-render without reconciling the DOM?

Jokcy:
@ianstormtaylor Not only in chrome. 

I use this solution because I can't find a way to notice whether a decroation should be updated. As I said this PR is more about talking then just fix it.

ianstormtaylor:
@Jokcy which other browsers do you see this bug in?

I wonder if the entire `Text` component could be re-rendered in this case somehow, since we don't necessarily know which decoration it was in? I'm not totally sure.

Jokcy:
@ianstormtaylor Firefox. Not test other browser yet, but I guess it is a common issue not much relative to browser. The main issue is that we can't `preventDefault` the composition event from directly insert text into dom.

ianstormtaylor:
@Jokcy can you test the other browsers? I'm not sure it's common off the top of my head. Specifically Safari and Chrome.

Jokcy:
@ianstormtaylor OK, I will do it. You mean just test the bug exists or this PR can fix this bug?

Jokcy:
@ianstormtaylor  I tested in chrome, firefox and safari, result here:

chrome:
slate-bug-chrome

firefox:
slate-bug-firefox

safari is different. 
slate-bug-safari

Jokcy:
And reset selection in `TextString` won't work any more, I guess it has some issue with the `controlled slate`
```
editor.apply({
        type: 'set_selection',
        properties: editor.selection,
        newProperties: editor.selection,
      })
```

Jokcy:
@ianstormtaylor I update this PR to make it work with the controlled `slate` component. The main different from `master` is **sync selection from slate to realdom now is async**. This fix firefox, chrome and safira issues.

But I'm not quite sure if it will cause other issues.

Jokcy:
@ianstormtaylor Any progress on this PR?

ianstormtaylor:
@Jokcy the question is actually have _you_ made any progress on this PR. I'm not going to merge something that seems untested and that isn't written up well to be sure that it's the right way of doing something.

Jokcy:
@ianstormtaylor I totally understand that. But I do need some advice to make sure my PR will work fine. I did not code well because I just want to mention how we can fix this problem and whether it will cause some other issues.

Now, I passed all the unit test and I tested locally of the three browsers, the just work fine. What should I do to make this PR pass？

Since all of my change of code in `slate-react`, I did not see any unit test for it now, how to make sure it not break by my code.

packages/slate-react/src/components/editable.tsx
```diff
@@ -153,50 +153,56 @@ export const Editable = (props: EditableProps) => {
 
   // Whenever the editor updates, make sure the DOM selection state is in sync.
   useIsomorphicLayoutEffect(() => {
-    const { selection } = editor
-    const domSelection = window.getSelection()
-
-    if (state.isComposing || !domSelection || !ReactEditor.isFocused(editor)) {
-      return
-    }
+    setTimeout(() => {
```
ianstormtaylor: What's the need for this `setTimeout`?

packages/slate-react/src/components/string.tsx
```diff
@@ -58,8 +58,18 @@ const String = (props: {
...
   const { text, isTrailing = false } = props
+  const editor: any = useEditor()
+
+  const ref: React.RefObject<HTMLSpanElement> = useRef() as any
+
+  useEffect(() => {
+    if (ref.current && ref.current.innerText !== text) {
+      ref.current.innerText = text
+    }
+  })
```
ianstormtaylor: What's the goal of this `useEffect` statement here? Can you add a clear comment about with `COMPAT: ...` explaining this for future readers.

packages/slate-react/src/components/editable.tsx
Jokcy: As the gif shows, in safari, when ever I use composition input, the selection will reset to the beginning of the `String` compnent (as the `span` tag). It seems in safari do reset in `layoutEffect` is not ok. 

Change to `useEffect` instend of `useLayoutEffect` will work fine too.

Also I added **reset text** to the `String` component, it will cause selection reset to the beginning too.

Jokcy: If the *delay of reset selection* ok, I will change to `useEffect` instead of `setTimeout`

packages/slate-react/src/components/string.tsx
Jokcy: Sure, I will commit it soon

packages/slate-react/src/components/editable.tsx
ianstormtaylor: `useEffect` sounds like a much better choice than `setTimeout` to me.

Jokcy:
@ianstormtaylor I updated this PR, change to `useLayoutEffect` to reset `String` component, so I the default selection reset is enough in this case.

 And I remove the `setTimout` also. I found no matter to use `useEffect` or `setTimout`, it will cause cursor twinkle which is bad experience.

So I guess safari issue should be another case

packages/slate-react/src/components/string.tsx
Jokcy: Yeah, I want to find a more *React* way to force the `String` node to be updated at first.But then I found the main reason of this issue is because **React can't detect this kind change of dom**. If someone find a more *React* way to solve this issue, that will be better.

The reason why `useLayoutEffect` just fix this problem is **selection is reset at `Editable` component** which is the root of all `String` components. The direction of React run effects is from leaf to root, so the text reset will run first. And `useEffect` is async which run on nextTick, so it won't work with the selection reset in `layoutEffect`.

`reconciliation` just do something with virtual dom, it never touch the real dom

packages/slate-react/src/components/string.tsx
```diff
@@ -58,8 +59,26 @@ const String = (props: {
...
+  // but when slate state update, after decroate
+  // we will find the `decroation` node not need to update
+  // this cause the difference of *vdom <-> realdom*
+  // so we need to reset it if needed
+  // https://github.com/ianstormtaylor/slate/pull/3205
+  useIsomorphicLayoutEffect(() => {
+    if (ref.current && ref.current.innerText !== text) {
+      ref.current.innerText = text
+    }
+  })
```
EleanorMao: 你好，我觉得你好像没有考虑到isTrailing为true的情况。因为此时会多加一个\n，所以innerText对比必然是错误的。
另外如果这段string里面包含\n的话，使用innerText会导致加进去很多<br />。

packages/slate-react/src/components/string.tsx
```diff
@@ -58,8 +59,26 @@ const String = (props: {
 
 const TextString = (props: { text: string; isTrailing?: boolean }) => {
   const { text, isTrailing = false } = props
+
+  const ref: React.RefObject<HTMLSpanElement> = useRef() as any
```
EleanorMao: try `const ref = useRef<HTMLSpanElement>(null)`

packages/slate-react/src/components/string.tsx
Jokcy: isTrailing是什么？不是很明白你的意思，能给个例子么

packages/slate-react/src/components/string.tsx
```diff
@@ -58,8 +59,26 @@ const String = (props: {
...
+    if (ref.current && ref.current.innerText !== text) {
+      ref.current.innerText = text
+    }
+  })
+
   return (
-    <span data-slate-string>
+    <span data-slate-string ref={ref}>
       {text}
       {isTrailing ? '\n' : null}
```
EleanorMao: 这边如果isTrailing是true的话，会加一个\n

packages/slate-react/src/components/string.tsx
EleanorMao: [看这个](https://github.com/ianstormtaylor/slate/pull/3205/files#diff-f7099a3a2afcbc3f7da9adb71b71754dL64)

Jokcy: 哦～～～这是几个大版本升级过程中加的，merge的时候没注意，待会改一下。

Jokcy: 你对于这样的改动有什么建议么？或者你觉得可能会出现问题的地方

EleanorMao:
```javascript
useIsomorphicLayoutEffect(() => {
    if (ref.current) {
      let htmlText = ref.current.innerText
      if (isTrailing) {
        htmlText = htmlText.slice(0, -1)
      }
      if (htmlText !== text) {
        ref.current.innerHTML = text + (isTrailing ? '\n' : '')
      }
    }
  })
```
我这样子写了😂
用innerText如果有`\n`的话会加进去`<br />`

Jokcy:
I recently tried to fix this problem again, and then I updated the demo of *Markdown Preview*:
```tsx
const Leaf = ({ attributes, children, leaf }) => {
  const spanRef = useRef<HTMLSpanElement>()

  return (
    <span
      ref={spanRef}
      key={spanRef.current ? spanRef.current.innerText : leaf.text}
      {...attributes}
    // other code ......
  )
}
```
it worked.

So the main principl is that I force leaf node to rerender if I found it's innerText is not equal to `leaf.text`. So @ianstormtaylor  is this a good way to fix this problem?

ianstormtaylor:
@Jokcy that sounds like a much simpler solution, thanks!

Jokcy:
@ianstormtaylor so should we update the default Leaf component like this?

ianstormtaylor:
Ah then maybe it's not good if it makes rendering performance way worse.

dylans (changes requested):
@Jokcy are you still interested in pursuing this PR? I'm happy to work with you to review/land it.

dylans:

Issue #3152, opened 2019-11-28
Add super simple autocomplete to mentions example


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now the `mentions` example uses a `window.prompt` for simplicity.

#### What's the expected behavior?

I think we could use the `<datalist>` approach from [this tweet](https://twitter.com/TejasKumar_/status/1194689434656088065) to end up with a proper autocomplete-like behavior but while still keeping it very simple. 

![image](https://user-images.githubusercontent.com/311752/69832957-06677e80-11ff-11ea-85a2-ab44d2e2cd10.png)

It would: 

- Borrow the hovering menu from `hovering-menu` to position the autocompleting `<input>` in the right spot when `@` is pressed, and insert on <kbd>enter</kbd>.

- Remove the mention button, and only use `@` autocompletion instead.

timothyarmes:
Could this example please add a tiny bit of very useful complexity...

Rather than reusing the hover code, fo which we already have an example, it would be really useful if the dropdown could be pinned to the trigger character. It's not obvious how to do this, so the example add value.

We used to do this with an annotation.

ianstormtaylor:
Sure! It would be very similar, just without updating on selection changes.

timothyarmes:
Would that follow in scroll?

objectlegal:
To my understanding, omitting to updating on changes makes it a bit weird in cases of line breaks and possibly even scrolling. E.g. we need the dropdown to follow even when we start @typing far to the right and it line-breaks in the middle of the suggestion. Therefore, its a bit more overhead and logic than anticipated, but it should be working calculatingand creating a custom dom range for the trigger and get the position from there. 

See: https://codesandbox.io/s/slate-reproductions-cdw83

Mewle

PR #3206, opened 2019-12-03
Simplified mentions example with dropdown (#3152)

timothyarmes:
_improvement_

#### What's the new behavior?

An improved mentions example showing:

* How to react to the current input (to detect the trigger)
* How to pin a pulldown menu to a dom element (in this case the trigger)
* How to override the keyboard arrow keys when the dropdown is open

Of other interest:

* The Slate component simply has to include` <MentionsPortal />` and call `handleSuggestionsKeys` on the `onKeyDown` event.
* `MentionsPortal` is a specific usecase for responding to _mention_ triggers. Another portal could for example, respond to emoticons.  It uses a generic hook (`useSuggestions`) to parse out the current trigger search value and to fetch the dropdown positional information, and displays the results using `SuggestionsList`.

### DIFF
--- a/site/examples/mentions.js
+++ b/site/examples/mentions.js
@@ -1,131 +1,273 @@
-import React, { useMemo } from 'react'
-import { Editor, createEditor } from 'slate'
+import ReactDOM from 'react-dom'
+import React, { useMemo, useRef, useEffect, useState } from 'react'
+import { Slate, Editable, withReact, useSlate, ReactEditor } from 'slate-react'
+import { Editor, createEditor, Node, Range, Path } from 'slate'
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @timothyarmes thank you for this, it's awesome! 

I like how you implemented the portal, I've just opened https://github.com/ianstormtaylor/slate/pull/3214 that borrows your implementation but cleans it up a bit with some of the `Editor.*` helpers that can be used, and simplifies the rendering/styling a bit to make it easier to follow hopefully. I really appreciate you taking the time to showcase this!

Issue #3181, opened 2019-12-01
Using currying in withSchema


timothyarmes:
_feature_

Currently the `withSchema` function takes the schema as the last argument: `withSchema(editor, schema)`. This is unfortunate since it doesn't allow functional composition:

```
const editor = compose(
  withInlineStyles,
  withBlockStyles,
  withAlignment,
  withLinks,
  withImages,
  withVideos,
  withHistory,
  withReact,
)(createEditor());

// This isn't nice ->
this.editor = withSchema(editor, schema);
```

If we allowed withSchema to be curried, we could do this, which is much nicer and fits better with the very functional approach taken throughout the codebase:

```
const editor = compose(
  withInlineStyles,
  withBlockStyles,
  withAlignment,
  withLinks,
  withImages,
  withVideos,
  withHistory,
  withSchema(schema),
  withReact,
)(createEditor());
```

ianstormtaylor:
Sounds great to me, thanks @timothyarmes!

PR #3207, opened 2019-12-03
Change withSchema to allow currying (#3181)

timothyarmes:
I've reversed the order of the parameters so the rules come first.
The function is curried so that it can be called in two ways:

withSchema(rules, editor)
withSchema(rules)(editor)

Addresses #3181


ianstormtaylor:
Hey @timothyarmes I've been thinking about this, and since it's a breaking change anyways I think I want to go with exporting a `defineSchema` function instead which returns the plugin. That way there's only a single way to do things, instead of using currying. I've mocked this up in: https://github.com/ianstormtaylor/slate/pull/3252

PR #3210, opened 2019-12-03
Fix InsertImageButton in image example

mms-uret:
Bug

#### What's the new behavior?

The insert image button on the image example works again.


pchr-srf:
Thanks, lost 2 hours because of this

ianstormtaylor:
Thanks @mms-uret!

PR #3214, opened 2019-12-03
upgrade mentions example to use a portal with suggestions

ianstormtaylor:

Improvement.

#### What's the new behavior?

Adds an auto-complete popover to the mentions example.

It uses a `<Portal />` that waits to match a specific condition of the user having their cursor after an `@` for mentioning. And while the popover is active, hotkeys are intercepted to change the highlighted item and commit or cancel a suggestion.

Fixes: https://github.com/ianstormtaylor/slate/issues/3152
Fixes: https://github.com/ianstormtaylor/slate/pull/3206

### DIFF
--- a/site/examples/mentions.js
+++ b/site/examples/mentions.js
@@ -1,47 +1,139 @@
-import React, { useMemo } from 'react'
-import { Editor, createEditor } from 'slate'
+import React, { useMemo, useCallback, useRef, useEffect, useState } from 'react'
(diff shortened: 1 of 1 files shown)


timothyarmes:
Hi,

I like your simplified example. There's a 'bug' however - it gets triggered when typing in an email address. You need to check that there's either a space or a start of line before the trigger.

timothyarmes:
Also note that the app crashes when there's no text, since Editor.before returns undefined.

Issue #3209, opened 2019-12-03
Api for setting value from an external change


bmingles:
#### Do you want to request a _feature_ or report a _bug_?

Feature request

#### What's the current behavior?

There doesn't appear to be a way to set the editor value externally e.g. from a prop change or other event that would warrant updating the value from outside the editor.

-->

Slate: 0.XX.X
Browser: Chrome / Safari / Firefox / Edge
OS: Mac / Windows / Linux / iOS / Android

#### What's the expected behavior?

We should have a way to set the value externally. 

One proposal that has been made is to be able to configure the editor as a controlled component, but there are likely tradeoffs with this. This issue is not intended to champion that particular proposal but rather to discuss what the "right" api might be. e.g. there could be options to call a method on the `editor` or something more imperative.

NOTE: The idea has been discussed in #3200 of possibly using changes to `defaultValue` as a way to set the value externally, but the current behavior seems consistent with other uncontrolled components in React.

timothyarmes:
The standard way to handle this for uncontrolled components is to update the key.

PeterKottas:
@timothyarmes that's an awkward solution at best. You could use this reasoning for everything and we would have to swap key for every single input everywhere. 

Or to put it in other words, yes you are right that this is the case with uncontrolled components, the question remains if it's a good idea to have slate as an uncontrolled component. I personally prefer to have a value prop and be saved from swapping the key all the time.

timothyarmes:
Yes. I asked Ian on Slack why he chose to go with an uncontrolled component. He said:

> I changed it to being an uncontrolled because to me it seemed like the architecture was asking for it for a few reasons…
>
> (a) We’ve always had an onChange handler. And whenever it was called, the editor.value property was already the new value, so re-sending it as a prop was only a little trick to get React to re-render, but the editor itself already had the new value. So on every render pass we’d do editor.value = props.value just to be sure it was the latest
>
> (b) When moving the selection to editor.selection (and paving the way for other non-core tracked properties), we no longer have a single value prop that can be passed in. You’d need to pass in children and selection. And if you use any plugins that extend the editor with other tracked properties, you’d need to pass those in as well
>
> I’m down to try to work around those issues though, if someone wants to take a crack at it
> We could potentially not special case any prop on <Slate> such that it passed children={children} selection={selection} through just fine
> But if you also gave it annotations={annotations} it would attach that to the editor too
[...]

PR #3216, opened 2019-12-03
Change `<Slate>` to a controlled component

ianstormtaylor:

Improvement.

#### What's the new behavior?

It changes `<Slate>` to be a controlled component. 

It takes `value=` and `selection=` props and ensures they are updated on the editor whenever they change, such that people can use their own `useState` to store them. People must now store state for **both** of these.

Currently it is:

```jsx
<Slate defaultValue={initialValue}>
  // ...
</Slate>
```

With this PR it is:

```jsx
const [value, setValue] = useState(initialValue)
const [selection, setSelection] = useState(null)

<Slate 
  value={value} 
  selection={selection} 
  onChange={(value, selection) => {
    setValue(value)
    setSelection(selection)
  }}
>
  // ...
</Slate>
```

For an example in context, check out the [changes to the `rich-text` example](https://github.com/ianstormtaylor/slate/pull/3216/files#diff-a0f7e62dc9c4fe521e1ec5245c0a402c) on this branch.

Fixes: #3209


PeterKottas:
Looks great, well done man!

packages/slate-react/src/components/slate.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { useMemo } from 'react'
+import { Editor, Node, Operation, Range } from 'slate'
```
CameronAckermanSEL: moving this to a different module was a good decision i think, it took a little bit to find it just diving through github.

packages/slate-react/src/components/slate.tsx
```diff
@@ -0,0 +1,41 @@
...
+  onChange: (children: Node[], selection: Range | null) => void
+  [key: string]: any
+}) => {
+  const { editor, children, onChange, value, selection, ...rest } = props
+  const context: [Editor] = useMemo(() => {
+    editor.children = value
+    editor.selection = selection
+    return [editor]
+  }, [value, selection, ...Object.values(rest)])
+
```
CameronAckermanSEL: i wonder if there will be some gotchas around memoizing around the extra props. for the most part this is probably what a caller is going to need but this seems like a subtle trap for users who are not paying attention

CameronAckermanSEL:
This looks pretty good to me. I'm happy that this still works well with the context based distribution work flow.

packages/slate-react/src/components/slate.tsx
ianstormtaylor: Yeah, although by including them it's less restrictive (not more) so I think it's alright? Otherwise implementing your own custom editor properties becomes impossible I think?

focux:
Looks good to me, also, thank you for all your work @ianstormtaylor.

benjycui:
How can I get operations for collaborative editing after this MR?

wmertens:
I would prefer that the entire state gets controlled as `state` and the current value is a part of that (much like it was before).

* plugins can add their own editor state without changing the public API. E.g. a history plugin can add a `history` key to the state.
* only one state needs storing by the user

rikkit:
Not really been following this - is it possible to use Slate as an uncontrolled component after this? Or is this fully switching it over? I guess you can always have a wrapper component to make it uncontrolled.

Jokcy:
@ianstormtaylor https://docs.slatejs.org/concepts/xx-migrating#plugins 

So plugin deps on operation won't work anymore? What's the new way to build a plugin?

PeterKottas:
I think your question doesn't belong here @Jokcy, try slack, this PR for making slate into a controlled component again.

@ianstormtaylor , just out of curiosity, since this is a breaking change, how long do you think it will take to get this out? Not sure what the policy about stuff like this is with slate. Thanks!

ianstormtaylor:
> How can I get operations for collaborative editing after this MR?

You can access them from `editor.operations`. (Actually you can access the `children` and `selection` that way too, but I've changed the `onChange` handler to be more of a convenience for making things React-ish instead of "the only way to get things".)

ianstormtaylor:
> Not really been following this - is it possible to use Slate as an uncontrolled component after this? Or is this fully switching it over? I guess you can always have a wrapper component to make it uncontrolled.

This fully switches to a controlled component. I think using React's own helpers you can make it "uncontrolled" kind of. But I'd rather have only a single way to do it.

maksimsemenov:
I think docs still refer to the old API (an un-controlled). Not sure if it is a known issue, though

PR #3217, opened 2019-12-03
fix: ts

zbeyens:

Bug

`JSX.Element` does not accept null, undefined or boolean, so conditional rendering is not allowed.
`React.ReactNode` seems to be the default for `children`

Fixes: #
Reviewers: @


ianstormtaylor:
Thanks @zbeyens!

PR #3218, opened 2019-12-04
fix: support upper case search

zbeyens:

bug

#### What's the new behavior?

Before, only lower case search was working.

Now, support upper case search as well.

Fixes: #
Reviewers: @


ianstormtaylor:
Thanks @zbeyens!

Issue #3182, opened 2019-12-01
Forced layout bug


timothyarmes:
_bug_?

The forced layout code inserts a paragraph if the last block isn't of the right type:

```
case 'child_min_invalid': {
  const block = { type, children: [{ text: '', marks: [] }] }
  Editor.insertNodes(editor, block, { at: path.concat(index) })
  break
}
```

Rather than inserting the paragraph at the top level, it's trying to insert it as a child of the last element, which is incorrect and actually fails completely if the last child is a void.

IMO it should be:

```
case 'child_min_invalid': {
  const block = { type, children: [{ text: '', marks: [] }] }
  Editor.insertNodes(editor, block, { at: [index + 1] })
  break
}
```

At least, that works perfectly in my code to ensure that my last block isn't void.

Issue #3165, opened 2019-11-29
slate-schema validation does not respect children's `min` prop


flavordaaave:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

![](http://g.recordit.co/q6dXlmyTWt.gif)

See [https://www.slatejs.org/examples/forced-layout](https://www.slatejs.org/examples/forced-layout)

Slate: 0.50.0
Browser: Safari & Chrome
OS: Mac

#### What's the expected behavior?

When deleting the `title` block a `child_min_invalid` error should be thrown.

dmarkow:
It seems that some of the min/max counting isn't dealing with the `match` part of the rule properly.

Here's an example child_min_invalid test that should pass but doesn't. Since there is no `<element b>` the validation should fail.

Given the following test:

```js
export const schema = [
  {
    for: 'node',
    match: { a: true },
    validate: {
      children: [
        { match: { b: true }, min: 1 },
        { match: { c: true }, min: 1 },
      ],
    },
  },
]

export const input = (
  <editor>
    <element a>
      <element c>one</element> // test passes if I make this <element b> instead
    </element>
  </editor>
)

export const output = <editor />
```
- Validation should fail and the test should pass since there is no `<element b>`, but it doesn't. The test fails, leaving `<element c>` in the output.
- If you remove the `match` part of each rule (so it's just two `min` rules) the test passes.
- If you change `<element c>` to `<element b>` the passes.

PR #3219, opened 2019-12-04
Fix child_min_invalid checks

dmarkow:

Fixing a bug

#### What's the new behavior?

In some situations, a `child_min_invalid` wasn't being thrown when necessary, often involving `match` arguments in the `children` rules. This was causing the Forced Layout example to not work correctly (if you highlighted everything and deleted it, you were left with a single paragraph instead of a title and a paragraph).

- When checking child nodes, the count is now only incremented when the child node matches the `match` property (or the default empty match of `{}`)
- The Forced Layout code was expecting an `index` property in the returned error, but the `checkers.ts` file never adds an index. Instead, it looks like it sends back the full expected path of the missing node. So I updated the forced layout example to just use the full returned path instead.

Fixes: #3182 
Fixes: #3165

### DIFF
--- a/packages/slate-schema/src/checkers.ts
+++ b/packages/slate-schema/src/checkers.ts
@@ -130,7 +130,6 @@ export const checkAncestor = (
   let g = 0
 
   while (true) {
-    count++
     const group = groups[g] as ChildValidation | undefined
     const child = parent.children[index] as Descendant | undefined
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Thank you @dmarkow, this is awesome!

Issue #3211, opened 2019-12-03
refreshing any example resets to default rich text one


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

For some reason refreshing the examples doesn't properly render the actual example you were on (despite the URL being right) and changes instead to the `rich-text` example. Not sure why this is, some Next.js configuration thing...

PR #3220, opened 2019-12-04
Load with correct starting example

dmarkow:

Fixing a bug

#### What's the new behavior?

Reloading the page on one of the examples now stays on that example.

Seems like next.js only populates `router.query` after it pushes a route itself. There may be a cleaner solution, but parsing the starting example out of the `asPath` parameter seems to work fine for both initial load and navigating between pages and gets it fixed for now.

Fixes: #3211


ianstormtaylor:
Thank you so much @dmarkow!

PR #3232, opened 2019-12-04
Fix check list example

mms-uret:

Bug.

#### What's the new behavior?

The check list example uses `ReactEditor.findPath(editor, ...)` instead of `editor.findPath()`. This allows to uncheck and check the checkboxes again.


ianstormtaylor:
Thanks @mms-uret!

Issue #3231, opened 2019-12-04
Change marks array on a text node to properties of the text node


thesunny:
#### Do you want to request a _feature_ or report a _bug_?

Change to text node data structure

#### What's the current behavior?

In 0.5 marks are stored as a property on a text node like this

```js
const CURRENT_WAY = [
  { text: "This is ", marks: [] },
  { text: "bold", marks: [{ type: "bold" }] },
  { text: " and ", marks: [] },
  { text: "italic", marks: [{ type: "italic" }] },
  { text: "text", marks: [] },
]
```

#### What's the expected behavior?

Turn marks into properties on the text node

```js
const PROPOSED_WAY = [
  { text: "This is " },
  { text: "bold", bold: true },
  { text: " and " },
  { text: "italic", italic: true },
  { text: "text" },
]
```

This is easy for beginner’s to manipulate and removes cases like accidentally adding two bold marks. We just set property `{bold: true}`. In cases where multiple marks are desired (say, we  want to mark some text as for specific users) it would be encoded into the model by using an array  `{forUsers: ['john', 'jane']}`  which makes intent clear.

Note: Discussion on this issue in Slack on the `#contributing` channel

ianstormtaylor:
Thank you for opening this @thesunny! This is a really genius idea!

I think there are a few tradeoffs (I'm probably forgetting some), but the amount it would simplify the data model is so great that it would take some pretty big drawbacks to make it not worth it. And it plays into the new interface-based data model really well.

I've tried implementing it in #3235 and everything works pretty well!

One of the drawbacks is that rendering what used to be "marks" as multiple elements is a little more involved, because you have to keep track of what the "top" one is. Although it could be argued to just use a single element for the `Leaf` in the first place so simplify it.

Another thing that's not a drawback, but just a tradeoff, is that "unsetting nodes" becomes more important. And the trouble is that `undefined` isn't serializable to JSON, which leads to issues with unsetting. So I think we need to start special-casing `null || undefined` as a custom property to mean `delete` instead of set. But I think that's a reasonable thing to do, essentially treating `null` and `undefined` as the same. And this was probably going to be necessary already for custom properties on non-text nodes.

It's crazy how the pieces just keep getting simpler. Once you remove Immutable.js, and annotations, and all the other things that _seemed_ to "need" marks, marks themselves aren't even really necessary either! Such a good suggestion.

There's probably a few kinks to work out still on that branch, but hoping to merge it soon before people get too far in adopting the new data model.

PR #3235, opened 2019-12-04
remove marks, in favor of text properties

ianstormtaylor:

Debt.

#### What's the new behavior?

The `Mark` interface is removed, in favor of storing text-level formatting on the `Text` nodes themselves as custom properties.

Instead of having:

```js
{
  text: 'A line of text.',
  marks: [{ type: 'bold' }],
}
```

You have:

```js
{
  text: 'A line of text.',
  bold: true,
}
```

This is now possible because of the pure-JSON, interface-based data model!

Fixes: https://github.com/ianstormtaylor/slate/issues/3231


plbabin:
I really like that new direction with marks! 

Is it possible with this update to apply a mark directly on a text node when executing `Editor.insertText()` ?

ianstormtaylor:
@plbabin not directly with the `Editor.insertText` transform. You could achieve it being calling `Editor.setNodes` after inserting. Or you might achieve it by calling `Editor.insertNodes` instead with an already combined text-with-marks node.

TheSpyder:
I just started implementing transforms on the three `mark` operations and then realised in the changelog they were deleted last week 😂 

Does the `set_node` operation really cover all three of the previous mark cases? I am happy I have less operations to transform but I haven't looked at how `set_node` is triggered in much detail yet and I'm not quite sure what the equivalent operations will look like now.

Issue #3145, opened 2019-11-28
Editor broken(including examples on slatejs.org)


jessekelly881:
Currently all of the examples on slatejs.org are broken and when trying to use slate I get  "Error: Cannot resolve a DOM point from Slate point: {"path":[0,0],"offset":18}"

DamareYoh:
Hi @jessekelly881 , thanks for filing the issue. It's hard to diagnose your issue with the lack of information you provided. All of the examples are currently working for me. 

Please fill out the information missing on the template, at least with some information about the browser your using.

jessekelly881:
Hey. Apologies. The browser I am currently using is google chrome "Version 73.0.3664.3 (Official Build) dev (64-bit)" I update my browser and see if that fixes things.

jessekelly881:
Working well with Google Chrome version 78. Sorry again for the bother. Although, a little more back comparability seems needed. Chrome version 73 is only a few months old and slate is completely unusable on it.

DamareYoh:
No bother at all, thanks for the follow up :)

ianstormtaylor:
This likely because we rely on `beforeinput` events, maybe 73 hadn't had those yet. But either way, I think the market share of 73 is very low. (This is only a problem for the latest rewritten version of Slate, if you need older browser support use `0.47`.)

saltire:
I'm seeing this issue on some of the examples. If I try to make a selection that includes a checkbox in the Checklists example, or the video in the Embeds example, OP's error message shows up in the console. This happens in both Firefox 70 and Chrome 78.

Also, clicking a checkbox gives me `i.findPath is not a function`, though that may or may not be a separate issue.

cshen4:
I also have the above mentioned problem, i see the error `editor.findPath is not a function` when i run the examples repository and click on the checkbox in the Checklists example.

focux:
I'm also having this issue. If you go to the mention example in slatejs.org and select a mention it crashes.

gotoprototype:
I have this issue in the _images_, _Embeds_ and _Mentiones_ with all browsers I tested it with:

- Chromium 78.0
- Firefox 70.0
- Firefox 71.0
- Firefox Nightly 72.0

Couldn't reproduce on the _Checklists_ example.
---
However I found a makeshift solution:
**Give the `Element` nodes child an non empty string as text:**
`children: [{ text: 'NOTEMPTY', marks: [] }]`
even an blank space is enough.

[...]

PR #3244, opened 2019-12-05
Fix mention and images examples

focux:
Fixing a bug where Slate was crashing in mentions and images examples when clicking on a mention or image respectively.

#### What's the new behavior?

Those examples don't crash anymore.

Add content to the text of the elements.

Fixes: #3145
Reviewers: @ianstormtaylor

### DIFF
--- a/site/examples/images.js
+++ b/site/examples/images.js
@@ -72,7 +72,7 @@ const withImages = editor => {
 
       case 'insert_image': {
         const { url } = command
-        const text = { text: '', marks: [] }
+        const text = { text: 'NOTEMPTY', marks: [] }
         const image = { type: 'image', url, children: [text] }
         Editor.insertNodes(editor, image)
         break
@@ -159,7 +159,7 @@ const initialValue = [
     url: 'https://source.unsplash.com/kFrdX5IeQzI',
     children: [
       {
-        text: '',
+        text: 'NOTEMPTY',
         marks: [],
       },
     ],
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
This sounds like a bug with the selection determining logic. Since void nodes you be allowed to have empty text nodes I’d rather not fix it this way.

focux:
Oh I see, I thought that was the intended behavior, then I close this PR.

gotoprototype:
@ianstormtaylor 

I looked at the code and think that the problem is somewhere in `ReactEditor.toDomPoint()` (or below), while determining the true length (`trueLength`), here are the returned values:

```javascript
const { length } = domNode.textContent //  1
const attr = text.getAttribute('data-slate-length') //  "0"
const trueLength = attr == null ? length : parseInt(attr, 10) // 0
const end = start + trueLength // 0 

if (point.offset <= end) { //  false (1 <= 0)
```
and thus doesn't create a `domPoint` object.

Changing the ternary operators condition to `(attr == null || attr === '0')`:
```javascript
const trueLength = (attr == null || attr === '0') ? length : parseInt(attr, 10)
```
prevents it, but I don't know if that's actually the bug or just another symptom of it and I also don't know what other side effects this change brings.

For the time being I will simply sticking with creating a non-zero child element.
---
PS: Even though the child was created with an empty text (`children: [{ text: '', marks: [] }]`), the `domNode.textContent` is `"\ufeff"` - a ZERO WIDTH NO-BREAK SPACE.

ianstormtaylor:
I believe it may be due to the void nodes in Slate having `0` length text. But in the DOM we render a non-zero-length space character in their place, so it's probably trying to get a point at offset `1` which doesn't actually exist. Probably needs to adjust void nodes to always be offset `0`.

Issue #3255, opened 2019-12-05
Deleting across multiple text nodes merges parent element with previous element


dmarkow:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

With the cursor in a text node, hit backspace until you hit the previous text node (in the gif below, clicking just past the bolded text, then backspacing into the bolded text, or clicking in the bolded text and backspacing to the text before it). This causes the entire element node to get merged with the previous element node. 

![screencast 2019-12-05 16-01-08](https://user-images.githubusercontent.com/8009/70278193-70d96b00-1779-11ea-9016-459b2be7c155.gif)

Same thing happens if you highlight across multiple text nodes and delete:

![screencast 2019-12-05 16-12-50](https://user-images.githubusercontent.com/8009/70278599-34f2d580-177a-11ea-8ca0-bfd3b0d09102.gif)

Slate: 0.52.0
Browser: Chrome
OS: Mac

#### What's the expected behavior?

The element nodes shouldn't be merged.

PR #3258, opened 2019-12-06
Improved merging of blocks when deleting text

dmarkow:

Fixing a bug

#### What's the new behavior?

Deleting across inlines/text nodes within a single block no longer merges the parent block with the previous block.

Instead of using `isBlockAncestor`, I hoisted the block matching from the `startVoid`/`endVoid` checks and compared the paths of the starting/ending blocks to see if the selection was across multiple blocks.

I'm not 100% confident this is the right approach (not familiar enough with Slate's internals) but all tests are passing, along with the new test I added. Also, the Rich Text example now functions as expected when backspacing into the bolded text (it no longer merges the paragraph with the one above).

Fixes: #3255

### DIFF
--- a/packages/slate/src/interfaces/editor/transforms/text.ts
+++ b/packages/slate/src/interfaces/editor/transforms/text.ts
@@ -66,26 +66,31 @@ export const TextTransforms = {
       }
 
       let [start, end] = Range.edges(at)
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
Brilliant, thank you @dmarkow!

Issue #3151, opened 2019-11-28
Normalize `insert_text` on edges of inlines/marks


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Improvement.

#### What's the current behavior?

Right now we let the DOM determine where the selection ends up, and when an `insert_text` command is executed we insert wherever the selection is. But because of https://github.com/ianstormtaylor/slate/issues/3148 this can vary across browsers which is not intuitive for people.

#### What's the expected behavior?

I think we should have Slate's core behaviors canonicalized, so that it will:

- Always insert outside of an inline when it's at the edge.
- Always inside before a mark when it's at the start.
- Always inside inside a mark when it's at the end.

People can then choose to override these behaviors in their own plugins if they'd like to have "inline continuation", but at least in core it will be consistent.

beorn:
I know of #3148 — but would it be possible to give Slate's selection position be sticky and take precedence where possible, so that it's possible to use arrow keys to step through all the positions?  I.e., end of inline, start of next node as two different positions we can both move the selection to, and we can insertText into?

PR #3260, opened 2019-12-06
Insert outside inline when selection is at the end

dmarkow:

Fixing a bug

#### What's the new behavior?

screencast 2019-12-06 07-46-47

When inserting text at the end of an inline (e.g. a link) using a selection as the insertion point, the text will be added outside the inline. Passing an explicit point/path to `Editor.insertText` will retain the old behavior of inserting inside the inline.

Input:
```
<editor>
  <block>
    one
    <inline>
      two
      <cursor />
    </inline>
    three
  </block>
</editor>
```

Output from typing "four" or using `Editor.insertText(editor, "four")` (with no explicit `at` point/path)
```
<editor>
  <block>
    one
    <inline>
      two
    </inline>
    four
    <cursor />
    three
  </block>
</editor>
```

Output from calling `Editor.insertText(editor, "four", { at: { path: [0, 1, 0], offset: 3 } })`:
```
<editor>
  <block>
    one
    <inline>
      twofour
    </inline>
    three
  </block>
</editor>
```

I updated the `insertText` function to check if the cursor is at the end of an inline, and if so, it moves the selection outside of the inline before adding the text.

I have no idea if this is the right approach, so any feedback is greatly appreciated.

See #3151  - this addresses the inline portion of that issue.


ianstormtaylor:
Hey @dmarkow thank you for taking the time to investigate this!

I think we actually want to define this in the `insert_text` command, instead of changing the behavior of `Editor.insertText`. This way people can override `insert_text` and change the behavior (eg. for "sticky inlines" which has been request by a decent amount of people). And that way the `Editor.insertText` stays as predictable as possible for when using it in other situations.

dmarkow:
Just made another pass on this. I moved the logic to `insert_text`, and updated it to only apply when the range is collapsed. Dealing with the expanded range is tricky since `Editor.insertText` handles some stuff with voids/deletion, and it seems like the issue wasn't affecting expanded ranges in the same way anyway.

As you mentioned, overriding `insert_text` to not do the inline edge check would be very easy now (it would go back to being a single `Editor.insertText` command)

I did catch another bug while working on this. 

screencast 2019-12-06 20-26-02

If you highlight across the edge of an inline and type, the first letter you type becomes highlighted/selected, then the second letter you type replaces it. So if you quickly type `one` after selection across the edge, the `o` will be highlighted and then overwritten by `n`, with resulting text of `ne`. It looks like this was because the anchor/focus point of the selection were on difference edges of the inline. I updated it to set the selection anchor and focus to `at`.

Let me know how this looks. If it's good to merge, I can clean up the commit history on it first.

ianstormtaylor:
Looks great! Thanks for the detailed explanation too, that made it very easy to review. No need to cleanup the commits either because I just use GitHub's "squash and merge" strategy so it all goes into a single commit.

Thank you @dmarkow!

Issue #3239, opened 2019-12-05
Cannot resolve a DOM point from Slate point


joesonw:
#### Do you want to request a _feature_ or report a _bug_?

bug
#### What's the current behavior?

steps:
1. goto https://www.slatejs.org/examples/images
2. drop a image from computer to the editor
3. click empty area in editor

`Cannot resolve a DOM point from Slate point: {"path":[1,0],"offset":1}`

Slate: 0.50.X (official example)
Browser: Chrome
OS: Mac

#### What's the expected behavior?
Normal

dmarkow:
You don't even need to add another image. Just left click once on the existing image to highlight, then right click on it. Same error.

dmarkow:
Same issue with https://www.slatejs.org/examples/embeds -- try selecting some text above and continuing the selection into the embed.

![ScreenFlow](https://user-images.githubusercontent.com/8009/70244583-4ae0a600-173a-11ea-8eb7-20cc46693478.gif)

focux:
Duplicate of #3145

PR #3261, opened 2019-12-06
Force a zero offset when inside a void node

dmarkow:

Fixing a bug

#### What's the new behavior?

Selecting a void node no longer crashes.

`ReactEditor.toDOMPoint` forces an offset of 0 if the point is inside a void node.

I have no idea how to add tests for this, but can include some with a little guidance. 

Fixes: #3239


dmarkow:
This PR is a different approach to what #3244 was trying to solve.

ianstormtaylor:
Thank you @dmarkow!

lether-jdd:
this problem occurs on version 0.80.0, so the code has merged ?

Issue #3254, opened 2019-12-05
v0.52.0 React warning: useLayoutEffect does nothing on the server,


eddyw:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
I'm using NextJS which does SSR and React warns about usage of `useLayoutEffect` on server.
```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://fb.me/react-uselayouteffect-ssr for common fixes.
```

Slate: 0.52.0
Browser: Chrome / Safari / Firefox / Edge
OS: Mac / Windows / Linux / iOS / Android

#### What's the expected behavior?
It should default to `useEffect` if run on server, anyways neither `useEffect` nor `useLayoutEffect` is run on SSR.

Maybe?
```javascript
const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffect : useEffect
```

PR #3262, opened 2019-12-06
implement isomorphic layout effect

eddyw:

Fixing a bug

#### What's the new behavior?

No warning is displayed on SSR.

Suppresses warnings on SSR when using `useLayoutEffect` by falling back to `useEffect` when `window` is `undefined`.

This same approach is used in `react-redux` and `react-beautiful-dnd`.

Neither `useEffect` nor `useLayoutEffect` is run on server. Falling back to `useEffect` on server helps to keep consistency with the number and order of hooks within a component as well as preventing the warning from showing.

Fixes: https://github.com/ianstormtaylor/slate/issues/3254

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -1,10 +1,4 @@
-import React, {
-  useLayoutEffect,
-  useEffect,
-  useRef,
-  useMemo,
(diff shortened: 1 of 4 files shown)


ianstormtaylor:
Thanks @eddyw!

Issue #3188, opened 2019-12-01
Specific HTML can crash editor when pasted


scottfr:
#### What's the current behavior?

1. Run https://github.com/ianstormtaylor/slate/blob/master/site/examples/paste-html.js
2. Focus the editor
3. Run the following in the console:

```
var pasteData = new DataTransfer();
pasteData.setData('text/html', `<meta charset='utf-8'><p style="margin:0px !important;padding:0px !important;"><u>abc</u></p>`);
var e = new InputEvent('beforeinput', { inputType: 'insertFromPaste',
dataTransfer: pasteData,
          bubbles: true,
          cancelable: true,
          composed: true,
          composing: false})
document.querySelector('[role=textbox]').dispatchEvent(e);
```

The editor crashes.

Copying the html and pasting it normally (not via the console) also crashes.

Slate: 0.50
Browser: Chrome / Safari / Firefox / Edge
OS: Mac / Windows / Linux / iOS / Android

#### What's the expected behavior?

The editor shouldn't crash and should insert the contents.

ianstormtaylor:
Can you include the error that occurs?

scottfr:
Here you go:

```
backend.js:6 Error: Minified React error #152; visit https://reactjs.org/docs/error-decoder.html?invariant=152&args[]=qt for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at commons.f6d5446350e107a89ebc.js:1
    at Ga (commons.f6d5446350e107a89ebc.js:1)
    at Pu (commons.f6d5446350e107a89ebc.js:1)
    at Pl (commons.f6d5446350e107a89ebc.js:1)
    at _l (commons.f6d5446350e107a89ebc.js:1)
    at gl (commons.f6d5446350e107a89ebc.js:1)
    at commons.f6d5446350e107a89ebc.js:1
    at UdKW.t.unstable_runWithPriority (commons.f6d5446350e107a89ebc.js:1)
    at Ko (commons.f6d5446350e107a89ebc.js:1)
    at Zo (commons.f6d5446350e107a89ebc.js:1)
```

dmarkow:
Ran into this today. At least for me, nested formatting is causing it. Pasting `<em>One <u>Two</u></em>` crashes with the following output. Pasting `<em>One </em><em><u>Two</u></em>` (splitting the `em` tag) works fine.

```
index.es.js:429 Uncaught Error: The <text> hyperscript tag must only contain a single node's worth of children.
[...]

PR #3264, opened 2019-12-06
Fix paste example: nested formatting

dmarkow:

Fixing a bug.

#### What's the new behavior?

Pasting nested formatting, such as `<em>One <u>Two</u></em>` works instead of crashing.

Instead of a single `jsx("text", attrs, children)` call, the children are iterated over and return their own text nodes, which seems to handle multiple nested formats. I'm not sure if this ends up creating an excessive amount of text nodes, but it seems that `Editor.insertFragment` is smart enough to merge text nodes with identical attributes.

Fixes: #3188

### DIFF
--- a/site/examples/paste-html.js
+++ b/site/examples/paste-html.js
@@ -59,7 +59,9 @@ export const deserialize = el => {
     parent = el.childNodes[0]
   }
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Beautiful! Thank you very much @dmarkow.

PR #3266, opened 2019-12-06
Docs: Fix introduction links to examples

yozlet:
Existing links (to Github) were 404ing, so I changed them to point to slatejs.org.

Also: removed the final list item linking to the rest of the examples, since I can't see a page on slatejs.org that does that. Instead, I added a line after the examples list that talks about "View Source" and links to the Github folder listing.


ianstormtaylor:
Great idea, thanks @yozlet!

PR #3270, opened 2019-12-07
fix: ts

zbeyens:
Fix:

- Argument of type 'Point | undefined' is not assignable to parameter of type 'Location'.
  Type 'undefined' is not assignable to type 'Location'.
- Types of property 'zIndex' are incompatible.

### DIFF
--- a/site/examples/mentions.js
+++ b/site/examples/mentions.js
@@ -83,10 +83,14 @@ const MentionExample = () => {
         if (selection && Range.isCollapsed(selection)) {
           const [start] = Range.edges(selection)
           const wordBefore = Editor.before(editor, start, { unit: 'word' })
-          const before = Editor.before(editor, wordBefore)
-          const beforeRange = Editor.range(editor, before, start)
-          const beforeText = Editor.text(editor, beforeRange)
-          const beforeMatch = beforeText.match(/^@(\w+)$/)
+          let before
+          if (wordBefore) before = Editor.before(editor, wordBefore)
+          let beforeRange
+          if (before) beforeRange = Editor.range(editor, before, start)
+          let beforeText
+          if (beforeRange) beforeText = Editor.text(editor, beforeRange)
+          let beforeMatch
+          if (beforeText) beforeMatch = beforeText.match(/^@(\w+)$/)
           const after = Editor.after(editor, start)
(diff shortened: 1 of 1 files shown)


site/examples/mentions.js
```diff
@@ -83,10 +83,14 @@ const MentionExample = () => {
...
+          let before
+          if (wordBefore) before = Editor.before(editor, wordBefore)
+          let beforeRange
+          if (before) beforeRange = Editor.range(editor, before, start)
+          let beforeText
+          if (beforeRange) beforeText = Editor.text(editor, beforeRange)
+          let beforeMatch
+          if (beforeText) beforeMatch = beforeText.match(/^@(\w+)$/)
           const after = Editor.after(editor, start)
           const afterRange = Editor.range(editor, start, after)
```
ianstormtaylor: Would it be the same for these `after*` ones?

site/examples/mentions.js
```diff
@@ -83,10 +83,14 @@ const MentionExample = () => {
         if (selection && Range.isCollapsed(selection)) {
           const [start] = Range.edges(selection)
           const wordBefore = Editor.before(editor, start, { unit: 'word' })
-          const before = Editor.before(editor, wordBefore)
-          const beforeRange = Editor.range(editor, before, start)
-          const beforeText = Editor.text(editor, beforeRange)
-          const beforeMatch = beforeText.match(/^@(\w+)$/)
+          let before
+          if (wordBefore) before = Editor.before(editor, wordBefore)
```
ianstormtaylor: Instead of using the `if` statements, could you use `&&` like:

```js
const before = wordBefore && ...
```

site/examples/mentions.js
zbeyens: The `after*` are not throwing a ts error:

- `after` type is `Point | undefined` but
- `Editor.range` allows `undefined` for the 3rd parameter `to?: Point | Range | Path | undefined` and returns a `Range` (never undefined)

site/examples/mentions.js
zbeyens: Thanks for the reminder!

ianstormtaylor:
Beautiful, thanks @zbeyens!

Optional chaining would have been sweet, sounds like we need to tweak the Babel config for future stuff.

PR #3271, opened 2019-12-07
fix: ts

zbeyens:
Fix:

- Type 'undefined' is not assignable to type 'Range'.


ianstormtaylor:
Thanks @zbeyens, maybe we just remove the interim `suggest` variable completely then if we have to end up going around it. Would you be down to make that change?

PR #3280, opened 2019-12-09
Fix trailing new line in blocks

eddyw:

Fixing bug

#### What's the new behavior?

`String` inserts trailing new line when block `isLast`

A bad comparison with `node.children.length` was preventing `String` from inserting a trailing new line at the end of a block.

For instance:
```javascript
if (command.type === 'insert_break' && command.soft) {
  Editor.insertText(editor, '\n')
}
...
editor.exec({ type: 'insert_break', soft: true }) // <shift><enter>
```
Previously, if soft break was inserted at the end of the block, you'd need to insert it twice to see a new line because `isLast` was never going to be `true` 😜 

Fixes: 0.52.5
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/children.tsx
+++ b/packages/slate-react/src/components/children.tsx
@@ -69,7 +69,7 @@ const Children = (props: {
         <TextComponent
           decorations={ds}
           key={key.id}
-          isLast={isLeafBlock && i === node.children.length}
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Good catch, thank you @eddyw!

PR #3281, opened 2019-12-09
fix: typo

zbeyens:
`el.nodeNode` does not exist?


ianstormtaylor:
Haha thanks @zbeyens! Good catch.

Issue #3146, opened 2019-11-28
Unexpected scroll when start/end of a block is outside the viewport


zbeyens:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

https://i.imgur.com/p28xul5.gifv

https://www.slatejs.org/examples/rich-text 

Just paste a long paragraph. It scrolls everytime the start or the end of the block is outside the viewport. Scroll target seems to be the middle of the block?

Chrome 78.0.3904.108
Windows 10
Slate v0.50.0

#### What's the expected behavior?

It should not scroll when the cursor is in the viewport.

ejoebstl:
Proposed a setting to turn this on/off in #3284.

ianstormtaylor:
What’s the use case for adding a configuration to turn it off?

ejoebstl:
To prevent this behavior. 

An alternative would be to remove it entirely, but then it could break use cases I'm not aware of.

ianstormtaylor:
Prevent what behavior? And to what end?

ejoebstl:
The behavior described in this very issue by the original author. 

When a long paragraph which is partly out of the viewport gets selected, the window is scrolled so the paragraph is centered in the viewport. Also, the behavior is not consistent. Sometimes it does scroll, sometimes it does not.

I think this is rather unexpected and I'd like to turn it of, hence this PR.

Hope that helps!

ianstormtaylor:
Gotcha, yes the current bug of scrolling incorrectly should be fixed.

I don't think that merits adding more configuration to disable scrolling. It should instead be fixed so that it mimics the built-in browser behavior properly.

thallada:
My 2 cents: I think #3284 is worth merging regardless of the bug or not.

I have implemented my own scrolling logic for the editor that runs `onKeyDown` and `onSelect`. It does custom things that most users of Slate probably don't want (e.g. keeps a certain amount of buffer space between the cursor and the edge of the scroll container). But, this broken scrolling is running first and causes the scroll to jump back and forth. With no way to disable the Slate scrolling, I have to fork slate to remove the code.

tgonzalez-asapp:
I'm experiencing this issue too in a scrollable input. Turning off auto-scroll as proposed doesn't seem to improve things in this case:
![no-auto-scroll](https://user-images.githubusercontent.com/1714020/74609248-1472c000-50c7-11ea-9b60-6ddee80b4950.gif)
[...]

PR #3284, opened 2019-12-10
Make auto scrolling configurable.

ejoebstl:
Improves #3146 by adding a setting that can be used to turn auto-scrolling off.


ianstormtaylor:
Hey @ejoebstl thanks for taking the time to submit this. I'd like to get the auto-scrolling to be bug free, and then we can figure out if there are any cases where it really requires a config flag. Thanks for understanding.

PR #3288, opened 2019-12-10
Make Editable accept 'as' prop

eddyw:

Feature

#### What's the new behavior?

`Editable` is able to accept an optional `as` prop which can be an intrinsic element or a component (Class or Functional).

Example:
```typescript
<Editable as="section" /> // intrinsic element "section"
<Editable as={Component} /> // functional/class component
```
If a component is passed, it must forward refs. Added an example which shows how it should be done.

Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -57,22 +57,23 @@ export interface RenderLeafProps {
   }
 }
 
+export type EditableProps = {
+  decorate?: (entry: NodeEntry) => Range[]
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @eddyw! Ignore the Zeit thing, only got halfway on setting that up.

Can you remove the extra example though? I think it's not a common or complex enough need to have a full-fledged example for it. Thanks!

eddyw:
@ianstormtaylor ok!, removed last commit which added the example.

Issue #2051, opened 2018-08-08
try to prevent re-rendering at the Leaf level


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Idea/debt.

#### What's the current behavior?

Right now, we re-render the DOM in Slate with every change that occurs in the editor. This is different from Draft.js (which also uses React), which aborts re-rendering for "simple" changes to the DOM, like inserting or deleting characters. 

Our current approach has a lot of benefits in terms of simplicity... 

- It makes it easy to ensure that schema validations can be applied even at the text level. Since all changes will result in re-rendering, we don't have to differentiate and handle the "simple" changes in a more complex way to account for them already having been applied to the DOM, but now needing to be normalized.

- It makes it possible to create editors which render elements that update for any change, even basic ones like inserting text. Since every changes results in a re-render, plugins can render things like word counters, or block-specific styles, easily without having to account for the fact that some changes don't actually trigger renders.

- It makes all of the logic in the `Before/After` plugins a lot easier to follow, since they don't have to reverse engineer what the specific changes the browser may or may not have made in the `contenteditable` element out from under React.

However, even right now we're not doing it for 100% of DOM operations. We currently don't do it for spellchecking, since those changes are applied to the DOM immediately, without going through a usable event (at least in most browsers).

But this "always re-render" approach also has some downsides...

- It makes old browser and mobile browser support harder. Since these browsers use a system where the `beforeinput` event can't be prevented, preventing it makes it hard (and/or impossible) to get Slate to work in these browsers. https://github.com/ianstormtaylor/slate/issues/2047 https://github.com/ianstormtaylor/slate/issues/1720

- It makes IME support harder? Unsure of this one. But since IME is also a case where we can't actually prevent the defaults, since we need to read the text in the DOM, this might be a similar situation.

- It makes spellcheck glitchier, since browsers blink when text is re-rendered before they spellcheck it again. https://github.com/ianstormtaylor/slate/issues/1934

- It's less performant, in the case of simple insertions. Not terribly so, and for most editors this isn't a bottleneck, but it does require re-rendering the DOM even for "simple" cases like inserting a single character. This is not a big reason for doing this, since there are _many_ other places that would be important to improve first for this.

So, where does that leave us...

#### What's the expected behavior?

First, a look at what the render tree looks like for Slate:

```jsx
<UsersApp> (userland)
  <Editor>
    <Content>
      <Node>
        <UsersNode> (userland)
          <Text>
[...]

PR #3293, opened 2019-12-11
use native character insertion to fix browser/OS text features.

luddep:

fixing a bug

#### What's the new behavior?

Fixes spellcheck, text replacement, autocorrect and makes it so spellcheck squiggles dont blink in and out as you type.

This is a port of #3076 to 50+. It's fixing the same bug(s), with the same technique and comments from @ryanmitts.

Current behavior:
current

Fixed:
fixed

Single character insert events are not cancelled, as browser features like spellcheck only fires on user events and are removed on DOM updates. Native events are allowed to pass through, and <TextString /> will cancel its render if the effective text value hasn't changed.

Native Operations are queued and applied `onInput`, as for the <TextString /> check to work the native events need to propagate.

As #3076, this is pretty narrow and is only applied for `insertText` onBeforeInput events, with no selection, not at the end of inline nodes (as the selection is moved on `insert_text` elsewhere) or at the very beginning of a selection.

#3076 has more details.

Fixes: #2061 (and some of #2051?)
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -36,6 +36,7 @@ import {
   IS_FOCUSED,
   PLACEHOLDER_SYMBOL,
 } from '../utils/weak-maps'
+import { asNative, flushNativeEvents } from '../utils/native'
 
 // COMPAT: Firefox/Edge Legacy don't support the `beforeinput` event
 const HAS_BEFORE_INPUT_SUPPORT = !(IS_FIREFOX || IS_EDGE_LEGACY)
@@ -229,7 +230,47 @@ export const Editable = (props: EditableProps) => {
           return
         }
 
-        event.preventDefault()
+        let native = false
+        if (
+          type === 'insertText' &&
+          selection &&
+          Range.isCollapsed(selection) &&
+          // Only do it for single character events, for the simplest scenario,
+          // for now.
+          event.data &&
+          event.data.length === 1 &&
+          // Chrome seems to have issues correctly editing the start of nodes.
+          // I see this when there is an inline element, like a link, and you select
+          // right after it (the start of the next node).
+          selection.anchor.offset !== 0
+        ) {
+          native = true
+
+          // Skip native if there are marks, as that means
+          // `insertText` will insert a node, not just text.
+          if (editor.marks) {
+            native = false
+          }
+
+          // and because of the selection moving in `insertText` (create-editor.tx).
+          const { anchor } = selection
+          const inline = Editor.above(editor, {
+            at: anchor,
+            match: n => Editor.isInline(editor, n),
+            mode: 'highest',
+          })
+          if (inline) {
+            const [, inlinePath] = inline
+
+            if (Editor.isEnd(editor, selection.anchor, inlinePath)) {
+              native = false
+            }
+          }
+        }
+
+        if (!native) {
+          event.preventDefault()
+        }
 
         // COMPAT: For the deleting forward/backward input types we don't want
         // to change the selection because it is the range that will be deleted,
@@ -327,7 +368,13 @@ export const Editable = (props: EditableProps) => {
             if (data instanceof DataTransfer) {
               ReactEditor.insertData(editor, data)
             } else if (typeof data === 'string') {
-              Editor.insertText(editor, data)
+              // Only insertText operations use the native functionality, for now.
+              // Potentially expand to single character deletes, as well.
+              if (native) {
+                asNative(editor, () => Editor.insertText(editor, data))
+              } else {
+                Editor.insertText(editor, data)
+              }
             }
 
             break
@@ -486,6 +533,13 @@ export const Editable = (props: EditableProps) => {
           },
           [readOnly]
         )}
+        onInput={useCallback((event: React.SyntheticEvent) => {
+          // Flush native operations, as native events will have propogated
+          // and we can correctly compare DOM text values in components
+          // to stop rendering, so that browsers functions like autocorrect
+          // and spellcheck work as expected.
+          flushNativeEvents(editor)
+        }, [])}
         onBlur={useCallback(
           (event: React.FocusEvent<HTMLDivElement>) => {
             if (
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -235,10 +235,12 @@ export const Editable = (props: EditableProps) => {
           type === 'insertText' &&
           selection &&
           Range.isCollapsed(selection) &&
-          // Only do it for single character events, for the simplest scenario,
-          // for now.
+          // Only do it for single character a-z or space for now.
+          // Long-press events (hold a + press 4 = ä) to choose a special character otherwise
+          // causes duplicate inserts.
           event.data &&
           event.data.length === 1 &&
+          /[a-z ]/i.test(event.data) &&
           // Chrome seems to have issues correctly editing the start of nodes.
           // I see this when there is an inline element, like a link, and you select
           // right after it (the start of the next node).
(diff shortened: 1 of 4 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -234,7 +234,47 @@ export const Editable = (
...
+          // for now.
+          if (event.data && event.data.length === 1) {
+            native = true
+          }
+
+          // Chrome seems to have issues correctly editing the start of nodes.
+          // I see this when there is an inline element, like a link, and you select
+          // right after it (the start of the next node).
+          const { anchor } = selection
+          if (anchor.offset === 0) {
```
ianstormtaylor: Can you collapse this `anchor` if statement and the previous `event.data` if statement into the one above it, just to do it all in one go. (While preserving this great comment about Chrome compatibility.)

packages/slate-react/src/components/editable.tsx
```diff
@@ -234,7 +234,47 @@ export const Editable = (
...
+              native = false
+            }
+          }
+
+          // Chrome seems to have issues correctly editing the start of nodes.
+          // I see this when there is an inline element, like a link, and you select
+          // right after it (the start of the next node).
+          //  if (selectionStartPoint.isAtStartOfNode(startTextNode)) {
+          //   canNativelyEdit = false
+          // }
```
ianstormtaylor: I think this can be removed now with the offset check above.

packages/slate-react/src/components/string.tsx
```diff
@@ -55,15 +55,42 @@ const String = (props: {
 /**
  * Leaf strings with text in them.
  */
+type TextStringProps = { text: string; isTrailing?: boolean }
+class TextString extends React.Component<TextStringProps> {
```
ianstormtaylor: Can you keep this as a functional component and use `useEffect` and `React.memo` instead to achieve the changes?

luddep: Not entirely sure how to get the same behavior of both (1) stopping the render, and (2) making sure React still reconciles (the `forceUpdateFlag` part), using hook mechanics.

Spent a few hours experimenting, and was able to get (1) with a React.memo wrap on a subcomponent (could probably be lifted up to `<Leaf />`) and not updating render value if `textContent` is the same as `text` prop, but haven't been able to solve (2).

(2) Manifests itself as seeing ghost characters on delete (e.g. typing `hi|`, delete, expect to see `h|`, but see `h|i`) or selection moving around (`hello |`, delete, `|hello`.

I'm a little stumped, but also haven't quite grokked functional components enough to think of a clever way of doing it. Happy to give it another go if anyone's got any ideas.

packages/slate/src/create-editor.ts
```diff
@@ -25,11 +25,23 @@ export const createEditor = (): Editor => {
   const editor: Editor = {
     children: [],
     operations: [],
+    nativeOperationsQueue: [],
     selection: null,
     isInline: () => false,
     isVoid: () => false,
     onChange: () => {},
+    flushQueuedNativeOperations: () => {
```
ianstormtaylor: I'd like to achieve this without exposing any user-facing pieces, so that it's completely under the covers, and could even be eliminated one day if browser behavior improves.

I believe this would be possible by using the `WeakMap` pattern that's used elsewhere in the codebase to keep track of internal state. In this case we'd probably need two:

```ts
const AS_NATIVE: WeakMap<Editor, boolean> = new WeakMap()
const NATIVE_OPERATIONS: WeakMap<Editor, Operation[]> = new WeakMap()
```

With an associated helper:

```js
asNative(() => {
  editor.exec(...)
})
```

Such that it ensures all operations that happen inside that function are queued as native operations. And the queue can be checked still in `apply`.

They should prevent having to add any new functions to the `Editor` interface, or exposing any `native?:` flags to transforms/operations as well.

luddep:
Makes sense! Moved it all into `slate-react`, and removed any new interface (no more `native?:` flags, etc.)

ryanmitts:
Thanks for porting this to the new version for me!

luddep:
updated to work with latest changes, and fix an insertion bug when using marks.

bkrausz:
Strong +1 to this PR: it fixes some other quirks. For example, we've been working on single-line Slate and this allows horizontal scroll as you type (like an input field) to work properly. Seems like the path forward.

I believe this breaks the [markdown shortcuts example](https://github.com/ianstormtaylor/slate/blob/master/site/examples/markdown-shortcuts.js): since all of the operations get queued up, it's hard to figure out what to do to effectively "undo" the typing that will happen beforehand. I wonder if there's a good heuristic to still preventDefault if we're queueing up certain operations, or if we should provide some hook for plugins to tell that they don't want this functionality because they're messing with structure.

bobjflong:
I'm not 100% sure, and not familiar with Slate internals, but I _think_ this breaks the accented character input for us on mac:

Screen Shot 2020-02-19 at 5 02 13 PM

I see an error like this upon selecting one of the accented forms:

```
Error: Cannot resolve a DOM point from Slate point: {"path":[0,0],"offset":2}
```

luddep:
@bkrausz good catch. By bailing out of the native queue and immediately flushing if any non insert_text ops were being applied, the example works again.

@bobjflong I'm unable to reproduce that bug. Still happening? (was maybe fixed in master?)

packages/slate-react/src/components/string.tsx
ajorkowski: @luddep @ianstormtaylor The problem is that you want to prevent a normal render unless you want to reconcile the textContent with the native element, so it is not so simple to do it with hooks. I think you will need to have two hook based components to do this, something like the following (sorry if I mangled the TS a little bit)?

```
const TextString = (props: { text: string; isTrailing?: boolean }) => {
  const { text, isTrailing = false } = props

  const ref = useRef(null);
  const forceUpdateFlag = useRef(false);

  if (ref.current && ref.current.textContent !== text) {
    forceUpdateFlag.current = !forceUpdateFlag.current;
  }

  return <TextStringMemo ref={ref} text={text} isTrailing={isTrailing} forceUpdate={forceUpdateFlag.current} />
}

const TextStringMemo = memo(forwardRef((props: { text: string; isTrailing: boolean, forceUpdate: boolean }, ref) => {
  return (
    <span
      data-slate-string
      ref={ref}
      key={forceUpdate ? 'A' : 'B'}
    >
      {text}	
      {isTrailing ? '\n' : null}
    </span>
  )
}));
```

ajorkowski: Actually, if I understand what you are trying to do is just make sure that the render is forced on text change it should be possible with a single component after you make it `memo`:

```
const TextString = React.memo((props: { text: string; isTrailing?: boolean }) => {
  const { text, isTrailing = false } = props

  const ref = useRef(null);
  const forceUpdateFlag = useRef(false);

  // Always change the key when a change to the text field, which will force react to render
  forceUpdateFlag.current = !forceUpdateFlag.current;

  return (
    <span
      data-slate-string
      ref={ref}
      key={forceUpdateFlag.current ? 'A' : 'B'}
    >
      {text}	
      {isTrailing ? '\n' : null}
    </span>
  );
});
```

ALITHEALI:
Hello! Any update on this one?

gztomas:
Is there anything blocking this PR to be merged? It dramatically improves the UX and fixes a number of issues.

emtesenair:
What's the status of this PR?

timbuckley:
Per Ian's comment, can we make that class component into a functional component using hooks?

brycedewitt:
Checking in on this, is anyone actively working on a functional conversion right now?

bkrausz:
I have the conversion, can put up a new PR tomorrow with it.

bkrausz:
H'okay, #3888 is out with the requested change.

dylans:
Superseded by #3888.

Issue #3275, opened 2019-12-08
change `normalizeNode` to receive only a path


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Debt.

#### What's the expected behavior?

With the change in https://github.com/ianstormtaylor/slate/issues/3273 there is a chance that people normalize a node and change the `node` in the document, but still pass on the existing `entry` object:

```js
const withPlugin = editor => {
  const { normalizeNode } = editor

  editor.normalizeNode = entry => {
    const [node, path] = entry

    // Do something that changes the `node` in the document.
    // ...

    return normalizeNode(entry)
  }

  return editor
}
```

But there's a bug there because the original `normalizeNode` receives an entry where the `node` is outdated.

So instead, I think it should change to have a signature of `normalizeNode(path)` and people can always just lookup the node themselves when they want to normalize it.

PR #3294, opened 2019-12-11
Fix forced layout example

dmarkow:

Fixing a bug

#### What's the new behavior?

Forced layout example works. In 0.53.0, it wasn't, due to a couple issues.

- Forced layout example was destructuring the first element of a `[node, path]` array into `[path]`, so `path` wasn't actually a path. And then it was forwarding `[path]` to the next pass, instead of `[node, path]`.
- It called `Node.children` but never imported it.

Note: If the plan is to execute #3275, then this PR should be ignored and `normalizeNode` should instead be changed to only use a path. But in the meantime, the example's broken.


ianstormtaylor:
Thanks @dmarkow!

PR #3297, opened 2019-12-11
Allow links to be pasted without a selection

sidwyn:
Improving a feature

#### What's the new behavior?
For the Links example, allows links to be pasted even when there is no text selected.

Dec-11-2019 09-31-27

It checks if the current selection's anchor == focus, and if so that means that there is no selection, and it should insert text before wrapping the link.

Fixes: #
Reviewers: @ianstormtaylor


ianstormtaylor:
Thanks @sidwyn!

PR #3310, opened 2019-12-12
Fix emoji offsets

rinoc:
Fixes regression in latest release for #3044 that causes cursor to get stuck in some emoji's.

#### What's the new behavior?
Arrowing through emoji's works as expected.

I just added the code back from this PR #3044; it seems to solve the issue.


packages/slate/src/interfaces/editor/queries/location.ts
```diff
@@ -854,3 +853,146 @@ const getWordDistance = (text: string): number => {
...
+    code === 0x2642 || // male (♂)
+    code === 0x2640 || // female (♀)
+    code === 0x2620 || // scull (☠)
+    code === 0x2695 || // medical (⚕)
+    code === 0x2708 || // plane (✈️)
+    code === 0x25ef // large circle (◯)
+  )
+}
+
+const getCharOffset = (text: string, forward?: boolean): number => {
```
ianstormtaylor: This `forward` argument doesn't ever appear to be used?

rinoc: You're right! Removed it.

packages/slate/src/interfaces/editor/queries/location.ts
```diff
@@ -854,3 +853,125 @@ const getWordDistance = (text: string): number => {
...
+
+const getCharOffset = (text: string): number => {
+  let offset = 0
+  // prev types:
+  // SURR: surrogate pair
+  // MOD: modifier (technically also surrogate pair)
+  // ZWJ: zero width joiner
+  // VAR: variation selector
+  // BMP: sequenceable character from basic multilingual plane
+  let prev = null
```
ianstormtaylor: Can you set the type here to get this to typecheck?

ianstormtaylor:
Thank you @rinoc, I really appreciate you taking the time to port this one over. This is great.

PR #3326, opened 2019-12-15
add guard checks for value and editor in <Slate />

rockettomatooo:

Bug

#### What's the new behavior?

An error message is displayed when the user passes faulty values to `<Slate />`:
image
(I passed `null` as value)

`value` and `editor` props passed to `<Slate />` are checked with `Node.isNodeList()` and `Editor.isEditor()`.

--


rockettomatooo:
Unfortunately, there are no tests setup for `slate-react`. I tested this with an example (See the screnshot above).

Also the linter doesn't like my `console.error()`. I dunno if it's maybe better to throw an error instead (for a stacktrace maybe) but that would also kill the whole editor.

ianstormtaylor:
Hey @rockettomatooo can you use the `tiny-invariant` package for this? That's what we've used in the past. 

Also I think those lines should be moved inside the `useMemo` callback, just to make them slightly less frequent (although it's still fairly often).

rockettomatooo:
Hi @ianstormtaylor I now added the `tiny-invariant` package as a dependency to `slate-react` and moved the guard checks into the `useMemo` callback.
Also I merged my branch with the current master as there was a conflict with my first commit.

CameronAckermanSEL:
(FYI I resolved the merge conflict in package.json)

ianstormtaylor:
Thanks @rockettomatooo! Sorry for the delay.

Issue #3242, opened 2019-12-05
Remove regenerator runtime from distributed ES modules


jackturnbull:
#### What's the current behavior?

`regenerator-runtime` is being bundled in `slate/dist/index.es.js`. It does seem like I'm in a minority here but I'm unable to use slate.js at all because of two things;

- I run a CSP that does not allow for unsafe-eval.
- regenerator-runtime makes a call to `Function()` on `slate/dist/index.es.js:2164` which triggers the CSP warning, blocking slate.

This would work if the file which contains the slate code in my project is not set to strict mode, or if I change my CSP to allow for unsafe evaluation, but both feel like I'm working around the root cause. I guess one workaround would be to vendor this dist slate.js into its own script file but I also lose tree-shaking and dynamic imports.

Realistically though any system, bundler, or browser that supports ES modules will also support the async/await syntax so transforming async/await should probably not be happening at the package level the ES module source, you can leave the async/awaits as-is in your source and either let the browser handle it natively or let the package consumers choose their own async transformation strategy (I use promises instead of regenerator for instance.)

My gut feeling is you'll also reduce your bundle size a little this way so it'll be a net positive.

ref:
- browser support for async: https://caniuse.com/#feat=async-functions
- browser support for es modules: https://caniuse.com/#feat=es6-module

Slate: 0.50.8
Browser: **All**
OS: **All**

#### What's the expected behavior?

Not to bundle regenerator runtime; let the consumer of the package choose to inject it.

PR #3337, opened 2019-12-17
build: relax transpilation for es modules

jackturnbull:

Fixing a bug.

#### What's the new behavior?

ES module output currently includes regenerator runtime functions which can block strict content security policies. For an ES module output this isn't required because any runtime supporting ES modules will also support async/await natively. It is also highly likely that users of the ES module would also be performing their own transipliation of async/await and we should not make assumptions on their behalf (some may prefer promises to regenerator, for example). Only the ES module output will be changed

Although I expect this change to have little effect to the vast majority, if not all consumers of the ES modules, the change is somewhat fundemental and would be moving from providing an asynchronous runtime, to no longer providing it. Having never ran a project of this size, I'd only be guessing the best way to release this, but my gut feeling is to encourage a 'beta'/'@next' release of some kind to ensure that there aren't any edge-cases.

This commit splits the babel config into two; an unchanged one for the UMD output, and another for module output.

The regenerator runtime has been removed in the new ES module output.

_Before:_
```
slate on master [✘!?] ➜ rg 'regenerator' -l --no-ignore packages/
packages/slate/dist/index.js
packages/slate/dist/index.es.js
packages/slate/dist/slate.js
packages/slate/dist/slate.min.js
packages/slate/dist/index.js.map
packages/slate/dist/index.es.js.map
```
_After:_
```
slate on master [✘!?] ➜ rg 'regenerator' -l --no-ignore packages/
packages/slate/dist/slate.min.js
packages/slate/dist/slate.js
```

It also looks like the bundle size will drop slightly (for ES modules). This was to be expected as the consumers no longer vendor regenerator-runtime but also due to targeting newer browsers (requiring less transpilation).

_Before:_
```
slate on master ➜ du -k packages/**/dist | grep -v interfaces
108     packages/slate-history/dist
160     packages/slate-hyperscript/dist
24      packages/slate-react/dist/plugin
40      packages/slate-react/dist/utils
56      packages/slate-react/dist/components
48      packages/slate-react/dist/hooks
1916    packages/slate-react/dist
1660    packages/slate/dist
```
_After:_
```
slate on module-output-runtime took 5s 148ms ➜ du -k packages/**/dist/ | grep -v interfaces
104     packages/slate-history/dist
128     packages/slate-hyperscript/dist
24      packages/slate-react/dist/plugin
40      packages/slate-react/dist/utils
56      packages/slate-react/dist/components
48      packages/slate-react/dist/hooks
1900    packages/slate-react/dist
1384    packages/slate/dist
```

Fixes: #3242 
Reviewers: <any>


config/rollup/rollup.config.js
```diff
@@ -86,12 +86,21 @@ function configure(pkg, env, target) {
     builtins(),
 
     // Use Babel to transpile the result, limiting it to the source code.
-    babel({
-      runtimeHelpers: true,
-      include: [`packages/${pkg.name}/src/**`],
-      extensions: ['.js', '.ts'],
-      configFile: './config/babel/babel.config.cjs',
-    }),
```
ianstormtaylor: Instead of splitting the `babel.config.cjs` file in two. Could we inline it here in the Rollup config? Since that's the only place we're using it. That way we can have the smallest possible footprint and most clarity on specifically what `isUmd` enables?

I'm imagining something like:

```js
    babel({
      runtimeHelpers: true,
      include: [`packages/${pkg.name}/src/**`],
      extensions: ['.js', '.ts'],
      presets: [
        '@babel/preset-typescript',
        ['@babel/preset-env', { 
          exclude: isUmd ? ... : ... ,
          modules: false }
        ],
        '@babel/preset-react',
      ],
      plugins: [
        ['@babel/plugin-transform-runtime', isUmd ? {...} : {}],
        '@babel/plugin-proposal-class-properties',
      ],
    })
```

And then we can get rid of `babel.config.js` altogether!

I'm just thinking about trying to figure out what necessitated the multiple files in the future and being very confused.

Thank you!

jackturnbull: Good call - I'll take a look tonight. Had no idea the rollup plugin handled this and it's actually a decent improvement I can add to a couple of my other projects so thanks for pointing this out.

jackturnbull:
@ianstormtaylor changes made, see what you think. Now to do my own projects :)

ianstormtaylor:
That is beautiful, thank you @jackturnbull!

I'll add it to the current breaking release that's about to go out.

jackturnbull:
Hero! Looking forward to getting to use it

Issue #3292, opened 2019-12-11
IME input bug in richtext demo


icarusysuper:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
![demo](https://user-images.githubusercontent.com/3395277/70581749-3c3f3680-1bf3-11ea-8f65-226ff1573b5f.gif)

Slate: 0.53.X
Browser: Chrome
OS: Windows

#### Reproduce steps

1. visit https://www.slatejs.org/examples/richtext
2. switch to IME input (Chinese pinyin)
3. select multiple lines
4. input something (IME input)

#### What's the expected behavior?

IME input work correctly in situation above

tianpanpan66:
https://user-images.githubusercontent.com/28037665/111422707-0fe56480-872a-11eb-801b-c4406f750a11.mp4

type chinese, crash

PR #3342, opened 2019-12-18
Fix IME input bug (#3292)

kena0ki:
bug #3292

#### What's the new behavior?
Error doesn't happen after IME input.

The error happens because React tries to remove DOM nodes which is already deleted when IME input started. So I added deletion of DOM nodes on compositionstart event.
And I moved `state.isComposing = true`  to the listener for compositionupdate event, this is because cursor doesn't work well properly if it is left in compositionstart event.

Fixes:  #3292
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -539,13 +539,24 @@ export const Editable = (props: EditableProps) => {
           },
           [attributes.onCompositionEnd]
         )}
+        onCompositionUpdate={useCallback(
+          (event: React.CompositionEvent<HTMLDivElement>) => {
+            if (
+              hasEditableTarget(editor, event.target) &&
+              !isEventHandled(event, attributes.onCompositionUpdate)
+            ) {
+              state.isComposing = true
+            }
+          },
+          [attributes.onCompositionUpdate]
+        )}
         onCompositionStart={useCallback(
           (event: React.CompositionEvent<HTMLDivElement>) => {
             if (
               hasEditableTarget(editor, event.target) &&
               !isEventHandled(event, attributes.onCompositionStart)
             ) {
-              state.isComposing = true
+              editor.exec({ type: 'delete_fragment' })
             }
           },
           [attributes.onCompositionStart]


Jokcy:
@ianstormtaylor this seems another **composition event change dom before react** error. If composition event delete any node rendened for `decoration` or `mark`, then the dom deleted before react update. When React do delete, the node already deleted.

packages/slate-react/src/components/editable.tsx
```diff
@@ -539,13 +539,24 @@ export const Editable = (props: EditableProps) => {
...
+          [attributes.onCompositionUpdate]
+        )}
         onCompositionStart={useCallback(
           (event: React.CompositionEvent<HTMLDivElement>) => {
             if (
               hasEditableTarget(editor, event.target) &&
               !isEventHandled(event, attributes.onCompositionStart)
             ) {
-              state.isComposing = true
+              editor.exec({ type: 'delete_fragment' })
```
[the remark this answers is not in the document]
kena0ki: Because cursor position needs to be in sync after DOM nodes are removed by executing the following lines.
https://github.com/ianstormtaylor/slate/blob/a55396230cb2c20a614777c94623e401b7860d17/packages/slate-react/src/components/editable.tsx#L181-L200

Otherwise cursor doesn't work properly, like this:

ezgif com-crop

kena0ki:
@Jokcy 
Are you still reviewing? 
Or this fix cause something bad? 
If so, I want to know how can I reproduce it.

Jokcy:
No, I already figure out my question, to me it's ok now.

kena0ki:
@Jokcy 
Thank you for reviewing.
Could you approve this fix by submitting feed back?
Screenshot from 2019-12-21 21-09-35

Jokcy:
I do can approve it, but I don't quite understand what I do this mean for. I review this PR just because I was dealing some issue quite like this one. 

If you want to merge this PR, ianstormtaylor needs to know whether this code will cause other problems.

But if there is any necessity that I need to approve, let me know

kena0ki:
@Jokcy
According to Github help, a pull request is not automatically merged just being approved. Merge needs to be done by maintainers.
So, If your review has been done, I'd like you to approve this fix.

About reviews:
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-reviews

About merge:
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/merging-a-pull-request

packages/slate-react/src/components/editable.tsx
```diff
@@ -550,13 +550,27 @@ export const Editable = (props: EditableProps) => {
...
         onCompositionStart={useCallback(
           (event: React.CompositionEvent<HTMLDivElement>) => {
             if (
               hasEditableTarget(editor, event.target) &&
               !isEventHandled(event, attributes.onCompositionStart)
             ) {
-              state.isComposing = true
+              const { selection } = editor
+              if (selection && Range.isExpanded(selection)) {
+                Editor.deleteFragment(editor)
```
EleanorMao: I think you should set selection to anchor after delete fragment.
Selection will be set to focus according [here](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/transforms/text.ts#L181).  This causes the preview text to be misplaced when you typing.

EleanorMao: I think you should set selection to anchor after delete fragment.
Selection will be set to focus according [here](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/transforms/text.ts#L181).  This causes the preview text to be misplaced when you typing.

packages/slate-react/src/components/editable.tsx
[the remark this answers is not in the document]
EleanorMao: when you typing あい was at wrong place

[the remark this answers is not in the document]
kena0ki: The above gif shows wrongly fixed version (i.e. `state.isComposing = true` is in composition**start** event not in composition**update** event).

packages/slate-react/src/components/editable.tsx
kena0ki: Thank you for the comment.
I don't think `Transforms.select(editor, { anchor, focus})` is necessary after delete fragment.
`Editor.deleteFragment(editor)` should set selection if it is needed.

You can test if fixed version works correctly or not from Netlify link below. If it doesn't work as you expected, please let me know.

Screenshot from 2019-12-28 13-51-51

EleanorMao: You are right. Actually I had misplace problem in v0.56.0, and i fixed by that code. Your code do not have this problem. I think I‘ll figure it out later

yyjhao:
Since this is approved, any change you can merge it @kena0ki ?

kena0ki:
@yyjhao 
Sorry, I don't have write access😥.

FYI : You can use the version this patch applied by specifying commit-ish or forked repository in `package.json`.
See https://docs.npmjs.com/files/package.json#dependencies.

yyjhao:
@EleanorMao can you merge this? What is the usual process of merging PRs for slate.js?

EleanorMao:
@yyjhao 
Sorry, I can't

packages/slate-react/src/components/editable.tsx
baijiahei: Chrome:81.0.4044.129
macOS:10.15.4 (19E266)
slate:0.58.0
The problem still exists！

clmystes:
any progress here?

alex521:
+1

mfranzs:
+ 1 :) What are the remaining blockers here?

laobeijing0806:
hey guys, I've just met this problem, when will this PR be merged?

codingcn:
@ianstormtaylor Please merge this request, thanks.

Q-Qplus1s:
2020-12-24 same problem +1

miniwangdali:
Hi, we are facing the same problem. Do we know when will this fix get merged?

tianpanpan66:
same problem + 1

ianstormtaylor:
Thank you @kena0ki! Sorry for the delay.

Issue #3345, opened 2019-12-18
<Editable>, the onBeforeInput callback is created the wrong way


cvlmtg:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

inside `editable.tsx` the `onBeforeInput` is created this way:

```
        onBeforeInput={useCallback((event: React.SyntheticEvent) => {
          // COMPAT: Firefox doesn't support the `beforeinput` event, so we
          // fall back to React's leaky polyfill instead just for it. It
          // only works for the `insertText` input type.
          if (IS_FIREFOX && !readOnly) {
            event.preventDefault()
            const text = (event as any).data as string
            editor.exec({ type: 'insert_text', text })
          }
        }, [])}
```

the array of dependencies is empty, which means that if I render `<Editable readOnly={false} />` and then `<Editable readOnly={true} />`, inside the callback the value of readOnly will always be false. This breaks my app because the Editable components calls `preventDefault()` on events generated by an `<input>` tag, preventing the user from entering anything.

Slate: 0.52.2
Browser: Firefox
OS: all

#### What's the expected behavior?

Modify the dependencies array to be `[ readOnly ]`, so that everything works as expected

PR #3347, opened 2019-12-18
fix onBeforeInput memoization

cvlmtg:

bugfix

#### What's the new behavior?

the onBeforeInput callback is created on first render and then never changed, this means that if the user renders `<Editable>` with a different value of the `readOnly` prop, the callback will never see the updated value.

it correctly memoizes the callback passing `[readOnly]` in the dependencies

Fixes: #3345
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -442,7 +442,7 @@ export const Editable = (props: EditableProps) => {
             const text = (event as any).data as string
             editor.exec({ type: 'insert_text', text })
           }
-        }, [])}
+        }, [readOnly])}
         onBlur={useCallback(
           (event: React.FocusEvent<HTMLDivElement>) => {
             if (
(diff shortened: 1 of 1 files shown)


cvlmtg:
Update: when erasing a void node via `Editor.removeNodes` I was seeing error messages like this one:

    Error: Cannot find a descendant at path [0,1] in node: {"children":[{"type":"paragraph","children":[{"text":"prova  fffkf"}]},{"type":"paragraph","children":[{"text":"dssdf"}]}],"operations":[{"type":"remove_node","path":[0,1],"node":{"type":"formula","formula":"1+1=3","children":[{"text":""}]}},{"type":"merge_node","path":[0,1],"position":6,"target":null,"properties":{}}],"selection":null,"marks":null,"history":{"undos":[[{"type":"set_selection","properties":{"anchor":{"path":[0,0],"offset":0}},"newProperties":{"anchor":{"path":[1,0],"offset":5}}}],[{"type":"remove_node","path":[0,1],"node":{"type":"formula","formula":"1+1=3","children":[{"text":""}]}},{"type":"merge_node","path":[0,1],"position":6,"target":null,"properties":{}}]],"redos":[]}}

after fixing the memoization of the `onClick` callback like I have done for the `onBeforeInput` one I decided to fix all callbacks using `readOnly`

ianstormtaylor:
Thank you for investigating this @cvlmtg, this is great!

PR #3349, opened 2019-12-18
Swap remove_mark/add_mark example

dmarkow:

Fixing a bug.

#### What's the new behavior?

Changelog incorrectly called `add_mark` when `isActive` is true. Should be the opposite.


ianstormtaylor:
😄 thanks @dmarkow!

Issue #3343, opened 2019-12-18
remove redundant method Editor.unsetNodes


thesunny:
#### Do you want to request a _feature_ or report a _bug_?

remove unnecessary method

#### What's the current behavior?

Currently we can use `setNodes` with `null` property values or `unsetNodes` with an array to remove properties from nodes.

Slate: 0.50+

#### What's the expected behavior?

As per Slack discussion with Ian, we can remove the `unsetNodes` method since it's now unnecessary.

dylans:
Given the feedback in https://github.com/ianstormtaylor/slate/pull/3353#issuecomment-894975949 , it seems like keeping the `unsetNodes` method may be convenient enough to not remove. Let me know if we should close this out or proceed with removing.

thesunny:
I'm okay with keeping it as a convenience feature. If we are doing a bigger cleanup of the API removing unnecessary methods, maybe we could reconsider it then.

PR #3353, opened 2019-12-18
Remove unsetNodes method

thesunny:

Simplifying the API

#### What's the new behavior?

Removes the method `unsetNodes` because the same functionality is available using `setNodes` with a `null` value

Removes the method and changes any uses of the method to use the `setNodes` method instead

Fixes: #3343 
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate/src/create-editor.ts
+++ b/packages/slate/src/create-editor.ts
@@ -277,10 +277,14 @@ export const createEditor = (): Editor => {
 
       if (selection) {
         if (Range.isExpanded(selection)) {
-          Transforms.unsetNodes(editor, key, {
-            match: Text.isText,
-            split: true,
-          })
+          Transforms.setNodes(
+            editor,
+            { [key]: null },
+            {
+              match: Text.isText,
+              split: true,
+            }
+          )
         } else {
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Thank you @thesunny this is great! 

I'm going to wait to release since I just cut a breaking release a few hours ago. Whenever the next one is ready, or if nothing else comes in a bit, then I'll merge.

thesunny:
Sounds good.

BrentFarese:
@CameronAckermanSEL do you want to merge this in?  I can unless you have any comments/reservations?

thesunny:
@BrentFarese @CameronAckermanSEL 

Just a reminder to indicate through the release number that this is a breaking change when a release is made.

CameronAckermanSEL:
Damn I really wish I remember this one when we rolled out 0.58

dylans (changes requested):
Assuming we still want to proceed with this, it needs a rebase and changset.

TheSpyder:
I find `unsetNodes` useful for removing marks, where the key is dynamic so I'd need to do `{ [key]: null }` myself with this change.

Issue #3338, opened 2019-12-18
editor.selection is delayed so selection is undependable to find current element


thesunny:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

Press a cursor key and the cursor position is not updated in `editor.selection` for between 100ms and 200ms on a fast computer.

This causes funkiness like holding down `tab` in a table. We read the current `td` and move to the next but because of the slow selection updates, the cursor jumps several times to the same `td` because it is returning the old selection (i.e. cursor has moved but we keep getting the `td` from before the move because of the old selection).

Also the cause of other issues to do with cursor movement.

Slate: 0.55.3
Browser: Chrome
OS: Mac

#### What's the expected behavior?

In earlier Slate like 0.47, the `editor.selection` was updated within a `requestAnimationFrame`.

thesunny:
I think I found the cause of the delay and it's in the `debounce` in the `onDOMSelectionChange` method here https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/editable.tsx

@ianstormtaylor I noticed that you switched lodash's `_.throttle` to the `debounce` package and this is probably what is causing the issue. Looking at the docs, lodash's throttle invokes on both the leading and trailing edge of the call by default and you are using the default call. This means that in 0.47 we get the first selection change immediately.

This differs from `debounce` which always waits until [x]ms has ellapsed before firing for the first time.

Are you okay with switching back to `lodash's throttle method?

ianstormtaylor:
Hey @thesunny thanks for investigating this. If I remember right, there was some reason I ended up using `debounce` instead of throttle, but I can't remember why.

I think the _ideal_ would be to use neither, and just let the code run very frequently? But I feel like I remember there being an issue with that too (other than performance, which I'm okay to wait and see if it's even a problem).

If you can verify that switching to `_.throttle` fixes it I'm down to revert that. Although I'd prefer to use a smaller dependency, since it seemed like `lodash` was always really complex to ensure that the bundle size stayed small.

Ideally I'd love to get rid of throttling or debouncing entirely.

thesunny:
There's two ways to keep the bundle size small.

1. Use the npm package `lodash.throttle` which is just the throttle code extracted.
2. Use the `lodash` package but import it with a path like `import throttle from 'lodash/throttle'`

[...]

PR #3355, opened 2019-12-19
Use lodash throttle instead of debounce for updating selection change

thesunny:

Fixing a bug

#### What's the new behavior?

Switched slate to use `lodash/throttle` instead of `debounce` package for selection changes in slate.

I just switched the method but the main change is the throttle will execute on the leading and trailing edge of the call while debounce only executed on the trailing edge. This means that the initial change was not reflected immediately causing some bugs when we grab the selection.

NOTE: I'm not experienced with typescript. I added `@types/lodash` package but am not sure if I did this properly, whether I grabbed the right version, or whether it's necessary at all.

Fixes: #3338 
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/package.json
+++ b/packages/slate-react/package.json
@@ -14,9 +14,8 @@
     "dist/"
   ],
   "dependencies": {
-    "@types/debounce": "^1.2.0",
     "@types/is-hotkey": "^0.1.1",
-    "debounce": "^1.2.0",
+    "@types/lodash": "^4.14.149",
     "direction": "^1.0.3",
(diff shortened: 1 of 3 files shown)


karelskopek:
@ianstormtaylor @thesunny Hey fellows! 🙌

Can we please merge this lovely PR? 🙏 We encountered the same issue rendering selection change highly unreliable (especially when using up and down arrows). 🙊 

Or does it require further testing? 💯

CameronAckermanSEL:
Thanks @thesunny , I tested this locally and it appears to work well. Thanks for the fix!

PR #3359, opened 2019-12-19
fix: Slate editor props

zbeyens:

Bug

 TS2322: Type 'Editor' is not assignable to type 'ReactEditor'.

### DIFF
--- a/packages/slate-history/src/with-history.ts
+++ b/packages/slate-history/src/with-history.ts
@@ -7,7 +7,7 @@ import { HistoryEditor } from './history-editor'
  * editor as operations are applied to it, using undo and redo stacks.
  */
 
-export const withHistory = <T>(editor: T): T & HistoryEditor => {
+export const withHistory = <T extends Editor>(editor: T): T & HistoryEditor => {
   const e = editor as HistoryEditor
   const { apply } = e
   e.history = { undos: [], redos: [] }
--- a/packages/slate-history/src/with-history.ts
+++ b/packages/slate-history/src/with-history.ts
@@ -7,8 +7,8 @@ import { HistoryEditor } from './history-editor'
  * editor as operations are applied to it, using undo and redo stacks.
  */
 
-export const withHistory = <T extends Editor>(editor: T): T & HistoryEditor => {
-  const e = editor as HistoryEditor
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
@zbeyens can you give me a code sample of how you're using this?

zbeyens:
I'm using this helper function to create an editor with an array of plugins. The output type is `Editor` and `Slate` currently only accepts `ReactEditor`.

```js
export const useCreateEditor = (
  editorPlugins: PluginEditor[] = [],
  plugins: SlatePlugin[] = []
) =>
  useMemo(() => {
    let editor = createEditor();

    const newEditorPlugins = createEditorPlugins(editorPlugins, plugins);

    newEditorPlugins.forEach(plugin => {
      editor = plugin(editor);
    });

    return editor;
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

zbeyens:
A simpler example everyone is using:

```js
const editor = useMemo(() => withHistory(withReact(createEditor())), []);
``` 
`editor` type is `HistoryEditor`

Typescript error:

> Property 'insertData' is missing in type 'HistoryEditor' but required in type 'ReactEditor'.

ianstormtaylor:
Dang. I think we need to figure out a way to have the plugins return an editor with the proper fields instead of loosening the expectation. Because a `ReactEditor` needs `insertData`. And a `HistoryEditor` needs `undo`, `redo` and `history`.

zbeyens:
Why not just setting the type to `Editor`? All examples are working well on my end with this change.

`[key: string]: any;` allows new fields anyways.

ianstormtaylor:
@zbeyens because then there's no type safety on those extra pieces? That doesn't sound good to me. Shouldn't the plugins be using `Editor & {...}` to instead denote that they've added extra stuff?

zbeyens:
True!

What about generic types for all plugins?

`export declare const withReact: <T>(editor: T) => T & ReactEditor;`
`export declare const withHistory: <T>(editor: T) => T & HistoryEditor;`

Works well on my end

ianstormtaylor:
Beautiful, thank you @zbeyens!

Issue #3360, opened 2019-12-19
can't programmatically focus the editor anymore...


cvlmtg:
#### Do you want to request a _feature_ or report a _bug_?

feature

#### What's the current behavior?

There's no longer a way to give programmatically focus to the content-editable element.
In slate 0.4x it was possibile to attach a ref to the `<Editor>` component. This is no longer the case because `<Editable>` is a stateless component

Slate: 0.57.0
Browser: all
OS: all

#### What's the expected behavior?

It should be possible to give focus programmatically to the editor

Killavus:
Isn't `ReactEditor.focus` function does what you want to achieve?

cvlmtg:
@Killavus no idea because it's basically undocumented and I didn't know it existet. I'll try it.

cvlmtg:
@Killavus yap it works, thanks!

MarkPare:
Is this documented anywhere?

paullaffitte:
@MarkPare I found that you can import ReactEditor from slate-react and it has a [focus](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugin/react-editor.ts#L116) method. Here is how you can use it.

```js
import { ReactEditor } from 'slate-react';
ReactEditor.focus(editor);
```

With `editor` being an instance of your editor. You can get it from `createEditor()` or `useSlate()` depending on where do you need it.

dons20:
This really needs to be documented...

PR #3361, opened 2019-12-19
allow to get ref on the contentEditable div

cvlmtg:

feature

#### What's the new behavior?

It is now possible to pass a `innerRef` prop to the `<Editable>` component, so the parent can e.g. programmatically call `ref.current.focus()` to focus the editor

The editable component uses the passed innerRef or, if none, its own like in the previous version

Fixes: #3360
Reviewers: @ianstormtaylor 

NB @ianstormtaylor I've never used typescript, so please check that my changes are indeed correct (lint passes though...). There's also a little catch: to make the focus work as in slate 0.4x I have to save the selection before putting the editor in read-only mode and then restore it before calling `focus()`. this is because I click on a custom element inside the document, no idea if it happens in other circumstances, so I wonder is it could be useful to save/restore selection automatically when switching the `readOnly` prop

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -76,6 +76,7 @@ export type EditableProps = {
   readOnly?: boolean
   role?: string
   style?: React.CSSProperties
+  innerRef?: React.Ref<HTMLDivElement>
   renderElement?: (props: RenderElementProps) => JSX.Element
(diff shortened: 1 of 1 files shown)


cvlmtg:
@ianstormtaylor ok I suppose this pull-request is useless as I've discovered that there's `ReactEditor.focus()`

PR #3364, opened 2019-12-19
feat: delete current line when deleting backward with line unit

### LINKED ISSUE #1623: cmd+backspace deletes the entire block, not just the current line

Bug.

Here's a gif of the problem in our Slate editor, compared with a native mac app like textedit. Command and backspace currently deletes the entire paragraph/block. Although the gif shows our use case, this is easily reproducible on the examples at http://slatejs.org/.

image

Only the current line should be deleted.

Note: cmd+shift creates the correct selection, so one way of phrasing the expected behavior is "it should do what cmd+shift, followed by backspace, does today".

austingreendev:
#### New Feature

This PR resolves #1623 and allows a `deleteBackward({ unit: 'line' })` to delete the current line.

#### What's the new behavior?

This PR mimics native text editor functionality and allows the `CMD + BACKSPACE` or (`deleteBackward({ unit: 'line' })` command) to delete the current line where the selection lives.

##### Before

before

##### After

delete_backward

A quick highlight of the logic for this addition:

1. Runs during the `deleteBackward` command is ran with `unit='line'`
2. The editor must have a selection value that is collapsed
3. It then finds a parent block element
4. If a parent element exists it will determine a `Range` for the current line
  - This range is found by getting an array of valid positions within the parent element
  - We then compare the `BoundingClientRect` of positions against the current selection to find the maximum range which is on the same line
  - This uses binary search to make this reasonably efficient
5. If that range is not collapsed (i.e. selection is at beginning of line) it will delete the range

The plugin API has changed a few times since I've started this PR 😄so if there are any better ways to gather positions, ranges, etc. please let me know!

#### Testing

There are currently no tests for `slate-react` and this logic requires a DOM environment to validate so I'm not sure if this can be tested fully.

N/A

Fixes: #1623

### DIFF
--- a/packages/slate-react/src/plugin/with-react.ts
+++ b/packages/slate-react/src/plugin/with-react.ts
@@ -15,14 +15,14 @@ export const withReact = <T extends Editor>(editor: T) => {
   const e = editor as T & ReactEditor
   const { apply, onChange, deleteBackward } = e
 
-  e.deleteBackward = (unit) => {
+  e.deleteBackward = unit => {
     if (unit !== 'line') {
       return deleteBackward(unit)
     }
 
     if (editor.selection && Range.isCollapsed(editor.selection)) {
       const parentBlockEntry = Editor.above(editor, {
-        match: (n) => Editor.isBlock(editor, n),
+        match: n => Editor.isBlock(editor, n),
         at: editor.selection,
       })
 
@@ -108,7 +108,7 @@ export const withReact = <T extends Editor>(editor: T) => {
     let attach = contents.childNodes[0] as HTMLElement
 
     // Make sure attach is non-empty, since empty nodes will not get copied.
-    contents.childNodes.forEach((node) => {
+    contents.childNodes.forEach(node => {
       if (node.textContent && node.textContent.trim() !== '') {
         attach = node as HTMLElement
       }
@@ -136,7 +136,7 @@ export const withReact = <T extends Editor>(editor: T) => {
     // Remove any zero-width space spans from the cloned DOM so that they don't
     // show up elsewhere when pasted.
     Array.from(contents.querySelectorAll('[data-slate-zero-width]')).forEach(
-      (zw) => {
+      zw => {
         const isNewline = zw.getAttribute('data-slate-zero-width') === 'n'
         zw.textContent = isNewline ? '\n' : ''
       }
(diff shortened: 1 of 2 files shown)


packages/slate-react/src/plugin/with-react.ts
```diff
@@ -1,17 +1,45 @@
...
-  const { apply, onChange } = e
+  const { apply, onChange, deleteBackward } = e
+
+  e.deleteBackward = unit => {
+    if (unit !== 'line') {
+      return deleteBackward(unit)
+    }
+
+    if (editor.selection && Range.isCollapsed(editor.selection)) {
+      const parentBlockElement = Editor.above(editor, {
```
BrentFarese, line 23:
```suggestion
      const parentBlockEntry = Editor.above(editor, {
```

packages/slate-react/src/plugin/with-react.ts
```diff
@@ -1,17 +1,45 @@
...
+        match: n => Editor.isBlock(editor, n),
+        at: editor.selection,
+      })
+
+      if (parentBlockElement) {
+        const parentElementRange = Editor.range(
+          editor,
+          parentBlockElement[1],
+          editor.selection
+        )
```
BrentFarese, line 33:
```suggestion
      if (parentBlockEntry) {
        const [, parentBlockPath] = parentBlockEntry;
        const parentElementRange = Editor.range(
          editor,
          parentBlockPath,
          editor.selection.anchor
        )
```

BrentFarese:
Related to #3644 (discussion in there about the `line` unit not actually deleting a line).

It'd be great if this could have some tests.  Any way to include?

packages/slate-react/src/plugin/with-react.ts
austin94: Updated with suggestion.

packages/slate-react/src/plugin/with-react.ts
austin94: Much cleaner. Updated with suggestion.

austin94:
Hi @BrentFarese thanks for taking a look. It had been a while since I'd looked at this PR so I've rebased on `master` and added your suggested changes.

> It'd be great if this could have some tests. Any way to include?

The `slate-react` package [doesn't include any tests at the moment](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/test/index.js). Without any existing text fixtures and line computation relying heavily on DOM APIs this would be difficult to implement. Should I open an issue to create tests for this package?

> Relating to #3644

Based on that discussion do you think this should still live within `slate-react`?

ianstormtaylor:
Thanks @austingreendev!

Issue #3352, opened 2019-12-18
Cannot decorate siblings in decorator callback


AWildFoxAppeared:
#### Do you want to request a _feature_ or report a _bug_?
Bug (?)

#### What's the current behavior?
When a decorator is called on a leaf node **A**, and the range returned from the decorator includes **B, C** which are siblings of **A**, the resulting decoration does not span from **A** to **C** Rather, the provided anchor and focus offset are treated as if applied to the path of **A**

**Sandbox:** https://codesandbox.io/s/slate-reproductions-pf2ip
**Slate:** 0.55.0
**Browser:** Chrome / Firefox 
**OS:** Linux

#### What's the expected behavior?
The range returned from the decorator should be interpreted independently from where it was generated. Alternatively, the limitation of being only able to specify paths including the node itself and its children should be made explicit.

AWildFoxAppeared:
In the sandbox, the bugFix Boolean toggles between desired and bug behavior. When the bugFix Boolean is turned on, the range is specified from the parent node, and everything works fine. Currently, the problem with this approach is that all Elements are updated on changes made to the editor. If a user was implementing a decorator intended to highlight sentences which may include marks, he would have to apply the decorator on the Element/Paragraph level. And since all Elements are updated on each change event, this might get very inefficient.

An alternative solution to allowing siblings to be set from inside another sibling, would be to only feed updated Elements to the decorator callback.

The limitation of not being able to decorate siblings seems to stem from the following piece of code:
https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/children.tsx#L45

Unless I'm misunderstanding, it seems that with the current architecture, the range returned from a decorator is not universal in the sense it is only passed on to its children. For it to be universal there would have to be a separation between when we gather all ranges and when we apply the decorations to these ranges. 

I'm not very familiar with Slate yet, as I've only started to look into it, but I suspect the aforementioned change might require a bit of work.

ianstormtaylor:
@AWildFoxAppeared thanks for opening this!

I forgot that this is indeed a limitation of how decorations work. (At least currently!) Because decorations are only calculated for each node as the rendering happens, going down the tree. And they only ever update when the node they were calculated for changes. So decorating adjacent nodes isn't possible, but also wouldn't be updated in the decorations.

I'm not sure if there's a way around this? I'd be down if so. Or I'm down for a way to make it more clear and explicitly prevent these.

PR #3368, opened 2019-12-20
Moved decorator calls for Elements to inside the memoized Element

AWildFoxAppeared:
Feature

#### What's the new behavior?
The need for Element-level decorators is demonstrated by the GIF below, where a search term is highlighted across multiple children of an Element. 
element-decorator

In the original example this was also possible, but it would not scale, as the Element-level decorator would be called on any change.

Decorator calls used to be called regardless of whether an Element had changed. This change moves the decorator call to inside the memoized Element. This makes it more efficient to implement Element-level decorators.

An important side-effect is that render useCallback hooks (such as those used for renderLeaf) require the decorators as callback dependencies.

- [ ] The new code matches the existing patterns and styles. **Not sure about the typing at slate-react/src/components/element.tsx:50 , advice is welcome**
- [X ] The tests pass with `yarn test`.
- [X ] The relevant examples still work. (Run examples with `yarn start`.)

Fixes: #3352 , though it's still not possible to decorate siblings, the need for this might gone, as we can efficiently decorate Elements and their children.
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/children.tsx
+++ b/packages/slate-react/src/components/children.tsx
@@ -41,23 +41,16 @@ const Children = (props: {
     const n = node.children[i] as Descendant
     const key = ReactEditor.findKey(editor, n)
     const range = Editor.range(editor, p)
+
     const sel = selection && Range.intersection(range, selection)
-    const ds = decorate([n, p])
-
-    for (const dec of decorations) {
-      const d = Range.intersection(dec, range)
-
-      if (d) {
-        ds.push(d)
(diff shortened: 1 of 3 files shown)


dylans (changes requested):
Hi @AWildFoxAppeared this PR hasn't received any attention since you opened it. If you're still interested in this, please rebase your PR and I'll review it. Thanks!

dylans:

Issue #3384, opened 2019-12-24
Checklist Example Bug


arpit016:
#### Do you want to request a _feature_ or report a _bug_?
A bug

#### What's the current behavior?
Currently, if the last item in the checklist is checked and if we press enter the next checklist item is automatically in checked state. And if the last item in the checklist is unchecked then on enter the next checklist item is unchecked. Basically the next item copies the checked state of previous checklist item. 

[Gif/Video of the issue](https://www.loom.com/share/b2c071abc5e347b2a7dd7f3f24a70bdb)

[Sandbox Environment reproducing the issue](https://codesandbox.io/s/xenodochial-dust-pkvi5)

Slate: 0.57.0
Browser: Chrome / Safari / Firefox / Edge
OS: Mac / Windows / Linux / iOS / Android

#### What's the expected behavior?
The expected behavior by default should be that on enter the new checklist item should always be in unchecked state.

PR #3385, opened 2019-12-24
Bug fix: By default checklist item should be unchecked

arpit016:

Bug Fix

#### What's the new behavior?
As shown in the gif below:
- If the last item is checked and user presses enter then the new checklist item by default will remain unchecked
- The line-through on text is shown for the checked item and not for unchecked item
- Pressing enter on an empty checklist replaces it with a new paragraph now.

chrome-capture-solved

[Code Sample showing new behavior](https://codesandbox.io/s/blissful-cache-qy97g)

Fixes: #3384 

Reviewers: @ianstormtaylor

### DIFF
--- a/site/examples/check-lists.js
+++ b/site/examples/check-lists.js
@@ -58,25 +58,24 @@ const withChecklists = editor => {
         const after = Editor.after(editor, start, { unit: 'line' })
         const afterRange = Editor.range(editor, start, after)
         const afterText = Editor.string(editor, afterRange)
+        const checklist = {
+          type: 'check-list-item',
+          checked: false,
+          children: [{ text: afterText }],
+        }
         if (afterText.length > 0) {
-          const checklist = {
-            type: 'check-list-item',
-            checked: false,
-            children: [{ text: afterText }],
-          }
           if (selection.anchor.offset === 0) {
             Transforms.delete(editor, { unit: 'line' })
             Transforms.insertNodes(editor, checklist)
+            Transforms.setSelection(editor, {
+              anchor: { path: editor.selection.anchor.path, offset: 0 },
+              focus: { path: editor.selection.anchor.path, offset: 0 },
+            })
             return
           }
           Transforms.splitNodes(editor)
           Transforms.setNodes(editor, checklist)
         } else {
-          const checklist = {
-            type: 'check-list-item',
-            checked: false,
-            children: [{ text: '' }],
-          }
           Transforms.insertNodes(editor, checklist)
         }
         return
--- a/site/examples/check-lists.js
+++ b/site/examples/check-lists.js
@@ -41,27 +41,39 @@ const withChecklists = editor => {
     })
     if (match) {
       const [matchingNode] = match
-      if (matchingNode.type !== 'paragraph') {
-        if (
-          matchingNode.children[matchingNode.children.length - 1].text
-            .length === 0
-        ) {
-          Transforms.delete(editor)
-          Transforms.insertNodes(editor, {
-            type: 'paragraph',
-            children: [{ text: '' }],
-          })
-          return
-        }
-        if (matchingNode.type === 'check-list-item') {
+      if (
+        matchingNode.children[matchingNode.children.length - 1].text.length ===
+        0
+      ) {
+        Transforms.delete(editor)
+        Transforms.insertNodes(editor, {
+          type: 'paragraph',
+          children: [{ text: '' }],
+        })
+        return
+      }
+      if (matchingNode.type === 'check-list-item') {
+        const [start] = Range.edges(editor.selection)
+        const after = Editor.after(editor, start, { unit: 'word' })
+        const afterRange = Editor.range(editor, start, after)
+        const afterText = Editor.string(editor, afterRange)
+        if (afterText.length > 0) {
+          const checklist = {
+            type: 'check-list-item',
+            checked: false,
+            children: [{ text: afterText }],
+          }
+          Transforms.splitNodes(editor)
+          Transforms.setNodes(editor, checklist)
+        } else {
           const checklist = {
             type: 'check-list-item',
             checked: false,
             children: [{ text: '' }],
           }
           Transforms.insertNodes(editor, checklist)
-          return
         }
+        return
       }
     }
     insertBreak()
--- a/site/examples/check-lists.js
+++ b/site/examples/check-lists.js
@@ -53,8 +53,9 @@ const withChecklists = editor => {
         return
       }
       if (matchingNode.type === 'check-list-item') {
-        const [start] = Range.edges(editor.selection)
-        const after = Editor.after(editor, start, { unit: 'word' })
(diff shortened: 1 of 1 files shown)


site/examples/check-lists.js
```diff
@@ -32,7 +32,45 @@ const CheckListsExample = () => {
...
+        if (
+          matchingNode.children[matchingNode.children.length - 1].text
+            .length === 0
+        ) {
+          Transforms.delete(editor)
+          const paragraph = {
+            type: 'paragraph',
+            children: [{ text: '' }],
+          }
+          Transforms.insertNodes(editor, paragraph)
```
zbeyens: What about:

`Transforms.setNodes(editor, { type: 'paragraph});`

eden-lane:
Please tell me there is an easier way to do this 🙈

site/examples/check-lists.js
arpit016: @zbeyens  As you can see below in the screenshot of second console Array position 7th if we use `Transforms.setNodes(editor, { type: 'paragraph' });`, then when an empty checklist is converted to paragraph it also contains the checklist additional property `checked:false` which is not correct in my view. 


Screenshot 2020-01-05 at 3 13 22 AM

zbeyens: True, it would be nice to have an option to "reset" the node state on `setNodes` (remove additional properties) @ianstormtaylor

arpit016: @zbeyens wouldn't it be better if the `setNodes` by default removes the additional properties and just copies the `type` and empty `children` ? or is there a scenario where doing this implicitly will break the code?

site/examples/check-lists.js
```diff
@@ -32,7 +32,45 @@ const CheckListsExample = () => {
...
+          Transforms.unwrapNodes(editor, {
+            match: n => n.type === 'check-list-item',
+            split: true,
+          })
```
zbeyens: This doesn't do anything as you just removed the check-list-item above, there is no "parent" check-list-item to unwrap?

arpit016: True. Let me update and test it once and I'll update the pull request

arpit016: @zbeyens I tested the code with the suggested change and there were no issues with removing it. I have updated my pull request to reflect the same.

eden-lane:
I managed to achieve the same result with less code:
```
editor.insertBreak = () => {
        const match = Editor.above(editor, {
          match: n => n.type === ELEMENT_TYPE
        });

        if (match) {
          Transforms.insertNodes(editor, {
            type: ELEMENT_TYPE,
            checked: false,
            children: [{ text: '' }]
          });
          return;
        }

        insertBreak();
      };
```

Please review it, I might be missing something

site/examples/check-lists.js
```diff
@@ -32,7 +32,39 @@ const CheckListsExample = () => {
...
+  const { deleteBackward, insertBreak } = editor
+  editor.insertBreak = () => {
+    const match = Editor.above(editor, {
+      match: n => {
+        return Editor.isBlock(editor, n)
+      },
+    })
+    if (match) {
+      const [matchingNode] = match
+      if (matchingNode.type !== 'paragraph') {
```
CameronAckermanSEL: I'm not sure you need to handle the paragraph type here. The existing insertBreak handler you're calling at the end should be able to deal with this case handily.

site/examples/check-lists.js
```diff
@@ -32,7 +32,39 @@ const CheckListsExample = () => {
...
+            type: 'paragraph',
+            children: [{ text: '' }],
+          })
+          return
+        }
+        if (matchingNode.type === 'check-list-item') {
+          const checklist = {
+            type: 'check-list-item',
+            checked: false,
+            children: [{ text: '' }],
```
CameronAckermanSEL: This logic doesnt handle the case where the checklist item block is being split in when selection is collapsed in the middle. The end half of the line is truncated instead of the text being carried down to the new block.

CameronAckermanSEL:
Thanks so much for the improvement! It's really close. I think it can be simplified and I found a small bug that I think that should be addressed prior to merging.

site/examples/check-lists.js
arpit016: @CameronAckermanSEL  I agree the text doesn't get carried forward, its a bug, thanks for finding that out. Any idea on how can I handle that case?

arpit016:
> I managed to achieve the same result with less code:
> 
> ```js
> editor.insertBreak = () => {
>   const match = Editor.above(editor, {
>     match: n => n.type === ELEMENT_TYPE
>   });
> 
>   if (match) {
>     Transforms.insertNodes(editor, {
>       type: ELEMENT_TYPE,
>       checked: false,
>       children: [{ text: '' }]
>     });
>     return;
>   }
> 
>   insertBreak();
> };
> ```
> 
> Please review it, I might be missing something

The code is really concise, one thing that needs to be handled here is if the user presses enter on an empty checklist element then it should just get converted to a paragraph instead of creating another empty checklist item.

site/examples/check-lists.js
CameronAckermanSEL: @arpit016 have you tried overriding splitNodes instead?

arpit016: @CameronAckermanSEL  splitNodes will work if the selection or the cursor is in between the text, but if its at the end of the text, it wont work, so is there a method where I can check that the enter pressed is from the end of the text and not from between the text?

arpit016: @CameronAckermanSEL  I think I figured it out, updating the pull request in few minutes

arpit016: @CameronAckermanSEL I have updated the code to handle the bug that was there. Please checka nd review once

site/examples/check-lists.js
```diff
@@ -116,7 +166,7 @@ const CheckListItemElement = ({ attributes, children, element }) => {
         className={css`
           flex: 1;
           opacity: ${checked ? 0.666 : 1};
-          text-decoration: ${checked ? 'none' : 'line-through'};
+          text-decoration: ${checked ? 'line-through' : 'none'};
```
dylans, line 169: fwiw this fix has already landed in master.

If you have time to rebase your PR, I'll review and land it as this has been open for far too long in my opinion. :)

dylans (changes requested):

dylans:
As noted this change has become very stale and the main fix landed in a separate PR. Feel free to reopen or create another PR if there's interest in fixing this. Thanks.

eden-lane:
Found a better solution:
```typescript
editor.apply = (operation) => {
  if (operation.type === 'split_node' && 'type' in operation.properties) {
    const node = operation.properties;
    return apply({
      ...operation,
      properties: {
        type: node.type,
      },
    });
  }

  apply(operation);
};
```

In this case we override default `split_node` behaviour and **_don't copy_** existing props.

Issue #3321, opened 2019-12-13
Toggling readOnly prop from true to false breaks the Editor


martonlanga:
#### Do you want to request a _feature_ or report a _bug_?
Report a bug

#### What's the current behavior?
[Code Sandbox](https://codesandbox.io/s/slate-reproductions-jb0td)
If you init the Editor with readOnly set to true, then toggle it to false, the onChange method won't be called when making changes. But you can still edit the value, and see the changes.
If you init the Editor with readOnly set to false, everything works as expected.

Slate: 0.54.2
Browser: Chrome / Firefox
OS: Mac

#### What's the expected behavior?
Call `onChange` even after toggling the `readOnly` prop.

[Slack thread](https://slate-js.slack.com/archives/C1RH7AXSS/p1576227027259800)

rockcs1992:
@martonlanga  I'm also struggling with this bug. Do you have any fix or any work around?

martonlanga:
Hey @rockcs1992, I think the only workaround for now is to init the Editor to with readOnly set to false, then change it to true.
@kena0ki made a pull request that should fix this bug, but it's still not merged :/
https://github.com/ianstormtaylor/slate/pull/3388

rockcs1992:
@martonlanga Thanks for the update. I think I will go with your workaround at the moment. 😄

Francesco-Lanciana:
Another workaround until #3388 is merged is that you can set a key on the Slate component to be equal to your readOnly state, this seems to force Slate into the correct state.

martonlanga:
@Francesco-Lanciana That makes sense, great workaround!

rockcs1992:
@Francesco-Lanciana Great that works!

PR #3388, opened 2019-12-25
Don't use stale readOnly prop. (Fix bug #3321)

kena0ki:
Bug #3321

#### What's the new behavior?
Even after readOnly is switched from true to false, onChange is called.

Demo after fixed:
https://issue3321-test-slate.netlify.com/examples/read-only
https://github.com/kena0ki/slate/commit/566432e5f46e0b90431d34c52f9de58045f206b5#diff-3377000c0769495e84953bfcdd9bb9fc (Source code)

In order for onDOMSelectionChange and onDOMBeforeInput to be updated, I added dependencies to the second argument of two useCallback hooks and two useIsomorphicLayoutEffect hooks.

Although dependencies I added to useIsomorphicLayoutEffect hooks is onDOMSelectionChange and onDOMBeforeInput, an undefined variable error occurs if they are at the current location. So I moved them down below where they are defined.

Fixes: #3321
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -125,40 +125,6 @@ export const Editable = (props: EditableProps) => {
     }
   })
 
-  // Attach a native DOM event handler for `selectionchange`, because React's
-  // built-in `onSelect` handler doesn't fire for all selection changes. It's a
-  // leaky polyfill that only fires on keypresses or clicks. Instead, we want to
-  // fire for any change to the selection inside the editor. (2019/11/04)
-  // https://github.com/facebook/react/issues/5785
-  useIsomorphicLayoutEffect(() => {
-    window.document.addEventListener('selectionchange', onDOMSelectionChange)
-
-    return () => {
-      window.document.removeEventListener(
-        'selectionchange',
-        onDOMSelectionChange
-      )
-    }
-  }, [])
-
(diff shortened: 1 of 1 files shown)


Jmaharman:
I was having a problem with editors having an initial readOnly state of true, then being changed to false, where by the editor was not editable.

This has resolved that and neatly solves any issue of onDOMSelectionChange or onDOMBeforeInput having other dependencies changed in the future without similar issues arising.

ryanmitts:
This will also provide a resolution to https://github.com/ianstormtaylor/slate/issues/3412

packages/slate-react/src/components/editable.tsx
```diff
@@ -392,9 +376,25 @@ export const Editable = (props: EditableProps) => {
         }
       }
     }, 100),
-    []
+    [readOnly]
```
ryanmitts: The debounce for this function will somehow need to be cleared too, otherwise it's possible it memoizes and uses the stale value while it's waiting during debouncing.

ryanmitts: [This PR](https://github.com/ianstormtaylor/slate/issues/3338) will fix that it looks like, so all good if both are merged.

CameronAckermanSEL:
Thanks for fixing this readonly bug, really appreciate the contribution @kena0ki

PR #3389, opened 2019-12-26
fix: check isContentEditable of target element in ReactEditor.hasDOMNode

efimweb:
fixing a _bug_:
```ReactEditor.hasDOMNode``` checks ```isContentEditable``` of editor element. When element with ```contentEditable={false}```, which is not exists in editor value (e.g., some helper element), get selection, ```onDOMSelectionChange``` [checks](https://github.com/ianstormtaylor/slate/blob/22d9095c39a0e201878e1df04ef5e35d4d86a596/packages/slate-react/src/components/editable.tsx#L385-L386) work incorrectly.

demo: 
- https://codesandbox.io/s/slate-reproductions-lblhy
- on official [check-lists demo](https://www.slatejs.org/examples/check-lists) page (click near the left of the checkbox):
Kapture 2019-12-26 at 12 31 36

#### What's the new behavior?
```ReactEditor.hasDOMNode``` now checks ```isContentEditable``` prop of target element instead of editor element.

### DIFF
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -149,15 +149,15 @@ export const ReactEditor = {
     options: { editable?: boolean } = {}
   ): boolean {
     const { editable = false } = options
-    const el = ReactEditor.toDOMNode(editor, editor)
-    let element
+    const editorEl = ReactEditor.toDOMNode(editor, editor)
+    let targetEl
 
     // COMPAT: In Firefox, reading `target.nodeType` will throw an error if
     // target is originating from an internal "restricted" element (e.g. a
     // stepper arrow on a number input). (2018/05/04)
     // https://github.com/ianstormtaylor/slate/issues/1819
     try {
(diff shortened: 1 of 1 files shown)


ryanmitts:
This change is great!

It will also fix #3426 and #3425 (with PR #3436 needed for Firefox support).

0ctothorp:
@ryanmitts I'm afraid this is not really fixing #3426 and #3425. I tried this fix and indeed it's getting rid of the mentioned error, but the mentioned issues deal with more problems and this is not fixing them 😢

ryanmitts:
> @ryanmitts I'm afraid this is not really fixing #3426 and #3425. I tried this fix and indeed it's getting rid of the mentioned error, but the mentioned issues deal with more problems and this is not fixing them 😢

My bad, I assumed they both wanted void nodes that can have whatever they want inside, but I see it might not be that simple.

This does allow void nodes to have whatever you want inside though.

CameronAckermanSEL:
Thank you so much for this fix! Looks great to me!

PR #3393, opened 2019-12-29
fix some doc's bug (with v0.57.1)

loveloki:
#### What's the new behavior?

fix: editor's exec invoke
    now, editor.exec() is not available in Slate v0.57.1
    so, use editor.insertText() to instead it

and other wrong in doc: 
- code blocks's info string
- delete a nerver used value


ianstormtaylor:
Thanks @loveloki!

Issue #3372, opened 2019-12-21
duplicated code


cvlmtg:
`src/transforms/general.ts` and `src/interfaces/general` both contains the `transform` function which contains the same code.

cvlmtg:
@ianstormtaylor supposing, again, I've understood the code organization of slate... shouldn't `Editor.normalize` really be `Transforms.normalize`?

PR #3395, opened 2019-12-29
remove duplicated code

cvlmtg:
#### Is this adding or improving a feature or fixing a bug?

bugfix

#### What's the new behavior?

`src/transforms/general.ts` and `src/interfaces/general.ts` both contains the `transform` function which contains the same code. Supposing I understood correctly the changelog for slate 0.57.0, I removed the code from the editor and left the one in the Transforms helper.

- [x] The tests pass with yarn test.
- [x] The linter passes with yarn lint. (Fix errors with yarn fix.)
- [x] The relevant examples still work. (Run examples with yarn start.)

Fixes: #3372
Reviewers: @ianstormtaylor


packages/slate/src/interfaces/editor.ts
```diff
@@ -1241,275 +1241,6 @@ export const Editor = {
     return text
   },
 
-  /**
```
maxov: the removed snippet is the only reference to `immer`, so you can remove the imports on the second line

cvlmtg: good catch, thanks!

CameronAckermanSEL:
Hey @cvlmtg  thanks for the PR. Am I correct in understanding that this is a breaking API change? Would it be better to just remove the `transforms` call from the `Editor` object, or would it be better for it to be an alias for `Transforms.transforms`?

cvlmtg:
@CameronAckermanSEL Hi, it's been a while since I fiddled with slate, but after a quick search in the current docs I see that `Transforms.transform` is the one that's documented, so my PR should not change any API. Removing `transform` from Editor should be safe and the best option, imho.

Issue #3394, opened 2019-12-29
<Editable> onPaste works only on Firefox


cvlmtg:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

if you render `<Editable onPaste={onPasteHandler} />`, the *onPasteHandler* function is called only on Firefox. On Safari and Chrome it is totally ignored. You can try with https://jsfiddle.net/wfzraptj/ and see that the alert is seen only on Firefox

Slate: 0.57.1
Browser: Chrome / Safari
OS: Mac

#### What's the expected behavior?

the onPaste handler should work on all browsers

PR #3396, opened 2019-12-29
On paste

cvlmtg:

bigfix

#### What's the new behavior?

if the user specify the `onPaste` props it is now called on every browser and not just Firefox

it calls the `onPaste` handler on every browser and not just Firefox

Fixes: #3394
Reviewers: @ianstormtaylor


ianstormtaylor:
Thanks @cvlmtg!

PR #3398, opened 2019-12-31
fix: incorrect unwrapping list items and quotes in MarkdownShortcuts …

eden-lane:

Improving

#### What's the new behavior?

Currently if in Markdown Shortcut example you remove second list item it will unwrap the whole list (so `<li>` elements will move to the root of the document). If you try to unwrap nested list Slate will change it to paragraph but will leave it inside list which is incorrect.

Now if you try to remove nested list element Slate will unwrap it and if list item is not nested it will change it to paragraph and will lift it to the root of the document.

### DIFF
--- a/site/examples/markdown-shortcuts.js
+++ b/site/examples/markdown-shortcuts.js
@@ -91,11 +91,16 @@ const withShortcuts = editor => {
           block.type !== 'paragraph' &&
           Point.equals(selection.anchor, start)
         ) {
-          Transforms.setNodes(editor, { type: 'paragraph' })
+          let type = block.type
+
+          if (path.length <= 2) {
+            type = 'paragraph'
+            Transforms.setNodes(editor, { type })
+          }
 
           if (block.type === 'list-item') {
-            Transforms.unwrapNodes(editor, {
-              match: n => n.type === 'bulleted-list',
+            Transforms.liftNodes(editor, {
+              match: n => n.type === type,
             })
           }
 


zbeyens:
https://github.com/ianstormtaylor/slate/pull/3403

What about this (see linked comment)?

eden-lane:
@zbeyens it looks way better!
But it's still possible to get `<p>` inside `<ul>` tag which is incorrect from my point of view. I might be wrong, please tell me what do you think.
image

zbeyens:
On delete? Can you reproduce on the [preview](https://deploy-preview-3403--slatejs.netlify.com/examples/markdown-shortcuts)?

eden-lane:
Yes, I recorded it:
https://recordit.co/8C3fhgH7aa

zbeyens:
Ah sorry, didn't read it was about nested lists. Nice then!

By the way, I see another issue on delete:

- set an empty list item
- next line, set an empty paragraph
- delete backward
- the empty list item is replaced by the paragraph

RFAAGYmZPX

macrozone:
> Ah sorry, didn't read it was about nested lists. Nice then!
> 
> By the way, I see another issue on delete:
> 
> * set an empty list item
> * next line, set an empty paragraph
> * delete backward
> * the empty list item is replaced by the paragraph
> 
> RFAAGYmZPX

I am about to implement slate 0.57.7 into https://github.com/react-page/react-page and i am struggling with lists for some time now. I got it nearly working, but `Transforms.unwrapNodes` seems to not do what i want and I think, its a bug in slate itself and the one in the example might be because of the same bug.

i would expect that `Transforms.unwrapNodes(editor, {at: path})` just unwraps that one node at `path` but instead it tries to unwrap the whole subtree where path is at.

CameronAckermanSEL:
Hey @eden-lane , I'll play with this more later today. I think at the very least its an improvement. Do you think that perhaps a schema that kicks paragraph nodes out of the list would be an easy way to solve the problem?

dylans:
This PR is stale and doesn't look like it's actively being worked on any longer. Feel free to open a new PR or reopen if I am incorrect.

PR #3402, opened 2020-01-02
docs: walkthroughs: make implicit React dependencies explicit

davidcalhoun:


ianstormtaylor:
Thanks @davidcalhoun!

Issue #3414, opened 2020-01-09
Paste and Match Style / Paste without Formatting is not properly handed by Slate


Killavus:
#### Do you want to request a _feature_ or report a _bug_?

Report a bug.

#### What's the current behavior?

Right now if you try to use Paste without Formatting (a.k.a. Patch and Match Style) in Slate it results in a successful paste of content which is not managed by Slate - you can either not remove it, or it gets removed entirely when you start typing anything (browser-dependent). This is caused by lack of `beforeinput` event when using this option.

Here is the bug in action: https://v.usetapes.com/gXVeb5sv64

**Browsers:** macOS Chrome, macOS Firefox, macOS Safari. I'm pretty sure all supported browsers are exhibiting this behaviour.
**Slate version:** 0.57.1

#### What's the expected behavior?

Paste without Formatting / Paste and Match Style pastes proper text nodes that can be edited by users.

I've created a [pull request](https://github.com/ianstormtaylor/slate/pull/3415) solving this issue. Since `beforeinput` is not emitted for these events, they need to be handled in `onPaste` - `paste` event is properly emitted for these types of pastes.

Killavus:
@ianstormtaylor do you need more details before you review the PR attached to this issue?

heymanhn:
Tested this on `v0.57.2` and it still doesn't work. Again, I can see the text nodes being inserted, but as with the original behavior, it seems like the content is still not managed by Slate. The selection doesn't change, and once I start typing something, the pasted content disappears.

https://share.getcloudapp.com/6qu5Y50B

PR #3415, opened 2020-01-09
Fix Paste without Formatting / Paste and Match Style producing uneditable text nodes.

Killavus:

Fixing a bug.

#### What's the new behavior?

This fixes https://github.com/ianstormtaylor/slate/issues/3414. Basically right now Paste and Match Style type of pasting from Chrome is working properly - as well as `mod+shift+v` shortcut from other browsers.

Since these types of pastes are not emitting `beforeinput` events, I handle them in `paste` event handler. After that it's just a matter of simply passing them to `insertData` of `ReactEditor`. I want to minimize the surface area of change, so this behavior is triggered only for pastes having exclusively `text/plain` ClipboardData in them.

Fixes: https://github.com/ianstormtaylor/slate/issues/3414
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -231,7 +231,6 @@ export const Editable = (props: EditableProps) => {
         isComposing: boolean
       }
     ) => {
-      console.log('onDOMBeforeInput', event)
       if (
         !readOnly &&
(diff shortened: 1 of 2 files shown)


CameronAckermanSEL:
Thanks so much for fixing this bug @Killavus !

PR #3431, opened 2020-01-15
Fix Getting Started links

JLRishe:
The Getting Started and Walkthroughs links both currently lead to a 404 page. The modification in this change should fix them to link to the correct location.

Fixing a documentation bug

#### What's the new behavior?

The Getting Started and Walkthroughs links on the Readme page should now correctly link to the walkthroughs

Removed the part at the end of the URLs that probably worked at one point but does not because the URL formats have apparently changed. Linking to /walkthroughs seems to auto-redirect to the first walkthrough.

- [N/A] The new code matches the existing patterns and styles.
- [N/A] The tests pass with `yarn test`.
- [N/A] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [N/A] The relevant examples still work. (Run examples with `yarn start`.)

Fixes: #
Reviewers: @

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -118,9 +118,9 @@ If you have an idea for an example that shows a common use case, pull request it
 
 ### Documentation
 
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @JLRishe, I believe this has been fixed by another earlier PR.

Issue #3426, opened 2020-01-14
onChange event not firing


Josh-McFarlin:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?
Inputs and other tags inside custom elements can only be changed if they are within a portal. 
onKeyDown fires correctly, but onChange never fires. Others on the slack channel have also come across this bug, and state the only solution is placing the custom element within a portal.

CodeSandbox: https://codesandbox.io/s/slate-reproductions-wcko1

```
<input
    onChange={() => console.log('this never fires')}
    onKeyDown={() => console.log('but this works')}
/>
```

Slate: 0.57.1
Browser: Chrome
OS: Mac

#### What's the expected behavior?
onChange should fire in custom elements without needing to be inside a portal / pop-up.

Francesco-Lanciana:
Hey this is quite the hack but if you wrap your input inside a div and give it the attribute ```data-slate-editor``` then the onChange handler should call. This fools Slate into thinking that the input is outside the editor so your input will work as normal. The issue I just created #3425 is more or less regarding the same thing, but my case was with CodeMirror which has a textarea whose onInput never fires. From what I can tell with some experimenting the culprit seems to be the Editable component that steals focus from inputs/textareas as you type. I would really like to get this resolved but in the meantime this hack seems to work nicely

Josh-McFarlin:
Can confirm @Francesco-Lanciana fix works

cvlmtg:
not sure if the problem I encountered is the same, but IIRC I fixed the problem by just wrapping my component (the one who was rendering the input element) inside a `<span {...attributes}>`

LiuJi-Jim:
> Hey this is quite the hack but if you wrap your input inside a div and give it the attribute `data-slate-editor` then the onChange handler should call. This fools Slate into thinking that the input is outside the editor so your input will work as normal. The issue I just created #3425 is more or less regarding the same thing, but my case was with CodeMirror which has a textarea whose onInput never fires. From what I can tell with some experimenting the culprit seems to be the Editable component that steals focus from inputs/textareas as you type. I would really like to get this resolved but in the meantime this hack seems to work nicely

ugly, but it works😰😂

alfechner:
Thanks for the work around!

It works in one way but it breaks jumping with the arrow keys before the custom element if it's for example the first node in the editor.

Francesco-Lanciana:
[...]

PR #3436, opened 2020-01-18
Fix inputs from not being able to be used within void nodes in Firefox

### LINKED ISSUE #3425: Create isolated elements inside Slate
Feature Request

If you have an element inside the React Editor that doesn't contain a text node and can receive focus (and therefore steal it from Slate) then you get the error: Cannot resolve a Slate node from DOM node.

This is an issue when you want to add something like CodeMirror inside the editor as it needs focus. 

It would be great if Slate gave you the ability to have isolated elements inside the Editor so you could have an escape hatch from slates event handlers. This would be useful for adding in elements that need to steal focus from the editor, for example a code editor like CodeMirror. It would also be useful for adding things like the hover toolbar that Notion has, where it needs to be positioned relative to one of slates elements.
Screen Shot 2020-01-07 at 11 42 09 am

Let me know if I'm just missing a key concept that makes this doable. Otherwise I'm happy to contribute a PR, I would just need a little guidance in terms of scoping the issue - i.e. what are all the things I haven't considered.

ryanmitts:
This change builds on top of PR #3389 (I can rebase when it is merged since I copied into the commit).

That PR allows void nodes to have any kind of element in them and the editor events will not respond to them if the void node renders with `contentEditable={false}`, this will treat all it's children as not editable unless specified otherwise, therefore events within the void node will be treated like they are outside the editor. Currently they cannot have inputs or any other content editable areas or the editor event handlers try to interpret them as part of the main editor document, when if they are in void nodes with `contentEditable` turned off, they should be treated as isolated islands.

This behavior could be achieved with the pre 0.50.x versions, therefore some of our projects will have regressions.

That PR will fix #3425 and #3426.

This PR fixes a bug specific to Firefox where input events within a void nodes prevent inputs from working.

I included an example of having these isolated void nodes in case it's of value for future testing and and to serve as an example.

Bug

#### What's the new behavior?

See the example. A void node can be made to be completely isolated.

The previous PR would stop all events on the editor from responding to non editable targets. This CR will stop the Firefox specific onBeforeInput event from firing in the same case.

Fixes: #3425, #3426, the other PR (#3389) listed above fixes it mostly, this makes it work in Firefox to resolve the same issues.

Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -442,11 +442,16 @@ export const Editable = (props: EditableProps) => {
           ...style,
         }}
         onBeforeInput={useCallback(
-          (event: React.SyntheticEvent) => {
+          (event: React.FormEvent<HTMLDivElement>) => {
             // COMPAT: Firefox doesn't support the `beforeinput` event, so we
             // fall back to React's leaky polyfill instead just for it. It
             // only works for the `insertText` input type.
-            if (IS_FIREFOX && !readOnly) {
+            if (
+              IS_FIREFOX &&
+              !readOnly &&
+              !isEventHandled(event, attributes.onBeforeInput) &&
+              hasEditableTarget(editor, event.target)
+            ) {
               event.preventDefault()
               const text = (event as any).data as string
               Editor.insertText(editor, text)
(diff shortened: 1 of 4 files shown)


huttunenMarco:
Running in to this problem as well. Looking forward to a merge 👍

CameronAckermanSEL:
Hey @ryanmitts ! Thanks so much for fixing this problem, this has been a real issue on Chrome and Firefox for this version of slate and I think people will appreciate that you were able to resolve it. Thank you. I tested this on chrome and firefox using the embeds example (which is still going to be busted after merging this PR, I'll have a follow up that fixes it up).

Issue #3444, opened 2020-01-20
Example site does not have Slate library source maps available to debug with


ryanmitts:
#### Do you want to request a _feature_ or report a _bug_?

Feature

#### What's the current behavior?

The example site only has the entire libraries index.js file available for each package to debug with.

It's usable to debug with, but it would be nice to have the original files as well.

Slate: 0.50.x
Browser: Chrome / Safari / Firefox / Edge
OS: Mac / Windows / Linux / iOS / Android

#### What's the expected behavior?

Have each .ts/.tsx file available as a source map for the examples site.

PR #3445, opened 2020-01-20
Make source map's of Slate libraries available to the example site Webpack config

ryanmitts:

Adding a feature.

#### What's the new behavior?

Source maps are available to the Webpack config for the Slate libraries to be included with the source maps of the site.

The source map loader makes linked source maps available to Webpack to process like the rest of the first party files.

Fixes: #3444
Reviewers: @

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "slate-history": "*",
     "slate-hyperscript": "*",
     "slate-react": "*",
+    "source-map-loader": "^0.2.4",
     "typescript": "^3.7.2"
   }
 }
--- a/site/next.config.js
+++ b/site/next.config.js
@@ -20,4 +20,12 @@ module.exports = {
 
     return pages
   },
+  webpack: config => {
+    config.module.rules.push({
+      test: /\.js$/,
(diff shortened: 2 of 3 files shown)


CameronAckermanSEL:
Hey @ryanmitts , this seems like a really useful PR but I'm having some trouble verifying that it's working (specifically I can't seem to get a production build of the docs site in a local environment). What did you do to test it locally?

CameronAckermanSEL:
Ah, just realized there's a netlify preview 😁  

However, I don't see any indication that it's picking up the source map. Am I missing something?

ryanmitts:
@CameronAckermanSEL Assuming the deploy preview is using a production build, I believe that Next.JS won't output/host the source maps.

In development (`npm run start`) I see them in the Chrome source tabs under the Webpack folder.

One thing though is it seems to not be able to distinguish between the packages within the repo, and the folders get merged, -- I don't know how to get around that.

slate_source_maps

Let me know if it's not working for you and I can have another look tomorrow.

CameronAckermanSEL:
Thanks so much! As far as I can tell this works great. :)

PR #3461, opened 2020-01-23
Add className to placeholder

lcswillems:

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/leaf.tsx
+++ b/packages/slate-react/src/components/leaf.tsx
@@ -33,6 +33,7 @@ const Leaf = (props: {
       <React.Fragment>
         <span
           contentEditable={false}
+          className="slate-placeholder"
           style={{
             pointerEvents: 'none',
             display: 'inline-block',


CameronAckermanSEL:
Thanks for the PR. I think the general approach makes sense but I'm feeling that maybe this is a bit too constricting and instead there should be a prop or something that allows the user to pass in their own class name. But then I'm also worried about what the surface area of that API would look like. I'll post back some ideas if I have some, but this might be something that Ian needs to look at since it's changing the API.

lcswillems:
Okay, I understand. @ianstormtaylor What would be the best solution here? Using a classname or properties?

delijah:
Why do you want to pass a `className`? You could also just pass your own component and style it the way you want it... like `<Editable placeholder={<Placeholder>{placeholder}</Placeholder>} />`.

ianstormtaylor:
Hey @lcswillems thanks for working on this. I agree with @CameronAckermanSEL that it feels like we should provide a more customizable version of this idea. I wrote up a solution in https://github.com/ianstormtaylor/slate/issues/4123 that I think would work, if any one wants to contribute! Thanks.

Issue #3416, opened 2020-01-09
Consider removing `[key: string]: any` from interfaces, and use type generics in function definitions.


timbuckley:
#### Do you want to request a _feature_ or report a _bug_?

Feature (for TypeScript users of Slate).

#### Problem discussion

In the current version of Slate core (`v0.57.1`), interfaces are declared in the following way
```ts
interface Text {
  text: string;
  [key: string]: any;
}
```

...and most Slate functions are written in a form like the following:
```ts
const Text = {
  //...
  decorations(node: Text, decorations: Range[]): Text[] {/* ... */}
}
```

The problem with this is that TypeScript users would likely want to constrain their `Text` nodes in their own way, above and beyond the underlying `Text`, `Editor`, etc interfaces, while still allowing them to be used directly in helper functions without using `as` type coercions.

For example:
```ts
interface CompanyText extends Text {
  isBold: boolean
  fontSize: number
}

const myTextNode: CompanyText = {text: 'foo', isBold: false, fontSize: 13}

Text.decorations(myTextNode, [])
// Type error! `myTextNode` is not of type `Text`
```

The `CompanyText` type is still overly permissive because a typo like `isBeld` is still allowed, despite the user's extended interface (thanks to key-any). And the user would also have to use `myTextNode as Text` when calling `Text.decorations()` on it, otherwise TypeScript will complain.

#### Proposed solution

Use type generics in functions, and remove the `[key: string]: any` from interfaces
```ts
interface Text {
  text: string;
}

const Text = {
  // ...
  decorations<T extends Text>(node: T, decorations: Range[]): T[]   {/* ... */}
}
```

This will allow TypeScript consumers of Slate to use it in the `CompanyText` example above.

miracle2k:
This would also be a huge help in migrating existing code over to the new API, as I am currently undertaking. For example, an old-style call such as `editor.setNodeByKey` does not show a typing error once I switch the type over the new new `Editor` (which has an index-signature).

PR #3473, opened 2020-01-29
[WIP] Use Generics for top-level functions 

timbuckley:

Improvement

#### What's the new behavior?

No behavior change, other than that TypeScript consumers of the library should be able to use their own extensions of top-level interfaces (eg `interface MyEditor extends Editor {...}`)

Fixes: #3416

### DIFF
--- a/docs/concepts/01-interfaces.md
+++ b/docs/concepts/01-interfaces.md
@@ -5,7 +5,6 @@ Slate works with pure JSON objects. All it requires is that those JSON objects c
 ```ts
 interface Text {
   text: string
-  [key: string]: any
 }
 ```
 
@@ -22,7 +21,6 @@ To take another example, the `Element` node interface in Slate is:
 ```ts
 interface Element {
   children: Node[]
-  [key: string]: any
 }
 ```
 
--- a/docs/concepts/02-nodes.md
+++ b/docs/concepts/02-nodes.md
@@ -55,7 +55,6 @@ Elements make up the middle layers of a richtext document. They are the nodes th
 ```ts
 interface Element {
   children: Node[]
-  [key: string]: any
 }
 ```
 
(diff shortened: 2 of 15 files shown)


timbuckley:
Tests do not pass yet, as I did not resolve all fixes in the types. In theory, there should be no breaking changes for TypeScript users (I believe), it should just create new flexibility in the constraints one puts in their types.

ianstormtaylor:
@timbuckley this approach sounds good to me! Let me know when you think it's ready to merge and I can give it a full review. Seems like we'll need to use generics for every API call is that right?

timbuckley:
@ianstormtaylor The goal is for current consumers of the TypeScript version not needing to do anything, but have the ability to pass in their own types extending the default.

This can be accomplished by my changing generics from...
```ts
edges<E extends Editor, El extends Element>() {}
```
to...
```ts
edges<E extends Editor = Editor, El extends Element = Element>(...) {}
```
... which will mean that usages such as `editor.edges(...)` still valid in TS, since the default value for the `E` type variable is just good ol' `Editor`.

This PR is a little difficult because there are no current tests written in TypeScript against which to check that usage behavior is preserved. What do you think about moving all tests (or at least unit tests) to TypeScript as well? If so that would be a another blocking PR of course.

jerkly:
Are there any updates on this pull?

thesunny:
This proposal solves the same problem of custom types but doesn't require you to use generics everywhere. You can extend custom types once and it gets used everywhere.

It it also has benefits over the declaration merging interface proposal in that it supports type discrimination and type unions.

https://github.com/ianstormtaylor/slate/issues/3725

dylans:
I believe this is superseded by #3725 and other work to improve TS definitions for Slate.

Issue #3471, opened 2020-01-28
Module parse failed: Unexpected token. You may need an appropriate loader to handle this file type


vmarquet:
#### What's the current behavior?

I simply added the library to my package.json, and now I can't even bundle my code:

<img width="1440" alt="Capture d’écran 2020-01-28 à 21 46 31" src="https://user-images.githubusercontent.com/6226610/73304416-81a5da80-4218-11ea-9160-c471edff77f3.png">
<img width="660" alt="Capture d’écran 2020-01-28 à 21 47 11" src="https://user-images.githubusercontent.com/6226610/73304426-85d1f800-4218-11ea-89e3-ba1cfd38b194.png">

Seems like it comes from the following files:

```
➜  node_modules master✗ ack 'autoFocus, decorate = defaultDecorate, onDOMBeforeInput'
slate-react/dist/slate-react.js
1215:      const { autoFocus, decorate = defaultDecorate, onDOMBeforeInput: propsOnDOMBeforeInput, placeholder, readOnly = false, renderElement, renderLeaf, style = {}, as: Component = 'div', ...attributes } = props;

slate-react/dist/index.js
579:    const { autoFocus, decorate = defaultDecorate, onDOMBeforeInput: propsOnDOMBeforeInput, placeholder, readOnly = false, renderElement, renderLeaf, style = {}, as: Component = 'div', ...attributes } = props;

slate-react/dist/index.es.js
572:    const { autoFocus, decorate = defaultDecorate, onDOMBeforeInput: propsOnDOMBeforeInput, placeholder, readOnly = false, renderElement, renderLeaf, style = {}, as: Component = 'div', ...attributes } = props;
```

Slate: 0.57.1
Browser: Chrome 79 & Firefox 72
OS: macOs

My **package.json**:

```
    "@rails/webpacker": "3.4",
    "babel-plugin-transform-runtime": "^6.23.0",
    "babel-preset-react": "^6.24.1",
    "babel-preset-stage-2": "^6.24.1",
    "react-dom": "^16.10.2",
    "react-hot-loader": "4",
    "ts-loader": "3.5.0",
    "typescript": "^3.7.2",
    "webpack-dev-server": "2.11.2",
    "webpack-merge": "^4.1.2",
    "react": "^16.10.2",
    "slate": "^0.57.1",
    "slate-react": "^0.57.1",
```

My **.babelrc**:

```
{
  "presets": [
    [
      "env",
      {
        "modules": false,
        "targets": {
[...]

Issue #3400, opened 2020-01-02
Can not open slatejs example in Edge browser


wenyuyangpku:
Can not open [slatejs example](https://www.slatejs.org/examples/richtext) in Edge browser.

![image](https://user-images.githubusercontent.com/47671625/71650038-6e6b2200-2ce1-11ea-9a2a-450bc9ebe200.png)

Killavus:
This is because Slate-React is being bundled with target TSconfig set to `esnext`. You may want to change that to `ES6` - this way object spread in `Editable` component will get transpiled properly.

Editor is still kind of broken after that, though. I'm looking for polyfills necessary to make it usable for now.

DamareYoh:
does it work on the chromium based edge? I suspect the legacy edge is unlikely to be supported out of the box

stephenjelfs:
@DamareYoh You're quick: I think the "legacy" non-chromium Edge has only been "legacy" for a couple of hours! It would still be nice to get this fixed, if possible, for people who like to support at least the latest two versions of Chrome, Firefox and Edge :)

tylde:
It is possible to change in tsconfig from `esnext` to `es6`?

PR #3474, opened 2020-01-30
Enable processing of tsx files with babel in rollup.config.js

kamilmielnik:

Fixes https://github.com/ianstormtaylor/slate/issues/3471
Fixes https://github.com/ianstormtaylor/slate/issues/3400

#### What's the new behavior?

No new behavior. This is a fix in babel setup in `rollup.config.js`.

[Experimental JS "Rest in objects" feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) in `packages/slate-react/dist/index.es.js` & `packages/slate-react/dist/index.js` becomes a valid JS that works in older browsers.

Example:

```js
    const { autoFocus, decorate = defaultDecorate, onDOMBeforeInput: propsOnDOMBeforeInput, placeholder, readOnly = false, renderElement, renderLeaf, style = {}, as: Component = 'div', ...attributes } = props;
```

becomes:
```js
var {
    autoFocus,
    decorate = defaultDecorate,
    onDOMBeforeInput: propsOnDOMBeforeInput,
    placeholder,
    readOnly = false,
    renderElement,
    renderLeaf,
    style = {},
    as: Component = 'div'
  } = props,
      attributes = _objectWithoutProperties(props, ["autoFocus", "decorate", "onDOMBeforeInput", "placeholder", "readOnly", "renderElement", "renderLeaf", "style", "as"]);
```

I have also checked it in MS Edge 18 :heavy_check_mark: 

Unfortunately IE 11 still does not work. I'm getting a syntax error (arrow function).
**Is there a plan to support IE 11?**

Fixes: https://github.com/ianstormtaylor/slate/issues/3471, https://github.com/ianstormtaylor/slate/issues/3400
Reviewers: @ianstormtaylor

### DIFF
--- a/config/rollup/rollup.config.js
+++ b/config/rollup/rollup.config.js
@@ -89,7 +89,7 @@ function configure(pkg, env, target) {
     babel({
       runtimeHelpers: true,
       include: [`packages/${pkg.name}/src/**`],
-      extensions: ['.js', '.ts'],
+      extensions: ['.js', '.ts', '.tsx'],
       presets: [
         '@babel/preset-typescript',
         [


e1himself:
Ping @ianstormtaylor.

We need this to continue development of Slate-based editor inside our application.

eslachance:
@kamilmielnik I attempted to checkout your branch, npm link it, and it does not seem to resolve the issue described in https://github.com/ianstormtaylor/slate/issues/3471 , or at least it doesn't for me. Is there anything I might be missing? Tried on all of node 10.16, 12.16 LTS and 13.8 Current.

kamilmielnik:
@eslachance you have to rebuild slate (`npm run build`).

eslachance:
> @eslachance you have to rebuild slate (`npm run build`).

I pulled your branch, I run `npm run build`, then `npm link`, and in my project I ran `npm link slate` and `npm link slate-react`, and I'm still getting the issue. 
image

kamilmielnik:
@eslachance 
In the PR description, in "How does this change work?" section, you can find an example code 
```jsx
     const { autoFocus, decorate = defaultDecorate, onDOMBeforeInput: propsOnDOMBeforeInput, placeholder, readOnly = false, renderElement, renderLeaf, style = {}, as: Component = 'div', ...attributes } = props;
```
and what should happen to it, i.e.
```jsx
var {
    autoFocus,
    decorate = defaultDecorate,
    onDOMBeforeInput: propsOnDOMBeforeInput,
    placeholder,
    readOnly = false,
    renderElement,
    renderLeaf,
    style = {},
    as: Component = 'div'
  } = props,
      attributes = _objectWithoutProperties(props, ["autoFocus", "decorate", "onDOMBeforeInput", "placeholder", "readOnly", "renderElement", "renderLeaf", "style", "as"]);
```

1. Can you see that change when you debug slate code in your app?
2. What error are you getting?

eslachance:
I'm not sure what code you'd like to see, I've simply followed the tutorial. What you're showing is internal slate code, would I need to fix that code for slate to work? Is there a PR to also correct that problem? I'm confused, here.

RobMayer:
Forked it, rebuilt it with the tsx added - can confirm now working in edge and passes tests.

CameronAckermanSEL:
This looks good to me. Thanks for fixing this configuration issue. <3

kamilmielnik:
@CameronAckermanSEL thanks!

Any chance of releasing this soon?
It's blocking us at [Prezly](https://www.prezly.com/) from upgrading slate to 0.50+, as number of our customers use Edge.

jarandmi:
Have anyone managed to install Slate from master with Yarn? As long as this is in master i thought it would be possible to install it, but maybe not since the `/dist` files is not checked in. If not, do we need to wait for a new versio from master before this PR can be used?

kamilmielnik:
@jarandmi 

> If not, do we need to wait for a new versio from master before this PR can be used?

It appears that `/dist` is only included in the npm/yarn package, so yes, we need to wait for a new release. The easiest workaround I can think of is to fork it and release as your own package :man_shrugging:

jungRoit:
when will this be released? it is currently a blocker to use the package

naortor:
Hi,

are there any plans on releasing it any time soon?
I would like to start using Slate, but my webpack build fails.

thanks

eslachance:
So, I commented higher up on this, and feel like anyone wanting to use slate right now might run into similar issues. The way we resolved it here was that we're including the path of slate in our webpack.config.babel.js file (our setup is complicated, perhaps this file is named differently, usually). Basically this bit: 

```js
        rules: [
            {
                test: /\.js$/,
                loader: 'babel-loader',
                include: [SRC_PATH, CONFIG_PATH, LIB_PATH, path.resolve(__dirname, './node_modules', 'slate-react/dist')],
                options: {
                    cacheDirectory: DEBUG,
                },
            },
```

Perhaps this can help others having the same issue right now.

jdmoody:
It looks like the Slate npm package provides a few different files to support different js formats. It looks like these files are enumerated in the package.json:
https://github.com/ianstormtaylor/slate/blob/60b56651b2e1e6f58bcf29a1a876612720a4fe2c/packages/slate/package.json#L7-L11

The `module` format maps to the `index.es.js` file. When I install `slate` locally, this is the only file that uses the rest operator. For some reason, my Webpack bundles were choosing the `index.es.js` file over the `index.js` file...

---

The problem seems to be that, [by default, Webpack chooses the `module` format over the `main` format when resolving node modules](https://webpack.js.org/configuration/resolve/#resolvemainfields) (I think [Rollup does, too](https://github.com/rollup/plugins/tree/master/packages/node-resolve#mainfields)?). From their docs:
> When the target property is set to webworker, web, or left unspecified:
> 
> webpack.config.js
> ```
> module.exports = {
>   resolve: {
>     mainFields: ['browser', 'module', 'main']
>   }
> };
> ```

You can also see that the `browser` format is preferred over `module`. I think if we add a `browser` field to each slate packages' package.json, it could solve this issue by using the properly transpiled index.js?

I'm happy to put up a PR if this sounds like the correct solution.

@eslachance this might help resolve your issue without the need for your Webpack build to transpile Slate. Given that Slate is meant for the browser, I think it would be nice to have syntax that is supported by IE11 so any scripts that include slate can be at least be _parsed_. Polyfills are a different issue and I think it makes sense that they should be provided by the user as described in the [FAQ](https://github.com/ianstormtaylor/slate/blob/main/docs/general/faq.md#what-browsers-and-devices-does-slate-support).

PR #3475, opened 2020-01-30
Change lernajs to correct URL

huttunenMarco:
The URL to the lerna website in README.md is not working, I noticed they changed it to a new one and updated the readme


CameronAckermanSEL:
Thanks for correcting this! It was probably really confusing

PR #3483, opened 2020-02-04
Avoid raising exception `Cannot resolve a Slate point from DOM point...` in function react-editor.ts::toSlatePoint for some kind of Elements.

einsxiao:

fixing a _bug_

For some self-defined Element, the dom-point (by clicking) may not be the leaf-node, in such case, the slate will raise Exception of  `Cannot resolve a Slate point from DOM point...` in function toSlatePoint in react-editor.ts. 

#### What's the new behavior?

The exception won't raise.

The fix add a condition to return the start point of the node to avoid exception raise.

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -149,15 +149,17 @@ export const ReactEditor = {
     options: { editable?: boolean } = {}
   ): boolean {
     const { editable = false } = options
-    const el = ReactEditor.toDOMNode(editor, editor)
-    let element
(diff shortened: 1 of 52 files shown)


packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -406,6 +406,13 @@ export const ReactEditor = {
         textNode = leafNode.closest('[data-slate-node="text"]')!
         domNode = leafNode
         offset = domNode.textContent!.length
+      }else{
```
CameronAckermanSEL:
```suggestion
      } else {
```

PR #3484, opened 2020-02-05
Fix test folders name swap for interfaces/Editor/nodes

oguzgelal:

I realized the folder names for the tests in `interfaces/Editor/nodes/mode-all` and `interfaces/Editor/nodes/mode-highest` were swapped (ie. tests for mode-all was in mode-highest and vice versa). So I fixed that.

#### What's the new behavior?

N/A

N/A

N/A

N/A


CameronAckermanSEL:
Thanks a ton for the contribution @oguzgelal ! Great catch

PR #3491, opened 2020-02-10
Allow zero as offset on <anchor> and <focus> inside <selection>

aglundahl:

Fixing a bug.

#### What's the new behavior?

This currently fails on `master`:

```jsx
<selection>
  <anchor offset={0} path={[0]} />
  <focus offset={0} path={[0]} />
</selection>
```

The reason is because the guards for checking the truthiness of offset in `createSelection` look like this:

```js
if (!anchor || !anchor.offset || !anchor.path) {
```

This won't work for zero, since that's falsy in JavaScript. This PR fixes this.

I check if the offset if falsy and explicitly not zero. I wanted to use something like `Number.isFinite` instead, but unfortunately that doesn't work for a `number | undefined` union.

### DIFF
--- a/packages/slate-hyperscript/src/creators.ts
+++ b/packages/slate-hyperscript/src/creators.ts
@@ -152,13 +152,13 @@ export function createSelection(
   const anchor: AnchorToken = children.find(c => c instanceof AnchorToken)
   const focus: FocusToken = children.find(c => c instanceof FocusToken)
 
-  if (!anchor || !anchor.offset || !anchor.path) {
+  if (!anchor || (!anchor.offset && anchor.offset !== 0) || !anchor.path) {
     throw new Error(
       `The <selection> hyperscript tag must have an <anchor> tag as a child with \`path\` and \`offset\` attributes defined.`
     )
   }
 
-  if (!focus || !focus.offset || !focus.path) {
+  if (!focus || (!focus.offset && focus.offset !== 0) || !focus.path) {
     throw new Error(
       `The <selection> hyperscript tag must have a <focus> tag as a child with \`path\` and \`offset\` attributes defined.`
     )
--- a/packages/slate-hyperscript/src/creators.ts
+++ b/packages/slate-hyperscript/src/creators.ts
@@ -152,13 +152,13 @@ export function createSelection(
   const anchor: AnchorToken = children.find(c => c instanceof AnchorToken)
   const focus: FocusToken = children.find(c => c instanceof FocusToken)
 
-  if (!anchor || (!anchor.offset && anchor.offset !== 0) || !anchor.path) {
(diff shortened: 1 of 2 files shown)


packages/slate-hyperscript/src/creators.ts
```diff
@@ -152,13 +152,13 @@ export function createSelection(
   const anchor: AnchorToken = children.find(c => c instanceof AnchorToken)
   const focus: FocusToken = children.find(c => c instanceof FocusToken)
 
-  if (!anchor || !anchor.offset || !anchor.path) {
+  if (!anchor || (!anchor.offset && anchor.offset !== 0) || !anchor.path) {
```
skogsmaskin: Could this be written like this, to make it a little more readable?

`if (!anchor || !Number.isInteger(anchor.offset) || !anchor.path)`

aglundahl: I would also prefer that, but as I mentioned in the description `Number.isFinite` (and `Number.isInteger`) doesn't accept a `number | undefined` union.

sgurenkov:
```suggestion
  if (!anchor || anchor.offset === undefined || anchor.path === undefined) {
```

packages/slate-hyperscript/src/creators.ts
```diff
@@ -152,13 +152,13 @@ export function createSelection(
...
     throw new Error(
       `The <selection> hyperscript tag must have an <anchor> tag as a child with \`path\` and \`offset\` attributes defined.`
     )
   }
 
-  if (!focus || !focus.offset || !focus.path) {
+  if (!focus || (!focus.offset && focus.offset !== 0) || !focus.path) {
```
sgurenkov:
```suggestion
  if (!focus || focus.offset === undefined || focus.path === undefined) {
```

packages/slate-hyperscript/src/creators.ts
aglundahl: Thanks for the suggestion, @sgurenkov! What I don't like about explicitly checking equality with `undefined` is that it doesn't include `null`. Which is fine in this case as long as you're in TypeScript-land, but not so much for JS. However, the chances for a `null` value slipping through to this point seem pretty slim.

Personally, I don't really care, though. Since you are both Slate contributors I'll leave the decision to you. :)

skogsmaskin: `anchor.offset == null` is probably better. It will return false for `0` and true for `null` and `undefined`.

aglundahl: Neat! Fixed in 4406a26. Thanks!

ianstormtaylor:
Looks great, thanks @aglundahl!

Issue #3500, opened 2020-02-14
<input /> onChange is not triggered in Firefox


daldoy:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
input elements rendered in renderElement don't work in Firefox: onChange is not triggered.

https://codesandbox.io/s/slate-reproductions-b5363

Slate: 0.57.1
Browser: Firefox

CameronAckermanSEL:
I believe this is fixed from merging https://github.com/ianstormtaylor/slate/pull/3436

PR #3501, opened 2020-02-14
fix: <input /> onChange is not triggered in Firefox #3500

daldoy:
Fixing a bug.

Now Editable.onBeforeInput only calls event.preventDefault() when the editor is focused.


CameronAckermanSEL:
Hey @daldoy , I merged another PR that fixed this problem from a couple of months ago before I noticed this one. I'm really sorry about that. But thank you very much for the PR nonetheless. Please feel free to reopen a PR if the issue is still unresolved.

Issue #3502, opened 2020-02-15
Enforce Layout demo is broken


adamsanderson:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
1. Visit the demo page: https://www.slatejs.org/examples/forced-layout
2. Place the cursor at the start of the paragraph node, 
3. Hit delete

You will see the following exception:

```
[example].js:1 Uncaught ReferenceError: Transforms is not defined
    at Object.a.normalizeNode ([example].js:1)
    at [example].js:1
    at Object.withoutNormalizing ([example].js:1)
    at Object.normalize ([example].js:1)
    at Object.withoutNormalizing ([example].js:1)
    at Object.delete ([example].js:1)
    at Object.deleteBackward ([example].js:1)
    at Object.deleteBackward ([example].js:1)
    at HTMLDivElement.<anonymous> ([example].js:1)
```

Slate: 0.XX.X (master)
Browser: Chrome
OS: Mac

#### What's the expected behavior?

Oh golly, I think it's supposed to prevent deleting the paragraph, but I don't know for sure ;)
[It looks like `Transforms` just isn't imported](https://github.com/ianstormtaylor/slate/blob/master/site/examples/forced-layout.js), but I haven't looked into it beyond that.

Kala-kala:
How can we import transforms? It is not there in the slate.js official documentation also.

PR #3503, opened 2020-02-15
Adds Transforms import to forced-layout demo

adamsanderson:
The forced layout demo included Editor, but not Transforms, and would crash when it needed to enforce the layout.

Fixing a bug

#### What's the new behavior?
Fixes: #3502

This adds the missing imports to the demo.

- [X] The relevant examples still work. (Run examples with `yarn start`.) Oh yes indeed!

Fixes: #3502
Reviewers: @


CameronAckermanSEL:
Thanks so much for the contribution @adamsanderson !

Issue #3380, opened 2019-12-23
Last instance steals focus when multiple instances are present


marcandrews:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

In my use case, I have two instances present, one for making edits, and another for previewing the result. 

When multiple instances are present, and you make edits to any instance but the last in the DOM, the last instance steals focus. It then becomes impossible to focus any instance but the last.

I know this issue was fixed before, but perhaps it resurfaced post-`0.50.0`. 

![output2](https://user-images.githubusercontent.com/3518829/71366223-6ab3fe80-256f-11ea-9618-03a7db1fddec.gif)

Codesandbox: https://codesandbox.io/s/spike-slate-u8zpz

Slate: 0.57.1
Browser: Chrome
OS: Mac

#### What's the expected behavior?

When multiple instances are present, the user should be able to interact and edit an instance without another instance stealing focus.

#### Are there any potential workarounds?

In my use case, I only require one instance to be editable, so ensuring that this instance is last in the DOM is working for me.

marcandrews:
@ianstormtaylor is the potential fix for this in `slate` or `slate-react`?

am1rb:
Hi, I have the same issue, any update?

jaredpalmer:
Same issue. Can't use current release

Haaxor1689:
We have a bit different issue. Again 2 different instances of editor but the problem is that when clicking focus directly from one to another both seem to loose focus.

am1rb:
Hey @Haaxor1689, I checked PR #3506 again and it seems that everything is fine in this branch.
I can not reproduce your issue, would you please put a sample code in [codepen.io](https://codepen.io/) or similar websites?
![check-focus-again](https://user-images.githubusercontent.com/573937/75096701-5d3ee300-55b7-11ea-9d9a-9e86ad87167a.gif)

CameronAckermanSEL:
I'm pretty sure this bug is specifically caused by reusing the same value in two different editor components. The code sandbox provided shows (maybe) a reasonable case where you'd want to do this but I'm suspecting there's a pretty good chance that this is avoidable in a lot of other use cases where the value does not need to be shared between the editor components.

Double check that each editor is initialized with a fresh value object as opposed to having a single const for it.

matteofrana:
Unfortunately for this problem I have to stick to 0.47 for [React Bricks CMS](https://www.reactbricks.com).

[...]

Issue #3497, opened 2020-02-13
Editor loses focus if parent component triggers unrelated state change


skokenes:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
I have a component that contains a slate editor and another input object which has its own state being maintained with a React useState hook.

When I press enter in my input box, I trigger a state update. I also try to focus the slate editor using `ReactEditor.focus(editor)`. However, the focus gets removed from editor immediately. If I comment out the state update, the focus goes to the editor properly.

![slate-focus](https://user-images.githubusercontent.com/5554373/74393179-4b9b5580-4dd7-11ea-9c94-ea632bfc5d06.gif)
https://codesandbox.io/s/intelligent-field-o2y8u?fontsize=14&hidenavigation=1&theme=dark

OSX, Chrome, Slate 0.57.1

Slate: 0.XX.X
Browser: Chrome / Safari / Firefox / Edge
OS: Mac / Windows / Linux / iOS / Android

#### What's the expected behavior?
The expected behavior is that when imperatively setting focus on the editor, the focus should be maintained even if an unrelated state update happens in the component.

debuggermalhotra:
Hey! any updates on this ? I think I am facing an identical bug. The focus tends to shift and the element gets added at the lest of the text when i insert and Inline Node.

tiberiuichim:
The problem is caused by the initialValue that's being used in both editors. Use cloneDeep to wrap it, it will work properly after that.

sedatbasar:
Same here. I have two different editors in same page. I am editing 1st one, when the 2nd one state changes and get rerendered, 1st one loses the focus. Actually ReactEditor.isFocused still returns true but cursor is not there and not allowed to type anything.

cloneDeep didnt work for me which @tiberiuichim suggested btw.

vangelov:
Any news on this? I'm faciting the same issue when there are two editors on the users switches by clicking Tab.

PR #3506, opened 2020-02-17
Fix the stealing focus problem when multi instances are present

am1rb:
Fixing a bug

#### What's the new behavior?
Fixes: #3380 

The 'isRangeEqual' method does not work correctly and returns false most of the time. This causes the editor to remove and add all the selection ranges again on every render of the Editable component. 

Also, the above behavior causes some problems when multiple instances are present and the last instance steals focus.

**Before:**
before-bugfix-optimized

**After:**
after-bugfix-optimized

Fixes: #3380

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -191,7 +191,7 @@ export const Editable = (props: EditableProps) => {
         ) {
           return
         }
-      } catch {
+      } catch (e) {
         return
       }
     }
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -175,11 +175,12 @@ export const Editable = (props: EditableProps) => {
       return
     }
 
-    const newDomRange = selection && ReactEditor.toDOMRange(editor, selection)
+    let newDomRange: DOMRange | null = null
+    try {
+      newDomRange = selection && ReactEditor.toDOMRange(editor, selection)
 
-    // If the DOM selection is already correct, we're done.
-    if (hasDomSelection) {
-      try {
+      // If the DOM selection is already correct, we're done.
+      if (hasDomSelection) {
         const newEditorRange = ReactEditor.toSlateRange(
           editor,
           domSelection.getRangeAt(0)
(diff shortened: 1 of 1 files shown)


ryanmitts:
This appears to also fix #3497.

greylemon:
I'm facing another issue after I merged this PR on a clean fork. When using left and right arrow keys (up and down works), the selection blinker goes to the last instance. I will check it out

greylemon:
https://github.com/ianstormtaylor/slate/issues/3380#issuecomment-591261284 

Fixed my issue (Thanks!) Though, I still needed this PR for the correct selection

greylemon:
https://github.com/ianstormtaylor/slate/issues/3380#issuecomment-591261284

Fixed my issue (Thanks!). On 0.57.1 and this PR, it seems there are selection issues from right to left

am1rb:
@greylemon oops! I will check the arrow keys issue on my free time

Haaxor1689:
@greylemon @am1rb any updates on this MR?

rohmanhm:
Any workaround to fix this issue?

cc @ianstormtaylor

CameronAckermanSEL:
Hi @am1rb, thanks for this contribution. I see theres an outstanding issue so I’ll hold off on review until an update.

Personally, my opinion is that this behavior is a bit on the “working as intended side” and instead we should attack this with better documentation. This issue is caused by a user sending the same value to multiple editor components https://github.com/ianstormtaylor/slate/issues/3380#issuecomment-612432934. Because the editor components use the referential address of a given block to map to DOM nodes, this leads to selection between two editors to be handled oddly if there are multiple using the same initial values, and the fix is a matter of generating a value for each editor separately.

I think this is a bit counter intuitive for new users but is ultimately a good thing for the editor to require its own separate value object. And we can add a little guide for how to have multiple editors instead.

What are your thoughts?

hanford:
@CameronAckermanSEL

I think that would be a great start! It was really confusing to me at first, as my default value was outside of component scope .. which triggers this problem

```js
const defaultValue = [..]


const Editor (p) => {
	const [value, setValue = React.useState(defaultValue)
	..
}
```

Once I changed that, everything was working 100%.

CameronAckermanSEL:
Hey @am1rb , thanks a lot for this contribution. At this time, I think we'll decline this PR because this seems to fall more under user error that should be addressed with documentation rather than misoperation of the editor. It is (likely) by design that each editor instance should have its own separate value and lately the circulation of this advice seems to have alleviated these problems for most people.

This certainly isn't a final decision or anything! If you feel like we should reconsider having this merged in, please feel free to reach out! 

Appreciate your understanding and your donation of time. Thank you.

Issue #3507, opened 2020-02-19
All nodes are re-rendered in a controlled value update


pzhine:
#### Do you want to request a _feature_ or report a _bug_?

Request a feature

#### What's the current behavior?

1. An `onKeyDown` event on the `<Editable />` component triggers a mutation on a cloned Slate `value`.
2. On the next render, we pass the mutated `value` to the `<Slate />` component.
3. The render method is called for each node, and each node has a new React `key` prop, which forces a DOM update.

```js
const onKeyDown = event => {
    event.preventDefault()
    const _value = JSON.parse(JSON.stringify(value))
    _value[0].children[0].text += event.key
    setValue(_value)
};
```

![slate-renderer](https://user-images.githubusercontent.com/2692844/74887160-73139480-537a-11ea-9693-ea248ad13df0.gif)

https://jsfiddle.net/75ezr96w/2/

Slate: 0.57.1
Browser: Chrome
OS: Mac

#### What's the expected behavior?

For even moderately large documents, re-rendering and updating the DOM for all nodes on every `value` mutation can be much too slow. Slate 0.5x tries to optimize the re-renders by memoizing the render callback based on object equality. That is, if the pre-update node is the same object as the post-update node, it skips the post-update render call. However, in a controlled component, we would like to clone our editor state before updating to enforce immutability, so this memoization check fails to optimize.  

Similarly, Slate 0.5x tries to optimize DOM updates by generating the `key` prop using a weak map where the node objects are the keys, so as long as the pre-update and post-update nodes are object-equivalent, the `key` prop stays the same and the DOM is left unchanged. However, in our controlled component with immutable `value`s, the `key`s will be updated on all the nodes every time we mutate a node.

#### Suggested feature

We propose that we follow React's pattern for controlling DOM node updates, which allows us to explicitly pass a `key` prop to our components. If we keep this `key` prop constant, React will only update the corresponding DOM node if the component's contents have changed. On the other hand, if we change the `key` prop, React will re-mount our component and force an update on the DOM node.  

So, we could optionally override the `key` prop on our components through the `value`, like so:

```js
const initialValue = [{
  type: 'paragraph',
  key: 0,
  children: [{
    text: `line ${idx}`
   }]
}
```

Keeping that `key` constant would tell Slate to skip this node in the next render, so now it is the controlling application's responsibility to update the `key` whenever the contents of that node change. If the `key` field is left out of a `value` node, Slate generates it based on the current weak map implementation.

[...]

Issue #3489, opened 2020-02-09
Unique identifier for Nodes


alfechner:
#### Do you want to request a _feature_ or report a _bug_?

Request a feature.

#### What's the current behavior?

I migrate my implementation from a older version of Slate to the current one. The API has been rewritten and improved.

I've got a use case in my old case witch doesn't seem to be covered anymore. Previously Nodes had unique keys assigned which could be used in i.e. `replaceNodeByKey`. My code heavily relies on this feature. Actions taken in an Element change it's definition.

For example: An Image Element which holds an URL. I need a edit button to change the URL to another one. So far I replaced the Node's definition by the new one using the method mentioned above.

I asked about that in the Slack channel. People proposed to set the new Node by path using `setNodes`. But nobody knew how to get the path of the Node without having some kind of unique identifier.

If this kind of identifier is really needed to perform such actions I'd love to see it back in the library.

CameronAckermanSEL:
Hey @alfechner , the API was rewritten to effectively not rely on keys to locate nodes anymore, instead this is done with paths. This alleviates the need for the API to recursively search the node hierarchy to locate a node by its key and is a performance improvement. I know it's a bit difficult to transition to the new paradigm, but I'd be surprised to see keys come back as a first class way of accessing nodes in response to this feature.

If you would like to bind keys to your nodes, the API allows this because you can freely add arbitrary properties to your nodes as you see fit (as long as you don't run into name collision with properties on the core interface). You can use this to write your own locate by key utilities and then that should yield you a path. If you like.

alfechner:
@CameronAckermanSEL thanks for clarifying.

alfechner:
A good example on how to do this can be found in the [embedded](https://github.com/ianstormtaylor/slate/blob/master/site/examples/embeds.js) example.

```js
 onChange={value => {
    const path = ReactEditor.findPath(editor, element)
    Transforms.setNodes(editor, { url: value }, { at: path })
}}
```

raitucarp:
I've tried assign ID to nodes when normalizing, but it seems I catch bug when applying Transforms.setNodes .. 

https://github.com/ianstormtaylor/slate/issues/4641

PR #3508, opened 2020-02-20
Add node key override to allow performant immutable controlled updates

pzhine:

Improving a feature.

#### What's the new behavior?

We propose that we follow React's pattern for controlling DOM node updates, which allows us to explicitly pass a `key` prop to our components. If we keep this `key` prop constant, React will only update the corresponding DOM node if the component's contents have changed. On the other hand, if we change the `key` prop, React will re-mount our component and force an update on the DOM node.  

So, we could optionally override the `key` prop on our components through the `value`, like so:

```js
const initialValue = [{
  type: 'paragraph',
  key: 0,
  children: [{
    text: `line ${idx}`
   }]
}
```

Keeping that `key` constant would tell Slate to skip this node in the next render, so now it is the controlling application's responsibility to update the `key` whenever the contents of that node change. If the `key` field is left out of a `value` node, Slate generates it based on the current weak map implementation.

The following screencast and snippet are from `site/examples/controlled-keys.js`, a new example that demonstrates the key overriding with controlled `value`s.

slate-override

```js
const onKeyDown = event => {
    if (event.key.length > 1) {
      return
    }
    event.preventDefault()
    const { path, offset } = editor.selection.focus
    const _value = JSON.parse(JSON.stringify(value))
    const _textNode = _value[path[0]].children[0]
    _textNode.text =
      _textNode.text.slice(0, offset) + event.key + _textNode.text.slice(offset)
    _value[path[0]].key.id += '1'
    setValue(_value)

    const _selection = JSON.parse(JSON.stringify(editor.selection))
    _selection.focus.offset += 1
    _selection.anchor.offset += 1
    setSelection(_selection)
  }
```

To implement this, we had to refactor the `KEY_TO_ELEMENT` from a WeakMap to a Map, updating methods that use this map to index using the string `key.id` instead of the whole `key` object. 

Fixes: #3507

### DIFF
--- a/packages/slate-react/src/components/children.tsx
+++ b/packages/slate-react/src/components/children.tsx
@@ -5,7 +5,7 @@ import ElementComponent from './element'
 import TextComponent from './text'
 import { ReactEditor } from '..'
 import { useEditor } from '../hooks/use-editor'
-import { NODE_TO_INDEX, NODE_TO_PARENT } from '../utils/weak-maps'
+import { NODE_TO_INDEX, NODE_TO_PARENT } from '../utils/maps'
 import { RenderElementProps, RenderLeafProps } from './editable'
 
 /**
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -34,7 +34,7 @@ import {
   NODE_TO_ELEMENT,
   IS_FOCUSED,
   PLACEHOLDER_SYMBOL,
-} from '../utils/weak-maps'
+} from '../utils/maps'
 
 /**
  * `RenderElementProps` are passed to the `renderElement` handler.
--- a/packages/slate-react/src/components/element.tsx
+++ b/packages/slate-react/src/components/element.tsx
@@ -134,7 +134,8 @@ const MemoizedElement = React.memo(Element, (prev, next) => {
   return (
     prev.decorate === next.decorate &&
     (prev.element === next.element ||
-      (prev.element.key && next.element.key &&
+      (prev.element.key &&
+        next.element.key &&
         prev.element.key.id === next.element.key.id)) &&
     prev.renderElement === next.renderElement &&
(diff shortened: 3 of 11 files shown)


alfechner:
The editor will not specify any key for you, right? This has to be done manually. 

I guess the old API (<0.5) had a key concept in place which was used to identify nodes. I don't know if it served any render optimization, though.

I opened #3489 regarding a unique identifier for nodes (like the key in the previous API). For my use case I'd like it to be created automatically, though. Is this covered by your PR?

CameronAckermanSEL:
Hi @alfechner , if I understand this PR, then the feature you are requesting will not be covered. 

Slate no longer, by design, exposes an automatically generated key as a way of uniquely identifying nodes. And I'd honestly be a little surprised if a core utility was added to the library to do that. I'd instead suggest exploring a plugin that did this for you as nodes are split or inserted. Nodes can accept arbitrary properties today, including a `key`, so this the model should be conducive to it, even if it ignores that the keys exist at all.

pzhine:
I am closing this because the solution moves in the wrong direction. Changing `KEY_TO_ELEMENT` from a WeakMap to a Map will introduce a memory leak because unused keys and their associated node values will not get garbage-collected.

A better solution is in #3509 , which uses the Plugin pattern to solve this problem while keeping the memory efficiency of the WeakMap.

pzhine:
@alfechner The editor does assign keys to each node internally. See the notes for #3509 for an explanation of this architecture. If you need functionality equivalent to `replaceNodeByKey`, you will have to maintain your own map of keys to nodes in your application write your own `replaceNodeByKey` method. 

However, the bigger issue is that `setNodes` is architecturally different than `replaceNodesByKey` in that it only sets properties on a node and _cannot_ set the `text` property. To solve this, we're controlling the `value` that is passed to Slate, which allows us to replace any or all of the node. However, the drawback is that it forces a re-render until #3509 is merged, which will allow persisting keys across immutable values.

PR #3509, opened 2020-02-21
Make ReactEditor.findKey overridable using the plugin pattern

pzhine:

Improving a feature.

#### What's the new behavior?

All nodes have a `key` that is generated automatically by the [findKey](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugin/react-editor.ts#L38) method and used internally to map Slate nodes to DOM nodes. It is also used as the `key` prop on the React element that contains your node's children. 

The `key` is stored in a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap), where the map key is the node object from the Editor `value`. This means that the `key` is stable as long as the instance of the node object in `value` remains unchanged.

However, if we're trying to use Slate as a controlled input, we run into performance problems if managing our state using an [immutable update pattern](https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns/) like Redux. These patterns return a mutated deep clone as the new state, which means our underlying node object instances would always change after an update even if the node data remained constant. This would lead to a lot of unnecessary re-rendering and re-mounting in the DOM, because all of our node keys would change on every render. 

As a solution, we propose allowing developers to override the `editor.findKey` method using the  [plugins](https://docs.slatejs.org/concepts/07-plugins) pattern. For example, let's say we wanted to use an `id` field on the node as our `key`. The `editor.findKey` interface expects an object as a return value, so we would write an HOC like `withKeys` in the example below that maps string IDs to key objects.

```js
import React, { useRef } from 'react'

const nodeKeysRef = useRef({})

const withKeys = editor => {
  const { findKey } = editor
  editor.findKey = element => {
    const { id } = element
    if (!nodeKeysRef.current[id]) {
      nodeKeysRef.current[id] = findKey(element)
    }
    return nodeKeysRef.current[id]
  }
  return editor
}
const editor = withKeys(createEditor())
```

Note that we are now responsible for maintaining our map between `id` strings and `key` objects. For memory efficiency, we should remove the key from our map when we remove a node so that it gets garbage collected correctly. In the example above, this would be achieved by calling `delete` on that map entry, as in:

```js
const removeKey = id => {
  delete nodeKeysRef[id]
}
```

Fixes: #3507
Reviewers: @ianstormtaylor


CameronAckermanSEL:
Thanks for the PR, it looks really well crafted and (in my opinion) it seems like a really nice "granular control" feature to expose. 

However, one thing I wanted to mention is that I think perhaps the workflow of doing a deep clone of the editor state in redux might not be the intended workflow. In the immutable js days, the immutability was managed by immutable js in a pure way that prevented the need to do a deep clone via spread, AND prevented the ability to do a deep clone via spread at all, so the expected usage was to basically dump the immutable js object directly into redux and call it a day. Here, we're getting immutability via immer (see https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/transforms/general.ts), and I'm pretty sure it's also meant to be dropped into redux directly. So I'm not sure whether the deep spread clone into the redux store is a gain or preventable.

I'm interested to hear your thoughts on it. I'm perfectly aware that I'm probably not as well versed in this problem as you are given all the research you did 😆  but I'm hoping to learn a little bit more about how you're approaching this problem.

Thanks again, this looks like really well crafted work and it's appreciated. 

(also, I'm not blocking this PR, I think this is something Ian will have to look at)

JirkaVebr:
I believe that deep-cloning the editor state could and should be avoided. There is, however, another use case that this would happen to help with.

In my editor there are void elements that contain inputs. When the user modifies their values, I need to somehow update the corresponding slate element (e.g. in my `apply` override). In doing so, I can either mutate the object directly, thereby preserving its referential equality, or create a new one. Normally, the correct way would be the latter but unfortunately, neither work in this case.

If I change the referential equality, slate-react generates a new key for it and my void element gets unmounted and rendered again. That in turn causes the inputs to lose focus after each keystroke which completely breaks the UI.

Alternatively, if I do preserve the referential equality, the `prev.element === next.element` check in `element.tsx` that this PR removes causes Slate to bail out of the update which prevents the element from ever re-rendering.

In fact, I believe that the [embed example](https://www.slatejs.org/examples/embeds) suffers from this very issue as well ‒ even if you make the input always visible, it still never updates (or at least it doesn't for me).

**So fundamentally, this is about updating elements that are rendered by stateful components that would preferred to be updated as opposed to remounted.**

There is a hacky user-space workaround:
1. Avoid the remounts by preserving element referential equality.
2. Make a special React context intended for forcing re-renders
3. When rendering each relevant void element, wrap it in a consumer of said context.
4. Modify the context value in each re-render of the whole editor.

Obviously, that doesn't come without a performance hit but it is not the end of the world since the void element can bail out of updates on its own if necessary. Nevertheless, I would say that a better way of dealing with the keys would be nice.

pzhine:
Thank you @CameronAckermanSEL and @JirkaVebr for the useful feedback! I wasn't aware the Slate was using `immer` internally, and, additionally, I hadn't used `immer` in practice in my projects (although it was on the list for a current refactor). 

I created PR #3515 that replicates the "Controlled Markdown Shortcuts" example in this PR but uses `immer` to modify the state. Lo and behold, everything works as expected 🤯🎉. So the `findKey` override is not really necessary to achieve performant controlled updates.

As for the check in `MemoizedElement` for `prev.element === next.element` that is causing problems for @JirkaVebr, I found that removing that check didn't seem to make any difference in the execution or performance. So I'll make a separate PR with that removal and we can move this conversation there.

Even though someone may find the `findKey` override useful, I think this PR should be closed until we find a use case that necessitates it. What do you think?

pzhine:
> In my editor there are void elements that contain inputs. When the user modifies their values, I need to somehow update the corresponding slate element (e.g. in my apply override). 

@JirkaVebr can you post a CodeSandbox that reproduces this setup?

tomconnors:
@JirkaVebr Did you ever find a less hacky workaround? I have a similar situation: my Slate editor contains complex entities that can be selected and modified independently of Slate. Every modification of one of those entities causes a remount of the component in the Slate editor because the element in the slate value is a new object.

Perhaps instead of allowing overrides of `findKey`, Slate could add a new optional property to its schema, `slateKey`, that allows users to provide the key, instead of Slate always basing the key on the object's identity.

PR #3514, opened 2020-02-23
docs: update an example's link

chengluyu:
The "hovering menu" example doesn't exists. Instead, there's an example called "hovering toolbar". So I update the link to that example.

Nope.

#### What's the new behavior?

Nope.

I just fixed an unreachable link.

No.


CameronAckermanSEL:
Thanks so much for catching this issue! Really appreciate the PR :)

PR #3515, opened 2020-02-23
Add controlled updates example

pzhine:

This PR adds an example and a new Walkthrough in the docs.

#### What's the new behavior?

One of the powerful features of Slate is the ability to use it as a controlled input. This means that in our `onChange`, we can mutate the `value` and pass it back to the `<Slate />` component. If we mutate it correctly, Slate will efficiently render the new document. However, we must be cautious. Slate uses [immer.js](https://immerjs.github.io/) to handle state updates internally, which means that `value` is [frozen](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze), e.g. immutable, and the JS runtime will throw an error if you try to mutate it in place. We can deep-clone the `value`, mutate it, and pass it back to Slate, but this is not recommended for reasons covered in the discussion of #3509. 

Instead, the recommended pattern is to use `immer.js` to create a new "draft" of the `value` using `createDraft` and then mutate the draft. When we pass the new `value` returned by `finishDraft` to Slate, everything works as intended. This pattern is described in a new "Controlled Updates" section of the docs site, which uses sample code tested in this [CodeSandbox example](https://codesandbox.io/s/controlled-updates-example-u3v70).

Fixes: #3507

### DIFF
--- a/docs/walkthroughs/XX-controlled-updates.md
+++ b/docs/walkthroughs/XX-controlled-updates.md
@@ -1,0 +1,42 @@
+One of the powerful features of Slate is the ability to use it as a controlled input. This means that in our `onChange`, we can mutate the `value` and pass it back to the `<Slate />` component. 
+
+For example, let's say we have a document in which we always want the first paragraph to be uppercase. In a controlled implementation, we would want something like the code below:
+
+```js
+const App = () => {
+  const [value, setValue] = useState([
+    {
+      children: [{ text: 'This line should be uppercase' }],
+    },
+  ])
+
+  const onChange = value => {
+    value[0].children[0].text = value[0].children[0].text.toUpperCase()
+    setValue(value)
+  }
+
+  return (
+    <Slate editor={editor} value={value} onChange={onChange}>
+      <Editable renderElement={renderElement} autoFocus />
+    </Slate>
+  )
+}
+```
+
+However, if we try this we'll get an error like `Cannot assign to read only property 'text' of object`. This is because the Slate `value` is [frozen](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze), e.g. immutable. 
+
+To correctly mutate our controlled `value`, we use the [immer.js](https://immerjs.github.io/) library, which is what Slate uses internally to manage state updates. We simply use the `createDraft` of immer to create a mutable "draft" of our `value`, make our changes, and then call `finishDraft` on our draft `value` before passing it back to Slate.
+
+```js
+import { createDraft, finishDraft } from 'immer'
+
+const onChange = value => {
+  const _value = createDraft(value)
+  _value[0].children[0].text = _value[0].children[0].text.toUpperCase()
+  setValue(finishDraft(_value))
+}
+```
+
+With the working code, we'll see our first line transformed to `THIS LINE SHOULD BE UPPERCASE` and if you type anything else on that line, it will be converted to uppercase as you type!
+
+You can find a more advanced example of controlled updates in the [Controlled Updates](https://slatejs.org/examples/controlled-updates) example.
--- a/package.json
+++ b/package.json
@@ -84,5 +84,8 @@
     "slate-hyperscript": "*",
     "slate-react": "*",
     "typescript": "^3.7.2"
+  },
+  "dependencies": {
+    "immer": "^5.3.6"
   }
 }
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "eslint-plugin-react": "^7.16.0",
     "faker": "^4.1.0",
     "image-extensions": "^1.1.0",
+    "immer": "^5.3.6",
     "is-hotkey": "^0.1.6",
     "is-url": "^1.2.2",
     "lerna": "^3.19.0",
@@ -84,8 +85,5 @@
     "slate-hyperscript": "*",
     "slate-react": "*",
     "typescript": "^3.7.2"
-  },
-  "dependencies": {
-    "immer": "^5.3.6"
   }
 }
--- a/site/examples/controlled-updates.js
+++ b/site/examples/controlled-updates.js
@@ -18,25 +18,19 @@ const SHORTCUTS = {
   '######': 'heading-six',
 }
 
-let i = 0
-
 const initialValue = [
   {
     type: 'heading-one',
-    id: ++i,
     children: [
       {
-        id: `${i}leaf`,
         text: 'Controlled Update Example',
       },
     ],
   },
   {
     type: 'paragraph',
-    id: ++i,
     children: [
       {
-        id: `${i}leaf`,
         text:
           'This example has the same functionality as the "Markdown Shortcuts" example. However, in our implementation we are using `immer.js` to draft a new state in onChange, mutate it, and render the results.',
       },
@@ -44,10 +38,8 @@ const initialValue = [
   },
   {
     type: 'paragraph',
-    id: ++i,
     children: [
       {
-        id: `${i}leaf`,
         text:
           'Try starting a line with a markdown modifier, like "#", "##", ">" or "-" and then press SPACEBAR to see the formatting applied.',
       },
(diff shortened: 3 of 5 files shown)


CameronAckermanSEL:
Thank you very much for the contribution. I think it is high value to have an example of slate used as a controlled editor, but I think this particular method of handling markdown shortcuts in the onChange conflicts with the other example and I'm unsure if this is the right usage of the onChange event. 

In particular, the thing I'm most wary about is that mutating during an onChange bypasses some of the normalization events (if I'm not mistaken). Generally we recommend strongly that users lean on transformations to make mutations rather than manually mutating the document state via immer. 

It's good for a user to know that immer exists but using it directly borders a bit on antipattern unless there's a specific reason a user needs to mutate the editor value directly, and I can't really think of a good use case for this at this time. 

I'd recommend that we cut out the guts of the onchange and put the value directly in state because this will give users guidance for the recommended controlled usage pattern.

pzhine:
Re: @CameronAckermanSEL, thank you for taking the time to review and comment on this PR. I think this discussion cuts to the heart of a very important and as-yet undocumented area of the 0.5x Slate package.

I purposefully chose to copy the functionality of the _Markdown Shortcuts_ example so we could compare two different methods of accomplishing the same thing in Slate, the first using the plugin and override architecture and the second using a "controlled component" pattern. 

As per the authors of React, an "uncontrolled component" maintains its own state, which must be kept in sync with the application state. This synchronization step adds a lot of side effects, and complexity to our code. Instead of two parallel states, the authors [suggest](https://reactjs.org/docs/forms.html#controlled-components) that

> We can combine the two by making the React state be the “single source of truth”. Then the React component that renders a form also controls what happens in that form on subsequent user input. An input form element whose value is controlled by React in this way is called a “controlled component”.

The _Markdown Shortcuts_ example that uses plugins and overrides uses Slate as an uncontrolled component. That is, it assumes Slate has mutated its own state based on user interaction and  applies a secondary mutation to the Slate state by way of the `Transforms` API. This indeed adds a lot of complexity. Not only is the mutate code more than twice as many lines than the "controlled component" implementation, it forces the user to learn (in addition to the plugin/override API) an entire mutation API, including the specification of 11 new functions: `Range.isCollapsed`, `Editor.above`, `Editor.isBlock`, `Editor.start`, `Editor.string`, `Transforms.select`, `Transforms.delete`, `Transforms.setNodes`,  `Transforms.wrapNodes`, `Point.equals` and `Transforms.unwrapNodes`.

The "controlled component" _Markdown Shortcuts_ example, on the other hand, just uses plain JavaScript string manipulation to compose a new `node.text` value. That is, it treats the `Element` (e.g. `node`) as a controlled component. The only `Transforms` API method we have to learn is `setSelection` to update our cursor once the text has changed. It's a clear use case unto itself: we want, as application developers, to mutate our state (our document, in this case) using operations that are terse and independent of our UI architecture. Then we want to pass that state to our UI architecture, which is Slate in this case, with as little friction and boilerplate as possible. Although there may be cases where it makes sense for the Slate state to be the "single source of truth" and only use the `Transforms` API to mutate it, it is quite restrictive to assume that is the only "correct" way to use Slate and that any other use is an "anti-pattern". I can imagine many applications that would not want their higher-order state management code to rely on a vendor API for mutations. Our application happens to be one of them :)

As for the concern about normalization events, a [quick CodeSandbox](https://codesandbox.io/s/controlled-updates-with-normalize-le3ty) seems to indicate that `editor.normalizeNode` is called as expected in our controlled component example.

slate-controlled-normalize

rockettomatooo:
I might be missinterpreting this but here are my 2 cents:

I understand that the way slate does mutations on its documents is a little confusing and probably won't land on the showcase page for reacts' best-practices. I also agree that the API isn't that obvious at first (that is due to 2 reasons IMO: slate is of very abstract nature and secondly doesn't have an API reference yet). Really, we had cases in slack where people needed to transform a slate document on the server side and we suggested that for simple transformations, they might be better of doing it themselves rather than pulling in slate for this.
I could completely understand your usecase in a server environment and for relatively easy transformations (maybe inserting a node or making a text node uppercase). However, I can't wrap my head around, why one would do this within the client? Can you maybe describe a more complex example on where / why you use this approach in your application?
Beside the API not beeing entirely obvious and lacking a reference I think it would be a real step backwards to advertise manipulating a slate document by hand. Really, the API incorporates a lot of edge-cases and functionality like splitting nodes that everybody would have to implement themselves. Also the API is so abstracted that you should be able to accomplish almost anything with it. I don't think it limits users to something they couldn't achieve with it.

I don't agree on the uncontrolled component part tho. The current implementation of the `<Slate />` component might not land on the showcase page for react best-practices but slate already is completely a controlled component IMO. You pass a `value` to `<Slate />` and if you don't update that state, your editor won't change. That's also the reason you pass an `onChange` function to slate so it can tell you the new state that you have to then pass back down to your component again. For me, this fully qualifies for beeing a completely controlled component.
Maybe to clarify on how slate would look like uncontrolled IMO is, that you could at most pass an `defaultValue` to `<Slate />` which will then completely control its state internally.

CameronAckermanSEL:
I think I'd be more open to an example where we were doing manual mutations through immer if there was a compelling usecase for it. But for now there are a couple of areas I'm reluctant to advertise as "best practice" for slate usage:

1. Mutations in an onChange are, in my opinion, not the best lifecycle. I could see if we were doing mutations in a more specific event handler, like onKeyDown or onPaste, etc. But onChange gets fired extremely liberally and it should be as fast as possible in order to keep the editing experience performant. 

2. Again, mutating the document state manually in lieu of using the available transforms will guide users to a workflow where they are more likely to run into strange edge cases, particularly with respect to browser compatibility issues, as @rockettomatooo pointed out. 

It is really important for these examples that we're careful about what practices we're suggesting to people. Using slate in this controlled workflow is a good example to have, and having documentation on how to deal with immer _if you really must do this_ is also useful, I'm still not convinced that an example that uses immer directly is what we're shooting for.

pzhine:
This is a really productive discussion! I agree that the examples site might not be the best place to show off this type of usage. I wasn't thinking of it as a best practices, but more of a repository of "recipes" that are possible with Slate. I think a link to a CodeSandbox in the issue is sufficient. I also agree that the distinction I was making between “controlled” and “uncontrolled” isn’t all that helpful. The `Transforms` library is really helpful for the kind of string math and object mutations that we need in a text editor. It is, in fact, a full-fledged “reducer” for controlling Slate. What I intended to exemplify was that that there are use cases where you might want to “control” or override these low level operations (e.g. what is actually happening in the `Transforms` functions), and that Slate allows this. You’re both right in that it lets you break things if you do mutate incorrectly, and perhaps that’s the best reason not to include it in the examples site.

If you’re curious, this experiment is derived from the use case of our app, [databyss](https://databyss.org). It’s a relational database editor (e.g. CRUD interface) that uses structured text rather than forms to give writers and researchers a more natural way to take notes and compile sources, chapters, and themes. Like a Notion tailored to higher ed and research. If you've played with Notion (and it's amazing, I suggest you do), you'll notice that there's a ton of two-way binding happening not only in the client (it runs offline) but within the documents themselves, because, in the end, the document is the primary interface and application state. As it’s a kind of compiler, we could have implemented it like a code editor in which the AST is built on a separate thread and updates the UI asynchronously. However, it’s a more of a word-processor (the dreaded WYSIWYG editor) than a code editor, so the AST and UI need to be updated synchronously.  It is _totally possible_ to use the Slate `value` as a first-class data structure in our application code, but we decided to decouple them for a few reasons, including interoperability with native platforms (Android and iOS) that will require different underlying rich text components with their own transform APIs. 

Screenshot 2020-02-25 23 37 03

I hoped that might clear things up, but I worry I may have made things more confusing. You can look at our source at https://github.com/databyss-org/databyss if you’re reaaaaallly curious 😋. 

I’m gonna go ahead and close this PR. I’ll put a link to the CodeSandbox example in the linked issue.

CameronAckermanSEL:
Hey @pzhine , wanted to reiterate again that the contribution is/was appreciated. I think there's some value in having a guide that talks about immer (if only to prevent some of the churn you had to do in response to trying to add immutability on top of it), so if it's alright with you I might dig that guide out of this PR at a later date and put something up.

Thanks again

PR #3516, opened 2020-02-26
fixed some bugs with the input event handler on the embeds example

CameronAckermanSEL:

Bug

#### What's the new behavior?

This PR simplifies some selection based conditional rendering that was making it hard to edit the embedded input. 

* Removed conditionals that were rendering the input only when the VideoEmbed element was selected and focused. 
* Removed the div that was providing an outline to the VideoEmbed element - this was disrupting the input's select state.

Fixes: #
Reviewers: @

### DIFF
--- a/site/examples/embeds.js
+++ b/site/examples/embeds.js
@@ -1,5 +1,5 @@
 import React, { useState, useMemo } from 'react'
-import { Editor, createEditor } from 'slate'
+import { Transforms, createEditor } from 'slate'
 import {
   Slate,
   Editable,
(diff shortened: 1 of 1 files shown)


site/examples/embeds.js
```diff
@@ -83,20 +65,18 @@ const VideoElement = ({ attributes, children, element }) => {
...
+          value={url}
+          onClick={e => e.stopPropagation()}
+          style={{
+            marginTop: '5px',
+            boxSizing: 'border-box',
+          }}
+          onChange={e => {
+            const path = ReactEditor.findPath(editor, element)
+            Transforms.setNodes(editor, { url: e.target.value }, { at: path })
+          }}
```
CameronAckermanSEL: the event handler was quite busted because it was expecting a value string but receiving an input event, and Transforms wasn't being imported.

site/examples/embeds.js
```diff
@@ -41,30 +41,12 @@ const Element = props => {
 
 const VideoElement = ({ attributes, children, element }) => {
   const editor = useEditor()
-  const selected = useSelected()
-  const focused = useFocused()
   const { url } = element
   return (
     <div {...attributes}>
       <div
         contentEditable={false}
```
CameronAckermanSEL: removed conditional rendering that was based on selection state. The input being selected is not treated as the element being selected. This might be a bug but I think it's tangential to this example.

CameronAckermanSEL:
Since someone has looked at this and it's just an example I'm going to self merge. Thank you!

PR #3525, opened 2020-02-27
next.js fix for portals

bodrovphone:

hovering-toolbar.js example doesn't work with Next.js. Reason of this is simple - no portals on ssr.

#### What's the new behavior?

hoverting-toolbar example works with next.js fine (although their is console error saying about client-server render inconsistency.

### DIFF
--- a/site/components.js
+++ b/site/components.js
@@ -122,11 +122,11 @@ export const Menu = React.forwardRef(({ className, ...props }, ref) => (
 ))
 
 export const Portal = ({ children }) => {
-  let portal = null;
+  let portal = null
   if (globalThis.document) {
-    portal = ReactDOM.createPortal(children, document.body);
+    portal = ReactDOM.createPortal(children, document.body)
   }
-  return portal;
+  return portal
 }
 
 export const Toolbar = React.forwardRef(({ className, ...props }, ref) => (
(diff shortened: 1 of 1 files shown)


lcswillems:
I don't think it is a good idea to change this example to make it working with Next.js, because it is not a Next.js example. Following your modification will complexify the example for nothing.

bodrovphone:
Thank you for your feedback, @lcswillems.
I still believe it wouldn’t hurt and help people who want to quickly take slate on their project working right away. 
took me quite a while to find out what’s going on before make it work. Hopefully will help others who is dealing with Next.js+Slate

lcswillems:
I think it is good to know how to make Slate working with Next.js, but I don't think your solution is the good one. Indeed, it will make it easier for Next.js users but complexify it for all the other users.

From my point of view, a better solution would be to rather do a "Slate + Next.js" page.

JedWatson:
If we considered this as not specific to next.js but relevant to anyone using Slate in a SSR environment, would be be acceptable to simply guard the portal in the example?

```
export const Portal = ({ children }) => {
  // the document does not exist in SSR, so guard the Portal
  return (typeof document === 'object' && document.body)
    ? ReactDOM.createPortal(children, document.body)
    : null;
}
```

lcswillems:
I am not a maintainer of the repo but I would say that if you do SSR, you know easily how to fix the code (the fix is not specific to Slate here), so there is no need to include this fix that will make code more complicated to understand.

Another thing: I think `typeof window !== "undefined"` is the simplest way to test if on the client.

Issue #3299, opened 2019-12-11
Allow for more complicated `matches` handling on Text & Element nodes


timbuckley:
#### Do you want to request a _feature_ or report a _bug_?

Feature

#### What's the current behavior?

`Text.matches` and `Element.matches` cannot handle matches where the values for props are objects.

For example:
```js
const textNode = {text: 'hiii', bold: {active: true}}
const searchProps = {bold: {active:true}}

 // Does not consider this a match because of JavaScript's !== is a reference check, eg: {} !== {}
Text.matches(textNode, searchProps) === false
```

#### What's the expected behavior?

Since the concept of the `marks` array is gone, properties on `Text` and `Element` nodes may become more complex, and so we should allow for more robust searching of properties on these Nodes.

Expected behavior:
```js
const textNode = {text: 'hiii', bold: {active: true}}
const searchProps = {bold: {active:true}}


Text.matches(textNode, searchProps) === true
```

More complicated expected matching behavior:
```js
const textNode = {text: 'hiii', bold: {active: true, inherited: false}}
const searchProps = {bold: {active:true}}

// Nested objects should also allow for partial matches.
Text.matches(textNode, searchProps) === true
```

I'll have a PR for this change shortly.

ianstormtaylor:
What use case are you needing the nested objects for?

And can you give a code snippet of where are you using the `Node/Text.matches` function?

timbuckley:
@ianstormtaylor Since the concept of Marks is removed in favor of attributes on nodes, more robust searching on those properties is useful.

```js
import { Text } from 'slate'

const exampleTextNode: Text = {
  text: 'Some bold text.',
  bold: {
    active: true,
    origin: 'INHERITED',
  }
};

// Now I want to find bolded Text nodes, regardless of their `origin`.
Text.matches(exampleTextNode, {bold: {active: true}})
// But ^ this returns false, not the desired behavior
```

ianstormtaylor:
What's your use case? Explain the domain-specific behaviors you're looking for.

I'm not likely to expand this contract in the API easily. No other parts of the API work that way to my knowledge, and it's generally common for things to not be loosely matched? Not to mention if you really need this you can always write one-off logic to test nested structures.

I'm not opposed, but I want to see strong merit.

timbuckley:
In our use case, we need to know two things about any given text property (bold, italics, etc):

[...]

PR #3532, opened 2020-03-02
`Text.equals` now works with more nested Text nodes, and non-primitive attributes.

timbuckley:

Feature

#### What's the new behavior?

Related to the problem with the `Text.matches` function called out #3299, the `Text.equals(nodeA, nodeB)` function would incorrectly return false if a property on both nodes were equivalent objects/arrays.

This means that some consumers of this library (such as my company), who would expect Slate to merge Text nodes together if they have similar properties, to find that expectation false if they use non-primitive values for Text node properties.

The new logic merely makes use of `lodash`'s  `isEqual` function to recurse through the objects looking for discrepancies. It will also omit the `text` property for each node under `loose` mode.

### DIFF
--- a/packages/slate/package.json
+++ b/packages/slate/package.json
@@ -18,6 +18,7 @@
     "esrever": "^0.2.0",
     "immer": "^5.0.0",
     "is-plain-object": "^3.0.0",
+    "lodash": "^4.17.4",
     "tiny-warning": "^1.0.3"
   },
   "keywords": [
--- a/packages/slate/src/interfaces/text.ts
+++ b/packages/slate/src/interfaces/text.ts
@@ -1,4 +1,6 @@
 import isPlainObject from 'is-plain-object'
+import isEqual from 'lodash/isEqual'
+import omit from 'lodash/omit'
 import { Range } from '..'
 
 /**
@@ -24,27 +26,10 @@ export const Text = {
   ): boolean {
     const { loose = false } = options
 
-    for (const key in text) {
-      if (loose && key === 'text') {
-        continue
-      }
-
-      if (text[key] !== another[key]) {
-        return false
-      }
-    }
-
-    for (const key in another) {
-      if (loose && key === 'text') {
-        continue
-      }
-
-      if (text[key] !== another[key]) {
-        return false
-      }
-    }
-
-    return true
+    return isEqual(
+      loose ? omit(text, 'text') : text,
+      loose ? omit(another, 'text') : another
+    )
   },
 
   /**
--- a/packages/slate/test/interfaces/Text/equals/complex-exact-equals.js
+++ b/packages/slate/test/interfaces/Text/equals/complex-exact-equals.js
@@ -1,0 +1,20 @@
+import { Text } from 'slate'
+
+export const input = {
+  textNodeA: {
+    text: 'same text',
+    bold: true,
+    italic: { origin: 'inherited', value: false },
+  },
+  textNodeB: {
+    text: 'same text',
+    bold: true,
+    italic: { origin: 'inherited', value: false },
+  },
+}
+
+export const test = ({ textNodeA, textNodeB }) => {
+  return Text.equals(textNodeA, textNodeB, { loose: false })
+}
+
+export const output = true
--- a/packages/slate/test/interfaces/Text/equals/complex-exact-not-equal.js
+++ b/packages/slate/test/interfaces/Text/equals/complex-exact-not-equal.js
@@ -1,0 +1,20 @@
+import { Text } from 'slate'
+
+export const input = {
+  textNodeA: {
+    text: 'same text',
+    bold: true,
+    italic: { origin: 'inherited', value: false },
+  },
+  textNodeB: {
+    text: 'same text',
+    bold: true,
+    italic: { origin: 'inherited', value: true },
+  },
+}
+
+export const test = ({ textNodeA, textNodeB }) => {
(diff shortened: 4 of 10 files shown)


majelbstoat:
I have a use-case in which this would be useful. I'm storing provenance of text as props on the node. The full set of possible props in my application looks like:

```ts
type Props = {
  adoption?: boolean
  attribution?: {
    original?: string,
    url?: string,
    unknown?: string,
    root?: Source
    via?: Source
  }
}

type Source = {
  id: string,
  edition: string,
  range: Range
}
```

I can do a useRef to a common object for some use-cases. But if I copy-paste from Slate to Slate, the fragments are serialised and deserialised which means they fail a comparison check, and they're not merged correctly. Which means I have to do my own merge normalisation.

 I was surprised to find that similarly property'd nodes were not automatically merged. I understand that design choice, but `isEqual` equality is much more inline with how I thought about them.

majelbstoat:
How about using https://github.com/epoberezkin/fast-deep-equal instead? It's used by a bunch of projects and seems to be much faster.

ianstormtaylor:
Thanks @timbuckley!

TheSpyder:
I thought we were encouraging node properties to be flat? I've been arguing against adding lodash to Slate, and deep equality in general, a few times (e.g. #4080). Performance is also a concern although I don't know how much `Text.matches` is used in practice.

JonasKruckenberg:
I have

JonasKruckenberg:
The addition of those two functions increases the bundle size by about 100kB which seems a little excessive.
I submitted #4245, which adresses at least the lodash/omit function and reduces bundle size by ~50kB, but I'd be all for using fast-deep-equal as well 👍🏻

ianstormtaylor:
I think there are some decent arguments for the deep properties, for use cases that aren’t possible without them. (I think having multiple highlights on a single span of text is one use case.)

But I’m definitely in favor of smaller bundle sizes, and not using Lodash if that’s the way to achieve it.

TheSpyder:
> I think there are some decent arguments for the deep properties

I'm glad you said that, because my team uses a nested object on block elements 😅 they're also useful to add a collection of editor-specific properties that might otherwise conflict with user-defined HTML attributes.

To maintain performance, though, I'm wondering if instead of a proper deep equal we just specifically support nested object comparison. `fast-deep-equal` looks better than lodash for Slate's purposes, but it still has specific comparisons for a lot of things that are unlikely to ever appear on a Slate node (`Array`, `Map`, `Set`, `Date`, `RegExp`...)

@JonasKruckenberg I'd be happy to accept replacing `isEqual` in your PR.

jason-codaio:
Late to the party here, but this change breaks the ability to have optional keys merge cleanly with serialization in place. Before this PR `{text: 'foo'}` and `{text: 'bar', isBold: undefined}` would merge cleanly. Because the value of isBold is undefined on both nodes. With the new libraries (both here and the replacement PR) check the key lengths are the same and assume these are not equal and cause the nodes to not be merged. This seems pretty unfortunate....

TheSpyder:
> With the new libraries (both here and the replacement PR) check the key lengths are the same and assume these are not equal and cause the nodes to not be merged

I have been considering removing even the small `fast-deep-equal` library used in the replacement PR - it has comparison logic for things that should never exist on a Slate node - and switching back to what we had before with simple added recursion for nested objects. This seems like a good reason to do that.

BrentFarese:
> Before this PR {text: 'foo'} and {text: 'bar', isBold: undefined} would merge cleanly. 

I don't really understand the use case for this, but if there is some other deep equal comparison that is appropriate, that may work.  I do think the prior implementation that did not deeply check an object's properties was insufficient though as Slate supports and many people use nested objects (including us).

jason-codaio:
@BrentFarese The use case is basically trying to reduce bloat. There are a number of properties to describe text style most of the time these styles are not applied, so having every single text node in a large document have a ton of properties all set to false causes overhead on the wire as well as just in memory (albeit a bit at the will of V8 optimizations).

All of our boolean properties on the text node are `property?: true` so the only valid values are `true | undefined`. This means there are a few transitions that are problematic. 

A simple serialization and storage of a document with something like JSON.stringify and later JSON.parse is going to strip these undefined keys off of the textNode. It feels particularly bad that something like serialization could cause change in behavior.

My understanding of the invariant with merging nodes is that if they have the same styling they should merge. In my example those nodes have the same styling.

I'm definitely not arguing against deeply nested checks, but it would be nice if they didn't come at the expense of previous functionality.

BrentFarese:
That use case makes sense @jason-codaio!  I personally am completely fine w/ some implementation that does a deep check and accommodates your use case.

JonasKruckenberg:
Yeah so a custom semantic-equal function it is, the idea tat nodes that render the same thing should also be equal is great 👍🏻

From my understanding your not saving any memory though @jason-codaio, since properties with the _value_ undefined are still defined properties that merely have the value undefined. All meta information is still preserved.

This still saves space in the serialized representation, not in memory probably though.

You could delete properties instead, that would work with the current isEqual and json but as you know horribly slow sadly...

TheSpyder:
OK, I'll work on a PR.

I wonder if, with this reverted, we could look at swapping `set_node` to set undefined and improve performance instead of using `delete`? That seems like a potentially risky change though.
https://github.com/ianstormtaylor/slate/blob/796389c7d3d9cead1493abcba6c678cb9dfa979f/packages/slate/src/transforms/general.ts#L202-L206

TheSpyder:
done #4276

PR #3533, opened 2020-03-02
Fix for backspace press in list

V0XNIHILI:
Made a fix for the fact that if you were standing in a `<li>` wrapped in an `<ul>` with no text and pressed backspace, the `<ul>` would go away while the `<li>`'s stay.

Fixing a bug

#### What's the new behavior?

When initially you have this in your editor (with the cursor in the last list item):

image

And then press backspace, the following happens:

image

With this fix, the following (correct behaviour) occurs:

image

I added `split=true` at the `Transforms.unwrapNodes()` point, just like in the example on the site with lists in rich text.

Fixes: #
Reviewers: @


CameronAckermanSEL:
Thanks so much @V0XNIHILI !!

Issue #3536, opened 2020-03-05
Microsoft Edge Legacy doesn't support 'beforeinput' event


pavlyna:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?

Children and selection don't get synced in Microsoft Edge Legacy.
Microsoft Edge Legacy doesn't support 'beforeinput' event https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforeinput_event
Apparently, no workaround has been introduced.

![ezgif com-video-to-gif (1)](https://user-images.githubusercontent.com/7562331/76009763-66667300-5f1a-11ea-932e-cac171a6e047.gif)

Slate: 0.57.1
Browser: Edge <79
OS: Windows

#### What's the expected behavior?

Whereas Microsoft Edge Legacy is not the newest, but is the latest of its kind, I would suggest consider providing support for it, in the same way as Firefox is handled since it has the same issue.

In addition, would like to mention the next issue https://github.com/ianstormtaylor/slate/issues/2060

Created PR just in case, #3537

p0lax:
That's would be great to fix it, because currently new version of Edge is not installing automatically, so majority of users which are used Edge will proceed using version ~44

eivindw:
We have finished upgrading slate - but will probably have to revert to old version if Edge issues are not fixed. Would be good if support for Edge could be kept until most Windows users are on the new version. Cutting old IE versions is fine - but Edge is still used by some of our users unfortunately.

PR #3537, opened 2020-03-05
3536 cover beforeinput behavior on edge legacy browser

pavlyna:
Fixing bug

#### What's the new behavior?

In Edge Legacy, children and selection become in sync with operations
In the same way as Firefox was covered as a workaround due to lack of support

Fixes: #3536
Reviewers: @

### DIFF
--- a/packages/slate-hyperscript/src/creators.ts
+++ b/packages/slate-hyperscript/src/creators.ts
@@ -24,6 +24,17 @@ import {
 
 const STRINGS: WeakSet<Text> = new WeakSet()
 
+function flattenDescendants(array: any[]) {
+  return array.reduce((acc, el) => {
+    if (Array.isArray(el)) {
+      acc.push(...flattenDescendants(el))
+      return acc
+    }
+    acc.push(el)
+    return acc
+  }, [])
+}
+
 const resolveDescendants = (children: any[]): Descendant[] => {
   const nodes: Node[] = []
 
@@ -73,7 +84,8 @@ const resolveDescendants = (children: any[]): Descendant[] => {
     }
   }
 
-  for (const child of children.flat(Infinity)) {
+  const flattened = flattenDescendants(children)
+  for (const child of flattened) {
     addChild(child)
   }
 
--- a/packages/slate-hyperscript/src/creators.ts
+++ b/packages/slate-hyperscript/src/creators.ts
@@ -24,17 +24,6 @@ import {
 
 const STRINGS: WeakSet<Text> = new WeakSet()
 
-function flattenDescendants(array: any[]) {
-  return array.reduce((acc, el) => {
-    if (Array.isArray(el)) {
-      acc.push(...flattenDescendants(el))
-      return acc
-    }
-    acc.push(el)
-    return acc
-  }, [])
-}
-
 const resolveDescendants = (children: any[]): Descendant[] => {
   const nodes: Node[] = []
 
@@ -84,8 +73,7 @@ const resolveDescendants = (children: any[]): Descendant[] => {
     }
   }
 
-  const flattened = flattenDescendants(children)
-  for (const child of flattened) {
+  for (const child of children.flat(Infinity)) {
     addChild(child)
   }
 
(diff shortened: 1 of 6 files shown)


packages/slate-hyperscript/src/hyperscript.ts
```diff
@@ -91,8 +91,9 @@ const createFactory = <T extends HyperscriptCreators>(creators: T) => {
       attributes = {}
     }
 
-    children = children.filter(child => Boolean(child)).flat()
-    const ret = creator(tagName, attributes, children)
+    const flattened: any[] = []
+    flattened.push(...children.filter(child => Boolean(child)))
+    const ret = creator(tagName, attributes, flattened)
```
CameronAckermanSEL: On reading, it doesn't look like this is completely identical to the previous code because it's missing the flat() call.

packages/slate-react/src/utils/environment.ts
```diff
@@ -14,3 +14,7 @@ export const IS_FIREFOX =
 export const IS_SAFARI =
   typeof navigator !== 'undefined' &&
   /Version\/[\d\.]+.*Safari/.test(navigator.userAgent)
+
```
CameronAckermanSEL: whats the userAgent for modern edge?

site/examples/paste-html.js
```diff
@@ -58,10 +58,7 @@ export const deserialize = el => {
   ) {
     parent = el.childNodes[0]
   }
-
-  const children = Array.from(parent.childNodes)
-    .map(deserialize)
-    .flat()
+  const children = [].push(...Array.from(parent.childNodes).map(deserialize))
```
CameronAckermanSEL: I'd recommend instead using a polyfill to incorporate the flat call for legacy edge

packages/slate-hyperscript/src/creators.ts
```diff
@@ -24,6 +24,17 @@ import {
 
 const STRINGS: WeakSet<Text> = new WeakSet()
 
+function flattenDescendants(array: any[]) {
```
CameronAckermanSEL: Instead of doing this, traditionally we've recommended that consumers of slate and it's peripheral libraries use polyfills to get access to calls that aren't available on dead browsers

packages/slate-hyperscript/src/hyperscript.ts
CameronAckermanSEL: It looks like the flat call was removed, was this intentional? What are the impacts on the behavior of this call without it?

CameronAckermanSEL:
Thanks for the PR! 

The compatibility fix to get legacy edge to be treated like firefox seems like the right way to deal with this problem, however I'm a little concerned about the removal of `flat()` calls through the code base. In this case, it seems much more tenable for consumers of the library to install polyfills to get this functionality working on browsers they need to support that are basically deprecated.

site/examples/paste-html.js
CameronAckermanSEL: Following up on this, if we went with the polyfill, it'd need to be scoped to the dependencies of the example site, not the slate packages.

packages/slate-react/src/utils/environment.ts
dominiceden: It's "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.74 Safari/537.36 Edg/79.0.309.43"

According to https://docs.microsoft.com/en-us/microsoft-edge/web-platform/user-agent-string

CameronAckermanSEL:
```suggestion
// "modern" edge was released at 79.x
```

site/examples/paste-html.js
pavlyna: @CameronAckermanSEL thanks for the feedback!
Got it. I have removed local support of flat. Let me know if anything else needs attention.

pavlyna:
@CameronAckermanSEL comments have been addressed. pls review

CameronAckermanSEL:
Thanks @pavlyna ! This looks like it works really well! Greatly appreciate the help with this compatibility issue!

eivindw:
@CameronAckermanSEL Any chance you could make a quick release soon to get changes like this out? Would be awsome if we could start using the new slate for our customers :)

CameronAckermanSEL:
No. I don’t have release permissions.

I get asked this constantly so please forgive the terseness.

eivindw:
Ok, sorry.

@ianstormtaylor - maybe you could make a release including this change?

PR #3541, opened 2020-03-06
Avoid mutating the selection in Transforms.setPoint

aglundahl:

Fixing a bug.

#### What's the new behavior?

The test case included in this PR and this code currently fail on `master`:

```javascript
Transforms.move(editor); // This can be anything that triggers a selection change.
Transforms.setPoint(editor, { offset: 0 }, { edge: 'focus' }); //=> TypeError: Cannot assign to read only property 'offset' of object '#<Object>'
```

The reason for this is that after the first selection change (`Transforms.move(editor)`), Immer has frozen the selection. In `Transforms.setPoint` there's a `Object.assign` call on the selection point, which throws the TypeError. This PR fixes this.

I never try to mutate any selection points. Instead I just create a new object.

### DIFF
--- a/packages/slate/src/transforms/selection.ts
+++ b/packages/slate/src/transforms/selection.ts
@@ -157,13 +157,10 @@ export const SelectionTransforms = {
 
     const { anchor, focus } = selection
     const point = edge === 'anchor' ? anchor : focus
-    const newPoint = Object.assign(point, props)
 
-    if (edge === 'anchor') {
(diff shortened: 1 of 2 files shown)


aglundahl:
As a side note, I'm not sure if there are more places where `children` or `selection` are mutated directly. But I guess they will surface sooner or later, since [auto freezing](https://immerjs.github.io/immer/docs/freezing) isn't disabled (which is probably a good thing).

Issue #3518, opened 2020-02-27
Cannot focus inline void elements


knpwrs:
#### Do you want to request a _feature_ or report a _bug_?

_bug_

#### What's the current behavior?

When arrowing through the text in the mentions example, as soon as a mention is highlighted, focus is lost and user can no longer use arrow keys.

![2020-02-26 19-54-03 2020-02-26 19_55_56](https://user-images.githubusercontent.com/174864/75402168-2aa42b80-58d2-11ea-8206-15497aceda5a.gif)

Slate: 0.57.1
Browser: Chrome
OS: Mac

#### What's the expected behavior?

The mentions should become highlighted and arrow keys should continue to work.

CameronAckermanSEL:
I think this is related to one of the PRs I merged in that was attempting to fix inputs in void nodes.

dalevy:
Is this related to Issue [#2685 ](https://github.com/ianstormtaylor/slate/issues/2685) and [#2019](https://github.com/ianstormtaylor/slate/issues/2019) ?

cj1128:
Examples with input, `editable-voids` and `embeds`  are all broken now, You just can't type anything in this input element.

Slate version: 0.57.1

CameronAckermanSEL:
Hi @cj1128 , the examples that are currently hosted should have functional input voids. Please double check and let me know if that's not the case. The live examples work for me in FF and Chrome.

dalevy:
> Hi @cj1128 , the examples that are currently hosted should have functional input voids. Please double check and let me know if that's not the case. The live examples work for me in FF and Chrome.
> 
> A version of slate that fixes the problem with the input voids has not yet been released.

By any chance, do you have a timeline for the release of the fix @CameronAckermanSEL ?

CameronAckermanSEL:
Hey @kilrain, the version of the library that has this bug should not have been released yet, it should only impact the live examples right now. Are you able to reproduce this on the latest release in npm?

kilrain:
Ah - that's good news. I'm in the middle of upgrading from .47 and haven't yet migrated the inline-void parts of my codebase. Thanks for the quick response @CameronAckermanSEL !

cj1128:
> Hi @cj1128 , the examples that are currently hosted should have functional input voids. Please double check and let me know if that's not the case. The live examples work for me in FF and Chrome.
> 
> A version of slate that fixes the problem with the input voids has not yet been released.

[...]

PR #3542, opened 2020-03-07
Changed selection reconciliation in the React Editor to work with zero width characters added to void elements.

ryanmitts:
Void elements are selected by the insertion of a zero-width character
that the browser can select.

Currently, the selection reconciliation (making sure the Slate selection
state matches the DOM) will only select things that are within the
Editor it is responsible for, and the DOM element that is the target of
the selection is editable in the editor.

A lot of times, void nodes have their rendered root elements marked as not
content editable, thus making this zero-width character not marked as
editable, and the selection reconciliation will not mark the selection as
allowed.

A way around this is to always have an inner element in a void node
that is marked as not content editable if needed, leaving the zero-width
character out of the wrapping. This adds an unneeded DOM node however
and might get confusing without being explicit in the documentation.

This creates a new check if a node is selectable. Being selectable
means its editable or a void nodes zero-width element.

Fixing a bug, #3518 

#### What's the new behavior?

The expected behavior of the ticket. Also see the above description.

See above description.

Fixes: #3518 
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -367,8 +367,8 @@ export const Editable = (props: EditableProps) => {
 
         if (
           domRange &&
-          hasEditableTarget(editor, domRange.startContainer) &&
-          hasEditableTarget(editor, domRange.endContainer)
+          hasSelectableTarget(editor, domRange.startContainer) &&
+          hasSelectableTarget(editor, domRange.endContainer)
         ) {
           const range = ReactEditor.toSlateRange(editor, domRange)
           Transforms.select(editor, range)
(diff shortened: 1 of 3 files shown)


CameronAckermanSEL:
Hey @ryanmitts , thanks so much for fixing this bug! 

When I was testing this, it looks like the images in the image example is no longer focusable (the cursor just skips them entirely(, which is interesting because the mentions still are. Do you think that people will want to have the image be focusable? Is there a way to accomplish that after this change?

ryanmitts:
@CameronAckermanSEL Interesting, I'll give it a look, I tested out the image example, but I'll give it another look. Is it just using the arrow keys or mouse too?

CameronAckermanSEL:
It was specifically using arrow keys (up, down arrow)

ryanmitts:
@CameronAckermanSEL Looks like there is a couple of issue with selecting void nodes, I'll try to take a look when I have time. For instance, selecting across a void node (mention or image) with the mouse.

nahumzs:
using  `"slate-react": "^0.57.1"` if the element is void (ex. mention) the useFocus and useSelect is always true whenever the user is on the same paragraphs, can this be part of the same bug or its completely another thing?

dylans (changes requested):
Is this PR still relevant? @ryanmitts let me know if you have an update or if Slate already fixed this or if I should just close it out. Thanks.

dylans:
It looks like this was resolved in #3572. If I'm incorrect, feel free to reopen and rebase against main. Thanks!

Issue #3505, opened 2020-02-16
Copied text is missing line breaks


tgonzalez-asapp:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Multiline text copied from the slate editor does not include line breaks when pasted as plain text
<img src=https://user-images.githubusercontent.com/1714020/74613921-e2c31e80-50f1-11ea-803f-f4cd7a313f91.gif width=300>

To reproduce use https://codesandbox.io/s/slate-copy-text-issue-5ix0o

Slate: 0.57.1
Browser: Chrome
OS: Mac

#### What's the expected behavior?

DataTransfer object should have the line breaks included in its `text/plain` data.

The root cause of this issue seems to be calling `getComputedStyle` on nodes that are not attached to the page DOM (see https://stackoverflow.com/questions/35360711/getcomputedstyle-returns-a-cssstyledeclaration-but-all-properties-are-empty-on-a). The `display` property is then always an empty string. 

https://github.com/ianstormtaylor/slate/blob/e54b07eba8ab23a8c26b275f77e8ec207f8c9b55/packages/slate-react/src/components/editable.tsx#L1102

Because of that, `getPlainText` is not improving the `element.innerText` behavior. If we did attach the root div to the document, then both `getPlainText` and `element.innerText` would include the line breaks.

I can think of two possible fixes:

- Attach the node to the dom, call `getPlainText`, then remove. In that case we might want to consider replacing `getPlainText` with just `.innerText`, which also works.
- Use the range to set the document selection (or just assume the document selection to be the same as the slate selection) and then get the text from the selection with `.toString()`, which apparently does respect line breaks (https://stackoverflow.com/questions/1146730/problem-detecting-newlines-in-javascript-range-object)

I'm willing to create a PR if I can get some insight about which would be the preferred approach.

CameronAckermanSEL:
Hi @tgonzalez-asapp , did you check out the Paste HTML Example? If you want something beyond plain text pasting or slate editor to slate editor pasting, you'll need to hook up your own paste handler.

tgonzalez-asapp:
@CameronAckermanSEL, thanks. Please consider that this issue is not about pasting, but copying. It is in fact plain text copied to the clipboard what is wrong, as it does not have the line breaks. Hooking a paste handler does not change what slate copies to the clipboard AFAIK. Please let me know if I'm wrong.

CameronAckermanSEL:
[...]

PR #3543, opened 2020-03-07
Attach the cloned DOM node before coputing its plaintext representation when copying.

ryanmitts:
The cloned dom node constructed when copying must be attached to the dom for getComputedStyle to work.

This adds the div to the body as a hidden element and removes after computing its text representation.

Fixing a bug.

#### What's the new behavior?

See above description.

See above description.

Fixes: #3505 
Reviewers: @


CameronAckermanSEL:
Thanks for fixing this issue @ryanmitts ! I tested the example site on the netlify preview, and it appears to correct problems with copying -> pasting into multiline text fields, and pasting plaintext back into the editor.

Nice work!

PR #3546, opened 2020-03-11
fix input in editable void example

K-Kit:
It seems like this is required to make the text field editable?
New to slate but spent the past several hours trying to figure out why I couldn't edit the field in the void and this fixed it.

### DIFF
--- a/site/examples/editable-voids.js
+++ b/site/examples/editable-voids.js
@@ -73,6 +73,7 @@ const EditableVoidElement = ({ attributes, children, element }) => {
       >
         <h4>Name:</h4>
         <input
+          data-slate-editor
           className={css`
             margin: 8px 0;
           `}


CameronAckermanSEL:
Hi @K-Kit , I'm a little confused because it looks like the editable void examples work on Firefox, Chrome, and Legacy Edge. Could you explain a little bit more about how you're encountering this problem?

K-Kit:
Hi @CameronAckermanSEL I see the examples working perfectly on the slate site however when I tried to implement it myself it did not work without that prop.  I even tried copy-pasting the example to make sure my implementation wasn't just wrong and had the same result.

I was running it in a contentful extension if that matters.  Mac OS - chrome, firefox, and safari all did not work. Node 10.13.0

CameronAckermanSEL:
Hi @K-Kit , when you tried it locally, did you pull down the repo and use that in your build? The master release has a fix that makes inputs work in void elements, but this is not yet released (master also has a bug that breaks selection of void nodes)

CameronAckermanSEL:
Hi @K-Kit , I'm pretty confident this issue is fixed on latest already so I'm going to close this PR. If you're still having problems with 0.57.2 please let me know and we can reopen this. 

Thanks for the contribution!

Issue #3521, opened 2020-02-27
When hitting backspace, all remaining nodes are removed


pavlyna:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?

See https://codesandbox.io/s/clever-tu-3cmy4

Step to reproduce
1. Having json like in the example provided above 
2. Put the cursor at the beginning of the first element
3. Hit backspace

Selected node is merged with the previous line, the rest of the nodes after the selected node are removed

Slate: 0.57.1
Browser: Chrome
OS: Mac / Windows / Linux

![ezgif com-crop (1)](https://user-images.githubusercontent.com/7562331/75457680-6e0ba200-5985-11ea-853b-27c58570144e.gif)

#### What's the expected behavior?
Selected node is merged with the previous line, the rest of the nodes after the selected node remain.

This happens in `mergeNodes` when determining `emptyAncestor`

pavlyna:
just in case, PR #3547

juliankrispel:
Managed to reproduce this issue with the original codesandbox example, upgraded it and cannot reproduce it anymore, seems to now work as it should, here's a sandbox with an upgraded version of slate for reference https://codesandbox.io/s/zealous-golick-gs9vg

Hence closing this issue

PR #3547, opened 2020-03-11
3521 When hitting backspace, all remaining nodes are removed 

pavlyna:
fixing bug

#### What's the new behavior?

Only current node is merged with the previous line. The rest of the nodes below remain

Fixes: #3521 
Reviewers: @


juliankrispel:
Was going to look into merging this but issue seems to be fixed on the most recent version of slate, [here's an example](https://codesandbox.io/s/zealous-golick-gs9vg)

PR #3549, opened 2020-03-15
Correct onDOMBeforeInput typings

josias-r:

TS will report an error when trying to access event.inputType since it was declared of type "Event" and not "InputEvent".

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -75,7 +75,7 @@ export interface RenderLeafProps {
 
 export type EditableProps = {
   decorate?: (entry: NodeEntry) => Range[]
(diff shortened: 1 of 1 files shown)


josias-r:
Further down I just found another interface (when the method is being defined using useCallback).. I'm not sure if that one needs to be adjusted as well 😬
Typings are wrong anyways...

gmathieu:
`event.inputType` is very useful. It's even referenced in one of the examples:

https://github.com/ianstormtaylor/slate/blob/f4ee75d13d219214fd12c1b10448098f4c070b77/site/examples/hovering-toolbar.js#L20-L29

@ianstormtaylor is `Event.inputType` meant to be internal only?

ianstormtaylor:
Thanks @josias-r!

PR #3552, opened 2020-03-19
use state to avoid cursor jumping in controlled input

Entkenntnis:

Fixing a bug

#### What's the new behavior?

Cursor is not jumping to end in Embeds example

Controlled inputs need a separate state to handle proper cursor management. This PR adds a state to the input that stores the current url value.

Fixes: #
Reviewers: @


CameronAckermanSEL:
@Entkenntnis  Thanks a ton for catching this issue and fixing it! It looks like it works great!

Issue #3550, opened 2020-03-17
onChange handler is triggered after unmount


jdimond:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Minimal repro: https://codesandbox.io/s/serene-hooks-4dzks.

![repro](https://user-images.githubusercontent.com/1628592/76885775-0090ba00-683d-11ea-9f82-db1e878621ba.gif)

The onChange handler is triggered *after* the component is unmounted. As far as I can see this happens because

- the [onDOMSelectionChange](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/editable.tsx#L354) callback is run asynchronously due to the call to throttle.
- However, that callback still has a reference to the editor, which in turn avoids deallocating the callback stored in the weak reference map [here](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/slate.tsx#L35).
- The callback runs a Transforms.(de)select which then causes the onChange callback to be run (after the component has already been unmounted).

Note that this is not reproducible if the toggle switch is replaced with an actual button rather than a normal element with an onClick handler.

Slate: 0.57.1
Browser: Chrome
OS: Mac

#### What's the expected behavior?

onChange should stop triggering immediately after an unmount.

qin9smile:
@jdimond 
use ref instead.
https://codesandbox.io/s/stupefied-pine-ynedd

jdimond:
@qin9smile using ref (and that examples) still repros with the error for me.

qin9smile:
> @qin9smile using ref (and that examples) still repros with the error for me.

you are right.

PR #3553, opened 2020-03-20
fix(unmount): add destroyed flag, prevent apply change when component…

qin9smile:
fix: https://github.com/ianstormtaylor/slate/issues/3550

### DIFF
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -5,7 +5,7 @@ import { ReactEditor } from '../plugin/react-editor'
 import { FocusedContext } from '../hooks/use-focused'
 import { EditorContext } from '../hooks/use-editor'
 import { SlateContext } from '../hooks/use-slate'
-import { EDITOR_TO_ON_CHANGE, IS_DESTROYED } from '../utils/weak-maps'
+import { EDITOR_TO_ON_CHANGE } from '../utils/weak-maps'
 
 /**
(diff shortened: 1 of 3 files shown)


CameronAckermanSEL:
Thanks very much for the contribution, @qin9smile. Really appreciate you taking on this defect. 

I had a couple of questions I posted in the code

packages/slate-react/src/components/slate.tsx
```diff
@@ -33,6 +33,13 @@ export const Slate = (props: {
   }, [key, onChange])
 
   EDITOR_TO_ON_CHANGE.set(editor, onContextChange)
+  IS_DESTROYED.set(editor, false)
+
```
CameronAckermanSEL: Instead of having a separate property to track this, could we instead use the effect to clear the EDITOR_TO_ON_CHANGE property?

alexander-zolotykh:
Can we solve the problem this way or another way so that it can be fixed finally?

packages/slate-react/src/components/slate.tsx
qin9smile: i changed. haha.. i have been forgotten.

qin9smile:
> Can we solve the problem this way or another way so that it can be fixed finally?

maybe it can be fixed now.

CameronAckermanSEL:
Thanks @qin9smile ! This looks good to me!

Issue #3137, opened 2019-11-28
Images are now duplicated instead of being draggable


laufeyrut:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?

From the examples https://www.slatejs.org/examples/images 

Now images are duplicated instead of being moved around in the text when dragging them.

Chrome Version 78.0.3904.108
MacOs Catalina 10.15.1
Slate version 0.50.0

![Kapture 2019-11-28 at 13 01 08](https://user-images.githubusercontent.com/15036127/69808345-2d15bd00-11df-11ea-82c6-45c0034c46b7.gif)

#### What's the expected behavior?

In the last version Images were draggable

ejoebstl:
@ianstormtaylor Any leads on how to tackle this issue?
Is this a problem with the example or will it require digging deeper?

kamilmielnik:
I'd like to point out one more issue here: dragging an image does not work if image is not selected. Instead its' URL is dragged.

![Peek 2020-09-25 15-10](https://user-images.githubusercontent.com/6830683/94243258-59a27e80-ff41-11ea-9358-f80d81be4dc1.gif)

alexanderisora:
Thanks for opening this issue, @laufeyrut 
I have subscribed to follow any updates.

clauderic:
Solved by #4238

nikita-neganov:
@clauderic still duplicates at demo https://www.slatejs.org/examples/images

clauderic:
@nikita-neganov there are issues with the auto-deploy of the https://www.slatejs.org website when changes are merged onto the main branch, I've informed @ianstormtaylor of this issue.

Locally when you clone the project, the problem is fixed.

tjuandev:
For some reason the problem is still persisting in Firefox...

Issue #3522, opened 2020-02-27
Drag & drop of nested elements in 0.57.x


DianaLease:
#### Do you want to request a _feature_ or report a _bug_?
🐛 

#### What's the current behavior?

Slate: 0.57.1
Browser: Chrome
OS: Mac

https://codesandbox.io/s/slate-reproductions-7klqo
- Dragging and dropping a nested element (in this example, a div with a paragraph child) does not work as expected.
- Element is not placed where the user's cursor was when dropped.
- Original element is not removed after dropping.
![nw5YwtaEl7](https://user-images.githubusercontent.com/20543103/75476038-e19eb680-5967-11ea-81d4-78fe6d8e8562.gif)

#### What's the expected behavior?

- Element is placed where the user's cursor was when they dropped the element.
- The original element is removed.

Elvenisboy:
it seems like can write like this:

```
<div
    draggable="true"
    style={{ backgroundColor: 'red', width: '200px' }}
    onDragStart={...}
    onDragOver={...}
    onDrop={...}>
    ...
</div>
```

DianaLease:
> it seems like can write like this:
> 
> ```
> <div
>     draggable="true"
>     style={{ backgroundColor: 'red', width: '200px' }}
>     onDragStart={...}
>     onDragOver={...}
>     onDrop={...}>
>     ...
> </div>
> ```


Thanks for you comment, @Elvenisboy ! Do you have sandbox example using this? I just tried it with the sandbox I provided in the issue but it unfortunately does not fix the issue.

Elvenisboy:
> > 似乎可以这样写：
> > ```
> > <div
> >     draggable="true"
> >     style={{ backgroundColor: 'red', width: '200px' }}
> >     onDragStart={...}
> >     onDragOver={...}
> >     onDrop={...}>
> >     ...
> > </div>
> > ```
> 
> 感谢您的评论，@ Elvenisboy！您是否有使用此功能的沙箱示例？我只是在问题中提供的沙箱中尝试过，但是很遗憾，它无法解决问题。

Sure, just see here [https://codesandbox.io/s/example-for-issues-3522-1meyy](url)

I don't konw how to swap nodes by slate native, so I use context.

Although It need  more code, it works well

llevequeblup:
The same issue can be replicated on https://www.slatejs.org/examples/richtext
Browser: Firefox 74.0
OS : Ubuntu
[...]

PR #3562, opened 2020-03-25
Fix duplicated content after drag and drop

llevequeblup:

Fixing a bug

#### What's the new behavior?

When drag and dropping text on Firefox, or more complex objects on all browsers, actually move the drag and dropped item instead of simply copying it

Previously the beforeinput event with deleteByDrag inputType was relied on to clean dragged item, but this event is not thrown in Firefox, and neither in Chrome for non-textual content.
This change reproduce behaviour found in v0.47 : a flag is set when dragging starts, cleared if dragging stops, and checked when dropping content to delete the old selection.

Fixes: #3137, probably #3522


rabiesnoussi:
Hi @llevequeblup does these changes fix the problem ?

llevequeblup:
Yes they do

rabiesnoussi:
Cool !! Is it going to be merged soon, or does it need more tests before ?

CameronAckermanSEL:
Hey @llevequeblup, thank you for the PR. When I took a look at this and tested it on chrome, I noticed that while the latest version of the slate examples has functional drag and drop for normal text, the deployment for this PR doesn't remove the old text anymore. I haven't yet tried it from Firefox, and I also can't see what part of the PR might have lead to this regression.

When you have a free moment, lemme know what you think!

Arcainne:
Any update on this? Would love to see this issue fixed for dragging and dropping void elements. :)

llevequeblup:
When I tested drag and drop of normal text on Chrome, I figured out that it worked only if the source and target were not in the same span (I am not sure why).
If I drag and drop text from one place of a span to another place of the same span, the text always get dropped at the start of the span.

When I tried fix this I supposed it was the selection that got messed up and I wanted to use ReactEditor.findEventRange in the inputFromDrop branch of the beforeinput handler, but actually the InputEvent has no clientX/clientY, and therefore there is no way to know where the drop happened.

Since it looks that there is no way to make this work in every cases using beforeinput, my new changes do nothing in beforeinput handler when the inputType is deleteFromDrag or insertFromDrop, and enable back the handling in onDrop even if the browser HAS_BEFORE_INPUT_SUPPORT

keyserfaty:
no updates on this? it's a pretty big issue.

alexanderisora:
I'm waiting for this PR impatiently 😅

clauderic:
Hey @llevequeblup, thanks for your contribution. I'm going to be closing this PR as we recently merged #4238, which was largely inspired by you PR and fixes the same issues.

PR #3566, opened 2020-03-28
TypeScript Improvement: Use `[key: string]: unknown`, not `[key: string]: any`

timbuckley:

Improvement.
Mostly fixes issue #3416, superseding WIP PR attempt #3473.
(Arguably fixes a bug for TypeScript users)

#### What's the new behavior?

As described in issue #3416, the current type interfaces for most important Slate objects looks something like this:

```ts
interface Text {
  text: string;
  [key: string]: any;
}
```

This PR mainly changes this and similar types to:
```ts
interface Text {
  text: string;
  [key: string]: unknown;
}
```

---
The original purpose of the `[key: string]: any` was to say, effectively "This `Text` object is allowed to have other key-value pairs besides `text`", which is needed.

However, because the value for every arbitrary key is of type `any`, it muffles any errors or warnings TypeScript would otherwise give, causing potential runtime errors. See below:

Screen Shot 2020-03-28 at 7 11 59 PM

As in the example, `unknown` is the more appropriate type to use here. It allows any arbitrary keys on the `Text` object, but makes no assumption about its value, and requires the user to typecheck, preventing runtime errors.

This has two major benefits to the codebase.

1) Previous possible type errors in the source are caught. For example, `Node` objects now are no longer assumed to have `children` defined, because they can be `Text` nodes, which are not guaranteed to have that property (and generally don't). Now the codebase has to either guard against those cases, or inform the type system (via `as` assertions) when that is seen as reasonable (eg a parent to a `Node` will be an `Ancestor`).

2) Slate types now play well with user-extended types. When an extension defines another required field, TypeScript can make sure non-defined keys are not treated as defined:

Screen Shot 2020-03-28 at 7 20 23 PM

I don't believe this offers the full flexibility that full type generics offer, but helps us get there.

Fixes: #3416 (mostly)
Reviewers: @ianstormtaylor, @miracle2k (Because you previously commented on the related issue)

### DIFF
--- a/docs/api/locations.md
+++ b/docs/api/locations.md
@@ -22,7 +22,7 @@ type Path = number[]
 
 ## Point
 
-`Point` objects refer to a specific location in a text node in a Slate document. Its `path` refers to the lcoation of the node in the tree, and its offset refers to distance into the node's string of text. Points may only refer to `Text` nodes.
+`Point` objects refer to a specific location in a text node in a Slate document. Its `path` refers to the location of the node in the tree, and its offset refers to distance into the node's string of text. Points may only refer to `Text` nodes.
 
 ```typescript
 interface Point {
--- a/docs/api/locations.md
+++ b/docs/api/locations.md
@@ -28,7 +28,7 @@ type Path = number[]
 interface Point {
     path: Path
     offset: number  
-    [key: string]: any
+    [key: string]: unknown
 }
 ```
 
@@ -68,7 +68,7 @@ Options: `{affinity?: 'forward' | 'backward' | null}`
 interface Range {
     anchor: Point
     focus: Point
-    [key: string]: any
+    [key: string]: unknown
 }
 ```
 
--- a/docs/api/nodes.md
+++ b/docs/api/nodes.md
@@ -123,7 +123,7 @@ interface Editor {
   selection: Range | null
   operations: Operation[]
   marks: Record<string, any> | null
-  [key: string]: any
+  [key: string]: unknown
 
   // Schema-specific node behaviors.
   isInline: (element: Element) => boolean
@@ -212,7 +212,7 @@ Apply an operation in the editor.
 ```typescript
 interface Element {
   children: Node[]
-  [key: string]: any
+  [key: string]: unknown
 }
 ```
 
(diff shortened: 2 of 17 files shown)


miracle2k:
I'm not most suitable to review it, due to lack of knowledge of the Slate.js code base, but I'll give it a try and run it in my project to see if there are any issues.

What was reason to move away from using generics for now, has that been discussed somewhere?

timbuckley:
@miracle2k I'm definitely not opposed to the generics work, it's just a lot more work to implement. This is a much faster way to get most of the value generics would bring.

CameronAckermanSEL:
Hi @timbuckley , thank you very much for this contribution. This seems like a great way to strengthen the safety of code using slate. 

When we go to release this, would you consider this to be a breaking change to the existing interface? Would you recommend treating this as a minor version bump or would it just be a patch release?

Thanks!

timbuckley:
@CameronAckermanSEL I think this is best considered a minor version bump. It technically corrects a bug, but someone accepting all patches might will find that their build fails suddenly. So it's better considered a breaking change.

CameronAckermanSEL:
Hey @timbuckley , I'm going to revert this briefly and then roll out a release real quick of our existing bug fixes. Thanks so much!

CameronAckermanSEL:
Alright @timbuckley , I reapplied the change. Thanks so much! I'll make sure to release it as a minor version bump

majelbstoat:
Oh boy, I know is technically correct, and I'm going to be safer, but I now have 107 places that I have to typecheck my props 😂😬😭

Roll on actual generic type signatures!

flavordaaave:
@timbuckley @CameronAckermanSEL Have you guys also considered removing the index signature completely?
The reason I ask is that using index signature makes it impossible to to use type guards on types like `Node` when using some of the internal methods which return `Node`(s):

As it is now:
```
if ('children' in node) {
   const children = node.children // `unknown` instead of `Node[]`
}
```

When removing the index signature the user could still use module declaration/merging to add custom types:

```
declare module 'slate' {
  interface Element {
    myCustomProp: string
  }
}
```

And than the type-guards would work for as expected:

```
if ('children' in node) {
   const children = node.children // `Node[]`
  const myCustomProp = node.myCustomProp // string
}
```

PR #3572, opened 2020-04-01
fix issue where inline voids can no longer be selected

### LINKED ISSUE #3518: Cannot focus inline void elements

CameronAckermanSEL:
Bug

#### What's the new behavior?
This fixes #3518 by allowing zero width characters to be considered editable dom targets, thereby alloing `onDOMSelectionChange` to select them.

This adds zero-width characters to the set of editable targets that are considered selectable DOM children of the editor

Examples checked:
* Mentions Example
* Editable Voids Example
* Embeds Example
* Checklist example (This defect was introduced accidentally as part of https://github.com/ianstormtaylor/slate/pull/3389)

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -174,11 +174,9 @@ export const ReactEditor = {
 
     return (
       targetEl.closest(`[data-slate-editor]`) === editorEl &&
-      (
-        !editable ||
+      (!editable ||
         targetEl.isContentEditable ||
-        !!targetEl.getAttribute('data-slate-zero-width')
-      )
+        !!targetEl.getAttribute('data-slate-zero-width'))
     )
   },
 
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -174,7 +174,7 @@ export const ReactEditor = {
 
     return (
       targetEl.closest(`[data-slate-editor]`) === editorEl &&
-      (!editable || targetEl.isContentEditable)
+      (!editable || targetEl.isContentEditable || !!targetEl.getAttribute('data-slate-zero-width'))
     )
   },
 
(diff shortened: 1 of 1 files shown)


CameronAckermanSEL:
@ryanmitts  you've been poking around with selection recently so maybe you could provide some insight on whether this is a sane change? Appreciate your time if you can, if not, no problem at all :)

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -174,7 +174,9 @@ export const ReactEditor = {
 
     return (
       targetEl.closest(`[data-slate-editor]`) === editorEl &&
-      (!editable || targetEl.isContentEditable)
+      (!editable ||
+        targetEl.isContentEditable ||
+        !!targetEl.getAttribute('data-slate-zero-width'))
```
CameronAckermanSEL, line 179: adding slate-zero-width characters here means that onDOMSelectionChange will allow this to be a selectable region. The absence of this code causes the deselect routine to be called in the event handler. 

See: https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/editable.tsx#L371

Before this change, hasEditableTarget returns false for inline voids (because they're not `isContentEditable=true`)

CameronAckermanSEL:
Tests:

- Checked selection via keyboard on inline voids (forward and backward)
- Checked selection via keyboard on normal inlines and marks (forward and backward) (found same bugs that are on production)
- Checked selection via keyboard on void blocks (forward and backward) (found same bugs that are on production)
- Checked selection via mouse on inline voids (forward and backward)
- Checked selection via mouse on normal inlines and marks (forward and backward) (same bugs)
- Checked selection via mouse on void blocks (forward and backward) (found same bugs)
- Double checked that embedded inputs, and other editable void nodes could still be edited.
- Double checked selection across editable voids (it behaves oddly but this is not a new behavior)

Issue #3544, opened 2020-03-09
Backwards selection anchored at the end of an inline block has unexpected behavior


JonathanWbn:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Selecting text when starting at the end of the line and that end being a link the selecting doesn't work.
![Screen Recording 2020-03-09 at 11 42 54](https://user-images.githubusercontent.com/19997520/76206178-d6674880-61fb-11ea-83fb-483637376625.gif)

Slate: 0.57.1
Browser: Chrome
OS: Mac

kilrain:
I dug into this bug over the weekend a bit. The issue seems to have to do with the [zeroWidthString node](https://github.com/ianstormtaylor/slate/blob/16ff44d0566889a843a346215d3fb7621fc0ed8c/packages/slate-react/src/components/string.tsx#L73) inserted after an inline. Native DOM selection has a tendency to put the offset AFTER the zero width character (`offset: 1`) while Slate prefers to consider this state as `offset: 0`.

On the initial mousedown, the native selectionchange event sets the cursor to offset of 1... immediately, [Slate resets the offset to 0](https://github.com/ianstormtaylor/slate/blob/16ff44d0566889a843a346215d3fb7621fc0ed8c/packages/slate-react/src/components/editable.tsx#L165). After beginning to drag, native selection reverts to an anchor offset of 1 and Slate processes the update but, again, considers it as an offset of 0. This doesn't trigger a re-render (since Slate already thinks the selection anchor is offset 0), thus the [`useLayoutEffect`](https://github.com/ianstormtaylor/slate/blob/16ff44d0566889a843a346215d3fb7621fc0ed8c/packages/slate-react/src/components/editable.tsx#L133) does not get a chance to sync the DOM.

When Slate goes to re-render (as a result of the selection focus update) a native selectionchange event fires with the collapsed selection state we see in this bug. I could not successfully figure out where the collapsing selectionchange event is coming from.  If I dragged the mouse very slowly over a void inline node, I could get into a valid state where the void node was selected... but if I continue dragging another pixel, the selection would collapse on the other side of the void node (as we see in the original bug report).

JonathanWbn:
Thank you, @kilrain, this was helpful. I found a solution and opened this PR: #3574.

PR #3574, opened 2020-04-02
Fix: Backward selection with empty text nodes

JonathanWbn:
As described in #3544 selecting text by dragging backward causes issues if the anchor is an empty text node.

This is due to an inconsistency between slate and native DOM selection ranges (see also [this comment](https://github.com/ianstormtaylor/slate/issues/3544#issuecomment-602718866) from @kilrain). Native DOM selection points of empty text nodes have an offset of 1, while slates selection points have an offset of 0.
So when slate [sets a new DOM range](https://github.com/ianstormtaylor/slate/blob/16ff44d0566889a843a346215d3fb7621fc0ed8c/packages/slate-react/src/components/editable.tsx#L165) (starting and/or ending in an empty text node) based on the slate range, the DOM range point is set with an offset of 0, while native DOM would expect an offset of 1. This causes the described issue and is fixed by correcting this inconsistency when converting a slate range into a DOM range.

With this fix, the issue can no longer be observed:
u3GwLx3CZc

### DIFF
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -19,6 +19,7 @@ import {
   DOMSelection,
   DOMStaticRange,
   isDOMElement,
+  isEmptyTextNode,
   normalizeDOMPoint,
 } from '../utils/dom'
 
@@ -268,10 +269,18 @@ export const ReactEditor = {
       : ReactEditor.toDOMPoint(editor, focus)
 
     const domRange = window.document.createRange()
-    const start = Range.isBackward(range) ? domFocus : domAnchor
-    const end = Range.isBackward(range) ? domAnchor : domFocus
-    domRange.setStart(start[0], start[1])
-    domRange.setEnd(end[0], end[1])
+    const [startNode, startOffset] = Range.isBackward(range)
+      ? domFocus
+      : domAnchor
+    const [endNode, endOffset] = Range.isBackward(range) ? domAnchor : domFocus
+
+    // A slate Point pointing to a Leaf without text will always have an offset of 0.
+    // A native DOM selection point on the other hand will have an offset of 1.
+    // Therefore, DOM selection points are given an offset of 1 instead of 0.
+    // Not doing this causes issues when backward selcting (see issue #3544).
+    domRange.setStart(startNode, isEmptyTextNode(startNode) ? 1 : startOffset)
+    domRange.setEnd(endNode, isEmptyTextNode(endNode) ? 1 : endOffset)
+
     return domRange
   },
 
--- a/packages/slate-react/src/utils/dom.ts
+++ b/packages/slate-react/src/utils/dom.ts
@@ -69,6 +69,13 @@ export const isPlainTextOnlyPaste = (event: ClipboardEvent) => {
 }
 
 /**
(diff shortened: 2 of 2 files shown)


czechdave:
I've just ran into this issue as well and it might be better to check for the `data-slate-zero-width` attribute instead of checking the `textContent` of the node.

Maybe something like this?
```js
const startEl = (isDOMElement(startNode)
      ? startNode
      : startNode.parentElement) as HTMLElement
const isStartAtZeroWidth = !!startEl.getAttribute('data-slate-zero-width')
domRange.setStart(startNode, isStartAtZeroWidth ? 1 : startOffset)
```

JonathanWbn:
> I've just ran into this issue as well and it might be better to check for the `data-slate-zero-width` attribute instead of checking the `textContent` of the node.
> 
> Maybe something like this?
> 
> ```js
> const startEl = (isDOMElement(startNode)
>       ? startNode
>       : startNode.parentElement) as HTMLElement
> const isStartAtZeroWidth = !!startEl.getAttribute('data-slate-zero-width')
> domRange.setStart(startNode, isStartAtZeroWidth ? 1 : startOffset)
> ```

The `ZeroWidthString` node is a `<span>` and with that a `TEXT_NODE` but not an `ELEMENT_NODE`. `TEXT_NODE`s don't have the [`getAttribute`](https://developer.mozilla.org/de/docs/Web/API/Element/getAttribute) or the [`hasAttribute`](https://developer.mozilla.org/de/docs/Web/API/Element/hasAttribute) method.

kilrain:
> The ZeroWidthString node is a <span> and with that a TEXT_NODE but not an ELEMENT_NODE. TEXT_NODEs don't have the getAttribute or the hasAttribute method.

Plus, even if you were able to get a ref of the correct element, querying the DOM is going to be slower than just looking up a property on the node object.

czechdave:
> 
> The `ZeroWidthString` node is a `<span>` and with that a `TEXT_NODE` but not an `ELEMENT_NODE`. `TEXT_NODE`s don't have the [`getAttribute`](https://developer.mozilla.org/de/docs/Web/API/Element/getAttribute) or the [`hasAttribute`](https://developer.mozilla.org/de/docs/Web/API/Element/hasAttribute) method.

That is the reason for the `isDOMElement` check.

JonathanWbn:
Closing as this is obsolete with #3594.

JonathanWbn:
> > The `ZeroWidthString` node is a `<span>` and with that a `TEXT_NODE` but not an `ELEMENT_NODE`. `TEXT_NODE`s don't have the [`getAttribute`](https://developer.mozilla.org/de/docs/Web/API/Element/getAttribute) or the [`hasAttribute`](https://developer.mozilla.org/de/docs/Web/API/Element/hasAttribute) method.
> 
> That is the reason for the `isDOMElement` check.

I still don't get it. Can you help me out here, @czechdave? Here's what I don't understand: The `data-slate-zero-width` is set on a `span`. And a `span` is a text node. So how can `startEl.getAttribute('data-slate-zero-width')` ever be truthy?

czechdave:
`span` is an element that contains a `text` node. The selection is inside a `text` node inside a `span` so `startNode.parentElement` gets the `span` that has `getAttribute`.

PR #3579, opened 2020-04-03
Update wrapNodes to use lowest mode and create passing test.

BrentFarese:

Fixing a bug.

#### What's the new behavior?

Slate's `wrapNodes` function does not handle nested block structures with inlines.  When trying to have an inline wrap other nodes (e.g., other inlines or text), if the Slate document has a nested block structure, wrapNodes inserts the wrapping inline in the incorrect place.

This is a result of `wrapNodes` using `mode: 'highest'` [here](https://github.com/ianstormtaylor/slate/blob/1d7ab974292a3e831908a2ba0aab9fdd8a66fe10/packages/slate/src/transforms/node.ts#L798).

By switching to `mode: 'lowest'`, we get the correct `roots` for inline wrapping with nested block documents.

This PR changes to `mode: 'lowest'` and also adds a test with a nested block structure in place of `packages/slate/test/transforms/wrapNodes/inline/inline-across.js` (now re-named to `inline-across-nested.js`).

Reviewers: @CameronAckermanSEL, @timbuckley, @ianstormtaylor


BrentFarese:
@CameronAckermanSEL know you have been deep in the release, but give this a look when you can and then I can merge it if you don't have any comments.  Thanks!

CameronAckermanSEL:
Hi @BrentFarese , I'm really sorry that I'm leaving you hanging here. I will try to look at this PR this weekend. Thank you again for the contribution, it is greatly appreciated.

One thing though is that I'm deliberately withholding merging of PRs until release. The release is currently containing a great deal of bug fixes, and at this time I have a good confidence that we are not releasing any new bugs. I'd like to release what we have now, and then get in some other high value bug fixes out in a more controlled way shortly after.

Thanks for your patience and understanding :)

BrentFarese:
No problem at all @CameronAckermanSEL thanks for taking a look when you can!

CameronAckermanSEL:
Hey @BrentFarese , finally in a position to look at this and merge it. Thank you! I see that we modified an existing test , would it be redundant to have the old test case left alone and have a new file for deeply nested inlines?

BrentFarese:
No problem @CameronAckermanSEL!  Added back the test you referred to.  Also added .idea to .gitignore as that is the IDE I'm using.  Good to squash and merge?

CameronAckermanSEL:
This looks good! But I don't know why the netlify deployment is failing. I'm not too concerned about it yet, but just for safety could you merge/rebase with master explicitly and see if that clears the problem?

Thanks, @BrentFarese !

BrentFarese:
K just rebased off Master @CameronAckermanSEL and that seems to have done the trick.

Issue #3590, opened 2020-04-07
Editor.previous() returns the incorrect node if match isn't true for the specified location.


majelbstoat:
#### Do you want to request a _feature_ or report a _bug_?

report a _bug_

#### What's the current behavior?

In `Editor.previous()`, the second result of `Editor.nodes()` is always returned:

https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/interfaces/editor.ts#L1122

As a result, calling `Editor.previous()` with a location that doesn't satisfy the match function causes the wrong node to be returned.

Example:

https://codesandbox.io/s/crazy-engelbart-evxu3

Slate: 0.57.1
Browser: Chrome
OS: Mac

#### What's the expected behavior?

Editor.previous should always return the first element  that satisfies the match function before the supplied Location.

BrentFarese:
#3791, which I submitted appears to be a duplicate of this.

BrentFarese:
FWIW, in my own codebase my solution is to find the `Point` after the `at` location passed in and then return the first match after that `Point`.  It works well for me and maybe it'd work well enough for core...

BrentFarese:
Fixed by #3957.

PR #3591, opened 2020-04-07
fix: ensure Editor.previous() returns the correct node even when the `at` Location doesn't satisfy the match function.

majelbstoat:

fixing a _bug_

#### What's the new behavior?

The previous node is now returned even when the `at` location doesn't match the specified match function. Previously, calls with that characteristic would return the second previous location.

I added a failing test that outlines the behaviour, then fixed it.

We look at the first returned result and compare its to that of the first node at the specified location. If it is the same, then the result must be the second node. If it is not the same, which happens when the supplied location does not satisfy the match function, we return the first node.

Fixes: #3590

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -1126,17 +1126,19 @@ export const Editor = {
       mode,
       voids,
     })
+
     if (first) {
-      // If the first result is the first in the at, it can't be the previous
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Thanks for submitting this @majelbstoat, I believe this was fixed in another PR so I'm going to close. But if you think that's a mistake just let me know and we can re-open. Thanks!

majelbstoat:
No worries, nice to see you back ;)

PR #3593, opened 2020-04-08
Remove lodash from production dependencies

Andarist:

improving the bundlesize

#### What's the new behavior?

no behavior changes

No


majelbstoat:
lodash is tree-shakeable. are you sure this will reduce bundle sizes?

Andarist:
I'm pretty sure - lodash by default is not tree-shakeable (at least not by webpack, rollup might be a little bit smarter, but i don't have time to check this now). Webpack can only tree-shake ES modules and lodash doesn't distribute those right now (unless you are using `lodash-es`)

Also - lodash functions (although not sure about throttle) tend to drag into the bundle many other functions with them, because they are built on top of them etc.

CameronAckermanSEL:
Tree shaking is a central feature of rollup so the bundle that slate produces should not have any of the extraneous lodash code. Please double check and let us know if that's not accurate.

Andarist:
While Rollup is all about tree-shaking it still has its limits - to get max out of it modules should be authored in ESM (and lodash is still CJS-only) and code should avoid dynamic patterns (such as static properties assignments, top level property access, top level factory functions etc). JS is too dynamic and Rollup has to play "safe" and retain all code that it cannot be sure about its possible side effects - otherwise it could break somebody's production code.

You can check out a quick demo I've prepared - https://github.com/Andarist/slate-lodash-tree-shaking-demo . I've wanted to use `slate-react` package directly in the demo, but currently published version **doesnt** use lodash so I had to write simple input files. The situation there is 1to1 with what `react-slate` looks like - just boiled down to a much simpler repro case. Notice how `UseThrottled` **is not** in the final bundle, but `throttle` itself and whole lodash is still there.

CameronAckermanSEL:
Hi @Andarist , thanks for investigating this. The way `slate-react` imports `lodash/throttle` is using 
`import throttle from 'lodash/throttle';`

This is a small, but key, difference in the way the example project you provided is doing the import (`import { throttle } from 'lodash';`) because the path based import in the former statement does (or should be) providing tree shaking, even in commonjs. Other libraries such as material-ui leverage this path based importing to provide tree shaking to environments that can't get it from rollup or es6. 

I checked the `slate-react` bundle by hand (perhaps suboptimally :)) and I couldn't find any evidence of extraneous lodash functions being bundled in. 

If you investigate this again and find that indeed we are getting way more lodash code than intended, my next question is whether we should "vendor" (as in copy the lodash functions directly and commit them to the utils folder) that are making lodash/throttle function, so that we can maintain the exiting behavior. It seems like a good amount of research has gone into lodash/throttle and it's a tried and true implementation that hasn't changed in 3 years. Seems like a good bet to use that?

Thanks so much!

Andarist:
> because the path based import in the former statement does (or should be) providing tree shaking, even in commonjs

The situation here is actually slightly different. There is a small, but key, difference (😉) - in this very specific situation there is no tree-shaking involved. Tree-shaking is somewhat another term for dead code elimination and nothing gets shaken/eliminated here - you "just" use a direct import which makes tree-shaking irrelevant for this case as there is a low risk of other stuff being pulled with into the bundle with it. 

However, you are right about extra stuff not being pulled down for slate. A caveat in lodash's case is that:
```
require('lodash').throttle !== require('lodash/throttle')
```
because the main entry doesnt reuse those "direct imports". Those are copies. It's not much of a problem for you case though, as including just yet another throttle util is increasing the bundlesize anyway by "copying"/inlining its source (even if it gets reduced).

> If you investigate this again and find that indeed we are getting way more lodash code than intended

Yes - you still kinda are. For many, this could be an acceptable tradeoff though, so I'll let you decide if you want to change anything here. Currently `lodash/throttle` brings:
- `lodash/isObject`
- `lodash/isObjectLike`
- `lodash/isSymbol`
- `lodash/debounce`
- `lodash/now`
- `lodash/toNumber`
- `lodash/_baseGetTag `
- `lodash/_getRawTag `
- `lodash/_objectToString`
- `lodash/_root`
- `lodash/_freeGlobal`
- `lodash/_Symbol`

which adds up to a round 1024 bytes (min+gzipped), whereas the version included in this PR is merely 147 bytes.

> my next question is whether we should "vendor" (as in copy the lodash functions directly and commit them to the utils folder) that are making lodash/throttle function, so that we can maintain the exiting behavior

I don't think it would be worth it at all, given the fact that the current situation doesn't at least bring the whole lodash with it

Feel free to close this PR or ask me to adjust it anyhow - I don't mind either way.

TrySound:
👍

dylans:
Not sure if we're proceeding with this or not. Regardless I resolved the merge conflict with main but it's failing and will need some work to update it against the current codebase.

packages/slate-react/package.json
```diff
@@ -15,12 +15,10 @@
   ],
   "dependencies": {
     "@types/is-hotkey": "^0.1.1",
-    "@types/lodash": "^4.14.149",
     "direction": "^1.0.3",
     "is-hotkey": "^0.1.6",
     "is-plain-object": "^3.0.0",
-    "lodash": "^4.17.4",
-    "scroll-into-view-if-needed": "^2.2.20",
+    "scroll-into-view-if-needed": "^2.2.20"
```
jacekk, line 21: @Andarist @dylans A comma seems to be missing here :smile:

dylans, line 21:
```suggestion
    "scroll-into-view-if-needed": "^2.2.20",
```

dylans (changes requested):
To finally land this, we need to update  packages/slate-react/src/components/android/android-editable.tsx to also use the proposed throttle method.

@Andarist I know it's been a very long time, but if you want to update your PR, we'll land it as it does save on our default bundle size. Thanks!

dylans:
Closing due to inactivity. Feel free to reopen if you want to address the feedback and then I'll land it.

Issue #3531, opened 2020-02-28
Invalid links selection


Bacher:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

When I try to selection text of link by mouse cursor it's resetting or collapsing like on the video below.

Condition for bug:
* link must be last in a paragraph;
* selection should be started outside of link.

![Screen Recording 2020-02-28 at 15 50 16](https://user-images.githubusercontent.com/2598671/75550592-4549e200-5a43-11ea-8b91-ee5f0879d5cd.gif)

It can be reproduced in default example: https://www.slatejs.org/examples/links

Selection breaks because of `&#65279;` (Unicode BOM) symbols in the DOM:
![image](https://user-images.githubusercontent.com/2598671/75556835-fe161e00-5a4f-11ea-801f-0a00a193b5a6.png)

Slate: 0.57.1
Browser: Chrome / Safari / Firefox
OS: Mac / Linux / probably Windows

#### What's the expected behavior?

Expected that selection will extends following by cursor without drops.

Issue #3386, opened 2019-12-25
Selecting from empty paragraph backwards into paragraph content destroys selection


sirgru:
Press Enter at the end of the current paragraph, then try to select backwards
https://codesandbox.io/s/slate-reproductions-9glgw
![http://g.recordit.co/7SDriDsAtj.gif](http://g.recordit.co/7SDriDsAtj.gif)

Slate: 0.57.0
Browser: Chrome
OS: Windows

Issue #3333, opened 2019-12-17
In selection, anchor is always left of the focus


sirgru:
Bug (probably)

When a selection is made, even from right to left, anchor is always left of the focus. So isForward always returns true. 

#### What's the current behavior?
https://codesandbox.io/s/slate-reproductions-qm103
![http://g.recordit.co/WRwnEkbDQF.gif](http://g.recordit.co/WRwnEkbDQF.gif)
Slate: 0.55
Browser: Chrome
OS: Windows

#### What's the expected behavior?
Anchor should stay where selection begins, and focus should be where it ends.

kilrain:
I think the issue lies in these lines...  

https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugin/react-editor.ts#L467-#L470

jerkly:
Workaround at this time:
```typescript
const selection = window.getSelection();
const range = ReactEditor.toSlateRange(editor, selection);
const isBackward = Range.isBackward(selection);
```

Main problem is located in `onDOMSelectionChange:359` method. 
For some reason DOMSelection is "converted" to DOMRange, which is not store any information about direction of selection. As a result `Editor.selection` always has forward direction.
https://github.com/ianstormtaylor/slate/blob/1d7ab974292a3e831908a2ba0aab9fdd8a66fe10/packages/slate-react/src/components/editable.tsx#L353-L384

Issue #3143, opened 2019-11-28
Backwards keyboard selection expansion breaks when moving through an "empty" block


CameronAckermanSEL:
#### Do you want to request a feature or report a bug?
Bug

#### What's the current behavior?
On the rich text example, holding shift and using the arrow keys to expand selection backwards

Up arrow:
https://recordit.co/dyndP7b3M0

Right arrow:

https://recordit.co/PFHAl2raHy

This behavior works as expected when expanding selection forwards.

#### What's the expected behavior?
Selection continues to expand across empty blocks when using shift + arrow keys

PR #3594, opened 2020-04-08
Fix backwards selection

czechdave:

Fixing a bug with selection not working properly if it has backwards direction.

#### What's the new behavior?
Use `Selection` instead of `Range` so that the direction of selection is preserved.
Account for `data-slate-zero-width` nodes that have offset `0` in slate selection but `1` in DOM selection. Thanks to @JonathanWbn @kilrain for pointing this out.

Fixes: #3544
Fixes: #3531
Fixes: #3386
Fixes: #3333
Fixes: #3143
Fixes: #1969

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -145,13 +145,11 @@ export const Editable = (props: EditableProps) => {
       return
     }
 
-    const newDomRange = selection && ReactEditor.toDOMRange(editor, selection)
-
     // If the DOM selection is already correct, we're done.
     if (
       hasDomSelection &&
-      newDomRange &&
-      isRangeEqual(domSelection.getRangeAt(0), newDomRange)
+      selection &&
+      Range.equals(ReactEditor.toSlateRange(editor, domSelection), selection)
     ) {
       return
     }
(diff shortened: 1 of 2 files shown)


JonathanWbn:
Looks good and works! I'll go ahead and close https://github.com/ianstormtaylor/slate/pull/3574 in favor of this one.

vvav3:
Please merge, this is very annoying.

CameronAckermanSEL:
Thanks so much @czechdave , this is a huge improvement. Really appreciate the time you took to fix this issue!

plbabin:
Hello @czechdave!

I did track down an issue related to this little snippet, about detecting if the Dom selection is out of sync:

https://github.com/ianstormtaylor/slate/commit/67c4b3b330c3a62d626c4994d26f4526d18a4a1e#diff-0fa78153ea6f5b538c17b22d490c0ad6L152-R152

When inserting some text (or selecting all then delete) the cursor is moved out of the box (only on Safari).

Screen Recording 2020-05-08 at 11 31 AM

With the original code, the selection was restore by Slate, but now it thinks that it's all fine

Any idea on how to restore the original behavior, without impacting the fix you did with the selection direction?

czechdave:
Interesting @plbabin Do you have any details about your tracking down?

What's the difference between Slate's selection and window selection after each operation you're applying? I suspect your node becomes a zero-width node during your operations which throws the selection off?

Given that it only happens in Safari I would assume it's a Safari bug rather then Slate issue but there might be a way to hack around it for Safari if we have more details.

PR #3602, opened 2020-04-09
Add edge case in normalize to prevent excessive remove_node

### LINKED ISSUE #3600: Unexpected cursor behavior on ENTER key after mark
Bug

1. Go to the [rich text example](https://www.slatejs.org/examples/richtext)
2. Place your cursor after a mark range, e.g. after the italicized word "_much_"
3. Press ENTER key
4. Cursor remains in place even though line break is added

bug

Expected behavior is shown in the video above: when ENTER is pressed after non-marked text, the cursor follows the line break to the next line.

pzhine:
Fixes #3600

#### What's the new behavior?
When ENTER key is pressed after marked text (e.g. bold, italics), cursor advances to the following line and mark is maintained, as expected.

An edge case was added to the `normalizeNode` method, which by design cleans up empty nodes. The edge case prevents this cleanup if the current editor focus is on the empty node. This is the case when a `split_node` happens after an ENTER key press when the focus was at the end of a node. See video in #3600 for broken behavior and video below for corrected behavior.

fix

### DIFF
--- a/packages/slate/src/create-editor.ts
+++ b/packages/slate/src/create-editor.ts
@@ -221,6 +221,11 @@ export const createEditor = (): Editor => {
         const isInlineOrText =
           Text.isText(child) ||
           (Element.isElement(child) && editor.isInline(child))
+        let prevIsFocused = false
+        if (editor.selection) {
+          const [focusNode] = Editor.node(editor, editor.selection.focus)
+          prevIsFocused = focusNode === prev
+        }
 
         // Only allow block nodes in the top-level children and parent blocks
         // that only contain block nodes. Similarly, only allow inline nodes in
@@ -254,7 +259,9 @@ export const createEditor = (): Editor => {
             if (Text.equals(child, prev, { loose: true })) {
               Transforms.mergeNodes(editor, { at: path.concat(n), voids: true })
               n--
-            } else if (prev.text === '') {
+            } else if (prev.text === '' && !prevIsFocused) {
+              // Only clean up the empty node if it doesn't currently have focus.
+              // This prevents an overly aggressive removal when split happens between nodes.
               Transforms.removeNodes(editor, {
                 at: path.concat(n - 1),
                 voids: true,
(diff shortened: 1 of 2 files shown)


majelbstoat:
It would be cool to see a failing test that this fixed.

CameronAckermanSEL:
@pzhine Thanks so much for the PR! Really appreciate it. This looks like it'll be a nice behavior to fix. 
I do agree with @majelbstoat though, it'd be nice to have a test case for this if we can because it'll let us prevent this bug in the future.

Thanks!

pzhine:
@majelbstoat @CameronAckermanSEL 
Test should be good to go. Run it on `master` to see it failing.

majelbstoat:
Confirmed that this test fails on master@a1ce4477 👍

CameronAckermanSEL:
Thanks @pzhine . This looks great! I'm planning on merging it and releasing it sometime this week. :)

rskopal:
I think there is a problem with this solution. If you return back to the original position after the split and press a DELETE key, the previous node is deleted:
slate-pr3602-problem

pzhine:
Ah, good catch @rskopal ! I'll look into that shortly

spocke:
I don't think this should be fixed in the normalizer like this the problem is in splitNodes.

This line can be changed to path ref to the parent element:
https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/transforms/node.ts#L717

Then this could be changed to as Editor.start on that path:
https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/transforms/node.ts#L756

This way the caret will be moved over to the next line when split at the selection.

The next thing is to use the pending editor marks state so when you start typing the new text will be in that previous marks before the enter was pressed. This should probably go into insertBreak:
https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/create-editor.ts#L147

Something like:
```
let marks = editor.marks;
Transforms.splitNodes(editor, { always: true })
editor.marks = marks;
```

But maybe it's not something everyone would want to it might be a thing to leave up to the implementation.

dylans (changes requested):
Is there interest in addressing the feedback in https://github.com/ianstormtaylor/slate/pull/3602#issuecomment-735603123 or should we close this one?

dylans:
The underlying issue of this PR appears to have already been fixed, so I'm closing this out. Feel free to reopen if I missed a scenario where this is still an issue.

PR #3606, opened 2020-04-11
v0.47 documentation and dependency updates

kswenson:

For those of us still supporting a 0.47 installation, it is useful to be able to consult the 0.47 documentation and build the 0.47 examples. As things stand (at least for me) the 0.47 examples don't build and there are a number of links from 0.47-branch documentation to master-branch code and documentation that should be to the corresponding 0.47-branch target. This PR attempts to address these issues by fixing the dependencies and updating the documentation where appropriate.

#### What's the new behavior?

v0.47 branch tests pass, examples build, and documentation links target v0.47 pages where appropriate.

1. adds missing `webpack-cli` dependency (1.5.3 which I believe is the latest version that supports Webpack 3.11.0 used by the v0.47 branch)
2. downgrades `webpack-dev-server` dependency (to 2.11.5 which I believe is the latest version that supports Webpack 3.11.0 used by the v0.47 branch)
3. documentation uses GitHub-relative links where appropriate
4. documentation uses docs-relative links where appropriate
5. documentation uses full-path branch urls for links between GitHub and docs site
6. adds `./site` to list of directories removed by `clean` script; this directory remains in place when switching from the `master` branch and prevents the `lint` script from succeeding, for instance.

Fixes: #
Reviewers: @


CameronAckermanSEL:
Thanks so much for fixing these problems!

PR #3608, opened 2020-04-11
Fix(placeholder): allow ReactNode placeholder, allow multiline wrappi…

BJvdA:
…ng of placeholder, remove limiting opacity styling

Improvement of `placeholder` prop

#### What's the new behavior?

Before, the placeholder would always be one line and overflow the editor incase it's too long.
Now it will wrap, while still not being focusable. 
Also changed the type for the prop to include `ReactNode`.

Remove `display: inline-block`, add `user-select: none`, replace `white-space: nowrap` with `prewrap`.

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -1041,7 +1041,7 @@ const setFragmentData = (
   let attach = contents.childNodes[0] as HTMLElement
 
   // Make sure attach is non-empty, since empty nodes will not get copied.
-  contents.childNodes.forEach((node) => {
+  contents.childNodes.forEach(node => {
     if (node.textContent && node.textContent.trim() !== '') {
       attach = node as HTMLElement
     }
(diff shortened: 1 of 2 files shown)


CameronAckermanSEL:
Hi @BJvdA  thank you very much for the contribution. The changes requested make sense, however, I have two findings:
1. I think we should leave the opacity on. Most users expect there to be fainter text for the placeholder and if the api had made the decision previously to have that opacity, this might be a contract users are depending on.
2. I noticed there is a minor defect observable on this pr's netlify deployment example where the user can place the cursor at the end of the placeholder text. This is reproducible in the rich text example by removing all of the text:
image


Thanks again! Really appreciate the time you took to donate this improvement.

dylans (changes requested):
@BJvdA any interest in revisiting/updating this PR based on previous feedback or should we close it out?

dylans:

PR #3616, opened 2020-04-16
Do not store focus/blur in history

Andarist:
#### Description of a bug
loosing focus and crash

Notice how after text gets reinserted the cursor is gone and if we call `.undo` once again after that, the error pops up.

#### What's the new behavior?

Cursor gets preserved after such `undo` and calling `undo` after that doesn't cause an error.

This happens because
1. a batch to be undone contains: [removeText, focus]
2. this gets inversed to [blur, insertText]
3. after this gets processed the cursor is gone 😢 
4. we try to undo once more - undo batch contains now: [selectAll]
5. when processing this it crashes on: https://github.com/ianstormtaylor/slate/blob/16ff44d0566889a843a346215d3fb7621fc0ed8c/packages/slate/src/interfaces/editor.ts#L1438-L1445

I've thought about various possible fixes for this - at various stages/layers - and I came to the conclusion that it really doesn't make sense to store both "focus" and "blur" set_selections (or does it?). So if we just filter them out this problem won't happen because there won't be any attempts to undo them. 

### DIFF
--- a/packages/slate-history/src/with-history.ts
+++ b/packages/slate-history/src/with-history.ts
@@ -44,16 +44,7 @@ export const withHistory = <T extends Editor>(editor: T) => {
           const inverseOps = batch.map(Operation.inverse).reverse()
 
           for (const op of inverseOps) {
-            // If the final operation is deselecting the editor, skip it. This is
-            if (
-              op === inverseOps[inverseOps.length - 1] &&
-              op.type === 'set_selection' &&
-              op.newProperties == null
-            ) {
-              continue
-            } else {
-              e.apply(op)
-            }
+            e.apply(op)
           }
         })
       })
@@ -150,7 +141,10 @@ const shouldMerge = (op: Operation, prev: Operation | undefined): boolean => {
  */
 
 const shouldSave = (op: Operation, prev: Operation | undefined): boolean => {
-  if (op.type === 'set_selection' && op.newProperties == null) {
+  if (
+    op.type === 'set_selection' &&
+    (op.properties == null || op.newProperties == null)
+  ) {
     return false
   }
 
--- a/packages/slate-history/test/undo/cursor/keep_after_focus_and_remove_text_undo.js
+++ b/packages/slate-history/test/undo/cursor/keep_after_focus_and_remove_text_undo.js
@@ -1,0 +1,49 @@
+/** @jsx jsx */
+
+import assert from 'assert'
+import { Transforms, Editor } from 'slate'
+import { jsx } from '../..'
+
+export const run = editor => {
+  // focus at the end
+  Transforms.select(editor, {
+    anchor: { path: [0, 0], offset: 5 },
+    focus: { path: [0, 0], offset: 5 },
+  })
+  // select all
+  Transforms.select(editor, {
+    anchor: { path: [0, 0], offset: 5 },
+    focus: { path: [0, 0], offset: 0 },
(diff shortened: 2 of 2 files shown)


packages/slate-history/test/undo/cursor/keep_after_focus_and_remove_text_undo.js
```diff
@@ -0,0 +1,49 @@
...
+  })
+}
+
+export const input = (
+  <editor>
+    <block>Hello</block>
+  </editor>
+)
+
+export const output = {
```
Andarist, line 35: I've tried to use hyperscript for this, but couldn't make it work because the expected `output.children` have contained an extra empty text node for:
```
export const output = (
  <editor>
    <block>Hello</block>
    <cursor />
  </editor>
)
```
and I'm not sure how to fix this.

packages/slate-history/src/with-history.ts
```diff
@@ -44,16 +44,7 @@ export const withHistory = <T extends Editor>(editor: T) => {
           const inverseOps = batch.map(Operation.inverse).reverse()
 
           for (const op of inverseOps) {
-            // If the final operation is deselecting the editor, skip it. This is
```
Andarist, line 47: because "focus" (so "blur" when reversed) doesn't get stored in the history now, this if statement becomes obsolete

Andarist:
@CameronAckermanSEL friendly 🏓

CameronAckermanSEL:
Hey @Andarist , this contribution is really appreciated. I'm still mulling over the consequences of making this change because it's been a long standing (albeit _inconvenient_ in my opinion) behavior and I'm not sure what the ripple effects will be. This might be something @ianstormtaylor  has to review.

Andarist:
Ye - definitely hard to predict all the consequences of this, but I would say that on the high level the reasoning behind the change is good. You can also always release this as a new minor as then some breaking changes are OK for 0.x versions. If this causes any regressions - feel free to ping me and I will try to help with them. 

Of course, I'm OK with waiting for @ianstormtaylor's review but from what I've seen he has not been active in this repo lately.

hadrysmateusz:
This issue has been a source of a million headaches for me. It's great to see someone tackle it :)

Andarist:
@CameronAckermanSEL any updated on this? Is there anything I could do to move this forward anyhow?

interstates21:
@ianstormtaylor any update on this?

thesunny:
@Andarist Have you been using a version of the editor (including the new TypeScript stuff) with this PR in it already? If you are fairly confident that this plays nice, it might be easier for me to do a quick check and then merge it. It sounds like it makes sense to me.

I don't have a lot of bandwidth to check it in different scenarios myself. Unfortunately, Slate takes a lot of work IMO to check PRs because you can't easily test all the potential cross browser issues.

Either way, we would want to wait a bit at least because we just published a "next" release with a bunch of changes.

Andarist:
Yes, I'm using this patch in our application and didn't found any problems with it. If you get any problems caused by this after this gets published - feel free to ping me and I will try to do my best to help out. In the worst-case scenario, we'll have a new test case then :P

TheSpyder:
Well this is interesting. I didn't notice this was here when I logged #3756. It's certainly not the only potential cause of the incomplete operation error when performing undo but hopefully it fixed the problem for a lot of people.

PR #3620, opened 2020-04-17
Make setFragmentData and getFragment pluggable in ReactEditor

majelbstoat:

improving a _feature_, by fixing a regression.

#### What's the new behavior?

Previously (#1429), the equivalent of setFragmentData was pluggable, which meant we could alter the fragment used for a copy, cut, or drag. During a subsequent rearchitecture, it became impossible to do that without fully reimplementing setFragmentData.

This PR moves setFragmentData onto ReactEditor and makes it pluggable, such that a user can now do:

```ts
const withMyPlugin = <T extends ReactEditor>(
  e: T,
) => {
  const { setFragmentData } = e

  e.setFragmentData = (data: DataTransfer) => {
    console.log("My handler")
    setFragmentData(data)
  }
}
```

It also adds a new method `getFragment` to Editor and makes it pluggable in the same way. The default implementation is simply `Node.fragment(editor, selection)`.

Finally, setFragmentData is modified slightly to make a call to `editor.getFragment()` instead of `Node.fragment(editor, selection)`.

This facilitates use-cases where you don't need to reimplement the entirety of copying from the DOM, but just want to modify the encoded fragment that slate attaches. Specifically, for a use-case I have, this allows me to just modify the props before copy, while allowing slate to handle all the DOM cloning as standard.

Fixes: #1410

### DIFF
--- a/packages/slate-react/src/plugin/with-react.ts
+++ b/packages/slate-react/src/plugin/with-react.ts
@@ -1,9 +1,10 @@
 import ReactDOM from 'react-dom'
-import { Editor, Node, Path, Operation, Transforms } from 'slate'
+import { Editor, Node, Path, Operation, Transforms, Range } from 'slate'
 
 import { ReactEditor } from './react-editor'
 import { Key } from '../utils/key'
 import { EDITOR_TO_ON_CHANGE, NODE_TO_KEY } from '../utils/weak-maps'
+import { isDOMText, getPlainText } from '../utils/dom'
 
 /**
  * `withReact` adds React and DOM specific behaviors to the editor.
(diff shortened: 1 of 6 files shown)


packages/slate-react/src/plugin/with-react.ts
```diff
@@ -56,6 +57,91 @@ export const withReact = <T extends Editor>(editor: T) => {
     }
   }
 
+  e.setFragmentData = (data: DataTransfer) => {
```
majelbstoat, line 60: This method is an exact copy of `setFragmentData` with `editor` replaced by `e`, and one other change noted below:

packages/slate-react/src/plugin/with-react.ts
```diff
@@ -56,6 +57,91 @@ export const withReact = <T extends Editor>(editor: T) => {
...
+      const span = document.createElement('span')
+      // COMPAT: In Chrome and Safari, if we don't add the `white-space` style
+      // then leading and trailing spaces will be ignored. (2017/09/21)
+      span.style.whiteSpace = 'pre'
+      span.appendChild(attach)
+      contents.appendChild(span)
+      attach = span
+    }
+
+    const fragment = e.getFragment()
```
majelbstoat, line 129: This is the only consequential change. This line used to be `Node.getFragment(editor, selection)`

packages/slate-react/src/utils/dom.ts
```diff
@@ -145,3 +145,32 @@ export const getEditableChild = (
...
 }
+
+/**
+ * Get a plaintext representation of the content of a node, accounting for block
+ * elements which get a newline appended.
+ *
+ * The domNode must be attached to the DOM.
+ */
+
+export const getPlainText = (domNode: DOMNode) => {
```
majelbstoat, line 156: Moved this into utils, so it could be accessed more easily, and seemed like a reasonable home.

packages/slate/src/create-editor.ts
```diff
@@ -135,6 +135,15 @@ export const createEditor = (): Editor => {
       }
     },
 
+    getFragment: () => {
```
majelbstoat, line 138: Second new piece of code allowing just the fragment to be changed, if you don't need to mess with the dom copying stuff. (This is actually the only piece I need for my specific use-case, but it seemed more generally useful to also allow pluggability of the entire dataTransfer.)

CameronAckermanSEL:
Thanks @majelbstoat ! This looks great to me!

skogsmaskin:
I think you forgot to declare `getFragment: () => Descendant[]` in the `ReactEditor` interface def. (packages/slate-react/src/plugin/react-editor.ts) 

ReactEditor and Editor are now incompatible.

CameronAckermanSEL:
@skogsmaskin Thanks for the double check. From what I can see, this method is defined on the Editor interface, which ReactEditor extends:
https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/interfaces/editor.ts#L55

https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugin/react-editor.ts#L29

Might be something wrong with this because of name collision with the `Editor` object and the `Editor` interface?

skogsmaskin:
@CameronAckermanSEL - hmm...yes I see that now. That's weird, I got an error yesterday unless I defined the method in the ReactEditor interface, but today I can't seem to reproduce it. I suppose it must have been something wrong in my end! Sorry for the noise.

skogsmaskin:
This should work, right?

```
<T extends ReactEditor>(editor: T) => {
  const {setFragmentData, getFragment} = editor
  editor.setFragmentData = (data: DataTransfer) => {
    debug('set fragment data')
    setFragmentData(data)
  }
  editor.getFragment = () => {
    debug('get fragment data')
    getFragment()
  }
  return editor
}
```

Here I get an error saying that `Property 'getFragment' does not exist on type 'T'`

majelbstoat:
Lemme double check, but I think you might be right.

majelbstoat:
I went and checked this locally @skogsmaskin, and I can't reproduce. You are right that that fragment should work, but for me, it does. Meaning, the destructuring works, and it is correctly typed.

skogsmaskin:
Thanks for checking @majelbstoat - I'll have another look a soon. It could very well be my code, I'm currently in a linked state with forks and what not. :octopus:

lcswillems:
Because of this commit, copy-pasting void nodes doesn't work anymore for me.

@majelbstoat Did you change the logic of the `setFragmentData` and `getFragment`? I can't identify it from the diff

majelbstoat:
Nope, didn't change the logic. I noted the consequential changes here: https://github.com/ianstormtaylor/slate/pull/3620/files

What doesn't work about void nodes? And have you isolated it to this specific commit?

lcswillems:
Thanks for your answer!

Copy/pasting of void nodes works correctly for me on Slate 0.57.2 but not on 0.57.3, and only this commit might have altered copy/pasting

majelbstoat:
Sorry for the trouble. Can you open a new issue for this? A code sandbox example where we could compare the behaviour would be helpful. A failing text case would be ideal (though it's kind of hard to write them for ReactEditor ATM).

lcswillems:
I will try to write one tonight, but not sure I will succeed... Thanks for your quick answers :)

evasteingrims:
> I will try to write one tonight, but not sure I will succeed... Thanks for your quick answers :)

Did you create the issue yet? Copy/pasting a void node is not working for me either and I have narrowed it down to this commit

lcswillems:
No, I didn't have the time...

Issue #3196, opened 2019-12-02
Incorrect cursor after composition end when using IME on Safari


minwe:
#### Do you want to request a _feature_ or report a _bug_?

Report bug.

#### What's the current behavior?

![ime-safari](https://user-images.githubusercontent.com/4370770/69948701-10ed8500-152b-11ea-8170-1462d081bfd2.gif)

The cursor is going to the start of paragraph after IME composition end on Safari.

Slate: 0.50.6
Browser: Safari
```
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.3 Safari/605.1.15
```
OS: Mac

#### What's the expected behavior?

The cursor should on the end of inserted text.

PR #3626, opened 2020-04-21
Fix safari selection issue with IME

yyjhao:
#### Fixing a _bug_

This fixes https://github.com/ianstormtaylor/slate/issues/3196

#### What's the new behavior?

slatefix

It looks like if we override the default behavior of `insertFromComposition` in Safari, the selection will jump to the start of the text node. We can fix that by not overriding it and instead handle the input similar to Chrome which doesn't fire the event.

Fixes: https://github.com/ianstormtaylor/slate/issues/3196
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -213,7 +213,12 @@ export const Editable = (props: EditableProps) => {
         // cancelled. Let them through and wait for the composition to end.
         if (
           type === 'insertCompositionText' ||
-          type === 'deleteCompositionText'
+          type === 'deleteCompositionText' ||
+          // COMPAT: in Safari, if we override the insertFromComposition
+          // event, the selection will go to the beginning. To fix this
+          // we don't handle the event here but instead handle it in
+          // onCompositionEnd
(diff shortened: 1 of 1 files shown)


tizee:
I had applied this patch to my forked local repo and it did solve the bug that the cursor was misplaced when using IME in Safari. However, it brought another bug leading to duplicate characters. 

slate-bug

To fix up this,  we also need to remove the case `insertFromComposition` from `onDOMBeforeInput` handler in order to not to handle the insertFromComposition like Chrome.
```js
// slate-react/src/components/editable.tsx
...
const onDOMBeforeInput = useCallback(
         ...
          // do not fire insertFromComposition
          // case 'insertFromComposition':
          case 'insertFromDrop':
          case 'insertFromPaste':
          case 'insertFromYank':
          case 'insertReplacementText':
          case 'insertText': {
...
```
Finally, it works.
slate-bug-fixed

pubuzhixing8:
#3698 this is my idea for same program @yyjhao

ianstormtaylor:
I believe this is solved by https://github.com/ianstormtaylor/slate/pull/3698, so I'm going to close but feel free to comment if not and we can re-open. Thanks!

PR #3632, opened 2020-04-23
docs(resources): add product and editor

irmerk:
Proposing Clause as an example of usage, which is built on top of the Accord Project Markdown Editor - based on Slate.


CameronAckermanSEL:
Thanks @irmerk !

PR #3639, opened 2020-04-26
Fix: parentElement changed to parentNode because of IE 11 issues

Opty1712:
####  _bug_

Calling `parentElement` from `textNode` returns undefined in IE11, so IE11 crushes. For example it crushes on `scrollIntoView` from 'scroll-into-view-if-needed' because `target` is `undefined`
```
!target.ownerDocument.documentElement.contains(target);
```

#### What's the new behavior?

This PR changes `parentElement` to more stable `parentNode`

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -163,7 +163,7 @@ export const Editable = (props: EditableProps) => {
 
     if (newDomRange) {
       domSelection.addRange(newDomRange!)
-      const leafEl = newDomRange.startContainer.parentElement!
+      const leafEl = newDomRange.startContainer.parentNode!
       scrollIntoView(leafEl, { scrollMode: 'if-needed' })
     }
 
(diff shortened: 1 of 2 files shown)


Opty1712:
Here is the fix for IE if anyone needs it

```js
(function () {
  'use strict';

  var elem = document.createElement('p');
  elem.appendChild(document.createTextNode('string'));
  document.body.appendChild(elem);

  var node = document.querySelector('p');
  var isBrokenParentElement = !node.childNodes[0].parentElement;

  document.body.removeChild(elem);

  if (isBrokenParentElement) {
    var parentElementProperty = {
      configurable: false,
      enumerable: false,
      get: function () {
        return this.parentNode || null;
      },
    };

    var properties = [
      Attr.prototype,
      Text.prototype,
      Element.prototype,
      Document.prototype,
    ];

    for (var i = 0; i < properties.length; i++) {
      Object.defineProperty(
        properties[i],
        'parentElement',
        parentElementProperty
      );
    }
  }
})();

```

ianstormtaylor:
Hey @Opty1712 thanks for submitting this. We don't support IE11 so unfortunately I'd rather not change things in core to work around its limitations—there are likely many other things that are broken that would need to be fixed. Thanks for understanding.

PR #3644, opened 2020-05-01
Fixes #3458 Editor.positions() bug, refactors advance loop to make more clear, and documents algorithm to improve DX

### LINKED ISSUE #3458: Arrow navigation issue with single-character text node adjacent to inline element

Bug

https://codesandbox.io/s/slate-reproductions-qdi9i

Click anywhere in the first link. Use the right/left arrow keys to move around. You cannot move the cursor to the left of the first x or the right of the last x. Clicking before/after them works fine though.

If the text nodes have more than 1 character, they work fine (see the second editor in the example)

Slate: 0.57.1
Browser: Chrome
OS: Mac

Arrow keys should navigate to the end of the line

beorn:
Editor.positions() would miss a position in at least the case where the final node has a single character.

The code of Editor.positions() has some pretty darn complicated logic and could do with some more documentation, but that's a separate issue, and this PR just tries to fix this one correctness issue.

This does not fix all of the issues mentioned in #3458 - but it does fix the case of a missing last position.

Fixing a bug.

- [x] The new code matches the existing patterns and styles. (not sure?)

Fixes: #3458 (partly)
Reviewers: @


packages/slate/src/interfaces/editor.ts
```diff
@@ -1049,12 +1058,13 @@ export const Editor = {
 
       if (Text.isText(node)) {
         const isFirst = Path.equals(path, first.path)
-        available = node.text.length
-        offset = reverse ? available : 0
 
         if (isFirst) {
-          available = reverse ? first.offset : available - first.offset
-          offset = first.offset
+          available = reverse ? first.offset : node.text.length - first.offset
```
majelbstoat, line 1063: These 5 lines don't appear to be logically different from what came before. Why the change?

beorn: No they aren't.  I just thought it expressed the intent better to show the two cases - isFirst or not - instead of assuming !isFirst and right after overrdiing both available and offset again if isFirst.  It is probably more performant too.  But it's totally optional whether to do it this way or not.  The main change is the textOverflowed flag - which I admit is a bit hackish.

This whole function is rather complex - I assume with some deep thought it might be possible to express it more clearly.

Also, I made a lot of comments for my own understanding of this function, and I thought perhaps it'd be helpful to try to add these as documentation for others to more easily be able to trace through the flow of this function.  Happy to do so, but as I'm not the original author, it'd be good to have someone that's a bit more familiar do it, or oversee/fix my comments.

beorn: @majelbstoat I just submitted another commit that adds documentation and renames some variables to make the algorithm more clear, while keeping the logic untouched.

I think it'll help more people understand the algorithm, and therefore empower them to troubleshoot and refactor it.

It's such a fundamental algorithm used by many functions inside of Slate, and it's really rather complicated, so I think it's worth making it more approachable in this way.

packages/slate/src/interfaces/editor.ts
```diff
@@ -987,100 +986,166 @@ export const Editor = {
...
-      // Subtract the distance traveled from the available text.
-      available = available - distance!
-      // If the available had room to spare, reset the distance so that it will
-      // advance again next time. Otherwise, set it to the overflow amount.
-      distance = available >= 0 ? null : 0 - available
+      // Advance part 2: Advance leafText.
+
+      // Step leafText offset and remaining counters by distance.
+      leafTextOffset = reverse ? leafTextOffset - distance : leafTextOffset + distance
+      leafTextRemaining = leafTextRemaining - distance!
```
majelbstoat, line 1040: Is the `!` necessary here?

packages/slate/src/interfaces/editor.ts
```diff
@@ -987,100 +986,166 @@ export const Editor = {
...
+    // of block nodes in blockText, and text nodes in leafText.
+    // Exploits the fact that nodes are sequenced in such a way that we first
+    // encounter the block node, then all of its text nodes, so when iterating
+    // through the blockText and leafText we just need to remember a window of
+    // one block node and leaf node, respectively.
     for (const [node, path] of Editor.nodes(editor, { at, reverse })) {
       if (Element.isElement(node)) {
-        // Void nodes are a special case, since we don't want to iterate over
-        // their content. We instead always just yield their first point.
+        // Void nodes are a special case in that tey are not iterated over,
```
majelbstoat, line 1064: nit: typo

packages/slate/src/interfaces/editor.ts
```diff
@@ -987,100 +986,166 @@ export const Editor = {
...
-        }
+        // Inline element nodes are ignored as they don't themselves
+        // contribute to blockText or leafText - their parent and
+        // children do, respectively.
+        if (editor.isInline(node)) continue
 
+        // Block element node - set blockText to its text content.
         if (Editor.hasInlines(editor, node)) {
+          // We always exhaust block nodes before encountering a new one:
+          //   console.assert(blockText === '',
```
majelbstoat, line 1079: is this meant to be commented?

packages/slate/src/interfaces/editor.ts
```diff
@@ -987,100 +986,166 @@ export const Editor = {
...
 
+      // We are on new text leaf node.  Iterate through its text
+      // content, yielding positions every `distance` offset.
       if (Text.isText(node)) {
         const isFirst = Path.equals(path, first.path)
-        available = node.text.length
-        offset = reverse ? available : 0
 
+        // We always exhaust text nodes before encountering a new one:
+        //   console.assert(leafTextRemaining <= 0,
```
majelbstoat, line 1107: is this meant to be commented?

packages/slate/src/interfaces/editor.ts
```diff
@@ -987,100 +986,166 @@ export const Editor = {
...
+          // Successfully took a `distance` step through text on both a
+          // block and leaf level, so yield another position in this leaf node.
+          yield { path, offset: leafTextOffset }
         }
 
         isNewBlock = false
       }
     }
+    // Upon completion, text is exahusted on both block and leaf level:
+    //   console.assert(leafTextRemaining <= 0, "leafText wasn't exhausted")
```
majelbstoat, line 1147: commented out

packages/slate/src/interfaces/editor.ts
beorn: @majelbstoat It irked me too much, so I refactored the `advance` part of the function to make the logic more clear and remove some smelly code, like the flag.  I'm pretty much done tweaking this function now, so it's safe to code review.  Sorry about the repeated commits.

packages/slate/src/interfaces/editor.ts
beorn: It was there before as `distance` was sometimes set to null, but that's no longer the case so I'll remove it.

packages/slate/src/interfaces/editor.ts
beorn: No problemey.  Fixed. :)

packages/slate/src/interfaces/editor.ts
beorn: Yes, I wanted to make the assertions that the algorithm makes explicit - it assumes the next node returned is a text node or a block node, and we'll always exhaust block nodes before we encounter a new one, etc.  But I was reluctant to leave the statements in there it might impact performance unnecessarily.  Should I uncomment them?

packages/slate/src/interfaces/editor.ts
beorn: Same as above.

packages/slate/src/interfaces/editor.ts
beorn: Same as above.

beorn:
Ok, should be all ready for review.

BrentFarese:
I'm not sure if the comments and changes to the function make things clearer.  Where is the actual bug fix in the code?  It might be clearer to leave the function as it was stylistically but to put the bug fix in place and perhaps have a more in-depth comment before the function to broadly state what it does/how it works?  Or, you could put the comments in the docs as that might be a better place to describe the algorithm vs. source-code comments.

beorn:
@BrentFarese You can look at the individual commits and actually get a better sense of the different aspects of the change, from the simple bug fix, documentation, and the minor refactoring:

**(1) https://github.com/ianstormtaylor/slate/pull/3644/commits/9b49fd24ef2bf461b418076e870906322444a0b0 + https://github.com/ianstormtaylor/slate/pull/3644/commits/de05c1b390c665b596d96ee3dbef89ebb4fc0d27 is the bug fix with a new passing test case.**

At this point, still, if you try to trace what's going on, especially around the edge cases, your head will explode. :)  It is just very hard to look at the code and feel confident it's doing what it should do.  Tests are critical, but as was shown, they don't catch bugs you don't test for.

**(2) https://github.com/ianstormtaylor/slate/pull/3644/commits/afe5055bf6a20d98f735638b05cf9713fef3b1b7 + https://github.com/ianstormtaylor/slate/pull/3644/commits/6e176ca0e452c2123d439e46bd56cc7a18961878 adds documentation and tries to explain what the algorithm does**, and a few (commented out) console.asserts that show the underlying assumptions that the algorithm make to work (and you can verify if you uncomment them and run the code).  It's still complicated, but now it's documented, so my hope was that this would enable others to more quickly "get" the algorithm, necessary for bug fixing and tracing.

**(3) https://github.com/ianstormtaylor/slate/pull/3644/commits/b449730c17b8b8e4730f295f3694ec0231e178ff + https://github.com/ianstormtaylor/slate/pull/3644/commits/4259e04326acf931c1b69ff5753dd474a9190051 refactors the function** a bit to try to make it more clear:

(a) It pulls together the logic of the `advance` function with the while loop, refactoring out the non-essential logic part (getDistance()) as a helper.  This makes it easier to see what the entire logic of this inner loop is, while keeping it very small, and it relieves the necessity of having a (very hackish) flag to signal when the text overflowed.

(b) Instead makes a few renames to make it more clear that we are iterating over two instantiations of text, collected at the block level and the text leaf level, and that we are moving them forward in lock-step fashion. Instead of talking about distance, available, string, offset - which does not give any indication as to whether they refer to data structures used by the leaf- or block-level iteration, I used these names:
- `blockText` instead of string
- `leafTextRemaining` instead of available
- `leafTextOffset` instead of offset
- `distance` the distance the leafText iteration still has to catch up to blockText

(c) Overall there are several other small improvements:
- `distance` is no longer sometimes null and sometimes a number
- leafText variables are no longer initialized and then overridden if `isFirst` (confusing since an either-or not an and-if situation), they are just set differently depending on whether this is the first node or not
- `isNewBlock` is reset to false closer to the condition that uses it (just a subjective preference)
- the overall structure of iterating through all nodes, and then having two big branches depending on if we're on an element node or leaf text node, is arguably a bit more clear

Especially if looking at the refactoring change, it is probably easier to look at the old function and the new function side-by-side.

**THOUGHTS**
While I think the function+algorithm after these changes is easier to reason about, but I still think it could be made more clear.  In particular by making the (currently implicit) leaf and block node iterations explicit without breaking out of the inner loop - perhaps it would be possible to yield the next block and next leaf, or retrieve leaf nodes form block nodes and have two nested for loops instead:

    for each block node:
        for each text node in block:
            yield position

At least then it's explicit whether the next node you're dealing with is a block node or text node.

BrentFarese:
Thanks for the explanation!  A couple of comments/feedback that might be included in this fix:

1. I don't think `void` nodes are even being captured in the `*positions` function.  You can see that `Editor.nodes` is called without `voids` being mentioned at all, so I don't think `Editor.nodes` is going to return any `void` nodes?  So, the `*positions` API might not be giving positions of `void` nodes at all.  This can simply be fixed by giving the `*positions` function a `void?: boolean` parameter that defaults to `false` (seems to be very common in Slate core) and then is passed directly to `Editor.nodes`.
2. I think we might want to get rid of the `line` unit in the `*positions` API?  It is no different than the `block` unit actually and its inclusion in the API is misleading in my view.
3. I would consider maybe removing the `console.asserts` and tightening up the comments a bit.  There is a lot there which seems atypical for Slate core.

beorn:
> Thanks for the explanation! A couple of comments/feedback that might be included in this fix:
> 
> 1. I don't think `void` nodes are even being captured in the `*positions` function.  You can see that `Editor.nodes` is called without `voids` being mentioned at all, so I don't think `Editor.nodes` is going to return any `void` nodes?  So, the `*positions` API might not be giving positions of `void` nodes at all.  This can simply be fixed by giving the `*positions` function a `void?: boolean` parameter that defaults to `false` (seems to be very common in Slate core) and then is passed directly to `Editor.nodes`.

I didn't notice that.  I'm surprised that the common case is not returning positions for voids.

> 2. I think we might want to get rid of the `line` unit in the `*positions` API?  It is no different than the `block` unit actually and its inclusion in the API is misleading in my view.

I thought blocks are element nodes that are not inline (so a <P>), whereas line is based on the actual line being rendered in the DOM - no?  I can see that a "move-to-end-of-line" selection transformation would want to find the end of line, and such would need `*positions` for that.

> 3. I would consider maybe removing the `console.asserts` and tightening up the comments a bit.  There is a lot there which seems atypical for Slate core.

There's some redundancy in the explanations too - often both in summary form and in-context - so that could perhaps be removed (but which?).  At the same time, I felt it was a particularly difficult algorithm to navigate, and without the comments there'd be a lot of questions unanswered.  If it was as straightforward as a double for-loop, it probably wouldn't need much documentation.

BrentFarese:
For (2), take a look at the `*positions` implementation and the tests for line vs. block.  They are duplicative as far as I can tell and are 100% the same, but the API gives the impression that you can iterate over "line" positions when you really can't.  You can only really iterate over block positions.  Let me know if you agree.

beorn:
You are right!  But that means that none of the `unit='line'` logic works, since this function is at the bottom of it all: `Editor.delete` => `Editor.before` => `Editor.positions`.

How would we even do it?  It would have to delegate to the browser somehow for that.

BrentFarese:
> You are right! But that means that none of the `unit='line'` logic works, since this function is at the bottom of it all: `Editor.delete` => `Editor.before` => `Editor.positions`.
> 
> How would we even do it? It would have to delegate to the browser somehow for that.

I just tried changing, for instance, the transforms/delete/unit-line tests all to use `block` instead of `line` and all the tests pass as the behavior is the same.  So, I'm pretty sure the `line` logic does not work any differently than the `block` logic.

My opinion is you **can't** implement `line` logic for Slate in core.  Slate core has no concept of "lines" and I don't think there is any viable way to do that.  I think `line` should probably be removed from the API as a whole and the tests that use `line` simply changed to `block`. 

@CameronAckermanSEL do you have any view on this discussion?  Thanks.

ianstormtaylor:
Hey @beorn I'm very sorry for the long delay on this. Your PR looks amazing, and the detailed explanations look great. I tried to rebase/resolve conflicts with master but I fear I may have not done it properly. If you can resolve the conflicts I'd be happy to merge this. Thank you!

beorn:
@ianstormtaylor Sure, I'll try to get the PR updated next week :)  Will be happy to see it included!

I also see that at least one of the related inline element cursor bugs in Chromium (the cursor rendering position) are fixed too.

beorn:
Hey, so this has actually been fixed in parallel some months ago - #4073 - and the fix was recently incorporated into the code base.

The remaining benefit of this `Editor.positions()` code is all the [refactoring and documentation that was done to it as described in steps (2) and (3) as described above](https://github.com/ianstormtaylor/slate/pull/3644#issuecomment-623120703).  It's a drop-in replacement of the "old" version of this function - it passes all the same tests.

There are two options:

1) If we want to keep these [refactorings](https://github.com/ianstormtaylor/slate/pull/3644#issuecomment-623120703), then this PR can be changed into one that just replaces Editor.positions().  I can do this is necessary - the simple way is to just replace the entire function, but I guess I rebase the branch as well.

2) If not, then we can just close this PR.

lmk

ianstormtaylor:
@beorn ah good to know. Well I'd be happy to merge in the refactored/commented code if you're down to fix it up. The changes you made looked good for future understandability to me.

PR #3666, opened 2020-05-08
[slate-react] Prefer domSelection.setBaseAndExtent

skogsmaskin:
This is faster than removing ranges and creating new ones (probably).
Also remove Firefox workaround, is it really needed? It was a workaround created in 2016.

This is improving performance.

#### What's the new behavior?
Instead of removing and creating new ranges, adjust the current one. This is faster. What do you think?

Fixes: #
Reviewers: @


ianstormtaylor:
Thanks @skogsmaskin! I think this has been incorporated into `master` already now, so I'm going to close this out.

Issue #3661, opened 2020-05-07
Broken Hovering Toolbar Example


adelsz:
Thanks for the amazing library!
I have noticed a bug on the `hovering-toolbar` example on `slatejs.com/examples`.
Editor crashes when pressing CMD-B with an active selection. We are running into the same bug on a project based off the `hovering-toolbar` configuration.

#### What's the current behavior?

1. Visit [https://www.slatejs.org/examples/hovering-toolbar](https://www.slatejs.org/examples/hovering-toolbar)
2. Select any word
3. Press CMD-B
4. Page crashes, printing:
```
Cannot resolve a DOM point from Slate point: {"path":[0,0],"offset":35}
```

Slate: 0.XX.X (version live on the demo website)
Browser: Chrome 81
OS: Mac OS

#### What's the expected behavior?

Page doesn't crash.

CameronAckermanSEL:
Thanks for the report @adelsz . That's a pretty weird one. I was initially going to respond to you that it was possibly crashing because this example doesn't have any keyboard shortcut support, but it looks like this example is unique in the crashing behavior.

I'll mark this as `help please` and maybe someone can take a closer look. I might be able to look at it later too.

PR #3667, opened 2020-05-09
fix hovering toolbar format hotkeys

whyour:

Fixing the hovering toolbar hotkeys error

#### What's the new behavior?

The default event has not been blocked before, so the code in the red box will be executed. I don't think this part of the code is necessary. Select should be handled by the plug-in itself
image

N/A

Fixes: #3661 
Reviewers: @


CameronAckermanSEL:
Thanks a ton for tracking this down @whyour ! Really appreciate the fix!

Issue #3668, opened 2020-05-09
onPaste not firing


knubie:
#### What's the current behavior?

```jsx
<Editable onPaste={() => console.log("paste")} />
```
The handler is never fired.

Slate: 0.58.0
Browser: Chrome / Safari / Firefox / Edge
OS: Mac / Windows / Linux / iOS / Android

#### What's the expected behavior?
The handler should fire.

Based on my testing it seems that the call to the handler is being short-circuited in this conditional:
https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/editable.tsx#L926-L932

In other words, if one of the other predicates in the conditional is false, the handler never gets called.

In an unrelated note I think that having a predicate cause side effects is bad practice, and would recommend refactoring the way these handler work in that regard.

mufasa71:
I am not sure but based on example given you need to wrap it with function

```
<Editable onPaste={() => console.log("paste")} />
```

knubie:
@mufasa71 Just a mistake when writing the issue.

PR #3670, opened 2020-05-10
Fix Editable onPaste handler not getting called

knubie:
#### Bug fix
Previously the `onPaste` handler was not being called, it was being short-circuited in the conditional.

#### What's the new behavior?
`onPaste` handler now gets called.

Check `isHandledEvent` first in the conditional so that it does not get short-circuited.

Fixes: #3668
Reviewers: @ianstormtaylor @CameronAckermanSEL


mattpardee:
Is there a way of speeding up this merge without being abrasive? It's pretty fundamental :-\

CameronAckermanSEL:
Thanks for fixing this problem @knubie !

PR #3672, opened 2020-05-10
Fix addMark hanging anchor bug

### LINKED ISSUE #3671: Editor.addMark adds mark to wrong node when anchor is on edge

Report a bug

When using `Transforms.move` to create a selection and then apply a mark, we have observed unexpected behavior when the `anchor` is at the end of node and the `focus` is in the next node. When we call `Editor.addMark`, it applies the mark to the previous node _and_ the next node.

For example, let's start with the following text:
Screenshot 2020-05-10 19 21 15

then we run the following on this editor:

```js
const _zero = { path: [0], offset: 0 }
Transforms.setSelection(editor, { anchor: _zero, focus: _zero })
Transforms.move(editor, { edge: 'focus', distance: 12 })
Transforms.move(editor, { edge: 'anchor', distance: 5 })
```

we would see this:
Screenshot 2020-05-10 19 20 16

now we run
```js
Editor.addMark(editor, 'italic')
```

and we end up with this:
Screenshot 2020-05-10 19 23 49

[This CodeSandbox](https://codesandbox.io/s/editoraddmark-bug-45vu2?file=/index.js:291-424) uses hyperscript to show how we would reproduce this bug in a test.

As the `output` hyperscript shows in the CodeSandbox, we expect the `italic` mark to only be applied to the text "one two", like so:

Screenshot 2020-05-10 20 11 27

pzhine:
Fixes #3671 

#### What's the new behavior?
When `Editor.addMark` is called on a selection with a "hanging" anchor (see #3671 for illustration), Slate will "unhang" the anchor before applying the mark. 

- Adds option flag `hangingAnchor` to `setNodes`
- Defaults it to `true` to preserve current behavior for operations that use `setNodes`
- Defaults it to `false` in `addMark` to fix unexpected mark behavior when anchor was hanging at the end of a node (see #3671 for example)


TheSpyder:
I think I like this, but is there any harm in making it always unhang both anchor and focus?

TheSpyder:
Turns out this is already fixed (see my comment on the original issue). This can probably be closed.

PR #3673, opened 2020-05-11
Bugfix/fix bundles

CameronAckermanSEL:
Bug

#### What's the new behavior?
The commonjs bundle was not being correctly transpiled as commonjs, instead it was a duplicate es6 module.

There was some (probably confusing) logic that was trying to handle an es6 and commonjs in the same build step in the rollup config factory. I simplified this by just creating an explicit commonjs step instead. Probably more refactoring is warranted at a later time.

Fixes: #
Reviewers: @

### DIFF
--- a/config/rollup/rollup.config.js
+++ b/config/rollup/rollup.config.js
@@ -22,6 +22,7 @@ function configure(pkg, env, target) {
   const isProd = env === 'production'
   const isUmd = target === 'umd'
   const isModule = target === 'module'
+  const isCommonJs = target === 'cjs'
   const input = `packages/${pkg.name}/src/index.ts`
   const deps = []
     .concat(pkg.dependencies ? Object.keys(pkg.dependencies) : [])
@@ -103,7 +104,7 @@ function configure(pkg, env, target) {
                 ],
                 modules: false,
                 targets: {
-                  esmodules: true,
+                  esmodules: isModule,
                 },
               },
         ],
@@ -116,7 +117,7 @@ function configure(pkg, env, target) {
             ? {}
             : {
                 regenerator: false,
-                useESModules: true,
+                useESModules: isModule,
               },
         ],
         '@babel/plugin-proposal-class-properties',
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "build": "yarn build:rollup && yarn build:next",
     "build:next": "cd ./site && next build && next export",
     "build:rollup": "rollup --config ./config/rollup/rollup.config.js",
-    "clean": "rm -rf ./node_modules ./packages/*/{dist,lib,node_modules} ./site/{.next,out}",
+    "clean": "rimraf ./node_modules ./packages/*/{dist,lib,node_modules} ./site/{.next,out}",
     "fix": "yarn fix:prettier && yarn fix:eslint",
```
CameronAckermanSEL, line 12: minor change, `rimraf` is a command built into npm/yarn that can do a cross paltform `rm -rf`

config/rollup/rollup.config.js
```diff
@@ -147,17 +148,12 @@ function configure(pkg, env, target) {
     }
   }
 
-  if (isModule) {
+  if (isCommonJs) {
     return {
       plugins,
       input,
       onwarn,
       output: [
```
CameronAckermanSEL, line 156: instead of this one step doing es and cjs output, it now just does common js output

config/rollup/rollup.config.js
```diff
@@ -116,7 +117,7 @@ function configure(pkg, env, target) {
             ? {}
             : {
                 regenerator: false,
-                useESModules: true,
+                useESModules: isModule,
```
CameronAckermanSEL, line 120: these flags are conditional to whether this is an es6 build

config/rollup/rollup.config.js
```diff
@@ -173,6 +169,27 @@ function configure(pkg, env, target) {
       },
     }
   }
+
```
CameronAckermanSEL, line 172: this is identical to what it was before but now just handles es6 only, no cjs output here

package.json
TheSpyder: It's not a built-in, it's just a package that is so common it turns up everywhere. I recommend adding it as an explicit dependency if you're going to use it.
https://www.npmjs.com/package/rimraf

`yarn why rimraf` shows it's currently hoisted from numerous transitive dependencies :)

CameronAckermanSEL: Thanks for catching this mistake! That could have been a little unpleasant 😆

PR #3675, opened 2020-05-11
Fix path transform against sibling move operations

jason-krypton:
Fixing a bug I discovered while working on Slate collab OT. Added unit tests to around the fix. 

Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate/src/interfaces/path.ts
+++ b/packages/slate/src/interfaces/path.ts
@@ -391,6 +391,12 @@ export const Path = {
             }
 
             return copy.concat(p.slice(op.length))
+          } else if (Path.isSibling(op, p) && Path.equals(onp, p)) {
+            if (Path.endsBefore(op, p)) {
+              p[op.length - 1] -= 1
+            } else {
+              p[op.length - 1] += 1
+            }
           } else if (
             Path.endsBefore(onp, p) ||
             Path.equals(onp, p) ||
--- a/packages/slate/test/interfaces/Path/transform/move_node/ends-before-to-ends-equal.js
+++ b/packages/slate/test/interfaces/Path/transform/move_node/ends-before-to-ends-equal.js
@@ -1,0 +1,15 @@
+import { Path } from 'slate'
+
+const path = [0, 1]
+
+const op = {
+  type: 'move_node',
+  path: [0, 0],
+  newPath: [0, 1],
+}
+
+export const test = () => {
+  return Path.transform(path, op)
+}
+
+export const output = [0, 0]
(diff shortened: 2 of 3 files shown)


packages/slate/test/interfaces/Path/transform/move_node/ends-before-to-ends-equal.js
```diff
@@ -0,0 +1,15 @@
+import { Path } from 'slate'
+
+const path = [0, 3]
+
+const op = {
+  type: 'move_node',
+  path: [0, 0],
+  newPath: [0, 1],
```
TheSpyder, line 8: This doesn't look like "ends equal"? Is the test name wrong?

jason-krypton: Sorry, I messed up porting code over to this new fork. I just fixed the test. Thanks for pointing that out as this is the most important test.

packages/slate/src/interfaces/path.ts
```diff
@@ -391,11 +391,22 @@ export const Path = {
...
-            Path.equals(onp, p) ||
-            Path.isAncestor(onp, p)
-          ) {
+          } else if (Path.endsBefore(onp, p) || Path.isAncestor(onp, p)) {
+            if (Path.endsBefore(op, p)) {
+              p[op.length - 1] -= 1
+            }
+
+            p[onp.length - 1] += 1
+          } else if (Path.equals(onp, p)) {
```
TheSpyder, line 400: rather than splitting out `Path.equals` with duplicate code if `op` isn't a sibling, and using an early `break`, it might be clearer to leave the code as it was and add this as a condition above?

```
// patch code
          } else if (Path.isSibling(op, p) && Path.equals(onp, p)) {
            if (Path.endsBefore(op, p)) {
              p[op.length - 1] -= 1
            } else {
              p[op.length - 1] += 1
            }
// original code
          } else if (
            Path.endsBefore(onp, p) ||
            Path.equals(onp, p) ||
            Path.isAncestor(onp, p)
          ) {
...
```

Is that the same logic? It's a bit hard to tell from the diff.

jason-krypton:
Sorry I've been busy on other work. Is the slight change to the way the code is structured the only request before being able to check this in? @CameronAckermanSEL @TheSpyder

TheSpyder:
I was speculating in Slack that since the core depends on the path transformation to apply `move_node` there's more risk to these changes than might first appear. I'm not saying you need to do this, but in theory it should be replicable in an editor test 🤔

jason-krypton:
@TheSpyder I agree there is more risk, but it is broken or do you disagree on that?  Seems best to fix clearly broken logic.

TheSpyder:
Without Ian to give advice I guess we'll just have to trust that there are enough tests for the core move node implementation to cover this.

I think the logic is sound, but I found it easier to reason about the changes with the code suggestion I made than the duplication+`break` as you used.

jason-krypton:
Updated @TheSpyder

CameronAckermanSEL:
Thanks for the contribution here @jason-krypton !

Issue #3463, opened 2020-01-23
Jumpy editor behavior in scrollable container


stapleup:
#### Do you want to request a _feature_ or report a _bug_?
_bug_

#### What's the current behavior?
Editor jumps up and down when it is placed inside a scrollable container. This issue is different from other issues previously reported because they all target older versions of Slate.
GIF: https://share.getcloudapp.com/7KuyN5zo

Slate: 0.57.1
Browser: Chrome
OS: Windows

#### What's the expected behavior?

Slate should scroll inside its container and not scroll the parent container.

#### Possible Reason
Slate 0.50+ seems to be using https://github.com/stipsan/scroll-into-view-if-needed in order to bring selected view into focus. Scroll-into-view have a property  `boundary`  which allows us to define a parent element inside which it should position the item.. This parameter seems to be missing here https://github.com/ianstormtaylor/slate/blob/e54b07eba8ab23a8c26b275f77e8ec207f8c9b55/packages/slate-react/src/components/editable.tsx#L197 and it causes the parent container to scroll along with editor. This issue can be reproduced on the official site for said library here (https://scroll-into-view-if-needed.netlify.com/) if you disable the boundry check box next to **Limit propagation** you'll see it scrolls the whole document when positioning the element.

samosaboy:
I too am experiencing this issue. It is quite apparent when selecting text as well.

Francesco-Lanciana:
I'm also experiencing this when I format my text, even if the text is completely within the viewport which is wierd since that shouldn't happen with scrollMode: 'if-needed'.

I can't seem to figure why we actually have this scrollIntoView though. My best guess is that it's simply a convenience for users of Slate, so if they programatically update their selection it will also scroll to this new selection? 

For my specific use case I just did a bit of digging and it turns out that there is a quirk in browsers where if you set the HTML element height to be 100% then getClientBoundingRect gives a negative value for bottom (which makes no sense). The scroll-into-view-if-needed uses a library called compute-scroll-into-view to check if the element in question is within the viewport (which it does by getting the HTML elements rect and the element in questions rect and doing some checks). If it is and scrollMode is 'if-needed' then it won't scroll. Unfortunately the negative value causes this check to fail and it scrolls anyway. I have raised an issue about this https://github.com/stipsan/compute-scroll-into-view/issues/652

[...]

Issue #3677, opened 2020-05-12
Weird jumping with long texts


svenadlung:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

With long texts scroll position randomly jumps while typing. Please see sandbox example: https://jsfiddle.net/h94wn1Lf/

![May-12-2020 16-15-27](https://user-images.githubusercontent.com/11420688/81702699-e9043500-946b-11ea-9d68-2559023bb477.gif)

1. Place cursor in the beginning of the text and start typing

Maybe this is related to https://github.com/ianstormtaylor/slate/issues/3463?

Slate: 0.57.1
Browser: Chrome / Firefox
OS: Mac

#### What's the expected behavior?

The view should not leave the cursor while writing.

CameronAckermanSEL:
Hey @svenadlung  thanks for filing this. I think you're right that it's a potential duplicate of #3463, but I'm not 100% sure, so for the time being I'll leave it open. Would you be able to try a small change 
https://github.com/ianstormtaylor/slate/commit/1652ecab52088e64ffbaa130fa4049f60cdce57e
and see if this fixes the issue for you? That'll help us be sure.

Thanks!

svenadlung:
Hi @CameronAckermanSEL thanks for your quick reply. 

I patched our code and jumping around stopped. Within scrollable container it seems to work well. Unfortunately it doesn't scroll anymore at all without a scrollable parent. But this could be related to our source as well. Maybe you can keep an eye on that while reviewing https://github.com/ianstormtaylor/slate/commit/1652ecab52088e64ffbaa130fa4049f60cdce57e

Another issue besides wrong calculations in scrollable containers that we encountered concerns long texts. 
`scroll-into-view` by default uses `block: "center"` which causes scrolling to the vertical center of texts which might lead to a mispositioning in that case.

Still I am not sure how everything works with overlaying element like a fixed header. There is no easy way to change current calculation and add an offset f. ex.

Let me know if there is anything else I can do.

Best whishes.

CameronAckermanSEL:
Thanks for taking the time to try this and for the feedback. That's really helpful. I'll be sure to look in that area when reviewing the PR

CameronAckermanSEL:
I merged a PR that appears to fix this problem. When I tried to reproduce the issue you mentioned, I didn't have much luck running into it. If you still have that problem and have a reproduction of it handy, please let me know and we can make sure it gets addressed.

Thanks!

[...]

PR #3678, opened 2020-05-13
Fix scrolling boundary detection (see #3463)

mitchellbusby:

Fixing a bug

#### What's the new behavior?

Supplies the selected element specifically as a ["boundary"](https://github.com/stipsan/compute-scroll-into-view#boundary) to the scroll-into-view-if-needed library to determine whether or not Slate's selection is in the viewport.

This seems to solve well for cases where we may have multiple editors in a page, or where the bounding box of the page is not the same as the bounding box of the Slate editor.

I think longer term this behaviour _should_ be configurable in user space but this is a tactical solution to fix a broad use case and a blocker to adoption of new Slate for me and others.

See above.

Fixes: #3463 
Reviewers: @CameronAckermanSEL

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -175,7 +175,10 @@ export const Editable = (props: EditableProps) => {
     if (newDomRange) {
       domSelection.addRange(newDomRange!)
       const leafEl = newDomRange.startContainer.parentElement!
-      scrollIntoView(leafEl, { scrollMode: 'if-needed' })
+      scrollIntoView(leafEl, {
+        scrollMode: 'if-needed',
+        boundary: el,
+      })
     }
 
     setTimeout(() => {


CameronAckermanSEL:
Thanks a ton for the fix @mitchellbusby ! Seems to work well. I also tried reproducing the case mentioned here and https://github.com/ianstormtaylor/slate/issues/3677#issuecomment-628578878 wasn't able to observe the problem this user was mentioning. But wanted to record it for posterity nonetheless in case someone else runs into it.

ulion:
For certain conditions, according to https://github.com/stipsan/compute-scroll-into-view/blob/master/src/index.ts the boundary will not be treat as a scrollable parent frame due to the parameter we passed in this PR.
which cause it does not scroll at all in some browsers as pointed in https://github.com/ianstormtaylor/slate/issues/3750#issuecomment-652590942

the boundary parameter indeed means that the compute function will not consider it and its parent as scrollable frames and try to do any scroll on them. but sometimes, the boundary we passed in here is just the only one scrollable frame we have.
I believe this PR is not good, due to the boundary is not consider scrollable.

PR #3679, opened 2020-05-13
Fix a few issues with the documentation

ksimons:
When migrating to the latest and greatest slate, I noticed a few issues in the documentation.

Updating documentation and adding Kitemaker to the list of sites using Slate

#### What's the new behavior?

None

N/A

No


CameronAckermanSEL:
Thanks very much @ksimons !

PR #3688, opened 2020-05-17
Fix: ESlint Error (eslint: no-shadow)

mrfambo:
### Docs Fix

PROBLEM: "Value is defined in the upper scope (eslint: no-shadow)"
If you are trying to follow the guide using eslint there is "no-shadow" error because "value" is defined as a state variable and being used again in onChange.

Screenshot 2020-05-17 at 6 12 11 PM

FIX:
replace value with "newValue" on onChange.

### DIFF
--- a/docs/walkthroughs/01-installing-slate.md
+++ b/docs/walkthroughs/01-installing-slate.md
@@ -70,7 +70,7 @@ const App = () => {
   const [value, setValue] = useState([])
   // Render the Slate context.
   return (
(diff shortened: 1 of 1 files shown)


mrfambo:
P.S: Its my first contribution :P I don't know if i messed up something :D

CameronAckermanSEL:
Cool! Thanks for the adjustment @mrfambo !

PR #3690, opened 2020-05-18
Only normalize once per call to normalizeNode

### LINKED ISSUE #3689: Normalization incorrectly removes node

Bug

Slate release: 0.58.1

How to reproduce:
This bug was exposed by a document state that is reproduced in PR #3602. You can see the behavior in the GIF below, which is from the PR comment by @rskopal

82084106-16aae180-96eb-11ea-94d2-24d0f26b0016

I was able to reproduce the bug with the following test code:

```js
const input = (
  <editor>
    <block>
      <text a>1</text>
      <text b>
        <cursor />
      </text>
      <text c>2</text>
    </block>
  </editor>
)
const output = (
  <editor>
    <block>
      <text a>1</text>
      <text b>
        <cursor />
      </text>
      <text c>2</text>
    </block>
  </editor>
)
Editor.normalize(input, { force: true })
assert.deepEqual(input.children, output.children)
```

Which yields the following result:

Screenshot 2020-05-17 15 41 26

We should be able to test this behavior in the Netlify deployment for the PR, but that is currently failing for reasons unrelated to the PR (as far as I know).

As the test output shows, the expected behavior is that the node with text "1" is not removed.

pzhine:

Fixes #3689

#### What's the new behavior?

`editor.normalizeNode` loops over the children of the node and looks for normalization criteria. Currently, the loop continues even after normalization has happened. It updates the array index after the normalization (e.g. decrements if a child was merged or removed), but the array still points to the original, unmutated node children. This causes problems when the next normalization occurs because it is applied to the wrong node.

One option is to re-sync the array at the end of each iteration of the loop. However, this caused several tests to fail. Instead, we bail out of the loop after the first normalization. This solution passes all the tests, and it should be sufficient because `normalizeNode` is itself [called in the loop that iterates over "dirty" paths](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/interfaces/editor.ts#L762). In other words, if there are multiple normalizations needed on a node, the node should have multiple entries in `DIRTY_PATHS`.


TheSpyder:
Your commit history in this PR is quite weird. Looks like a merge gone wrong. Can I suggest a `git rebase -i upstream/master` and delete everything except the commit you meant to include in the PR?

TheSpyder:
As for the fix itself, there don't appear to be any tests for nodes that require multiple normalisations when the document is first loaded? The `DIRTY_PATHS` logic you mentioned would only apply to changes during editing, as far as I understand.

packages/slate/src/create-editor.ts
```diff
@@ -236,45 +232,45 @@ export const createEditor = (): Editor => {
         // other inline nodes, or parent blocks that only contain inlines and
         // text.
         if (isInlineOrText !== shouldHaveInlines) {
-          Transforms.removeNodes(editor, { at: path.concat(n), voids: true })
-          n--
+          Transforms.removeNodes(editor, { at: path.concat(i), voids: true })
+          break
```
BrentFarese, line 236: Why all the `break` statements?  The point of the loop is to continue and to use `n` to keep track of removed or added nodes so it points to the right index/node.  Was there a bug you experienced with this code as it is?

dylans (changes requested):
This is going to need more explanation to address the previously provided feedback (now over a a year old), and it's stale.

If you have interest in revisiting this please rebase and update. Otherwise I'll close in a week or so.

dylans:

PR #3691, opened 2020-05-18
Fix inverse of sibling move operation

jason-krypton:
Another breaking fix, but should be safer as it seems like only slate-history uses inverse in the main packages. 

PTAL @CameronAckermanSEL @ianstormtaylor

### DIFF
--- a/packages/slate/src/interfaces/operation.ts
+++ b/packages/slate/src/interfaces/operation.ts
@@ -235,9 +235,9 @@ export const Operation = {
         }
 
         // If the move does not happen within a single parent it is possible
-        // that the move impact the true path to the location of where the node
-        // was removed and where it was inserted. We have to adjust for this
-        // and the original path. We can accomplish this (only in non-sibling)
+        // for the move to impact the true path to the location where the node
+        // was removed from and where it was inserted. We have to adjust for this
+        // and find the original path. We can accomplish this (only in non-sibling)
         // moves by looking at the impact of the move operation on the node
         // after the original move path.
         const inversePath = Path.transform(path, op)!
--- a/packages/slate/src/interfaces/operation.ts
+++ b/packages/slate/src/interfaces/operation.ts
@@ -228,9 +228,18 @@ export const Operation = {
           return op
         }
 
-        // We need to get the original path here, but sometimes the `newPath`
-        // is a younger sibling of (or ends before) the original, and this
-        // accounts for it.
+        // If the move happens completely within a single parent the path and
+        // newPath are stable with respect to each other.
+        if (Path.isSibling(path, newPath)) {
+          return { ...op, path: newPath, newPath: path }
+        }
+
+        // If the move does not happen within a single parent it is possible
+        // that the move impact the true path to the location of where the node
+        // was removed and where it was inserted. We have to adjust for this
+        // and the original path. We can accomplish this (only in non-sibling)
+        // moves by looking at the impact of the move operation on the node
+        // after the original move path.
         const inversePath = Path.transform(path, op)!
         const inverseNewPath = Path.transform(Path.next(path), op)!
         return { ...op, path: inversePath, newPath: inverseNewPath }
--- a/packages/slate/test/interfaces/Operation/inverse/moveNode/child-to-parent.js
+++ b/packages/slate/test/interfaces/Operation/inverse/moveNode/child-to-parent.js
@@ -1,5 +1,8 @@
 import { Operation } from 'slate'
 
+// This test covers moving a child to the location of where the current parent is (not becoming its parent).
+// When the move happens the child is inserted infront of its old parent causing its former parent's index to shiftp
+// back within its former grandparent (now parent).
 export const input = { type: 'move_node', path: [0, 2, 1], newPath: [0, 2] }
 
 export const test = value => {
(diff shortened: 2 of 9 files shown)


TheSpyder:
how does this relate to #3675? 🤔

`previous-sibling-of-parent` is a bit mind bending :joy:

packages/slate/src/interfaces/operation.ts
```diff
@@ -228,6 +228,10 @@ export const Operation = {
           return op
         }
 
+        if (Path.isSibling(path, newPath)) {
```
TheSpyder, line 231: Can you please add a comment here about the reasoning? The comment below also talks about siblings, so it might be confusing to follow later.

jason-krypton:
@TheSpyder Open to suggestions on the test names. Need to describe the source and destination locations.

The PRs are unrelated. Just two bugs of a similar nature. It seems like the complicated cases were considered carefully but not the simple ones.

jason-krypton:
@TheSpyder Updated comment to hopefully add some clarity. Also renamed the tests. I did add some more tests as I realized I should flush out some use cases.

packages/slate/src/interfaces/operation.ts
```diff
@@ -228,9 +228,18 @@ export const Operation = {
           return op
         }
 
-        // We need to get the original path here, but sometimes the `newPath`
-        // is a younger sibling of (or ends before) the original, and this
-        // accounts for it.
+        // If the move happens completely within a single parent the path and
```
TheSpyder, line 231: This helped more than I expected, it wasn't immediately obvious that `Path.isSibling` was within a single parent until I took a close look at the code 😂

packages/slate/test/interfaces/Operation/inverse/moveNode/child-to-parent.js
```diff
@@ -0,0 +1,9 @@
+import { Operation } from 'slate'
+
+export const input = { type: 'move_node', path: [0, 2, 1], newPath: [0, 2] }
```
TheSpyder, line 3: this is "child to _replace_ parent"? 🤔

TheSpyder:
That's much better! I like all the new tests.

packages/slate/test/interfaces/Operation/inverse/moveNode/child-to-parent.js
```diff
@@ -0,0 +1,9 @@
...
+
+export const test = value => {
+  return Operation.inverse(value)
+}
+
+export const output = { type: 'move_node', path: [0, 2], newPath: [0, 3, 1] }
```
CameronAckermanSEL, line 9: should the path mutate here? `newPath: [0, 3, 1] `

packages/slate/test/interfaces/Operation/inverse/moveNode/ends-before-parent-to-child.js
```diff
@@ -0,0 +1,9 @@
+import { Operation } from 'slate'
+
+export const input = { type: 'move_node', path: [0, 1], newPath: [0, 2, 1] }
+
+export const test = value => {
+  return Operation.inverse(value)
+}
+
+export const output = { type: 'move_node', path: [0, 1, 1], newPath: [0, 1] }
```
CameronAckermanSEL, line 9: should the path mutate here?

packages/slate/test/interfaces/Operation/inverse/moveNode/child-to-parent.js
jason-krypton: Might be debating semantics in the english language, but it isn't replacing its parent, it is moving to the location fo where its parent currently is, which in turn shifts the location of its current parent before the move.

packages/slate/test/interfaces/Operation/inverse/moveNode/child-to-parent.js
jason-krypton: Yup, when the node is moved to `[0, 2]` it shifts what was its parent `[0, 2]` back inside its grandparent to `[0,3]`. I find it helpful to grab a whiteboard or piece of paper to draw these out when working on them.

packages/slate/test/interfaces/Operation/inverse/moveNode/ends-before-parent-to-child.js
jason-krypton: Yes, when the node moving is removed from its starting location `[0, 1]` the node `[0, 2]` slides backwards to fill the gap.

packages/slate/test/interfaces/Operation/inverse/moveNode/child-to-parent.js
CameronAckermanSEL: Could we add this comment to the test cases? I know that's not a really pervasive convention at this time but I think it would help for posterity.

CameronAckermanSEL:
Thanks very much for these fixes and additional tests!

packages/slate/test/interfaces/Operation/inverse/moveNode/child-to-parent.js
TheSpyder: What you're describing is what I meant by replacing, but you're correct that was the wrong way to put it. The comment you added does help.

PR #3692, opened 2020-05-19
docs(resources): slate-plugins-next

zbeyens:
Adding slate-plugins-next in the resources


CameronAckermanSEL:
Thanks @zbeyens ! These plugins are really awesome and a huge boon to the community :) Thanks for all of your contributions

PR #3694, opened 2020-05-20
Remove Outline since they switched editors

sparkertime:
See https://github.com/outline/rich-markdown-editor/pull/150 

Neither - fixing the docs.


CameronAckermanSEL:
Thanks @sparkertime !

PR #3697, opened 2020-05-22
Add OperationType Enum

jason-krypton:
#### What's the new behavior?

No behavior change

N/A

- [?] The new code matches the existing patterns and styles.


CameronAckermanSEL:
I like this a lot personally. We're a bit hot off of a breaking change to the types so while this is high value to me personally we might want to wait to merge this. Also, interested in Ian's opinions of the enums; haven't really gleaned much about his view on types.

jason-krypton:
Rebased to update new tests I added in another PR. Have things settled down sufficiently to check this in now? @CameronAckermanSEL @BrentFarese

packages/slate/src/interfaces/operation.ts
```diff
@@ -1,23 +1,35 @@
...
+  InsertNode = 'insert_node',
+  InsertText = 'insert_text',
+  MergeNode = 'merge_node',
+  MoveNode = 'move_node',
+  RemoveNode = 'remove_node',
+  RemoveText = 'remove_text',
+  SetNode = 'set_node',
+  SetSelection = 'set_selection',
+  SplitNode = 'split_node',
+}
```
AndrewScull, line 14: If is possible to use a [const enum](https://www.typescriptlang.org/docs/handbook/enums.html) here so there's no extra generated code, as was the case with the string literal types that were previously used?

ianstormtaylor:
Just curious… are there use cases where using the enums gives better type safety? Because my initial reaction is against simply because it introduces an extra concept/import/object to keep track of when strings are actually more clear as to what's going on underneath. However if this actually makes typings safer that changes the tradeoffs calculation.

BrentFarese:
> Just curious… are there use cases where using the enums gives better type safety? Because my initial reaction is against simply because it introduces an extra concept/import/object to keep track of when strings are actually more clear as to what's going on underneath. However if this actually makes typings safer that changes the tradeoffs calculation.

We try and use them a lot in our codebase vs. string literals for a couple of reasons.

1. Now you have an enum to use across the codebase vs. a string literal so if you ever change the enum value later it's easy to change it everywhere in the code vs. searching for string literals.

2. You can use things like `Object.values(enum)` to get an array of all possible enum values.  Comes in handy.

3.  Other reasons that are escaping me at present (haha).

As far as I know you don't get any better type-safety out of enum vs. string literal but we have found enums to be far better to work with from a DX perspective.  Let me know if you have a diff experience.  Thanks!

BrentFarese:
Yes that works for me!  There is not much concrete benefit immediately.

PR #3700, opened 2020-05-26
Use last available DOM node even if point offset is too large

### LINKED ISSUE #3699: Incorrect selection auto update after inserting manual decorator
Hello. I am receiving famous error `Cannot resolve a DOM point from Slate point` after inserting manual decorator. This code was recently updated by the following pull request #3594. But this code is still broken.

1. Please apply the following [patch](https://github.com/ianstormtaylor/slate/files/4684538/mentions2.zip) for `mentions` example. This code will try to replace last child with `@Amee`.
2. Open `mentions` example.
3. Set cursor to the last possible position.
4. Press `insert` key.
5. You will receive `Cannot resolve a DOM point from Slate point`.

In this example cursor appears **after** last child and `toDOMPoint` can't find point. This is fine and should be handled properly: selection end should be limited to the last available point from `DOM`.

andrew-aladev:
Fix is very simple - please use last available DOM node instead of throwing error, because it is possible that point offset will be too large during custom value manipulations.

Fixed #3699.


dylans (commented):
Looking at the history of this one, I have mixed feelings:

Pros: out of range errors are a common problem for people new to slate
Cons: silently just treating the last valid point of the end could cause more issues

If we were to proceed with this change, I would probably advocate for a console warning when the offset gets changed so that someone doesn't get silently unexpected behavior.

I don't have a sufficiently strong opinion so I'll talk with the core team and try to get feedback on how to proceed.

TheSpyder:
Silently correcting failures can too easily lead to compounding errors and unwanted side-effects. I also agree with the original reason #3699 was closed; Slate has specific methods to update the content and does not support direct mutation of the model.

PR #3704, opened 2020-05-28
Added CSS Class to Placeholder for Optional Styling

anatolzak:
improving a _feature_ 

#### What's the new behavior?
You can now apply custom styles to the placeholder thanks to having a CSS class tied to the placeholder.

DraftJS has a CSS class to the placeholder which not only allows you to style it according to your needs (without any plugins 😀 ) but also enables the user to change the following behavior most users would probably agree with is very undesirable.

When having no text in the editor, but toggling a block style. The placeholder matches the block style too. See the screenshots below:

image

image

So now thanks to having a placeholder with a CSS class, you can check if there isn't any text in the editor, and the block type isn't the default one (meaning i.e.: header, unordered list, etc.), and then add a dynamic className to the Editable component.

And then with CSS assuming the dynamic className provided to the Editable component is `Editable-HidePlaceholder` and the class name of the placeholder is `slate-placeholder`:

`.Editable-HidePlaceholder .slate-placeholder {
    display: none;
}`

Now, you successfully hid the placeholder when there is no text and when a custom block type is toggled! Congrats! 🎉

image

image

Here is an example of how you might achieve that behaviour:

```javascript
// the usual stuff
const editor = useMemo(() => withReact(createEditor()), []);
const [value, setValue] = useState([{ type: 'paragraph', children: [{ text: '' }] }]);

let EditorClassName;
// serialize the value to plain text
const editorText = value.map(n => Node.string(n)).join('')

// add placeholder class if there is no text and if the block type is not a paragraph
if (!editorText && editor.children[0]) {
    if (editor.children[0].type !== 'paragraph') {
        EditorClassName = "Editable-HidePlaceholder";
    }
}

return (
    <Slate 
        // the usual stuff
    >
        <Editable
            // the usual stuff
            className={EditorClassName} />
    </Slate>
```

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/leaf.tsx
+++ b/packages/slate-react/src/components/leaf.tsx
@@ -32,6 +32,7 @@ const Leaf = (props: {
     children = (
       <React.Fragment>
         <span
+          className="slate-placeholder"
           contentEditable={false}
           style={{
             pointerEvents: 'none',
--- a/packages/slate-react/src/components/leaf.tsx
+++ b/packages/slate-react/src/components/leaf.tsx
@@ -42,6 +42,9 @@ const Leaf = (props: {
             whiteSpace: 'nowrap',
             opacity: '0.333',
             userSelect: 'none',
+            fontStyle: 'normal',
+            fontWeight: 'normal',
+            textDecoration: 'none',
           }}
         >
           {leaf.placeholder as React.ReactNode}
--- a/packages/slate-react/src/components/leaf.tsx
+++ b/packages/slate-react/src/components/leaf.tsx
@@ -41,6 +41,7 @@ const Leaf = (props: {
             maxWidth: '100%',
             whiteSpace: 'nowrap',
             opacity: '0.333',
+            userSelect: none,
           }}
         >
           {leaf.placeholder as React.ReactNode}
--- a/packages/slate-react/src/components/leaf.tsx
+++ b/packages/slate-react/src/components/leaf.tsx
@@ -41,7 +41,7 @@ const Leaf = (props: {
             maxWidth: '100%',
             whiteSpace: 'nowrap',
             opacity: '0.333',
-            userSelect: none,
+            userSelect: 'none',
           }}
         >
           {leaf.placeholder as React.ReactNode}


CameronAckermanSEL:
This PR is the same as this other one: https://github.com/ianstormtaylor/slate/pull/3461 

And I still have the same concerns with it. I understand there's a common need here but I'm wondering whether a hard coded class name is the best API for this?

anatolzak:
I was trying to solve a problem, which I had with the placeholder still showing when you toggle headers, bullet lists, etc. which is not the behavior I want it.

So to solve that problem, I came with an alternative solution that doesn't involve any CSS class names:

```javascript
// the usual stuff
const editor = useMemo(() => withReact(createEditor()), []);
const [value, setValue] = useState([{ type: 'paragraph', children: [{ text: '' }] }]);

let placeholder = "Write something...";
// serialize the value to plain text
const editorText = value.map(n => Node.string(n)).join('')

// remove placeholder if there is no text and if the block type is not a paragraph
if (!editorText && editor.children[0]) {
    if (editor.children[0].type !== 'paragraph') {
        placeholder = null;
    }
}

return (
    <Slate 
        // the usual stuff
    >
        <Editable
            // the usual stuff
            placeholder={placeholder} />
    </Slate>
```

This is the only reason I thought the placeholder should have a CSS class. I like the placeholder styles the library provides and by coming up with an alternative solution for my placeholder problem, it makes the placeholder behave exactly the way I want it to. 

Thanks for appreciating my placeholder improvements 🙂

anatolzak:
But if the user DOES want to style the placeholder according to their needs, I would say that because slate is all about making you able to create a rich text editor according to your domain-specific needs and therefore, the library doesn't want to hard-code things like CSS classes because you want to give the user the full flexibility to create an editor according to their needs, probably the best solution would be to let the user set a custom placeholder CSS class prop in the Editable component and by doing so, you allow them to create exactly what they want to achieve with slate that they couldn't with other libraries, which is why slate was created.

That's just my 2 cents here 🙂

skogsmaskin:
In my fork I set the placeholder span with a data-attribute `data-slate-placeholder` and this can be used to style it from the outside. It's a data-attribute because that's what `slate-react` does for everything else (no className).

CameronAckermanSEL:
An attribute seems a more inline with how these slate provided components are treated

ianstormtaylor:
Hey @anatolzak thanks for working on this. I agree with @CameronAckermanSEL that it feels like we should provide a more customizable version of this idea. I wrote up a solution in https://github.com/ianstormtaylor/slate/issues/4123 that I think would work, if any one wants to contribute! Thanks.

PR #3712, opened 2020-06-01
Fix path transform for sibling ancestor moves

jason-krypton:

My previous fix https://github.com/ianstormtaylor/slate/pull/3675 did not go far enough. Extending that change to also fix sibling moves around ancestors.

Reviewers: @CameronAckermanSEL


packages/slate/src/interfaces/path.ts
```diff
@@ -386,12 +386,14 @@ export const Path = {
             const copy = onp.slice()
 
             if (Path.endsBefore(op, onp) && op.length < onp.length) {
-              const i = Math.min(onp.length, op.length) - 1
-              copy[i] -= 1
+              copy[op.length - 1] -= 1
```
jason-krypton, line 389: Note: This is just clean up. The length check in the if already guaranteed which path is less.

jason-krypton:
@TheSpyder The sample you provided is ancestor (parent) to ends after. Did you just want to see parent to moves after instead of grandparent?

TheSpyder:
@jason-krypton yes. I was looking at the `ancestor-to-*` tests and wondered how it would react if the direct parent was moved instead of the root ancestor. It'll probably be fine but an extra test can't hurt :)

jason-krypton:
@TheSpyder tests added

TheSpyder:
Thanks!

CameronAckermanSEL:
Thanks again for the PR @jason-krypton !

Issue #3693, opened 2020-05-19
Decorate is duplicating leaves


delijah:
#### What's the current behavior?

If you insert more than one soft hyphen after each other, they get duplicated. renderLeaf is called twice for every soft hyphen (except the first one).

![soft-hyphen](https://user-images.githubusercontent.com/237739/82337915-94713480-99ec-11ea-812e-ff7a61fd49df.gif)

Sandbox: https://codesandbox.io/s/slate-reproductions-v99e1
OS: macOS Mojave 10.14.6
Browser: Google Chrome 81.0.4044.138 (Official Build) (64-bit)
Slate version: 0.58.0

#### What's the expected behavior?

For every inserted soft hyphen, renderLeaf should only be called once.

It looks like the problem occurs over here:
https://github.com/ianstormtaylor/slate/blob/d82ffe49a5253de08adab8f36ac7f07879037977/packages/slate/src/interfaces/text.ts#L91

But what is happening there is too complex for me to decipher.

TheSpyder:
I can reproduce this in the search highlighting example:
https://www.slatejs.org/examples/search-highlighting

There is no visible issue, because search highlight adds a class without `::before` content, but a DOM inspector reveals 3 spans with the search highlight class (`css-evmfc5` in this screenshot).

![Screen Shot 2020-05-25 at 10 09 00 am](https://user-images.githubusercontent.com/298292/82768180-c2cd9480-9e70-11ea-8fe2-3275b79fde9a.png)

To replicate:
* insert the same text twice with no spaces (I used `eded`)
* search for that text (in my case `ed`)

It looks like something is adding a zero-width space in between the two matched decorate ranges, which is probably fine, but the zero-width space is given the same leaf properties as the matched ranges either side of it. I'm not sure if that is by design or not but it's certainly the cause of this issue.

PR #3714, opened 2020-06-02
Fixed decorations for single characters followed by each other

delijah:
#### What's the new behavior?

Looks like one if-statement was just done wrong before. Single character decorations followed by each other were not taken in consideration. Changing that if-statement (see changed files of this PR) fixes this edge case scenario. Existing tests and examples still work.

This code part is so complicated, it's hard for me to describe what changed. Please have a look at the code and you will see.... Since i am not 100% sure how this function works, it would be great if someone that is more aware of the internals of this function, could have a look at this PR.

Fixes: #3693


packages/slate/test/interfaces/Text/decorations/in-a-row.js
```diff
@@ -0,0 +1,51 @@
+import { Text } from 'slate'
+
+export const input = [
+  {
```
CameronAckermanSEL, line 4: This is a great test, but I'm wondering whether the name is correct. Are these two adjacent decorations or are they overlapping? 

```
0  1  2  3  4
a [b (c] d) e
```
Where decoration one is [] and decoration two is ().

Am I picturing this correctly, or does one of the ends of the decoration not actually contain the character at the index?

CameronAckermanSEL: Oh and a follow up, if this test is indeed "overlapping" decorations, then should we have one for adjacent decorations also?

delijah: Yeah sorry my english is not that good. This test is about adjacent decorations. I will change it.

delijah: Done :)

packages/slate/src/interfaces/text.ts
```diff
@@ -111,7 +111,7 @@ export const Text = {
 
         // If the range starts after the leaf, or ends before it, continue.
         if (
-          start.offset > offset + length ||
+          start.offset >= offset + length ||
```
cudr, line 114: @delijah @CameronAckermanSEL, it looks like collapsed decorators will never be rendered at the end of the line and intersected after this change.

CameronAckermanSEL: Thanks, this is a good catch

PR #3716, opened 2020-06-03
Allow void nodes to be copied. Fixes #3665

### LINKED ISSUE #3665: Copy/pasting a void element is not working
bug

Copy/pasting a void element is not working. This can be reproduce here: https://www.slatejs.org/examples/images and here: https://www.slatejs.org/examples/mentions

slate-bug-copy-paste

Slate: 0.58.0
Browser: Chrome
OS: Mac

Copy/paste should work

majelbstoat:

fixing a _bug_

#### What's the new behavior?

#3620 broke copying of void nodes with a slightly different default behaviour for getting the fragment to copy. Prior to that PR, it was not checking for an expanded range. I mistakenly introduced a stricter check, which prevented void nodes from being copied when they were the only thing selected.

With this PR, copying the image in the images example works again.

It removes the superfluous expanded range check.

Fixes: #3665
Reviewers: @CameronAckermanSEL, @lcswillems, @evasteingrims

### DIFF
--- a/packages/slate/src/create-editor.ts
+++ b/packages/slate/src/create-editor.ts
@@ -138,7 +138,7 @@ export const createEditor = (): Editor => {
     getFragment: () => {
       const { selection } = editor
 
-      if (selection && Range.isExpanded(selection)) {
+      if (selection) {
         return Node.fragment(editor, selection)
       }
       return []


majelbstoat:
If you can help me figure out a test for this so it doesn't regress in the future, i'd be obliged.

CameronAckermanSEL:
Thanks, really appreciate you taking the time to narrow the issue down. As far as regression tests, we need this really badly for the react frontend, but the only thing I can think of using is something like cypress or jest and I'm unsure what @ianstormtaylor 's thoughts are on the matter.

PR #3721, opened 2020-06-04
Add missing import in Normalization documentation snippet

hanford:
Add `Transforms` to slate import so code runs when copy/pasted 😃


CameronAckermanSEL:
Thanks very much!

PR #3722, opened 2020-06-05
fix older chrome support

justdanallen:
Fixing a bug.

This is the behavior on Chrome 75 and lower. I tested to Chrome 65. An extra character was being inserted during render that slate didn't know about. So it blows up when you try to type again, because it can't resolve a DOM point. 

In the GIF, I typed a single `k` but 2 are rendered. Then, when you type an additional char, the crash occurs. 

slate-blows-up

#### What's the new behavior?

This now works on Chrome 75 and lower. 

I found that in Chrome, versions below 75, when using the native `beforeinput` event was somehow causing React to fire an extra synthetic event called `textInput` that would inject an extra piece of text into the DOM. This extra character however was not inserted into the SLATE data object. 

Here is where the extra event was affecting the DOM:
Screen Shot 2020-06-04 at 2 24 01 PM

By not using the native `beforeinput` this extra event is not fired and works great! So I added a detection for chrome versions prior to 75 to not use the `beforeinput`. I am honestly, still not sure how that extra synthetic event was being triggered.

slate-doesnt-blow-up

I never opened an issue for this.

#### Testing
I couldn't have solved this with out BrowserStack. I used BrowserStack to manually test the `richText.js` example, by typing and using the bold and itals buttons and hot keys on the following:

*Windows 10*
Chrome: 83, 80,78,76,75,65,60, 50
Edge: 80, 18
Firefox: 75, 68, 60

*Mac High Sierra* 
Safari: 11.1
Chrome: 79, 69

The example worked as expected on all versions above. I also spot tested a few on Windows 8 with no issues. I spot tested Safari and Chrome on iOS 11 it worked. I also spot tested Android on Pixel 4 with Chrome and Firefox and it seemed to work.

IE 11 did not work, but also the example site didn't render. 

Would be good to implement write auto tests on BrowserStack to test against the different browsers. You get free access with an open source lib.

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -14,7 +14,12 @@ import scrollIntoView from 'scroll-into-view-if-needed'
 
 import Children from './children'
 import Hotkeys from '../utils/hotkeys'
-import { IS_FIREFOX, IS_SAFARI, IS_EDGE_LEGACY } from '../utils/environment'
+import {
+  IS_FIREFOX,
+  IS_SAFARI,
+  IS_EDGE_LEGACY,
+  IS_CHROME_LEGACY,
+} from '../utils/environment'
 import { ReactEditor } from '..'
 import { ReadOnlyContext } from '../hooks/use-read-only'
 import { useSlate } from '../hooks/use-slate'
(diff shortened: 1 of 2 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -347,13 +357,13 @@ export const Editable = (props: EditableProps) => {
...
       // @ts-ignore The `beforeinput` event isn't recognized.
       ref.current.addEventListener('beforeinput', onDOMBeforeInput)
     }
 
     return () => {
-      if (ref.current) {
+      if (ref.current && HAS_BEFORE_INPUT_SUPPORT) {
         // @ts-ignore The `beforeinput` event isn't recognized.
         ref.current.removeEventListener('beforeinput', onDOMBeforeInput)
       }
```
justdanallen, line 369: Because Chrome 75 and lower technically works with beforeinput, but not properly, I added a guard her to not register the event.

CameronAckermanSEL:
Thanks a ton @justdanallen ! Really appreciate the improvement here!

justdanallen:
@CameronAckermanSEL No problem! This is a really good library, happy to help.

Looks like a release was just cut 2 days ago. Any idea when the next one will go out?

Manwzy:
when will the next release go out?

Cremz:
would be awesome if this was released.. can we please get a release? pretty please :)

PR #3724, opened 2020-06-05
Fix placeholder text alignment

dmarkow:
Using verticalAlign: text-top was causing a slight height change when entering actual content. So typing the first letter caused the entire editor height to shrink by roughly 1px (notice the bottom edge of the editor jumping as I type)

screencast 2020-06-05 09-35-31

Fixing a bug

#### What's the new behavior?

The height no longer changes.

Remove `verticalAlign: "text-top"` from the placeholder's inline style.


CameronAckermanSEL:
Thanks @dmarkow !

PR #3726, opened 2020-06-07
Replace Iterable with Generator.

BrentFarese:

Fixing a bug.

#### What's the new behavior?

Replace the `Iterable` type used in Slate core with the correct `Generator` type.

The `Iterable` type used in Slate core is not technically correct.  What is returned from most `*functions` is a `Generator`.  The types should be updated to reflect this so you can use things like `.next()` and not have Typescript complain at you that `.next()` doesn't exist on the `Iterable` type.

Fixes: #
Reviewers: @CameronAckermanSEL, @timbuckley

### DIFF
--- a/config/rollup/rollup.config.js
+++ b/config/rollup/rollup.config.js
@@ -180,7 +180,7 @@ function configure(pkg, env, target) {
           file: `packages/${pkg.name}/${pkg.module}`,
           format: 'es',
           sourcemap: true,
-        }
+        },
       ],
       // We need to explicitly state which modules are external, meaning that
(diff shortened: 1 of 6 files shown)


BrentFarese:
@CameronAckermanSEL I see this is tagged for `0.59.0`.  Any idea timeline on that?  Thanks!

BrentFarese:
We have been using these types in our environment for 3-4 weeks and they seem to definitely be correct.  I am going to merge, unless you have comments @CameronAckermanSEL?  Thanks!

PR #3732, opened 2020-06-12
docs: add GraphCMS to products using slate

feychenie:

No, it is adding GraphCMS to the products actively using slate


CameronAckermanSEL:
Thanks! Welcome to the party!

PR #3735, opened 2020-06-14
Fix intersected Text decorations && add tests for that case

cudr:
#### What's the new behavior?

It seems that intersecting decorations do not separate the text correctly. This creates the problem of implementing multiple cursors when they intersect. 
I came across this when displaying the intersection of several cursors on the same line at https://slate-collaborative.herokuapp.com. Fixed it, added tests for this case.

Also relates and fix [#3714](https://github.com/ianstormtaylor/slate/pull/3714)

Now Text decorations function produce correct list of leaves.

Fixes: #3693

### DIFF
--- a/packages/slate/test/interfaces/Text/decorations/collapse.js
+++ b/packages/slate/test/interfaces/Text/decorations/collapse.js
@@ -1,0 +1,85 @@
+import { Text } from 'slate'
+
+export const input = [
+  {
+    anchor: {
+      path: [0],
+      offset: 1,
+    },
+    focus: {
+      path: [0],
+      offset: 2,
+    },
+    decoration1: 'decoration1',
+  },
+  {
+    anchor: {
+      path: [0],
+      offset: 2,
+    },
+    focus: {
+      path: [0],
+      offset: 2,
+    },
+    decoration2: 'decoration2',
+  },
+  {
+    anchor: {
+      path: [0],
+      offset: 2,
+    },
+    focus: {
+      path: [0],
+      offset: 3,
+    },
+    decoration3: 'decoration3',
+  },
+  {
+    anchor: {
+      path: [0],
+      offset: 4,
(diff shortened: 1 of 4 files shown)


delijah:
@CameronAckermanSEL do you think we can merge this?

ulion:
Just curious, the PR is approved, but when will it get merged?

packages/slate/test/interfaces/Text/decorations/collapse.js
```diff
@@ -0,0 +1,85 @@
...
+  },
+  {
+    text: 'b',
+    mark: 'mark',
+    decoration1: 'decoration1',
+  },
+  {
+    text: '',
+    mark: 'mark',
+    decoration1: 'decoration1',
```
ulion, line 67: wouldn't without decoration1 here be the more natural result? since decoration1 indeed just care the text 'b', not the zero-length item after that?

cudr: This is additional decoration between two character segments, so it should contains all decorations overlapped in this point. This will be useful, If you want to control multi-cursor overlap (for example). In this case you need to know, where they overlap in order to show this overlap in view.

delijah:
@BrentFarese could we merge this pull request?

BrentFarese:
> @BrentFarese could we merge this pull request?

I'm glad to, but have not really touched or dealt with the decorations code at all.  Ideally, I'd have someone else comment/approve it.  cc @ianstormtaylor any comment?  If not, I'll merge this in in a week or so as you have tests and it looks sound.  Ping me if I forget...

Thanks.

ianstormtaylor (approved):

delijah:
Whohoooooo, awesome! Thank you @BrentFarese and @ianstormtaylor ! 🥳

PR #3743, opened 2020-06-17
Remove legacy target operation property

jason-krypton:
Removing some operation properties that are not used. Looks like cruft from earlier revisions.


CameronAckermanSEL:
Thanks @jason-krypton for tidying up the operations again :)

PR #3746, opened 2020-06-20
fix: autoscroll when block is bigger than viewport

gztomas:

This is fixing an issue https://github.com/ianstormtaylor/slate/issues/3146

#### What's the new behavior?

[Before:](https://codesandbox.io/s/slate-scroll-issue-hlwv0)
before mov
After:
after mov

`scroll-into-view-if-needed` can only scroll **elements** into view. In this case, the element is bigger than the viewport, and we actually want to scroll the range of the selection into the viewport. `Range` already implements `getBoundingClientRect`. That should be all what `scroll-into-view-if-needed` needs to do its thing. But `scroll-into-view-if-needed` won't just work if we pass a range instead of an element. While being a bit hacky, overriding `getBoundingClientRect` of the block with the `getBoundingClientRect` from the range works wonderfully.

Fixes: #3146


rglusk:
@gztomas can you please tag @ianstormtaylor as a reviewer? maybe that will help to get this reviewed and merged sooner

packages/slate-react/src/components/editable.tsx
```diff
@@ -186,10 +186,14 @@ export const Editable = (props: EditableProps) => {
     if (newDomRange) {
       domSelection.addRange(newDomRange!)
       const leafEl = newDomRange.startContainer.parentElement!
+      leafEl.getBoundingClientRect = newDomRange.getBoundingClientRect.bind(
+        newDomRange
+      )
```
rglusk, line 191: when testing this fix, did you scroll around a lot? Worried about performance issues when building and destroying this every time you scroll

gztomas: I haven't found any performance issues so far. If that was the case however I think we would see the issue on typing rather than on scrolling given this piece of code is only called when the selection changes, not the scroll. I believe this should be a cheap operation either way. Or at least probably cheaper than the DOM access operations that happen on `toDOMRange` sequential to this.

CameronAckermanSEL:
Hi @gztomas , thank you so much for the contribution.

I went to go test this out on the deployed preview examples and it didn't seem to have any effect. What I tried was going to the rich text example, adding enough text for there to be a scollbar, and then I typed while the cursor was in and out of the view port. In both cases, the behavior was identical to what is now on latest. I also tried this by ensuring the text was in a single block and still no improvement.

Is there something I'm missing in the steps to reproduce?

gztomas:
Hi @CameronAckermanSEL 
Thanks for looking into this. What I have seen this change fixes happens when the slate editor height has a limit. I can repro the issue in the preview examples only after some css changes:
scroll-issue mov
Please let me know if that makes sense

rglusk:
@CameronAckermanSEL have you had a chance to re-review this? Is there anything else you need to repro?

rglusk:
@gztomas can we push this change out? its been a few months and there's been no comments around this change causing issues.

cc: @CameronAckermanSEL @ianstormtaylor

gztomas:
@rglusk I don't think there is anything to be done on our side. There are a lot of PRs unattended fixing really important stuff. This repository seems abandoned rather than community driven.

rglusk:
@timbuckley, i see you recently merged new code. Can you please either merge this for us, or give @gztomas and I write access to this repo? Thank you!!

anhphung97:
This is a problem for me as well, would like to see it gets fixed soon 👍 @timbuckley

gztomas:
In case it was useful for you, this is the hack I'm using to fix this on my side
```ts
// HACK for fixing Slate auto-scroll issue
// Overriding the rect for the Slate block with the selection block so Slate
// auto-scrolls to where the cursor is rather than auto-scrolling the whole
// block. If the block is big enough then the auto-scroll can make the cursor be
// outside of the viewport.
ReactEditor.toDOMRange = (editor, selection) => {
  const range = toDOMRange(editor, selection);
  range.startContainer.parentElement?.getBoundingClientRect = range.getBoundingClientRect.bind(
    range
  );
  return range;
};
```

rglusk:
thanks @gztomas, that works :) hopefully this change will get merged at some point so other people can use your fix too!

ianstormtaylor:
Thanks @gztomas!

packages/slate-react/src/components/editable.tsx
```diff
@@ -205,10 +205,14 @@ export const Editable = (props: EditableProps) => {
...
       }
       const leafEl = newDomRange.startContainer.parentElement!
+      leafEl.getBoundingClientRect = newDomRange.getBoundingClientRect.bind(
+        newDomRange
+      )
       scrollIntoView(leafEl, {
         scrollMode: 'if-needed',
         boundary: el,
       })
+      delete leafEl.getBoundingClientRect
```
juliankrispel, line 215: 🤔 this throws a typescript error could you elaborate why this is needed @gztomas?

gztomas: @juliankrispel this restores the override happening at line `208`, so any future call to `leafEl.getBoundingClientRect` will get the native `getBoundingClientRect` implementation from `leafEl` prototype. This is for cleanup purposes, as this code doesn't know who else might need to call `getBoundingClientRect` for this element. I guess TS can't know that even deleting this, `getBoundingClientRect ` will still be available, from the prototype.

juliankrispel: thanks for explaining @gztomas

PR #3749, opened 2020-06-24
Use shadow dom if available

davidruisinger:

_bug_

#### What's the new behavior?

If `<Editable />` is rendered within a shadowRoot, internal `editor.selection` is not updated correctly which (amongst other issues) leads to the cursor always jumping back to the beginning of the text while typing.

qbyHEGwORJ

`window.getSelection` (which is internally used in `slate-react`) refers to the shadow host selection. Instead of always relying on `window.document` I've added a `getDocumentOrShadowRoot` util which either returns the `ShadowRoot` (if active Element is rendered within a ShadowDom) or `Document`.
So instead of `window.getSelection()` we use `getDocumentOrShadowRoot ().getSelection()`


flavordaaave:
@CameronAckermanSEL would be great if you could have a look. THX!

packages/slate-react/src/utils/dom.ts
```diff
@@ -98,6 +98,15 @@ export const normalizeDOMPoint = (domPoint: DOMPoint): DOMPoint => {
   return [node, offset]
 }
 
+/**
+ * Returns ShadowRoot if active element is renderd within Shadow DOM,
+ * otherwise it returns the Document Object.
+ */
+
+export const getDocumentOrShadowRoot = (): Document | ShadowRoot => {
+  return window.document.activeElement?.shadowRoot ?? window.document
```
AndrewScull: This seems work for one level of shadowRoot but not if they are nested. I think it needs to be done bottom up from the element in question to find the shadowRoot that contains that element.

I've only come across searches up the node tree to find the appropriate root with something like below (illustrative only). It's probably something that's be computed once and stored on the editor for use when it's needed.

```
let n = editor;
while (n.
for (let n = n.parentNode; n; n = n.parentNode) {
  if (n instanceof Document || n instanceof ShadowRoot) break;
}
const documentOrShadowRoot = n || window.document;
```

flavordaaave: Hmm, interesting. Could you provide a simple example codesandbox (or something similar)?
You can look at the example I created in this PR as a starting point.

AndrewScull: I [forked your branch](https://github.com/AndrewScull/slate/commits/use_shadow_dom_if_available) of this pull request and updated the Shadow DOM example to also include a nested Shadow DOM.

AndrewScull: I've updated that branch with a commit that gets the nested version working too. It's far from being PR-ready and there are some behaviours that I can't yet fully explain so if you're more familiar with these APIs it'd be great to hear what you think.

AndrewScull: I had another go at this ([branch](https://github.com/AndrewScull/slate/tree/shadowdom)) after coming across `getRootNode()` to avoid the tree traversal. However, I'm still running into an issue with key events that seems to be a [React issue](https://github.com/facebook/react/issues/9242) with shadow DOM. On Chrome, the undo key combo doesn't trigger the event handler and so the undo doesn't happen.

flavordaaave: Nice work. I merged your changes into this PR. I've also updated a few smaller things (sorry for the mess with all those commits. I was trying out a few things...)

flavordaaave:
@CameronAckermanSEL , @ianstormtaylor , @BrentFarese , @timbuckley could one of you guys please have a look at this PR?
I really need to get slate working within ShadowDom and it seems that I'm not the only one with this use-case. THX!

ellenblaine:
Hey @flavordaaave thanks for putting this PR together! I also have this use case and am wondering what the status of this PR is. I've really enjoyed Slate in other projects and would like to use it again. Thank you!

flavordaaave:
> Hey @flavordaaave thanks for putting this PR together! I also have this use case and am wondering what the status of this PR is. I've really enjoyed Slate in other projects and would like to use it again. Thank you!

Hey @ellenblaine unfortunately none of the maintainers seems to be interested in reviewing/merging this PR which is why we have created our own (internal) for for now. We still hope that at some point slate will be compatible with beeing rendered within a ShadowDom.

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -81,6 +83,29 @@ export const ReactEditor = {
...
+   * Find the DOM node that implements DocumentOrShadowRoot for the editor.
+   */
+
+  findDocumentOrShadowRoot(editor: ReactEditor): Document | ShadowRoot {
+    const el = ReactEditor.toDOMNode(editor, editor)
+    const root = el.getRootNode()
+
+    if (!(root instanceof Document || root instanceof ShadowRoot))
+      throw new Error(
+        `Unable to find DocumentOrShadowRoot for editor element: ${el}`
```
ellenblaine, line 96: Noticed an interesting issue using this branch myself: The `selectionchange` event fires as the editor unmounts (we have a drawer component that contains the editor -- closing the drawer causes `selectionchange` to fire). When this happens, the `selectionchange` event fires and `findDocumentOrShadowRoot` is called. This error is hit like so:
```
// ... selectionchange event calls this function ...
const el = ReactEditor.toDOMNode(editor, editor) // el is the editor, which isn't currently in the DOM
const root = el.getRootNode() // root is another element that is not in the DOM or the shadow root
// error throws because root is not Document or ShadowRoot
```

Any ideas? No worries if not -- this could definitely be an issue that's unrelated to Slate.

tomdng:
> > Hey @flavordaaave thanks for putting this PR together! I also have this use case and am wondering what the status of this PR is. I've really enjoyed Slate in other projects and would like to use it again. Thank you!
> 
> Hey @ellenblaine unfortunately none of the maintainers seems to be interested in reviewing/merging this PR which is why we have created our own (internal) for for now. We still hope that at some point slate will be compatible with beeing rendered within a ShadowDom.

Hey @flavordaaave have you attempted to contact the maintainers through the [Slate Slack channel](https://slate-slack.herokuapp.com/) by any chance? It looks like the maintainers have their hands full trying to establish a release system for Slate and dealing with issues with the default branch for this repo. Perhaps you might find better luck there raising an issue there since there's a lot of noise with the active issues/PRs here.

flavordaaave:
> > > Hey @flavordaaave thanks for putting this PR together! I also have this use case and am wondering what the status of this PR is. I've really enjoyed Slate in other projects and would like to use it again. Thank you!
> > 
> > 
> > Hey @ellenblaine unfortunately none of the maintainers seems to be interested in reviewing/merging this PR which is why we have created our own (internal) for for now. We still hope that at some point slate will be compatible with beeing rendered within a ShadowDom.
> 
> Hey @flavordaaave have you attempted to contact the maintainers through the [Slate Slack channel](https://slate-slack.herokuapp.com/) by any chance? It looks like the maintainers have their hands full trying to establish a release system for Slate and dealing with issues with the default branch for this repo. Perhaps you might find better luck there raising an issue there since there's a lot of noise with the active issues/PRs here.

@tomdng Yes, I posted the PR there as well (`#contributing` channel) but I didn't help either.
There might be a small chance that this is working out of the box with React 17 (with the changes to the event delegation). Unfortunately I haven't had time to validate this but will do asap

cpmotion:
Can we make this fix happen? Regularly utilize Slate inside a shadow DOM. PR looks good.

packages/slate-react/src/plugin/react-editor.ts
davidruisinger: I couldn't reproduce this myself, yet. We are also unmounting our slate editor when the user navigates to a different section/page but so far I didn't get this error. 
Could you setup a minimal example illustrating this issue?
One idea how to get around this would be to call `blur()` before the drawer is closed but obviously this would just be a temporally workaround.

davidruisinger:
> > > > Hey @flavordaaave thanks for putting this PR together! I also have this use case and am wondering what the status of this PR is. I've really enjoyed Slate in other projects and would like to use it again. Thank you!
> > > 
> > > 
> > > Hey @ellenblaine unfortunately none of the maintainers seems to be interested in reviewing/merging this PR which is why we have created our own (internal) for for now. We still hope that at some point slate will be compatible with beeing rendered within a ShadowDom.
> > 
> > 
> > Hey @flavordaaave have you attempted to contact the maintainers through the [Slate Slack channel](https://slate-slack.herokuapp.com/) by any chance? It looks like the maintainers have their hands full trying to establish a release system for Slate and dealing with issues with the default branch for this repo. Perhaps you might find better luck there raising an issue there since there's a lot of noise with the active issues/PRs here.
> 
> @tomdng Yes, I posted the PR there as well (`#contributing` channel) but I didn't help either.
> There might be a small chance that this is working out of the box with React 17 (with the changes to the event delegation). Unfortunately I haven't had time to validate this but will do asap

Unfortunately the issue still remains with React 17. So these changes are still needed to get slate working within a shadow DOM. I updated this PR to resolve all open conflicts with the master branch.

Mangatt:
Is it possible that this PR is causing crash in Chromium? See #4170

juliankrispel:
> Is it possible that this PR is causing crash in Chromium, @davidruisinger ? See #4170

Indeed this brakes iframe support,[ here's is the code with the exception thrown:
](https://github.com/ianstormtaylor/slate/commit/0473d0bf93808b0e4e98abe833b7f7f4f5aff3b1#diff-06bf833c7913658864514eb335e23b8ddb7a55888d004bd68d8d1293933103cbR108-R110)
```ts
if (!(root instanceof Document || root instanceof ShadowRoot))
  throw new Error(
    `Unable to find DocumentOrShadowRoot for editor element: ${el}`
  )
```

for context: gets thrown if `root` refers to a document element inside an iframe.

packages/slate-react/src/plugin/react-editor.ts
zwily: FWIW, I'm seeing this same behavior with an unmounting editor... I tried `blur`ing it with a `useEffect` but that's probably too late.. still errors, but this time in the `ReactEditor.blur(editor)`.

juliankrispel: I'm trying to fix an issue related to this change, can you explain why we need to throw an exception here @davidruisinger? I believe I'm missing context. For reference here's my [draft pr](https://github.com/ianstormtaylor/slate/pull/4219) for fixing the iframe crash, would be good to get some 👀 on it before I add cypress tests

TheSpyder: At the very least, these `instanceof` checks should use `getWindow` the way other similar checks do

https://github.com/ianstormtaylor/slate/blob/6a1376332bbd2567336c444c57c1e64fdf706feb/packages/slate-react/src/components/editable.tsx#L383-L384

https://github.com/ianstormtaylor/slate/blob/8a2a60b052dd666ef4ec128863a56b251aec8ea8/packages/slate-react/src/utils/dom.ts#L67-L68

ottworks:
I'm getting `Unable to find DocumentOrShadowRoot for editor element: [object HTMLDivElement]` when I unmount an editor (it's in a react-bootstrap modal). getRootNode returns the now-detached modal div.

PR #3768, opened 2020-07-07
fix: IME repeat input

gyh9457:

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -494,8 +494,10 @@ export const Editable = (props: EditableProps) => {
               hasEditableTarget(editor, event.target)
             ) {
               event.preventDefault()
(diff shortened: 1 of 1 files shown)


timbuckley:
Is there a test you could write that would cover this change? Was there an issue that describes the bug being fixed?

gyh9457:
IME-repeat

as shown above

when I use IME to input, `onBeforeInput` and `onCompositionEnd` will be triggered at the same time
so it repeats the input twice

There are no unit test in `slate-react`，so I won’t add test cases for now

@timbuckley

githoniel:
I can confirm this bug exist in chrome x65 75.0.3770.80 in Win10

thezero:
Is there anything blocking this PR from being merged?

ianstormtaylor:
Thanks @gyh9457!

PR #3771, opened 2020-07-08
Fix/edge legacy

gerkensm:
Upgrade to immer 7.0.5, add code to only enable ES5 when needed


mfranzs:
Thanks for taking the time to set this up! I'm also running into problems with Immer on legacy browsers on the latest slate version -why did you end up closing this PR and is there anything that I can do to help?

gerkensm:
This was meant to be a PR for our fork of slate, I accidently submitted it to the wrong repo. It not only contains an upgrade of immer, but also two hacks to make slate enable immer's ES5 support on demand (which is required by more recent immer versions) for IE11 only and not for Edge legacy. I honestly assumed the problems with Edge legacy were a consequence of our exotic transpilation setup (Meteor with babel and a third-party Typescript support package), so I didn't bother filing a proper bug report. The demo page loads fine on IE11 and Edge legacy, so I blamed the problems on our transpilation setup and moved on. If there's anything I can help with, please let me know!

mfranzs:
Thanks for the info! I'm actually using a similar transpilation setup, which makes me wonder if that caused the problem originally.
Regardless, I've tried locally overriding Slate's version of Immer to the latest version which seems to have resolved my original problems. Upgrading Immer might thus be a good idea for Slate in general, although I haven't run the full test suite.

PR #3779, opened 2020-07-10
Resolves iamstormtaylor/slate#3778: Check if editable target before handling paste event

### LINKED ISSUE #3778: Slate doesn't check if the node pasted into is a Slate (editable) node before triggering onPaste prop
A bug 

#3670 has caused a regression where the onPaste handler may fire for non-Slate nodes that are in the Slate tree. 

image

This is especially bad if, for example, you have an input or textarea inside a void node; those paste events are caught by the onPaste handler because of the ordering of the checks in the onPaste handler.

This is a problem even if you render the input in a React portal as some contributors suggest doing, because the portal still respects React's tree for event propagation

Background thread: https://slate-js.slack.com/archives/C1RH7AXSS/p1594190718230800
Slate: 0.58.2+
Browser: any
OS: any

Slate's `Editable` component should check if the target is an editable Slate node before deciding to fire the onPaste prop.
I'm aiming to have a pull request up for this early next week.

mitchellbusby:
Resolves #3778 by checking for an editable event target before handling the paste event (calling the provided paste prop).

#### What's the new behavior?

Does what it says on the tin. We don't want the paste handler to be firing for non-Slate nodes (e.g. an input in the React tree). I think that #3670 accidentally reordered the two calls that were being moved, so this just reorders it correctly.

N/A

Fixes: #3778 
Reviewers: @CameronAckermanSEL (since you have context)

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -946,11 +946,11 @@ export const Editable = (props: EditableProps) => {
             // when "paste without formatting" option is used.
             // This unfortunately needs to be handled with paste events instead.
             if (
+              hasEditableTarget(editor, event.target) &&
               !isEventHandled(event, attributes.onPaste) &&
               (!HAS_BEFORE_INPUT_SUPPORT ||
                 isPlainTextOnlyPaste(event.nativeEvent)) &&
-              !readOnly &&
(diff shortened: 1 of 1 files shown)


CameronAckermanSEL:
Thanks @mitchellbusby !

Issue #3767, opened 2020-07-06
Request: Convert tests to TypeScript


timbuckley:
Slate tests are currently written in JavaScript.

Converting them to TypeScript would allow TypeScript consumers of the library to catch possible regressions or other changes to the API that are specific to TypeScript implementation.

PR #3784, opened 2020-07-13
Convert Slate tests to TypeScript

mdmjg:
feature

#### What's the new behavior?

The tests were converted to typescript, with the exception of the tests inside the slate-hyperscript folder because those would throw errors when converted.
I also modified the eslintrc file so that it would exclude .ts and .tsx files when checking  for import/no-extraneous-dependencies.

When you run yarn test, none of the tests should fail.

Fixes: #3767
Reviewers: @timbuckley


package.json
```diff
@@ -87,5 +88,8 @@
     "slate-react": "*",
     "source-map-loader": "^0.2.4",
     "typescript": "^3.7.2"
+  },
+  "dependencies": {
+    "react-js-to-ts": "^1.4.0"
```
timbuckley, line 93: It seems `react-js-to-ts` is in both deps and devDeps, it should just be in one if at all
```suggestion
```

Also, I don't seem to see it imported directly, is it needed?

Issue #3725, opened 2020-06-05
Add Custom TypeScript Types to Slate


thesunny:
This post describes a proposal to add custom types to Slate with the following features:

- Extend types in Slate like `Element` and `Text` with custom properties
- Does not require generics at every call site (define custom types once only)

Limitation:

- One custom type definition per project. You can't have multiple editors in one project **with different custom types**. You can have multiple instances of editors.

## Simple Schema Example

For simplicity, this is a model for a bare bones version of Slate's types without custom types. I use this to explain the proposal.

```ts
export type Element = {
  children: Text[]
}

export type Text = {
  text: string
}

// gets children text nodes from an element
export function getChildren(element: Element) {
  return element.children
}

// gets merged text from children text node
export function getText(element: Element) {
  return element.children.map((text) => text.text)
}
```

The goal:
- We want to be able to customize `Element` and `Text`.
- We need `getChildren` and `getText` to keep working.
- We need to use the customized `Element` and `Text` in our own code without generics at each call site.

## How To Use Custom Types

This section explains how to use Custom Types by the end user. I will then explain and show the code for how to add custom types to the schema in the sample above.

```ts
// import statement to `./slate` would actually be `slate` but I'm importing
// from a local file.
import { getChildren, Element } from "./slate"

// this is where you customize. If you omit this, it will revert to a default.
declare module "./slate" {
  export interface CustomTypes {
    Element:
      | { type: "heading"; level: number }
      | { type: "list-item"; depth: number }
    Text: { bold?: boolean; italic?: boolean }
  }
}

// This uses the custom element. It supports type discrimination.
// `element.heading` works and `element.depth` fails.
function getHeadingLevel(element: Element) {
  if (element.type !== "heading") throw new Error(`Must be a heading`)
  // Uncomment `element.depth` and you get a TypeScript error as desired
  // element.depth
  return element.level
}

// This shows that the regular methods like `getChildren` that are imported
// work as expected.
function getChildrenOfHeading(element: Element) {
  if (element.type !== "heading") throw new Error(`Must be a heading`)
  return getChildren(element)
}
```

Here's a screenshot showing what happens when we uncomment `element.depth` and that the proper typescript error shows up:

![image](https://user-images.githubusercontent.com/469587/83919903-77e53400-a730-11ea-9ffc-04fc96ee59f5.png)

## How It Works

Here is the source code for `./slate.ts`

```ts
// This would be Element as per Slate's definition
export type BaseElement = {
[...]

Issue #3680, opened 2020-05-13
 Support TypeScript declaration merging


ccorcos:
Slate: 0.58.1

I'd really like to refine the internal editor types. Typescript supposedly has support for this:

https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation

But I think its not working currently, because for example text.d.ts exports a interface Text and a const Text so I think Typescript is having trouble with that. 

Ideally, I can just write this code:

```ts
declare module "slate" {
	interface Text {
		bold?: string
	}
}
```

And now the text type everywhere in my codebase has a bold property.

ccorcos:
I can confirm that renaming `interface Text` → `interface IText` solves this problem.

ccorcos:
Just to give you a better sense of what I'm going after, this code is typesafe and lets me modularlize features.

```ts
declare module "slate" {
	interface IText {
		bold?: string
	}
	interface IEditor {
		bold: ReturnType<typeof createBoldInterface>
	}
}

function createBoldInterface(editor: IEditor) {
	return {
		isActive() {
			const [match] = Editor.nodes(editor, {
				match: (n) => n.bold === true,
				universal: true,
			})
			return !!match
		},
		toggle() {
			const isActive = editor.bold.isActive()
			Transforms.setNodes(
				editor,
				{ bold: isActive ? null : true },
				{ match: (n) => Text.isText(n), split: true }
			)
		},
	}
}

function withBold(editor: IEditor) {
	editor.bold = createBoldInterface(editor)
	return editor
}
```

That said, something I'm not super clear about yet -- what's the point of having methods on the editor instance and separating out all the static Editor helper functions that accept the editor as the first argument?

CameronAckermanSEL:
This would seem like a good improvement to me, personally

ccorcos:
I'll make a PR

ccorcos:
Hmm. It won't let me make a PR... 
![image](https://user-images.githubusercontent.com/1794527/81845517-8495b400-9505-11ea-8807-094de8249fe1.png)

Here's what I did though: https://github.com/ianstormtaylor/slate/compare/master...ccorcos:master

I've tried to only make a minimal change (hence the last two reverts). But I think in general, it's a bad pattern to shadow variable names with types.

TheSpyder:
Side note
>  it's a bad pattern to shadow variable names with types

[...]

PR #3762, opened 2020-07-02
syntax highlighting

mdmjg:
#### Is this adding or improving a _feature_ or fixing a _bug_?
feature

#### What's the new behavior?
Syntax highlighting for languages: javascript, java, python, php, html, css, sql

#### How does this change work?
The Editable section of the component, is decorated with a function that labels each leaf according to Prism.tokenize(text, grammar). Since language is a state, this tokenize function will be called whenever the user selects a new language. The leaf component will assign a some CSS to each leaf depending to the token assigned by Prism. 
If you want to extend the available languages, you must import them at the top of the file.

#### Have you checked that...?

- [x ] The new code matches the existing patterns and styles.
- [ x] The tests pass with `yarn test`.
- [ ] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.) (It tells me the command failed)
- [ x] The relevant examples still work. (Run examples with `yarn start`.)

#### Does this fix any issues or need any specific reviewers?

Fixes: #3733
Reviewers: @

PR #3785, opened 2020-07-13
Custom and Extended Types

mdmjg:
feature

#### What's the new behavior?
Uses interfaces to support declaration merging and types for type discrimination. 

In order to use declaration merging, the user must augment the Slate module. This detail could be explained in an additional example that could be added for this feature.

- [ x] The tests pass with `yarn test`.
- [ x] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [ x] The relevant examples still work. (Run examples with `yarn start`.)

Fixes: #3725
Reviewers: @thesunny @timbuckley


packages/slate/src/interfaces/custom-extensions.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * The `CustomExtensions` interface extends types
+ */
+
+export interface CustomExtensions {
```
BrentFarese, line 5: I still think we need to update Slate's types for `Text` and `Element` to use `ExtendedType` below?  See @thesunny suggestion in the implementation issue - https://github.com/ianstormtaylor/slate/issues/3725.  For instance:

```
// This would be Element as per Slate's definition
export type BaseElement = {
  children: Text[]
}

// This would be Text as per Slate's definition
export type BaseText = {
  text: string
}

export type Text = ExtendedType<"Text", BaseText>
export type Element = ExtendedType<"Element", BaseElement>
```

Then, the exported new types for `Text` and `Element` would be used in Slate's codebase and any developer's use of `CustomExtensions` and `ExtendedType` would propagate to the rest of the codebase.  

Let's discuss if needed.

mdmjg: The issue with this is that when you use `ExtendedType` you are redefining Element or Text according to the definition you gave it inside your CustomExtension. If make the extended Element the new Element, a bunch of errors will be thrown during build, and pretty much 60% of the testcases will fail. Something that I found while making this PR was that in order to call Element's functions like `isElement,` you need to use an alias. If you take a look at my isElement test case, you will see I am calling BaseElement.isElement. If I were to use Element.isElement, the program would not recognize the function because the type Extension is not able to fetch the functions as well.

thesunny: Sorry, I haven't had time to dive into the code so perhaps you could explain the issue and I might be able to help with a solution.

My question is, if the original object is defined as BaseText or BaseElement, and the extended type is created as Text and Element, are you saying that the methods don't come through to the extend type? It looks like in the typing we are doing:

```ts
export type ExtendedType<K extends string, B> =
  unknown extends CustomExtensions[K]
  ? B
  : B & CustomExtensions[K]
```

So shouldn't we get the proper types from both the original with the extensions added (ie. `B & CustomExtensions[K]`)?

I may be misunderstanding the issue in which case a clarification would be helpful. Thanks.

packages/slate/src/interfaces/node.ts
```diff
@@ -468,7 +472,7 @@ export const Node = {
     if (Text.isText(node)) {
       return node.text
     } else {
-      let children = node.children as Array<Descendant>
+      const children = node.children as Array<Descendant>
```
timbuckley, line 475: `Array<x>` is more commonly written as `x[]`

```suggestion
      const children = node.children as Descendant[]
```

packages/slate/src/create-editor.ts
```diff
@@ -217,7 +217,7 @@ export const createEditor = (): Editor => {
           (editor.isInline(node) ||
             node.children.length === 0 ||
             Text.isText(node.children[0]) ||
-            editor.isInline(node.children[0]))
+            editor.isInline(node.children[0] as Element))
```
BrentFarese, line 220: I think we should be staying away from type assertions, if at all possible.  There is a way to have the same code here w/o the type assertion.

packages/slate/src/interfaces/node.ts
BrentFarese: Tim is totally correct but we should avoid the type assertion, if we can.

mdmjg:
Note: You will get some errors when you run yarn build because of the placeholder property in the Leaf interface in slate-react. Some help with that would be super appreciated!

packages/slate-react/src/components/leaf.tsx
```diff
@@ -28,6 +27,8 @@ const Leaf = (props: {
     <String isLast={isLast} leaf={leaf} parent={parent} text={text} />
   )
 
+  // const placeholder = 'placeholder' in leaf && leaf.placeholder
```
BrentFarese, line 30: Do we need this @mdmjg?  Let's remove.

packages/slate-react/src/components/slate.tsx
```diff
@@ -14,7 +14,7 @@ import { EDITOR_TO_ON_CHANGE } from '../utils/weak-maps'
 
 export const Slate = (props: {
   editor: ReactEditor
-  value: Node[]
+  value: Element[]
   children: React.ReactNode
   onChange: (value: Node[]) => void
   [key: string]: unknown
```
BrentFarese, line 20: I think we might be able to get rid of this `[key: string]: unknown`.

packages/slate/src/interfaces/custom-types.ts
```diff
@@ -0,0 +1,18 @@
+import { Descendant } from './node'
+
+export interface BaseElement {
```
BrentFarese, line 3: I think I would move `BaseElement` next to the `Element` type and `BaseText` next to the `Text` type (i.e., back in the interface files for those types).  Easier to read/follow the source that way.

packages/slate-react/src/components/leaf.tsx
mdmjg: ill remove

packages/slate/src/interfaces/text.ts
```diff
@@ -63,6 +60,14 @@ export const Text = {
     return Array.isArray(value) && (value.length === 0 || Text.isText(value[0]))
   },
 
+  /**
+   * Check if some props are a partial of element.
+   */
+
+  isTextProps(props: any): props is Partial<Text> {
+    return !!props.children
+  },
```
BrentFarese, line 69: I'm making a couple fixes and will push a commit soon with this fix too, but it should be something like this:

```suggestion
  isTextProps(props: any): props is Partial<Text> {
    return (props as Partial<Text>).text !== undefined;
  },
```

packages/slate/src/transforms/general.ts
```diff
@@ -65,7 +65,8 @@ export const GeneralTransforms = {
         if (Text.isText(node) && Text.isText(prev)) {
           prev.text += node.text
         } else if (!Text.isText(node) && !Text.isText(prev)) {
-          prev.children.push(...node.children)
+          const children = node.children as Element[]
```
BrentFarese, line 68: I changed the type of `editor.children` to `Descendant[]` instead of `Element[]` and am going to revert some of these changes.  We can discuss but I think the more appropriate type for `editor.children` is `Descendant[]`.

mdmjg: okay! I'll wait for you push then

packages/slate/src/interfaces/custom-extensions.ts
BrentFarese: @thesunny could use a review from your end.  I think the code is looking more final now so it's probably a good time to review.  Thanks!

@CameronAckermanSEL can you review too?

BrentFarese:
This also fixes #3680.

lcswillems:
This seems an amazing PR!!

I don't get why @CameronAckermanSEL, @ianstormtaylor and @thesunny should review this PR? They don't seem active. Waiting for their review will just make this PR never get merged and I don't think it is the best thing for the community.

What I suggest is that this PR get merged and if there are issues, they will be raised and addressed in a future release, it is not like if the project was currently in a stable state...

Thank you @mdmjg for the awesome work you did in this PR and previous PRs (https://github.com/ianstormtaylor/slate/pull/3762, https://github.com/ianstormtaylor/slate/pull/3784)!

packages/slate-react/src/components/editable.tsx
```diff
@@ -9,6 +9,7 @@ import {
   Transforms,
   Path,
 } from 'slate'
```
CameronAckermanSEL, line 11: is this a potential circular dependency?

packages/slate/src/interfaces/node.ts
```diff
@@ -222,7 +241,7 @@ export const Node = {
         }
       }
 
-      delete r.selection
+      if (Editor.isEditor(r)) delete r.selection
```
CameronAckermanSEL, line 244: whats this for? is this fixing some other bug?

packages/slate/src/transforms/node.ts
```diff
@@ -729,7 +730,7 @@ export const NodeTransforms = {
 
         Transforms.liftNodes(editor, {
           at: range,
-          match: n => node.children.includes(n),
+          match: n => Element.isAncestor(node) && node.children.includes(n),
```
CameronAckermanSEL, line 733: does this fix something?

support/fixtures.js
```diff
@@ -25,7 +25,7 @@ export const fixtures = (...args) => {
       }
       if (
         stat.isFile() &&
-        (file.endsWith('.js') || file.endsWith('.ts') || file.endsWith('.tsx')) &&
```
CameronAckermanSEL, line 28: i think we probably need to readd the .ts extension

CameronAckermanSEL:
Thank you *so* much for all the effort put into this PR. This is going to be a great improvement for users who are in typescript. 

I had a few small questions about some code that looks like it was changed for reasons perhaps unrelated to this PR. Additionally, there was some code in the fixtures function that was changed that we might want to change back.

The final concern I had was that there was a pretty massive amount of churn in `yarn.lock` but no corresponding changes to `package.json`. I think we may want to revert `yarn.lock` to what it was, unless these changes can be justified. Surprise changes to the lock file are a little bit of a red flag to me.

Thanks again!

packages/slate/src/transforms/node.ts
timbuckley: Yes, `node` is not guaranteed to be an `Element` or `Editor`, and thus might not have a `children` property. Type-guarding it first as an Ancestor prevents attempts to to `node.children.includes` on a `Text` node

packages/slate/src/interfaces/node.ts
timbuckley: Same reason as below: Including the type-guard means TypeScript won't complain anymore since the `selection` key is not guaranteed on any `Node` other than an `Editor`.

Issue #3781, opened 2020-07-12
Fix types of Placeholder Element in Slate React


mwood23:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

The type for the placeholder property isn't correct.

The [prop is typed](https://github.com/ianstormtaylor/slate/blob/97dbab13a88cbeb7b9d41287fc7e6a2086211f1e/packages/slate-react/src/components/editable.tsx#L90):

```tsx
placeholder?: string
```

At the [callsite](https://github.com/ianstormtaylor/slate/blob/97dbab13a88cbeb7b9d41287fc7e6a2086211f1e/packages/slate-react/src/components/leaf.tsx#L49) it's invoked like this:

```tsx
{leaf.placeholder as React.ReactNode}
```

#### What's the expected behavior?

Placeholder should be typed as

```tsx
placeholder?: React.ReactNode
```

I think it would also be nice to add this to the docs or an example so users could see how they can render elements as a placeholder. It's also a workaround for https://github.com/ianstormtaylor/slate/issues/3780.

Let me know if you'd like a PR!

timbuckley:
@mwood23 That sounds like a good suggestion. Could you make a PR for it?

mwood23:
@timbuckley For sure, just put the PR.

Issue #3780, opened 2020-07-12
Add Data Attribute to Placeholder Property in React Editable


mwood23:
#### Do you want to request a _feature_ or report a _bug_?

Feature.

#### What's the current behavior?

Right now, you cannot easily style the placeholder property because the styles are inlined and there's no easy way to determine what's a placeholder. There's a PR, https://github.com/ianstormtaylor/slate/pull/3461, suggesting adding a `className`, but I think that opens up a can of worms on adding another prop to the base API, and also brings up the issue of where to draw the line on what components should get `className`s.

#### Proposed Behavior

It would be nice to be able to determine if something is a placeholder element not only for styling, but for testing purposes as well so you can asserts on the property existing. I propose adding a `data-state-placeholder="true"`to the placeholder element. Based on the [code](https://github.com/ianstormtaylor/slate/blob/97dbab13a88cbeb7b9d41287fc7e6a2086211f1e/packages/slate-react/src/components/leaf.tsx#L31), it should be a quick addition and maintains consistency with the `data-slate-` attributes on the other elements.

You can style the placeholder like this in your CSS:

```
[data-state-placeholder="true"] {
    /* Styles */
}
```

This keeps the surface of the API low, allows for styling, and gives us the ability to assert on placeholders existing in the DOM.

If this is something you'd be willing to take a PR for let me know!

kripod:
This is a great idea in my opinion, I would love to see an implementation!

_(Disclaimer: I'm **not** a maintainer of this project 😄)_

kripod:
There [seems to be a fork](https://github.com/ianstormtaylor/slate/pull/3704#issuecomment-638835101) of Slate where this is already implemented, just for future reference 😊

PR #3792, opened 2020-07-19
Fixes placeholder type and adds docs on usage

mwood23:

Improving a feature

#### What's the new behavior?

Corrects the types of the placeholder prop in Slate-React and adds usage docs.

The placeholder takes a React Node and the types originally said string. People have been looking at how to create a custom placeholder on Slack so I PR'ed to fix the types and add to the docs.

_Italicized placeholder_

Fixes: https://github.com/ianstormtaylor/slate/issues/3781
Reviewers: @timbuckley


site/examples/richtext.js
```diff
@@ -37,6 +37,7 @@ const RichTextExample = () => {
       <Editable
         renderElement={renderElement}
         renderLeaf={renderLeaf}
+        // See slate-react docs for how to style a custom placeholder.
```
timbuckley, line 40: I don't think this comment will compile because it's amongst JSX, I think it's okay to remove.

```suggestion
```

mwood23: These comments are valid if you're commenting within a component block and are invalid if you are commenting in the render block of JSX (need the fancier comments there). I'll go and remove to keep consistency with other examples.

https://codesandbox.io/s/awesome-feistel-1xuiz?file=/src/App.js

BrentFarese:
@mwood23 can you check this against #3725, which adds some types to `slate-react` for the placeholder using a new custom-types paradigm?  Are the types correct given your work?  If not, we can adjust...

mwood23:
@BrentFarese It does relate to it. Based on that PR it looks like `leaf.placeholder` is getting typed as a string (what my PR changes to `ReactNode`). I think this is a decision point about how Slate wants to type the placeholder prop. Technically, it can accept a `ReactNode` (`string | number | ReactElement | ReactFragment | ReactPortal | boolean | null | undefined`). However, maybe you want to limit it to just a `string` because the placeholder element may change a lot over time.

As it stands, there are no other ways to target the placeholder for styling. If there's not going to be many changes to the placeholder element I think it makes sense to make it a `ReactNode`, and if left as a string introduce a way to target it for styling (https://github.com/ianstormtaylor/slate/issues/3780). 

Since I'm not too familiar with future of the project I'm not sure what direction to recommend. I put in the PR because in the code is was being cast as a `ReactNode` and by that I assumed the intent was leave flexibility to the caller.

Issue #3791, opened 2020-07-17
Editor.next and Editor.previous return incorrect matches


BrentFarese:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

`Editor.next` and `Editor.previous` return the **2nd match**.

Slate: 0.58.4
Browser: Chrome
OS: Mac

#### What's the expected behavior?

`Editor.next` and `Editor.previous` should return the immediate next or previous node specified by the `match` after the relevant `Location`.  Instead, both functions return **the 2nd match**, which is counterintuitive and leads to bugs.

The behavior comes from [this line in Editor.next](https://github.com/ianstormtaylor/slate/blob/93fe25151722343488e9002a0ebd8ed3ba66ee95/packages/slate/src/interfaces/editor.ts#L590) and [this line in Editor.previous](https://github.com/ianstormtaylor/slate/blob/93fe25151722343488e9002a0ebd8ed3ba66ee95/packages/slate/src/interfaces/editor.ts#L1123) where you can see that the **2nd match** is destructured from the `Editor.nodes` call vs. the first match.

**Proposed fix:**  Simply destructure the first match - e.g., `const [next] = Editor.nodes(editor, { at: span, match, mode, voids })` and make sure all related tests pass/are fixed.

BrentFarese:
Or, an alternative behavior could be to find the `Point` **_after_** the `Location` and then return the next match immediately after that `Location`.  The API documentation says that the function returns the next match **_after_** a location, but instead the API returns the **_2nd matching node_** after a location.  Those are not the same things.

We might use `Editor.after` to find the `Point` after the location, and then search for the _**first match**_ from that point forward?

BrentFarese:
Fixed by #3957.

PR #3795, opened 2020-07-21
chore: Destructure first first match

kodekage:
Fixing a bug

#### What's the new behavior?
Destructure the first match when using `Editor.next` and `Editor.previous`

Fixes: #3791
Reviewers: @BrentFarese

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -587,7 +587,7 @@ export const Editor = {
       }
     }
 
-    const [, next] = Editor.nodes(editor, { at: span, match, mode, voids })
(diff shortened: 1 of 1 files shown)


awasum:
Why is this failing?

AndrewScull:
> Why is this failing?

This is a change in behaviour that has caused tests to fail. Looking at the CI logs, one such failure comes from a [test of Editor.next()](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/test/interfaces/Editor/next/block.tsx) that expects the previous behaviour and could suggest that other users of this function will also expect the previous behaviour, strange or confusing as it may be.

BrentFarese:
Going to close this as it is an incomplete implementation that needs to be thought out a bit more.  Will re-open if it gets picked up.  Thanks!

PR #3437, opened 2020-01-18
Fix element memoization checks. Allows placeholder to change dynamically based on the props. Allows renderLeaf to be memoized.

ryanmitts:
There are bugs with the memoization checks in the element, text, and leaf components.

The `renderLeaf` function to the Editable cannot be memoized as if it is, the rendering of leaves are not reconciled correctly on the DOM as the old node is allowed to stay despite the leaf being different. This is because the leaf element does not actually check it's text content in the memoization check.

The placeholder decorator is not re-rendered when the placeholder text is changed, this is because the element, leaf and text components do not check the decorator properties.

#### Is this adding or improving a _feature_ or fixing a _bug_?

Bug.

#### What's the new behavior?

The appropriate components check in the memoization comparison if two decorator range list items are placeholders, and if they are, if their text equals.

Demo

You will also be able to memoize the `renderLeaf` function when using decorators.

#### How does this change work?

Checking the decorator props, and the leaf checking it's text is the same in the memoization checks.

#### Have you checked that...?

- [x] The new code matches the existing patterns and styles.
- [x] The tests pass with `yarn test`.
- [x] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [x] The relevant examples still work. (Run examples with `yarn start`.)

#### Does this fix any issues or need any specific reviewers?

Fixes: #3423, #3432, #3447
Reviewers: @

PR #3799, opened 2020-07-24
Collapse selection according to reverse

aiwenar:

Unsure.

#### What's the new behavior?

`Transforms.delete` now collapses the selection forward if `options.reverse` is false and backward if `options.reverse` is true. Previously selection was always collapsed forward.

While this doesn't change behaviour much in normal operation (the behaviour is only different on the forward edge of inlines), it is significant if `remove_text` operations are prevented. In our use case we're building a suggestions mode (think Google Doc's suggesting or Word's tracking changes), in which when user deletes content it is visually marked and recorded as deleted, but not removed from the document itself. On master this causes backspace to behave unintuitively: 

- when selection is collapsed and backspace is pressed, the character before cursor is marked as deleted but the cursor doesn't move.

  remove-collapsed-before-2

  This is not very visible on this GIF, as for some reason new decorations are not rendered until cursor is moved, but note that deletion appears on ‘h’ only when cursor is moved away from it. You can also experiment with this in [this fiddle](https://jsfiddle.net/g62s5jbm/) (note that our plugin (used in GIF) is based on decorations, but fiddle uses text properties instead; this is because I couldn't get decorations to work correctly at all without #3437).

- when selection is expanded and backspace is pressed, selected text is removed but selection is collapsed at the end of the selection, rather that at the front

  remove-expanded-before

With this change delete button's behaviour is unchanged, but backspace now behaves correctly:

remove-collapsed-after

remove-expanded-after

- when collapsing `Transforms.delete` now prefers `startRef` over `endRef` if `options.reverse` is set to true;

- `editor.deleteFragment` and `Editor.deleteFragment` now accept an additional argument, `direction?: 'forward' | 'backward'`, specifying direction in which selection should be collapsed. This argument is optional and when omitted behaviour remains the same as it was before this PR.

Fixes: no
Reviewers:


ianstormtaylor:
Absolutely beautiful explanation and PR, thank you @aiwenar!

PR #3804, opened 2020-07-28
fix inconsistent render reversed selection

githoniel:

This PR would fix a bug.

According to MDN [Range/setStart](https://developer.mozilla.org/en-US/docs/Web/API/Range/setStart), there is no way to create a reverse selection. So `slate` will render a 'no-reverse' dom-selection when `editor.selection` is reverse.

It is broken when user re-render the component when selection is change. Here is an example:

```jsx
export default function Demo() {
  const [currentSelection, setCurrentSelection] = useState('')

  function onChange() {
    const { selection } = editor
    if (selection) {
      setCurrentSelection(JSON.stringify(currentSelection))
    }
  }

  return (
    <Slate editor={editor} value={value} onChange={onChange}>
      <Editable />
    </Slate>
  )
}
```

`Demo` will re-render when selection change, and then `Demo` will re-render `Editable`。

If selection is not reversed(eg selection offset is [X, Y], Y > X),  Editable will rendered dom-selection [X, Y] with `editor.selection` offset like [X, Y], and will return and do nothing to update DomSelection because of `Range.equals`.

If selection is reversed(eg selection offset is [X, Y], Y < X), Editable will rendered dom-selection [Y, X]  with `editor.selection` offset like [X, Y], and then found not `Range.equals` with `editor.selection`. It will try to re-render Editable while user is using mouse selecting text, result to anchor lost.

In the gif, I'm selecting text reverse using code like `Demo` above

xx

#### What's the new behavior?

make slate render reverse dom-selection when `editor.selection` is reverse to avoid unnessasery re-render due to  inconsistent render

use code from [stackoverflow](https://stackoverflow.com/a/4802994)

There is no test code for `slate-react` now.

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/utils/dom.ts
+++ b/packages/slate-react/src/utils/dom.ts
@@ -182,5 +182,5 @@ export const setReverseDomSelection = (
   const newRange = domRange.cloneRange()
   newRange.collapse()
   domSelection.addRange(newRange)
-  domSelection.extend(newRange.startContainer, newRange.startOffset)
+  domSelection.extend(domRange.startContainer, domRange.startOffset)
 }
--- a/packages/slate-react/src/utils/dom.ts
+++ b/packages/slate-react/src/utils/dom.ts
@@ -175,6 +175,14 @@ export const getPlainText = (domNode: DOMNode) => {
   return text
 }
 
+/**
+ * there is no way to create a reverse DOM Range using Range.setStart/setEnd
+ * according to https://dom.spec.whatwg.org/#concept-range-bp-set.
+ * Luckily it's possible to create a reverse selection via Selection.extend
+ *
+ * Note: Selection.extend is not implement in any version of IE (up to and including version 11)
+ */
+
 export const setReverseDomSelection = (
   domRange: DOMRange,
   domSelection: Selection
--- a/packages/slate-react/src/utils/dom.ts
+++ b/packages/slate-react/src/utils/dom.ts
@@ -174,3 +174,13 @@ export const getPlainText = (domNode: DOMNode) => {
 
   return text
 }
+
+export const setReverseDomSelection = (
+  domRange: DOMRange,
+  domSelection: Selection
+) => {
+  const newRange = domRange.cloneRange()
+  newRange.collapse()
+  domSelection.addRange(newRange)
+  domSelection.extend(newRange.startContainer, newRange.startOffset)
+}
--- a/packages/slate-react/src/utils/dom.ts
+++ b/packages/slate-react/src/utils/dom.ts
@@ -188,7 +188,10 @@ export const setReverseDomSelection = (
   domSelection: Selection
 ) => {
   const newRange = domRange.cloneRange()
+  // collapses the range to end
   newRange.collapse()
+  // set both anchor and focus of the selection to domRange's focus
   domSelection.addRange(newRange)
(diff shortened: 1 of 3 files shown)


packages/slate-react/src/utils/dom.ts
```diff
@@ -174,3 +174,13 @@ export const getPlainText = (domNode: DOMNode) => {
...
 }
+
+export const setReverseDomSelection = (
+  domRange: DOMRange,
+  domSelection: Selection
+) => {
+  const newRange = domRange.cloneRange()
+  newRange.collapse()
+  domSelection.addRange(newRange)
+  domSelection.extend(domRange.startContainer, domRange.startOffset)
```
TheSpyder, line 185: Hmm it would be better if this was all done in one place. Ideally `toDOMRange` could do this itself, rather than needing to patch what it returns, although that would require having it modify the global selection 🤔

I can't recall whether storing a backwards selection range and then re-adding it via `addRange` works, though, which would make that impossible.

githoniel: I found no way to create a reverse range according to https://dom.spec.whatwg.org/#concept-range-bp-set.

use `Selection.extend` is only way I found to make it work

TheSpyder: but if you make a backwards selection through `extend`, and then store the range to later set via `domSelection.addRange()`, would that work?

It's a bit of a round-about way to do it, but I'm suggesting we avoid splitting the set selection logic (any new code expecting `toDOMRange` to return a valid selection might also trigger a bug with backwards selections).

githoniel: https://w3c.github.io/selection-api/#definition

there is only one selection in document as above document. make a backwards selection through `Selection.extend` will change the selection of the document and will trigger onSelection change event.

I understand what your concern but it is dom API limit us.

CameronAckermanSEL: This seems like a useful concern to capture here for posterity. Could you add a note here to why we need to clone and mutate the range in this way?

CameronAckermanSEL:
Thank you @githoniel! Really appreciate you taking the time to fix this pretty irritating bug! 😆  

I had a request for a comment just because this area is a bit brittle and there's a lot of valuable domain knowledge you gathered that we could use. Other than that, ready to go and I'm excited to release it!

packages/slate-react/src/utils/dom.ts
CameronAckermanSEL: @TheSpyder  does leaving it as a comment work for you? We might be able to do something to improve it at a later time if it 💥  but this seems to work well

TheSpyder: I realise there's only one selection in a document, I was talking about storing the current selection, mutating it to the range desired, then storing the range and restoring the original selection. It's probably a bad idea but would keep everything contained.

I'd like to see a similar comment on `toDOMRange`, as that's where people are likely to look first for selection-related things rather than the `dom` util module.

githoniel: thank for preview and i have add comments for `setReverseDomSelection` and `toDOMRange`

CameronAckermanSEL:
@githoniel  thank you!

skogsmaskin:
@githoniel - thank you for this! This has been a pain point!

PR #3807, opened 2020-07-31
do not change selection while delete fragment

githoniel:

fixing a bug

#### What's the new behavior?

We will skip change selection while delete fragment as comment.

It will broken history merge while drag & drop content in slate

see below example

xx

original schema is

```
├─ 0
│  └─ children
│     └─ 0
│        └─ text: dragTest1
└─ 1
   └─ children
      └─ 0
         └─ text: dragTest2
```

1. select dragTest2
2. drop it to the end of dragTest2
3. new shema is
```
├─ 0
│  └─ children
│     └─ 0
│        └─ text: dragTest1dragTest2
└─ 1
   └─ children
      └─ 0
         └─ text: 
```
4. press `ctrl+z` to undo
5. new shema is
```
├─ 0
│  └─ children
│     └─ 0
│        └─ text: dragTest1
└─ 1
   └─ children
      └─ 0
         └─ text: 
```

We lost `dragTest2`

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -248,7 +248,7 @@ export const Editable = (props: EditableProps) => {
         // COMPAT: For the deleting forward/backward input types we don't want
         // to change the selection because it is the range that will be deleted,
         // and those commands determine that for themselves.
-        if (!type.startsWith('delete') || !type.startsWith('deleteBy')) {
+        if (!(type.startsWith('delete') || type.startsWith('deleteBy'))) {
           const [targetRange] = event.getTargetRanges()
 
           if (targetRange) {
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -248,7 +248,7 @@ export const Editable = (props: EditableProps) => {
         // COMPAT: For the deleting forward/backward input types we don't want
         // to change the selection because it is the range that will be deleted,
         // and those commands determine that for themselves.
-        if (!type.startsWith('delete') || type.startsWith('deleteBy')) {
+        if (!(type.startsWith('delete') || type.startsWith('deleteBy'))) {
```
AndrewScull, line 251: Since 'deleteBy' starts with 'delete' this condition would really only need to be `if (!type.startsWith('delete'))` i.e. removing the the bit that was selecting the `deleteFragment()` cases.

AndrewScull (commented):
I'm not sure the exact bug this is trying to fix. A detailed explanation or example to reproduce it would be helpful. I've made some suggestions assuming this is the correct fix.

githoniel:
thank you for review, sorry that I have forgot what this fix, but it did fix a bug in my slatejs based editor. 

I would add case later.

githoniel:
sorry @AndrewScull I found this pr is wrong, I will open a new one

PR #3809, opened 2020-07-31
Fixed html

mdmjg:

bug

#### What's the new behavior?

Underline now works in paste-html example

Code used to use "underline" and "underlined", which meant that the underline would not work.

- [x ] The new code matches the existing patterns and styles.

Fixes: #
Reviewers: @timbuckley @BrentFarese

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "workspaces": [
     "packages/*"
   ],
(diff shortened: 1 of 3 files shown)


CameronAckermanSEL:
Thanks so much!

PR #3820, opened 2020-08-11
fix unwrapNode when split && add in nested block

githoniel:

bug fix

#### What's the new behavior?

we will use this transform

```js
  Transforms.unwrapNodes(editor, {
    match: n => !!n.a,
    mode: 'all',
    split: true,
  })
```

schema before transform

```jsx
 <editor>
    <block a>
      <block a>
        <block>one</block>
        <block>
          <cursor />
          word
        </block>
        <block>now</block>
      </block>
    </block>
  </editor>
```

schema after transform

```jsx
 <editor>
    <block a>
      <block>
        <block>one</block>
      </block>
    </block>
    <block>
      <cursor />
      word
    </block>
    <block a>
      <block>
        <block>now</block>
      </block>
    </block>
  </editor>
```

we unwrap blocks contain `one` and `now` which are out of the selection.

`unwrapNode` will call `liftNode` which does not support `split` mode when nested.

if we apply `liftNode` transform like below in above example

```
  at: range, // [0, 0, 1, 0] to [0, 0, 1, 0]
  match: n => node.children.includes(n), // node' path is [0]
  voids,
})
```
we will lift all children node of  [0, 0]/

so `unwrapNode` should not call `unwrapNode` from root to deeper but from deeper to root

PS: or maybe we should fix in `liftNode`?

Fixes: #
Reviewers: @


packages/slate/src/transforms/node.ts
```diff
@@ -716,7 +716,9 @@ export const NodeTransforms = {
 
       const rangeRef = Range.isRange(at) ? Editor.rangeRef(editor, at) : null
       const matches = Editor.nodes(editor, { at, match, mode, voids })
-      const pathRefs = Array.from(matches, ([, p]) => Editor.pathRef(editor, p))
+      const pathRefs = Array.from(matches, ([, p]) =>
+        Editor.pathRef(editor, p)
+      ).reverse()
```
dylans, line 721: I'm not quite following this change? Maybe you want to make a call to range.isBackward and reverse if necessary ?

dylans (changes requested):

packages/slate/src/transforms/node.ts
githoniel: thanks for preview. this issue is not related to range direction.

`unwrapNode` will call `liftNode` which does not support split mode when nested. so if we not reverse the order and call it from top to the bottom, it will remove all  blocks that wrap target node. so we reverse the order

dylans: Thanks, that makes sense.

dylans, line 721:
```suggestion
      // unwrapNode will call liftNode which does not support splitting the mode when nested.
      // If we do not reverse the order and call it from top to the bottom, it will remove all blocks 
      // that wrap target node. So we reverse the order.
      ).reverse()
```

I think a comment helps here. I'm happy to be told this is unnecessary. :)

dylans:
@githoniel could you please run prettier locally to cleanup the PR, or give me edit access for the branch? My suggestion was probably too long. Also, if you have time, adding a changeset would be awesome. If not, I can add it. And then I think this one is ready to go.

githoniel:
I will do this soon

githoniel:
I think the deloy fail is not cause by this pr

packages/slate/src/transforms/node.ts
```diff
@@ -844,7 +844,13 @@ export const NodeTransforms: NodeTransforms = {
 
       const rangeRef = Range.isRange(at) ? Editor.rangeRef(editor, at) : null
       const matches = Editor.nodes(editor, { at, match, mode, voids })
-      const pathRefs = Array.from(matches, ([, p]) => Editor.pathRef(editor, p))
+      const pathRefs = Array.from(
+        matches,
+        ([, p]) => Editor.pathRef(editor, p)
+        // unwrapNode will call liftNode which does not support splitting the mode when nested.
```
dylans, line 850:
```suggestion
        // unwrapNode will call liftNode which does not support splitting the node when nested.
```

dylans (commented):
looks good, one small typo.

I will land this after we do a patch release today.

dylans (approved):

PR #3825, opened 2020-08-14
fix: mergeNodes by point would delete nested multi-child node

githoniel:

fix bug

#### What's the new behavior?

mergeNodes will remove emptyNode which has only children node

But if emptyNode has nested multi-children node, it will also be remove.

recursive to confirm the node has only one children node

Fixes: #
Reviewers: @


ianstormtaylor:
Thanks @githoniel!

PR #3831, opened 2020-08-17
fix: move_node path transformation when op and onp is sibling

akumatus:

fix a bug

#### What's the new behavior?

I'm using slate.js for my real-time collaborative editor and focusing on the operational transformation recently. I found out that `Path.transform` is useful, elegant and heuristic. But some logics of `move_node` path transform functions made me confused at the begining, and some calculations of them might be wrong.

For example: when `op` is sibling with `onp`, `onp` equals `p`, and `op` ends before `p`, the result of transformed path should be exactly the same as `p`.

image

Let me draw a picture to show why the result is `p`.

image

It is obvious that the transformed path of `move_node` in the aboved case is `[0, 2]`.

Here is another code snippet I found that may not be right. When `op` equals `p`, `op` ends before `onp`, and `op.length` equals `onp.length`. The result of transformed path should also do the logic of `copy[op.length - 1] -= 1`.

image

Let's take a look at the real world example again.

image

It is obvious that the transformed path of `move_node` in the aboved case is `[0, 2]`.

Then I  sort out the logics of `move_node` path transform and try to make it more beautiful. In my opinion, `move_node` is a bit like the composition of `remove_node` and `insert_node`. So the relationship of `move_node` paths (p, op, and onp)  can be divided into two parts. This idea instructs my changes to the `move_node` path tranform logic.

Moreover, I add this logic below in the code as a `quick return`:

image

It will leads to the same transformed path result without this `quick return` code snnipet. 

But this `quick return` should make sense since the `path` and `Path.next(path)` actually means the same position in the node movement. I draw another picture below to demonstrate this.

image

Let me know if I have some misunderstanding. Discussion is welcome.

Fixes: #
Reviewers: @ianstormtaylor @CameronAckermanSEL @jason-krypton

### DIFF
--- a/packages/slate/src/interfaces/path.ts
+++ b/packages/slate/src/interfaces/path.ts
@@ -377,28 +377,19 @@ export const Path = {
         case 'move_node': {
           const { path: op, newPath: onp } = operation
 
-          // If the old and new path are the same, it's a no-op.
-          if (Path.equals(op, onp)) {
+          // If the old and new path are the same or next sibling, it's a no-op.
+          if (Path.equals(op, onp) || Path.equals(Path.next(op), onp)) {
             return
           }
 
           if (Path.isAncestor(op, p) || Path.equals(op, p)) {
             const copy = onp.slice()
 
-            if (Path.endsBefore(op, onp) && op.length < onp.length) {
+            if (Path.endsBefore(op, onp)) {
               copy[op.length - 1] -= 1
             }
 
             return copy.concat(p.slice(op.length))
-          } else if (
-            Path.isSibling(op, onp) &&
-            (Path.isAncestor(onp, p) || Path.equals(onp, p))
-          ) {
-            if (Path.endsBefore(op, p)) {
-              p[op.length - 1] -= 1
-            } else {
-              p[op.length - 1] += 1
-            }
           } else if (
             Path.endsBefore(onp, p) ||
             Path.equals(onp, p) ||
@@ -410,7 +401,11 @@ export const Path = {
 
             p[onp.length - 1] += 1
           } else if (Path.endsBefore(op, p)) {
-            if (Path.equals(onp, p)) {
+            if (
+              Path.endsBefore(onp, p) ||
+              Path.equals(onp, p) ||
+              Path.isAncestor(onp, p)
+            ) {
               p[onp.length - 1] += 1
             }
 
--- a/packages/slate/test/interfaces/Path/transform/move_node/ancestor-sibling-ends-before-to-ancestor.tsx
+++ b/packages/slate/test/interfaces/Path/transform/move_node/ancestor-sibling-ends-before-to-ancestor.tsx
@@ -9,4 +9,4 @@ const op = {
 export const test = () => {
   return Path.transform(path, op)
 }
-export const output = [2, 3, 3]
+export const output = [3, 3, 3]
--- a/packages/slate/test/interfaces/Path/transform/move_node/ends-before-to-ancestor.tsx
+++ b/packages/slate/test/interfaces/Path/transform/move_node/ends-before-to-ancestor.tsx
@@ -1,12 +1,12 @@
 import { Path } from 'slate'
 
-const path = [0, 1]
+const path = [3, 3, 3]
 const op = {
   type: 'move_node',
-  path: [0, 0],
-  newPath: [0, 1],
+  path: [3, 2],
+  newPath: [3],
 }
 export const test = () => {
   return Path.transform(path, op)
 }
-export const output = [0, 1]
+export const output = [4, 2, 3]
--- a/packages/slate/test/interfaces/Path/transform/move_node/ends-before-to-ends-before.tsx
+++ b/packages/slate/test/interfaces/Path/transform/move_node/ends-before-to-ends-before.tsx
@@ -1,12 +1,12 @@
 import { Path } from 'slate'
 
-const path = [0, 1]
+const path = [3, 3, 3]
 const op = {
   type: 'move_node',
-  path: [0, 0],
-  newPath: [0, 1],
+  path: [3, 2],
+  newPath: [2],
 }
 export const test = () => {
   return Path.transform(path, op)
 }
-export const output = [0, 1]
+export const output = [4, 2, 3]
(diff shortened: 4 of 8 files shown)


packages/slate/test/interfaces/Path/transform/move_node/sibling-ends-before-to-ends-equal.tsx
```diff
@@ -9,4 +9,4 @@ const op = {
 export const test = () => {
   return Path.transform(path, op)
 }
-export const output = [0, 0]
+export const output = [0, 1]
```
jason-krypton, line 12: This is wrong. When you move 0 -> 1, Removal of the 0 node causes the 1 to shift forward and take the 0th place. Then what was 0 gets inserted at the newPath which is 1

Example:
Start: [A, B]
Remove 0: [B]
Insert A at 1: [B, A]

B's path has changed to [0, 0]

Suggest you look here: https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/transforms/general.ts#L91

jason-krypton (commented):
The move logic is a bit complicated, but I'm pretty sure the logic is correct as it stands. Can you write a unit test that shows it not working as expected?

In the first example you give in the PR description highlights the misunderstanding. onp is not before p it is === to p there is no concept of in between the nodes. 

I'd say the other thing is that the Transform code should be consistent with the op application path, which I currently believe it is.

packages/slate/test/interfaces/Path/transform/move_node/sibling-ends-before-to-ends-equal.tsx
akumatus: I think what you said is not exactly correct.

The `newPath` [0,1] is defined in the context before `move_node` operation. That means, you can not just insert node at path [0, 1] after remove node at path [0, 0], because the context has already changed. You need to transform the `newPath` [0,1] against `remove_node` operation, use the transformed `newPath` as `insert_node` path.

Example:
Start: [A, B]
Remove 0: [B]
Transform newPath: 1 -> 0
Insert A at 0: [A, B]

The key point is to understand the difference between operation definition context and execution context. Here are some information you can refer to.
[https://www3.ntu.edu.sg/scse/staff/czsun/projects/otfaq/#_Toc321146146](https://www3.ntu.edu.sg/scse/staff/czsun/projects/otfaq/#_Toc321146146)

ianstormtaylor:
@TheSpyder could you take a look at this? I can't remember whether `move_node`'s `newPath` is before or after the change, and I figure you've been dealing with this more recently and would know whether this is correct right away. Thank you!

ulion:
I recently also encounter some move_node treat newPath not as expected (before the move_node op context) cases, they are specifically the case of newPath is exactly same or next siblings (same level) cases. According to Path.transform, cases of newPath is descendants of next siblings are correctly handled, but the same level cases (same path or next siblings) are not handled the same way (for the correct cases, newPath is exactly where you currently can see the same position before move_node op, for the incorrect cases, newPath is exactly where you need first remove_node, then you can figure out where it is targeting).
```
if (Path.isAncestor(op, p) || Path.equals(op, p)) {
  const copy = onp.slice()

  if (Path.endsBefore(op, onp) && op.length < onp.length) {
    copy[op.length - 1] -= 1
  }

  return copy.concat(p.slice(op.length))
}
```
So, the move_node op's newPath is not identical logically defined for two cases:
If we have [0, 1, 2] 3 nodes, we want to move '0' to where 2 currently is (index 2), result is [1, 0, 2], in current implement, we have to do move_node (0 -> 1) with newPath [1] not where it currently looks like [2] (0 -> 2). which means for this next sibling case, you have to first do remove_node, then you can figure out where the target index is.
While if we want to move '0' under node '2', e.g. path [2, 1], then we do not do the shift on index 2, newPath is just [2, 1](where it currently looks should be) not [1, 1] (where after remove_node it is). which means for these cases, you do not do remove_node first, the newPath is exactly it looks currently will goes to. if you do remove_node first, then the newPath 'looks' incorrect.
In above 2 cases, newPath is either where it looks like, or not (require a remove_node first to match)
This behaviour maybe works with itself (all places do the same logic and generate newPath carefully), but it is not clearly defined. I thought newPath is where it currently (before move_node op) looks should be, but in the sibling case, they are not. The logic is not identical, and it caused some of my code have to be re-considered to fit the strange behaviour/definition of the move_node op.

TheSpyder:
Sorry I haven't had a chance to look into this. I was on leave 3 weeks ago and have been busy since then. I'll try to make time soon.

Off the top of my head, I can see the case for newPath to be the context before the move. But [the implementation uses after](https://github.com/ianstormtaylor/slate/blob/fd70dc0b2c0d06edb9490874fb831161b9759cba/packages/slate/src/transforms/general.ts#L104-L109), as @jason-codaio points out.

I'll need to check my notes but I had some problems with `move_node` in my OT implementation that lead me to invert one user's changes in complex conflicting scenarios rather than try to merge the results. Switching to `newPath` as the before context might help me avoid doing that.

If we wanted to make it before perhaps we could splice the node into its new index (meaning the node is in the document twice) and then remove it from the original path (using transform to find it). In this case _both_ `Path.transform` and `editor.apply` would need to be updated. Unfortunately this is likely a significant breaking change for an existing OT implementation based on the current behaviour.

I know we're not shy about breaking changes, but I want to be sure the change is worth it.

ulion:
yes, it is a breaking change. which will affect existed code which have already fitted the current newPath rule.
In my code (modified slate-yjs), I try to restore move_node/split_node/merge_node op from insert_node/remove_node/insert_text/remove_text combinations, because move/split/merge ops will avoid unnecessary lost of current caret (selection) position during those changes.
I write code like getPathBeforeOp and getPathAfterOp to get the original or future path before or after certain type of op. Then I found move_node does not like newPath be the one before the remove_node op for certain cases, which is not expected, but it can be understood. Now I know it is only for next sibling cases, so I wrote the other method to fix move_node ops I generated from the path before remove_node to fit this limited rule. Now my code works.

dylans:
I look at this PR somewhat regularly and think it would be cool to revisit if there's interest.

But if there's not, we should probably close it as it hasn't received attention for 2+ years.

@akumatus please let us know your current interest.

dylans:
Closing for now, feel free to reopen if you would like to revisit in the future.

Issue #3656, opened 2020-05-05
Avoid unnecessary re rendering of leaves within a block element


rafael-castelo:
#### Do you want to request a _feature_ or report a _bug_?
_bug_

#### What's the current behavior?
Hi, I'm using slate and been struggling with the rendering of leaves. Here is my use case, I need to have many leaves/text objects (with different properties) inside a block element, the thing is when a change is made to the editor value, all leaves in the block element are re-rendered, that's causing us an important performance issue. I'd like to know if there is a way that it could be hooked so it only re renders the current text object edited or if this behavior can be changed someway.

here is the code SandBox: https://codesandbox.io/s/slate-reproductions-q1nv7

As you can see, when a change is made, all leaves are re-rendered (look at the console.log in the renderLeaf function). With short paragraphs it is not a problem, but if you try with the largeEditorValue variable you will see how it laggs (editado)

Slate: 0.XX.X
Browser: Chrome / Safari / Firefox / Edge
OS: Mac / Windows / Linux / iOS / Android

#### What's the expected behavior?
I would like to be able to only render the current edited leaf, instead of rerendering all the children leaves of the element.

CameronAckermanSEL:
Hi @rafael-castelo , I think it's reasonable to expect slate to only rerender the leaves that changed and I think a PR that makes that change is likely to get merged. 

That said, I'm getting the sense that this particular implementation is avoidable and instead you should solve your use case in a different than having a render function for every single leaf that has runtime processing for styling.

What are you actually trying to accomplish?

jerkly:
@rafael-castelo, check this issue #3507 and pull with example #3515. You can find a start point to do it with controlled value.

rafael-castelo:
> 
> 
> Hi @rafael-castelo , I think it's reasonable to expect slate to only rerender the leaves that changed and I think a PR that makes that change is likely to get merged.
> 
> That said, I'm getting the sense that this particular implementation is avoidable and instead you should solve your use case in a different way than having a render function for every single leaf that has runtime processing for styling.
> 
> What are you actually trying to accomplish?

Glad to hear that a new PR which manages that is coming, by the way, I'm still loking for a solution meantime.
This is what i'm trying to acomplish: I'm building an app which shows the spoken words of a video or audio in the editor, allowing the user to edit/correct it. I need the start and end of each word(text node), to higlight them in sync with video, and also the confidence of each word, to colour words differently depending on the confidence. 
[...]

Issue #3740, opened 2020-06-16
Transform.moveNodes within nested container causes ALL children in container to be re-rendered


davidruisinger:
#### Do you want to request a _feature_ or report a _bug_?

_bug_

#### What's the current behavior?
Assuming you have a slate state like this:

```
[
  {
    type: "body",
    children: [
      {
        type: "video",
        children: [
          {
            text: "https://www.youtube.com/embed/22m5eU6uxeQ"
          }
        ]
      },
      {
        type: "paragraph",
        children: [
          {
            text:
              'Toggle this line to "bullet-list" => All nodes within "body" are re-rendered, causing video above to flicker.'
          }
        ]
      }
    ]
  }
]
```

And you use `Transforms.wrapNodes` (which internally calls `Transforms.moveNodes`) on the `paragraph` underneath the `video`, the `video` (and all other potentially existing nodes within the `body`) gets re-rendered even though it has not been touched.

[Codesandbox](https://codesandbox.io/s/prevent-re-render-of-children-gslvo?file=/src/slate-editor.tsx)

![P6NLJSkYcq](https://user-images.githubusercontent.com/5305150/84755247-b47f1e00-afc1-11ea-8cf7-b3728275b093.gif)

Slate: 0.58.3
Browser: Chrome
OS: Mac

#### What's the expected behavior?

Slate should only re-render the parts that actually changed so the video does not re-render (flicker).

flavordaaave:
My use-case where this behavior is really causing problem:

- Inlinebar is positioned relative to selection
- Social embeds are rendered
- Formatting is applied from Inlinebar (e.g. toggle `bullet-list`)
- Social embeds are re-rendered causing Inlinebar to be incorrectly placed

See this [Codesandbox](https://codesandbox.io/s/prevent-re-render-of-children-with-inlinebar-8go9z?file=/src/slate-editor.tsx):

![HyR24ybKVs](https://user-images.githubusercontent.com/5305150/84761238-74239e00-afc9-11ea-9eca-02a4fc3ec50a.gif)

TheSpyder:
Not that this shouldn't be fixed, but I thought most inline bars like that were absolutely positioned. Not actually embedded into the content.

flavordaaave:
> Not that this shouldn't be fixed, but I thought most inline bars like that were absolutely positioned. Not actually embedded into the content.

@TheSpyder The inlinebar IS actually absolutely positioned but relative to the current selection. 
[...]

PR #3832, opened 2020-08-19
fix node-to-key weakmap refresh 

githoniel:

fix a bug 

#### What's the new behavior?

We will update node-to-key weakMap when apply transform but it does not work well now.

1. revert transform do not update node-to-key weakmap
2. weakmap update before react batchUpdate
3. related element is not correct

RT see code

Notice: slate-react do not test code 

Fixes: #3656 #3740 partial fix
Reviewers: @

### DIFF
--- a/packages/slate-history/src/with-history.ts
+++ b/packages/slate-history/src/with-history.ts
@@ -52,7 +52,7 @@ export const withHistory = <T extends Editor>(editor: T) => {
             ) {
               continue
             } else {
-              e.apply(op)
+              editor.apply(op)
             }
           }
         })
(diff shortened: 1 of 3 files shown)


dylans (changes requested):
Any interest in updating this against main @githoniel and then I'll review and land it?

dylans (changes requested):
@githoniel is this PR still relevant? Let me know if you want to clean it up or if we should close it out.

githoniel:
I would close this because this is not implemented at that moment,  but now this has been implemented

PR #3845, opened 2020-08-27
Bulk upgrade core dependencies

jason-codaio:

Upgrading dependencies. Required minor test update.

#### What's the new behavior?

Hopefully none.

Reviewers: @timbuckley


site/pages/examples/[example].tsx
```diff
@@ -256,6 +267,7 @@ const ExamplePage = ({ example }: { example: string }) => {
   const [name, Component, path] = EXAMPLE
   return (
     <ErrorBoundary
+      FallbackComponent={ErrorFallback}
```
Comment (timbuckley) [inline site/pages/examples/[example].tsx:270]: Did you get a chance to manually test the result of an error here?

Comment (jason-krypton) [reply site/pages/examples/[example].tsx:270]: Honestly no, but I just copy pasted this off the package Readme

Comment (jason-krypton) [reply site/pages/examples/[example].tsx:270]: My assumption was this is not hit and I'd have to break one of the examples to hit this. Also given its test/examples code I was less concerned

jason-krypton:
Is this good to go? @timbuckley

PR #3846, opened 2020-08-27
Upgrade prettier

jason-codaio:

Update to latest prettier. Also updates code with `yarn lint --fix` to match new package.

#### What's the new behavior?

More idiomatic code style

Reviewers: @timbuckley


timbuckley:
How about just adding `"arrowParens": "avoid"` to `.prettierrc`?

jason-krypton:
@timbuckley I suppose thats up to you. I can update if that is the preference.

jason-krypton:
Updated with allowParens avoid

site/examples/markdown-preview.tsx
```diff
@@ -70,38 +70,38 @@ const Leaf = ({ attributes, children, leaf }) => {
...
+          font-size: 20px;
+          margin: 20px 0 10px 0;
+        `}
         ${leaf.list &&
-          css`
-            padding-left: 10px;
-            font-size: 20px;
-            line-height: 10px;
-          `}
+        css`
```
timbuckley, line 80: This indentation is a little odd, is this all new-prettier's doing? One would normally think the second operand of a conjunction would be indented relative to the previous if it must be on a new line...
What if you try to put `css` immediately after the `&&` like so?
```
${leaf.title && css`
```
Does that change the indentation a bit? I

jason-krypton: All code changes are from running `yarn lint --fix`. I can try manually changing the line manually and re-running, but will be a bit before I update.

jason-krypton:
I tried to manually change back the whitespace change around css that you called out @timbuckley but then lint will complain.

PR #3847, opened 2020-08-27
Add resolutions to override security alerts

jason-codaio:

Sadly there are a number of issues remaining which can not be simply fixed by updating a package as they are deeply nested and require new releases of webpack and other packages that move slowly.

Fixes the following security alerts in the package:
* https://github.com/rvagg/bl/pull/22
* https://github.com/advisories/GHSA-ff7x-qrg7-qggm
* https://github.com/advisories/GHSA-6c8f-qphg-qjgp
* https://github.com/advisories/GHSA-vh95-rmgr-6w4m
* https://github.com/advisories/GHSA-hxcc-f52p-wc94

#### What's the new behavior?

Less security alerts

Reviewers: @timbuckley

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,5 +86,12 @@
     "slate-react": "*",
     "source-map-loader": "^0.2.4",
     "typescript": "^3.7.2"
+  },
+  "resolutions": {
+    "bl": "^2.2.1",
+    "dot-prop": "^4.2.1",
+    "kind-of": "^6.0.3",
+    "minimist": "^0.2.1",
+    "serialize-javascript": "^4.0.0"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -86,5 +86,12 @@
     "slate-react": "*",
     "source-map-loader": "^0.2.4",
     "typescript": "^3.7.2"
+  },
+  "resolutions": {
+    "bl": "^2.2.1",
+    "dot-prop": "^4.2.1",
+    "kind-of": "^6.0.3",
+    "minimist": "^0.2.1",
+    "serialize-javascript": "^4.0.0"
```
TheSpyder, line 95: There is a webpack 4.44.0, this resolved the `serialise-javascript` 2.x dependency (and possibly others) for us. Specifically v1.4.4 of `terser-webpack-plugin` upgrades to `serialise-javascript` 3.1.0 fixing the security issue.

TheSpyder: oh webpack is coming in via `next`, sigh. I don't know if their most recent version includes the newer webpack

jason-krypton: It doesn't I've updated every package possible :) other than prettier and TS

TheSpyder: Hmm so #3845 should remove the need to override `serialize-javascript` because it upgrades `next` (which does upgrade webpack). Looking through the yarn lock, nothing depends on the old version yet it's still listed.

Yarn isn't perfect, try regenerating the lock file and see if that removes any of the old dependencies (at the very least it will remove `serialize-javascript@2.1.2`).

jason-krypton: I can try once that PR gets merged. I think there was still a dependency... but its a bit moot juggling multiple PRs until they get merged.

PR #3850, opened 2020-08-31
Upgrade immer to 7.0.0

aiwenar:

Upgrading a dependency.

#### What's the new behavior?

Hopefully none.

I'm proposing this upgrade because I've run across immerjs/immer#559, a fix for which is included in immer >= 6.0.4. This issue does not occur in Slate itself (hence why I didn't mark this PR as fixing a bug), but rather when a Slate type such as `Range` is stored in an immer-controlled container, and `transform`ed. For example this:
```typescript
let state = {
    anchor: { path: [0, 0], offset: 0 },
    focus: { path: [0, 0], offset: 10 },
}

state = produce(state, s => Range.transform(s, { type: 'insert_text', path: [0, 0], offset: 5, text: 'new' }))
```
will result in `state.focus.path` being a revoked proxy.

Since this changes a major version of a public dependency it is also a breaking change in Slate. However, I have verified that Slate works with Immer 5, 6, and 7, so in theory we could accept immer >= 5.00 < 8.0.0 and let the consumer choose which version of Immer they want, thus avoiding a breaking change. I'm not proposing that in this PR as I don't know how to do that in npm.

Fixes: 
Reviewers:

### DIFF
--- a/packages/slate/src/interfaces/node.ts
+++ b/packages/slate/src/interfaces/node.ts
@@ -292,7 +292,7 @@ export const Node: NodeInterface = {
       )
     }
 
-    const newRoot = produce(root, r => {
+    const newRoot = produce({ children: root.children }, r => {
       const [start, end] = Range.edges(range)
       const nodeEntries = Node.nodes(r, {
         reverse: true,
--- a/packages/slate-history/package.json
+++ b/packages/slate-history/package.json
@@ -14,7 +14,7 @@
     "dist/"
   ],
   "dependencies": {
-    "immer": "^5.0.0",
+    "immer": "^7.0.0",
     "is-plain-object": "^3.0.0"
   },
   "devDependencies": {
--- a/packages/slate/package.json
+++ b/packages/slate/package.json
@@ -16,7 +16,7 @@
   "dependencies": {
     "@types/esrever": "^0.2.0",
     "esrever": "^0.2.0",
-    "immer": "^5.0.0",
+    "immer": "^7.0.0",
     "is-plain-object": "^3.0.0",
     "tiny-warning": "^1.0.3"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -5693,10 +5693,10 @@ image-extensions@^1.1.0:
   resolved "https://registry.yarnpkg.com/image-extensions/-/image-extensions-1.1.0.tgz#b8e6bf6039df0056e333502a00b6637a3105d894"
   integrity sha1-uOa/YDnfAFbjM1AqALZjejEF2JQ=
 
-immer@^5.0.0:
-  version "5.3.6"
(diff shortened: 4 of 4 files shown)


packages/slate/src/interfaces/node.ts
```diff
@@ -197,7 +197,7 @@ export const Node = {
       )
     }
 
-    const newRoot = produce(root, r => {
+    const newRoot = produce({ children: root.children }, r => {
```
aiwenar, line 200: Since 7.0.0, Immer freezes all properties of the new state (see https://github.com/immerjs/immer/issues/462 for more details). This caused `editor.operations` to be frozen when calling `Editor.fragment(editor, ...)`.

BrentFarese: Confused by this comment.  Does this PR introduce an issue with `editor.operations` by freezing it if `Editor.fragment` is called?

BrentFarese:
It looks ok to me.  The 1 thing I worry about is the test coverage in core not covering all areas and some issue with freezing some portion of the editor state creeping up, or some other Immer issue.  But in general, I think we should upgrade this dependency.  It is a major dependency tho.

BrentFarese:
Glad to merge this in @aiwenar if you can clean up the conflicts and get it passing CI?  Thanks!

aiwenar:
Sure. Would you prefer me to merge master into this branch, or rebase this branch on master?

packages/slate/src/interfaces/node.ts
aiwenar: If I recall correctly, in immer 6 the way freezing works has changed. Previously immer (5) would only freeze the new object, starting with 6 it will also recursively freeze all properties. And because of structural sharing, this means that any property of `root` unchanged by the reducer function would also be frozen.

BrentFarese:
> Sure. Would you prefer me to merge master into this branch, or rebase this branch on master?

Either works for me.  Thanks!

aiwenar:
Squashed and rebased on top of master, otherwise unchanged. Since this PR was opened immer released versions 7.0.15 and 8.0.0, but I've left it at 7.0.7 as this is the version I've used when I opened this PR.

BrentFarese:
Fixes #4034.

niieani:
Any ETA on when we can expect a next beta containing this to be released? I'd love to be able to make use of the new TypeScript improvements, but cannot since the last one out (0.61) is broken due to #4034.

Many thanks for your help! 🙇

BrentFarese:
> @BrentFarese Any ETA on when we can expect a next beta containing this to be released? I'd love to be able to make use of the new TypeScript improvements, but cannot since the last one out (0.61) is broken due to #4034.
> 
> Many thanks for your help! 🙇

No exact ETA but hope to release something shortly so this issue gets resolved on the `@next` release.

williamstein:
I can confirm that @next is updated, and with that update, which includes this PR, #4034 is fixed for me.  Thanks!

PR #3851, opened 2020-08-31
Normalize moved node

aiwenar:

feature

#### What's the new behavior?

Normalize the node moved by a `move_node` operation.

When normalizations of a custom node depend on its parent, one currently must write them as

```typescript
editor.normalizeNode = entry => {
    const [node, path] = entry

    if (isCustomNode(node)) {
        // parent-independent normalization
    }

    for (const [child, childPath] of Node.children(editor, path)) {
        if (isCustomNode(child) && isCustomNodeParent(node)) {
            // parent-dependent normalization
        }
    }
}
```

with this change this can be instead written as

```typescript
editor.normalizeNode = entry => {
    const [node, path] = entry

    if (isCustomNode(node)) {
        // parent-independent normalization

        const [parent, parentPath] = Editor.parent(editor, path)
        if (isCustomNodeParent(node)) {
            // parent-dependent normalization
        }
    }
}
```

which helps keep related normalizations together and improves readability.

Fixes: 
Reviewers:

### DIFF
--- a/packages/slate/src/create-editor.ts
+++ b/packages/slate/src/create-editor.ts
@@ -353,7 +353,11 @@ const getDirtyPaths = (op: Operation) => {
         newAncestors.push(p!)
       }
 
-      return [...oldAncestors, ...newAncestors]
+      const newParent = newAncestors[newAncestors.length - 1]
+      const newIndex = newPath[newPath.length - 1]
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks @aiwenar!

PR #3859, opened 2020-09-04
Expose interface to setNormalizing

jason-codaio:
This PR exposes an API on Editor to control the normalization state. The reason for exposing this is `withoutNormalizing` is not sufficient in complex systems. Imaging a complex product where slate is just one collaborative surface. The main operation sync loop might look something like:

```
onOperationBatchReceived(ops: Operations): void {
  this.emit(START_BATCH);
  for (const op of ops) {
    // Not every op in a batch has the same objectId
    const obj = findObj(op.objectId);
    // Not every object here is a Slate Editor and the API is a shared interface for all collaborative objects
    obj.apply(op);
  }
  this.emit(END_BATCH);
}
```

The op batches are handled at a level and abstraction above slate which makes applying all operations in a withoutNormalizing callback difficult. Instead our SlateEditor listens to START_BATCH / END_BATCH events and toggles normalizing appropriately.

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -254,6 +254,7 @@ export interface EditorInterface {
   ) => RangeRef
   rangeRefs: (editor: Editor) => Set<RangeRef>
   removeMark: (editor: Editor, key: string) => void
+  setNormalizing: (editor: Editor, isNormalizing: boolean) => void
   start: (editor: Editor, at: Location) => Point
   string: (
     editor: Editor,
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @jason-codaio I see what you mean here. Can you refactor the existing `withoutNormalizing` to use this method under the covers, so that we are only using the map in one place? Thanks!

jason-codaio:
@ianstormtaylor Sorry about the delay this PR was not getting traction and I forgot about it. I have made the suggested updates.

jason-codaio:
@TheSpyder @thesunny Was wondering if I could get this merged now that I made the suggested changed.

PR #3866, opened 2020-09-09
Fixed issue where virtual keyboard closes when resynchronizing selection in Chrome on Android

wleroux:

This pull request fixes an issue where the virtual keyboard in Chrome on Android closes when all selections are removed. The code removes all selections and adds the new selection as part of its resynchronization. However, this causes the virtual keyboard to close.

#### What's the new behavior?

The virtual keyboard on Android does not close when resynchronizing the focus.

Instead of removing the selection and adding a new range, this change will set the selection's base and extent to the desired selection. As the selection is never lost, Chrome on Android doesn't close the virtual keyboard.

As part of this change, I also replaced `setReverseDomSelection` with `Selection#setBaseAndExtent`. As `setReverseDomSelection` assumes the selection has no range, it would not work with the previous changes.

`Selection#setBaseAndExtent` support forward and backward selection.

> If anchor is before focus, set the start the newRange's start to anchor and its end to focus. Otherwise, set the start them to focus and anchor respectively.
> https://w3c.github.io/selection-api/#dom-selection-setbaseandextent

> Note: If the focus position appears before the anchor position in the document, the direction of the selection is reversed — the caret is placed at the beginning of the text rather the end, which matters for any keyboard command that might follow. For example, Shift + ➡︎ would cause the selection to narrow from the beginning rather than grow at the end.
> https://developer.mozilla.org/en-US/docs/Web/API/Selection/setBaseAndExtent

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -33,7 +33,6 @@ import {
   isDOMText,
   DOMStaticRange,
   isPlainTextOnlyPaste,
-  setReverseDomSelection,
 } from '../utils/dom'
 import {
   EDITOR_TO_ELEMENT,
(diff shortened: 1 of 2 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -180,21 +179,23 @@ export const Editable = (props: EditableProps) => {
...
-        setReverseDomSelection(newDomRange, domSelection)
-      } else {
-        domSelection.addRange(newDomRange!)
-      }
+      domSelection.setBaseAndExtent(
+        newDomRange.startContainer,
+        newDomRange.startOffset,
+        newDomRange.endContainer,
+        newDomRange.endOffset
+      )
```
AndrewScull, line 191: IIUC, although `setBaseAndExtent` can create a reversed selection, the start and end of the range (`newDomRange`) will always be in the forwards direction so this will always be creating a forwards selection.

Previously, `Range.isBackward` was used to check the orientation of the selections's anchor and focus (note the difference from a range's start and end). If the selections is backwards, the end of the range needs to be used as the anchor rather than the start.

wleroux: I've updated the code to correct for that! Nice catch 👍

That behavior defied my intuition... If the `setBaseAndExtent` is capable of keeping track of the focus and anchor, it stands to reason (and wrongly so) that it was supported by `DOMRange`. It seems like that's a limitation of HTML itself, not just `newDomRange`.

Issue #3339, opened 2019-12-18
Can't backspace over first character when in second text node without incorrect block merge


thesunny:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

When backspacing in a block, if you are within the second text node in a block, you cannot backspace over the first text node with one character in it without also merging the block with the previous one which is undesired behavior.

![screencast](http://g.recordit.co/rKwTulC3hV.gif)

Slate: current
Browser: Chrome
OS: Mac

#### What's the expected behavior?

It shouldn't do that

udayanshevade:
Jumping over the first node with the cursor also has a similar effect (it skips the first position in the second block):

![cursor_jump_bug](https://user-images.githubusercontent.com/11330558/93727031-22f9ec00-fb6e-11ea-8e9d-8aa276e0072a.gif)

The edge position of an element node isn't yielded if the node is comprised of multiple text/inline nodes.

In the rich text example above, it yields the following [positions](https://github.com/ianstormtaylor/slate/blob/4c3e737dda95d5ab67a94af903b25512d926c7fe/packages/slate/src/interfaces/editor.ts#L976):

![cursor_back_bug_positions](https://user-images.githubusercontent.com/11330558/93727451-15ddfc80-fb70-11ea-80b6-ed3ee344b976.png)

Note that the location `{ path: [1, 0], offset: 0 }`, which corresponds to the point before the `S` is missing.

On the other hand, when the node contains uninterrupted text (e.g. "ce it's r" isn't bolded), the expected edge position appears:
![cursor_back_bug_plaintext_positions](https://user-images.githubusercontent.com/11330558/93727500-46be3180-fb70-11ea-816b-6b16fd844d1a.png)

---

Took a crack at it here: #3883 

After the update:
![richtext_char_backspace_fix](https://user-images.githubusercontent.com/11330558/93725213-048df380-fb62-11ea-9455-41ee4e2e3efd.gif)

Issue #3466, opened 2020-01-24
Paragraphs are unnecessarily merged


jazblueful:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?
https://codesandbox.io/s/slate-reproductions-x6nf8

Put cursor inside bold text
Tap backspace until you're at 2
Press backspace again
Number 2 is deleted and both paragraphs are merged into one

Slate: 0.57.1
Browser: Chrome
OS: Linux

#### What's the expected behavior?
Number 2 is deleted without merging the paragraphs into one

dimitriadamou:
Does this happen on the 0.58 version? https://codesandbox.io/s/slate-reproductions-2yr90?file=/index.js

Testing from Mac on Chrome, no replicable issue
![May-06-2020 19-24-31](https://user-images.githubusercontent.com/3094239/81161651-3e08fc80-8fcf-11ea-99e5-15bfbe67bbb4.gif)

jazblueful:
Yes, the bug is still there. I just tested it on both Windows and Linux version of Chrome. Btw you need to put the cursor in the MIDDLE of the bold text before tapping backspace in order to replicate the bug.

dimitriadamou:
Ah I see now, if I get time over the weekend I may have a play with it

udayanshevade:
Looks like a duplicate of #3339

I took a crack at fixing it:

![richtext_char_backspace_fix_2](https://user-images.githubusercontent.com/11330558/93726641-065cb480-fb6c-11ea-9ff5-a7698c0a1a4d.gif)

PR #3883, opened 2020-09-21
Fix skip in editor positions

udayanshevade:
Fixing a bug

#### What's the new behavior?
Does _not_ merge a block with its previous one when the user is backspacing over the first node and their cursor is in the second node. The fix works the same whether deleting backwards at the start of a block or deleting forwards at the end.

- With a styled text node:
richtext_char_backspace_fix

- Also works with inline nodes:
inline_char_backspace_fix

_Before the fix_: While scanning a text node for cursor positions, the loop [breaks early](https://github.com/ianstormtaylor/slate/blob/4c3e737dda95d5ab67a94af903b25512d926c7fe/packages/slate/src/interfaces/editor.ts#L1066) when there is no more string left. But if the [element node](https://github.com/ianstormtaylor/slate/blob/4c3e737dda95d5ab67a94af903b25512d926c7fe/packages/slate/src/interfaces/editor.ts#L1045) containing the string content is comprised of multiple inlines or text nodes, the edge position of the element node is effectively ignored and never [yielded](https://github.com/ianstormtaylor/slate/blob/4c3e737dda95d5ab67a94af903b25512d926c7fe/packages/slate/src/interfaces/editor.ts#L1075). This skip can be seen both while deleting over the first node and also while simply jumping over it with the cursor.

_The fix_: Constrain when to break the loop early using additional conditions.

- [ x ] The new code matches the existing patterns and styles.
          - I tried staying within the existing pattern. Please let me know if anything needs to be adjusted.
- [ x ] The tests pass with `yarn test`.
          - Existing tests pass, and new tests have been added.
- [ x ] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
          - The lint passes locally **except** for line endings (I'm working on Windows). Wasn't sure whether to update the eslint config to accommodate it? Will pose this question in slack and follow up with the change here.
- [ x ] The relevant examples still work. (Run examples with `yarn start`.)
          - Yes the examples still work.

Fixes: #3339 and #3466

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -990,6 +990,7 @@ export const Editor = {
     const range = Editor.range(editor, at)
     const [start, end] = Range.edges(range)
     const first = reverse ? end : start
+    const last = reverse ? start : end
     let string = ''
     let available = 0
     let offset = 0
@@ -1040,7 +1041,6 @@ export const Editor = {
           const s = Path.isAncestor(path, start.path)
             ? start
             : Editor.start(editor, path)
-
           const text = Editor.string(editor, { anchor: s, focus: e })
           string = reverse ? reverseText(text) : text
           isNewBlock = true
@@ -1062,8 +1062,12 @@ export const Editor = {
         }
 
         while (true) {
-          // If there's no more string, continue to the next block.
-          if (string === '') {
+          // If there's no more string (and we've exhausted the available space
+          // or we're at the end of the range), break from the current node.
+          if (
+            string === '' &&
+            (available === 0 || Point.equals(last, { path, offset }))
+          ) {
             break
           } else {
             advance()
--- a/packages/slate/test/interfaces/Editor/positions/all/block-multiple-inline-multiple-reverse.tsx
+++ b/packages/slate/test/interfaces/Editor/positions/all/block-multiple-inline-multiple-reverse.tsx
@@ -1,0 +1,51 @@
+/** @jsx jsx */
+import { Editor } from 'slate'
+import { jsx } from '../../../..'
+
+export const input = (
+  <editor>
+    <block>one</block>
+    <block>
+      two
+      <inline>
+        three
+        <inline>four</inline>
+      </inline>
+      <inline>five</inline>
+    </block>
+    <block>six</block>
+  </editor>
+)
+export const test = editor => {
+  return Array.from(Editor.positions(editor, { reverse: true, at: [] }))
+}
+export const output = [
+  { path: [2, 0], offset: 3 },
+  { path: [2, 0], offset: 2 },
+  { path: [2, 0], offset: 1 },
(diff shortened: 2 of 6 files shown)


packages/slate/test/interfaces/Editor/positions/range/inline-reverse.tsx
```diff
@@ -0,0 +1,44 @@
+/** @jsx jsx */
```
udayanshevade, line 1: I felt this test was missing (in keeping with the other existing tests). Including this helped correct a bad assumption on my part too, while trying to understand the issue, so I felt it might be good to include it here as well.

packages/slate/src/interfaces/editor.ts
```diff
@@ -1062,8 +1062,12 @@ export const Editor = {
         }
 
         while (true) {
-          // If there's no more string, continue to the next block.
-          if (string === '') {
+          // If there's no more string (and we've exhausted the available space
+          // or we're at the end of the range), break from the current node.
+          if (
+            string === '' &&
+            (available === 0 || Point.equals(last, { path, offset }))
```
udayanshevade, line 1069: Additional constraints for when to break early while scanning text content for possible cursor positions.

Either the available space has been exhausted or we've reached the end of the permitted range. These seem to accommodate existing use cases, while also fixing the bug.

udayanshevade:
One persisting issue I did see in some cases was the cursor jumping to the wrong block, despite the blocks not being merged:

richtext_word_backspace_fix_issue

It was seen specifically when there were multiple/nested styled text nodes, as shown above.

I added further tests to validate my changes in these scenarios and the change to `Editor.positions` still seemed to work as expected.

So I'm thinking it's a related but maybe separate issue. If it turns out that I should address it in this PR though, I can definitely do that.

packages/slate/test/interfaces/Editor/positions/all/block-multiple-inline-multiple-reverse.tsx
```diff
@@ -0,0 +1,51 @@
+/** @jsx jsx */
```
udayanshevade, line 1: I'm not sure about the naming for these new test files. Alternate suggestions would very much be welcomed.

I feel the names are getting out of hand in this case, but I wasn't sure how to reconcile the existing pattern with these compound criteria.

ianstormtaylor:
Hey @udayanshevade thanks for taking the time to work on this and sorry for the slow response. I've merged a few positions-related fixes recently, and I think it might have solved this. But if you find that it's still an issue please comment and I can reopen. Thank you!

udayanshevade:
Thank you! Was happy to try and help in any way I could. I wasn't able to reproduce the bug now which is awesome. 

I did notice a small issue with cursor placement while validating (at the end of the gif below). Not sure if an issue for that exists already, but it's probably a different thing.

cursor_placement_on_backspace_bug

PR #3888, opened 2020-09-25
Use native character insertion to fix browser/OS text features

bkrausz:
See #3293, which hasn't seen any action despite a change request. This takes that PR and makes TextString a functional component, the one requested change.

Huge thanks to @luddep for doing almost all of the work here.

Fixes: #2061 (and some of #2051?)
Reviewers: @ianstormtaylor

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -46,6 +46,7 @@ import {
   PLACEHOLDER_SYMBOL,
   EDITOR_TO_WINDOW,
 } from '../utils/weak-maps'
+import { asNative, flushNativeEvents } from '../utils/native'
 
 /**
  * `RenderElementProps` are passed to the `renderElement` handler.
@@ -268,7 +269,47 @@ export const Editable = (props: EditableProps) => {
           return
         }
 
-        event.preventDefault()
+        let native = false
+        if (
+          type === 'insertText' &&
+          selection &&
+          Range.isCollapsed(selection) &&
+          // Only do it for single character events, for the simplest scenario,
+          // for now.
+          event.data &&
+          event.data.length === 1 &&
+          // Chrome seems to have issues correctly editing the start of nodes.
+          // I see this when there is an inline element, like a link, and you select
+          // right after it (the start of the next node).
+          selection.anchor.offset !== 0
+        ) {
+          native = true
+
+          // Skip native if there are marks, as that means
+          // `insertText` will insert a node, not just text.
+          if (editor.marks) {
+            native = false
+          }
+
+          // and because of the selection moving in `insertText` (create-editor.tx).
+          const { anchor } = selection
+          const inline = Editor.above(editor, {
+            at: anchor,
+            match: n => Editor.isInline(editor, n),
+            mode: 'highest',
+          })
+          if (inline) {
+            const [, inlinePath] = inline
+
+            if (Editor.isEnd(editor, selection.anchor, inlinePath)) {
+              native = false
+            }
+          }
+        }
+
+        if (!native) {
+          event.preventDefault()
+        }
 
         // COMPAT: For the deleting forward/backward input types we don't want
         // to change the selection because it is the range that will be deleted,
@@ -380,7 +421,13 @@ export const Editable = (props: EditableProps) => {
             if (data instanceof window.DataTransfer) {
               ReactEditor.insertData(editor, data as DataTransfer)
             } else if (typeof data === 'string') {
-              Editor.insertText(editor, data)
+              // Only insertText operations use the native functionality, for now.
+              // Potentially expand to single character deletes, as well.
+              if (native) {
+                asNative(editor, () => Editor.insertText(editor, data))
+              } else {
+                Editor.insertText(editor, data)
+              }
             }
 
             break
@@ -552,6 +599,13 @@ export const Editable = (props: EditableProps) => {
             },
             [readOnly]
           )}
+          onInput={useCallback((event: React.SyntheticEvent) => {
+            // Flush native operations, as native events will have propogated
+            // and we can correctly compare DOM text values in components
+            // to stop rendering, so that browsers functions like autocorrect
+            // and spellcheck work as expected.
+            flushNativeEvents(editor)
+          }, [])}
           onBlur={useCallback(
             (event: React.FocusEvent<HTMLDivElement>) => {
               if (
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -274,7 +274,7 @@ export const Editable = (props: EditableProps) => {
           type === 'insertText' &&
           selection &&
           Range.isCollapsed(selection) &&
-          // Only do it for single character a-z or space for now.
+          // Only use native character insertion for single characters a-z or space for now.
           // Long-press events (hold a + press 4 = ä) to choose a special character otherwise
           // causes duplicate inserts.
           event.data &&
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -287,7 +287,7 @@ export const Editable = (props: EditableProps) => {
         ) {
           native = true
 
-          // Skip native if there are marks, as that means
+          // Skip native if there are marks, as
           // `insertText` will insert a node, not just text.
           if (editor.marks) {
             native = false
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -293,7 +293,7 @@ export const Editable = (props: EditableProps) => {
             native = false
           }
 
-          // and because of the selection moving in `insertText` (create-editor.tx).
+          // and because of the selection moving in `insertText` (create-editor.ts).
           const { anchor } = selection
           const inline = Editor.above(editor, {
             at: anchor,
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -274,10 +274,12 @@ export const Editable = (props: EditableProps) => {
           type === 'insertText' &&
           selection &&
           Range.isCollapsed(selection) &&
-          // Only do it for single character events, for the simplest scenario,
-          // for now.
+          // Only do it for single character a-z or space for now.
+          // Long-press events (hold a + press 4 = ä) to choose a special character otherwise
+          // causes duplicate inserts.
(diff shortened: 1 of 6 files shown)


maccman:
This spell-checking bug essentially makes Slate unusable in production in my opinion.

I know Slate has some new core-contributors - I urge them to look at this PR.

maccman:
Here's my hacky workaround I'm using for now:

```  const [spellCheck, setSpellCheck] = useState(true)
  const debouncedSetSpellCheck = useMemo(() => debounce(setSpellCheck, 300), [])

  const toggleSpellCheck = () => {
    // Slate has a spell checking bug where spell checking gets invoked too
    // aggressively. We can work around this by toggling it when the user types.
    // https://rb.gy/qycfpa
    setSpellCheck(false)
    debouncedSetSpellCheck(true)
  }
```

And:

```        <Slate
          onChange={newValue => {
              toggleSpellCheck()
            }
          }}>

          <EditablePlugins
            spellCheck={spellCheck}```

**Review by ianstormtaylor — COMMENTED** [review]
Looks good to me generally. It just needs someone to verify that this doesn't break other things, since this is a pretty core change.

**Comment by bkrausz** [comment]
I've been using this in production on a fork for months now with no issues, though all of our users are on Chrome.

The only issue I found was in doing complex normalizations on top of nativeOps, I just fixed that via the latest commit (only required an extra `withoutNormalization`.

Is there other verification you would like to see?

**Comment by luddep** [comment]
If you want to cherry-pick d7e444ab54d445d16b44b3471cf22831eb7752c6, it solves a macOS chrome bug where a long press to select a special character (hold a + press 4 = ä) would double insert until native events flushed -- solved for now by just sidestepping it entirely and limiting native events to single char a-z inserts. In my limited testing, those are the only keyboard events (on US keyboards) that trigger spell check corrections.

**Comment by MohsenElgendy** [comment]
Is there an estimation on when would this PR be merged with the master?

**Comment by cmditch** [comment]
Very glad a fix for the spellcheck flicker is out there. Appreciate the work @ryanmitts, @luddep, @bkrausz have put towards this. 🎉 

Pretty excited for this fix to get merged.

@thesunny do you see any issues with the logic of this PR?

**Review by dylans — CHANGES_REQUESTED** [review]
I hate to ask @bkrausz given how long this PR (and its previous form have been going), but if you could rebase it against the latest in main, I will get this landed asap. If you don't have time I'll do the rebase.

**Comment by bkrausz** [comment]
@dylans done. I'm having trouble getting tests to pass on my local machine, but the same 5 fail for me on main and on this branch, so hopefully nothing to fix.

**Comment by dylans** [comment]
> @dylans done. I'm having trouble getting tests to pass on my local machine, but the same 5 fail for me on main and on this branch, so hopefully nothing to fix.

Ok, we do have one integration test that appears to be legitimately failing now @bkrausz , the test in https://github.com/ianstormtaylor/slate/blob/main/cypress/integration/examples/code-highlighting.ts#L19-L34 

Here's the specific error: ```  1) code highlighting
       highlights javascript syntax:
     AssertionError: Timed out retrying after 4000ms: expected '<span.css-14qid8c>' to contain 'const'
      at Context.eval (http://localhost:3000/__cypress/tests?p=cypress/integration/examples/code-highlighting.ts:124:14)```

Please have a look when you have time.

**Comment by bkrausz** [comment]
Is it possible that this test is flaky? I'm seeing it working correctly locally, both via test passing and manually checking on a new build. Would you mind triggering a re-run to confirm and/or snagging the video from the test server? I don't have access to either.

**Comment by bkrausz** [comment]
On further inspection it looks like this PR isn't actually doing what it claims to do anymore...we still get flickering on spellcheck. I'm using it to get other native actions (specifically so horizontal scroll will work so Slate can emulate Input boxes). This PR needs deeper investigation: we are still avoiding calling `preventDefault`, but it seems that's no longer sufficient.

I doubt I'll have time to go that deep on this PR and fix it again, if anyone else is interested in taking this over that would be more than welcome.

**Comment by lukesmurray** [comment]
If anyone wants to do it you could potentially use git bisect to find the commit where the change broke.

**Comment by bkrausz** [comment]
FYI this PR does in fact still fix autocorrect support in Safari, just not the spellcheck squiggle blink.

**Comment by bkrausz** [comment]
Ok, I got nerdsniped enough to bisect this...it's from #4158, which causes every string to rerender. I'm pushing up a revert for it, can someone who knows how to test IME see if the bug described there is also fixed by this one? We may need to add a few extra characters to the [allowlist](https://github.com/ianstormtaylor/slate/pull/3888/commits/6a51623b905bc69e3de6edfc4f88a526224e9905#diff-1d0c52ec4c5562d96965f789988bd4efd4a6d24b88527918987d06bcc43d15adR282), as I don't know what characters it produces.

**Comment by dylans** [comment]
> Ok, I got nerdsniped enough to bisect this...it's from #4158, which causes every string to rerender. I'm pushing up a revert for it, can someone who knows how to test IME see if the bug described there is also fixed by this one? We may need to add a few extra characters to the [allowlist](https://github.com/ianstormtaylor/slate/pull/3888/commits/6a51623b905bc69e3de6edfc4f88a526224e9905#diff-1d0c52ec4c5562d96965f789988bd4efd4a6d24b88527918987d06bcc43d15adR282), as I don't know what characters it produces.

I know how to test IME. I’ll look at it over the weekend.

**Review by dylans — APPROVED** [review]
Overall this looks good to me @bkrausz... I'm going to let it sit for a couple more days so I can give it another look and in case anyone else wants to take a second look.

**Comment by brycedewitt** [comment]
Just piping in to check I'm clear on this - this fixes autocorrect on safari but not the squiggles issue?

**Comment by bkrausz** [comment]
> Just piping in to check I'm clear on this - this fixes autocorrect on safari but not the squiggles issue?

It does...something to improve blinking. With this PR all spellcheck on the line is removed until a space is pressed. This *almost* matches native browsing, which blocks the specific word's spellcheck while it's being edited. I do not know why this is the case.

https://user-images.githubusercontent.com/364070/128993742-f0bfdda2-b3ba-40c7-9b88-e18b3f891ecb.mov

File: packages/slate-react/src/components/editable.tsx [hunk 1]
```diff
@@ -268,7 +269,49 @@ export const Editable = (props: EditableProps) => {
...
+        ) {
+          native = true
+
+          // Skip native if there are marks, as that means
+          // `insertText` will insert a node, not just text.
+          if (editor.marks) {
+            native = false
+          }
+
+          // and because of the selection moving in `insertText` (create-editor.tx).
```
dylans, line 296:
```suggestion
          // and because of the selection moving in `insertText` (create-editor.ts).
```

packages/slate-react/src/components/editable.tsx
```diff
@@ -268,7 +269,49 @@ export const Editable = (props: EditableProps) => {
...
+          selection &&
+          Range.isCollapsed(selection) &&
+          // Only do it for single character a-z or space for now.
+          // Long-press events (hold a + press 4 = ä) to choose a special character otherwise
+          // causes duplicate inserts.
+          event.data &&
+          event.data.length === 1 &&
+          /[a-z ]/i.test(event.data) &&
+          // Chrome seems to have issues correctly editing the start of nodes.
+          // I see this when there is an inline element, like a link, and you select
```
dylans, line 284:
```suggestion
          // When there is an inline element, e.g. a link, and you select
```

packages/slate-react/src/components/editable.tsx
```diff
@@ -268,7 +269,49 @@ export const Editable = (props: EditableProps) => {
...
+          // right after it (the start of the next node).
+          selection.anchor.offset !== 0
+        ) {
+          native = true
+
+          // Skip native if there are marks, as that means
```
dylans, line 290:
```suggestion
          // Skip native if there are marks, as
```

packages/slate-react/src/components/editable.tsx
```diff
@@ -268,7 +269,49 @@ export const Editable = (props: EditableProps) => {
           return
         }
 
-        event.preventDefault()
+        let native = false
+        if (
+          type === 'insertText' &&
+          selection &&
+          Range.isCollapsed(selection) &&
+          // Only do it for single character a-z or space for now.
```
dylans, line 277:
```suggestion
          // Only use native character insertion for single characters a-z or space for now.
```

packages/slate-react/src/components/editable.tsx
```diff
@@ -552,6 +601,13 @@ export const Editable = (props: EditableProps) => {
             },
             [readOnly]
           )}
+          onInput={useCallback((event: React.SyntheticEvent) => {
+            // Flush native operations, as native events will have propogated
+            // and we can correctly compare DOM text values in components
+            // to stop rendering, so that browsers functions like autocorrect
```
dylans, line 607:
```suggestion
            // to stop rendering, so that browser functions like autocorrect
```

dylans (commented):
minor comments cleanup

githoniel:
@bkrausz sorry for the bug caused by my hacky PR that drove you crazy😥。

@dylans but this pr did not fix the IME double input when add mark bug. I have test in the main branch with this pr merged like below: 
xx

this anoy bug will cause slate to crash and this is very common used for the peope using IME

I have a new idea to fix this: create an node with place-holder text and move selection to it in advance before composition start, remove the node if there is no text after composition end. I will try to implement later

dylans:
> @bkrausz sorry for the bug caused by my hacky PR that drove you crazy😥。
> 
> @dylans but this pr did not fix the IME double input when add mark bug. I have test in the main branch with this pr merged like below:
> xx
> 
> this anoy bug will cause slate to crash when click on the text that should not exist.
> 
> I have a new idea to fix this without force re-render string component. I will try to implement later

ok @githoniel let me know. We've had quite a few IME related contributions lately and it would be great to be done with the double input issue resolved for good.

dylans:
@bkrausz fyi, this update seems to have caused a regression for code highlighting/decorations, see the notes in #4445.

jaked:
rather than defer the `insertText` call to the underlying `Editor`, would it make sense to accumulate native operations in `onDOMBeforeInput` and flush them in `onInput` entirely within `Editable`, so `insertText` is not called at all until `onInput`? It seems like a lot of the bugs around this (e.g. #4530, #4541) have to do with interactions with non-default `insertText` implementations.

If this seems like a reasonable idea I can try to do a PR.

juliankrispel:
Has this worked as part of these prs? I have built commits from both this pr and #4605 and have not found a working version. Pointers appreciated 🙏

Issue #3625, opened 2020-04-20
Cannot get the start point in the node at path [1] because it has no start text node.


lavaldi:
#### Do you want to request a _feature_ or report a _bug_?

Report a bug

#### What's the current behavior?

Bug GIF:

<img src="http://g.recordit.co/fA2hwhTqXq.gif" />

https://codesandbox.io/s/slate-reproductions-cm3r7?file=/index.js

Slate: 0.57.1
Browser: Chrome
OS: Mac

#### What's the expected behavior?

The expected behavior is that the string can be parsed to Slate object without throwing an error.

---

I found that the bug is in this function

https://github.com/ianstormtaylor/slate/blob/f6bfe034d707693488c38da77537fd36cb8856cf/packages/slate-hyperscript/src/creators.ts#L27-L81

because the `p` tag has no text, the `children` parameter is an empty array, therefore the `for` loop is never performed and returns the variable `node` as an empty array.

lavaldi:
Perhaps a validation should be added that if `children` is an empty array that it returns `[{ text: '' }]` 🤔

DamareYoh:
Hi @lavaldi , have you tried `<p />` to see if it makes a difference? 

(Also, the gif with the minimum reproduction is a bit hard for me to use for helping you so a minimum reproduction in a code sandbox would be cool if that doesn't work)

Thanks!

lavaldi:
@DamareYoh there is a link to the codesanbox in the issue description. 

Also, the error is thrown with the image tag `<img src=”[image_url]” />`. You can test it in the codesandbox.

CameronAckermanSEL:
Hi @lavaldi , the issue is that the paragraph blocks the deserialize function generates are allowed to have no children. This is an invalid state for slate; all blocks need to have at least one text node, even if it is empty. The deserializer function should be updated to correct that issue.

As for images, you should be able to pass a `void` flag to the image block to indicate that it is not intended to have text children. See the image example for how some examples of void blocks.

lavaldi:
I see, thank you for your help 👍

TheSpyder:
> Perhaps a validation should be added that if `children` is an empty array that it returns `[{ text: '' }]` 🤔

If #4208 is merged, this will happen when `Editor.normalize` is used (for example `Editor.normalize(editor, { force: true })` can be used to normalize an entire document).

juliana-witzke:
This thread was super helpful for me! 
[...]

PR #3889, opened 2020-09-25
Documentation: Add empty text if children are empty

nkovacic:
If line 170 is not added issues like https://github.com/ianstormtaylor/slate/issues/3625 start surfacing. 
Because the deserialize example is mostly directly copy-pasted into projects, the added line would save many head-scratching down the lane.

Fixes: #
Reviewers: @


karldanninger:
Hey @nkovacic you saved my day...

Only note on your PR, change `const` to `let` since you are modifying children

PR #3894, opened 2020-09-28
Fix error when trying to grab 

msc117:
fix(react-editor): text node spacer can be non-existent if editor is in readonly mode

#### fixing a _bug_

```
TypeError: Cannot read property 'closest' of null
    at Object.toSlatePoint 
    at Object.toSlateRange 
```

#### What's the new behavior?

Here's my usecase:
I have a void node with a click handler that selects itself, opens a dialog, and sets the editor to readonly to save the selection while focus is lost. Changing the selection causes `useIsomorphicLayoutEffect` to trigger which tries to grab selection (https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/editable.tsx#L183) and causes the stracktrace above when it eventually hits this line https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugin/react-editor.ts#L438. 

---

Here's the root cause I found:
When the editor is set to readonly, void nodes have their text spacers removed (https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/element.tsx#L96), but there is code elsewhere that depends on this spacer always being there (https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/plugin/react-editor.ts#L438). 

I get around this by doing a null check and defaulting the offset to 1, which would be the same offset if the spacer was there.

Fixes: #
Reviewers: @

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "lerna": "2.7.1",
-  "version": "0.59.0",
+  "version": "0.59.1",
   "npmClient": "yarn",
   "useWorkspaces": true
 }
--- a/packages/slate-react/package.json
+++ b/packages/slate-react/package.json
@@ -1,7 +1,7 @@
 {
   "name": "slate-react",
   "description": "Tools for building completely customizable richtext editors with React.",
-  "version": "0.59.0",
+  "version": "0.59.1",
   "license": "MIT",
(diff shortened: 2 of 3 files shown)


ianstormtaylor:
Thanks @msc117!

Issue #3896, opened 2020-09-29
TODO: Convert all remaining JavaScript files to TypeScript


timbuckley:
#### Do you want to request a _feature_ or report a _bug_?

Request

#### What's the current behavior?

Currently 1.9% of the codebase is written in JavaScript, with almost all of the rest in TypeScript.

#### What's the expected behavior?

We'd like to transform as much of the remaining project to TypeScript as possible. Mainly this will involve transforming the `slate-hyperscript` tests.

vanya829:
Hi. I'd like to work on this issue and convert `slate-hyperscript` tests to typescript. May I?

PR #3904, opened 2020-10-06
convert js file to tsx

vanya829:
No visual changes

Fixes: #3896


IsaacMSchultz:
Hey! I think we were both working on this issue at the same time. I submitted my PR too which is basically the same.

 I did notice that your `tsconfig.json` in the `slate-hyperscript` folder is unchanged, does it need to be updated with an `"include"` parameter to only look inside the src directory. I found that without it, I was getting errors in my IDE (Visual Studio Code) stating that all typescript files needed to be placed in the root directory.

vanya829:
@IsaacMSchultz Hi. I didn't get any errors, but I work in WebStorm, not VSCode. PRs are similar, so maintainer can deside which one to merge and close second one.

PR #3906, opened 2020-10-06
Converting JSX test cases to TSX

IsaacMSchultz:
Added 'include' parameter to tsconfig to tell the compiler to only look inside the src directory

fixing a _bug_

#### What's the new behavior?

No visual changes. Updates are to the test cases the from `slate-hyperscript` package, which now use `.tsx` like the other slate packages.

using js-to-ts converter on the JSX test cases from the `slate-hyperscript` package. I also added an 'include' parameter to tsconfig of that package to tell the compiler to only look inside the src directory, since there are now typscript files in the test directory.

Fixes: #3896


ianstormtaylor:
Thanks @IsaacMSchultz!

Issue #3902, opened 2020-10-05
Editor.normalize contains a bug


BrentFarese:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

In `Editor.normalize`, when Slate retrieves the node to be normalized from `DIRTY_PATHS`, it simply does a call to `Editor.node` found [here](https://github.com/ianstormtaylor/slate/blob/60b56651b2e1e6f58bcf29a1a876612720a4fe2c/packages/slate/src/interfaces/editor.ts#L761).  However, frequently during the course of normalizing the editor, a `DIRTY_PATH` might be removed from the tree altogether.  This causes the call to `Editor.node` to `throw` since a node at the now-removed `DIRTY_PATH` cannot be found.

Slate: 0.59.0
Browser: Chrome
OS: Mac

#### What's the expected behavior?

Slate should guard against the node not being there to normalize using `Node.has(editor, dirtyPath)` **_before_** calling `Editor.node` so that there will never be an error thrown during the course of normalization.  Put simply, if the node doesn't exist in the tree, it does not need to be normalized.

PR #3907, opened 2020-10-07
Check if a node exists before normalization in Editor.normalize

IsaacMSchultz:

fixing a bug

#### What's the new behavior?

Slate should now guard against a node not being there so that there will never be an error thrown during the course of normalization. If a node does not exist in the tree, it will not be normalized.

editor.normalize now checks if a node exists using `Node.has(editor, dirtyPath)` **before** calling `Editor.node`

Fixes: #3902

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -758,13 +758,13 @@ export const Editor = {
         }
 
         const dirtyPath = getDirtyPaths(editor).pop()!
-        m++
 
         // If the node doesn't exist in the tree, it does not need to be normalized.
-        if (!Node.has(editor, dirtyPath)) continue
-
-        const entry = Editor.node(editor, dirtyPath)
-        editor.normalizeNode(entry)
+        if (Node.has(editor, dirtyPath)) {
+          const entry = Editor.node(editor, dirtyPath)
+          editor.normalizeNode(entry)
+        }
+        m++
       }
     })
   },
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -757,9 +757,13 @@ export const Editor = {
           `)
         }
 
-        const path = getDirtyPaths(editor).pop()!
-        const entry = Editor.node(editor, path)
-        editor.normalizeNode(entry)
+        const dirtyPath = getDirtyPaths(editor).pop()!
+
+        // If the node doesn't exist in the tree, it does not need to be normalized.
+        if (Node.has(editor, dirtyPath)) {
+          const entry = Editor.node(editor, dirtyPath)
+          editor.normalizeNode(entry)
+        }
         m++
       }
     })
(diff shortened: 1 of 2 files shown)


packages/slate/src/interfaces/editor.ts
```diff
@@ -757,9 +757,13 @@ export const Editor = {
...
-        const entry = Editor.node(editor, path)
-        editor.normalizeNode(entry)
+        const dirtyPath = getDirtyPaths(editor).pop()!
+
+        // If the node doesn't exist in the tree, it does not need to be normalized.
+        if (Node.has(editor, dirtyPath)) {
+          const entry = Editor.node(editor, dirtyPath)
+          editor.normalizeNode(entry)
+        }
         m++
```
BrentFarese, line 767:
```suggestion
        const dirtyPath = getDirtyPaths(editor).pop()!

        // If the node doesn't exist in the tree, it does not need to be normalized.
        if (!Node.has(editor, dirtyPath)) continue 
        
        const entry = Editor.node(editor, dirtyPath)
        editor.normalizeNode(entry)
        m++
```

BrentFarese (changes requested):
Minor requested change in code style mainly.

packages/slate/src/interfaces/editor.ts
IsaacMSchultz: Thanks for the input! I made the change, but I noticed a potential conflict.

With the continue line, the rest of the code is skipped. Which would also mean line 766 (`m++`) will never happen, which could potentially create a scenario where an error is not thrown when it should be due an incorrect number of iterations being kept.

Because of this I have moved that counter to before the continue statement. If this is not needed, I will move it back to the end of the loop.

BrentFarese: I think you're right.  If you want to move it back to your original formulation, that actually may be a bit cleaner.  I can then merge it in.  Thanks!

IsaacMSchultz: Thanks! I went ahead and changed it back and I will resolve this thread

BrentFarese (approved):

TheSpyder:
I’m not sure this is a good idea. If a node tagged as dirty does not exist it means an invalid operation was applied. This is one of the ways I have been able to detect if my transforms fail.

Silently ignoring such a critical failure will likely only lead to weird bugs later on.

BrentFarese:
So the node not existing in the normalization is not a true indication that an invalid op was applied.  For instance, if you remove the node in question during the normalization, it may not exist **_later_** in that same normalization, even though the path was marked as "dirty".  This precise situation arises in our codebase from time to time (we have our own normalization process that does more than Slate's default).  We have patched our existing code with this check so that Slate will not throw an error in the normalization.  Really, if the node doesn't exist, Slate shouldn't be trying to normalize it.

TheSpyder:
I understand what you're saying, the error doesn't make sense during standard integrations and/or custom normalisation, but for those of us using default normalisation it's critical. I just tracked down a very obscure bug in my transforms that my tests found quite easily because they hit this error.

Particularly for those of us implementing collaboration this error is an essential safety net. Can an option perhaps be added to return to throwing an error here? I don't want to have to patch my copy just to bring it back, or not upgrade past the release that includes the fix. I only need to turn the error on when running my tests.

Issue #3857, opened 2020-09-04
Cutting (Ctrl + X) block element does not remove the element from the editor


kamilmielnik:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

Slate: 0.58.4
Browser: All / Google Chrome Version 85.0.4183.83 (Official Build) (64-bit)
OS: All / Ubuntu 20.04

1. Go to https://www.slatejs.org/examples/images
2. Wait for image to load
3. Click the image to select it
4. Hit <kbd>Ctrl</kbd>+<kbd>X</kbd>

Image is not removed from the editor (it gets copied to clipboard though).
If I expand the selection some more (e.g. <kbd>Shift</kbd>+<kbd>→</kbd>), then it works as expected.

#### What's the expected behavior?

Image is removed from the editor and lands in the clipboard.

markogresak:
The problem here is the `Range.isExpanded(selection)` check:
https://github.com/ianstormtaylor/slate/blob/b616e75d636733733503c724c9084ff53e88dde8/packages/slate-react/src/components/editable.tsx#L646-L648

And the same check appears in the `deleteFragment`:
https://github.com/ianstormtaylor/slate/blob/b616e75d636733733503c724c9084ff53e88dde8/packages/slate/src/create-editor.ts#L133-L135

I was able to track back the changes to this huge commit:
https://github.com/ianstormtaylor/slate/commit/4ff6972096953d5774b7c82571d702e304c3f55f#diff-0fa78153ea6f5b538c17b22d490c0ad6R505-R507
https://github.com/ianstormtaylor/slate/commit/4ff6972096953d5774b7c82571d702e304c3f55f#diff-a1d9556f831dcc92748a9b1f532495e5R112-R116

@ianstormtaylor why does this check exist? 

I believe the check at `onCut` is unnecessarily duplicating the one in `deleteFragment`, and it's breaking the behavior. If `onCut` calls `deleteFragment`, it would at least be possible to override the default behavior with a custom plugin. 
But I think the `deleteFragment` condition is missing one more check, when the focused node is a void element.

kamilmielnik:
FYI: the same thing happens when dragging elements - they are _copied_, while they should be _moved_.

rneiss:
Confirming this issue still exists on 0.59.0. Also fails in firefox, and extends to all void elements in a document.

rneiss:
The pull request above I believe fixes this issue.

PR #3910, opened 2020-10-08
Cutting (Ctrl + X) on a void element copies content to clip board AND removes the element from the editor

rneiss:
#### Cutting (Ctrl + X) on a void element copies content to clip board AND removes the element from the editor 
This is a fix for issue number [3857](https://github.com/ianstormtaylor/slate/issues/3857) -- and also fixes the image example allowing the image to be cut and pasted.

#### What's the new behavior?

When a void is selected and a user tries to cut the object to the clipboard, the element is both copied to the clipboard removed from the editor as expected.

Fixes: #3857


ianstormtaylor:
Hey @rneiss thank you for submitting this. I've merged another PR which I believe has fixed this, but feel free to comment if not and we can re-open. Thanks!

PR #3915, opened 2020-10-09
docs(libraries): better distinguish useEditor and useSlate

irmerk:

No, documentation update

#### What's the new behavior?

N/A

N/A

Fixes: Better distinguishes between the two similar hooks of [`useEditor`](https://docs.slatejs.org/libraries/slate-react#useeditor) and [`useSlate`](https://docs.slatejs.org/libraries/slate-react#useslate).

### DIFF
--- a/docs/libraries/slate-react.md
+++ b/docs/libraries/slate-react.md
@@ -36,7 +36,7 @@ React hooks for Slate editors
 
 ###### `useEditor`
 
-Get the current editor object from the React context.
+Get the current editor object from the React context. Differs from `useSlate` in that this is a context for sharing the editor object in a way that does **not** re-render the context whenever changes occur. To be used in most cases as `useSlate` can cause unnecessary re-renders of components that consume it.
 
 ###### `useFocused`
 
@@ -52,7 +52,7 @@ Get the current `selected` state of an element.
 
 ###### `useSlate`
 
-Get the current editor object from the React context.
+Get the current editor object from the React context. Differs from `useEditor` in that this is a context for sharing the editor object in a way that re-renders the context whenever changes occur. To be used if you need the component consuming it to re-render on `onChange`.
 
 ## ReactEditor
 
--- a/docs/libraries/slate-react.md
+++ b/docs/libraries/slate-react.md
@@ -36,7 +36,7 @@ React hooks for Slate editors
 
 ###### `useEditor`
 
-Get the current editor object from the React context. Differs from `useSlate` in that this is a context for sharing the editor object in a way that does **not** re-render the context whenever changes occur. To be used in most cases as `useSlate` can cause unnecessary re-renders of components that consume it.
+Get the current editor object from the React context. Differs from `useSlate` in that this is a context for sharing the editor object in a way that does **not** re-render the context whenever changes occur. Consider using `useEditor` if you do not explicitly need to trigger a re-render of the consuming component as it can result in improved performance.
 
 ###### `useFocused`
 
@@ -137,4 +137,3 @@ Adds React and DOM specific behaviors to the editor.
 ## Utils
 
 Private convenience modules
-
--- a/docs/libraries/slate-react.md
+++ b/docs/libraries/slate-react.md
@@ -52,7 +52,7 @@ Get the current `selected` state of an element.
 
 ###### `useSlate`
 
-Get the current editor object from the React context. Differs from `useEditor` in that this is a context for sharing the editor object in a way that re-renders the context whenever changes occur. To be used if you need the component consuming it to re-render on `onChange`.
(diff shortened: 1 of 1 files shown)


docs/libraries/slate-react.md
```diff
@@ -36,7 +36,7 @@ React hooks for Slate editors
 
 ###### `useEditor`
 
-Get the current editor object from the React context.
+Get the current editor object from the React context. Differs from `useSlate` in that this is a context for sharing the editor object in a way that does **not** re-render the context whenever changes occur. To be used in most cases as `useSlate` can cause unnecessary re-renders of components that consume it.
```
kamilmielnik, line 39: > To be used in most cases as `useSlate` can cause unnecessary re-renders of components that consume it.

Huh, in our project we've decided to default to `useSlate` as the lack of that extra re-render has caused multiple unexpected results/issues for us. So we treat `useEditor` as an optimization and we actually never use it as `useSlate` is fast enough (and solves easy-to-miss problems).

docs/libraries/slate-react.md
```diff
@@ -52,7 +52,7 @@ Get the current `selected` state of an element.
 
 ###### `useSlate`
 
-Get the current editor object from the React context.
+Get the current editor object from the React context. Differs from `useEditor` in that this is a context for sharing the editor object in a way that re-renders the context whenever changes occur. To be used if you need the component consuming it to re-render on `onChange`.
```
kamilmielnik, line 55: > To be used if you need the component consuming it to re-render on `onChange`.

Not only `onChange`, also for computations based on `editor.selection` value.

docs/libraries/slate-react.md
irmerk: This was pulled from our experience, as well as the discussion in this thread: https://slate-js.slack.com/archives/C1RH7AXSS/p1588329120389200

> What is the difference between `useEditor` and `useSlate` ?
The comment says that `useSlate` will share the editor in a way that 'the context re-renders whenever changes occur'.
Does this mean that whenever the slate Nodes change the context will re-render. And does this imply that the context will never re-render because of changes to the Nodes when using `useEditor` instead?

> `useSlate` will cause whatever component consumes it to re-render because of the context re-rerender.  `useEditor` provides you the latest `editor`, `editor.children`, etc. but does not cause a re-render of that component.
> We ended up using `useEditor` in most cases as `useSlate` was causing unecessary re-renders of components that consumed it.  But if you need the component to re-render on `onChange`, use `useSlate`.

> Thanks @brentfarese for your answer. So the `useSlate` context will re-render whenever the internal editor state changes, right? That means for example if a Text node is changed, do I get this right?

> Yes but you can get that information with `useEditor` too.
>The important thing is that `useSlate` will cause your entire component to re-render even if that component does not need to be (e.g., the data the component consumes has not changed).

docs/libraries/slate-react.md
irmerk: Added in ad89288

docs/libraries/slate-react.md
kamilmielnik: I understand your point.
Still, I'd refrain from this assumption:

> To be used in **most cases** 

as in our project it was the other way around. I think it's not a 0/1 statement.

To rephrase my initial comment: `useSlate` is _safer_ than `useEditor` - it can prevent many issues that one would experience with `useEditor`. I think `useEditor` should be treated as performance optimization (after all, as already mentioned, it helps with _unnecessary re-renders_). 

Even in a very simple, common example, `useSlate` is the way to go: https://github.com/ianstormtaylor/slate/blob/912d4b79da5bfafeaa955d1abdddc358c7220197/site/examples/richtext.tsx#L139-L142

----

To sum up, I'd say that if unsure, one should go with `useSlate` (unless it causes performance issues - but of course, premature performance optimization is the roof evil, bla bla bla, so one should default to the safer solution). 

----

Anyway, thanks for taking the time to explain this, because we did wonder for a bit :)

BrentFarese, line 39:
```suggestion
Get the current editor object from the React context. Differs from `useSlate` in that this is a context for sharing the editor object in a way that does **not** re-render the context whenever changes occur. Consider using `useEditor` if you do not explicitly need to trigger a re-render of the consuming component as it can result in improved performance.
```

BrentFarese: I added a minor suggestion.  Our experience is that we use `useEditor` almost everywhere and not `useSlate` but our setup is different.  I agree `useEditor` should be used if you care about performance, and if your app is performant enough and `useSlate` hides some oddities, then just use `useSlate`.  Let me know if the change is good.  Thanks for the PR!

irmerk: Great points!

ianstormtaylor:
I think this is obsolete now with the name change of `useEditor`. Feel free to reply if not!

Issue #3917, opened 2020-10-12
Using Slate inside iframe


greenxiii:
Bug
When you're using slate inside iframe you cant handle any events with custom handlers.
As I researched its caused by `isDOMNode(target)` function because you using `instanceof` to compare if target is Node. Since every environment (every iframe, frame, window) has it's own internal classes for Node, Array, Object, etc. it will be false in this cases so it makes slate pretty useless after update to 0.59.0 version

Slate: 0.59.0
Browser: Chrome
OS: MacOS

rndm2:
Any updates here?

greenxiii:
Up

rndm2:
@CameronAckermanSEL could you please take a look on this issue? We are created PR to fix this also

volkandkaya:
Was looking at the slate docs and was so impressed. Then forgot it doesn't work in an iframe.

rndm2:
We dropped slate usage and currently replacing it with draft js

volkandkaya:
@rndm2 i built a custom contenteditable for one use case, using draftjs for another use case.

Slate looks so good though :( 

Working with draftjs is a huge pain for me.

PR #3922, opened 2020-10-15
[#3917] Add to isDOMNode function check for iframe implementation

greenxiii:
Fixing a _bug_ #3917

#### What's the new behavior?

Now when you're using slate inside iframe there are will not be isDOMNode will not always return false and you can attach event handlers. The check didn't work before because iframe has its own internal Node and it's not equal to parent Node.

Checked with `yarn test`, `yarn lint` and `yarn start`

Fixes: #3917


dlin-me:
Please, I need this fixed

dlin-me:
Please look at this https://github.com/ckeditor/ckeditor5-utils/pull/208/files

dylans:

Issue #3198, opened 2019-12-02
Slate docs say to use useMemo, should be useRef


jas7457:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
In the docs, it says to use useMemo for the editor instance passed to the `<Slate>` component so it is "stable across renders". useMemo does not guarantee identity between renders (https://reactjs.org/docs/hooks-reference.html#usememo) and talking with @ianstormtaylor, it sounds like identity is needed. between renders.

Slate: 0.50.7

#### What's the expected behavior?
The docs should be updated to use a ref (or some other stable memo besides useMemo/useCallback/etc). Slate itself could memoize the value passed to it if identity is necessary, or inversion of control could be maintained for the user of Slate to pass a memoized version.

#### Note:
A simple hook that I wrote can help to initialize a ref if it's expensive to compute. It works similar to useState's lazy initialization, but there wasn't a way to lazily set a ref.

import {useRef, MutableRefObject} from 'react';

```javascript
// using an object here instead of something like "null" or "undefined" or "Infinity"
// that way we truly know if we need to call the init function or not because the user of this can never pass something that === initialValue
const initialValue = {};

/**
 * Allows for lazy refs that only call the callback once
 * This also helps guarantee values between renders, unlike useMemo or useCallback which makes no such guarantee
 */
export default function useLazyRef<T>( init:() => T ):MutableRefObject<T> {
	const ref = useRef<T | typeof initialValue>( initialValue );

	if ( ref.current === initialValue ) {
		ref.current = init();
	}

	return ref as MutableRefObject<T>;
}

```

ianstormtaylor:
Seems like we can use https://github.com/alexreardon/use-memo-one and maybe expose it as a `useStable` hook for people. Since it does seem to be asking a lot if people have to figure out the stability for themselves.

PR #3925, opened 2020-10-21
Add useStable hook to slate-react and update docs.

Enva2712:
#### Is this adding or improving a feature or fixing a bug?
This PR implements #3198 by adding a `useStable` hook to the `slate-react` package and updating the documentation to use it in place of `useMemo` for persisting the editor state; as `useMemo` doesn't guarantee persistence of the memoized value.

#### What's the new behavior?
From the updated docs:
```jsx
// Import React dependencies.
import React, { useEffect, useState } from "react";
// Import the Slate editor factory.

// Import the Slate components and React plugin.
import { Slate, Editable, withReact, useStable } from 'slate-react'

const App = () => {
  // Create a Slate editor object that won't change across renders.
  const editor = useStable(() => withReact(createEditor()))
  return null
}

```

The `useStable` hook uses a react ref to store the editor, and initializes on the first call using the function passed by the user.

Fixes: #3198

### DIFF
--- a/packages/slate-react/src/hooks/use-stable.ts
+++ b/packages/slate-react/src/hooks/use-stable.ts
@@ -7,7 +7,7 @@ const UNINITIALIZED = {}
 const isUninitialized = (v: unknown): v is typeof UNINITIALIZED =>
   v === UNINITIALIZED
 
-export function useStable<T>(init: () => T): T {
+export const useStable = <T>(init: () => T): T => {
   const ref = useRef<T | typeof UNINITIALIZED>(UNINITIALIZED)
   // If the ref hasn't been initialized, initialize it; else don't. then return the initialized value
   // This is written as a single expression to help typescript understand what's going on.
--- a/docs/concepts/08-rendering.md
+++ b/docs/concepts/08-rendering.md
@@ -10,10 +10,10 @@ For example if you wanted to render custom element components, you'd pass in the
 
 ```jsx
 import { createEditor } from 'slate'
-import { Slate, Editable, withReact } from 'slate-react'
+import { Slate, Editable, withReact, useStable } from 'slate-react'
 
 const MyEditor = () => {
-  const editor = useMemo(() => withReact(createEditor()), [])
+  const editor = useStable(() => withReact(createEditor()))
   const renderElement = useCallback(({ attributes, children, element }) => {
     switch (element.type) {
       case 'quote':
@@ -115,7 +115,7 @@ A common use case for this is rendering a toolbar with formatting buttons that a
 
 ```jsx
 const MyEditor = () => {
-  const editor = useMemo(() => withReact(createEditor()), [])
+  const editor = useStable(() => withReact(createEditor()))
   return (
     <Slate editor={editor}>
       <Toolbar />
--- a/docs/walkthroughs/01-installing-slate.md
+++ b/docs/walkthroughs/01-installing-slate.md
@@ -19,12 +19,12 @@ Once you've installed Slate, you'll need to import it.
 
 ```jsx
 // Import React dependencies.
-import React, { useEffect, useMemo, useState } from "react";
+import React, { useEffect, useState } from "react";
 // Import the Slate editor factory.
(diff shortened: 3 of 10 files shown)


kripod:
Great idea!

Have you considered reusing the hook from [use-constant](https://github.com/Andarist/use-constant)? (or even using it directly)

packages/slate-react/src/hooks/use-stable.ts
```diff
@@ -0,0 +1,15 @@
...
+// This helper function is required to help typescript track the type of ref.current below
+const isUninitialized = (v: unknown): v is typeof UNINITIALIZED =>
+  v === UNINITIALIZED
+
+export const useStable = <T>(init: () => T): T => {
+  const ref = useRef<T | typeof UNINITIALIZED>(UNINITIALIZED)
+  // If the ref hasn't been initialized, initialize it; else don't. then return the initialized value
+  // This is written as a single expression to help typescript understand what's going on.
+  return isUninitialized(ref.current) ? (ref.current = init()) : ref.current
+}
```
BrentFarese, line 15: I don't see why we have to use `UNINITIALIZED`.  Is it needed?  Can we just do:

```suggestion
export const useStable = <T extends Editor>(init: () => T): T => {
  const ref = useRef<T | null>(null)
  
  if (!ref.current) ref.current = init();
  
  return ref.current;
}
```

Enva2712: For the slate use case it's effectively the same; but if a slate user also uses the `useStable` hook for other values, it's possible that their initialization function returns a falsey value, causing it to re-run each time the hook is called & breaking the contract of only running a single time. Add side effects & there's a hard to find bug.

Because this is a helper being exported to users to make maintaining the editor state reference easier, and not just an internal helper, it should uphold the contract, & not lead to hard to debug errors when used in other ways

BrentFarese: Ah I see what you are saying.  Do we need to expose a general hook?  The use case here is to ensure referential equality on the `editor`, no?  I might suggest limiting its use and calling it `useStableEditor` or the like as it's not really the province of Slate to expose a general hook.

ianstormtaylor:
Dunno if this has been investigated, but could the `editor` constant be pulled outside the rendering layer entirely and just defined at the top-level of a file? If so, we could eliminate the need for a hook at all. And people can still use `useConstant` if they have more complex setups.

Enva2712:
@ianstormtaylor I hadn't thought of that, but it seems to make the most sense. Not only is it the pattern other libraries like redux use, it seperates the places where `slate` and `slate-react` packages are used, keeping the view layer seperate from the editor logic.

If everyone thinks that's the way to go, I can close this PR and create one for updating the docs in the next week or so

BrentFarese:
I think this would be totally fine/maybe better.  If the point is reference equality on the `editor` object, makes sense.  We don't use `slate-react` but would adopt a similar approach in our rendering layer as well.  Thanks for the upcoming PR @Enva2712!  I will close this then for now.

ArgonAlex:
That wouldn't work if you want to create the editor in a component that you can have more than one instance of.

I've typically used useState without a setter for stuff like this in the past.
```ts
const [editor] = useState(() => withReact(createEditor()));
```

Tokimon:
why not just use a `useMemo`?

```js
const editor = useMemo(() => withReact(createEditor()));
```

it would do the same wouldn't it?

dylans:
> why not just use a useMemo?

You can, but React warns against it for the future: https://reactjs.org/docs/hooks-reference.html#usememo

> You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.

Tokimon:
Well that sucks, that we cannot use `useMemo` like this anymore. But thank you for the clarification.

Issue #3923, opened 2020-10-15
Text highlighting and setting marks is broken


remingtonlang:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
If you double click on a word, set "Bold", "Italic", or any other mark. Navigate away to another character and then go back to the marked word with double click, and undo the prior mark set, it will unset the mark but enable the mark for surrounding marks. Reproduce:

1.  Go to: https://www.slatejs.org/examples/richtext
2. Double click "things" in the second paragraph; highlighting the word and click "B"(Bold)
3. Arrow away clearing the highlight
4. Double click "things" again.. highlighting it. The mark button is no longer highlighted despite it being bold.
5. Click Bold and the entire sentence is highlighted bold.
(Sorry I dont know how y'all are creating these fancy gif screens)

Slate: Whatever https://www.slatejs.org/examples/richtext is currently running
Browser Firefox Developer Edition (82.0b8 (64-bit))
OS: Mac 11.0 Beta

#### What's the expected behavior?
Highlighting the word should correctly set the mark, toggling the button should likewise work

vitorrd:
Couldn't reproduce it on Windows on Chrome 86.0.4240.75.

IsaacMSchultz:
I was able to reproduce this issue. Digging into the code, it seems it is related to the way slate selects what DOM nodes the user is highlighting using `window.getSelection()`.

# The bug
I was able to discover that after making a mark change, the selection is properly updated by examining the anchor and focus nodes of `window.getSelection()`:
![image](https://user-images.githubusercontent.com/44488203/96940962-0101cb00-1486-11eb-811a-68aefbb7676b.png)
Also note that the offsets count the correct number of characters after the anchor node.

But when the text is re-selected, the anchor node is set to the previous node (which I believe is correct), but the focus node is set to the node that comes after the actual selected node, and thus the offset of the focus node is 0, which is the case that is causing the bug.
![image](https://user-images.githubusercontent.com/44488203/96941076-5c33bd80-1486-11eb-8d28-b50aca1ba5c8.png)

# Solution
I believe this is fixable by checking for this case in the `toSlateRange()` function. I will start working on a solution for it.

BrentFarese:
Is this issue caused by #3756?  I am not sure this is a selection bug.  Perhaps it is a `slate-history` bug with how the history resolves the selection (in this case, it might be incorrect which causes the formatting bug)?

[...]

PR #3931, opened 2020-10-25
Fix: Text selection selecting incorrect slate nodes

IsaacMSchultz:

Fixing a bug

#### What's the new behavior?

Previously, the `toSlateRange` function in `react-editor.ts` did not check for the case of a selection being on the end of one node into another node. This causes issues for functions relying on the slate range, since they expect the anchor node to be the start of the selection. These two selections are semantically the same, but represented differently in the slate Point objects returned by `toSlateRange` There is a case when adding marks to text, updates would not happen correctly and the previous node's text would be updated instead (See #3923).

The change I made simply checks for this case, and if the beginning of a text selection is ***exactly*** at the end of the previous node, it sets the anchor to be the first entry in the next node.

the `toSlateRange` function will now check if the anchor node selection is at the end of the node, if it is, then it will move it to the start of the next node.

Fixes: #3923


packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -518,6 +518,23 @@ export const ReactEditor = {
...
 
+    // If the selection is at the very end of the anchor node, when the anchor and focus are not the same node.
+    if (
+      !isCollapsed &&
+      anchorNode !== focusNode &&
+      anchorNode.nodeValue != null &&
+      anchorNode.nodeValue.length === anchorOffset
+    ) {
+      // Get the next node
+      const newNodeEntry = Editor.next(editor, { at: anchor.path })
```
BrentFarese, line 529: I am not sure this is correct.  Shouldn't you be getting the next `Text` node, in which case I would consider passing a `match: Text.isText`?

BrentFarese:
I am not sure this is the correct fix.  Really, we should fix `Transforms.setNodes` to not apply properties to the node where the selection is at offset: 0, unless the caller of the function specifies that node should be included.

Related to #4038.

dylans:
Superseded by #4168.

Issue #3932, opened 2020-10-25
Adding images through image button doesn't validate image URL


kenzobanaag:
#### Do you want to request a _feature_ or report a _bug_?

Feature request

#### What's the current behavior?

For Image Example

Adding images from the image button does not check if the URL is an actual image while the paste functionality does.

![image_validation](https://user-images.githubusercontent.com/25941773/97097277-11ec4100-162c-11eb-8529-d77f8d581f3c.gif)

#### What's the expected behavior?

It doesn't break the code, but it would be nice to use the same functionality that is already present to validate the URL. 

The button should use the same validation as the paste functionality does.
Could also show an alert when the URL is not an image.

PR #3933, opened 2020-10-25
Added isImageUrl validation to button, added alert when not image

kenzobanaag:

Adding/improving a feature

#### What's the new behavior?

image_validation_fix

The new change checks if the URL provided is an image URL using the present function isImageUrl

Also shows an alert when the URL is not an image. 

Fixes:  #3932


ianstormtaylor:
Thanks @kenzobanaag!

Issue #3935, opened 2020-10-25
Paste Image URL functionality does not work on images from unsplash.com


kenzobanaag:
#### Do you want to request a _feature_ or report a _bug_?

_Bug_

#### What's the current behavior?

Pasting an image URL of pictures from unsplash.com does not insert the picture but pastes its URL.

![paste_url](https://user-images.githubusercontent.com/25941773/97098964-51bd2380-1640-11eb-9e9a-075587e0b5d9.gif)

[Image Example](https://www.slatejs.org/examples/images)

Slate: 0.59.0
Browser: Chrome
OS: Windows

#### What's the expected behavior?

Pasting the image URL should insert the actual image and not its URL.

dylans:
Closing this because it does not look like a current Slate bug.

This is an example limitation or environment issue, not a core bug.

If there is a minimal repro showing current Slate behavior is at fault, a narrower follow-up can be opened.

PR #3936, opened 2020-10-25
Paste URL works for image URLs that don't have image extensions

kenzobanaag:

Fixing a bug

#### What's the new behavior?

paste_url_fix

The fix checks the image URL if it includes an image header.

The recent isImageUrl function only checks the URL if it contains any of the image extensions from the [image extension package](https://www.npmjs.com/package/image-extensions). The problem was that images from unsplash.com do not include image extensions (eg. jpeg, png) in their image URLs, so the function evaluates all images from unsplash.com as not image URLs. 

The change builds off of isImageUrl and adds a last resort check.
It makes a get request to obtain the Content-Type of the URL sent using XMLHttpRequest.
If the response header has a content-type of image then it must be an image, so it will return true in this case. 

Also, I was running into some CORS Blocking issue so I had to use a proxy URL to bypass the CORS issue.

Fixes:  #3935

### DIFF
--- a/site/examples/images.tsx
+++ b/site/examples/images.tsx
@@ -130,15 +130,9 @@ const isImageUrl = url => {
   const ext = new URL(url).pathname.split('.').pop()
   if (imageExtensions.includes(ext)) return true // catch when url has image extension
   // last resort to check if url has an image content type
-  return nonAsyncIsHeaderImage(url)
-}
-
-const nonAsyncIsHeaderImage = url => {
-  const proxyurl = 'https://cors-anywhere.herokuapp.com/' // need proxyurl to bypass CORS
-  const req = new XMLHttpRequest()
-  req.open('GET', proxyurl + url, false)
-  req.send(null) // will fire CORS error if no proxy is used
-  return req.getResponseHeader('Content-Type').includes('image')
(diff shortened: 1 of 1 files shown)


CameronAckermanSEL:
Thank you very much for the contribution. I'm a little hesitant to accept this because it seems a little out of scope with the minimal example we're seeking to have and it deals with a CORS issue by delegating to heroku. 

Interested in how other contributors feel about the PR.

styxlab:
How about something along these lines instead?
```
  if(typeof ext === 'string' && imageExtensions.includes(ext)) return true
  const response = await fetch(url)
  if(response.ok && response.headers.get('Content-Type')?.includes('image')) return true
```
This works for me without running into CORS issues.

dylans (commented):
I've updated this PR based on feedback provided. If everyone is happy with this change I'll land it.

cameracker:
Hey folks, I'm going to decline this PR. I appreciate the submission and the discussion around it, but ultimately this kind of error handling is out of scope for an example, and like most examples this is going to wind up getting copy-pastad all over the place. 

It'll be here for posterity if someone runs into the same problem and needs this solution. 

Thanks!

Issue #3934, opened 2020-10-25
No tests for slate-react source code


IsaacMSchultz:
#### Do you want to request a _feature_ or report a _bug_?

request a feature

#### What's the current behavior?

I am noticing that there is a lack of test cases for the whole slate-react package. In my limited time working with the codebase, I am noticing that there are a lot of issues that arise from the code in this package that do not get discovered from tests. Most of these issues are things that could be mitigated or caught earlier by deploying more test cases in this package.

Slate: 0.59.0
Browser: Firefox
OS: Windows

#### What's the expected behavior?

Potentially introducing more tests specifically for this package, as well as more frameworks for testing. I notice that the project uses mocha for unit tests, but it does not have an assertion framework or a mocking framework, both of which could prove to be extremely useful for the maintainability of the project.

These tests do not need to be written all at once, but could be instead added as part of future PRs for the project.

IsaacMSchultz:
# Example improved testing
I have created a PR #3953 that provides an example of a possible introduction of more unit tests. This PR builds on #3931 and #3937, which have not been reviewed yet, so I don't expect it to get merged in its current state.

# Refactoring for testability
Testing the User Experience (UX) requires deep mocking of most of the functions involved, this required a switch to a different testing framework to allow for in-depth mocking (jest instead of mocha), as well as code re-factoring to move any code that may not be mock-able to its own function, so that specific functionality can be mocked. I proposed the possible switch to mocha from jest in #3937, and I have written tests for the code changes implemented in #3931. The function modified in #3931, `toSlateRange` is not easily testable because of the use of a `Selection` object, which as far as I understand is a part of the browser and cannot be mocked. For this reason, I moved that code out of the `toSlateRange` function into its own separate function so it's return value can be mocked.

Additionally, the test runner code was encountering an error with some of the import statements in slate-react's `dom.ts` file in the form `import DOMNode = globalThis.Node` when running tests with the test runner, I was unable to find a way to get babel to recognize this, so I resorted to renaming the variables `Node` to `SlateNode` and `Range` to `SlateRange` as a possible fix instead of trying to distinguish between those names and the global names. There is almost certainly a better solution than this.

PR #3937, opened 2020-10-26
Migrating from mocha to jest for testing

IsaacMSchultz:

Improving a feature (testing)

#### What's the new behavior?

Now using jest for testing. This allows for the use of features not available in mocha such as jest's better assertions, and mocking.

I changed the repository to use Jest instead of Mocha for testing. I had to modify a number of configuration files and I am still trying to figure out what is required vs what is not.

Fixes: #3934
Reviewers: @

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,0 +1,14 @@
+{
+    "presets": [
+        "@babel/preset-typescript",
+        [
+            "@babel/preset-env",
+            {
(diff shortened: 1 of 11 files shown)


dylans:
I don't think there's consensus to switch testing frameworks currently. Feel free to open an issue for discussion.

Issue #3797, opened 2020-07-23
Adding option to attach files


mdmjg:
#### Do you want to request a _feature_ or report a _bug_?

feature

#### What's the current behavior?
Currently, when you want to attach an image (like in the Image example), you are prompted with the URL of the image. I want to add the ability to select a file from your computer. 

Slate: 0.58.4
Browser: Chrome 
OS: Mac 

#### What's the expected behavior?
User should be able to click or drag a file into the editor, as long as having the nodes wrap around such file as needed.

CameronAckermanSEL:
Hi @mdmjg , thanks for the report. For clarity, are you asking for this to be added to the image example?

PR #3938, opened 2020-10-26
Need help with useEditor hook when adding an image from file explorer

kbrandonle:
Adding a feature

#### What's the new behavior?
A button now exists that allows a user to upload a file from their computer.

The form fires an onChange event when the user opens a file. That file is then run through the File Reader to get the image URL. What I cant figure out is how to call insertImage to insert the imageURL into the editor. When I try to create a hook for useEditor(), React complains that I broke a rule. Possibly because the hook is already being used in the Button component?

Fixes: #3797


dylans:
Closing this one as it's stale and doesn't seem to achieve the desired result of improving the image upload example. Feel free to re-open if I am misunderstanding.

Issue #3863, opened 2020-09-07
Cannot select text within void nodes


davidruisinger:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

- Open [Editable Voids example](https://www.slatejs.org/examples/editable-voids)
- Try selecting one of the labels using the mouse
- Selection is lost right away when releasing the mouse

![jPqQOxFnKu](https://user-images.githubusercontent.com/5305150/92390755-ca613280-f11b-11ea-92ba-505d1a868908.gif)

Slate: 0.58.4
Browser: Chrome & Safari
OS: Mac

#### What's the expected behavior?

Selection should stay

BrentFarese:
Can you add detail as to why you think it's needed to be able to select content inside a `void` node like that?  Does this issue have to do with the fact that the `EditableVoid` in the example has `contenteditable={false}`.  The selection you are referring to is actually not Slate's selection, but is the browser's selection for normal HTML.  I agree it's somewhat odd that the browser selection becomes unset on release, but why do you need to "select" that content to begin with?  As mentioned, the `void` node is not an editable element like other elements in Slate.

Thanks!

flavordaaave:
@BrentFarese In our application we use void elements for e.g. images and we render  render some additional info within the void elements that are esential to our editors in their daily workflow. So in case they added an image from our system to their written article they probably need some other info on that image later on in their article (e.g. a part of the description).

To illustrate what I mean I created a small example based on the "official" images example where you can show additional info on an image using the button underneath. Trying to copy a fraction of the text from the infobox is not working good because of the issue stated above.

[codesandbox](https://codesandbox.io/s/slate-void-selection-issue-n7ubr?file=/src/examples/images.tsx)

ryanmitts:
From having issues with this kind of thing for a couple years, it seems there are two conflicting ideas for what a void node is, and what isn't, and how they should work. It should probably be documented or described in the docs how to handle both cases? I could probably write that up, but have never been sure what the expectation is of a void node to know if I'm capturing the intent.

The two cases seem to be:

1. Things like images and videos, where you want them to be selected as a block, but otherwise behave as not editable.
2. Things that have interactions or text inside that you want completely unmanaged by the parent editor (like the Editable Void example, which I actually contributed).

For the case described, it's because of this logic on the `onClick` handler, [here](https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/src/components/editable.tsx#L598).

[...]

PR #3939, opened 2020-10-26
added html properties to void nodes

kbrandonle:
Fixing a bug

#### What's the new behavior?
Allows a user to select text in a void node.

An inefficient way to allow a user to select text in void nodes that inserts a property into the html tag. A proper fix would probably be to normalize all void nodes, but I can't figure out how.

Fixes: #3863

### DIFF
--- a/site/examples/editable-voids.tsx
+++ b/site/examples/editable-voids.tsx
@@ -28,6 +28,7 @@ const EditableVoidsExample = () => {
   )
 }
 
+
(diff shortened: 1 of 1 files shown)


dylans:
Given that this is stale and incomplete I'm going to close this out. @kbrandonle if you have interest in working on this, feel free to reach on the Slate Slack channel to discuss.

Issue #3849, opened 2020-08-30
useSlate vs. useEditor hook triggering behaviour unclear


aeplay:
#### Do you want to request a _feature_ or report a _bug_?

Naming/documentation improvement.

#### What's the current behavior?

From looking at the code and observed behavior:

- `useEditor` returns the editor instance once, but doesn't re-trigger when the editor's value changes.
- `useSlate` returns the editor instance and re-triggers every time the editor's value or selection changes.

Both are useful for different purposes.

Slate-React: 0.58.0
Browser: All
OS: All

#### What's the expected behavior?

It would be great if the naming of the hooks could make the expected behavior more clear.
Both the current naming and the documentation led me to believe that they were aliases for each other.

PR #3941, opened 2020-10-27
Differentiate between useSlate and useEditor hooks

songeunyou:
**Is this adding or improving a feature or fixing a bug?**
Naming/documentation improvement

**What's the new behavior?**
Renamed `useEditor` hook to `useSlateStatic` to indicate that the editor object does not re-render when changes occur. Updated documentation to differentiate between `useSlate` and `useSlateStatic`.

**How does this change work?**
`useSlateStatic` should be the hook that is used instead of `useEditor`.

**Have you checked that...?**
- [x]  The new code matches the existing patterns and styles.
- [x]  The tests pass with yarn test.
- [x]  The linter passes with yarn lint. (Fix errors with yarn fix.)
- [x]  The relevant examples still work. (Run examples with yarn start.)

Does this fix any issues or need any specific reviewers?
Fixes: [#3849](https://github.com/ianstormtaylor/slate/issues/3849)

### DIFF
--- a/docs/libraries/slate-react.md
+++ b/docs/libraries/slate-react.md
@@ -52,7 +52,7 @@ Get the current editor object from the React context. Re-renders the context whe
 
 ###### `useSlateStatic`
 
-Get the current editor object from the React context. A version of useSlate that does not re-render the context.
+Get the current editor object from the React context. A version of useSlate that does not re-render the context. Previously called `useEditor`.
 
 ## ReactEditor
 
--- a/docs/libraries/slate-react.md
+++ b/docs/libraries/slate-react.md
@@ -18,11 +18,11 @@ React components for rendering Slate editors
 
 The main Slate editor.
 
-###### `DefaultElement(props: RenderElementProps)` 
+###### `DefaultElement(props: RenderElementProps)`
 
 The default element renderer.
 
-###### `DefaultLeaf(props: RenderLeafProps)` 
+###### `DefaultLeaf(props: RenderLeafProps)`
 
 The default custom leaf renderer.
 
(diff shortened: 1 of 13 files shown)


docs/libraries/slate-react.md
```diff
@@ -52,7 +48,11 @@ Get the current `selected` state of an element.
 
 ###### `useSlate`
 
-Get the current editor object from the React context.
+Get the current editor object from the React context. Re-renders the context whenever changes occur in the editor.
+
+###### `useSlateStatic`
+
+Get the current editor object from the React context. A version of useSlate that does not re-render the context.
```
timbuckley, line 55: Please add a little note here that it used to be called `useEditor`

packages/slate-react/src/hooks/use-editor.tsx
```diff
@@ -2,6 +2,10 @@ import { createContext, useContext } from 'react'
 
 import { ReactEditor } from '../plugin/react-editor'
 
+/**
+ * @deprecated Use useSlateStatic instead.
+ */
+ 
 /**
  * A React context for sharing the editor object.
```
timbuckley, line 10: The `@deprecated` flag should move just below ^this line here, in the same comment section. The above error "Delete `·`" means it doesn't like that there's a comment section that's not directly above a function or something like that.

timbuckley (approved):
LGTM!

remingtonlang:
Is this a duplicate of #3925?

timbuckley:
@remingtonlang No they are different. This PR renames existing `useEditor` to `useSlateStatic` (keeping the old name available in the API). That PR #3925 is about creating a slate-specific replacement for `useMemo`.

PR #3946, opened 2020-10-28
Disable spellCheck/autoCorrect/autoCapitalize in browsers without onBeforeInput

hanford:
Currently you're unable to set spellCheck/autoCorrect/autoCapitalize attributes in browsers without support for the `onBeforeInput` event.

That's ok, because it's supposed to fix some buggy behavior, but `spellcheck` is enabled by default in firefox with no way of disabling it.

This PR fixes that, by explicitly setting these attributes to false, rather than undefined.


ianstormtaylor:
Thanks @hanford!

PR #3953, opened 2020-10-30
Example changes from updated testing

IsaacMSchultz:

Improving a _feature_ (Technically just for the devs because more testing makes it easier to find bugs 😊)

#### What's the new behavior?

In this pull request I am demoing the code changes that may be needed to implement improved testing for slate. This can greatly help with finding bugs early, and make future development easier by removing most of the manual testing steps. This is a draft PR that I don't expect to get merged (at least in its current state) since it builds on several code changes that are still awaiting review (#3937, #3931).

Testing the User Experience (UX) requires deep mocking of most of the functions involved, this required a switch to a different testing framework to allow for in-depth mocking (jest instead of mocha), as well as code re-factoring to move any code that may not be mock-able to its own function, so that specific functionality can be mocked. I proposed the possible switch to mocha from jest in #3937, and I have written tests for the code changes implemented in #3931. The function modified in #3931, `toSlateRange` is not easily testable because of the use of a `Selection` object, which as far as I understand is a part of the browser and cannot be mocked. For this reason, I moved that code out of the `toSlateRange` function into its own separate function so it's return value can be mocked.

Additionally, the test runner code was encountering an error with some of the import statements in slate-react's `dom.ts` file in the form `import DOMNode = globalThis.Node` when running tests with the test runner, I was unable to find a way to get babel to recognize this, so I resorted to renaming the variables `Node` to `SlateNode` and `Range` to `SlateRange` as a possible fix instead of trying to distinguish between those names and the global names. There is almost certainly a better solution than this.

Partially addresses: #3934
Reviewers: @


dylans:
@IsaacMSchultz there are definite improvements in here, but such a massive set of changes at once is difficult to review and can quickly get stale.

Given that this PR has stalled for almost a year, I'm going to close it out, but I encourage you to raise smaller PRs to address things you've identified to improve the typing and testability of Slate.

Issue #3770, opened 2020-07-08
Rename Master branch to Main


uipoptart:
#### Do you want to request a _feature_ or report a _bug_?
Neither?

Github has announced that they will be renaming the primary branch from "Master" to "Main". Opening this for Slate to do the same. This link does a better job that I can of explaining the why:
https://www.hanselman.com/blog/EasilyRenameYourGitDefaultBranchFromMasterToMain.aspx

My 2 cents:
It takes a few moments to run the terminal command and change the repository setting. We can what-if over the actual origins or if everyone decides later it doesn’t matter but the time spent over the hand wringing is less than the two terminal commands it would take to just change it and move on. 

The Slack thread:
https://slate-js.slack.com/archives/C1RH7AXSS/p1594073983171300

Thanks!

tograd:
First they came for my master branch, and I said nothing..

ianstormtaylor:
Hey @uipoptart thanks for opening this. I've been busy and hadn't checked any Slate-related notifications for many months, so I totally missed this issue. 

I agree that this is a great change to make. Not only does it eliminate a word that can bring up unnecessary stress for some people, it's also unambiguously clearer in terms of DX for people just getting into coding. I remember being confused as to what "master branch" meant when I first started using Git—"main branch" would have been one less piece of jargon to learn. Fits well with the Slate mindset of not being shy to make breaking changes for even minor improvements in DX.

ianstormtaylor:
I've just run the commands and switched the setting on GitHub, so we should be good to go. Thanks!

TheSpyder:
I appreciate the enthusiasm, but github hasn't yet released their tooling to make this seamless. All open PRs still target `master`. I recommend reversing this until github makes it easy to change by retargeting PRs (and may even offer seamless updates for `git fetch`).
https://github.com/github/renaming/#later-this-year-seamless-move-for-existing-repositories-

PR #3957, opened 2020-10-30
Editor.next and Editor.previous now return the first match

tomdng:

Fixing a bug

#### What's the new behavior?

`Editor.next` and `Editor.previous` now return the **1st match** instead of the 2nd match.

This changes works by using `Editor.before` and `Editor.after` to find the point before/after the current position passed into `Editor.previous` and `Editor.next` respectively.

However, this implementation relies in a modification to `Editor.before`, `Editor.after`, `Editor.positions`, and `Editor.string` where all four functions now accept an optional voids parameter so that you can iterate void nodes instead of treating them as a single point. By default the option is set to false to match original behavior, but passing in `voids: true` to these functions should enable the new behavior that make this implementation work.

Finally I had to adjust two test cases because of a new issue I found with `Editor.before`. It throws an error it finds a location without a text node that exists before the current location. So using `Editor.before` to resolve `Editor.previous` caused two tests cases to throw an error.

This only occurs during two `moveNodes` test cases because in the process of moving a node out of the first location of the editor, the Editor hasn't fully normalized that location yet, so when `Editor.before` is called, it then calls `Editor.start` (which ultimately calls `Editor.point` with the `edge: 'start'` option) which then throws an error because there is no start text node.

I know that changing the test cases is not recommended, as it introduces different behavior, so would anyone have a better method to get `Editor.next` and `Editor.previous` to return the 1st match while passing the original test case? Otherwise we could expose another function like `Editor.before` and `Editor.after` that doesn't end up throwing an error when running on un-normalized nodes. I would appreciate any help with this!

Fixes: #3791 
Fixes: #3590

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -121,16 +121,21 @@ export const Editor = {
     options: {
       distance?: number
       unit?: 'offset' | 'character' | 'word' | 'line' | 'block'
+      voids?: boolean
     } = {}
   ): Point | undefined {
     const anchor = Editor.point(editor, at, { edge: 'end' })
     const focus = Editor.end(editor, [])
     const range = { anchor, focus }
-    const { distance = 1 } = options
+    const { distance = 1, voids = false } = options
     let d = 0
     let target
 
-    for (const p of Editor.positions(editor, { ...options, at: range })) {
+    for (const p of Editor.positions(editor, {
+      ...options,
+      at: range,
+      voids,
+    })) {
       if (d > distance) {
         break
       }
@@ -155,12 +160,13 @@ export const Editor = {
     options: {
       distance?: number
       unit?: 'offset' | 'character' | 'word' | 'line' | 'block'
+      voids?: boolean
     } = {}
   ): Point | undefined {
     const anchor = Editor.start(editor, [])
     const focus = Editor.point(editor, at, { edge: 'start' })
     const range = { anchor, focus }
-    const { distance = 1 } = options
+    const { distance = 1, voids = false } = options
     let d = 0
     let target
 
@@ -168,6 +174,7 @@ export const Editor = {
       ...options,
       at: range,
       reverse: true,
+      voids,
     })) {
       if (d > distance) {
         break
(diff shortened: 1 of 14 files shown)


packages/slate/src/interfaces/editor.ts
```diff
@@ -121,16 +121,21 @@ export const Editor = {
...
-    const { distance = 1 } = options
+    const { distance = 1, voids = false } = options
     let d = 0
     let target
 
-    for (const p of Editor.positions(editor, { ...options, at: range })) {
+    for (const p of Editor.positions(editor, {
+      ...options,
+      at: range,
+      voids,
```
BrentFarese, line 137: Since we are spreading `...options` on 135 we can get rid of 137.

```suggestion
```

packages/slate/src/interfaces/editor.ts
```diff
@@ -1107,8 +1126,18 @@ export const Editor = {
...
+
+    if (!from) {
+      return
+    }
+
     const [, to] = Editor.first(editor, [])
+
+    // The search location is from the start of the document to the path of
+    // the point before the location passed in
     const span: Span = [from, to]
```
BrentFarese, line 1141: Will have to change this line if you accept my suggestion above.

```suggestion
    const span: Span = [pointBeforeLocation.path, to]
```

BrentFarese (changes requested):
Looking good @tomdng.  I think 1 thing missing from this PR is that we should test the new behavior of `Editor.positions` with void nodes.  Can you write a test or 2 for that use case?  Similarly, we can consider a test case or 2 for `voids = true` for `Editor.before` and `Editor.after`, if we want test coverage to be solid.

Thanks!

tomdng:
Thank you @BrentFarese! I went ahead and added the test cases to `Editor.before`, `Editor.after`, and `Editor.position` with the `voids: true` option. I also went ahead and added an additional test case to `Editor.string` because that also has the new `voids: true` behavior.

Let me know if you want more/less test cases, especially for `Editor.positions` because it looks like there's a ton of test cases for the non void elements.

BrentFarese (approved):
Test coverage looks great!  Thanks @tomdng.

BobElDevil:
@BrentFarese  FYI, noticed this while putting up my own PR. Due to the changes done here https://github.com/ianstormtaylor/slate/issues/3770 PR's should now be targeted towards 'main' not 'master', so this PR got merged into the 'wrong' branch. Probably want to point all open PRs to the new branch and re-merge these changes

lukesmurray:
@BrentFarese string is missing the new parameter options in the editor interface which leads to typescript issues.

https://github.com/ianstormtaylor/slate/blob/db3f52517fb1e1dd81301fb084cfaddc5d73f62f/packages/slate/src/interfaces/editor.ts#L249

BobElDevil:
I actually opened up a PR for this recently https://github.com/ianstormtaylor/slate/pull/4009

lukesmurray:
@BobElDevil amazing thank you!

BrentFarese:
> I actually opened up a PR for this recently #4009

Wish I had seen this!  #4014 is a duplicate of this.  1 of the PRs will get merged and this will be fixed.  Thanks!

Issue #3422, opened 2020-01-11
Path.previous should not throw if there is no previous path 


macrozone:
#### Do you want to request a _feature_ or report a _bug_?

feature request (or flaw)

#### What's the current behavior?

`Path.previous(path)` throws an error when path is first path (e.g. ends with 0):

"Cannot get the previous path of a first child path ... because it would result in a negative index."

#### What's the expected behavior?

It should not throw and instead return null. This is much easier to work with, in particular with typescript as you would mark the result type as optional.

This is in particular critical because:

- there is no method Path.hasPrevious to guard this situation
- Path objects should be opaque i guess (e.g. its internal representation as array should not be any concern to devs), thats why a manual check `if(path[path.length - 1] > 0)` is not recommended.

Similarly, other Path functions should not throw.

LiuJi-Jim:
I agree. I think it would be better if `prevous/next/above` etc. return `null/undefined`.

Issue #3641, opened 2020-04-27
We need to talk about... Slate's Draconian :) error handling


mpkelly:
Hello, 

I would like to start a discussion about error handling in Slate which I think is too strict as Errors are often thrown for things that could just as easily be ignored or replaced by a warning. These are in functions like `toDOMPoint` or `toSlateRange` when Slate is trying to do things like map a `HTMLElement` to a `Node` but cannot because there is no direct mapping (the `HTMLElement` is created indirectly by a Slate `Node`). In many cases Slate will throw an `Error` that will take down the whole page. I understand that various attributes can be applied to have Slate ignore certain regions of the editor but I've still found it a bit of a fight and I really don't understand why Slate can't just return early in cases where no Dom -> Slate mapping exists. In fact, I have forked Slate and disabled these Errors ([here](https://github.com/mpkelly/slate/commit/213b1fd97d4edc8a1b179df77975bddb30f4a358) and [here](https://github.com/mpkelly/slate/commit/2f9ad08d3ee97125770d02e89b628b2d5133aad7)) and made Slate return early and while this fork is WIP, I can tell you that my app is much better for it - it's less work for me and I will still find problems where things aren't wired-up correctly and I no longer have anxiety of deploying to production and worrying about the WSOD when the user's only crime was clicking on an empty `<div/>` (I know prod apps should have error handling but again it feels like I am having to handle more errors than is really necessary). 

Some ideas:

- As above, log a warning like React does when in development mode and only use `Error` when it is absolutely necessary (cannot be recovered/ignored). 
- Maybe Slate could add a `onError` callback to its API for users to learn about errors

Thoughts?

CameronAckermanSEL:
Perhaps some error handling could be relaxed, but I personally don't have a strong opinion.

However, I can't help but wonder why people aren't catching the errors and letting the page crash?  https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html

mpkelly:
Hi @CameronAckermanSEL, I use `componentDIdCatch` in production as it is necessary whether using Slate or not. However, it feels like Slate is the source of _too many errors_ and that's the key point. There is also the development experience when first getting started to consider, too - here people might not have proper error handling in place and have to constantly restart/refresh/whatever due to errors for things they probably don't even care about. 

[...]

PR #3963, opened 2020-11-02
Add hasPrevious method to Path

mmdonaldson:

Improving a feature 

#### What's the new behavior?

A new method to address some of the concern in #3422

Add's a new method to the Path API called `hasPrevious` which checks if the path of previous sibling node exists

as per the `previous` API which returns a path, but returns an error if the path of the previous sibling does not exist, this is safe check you can make before calling `previous`

- [Y] The new code matches the existing patterns and styles.
- [Y] The tests pass with `yarn test`.
- [Y] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [Y] The relevant examples still work. (Run examples with `yarn start`.)

Fixes: #3422 
Reviewers: @

### DIFF
--- a/packages/slate/src/interfaces/path.ts
+++ b/packages/slate/src/interfaces/path.ts
@@ -123,7 +123,7 @@ export const Path = {
    */
 
   hasPrevious(path: Path): boolean {
-    return path.length !== 0 && path[path.length - 1] > 0
+    return path[path.length - 1] > 0
   },
 
   /**
--- a/packages/slate/src/interfaces/path.ts
+++ b/packages/slate/src/interfaces/path.ts
@@ -119,6 +119,14 @@ export const Path = {
   },
 
   /**
+   * Check if the path of previous sibling node exists
+   */
+
+  hasPrevious(path: Path): boolean {
+    return path.length !== 0 && path[path.length - 1] > 0
+  },
+
+  /**
    * Check if a path is after another.
    */
 
--- a/packages/slate/test/interfaces/Path/hasPrevious/root.tsx
+++ b/packages/slate/test/interfaces/Path/hasPrevious/root.tsx
@@ -4,4 +4,4 @@ export const input = [0, 0]
 export const test = path => {
   return Path.hasPrevious(path)
 }
-export const output = false;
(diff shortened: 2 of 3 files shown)


packages/slate/src/interfaces/path.ts
```diff
@@ -118,6 +118,14 @@ export const Path = {
     )
   },
 
+  /**
+   * Check if the path of previous sibling node exists
+   */
+
+  hasPrevious(path: Path): boolean {
+    return path.length !== 0 && path[path.length - 1] > 0
```
BrentFarese, line 126: Do you need the first part of this check?  Otherwise, looks good!

mmdonaldson: Fixed

BrentFarese (approved):

lcswillems:
I don't think it solves the issue it says it solves. The issue asks for Slate not throwing an error. Even with this new method it will still throw an error.

lcswillems:
By the way, thank you @BrentFarese for the time you take to review these PRs!

BrentFarese:
Thanks @lcswillems!  I agree, it **_somewhat_** fixes the issue.  The issue really though is a larger issue that Slate `throw`s in various places where a better DX might be to return `null`.  This is something I might suggest we consider tackling on the road to 1.0 as it might be a better DX.  However, I'm comfortable closing out this issue as it pretty much provides a path around Slate `throw`ing, in this specific case.  

Here is a [related issue](https://github.com/ianstormtaylor/slate/issues/3641) that captures Slate liberally `throw`ing errors, so I think we are covered.

lcswillems:
I didn't know this issue about throw, so yes it covers it! You are right, it is a sufficient PR to considerably solve the issue.

lcswillems:
@BrentFarese Regarding 1.0, do you know if it will happen anytime soon? I assume it depends on the contribution from people, so the release date will be more or less random?

BrentFarese:
Yup there likely won't be any set release date, but we have a [milestone](https://github.com/ianstormtaylor/slate/milestone/7) to start tracking progress.

Issue #3967, opened 2020-11-02
Proposal - path parameter in match function


Mangatt:
#### Do you want to request a _feature_ or report a _bug_?

Feature

#### What's the current behavior?

Match function gets just `node` as parameter.

```js
const match = (node) => boolean
```

Slate: 0.59.0

#### What's the expected behavior?

Since location of node actually matters, it would be helpful to have that information available in match function like this:

```js
const match = (node, path) => boolean
```

It is minor non-breaking change that grealy simplifies code dependent on location.

BrentFarese:
Agree this could be a good improvement.  Do you want to draft a PR?  Thanks!

reinvanimschoot:
Is there any update on this? I am looking for ways to get a reference to the node parent when matching but it seems every function requires a reference to the path.

BrentFarese:
> Is there any update on this? I am looking for ways to get a reference to the node parent when matching but it seems every function requires a reference to the path.

Someone needs to submit a PR.  If you want to take a crack, it could get considered and possibly merged way faster.

reinvanimschoot:
@BrentFarese But there is an open PR? https://github.com/ianstormtaylor/slate/pull/3968

BrentFarese:
> @BrentFarese But there is an open PR? #3968

Should have looked at the linked PRs to this issue!  Thanks.  I think there are shortcomings to that PR that need to be fixed (see my comment), but otherwise I would be glad to take a look and merge if the comments are addressed.

PR #3968, opened 2020-11-03
Path parameter in NodeMatch function

Mangatt:

Feature

#### What's the new behavior?

Match function now gets second parameter with node path.

It is non-breaking change.

Fixes: #3967


ianstormtaylor:
I can't remember for sure, but I believe we used to do this before and changed it to only take a `node` because there are a few places where a `path` is not available to matching functions—potentially normalization. Not entirely sure, but worth checking on.

BrentFarese:
@Mangatt it seems this PR only changes the implementation for `Editor.nodes`, but there are other locations in the codebase where a `NodeMatch` function is used.  Can you change those locations too?  We shouldn't update the documentation or the TS interface for `NodeMatch`, I wouldn't think, until all places in the code that use a `NodeMatch` function are called with `match(n, p)`.

reinvanimschoot:
@BrentFarese I'm working on another PR with all match functions changed to `match(n, p)`. That being said, so far I've only noticed one other places where `match(n, p)` is explicitly called and not deferred to `Editor.nodes` which is `Editor.levels`. 

All `Transform` functions seem to make use of `Editor.nodes` for their match functionality and so do most of the other `Editor` functions.

 I am I missing another place where there is an explicit `match` function called?

BrentFarese:
> @BrentFarese I'm working on another PR with all match functions changed to `match(n, p)`. That being said, so far I've only noticed one other places where `match(n, p)` is explicitly called and not deferred to `Editor.nodes` which is `Editor.levels`.
> 
> All `Transform` functions seem to make use of `Editor.nodes` for their match functionality and so do most of the other `Editor` functions.
> 
> I am I missing another place where there is an explicit `match` function called?

I would look to see everywhere where the `NodeMatch` type is used in the code base, and then perhaps do a global find for `match` to see all the locations where the `match` parameter is used in a `Transform` or an `Interface`.

reinvanimschoot:
The thing is that the `Transforms` do not use the `NodeMatch` type, but `(node: Node) => boolean`, which might also be soemthing to clean up along the way. But I'll take a look at it.

BrentFarese:
> The thing is that the `Transforms` do not use the `NodeMatch` type, but `(node: Node) => boolean`, which might also be soemthing to clean up along the way. But I'll take a look at it.

Ah I see.  Yes if you're able to clean up the usage of that type, that would be a benefit too!

PR #3978, opened 2020-11-11
Handle sibling moves in moveNodes transform

BobElDevil:
Fixing a bug

#### What's the new behavior?
moveNodes transform properly handles sibling moves when multiple nodes are moved

During application, 'moveNodes' uses path refs to track the nodes that are being moved, as well as the destination path. This relied on the fact that when a node is inserted, the node that was originally at that location is pushed after, so the transform would ensure subsequent moves occur after the previous ones. 

However when a sibling move occurs from an earlier to later index, the node at that location actually shifts earlier. This means when we loop through the move ops, subsequent moves end up being targeted to before the previous ones, reversing the order of the nodes. 

Detect a sibling move, and adjust the destination of the move in the loop when that's the case

### DIFF
--- a/packages/slate/src/transforms/node.ts
+++ b/packages/slate/src/transforms/node.ts
@@ -486,6 +486,13 @@ export const NodeTransforms: NodeTransforms = {
         if (path.length !== 0) {
           editor.apply({ type: 'move_node', path, newPath })
         }
+
+        if (Path.isSibling(newPath, path) && Path.isAfter(newPath, path)) {
+          // When performing a sibling move to a later index, the path at the destination is shifted
+          // to before the insertion point instead of after. To ensure our group of nodes are inserted
+          // in the correct order we increment toRef to account for that
+          toRef.current = Path.next(toRef.current!)
+        }
       }
 
       toRef.unref()
--- a/packages/slate/src/transforms/node.ts
+++ b/packages/slate/src/transforms/node.ts
@@ -487,11 +487,15 @@ export const NodeTransforms: NodeTransforms = {
           editor.apply({ type: 'move_node', path, newPath })
         }
 
-        if (Path.isSibling(newPath, path) && Path.isAfter(newPath, path)) {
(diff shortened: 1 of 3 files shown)


jason-krypton:
LGTM. Similar to other bugs that have been fixed with sibling moveNodes.

BobElDevil:
@BrentFarese tagging you on this one. Thanks for taking a look!

packages/slate/src/transforms/node.ts
```diff
@@ -377,6 +377,13 @@ export const NodeTransforms = {
         if (path.length !== 0) {
           editor.apply({ type: 'move_node', path, newPath })
         }
+
+        if (Path.isSibling(newPath, path) && Path.isAfter(newPath, path)) {
+          // When performing a sibling move to a later index, the path at the destination is shifted
+          // to before the insertion point instead of after. To ensure our group of nodes are inserted
+          // in the correct order we increment toRef to account for that
+          toRef.current = Path.next(toRef.current!)
```
BrentFarese, line 385: I generally think the use of `!` in Slate core is an anti-pattern.  Can we just move to check that `toRef.current` exists in the if block on line 381?

BrentFarese (changes requested):
Apart from a minor requested change, this would look good to merge!  Make sure it's up to date w/ Master too please.  Thanks.

BobElDevil:
@BrentFarese Made the requested changes and rebased on master. Thanks for the review 👍

BrentFarese:
@BobElDevil I think if you install `@babel/helper-call-delegate` it will fix the deploy?  Give it a crack and then glad to merge.

BobElDevil:
@BrentFarese Sorry missed the notification for this, just rebased on the latest which pulled in the babel dependency, checks should be passing now.

BrentFarese (approved):

BrentFarese (approved):

PR #3979, opened 2020-11-12
Add button to remove link in Links example

frankdilo:
Just added a remove link button in the Links example to make things clearer (took me a while to figure out how to implement that).

Here is how it works:

CleanShot 2020-11-12 at 09 49 15


ianstormtaylor:
Love it, thanks @frankdilo!

PR #3988, opened 2020-11-17
rerender slate when focus changes.

### LINKED ISSUE #3987: Use Focus does not change when keyboard changes focus

Bug

- 
  - a bit hard to see in the gif but when I click the focus hook fires and changes as expected. When I press escape the focus hook does not change. I'm logging when the focus hook changes, and I'm using css styling to visualize focus on the editor. You can see when I press escape because the visual styling disappears but no focus event is logged until I click outside the editor
- [reproduction sandbox](https://codesandbox.io/s/create-react-app-with-typescript-forked-h7g73)
  - in the reproduction open up the log and try clicking in and then out of the editor. Now try clicking in the editor and pressing escape. You can see the focus hook does not update

Slate: 0.59.0
Browser: Chrome 
OS: Mac 

Expected behavior is the useFocused hook updates when the editor loses focus.

lukesmurray:
Rerender slate when focus changes. Fixes #3987

Fixing a bug

#### What's the new behavior?

I've shown the old behavior in the initial bug report. When the editor loses focus due to the user pressing tab or escape the `useFocused` hook does not update. While debugging I found that this is due to the `Slate` wrapper not rerendering, so the `useFocused` Context Provider never receives a new value.

Here is a code snippet which can be used to demonstrate the new behavior. Without the new code `is focused` only logs when you click outside the editor. With the new behavior `is focused` logs when you tab outside of the editor or press escape.

```tsx
import React, { useEffect, useMemo, useState } from 'react'
import { createEditor, Node } from 'slate'
import { Editable, Slate, useFocused, withReact } from 'slate-react'

const FocusTester = () => {
  const focused = useFocused()
  useEffect(() => {
    console.log('is focused', focused)
  }, [focused])

  return <></>
}

const FocusExample: React.FC = props => {
  const editor = useMemo(() => withReact(createEditor()), [])
  // Add the initial value when setting up our state.
  const [value, setValue] = useState<Node[]>([
    {
      type: 'paragraph',
      children: [{ text: 'A line of text in a paragraph.' }],
    },
  ])

  return (
    <Slate
      editor={editor}
      value={value}
      onChange={newValue => setValue(newValue)}
    >
      <Editable />
      <FocusTester />
    </Slate>
  )
}

export default FocusExample
```

When looking at the code I realized `ReactEditor.isFocused` does return the correct value at all time already. `Blur` and `Focus` events are already getting handled correctly. The only issue is `Slate` is not rendering when the `ReactEditor.isFocused` value changes.

In order to resolve the issue I just need to get slate to rerender when focus changes. My solution is in `slate.tsx`. In my solution I move `isFocused` to local react state. I listen to `focus` and `blur` events at the document level and then update the local state with `ReactEditor.isFocused`. I also update the local state on every render with `useEffect`. You can see this implementation in `slate.tsx`.  

If `useState` returns the same value as the previous state then react will not rerender. So this code should not cause any unnecessary renders.

I wanted to try and avoid the `focus` and `blur` handlers in `editable` because there is a bunch of edge case behavior handled there. All of those edge cases should still be handled the exact same way, because my change is essentially a read only change to read `ReactEditor.isFocused` when the focus changes on the page.

Fixes: #3987
Reviewers:

### DIFF
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -6,6 +6,7 @@ import { FocusedContext } from '../hooks/use-focused'
 import { EditorContext } from '../hooks/use-slate-static'
 import { SlateContext } from '../hooks/use-slate'
 import { EDITOR_TO_ON_CHANGE } from '../utils/weak-maps'
+import { useIsomorphicLayoutEffect } from '../hooks/use-isomorphic-layout-effect'
 
 /**
  * A wrapper around the provider to handle `onChange` events, because the editor
@@ -40,10 +41,40 @@ export const Slate = (props: {
     }
   }, [])
 
+  const [isFocused, setIsFocused] = useState(ReactEditor.isFocused(editor))
+
+  useEffect(() => {
+    setIsFocused(ReactEditor.isFocused(editor))
+  })
+
+  const focusHandler = useCallback(() => {
+    setIsFocused(ReactEditor.isFocused(editor))
+  }, [])
+
+  useIsomorphicLayoutEffect(() => {
+    document.addEventListener('focus', focusHandler, true)
+
+    return () => {
+      document.removeEventListener('focus', focusHandler, true)
+    }
+  }, [focusHandler])
+
+  const blurHandler = useCallback(() => {
+    setIsFocused(ReactEditor.isFocused(editor))
+  }, [])
+
+  useIsomorphicLayoutEffect(() => {
+    document.addEventListener('blur', blurHandler, true)
+
+    return () => {
+      document.removeEventListener('blur', blurHandler, true)
+    }
+  }, [blurHandler])
+
   return (
     <SlateContext.Provider value={context}>
       <EditorContext.Provider value={editor}>
-        <FocusedContext.Provider value={ReactEditor.isFocused(editor)}>
+        <FocusedContext.Provider value={isFocused}>
           {children}
         </FocusedContext.Provider>
       </EditorContext.Provider>


ianstormtaylor:
Hey @lukesmurray thanks for submitting this. It looks like as currently written it's going to be re-rendering all editors on the page whenever _anything_ is focused/blurred instead of specifically when the editor in question is focused/blurred? Which seems inefficient.

I wonder if that's a reason to move this logic into the `onFocus` and `onBlur` handlers to ensure that it's only happening for the specific editor in question.

lukesmurray:
Thanks for reviewing the PR! I didn't think about the multi editor scenario but I just checked and it seems to work correctly.

Here's a repro with multiple editors separated by horizontal rules, you can see as I tab through and click on different editors at most two editors are rererenderd, the one losing focus and the one getting focus.

multi-editor-example

You're right that we call `setState` whenever the focus changes at the document level but if the editor focus hasn't changed then it's a no-op.

> If you update a State Hook to the same value as the current state, React will bail out without rendering the children or firing effects. 

The reason this isn't in the `onFocus` and `onBlur` handles is because we need to change the value passed to the provider. Unless I'm mistaken if we want to change the value from a nested component we would have to add a setter to the provider which would be changing it's api.

Here's the multi editor code, adapted from the large document example.

```tsx
import faker from 'faker'
import React, { useCallback, useMemo, useState } from 'react'
import { Editable, Slate, withReact } from 'slate-react'

const HEADINGS = 1
const PARAGRAPHS = 1
const EDITORS = 5
const editorValues = []
for (let e = 0; e < EDITORS; e++) {
  const initialValue = []

  for (let h = 0; h < HEADINGS; h++) {
    initialValue.push({
      type: 'heading',
      children: [{ text: faker.lorem.sentence() }],
    })

    for (let p = 0; p < PARAGRAPHS; p++) {
      initialValue.push({
        children: [{ text: faker.lorem.paragraph() }],
      })
    }
  }
  editorValues.push(initialValue)
}

const MultipleEditors = () => {
  return (
    <div>
      {editorValues.map((initialValue, i) => (
        <div key={i}>
          <SlateEditor initialValue={initialValue} />
          <hr />
        </div>
      ))}
    </div>
  )
}

const SlateEditor = (props: { initialValue: any }) => {
  const [value, setValue] = useState(props.initialValue)
  const renderElement = useCallback(props => <Element {...props} />, [])
  const editor = useMemo(() => withReact(createEditor()), [])
  return (
    <Slate editor={editor} value={value} onChange={value => setValue(value)}>
      <Editable renderElement={renderElement} spellCheck autoFocus />
    </Slate>
  )
}

const Element = ({ attributes, children, element }) => {
  switch (element.type) {
    case 'heading':
      return <h1 {...attributes}>{children}</h1>
    default:
  }
}

export default MultipleEditors
```

ianstormtaylor:
@lukesmurray ah great call. Thank you!

Issue #3868, opened 2020-09-09
Transforms.removeNodes(editor) does not delete the current selection when it includes void nodes


markogresak:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

The location where `Transforms.removeNodes(editor)` is applied is offset. The target content is not entirely removed, but it removes the content before it.
I have reproduced it here: https://codesandbox.io/s/slate-reproductions-forked-52x60?file=/index.js

<details>
  <summary>Recording</summary>
 
![SlateBug](https://user-images.githubusercontent.com/6675751/92641862-a77f7b80-f2df-11ea-8182-49d28e94c1d8.gif)
  
</details>

If you do not trust the code which plays the scenario, you can try it out manually.
Steps to reproduce:
1. Open the example at https://codesandbox.io/s/slate-reproductions-forked-52x60?file=/index.js.
2. Click at the beginning of the "Paragraph 2".
3. Hold down <kbd>⇧ Shift</kbd> and press the <kbd>↓</kbd> (down key) 3 times.
4. Press <kbd>← Backspace</kbd>.

<details>
  <summary>Recording of manual steps</summary>
 
![SlateBugManual](https://user-images.githubusercontent.com/6675751/92642203-25dc1d80-f2e0-11ea-90cb-19a5cf5860f5.gif)
  
</details>

The result is not exactly the same, but it's also wrong.

Slate: 0.58.4
Browser: Chrome (and Firefox too, so I assume any browser)
OS: Mac (but I assume it's the same on other platforms)

#### What's the expected behavior?

`Transforms.removeNodes(editor)` removes exactly what's selected, nothing more, nothing less.

markogresak:
It's also possible to reproduce this with the images example at https://www.slatejs.org/examples/images.

Steps:
1. Duplicate the first paragraph of the editor.
2. Select the first paragraph and the image below.
3. Press backspace.

A recording: https://share.getcloudapp.com/NQurnA90 (Apologies for a 3rd party link - Github took too long to upload the gif and it won't accept the mp4 recording).

kamilmielnik:
> The result is not exactly the same, but it's also wrong.

[...]

Issue #3456, opened 2020-01-22
Inline void elements cannot be deleted when selected


laverdet:
#### Do you want to request a _feature_ or report a _bug_?
This is a bug report

#### What's the current behavior?
You can reproduce this easily on the Slate mentions example. Go here: https://www.slatejs.org/examples/mentions

*Click* one of the mentions so that it's highlighted with the custom rounded blue border. Now press backspace or delete. You can't delete it when it is selected. If you press right arrow and then backspace it will work fine.

![image](https://thumbs.gfycat.com/ColossalDisguisedKawala-size_restricted.gif)

It seems that block voids work, since `examples/images` works fine.

Slate: 0.57.1
Browser: Version 79.0.3945.117 (Official Build) (64-bit)
OS: macOS 10.15.1 (19B88)

#### What's the expected behavior?
I would expect that pressing delete would delete the mention.

laverdet:
Another issue with void elements, again reproducible in the examples. To reproduce:

* Visit the image example: https://www.slatejs.org/examples/images
* Click the image icon and paste an image URL (try https://source.unsplash.com/kFrdX5IeQzI)
* Cmd+A
* Delete

This will delete all text up to the first image, leaving the rest untouched. It seems to only happen when an image is the last element in the editor.

![image](https://thumbs.gfycat.com/CompletePopularCod-size_restricted.gif)

skogsmaskin:
> This will delete all text up to the first image, leaving the rest untouched. It seems to only happen when an image is the last element in the editor.

This also happens if there is an empty paragraph as last block.

abhishekbhardwaj:
I think it happens whenever a void element is at the start or end of the document.

if you put a checkbox at the top of the doc, it does the same to the actual checkbox.

abhishekbhardwaj:
@laverdet @skogsmaskin I was able to fix this last image issue by doing the following in my `withImages` plugin:

```
editor.deleteFragment = () => {
    const { selection } = editor;

    if (selection && Range.isExpanded(selection)) {
      const images = Array.from(
        Editor.nodes(editor, {
          match: n => n.type === TYPES.IMAGE
        })
      );

      if (!!images.length) {
        // We're only deleting the last image as that's what is always left behind.
        // Slate is handling the rest easily.
        const [, cellPath] = images[images.length - 1];
        Transforms.delete(editor, {
          at: cellPath,
          voids: true
        });
      }
    }

    deleteFragment();
  };
```

[...]

PR #3990, opened 2020-11-20
Fix delete behaviour when selection ends in void element

songeunyou:
Fixing a bug

#### What's the new behavior?
Using `Cmd+A` -> `Delete` properly deletes the selection when the last element in the editor is a void element or a void element followed by a newline.

Slate Issue 3456

<details>
  <summary>Intuitive cursor position after delete (Addressing [#3868])</summary>
This fix allows void nodes to be deleted correctly and places the cursor on a newline after deletion.

void-deletion

</details>

Previously, when using `Cmd+A` -> `Delete`, selections ending with a void element had their endpoint at offset: 0 (before the void element), causing `Editor.unhangRange` to trigger and omit the last elements when executing delete.

Now, `Transforms.delete` checks that the last element is not a void element before calling `Editor.unhangRange`. If it is a void element, `Editor.unhangRange` is skipped, allowing the selection to be deleted properly.

Additionally, voids is passed as true to `Editor.nodes` within `Editor.unhangRange` so that void elements will be iterated through while unhanging. This resolves the issue of buggy deletion on void elements followed by a newline and does not affect selections without void elements.

Two new tests were added to address these changes. The output of the test delete-void-and-newline-at-end has an extra `<block/>` than would be expected; this is another small bug that occurs when deleting selections ending in a newline. However, that is an issue that should be addressed in a separate PR.

Fixes two sub-issues mentioned in: [#3456]
Fixes: [#4005]
Fixes adjacent issue: [#3868]


packages/slate/src/interfaces/editor.ts
```diff
@@ -1265,6 +1265,13 @@ export const Editor = {
       at: end,
       match: n => Editor.isBlock(editor, n),
     })
+
+    // If last element is a void node, unhang range by including void node in range
+    if (Editor.isVoid(editor, endBlock[0])) {
```
BrentFarese, line 1270: I am not sure this the best solution.  Should we just simply not unhang the range in `Transforms.delete` if the selection end is inside a void node?  I think maybe we should handle this in `Transforms.delete` vs. changing the default behavior of `unhangRange`?  Let's pair on it.

songeunyou: I thought there were some path issues before, but I didn't realize that they're handled further down in `Transforms.delete`. I've moved the end void node handling to `Transforms.delete` so that it now skips `unhangRange`. If you think there are still changes that could be made I'm happy to pair. Thanks for the feedback!

packages/slate/src/interfaces/editor.ts
```diff
@@ -1274,7 +1275,7 @@ export const Editor = {
       at: before,
       match: Text.isText,
       reverse: true,
-      voids,
+      voids: true,
```
BrentFarese, line 1278: I think we might want to leave this as a variable and perhaps change how `Transforms.delete` is called in slate-react instead.

packages/slate/src/transforms/text.ts
```diff
@@ -70,7 +70,15 @@ export const TextTransforms = {
...
-        at = Editor.unhangRange(editor, at, { voids })
+        const [, end] = Range.edges(at)
+        const endBlock = Editor.above(editor, {
+          at: end,
+          match: n => Editor.isBlock(editor, n),
+        })
+
+        if (endBlock && !Editor.isVoid(editor, endBlock[0])) {
+          at = Editor.unhangRange(editor, at, { voids })
+        }
```
BrentFarese, line 81: I think this might result in another set of bugs that we don't really intend to introduce.  What if the `endBlock` is a `void` but it's not at the end of the document?  Then, the `void` node would be deleted because the range is not "unuhung".  Are there test cases for hanging ranges with voids in the middle of a document?  If not, we should likely add a test case(s).  And, maybe we consider doing something like the below:

```suggestion
        const [, end] = Range.edges(at)
        const endDoc = Editor.end(editor, []);

        if (!Point.equals(end, endDoc)) {
          at = Editor.unhangRange(editor, at, { voids })
        }
```

Need to think about this behavior a bit more.  The above might not be exactly what we want.  We should pair.

lukesmurray:
Potentially related to this. In next, if the last element is a void element, then you click away, and refocus, the void element gets mouse focus and you can't edit or do anything until you use the arrow keys to navigate away from the void element.

songeunyou:
@lukesmurray what is "next" in reference to? Also, will take a look into this + finish up this fix this weekend, thanks for the comment.

lukesmurray:
`slate@next` published on npm. Version 0.60.2.

songeunyou:
@BrentFarese, I looked into the default voids behavior in `Transforms.delete` that was raised during our pairing session, and the current implementation is correct (`voids: false` denotes that text within void elements can be deleted). Changing it introduces further issues.  

The current changes in this PR still fixes all of the issues, so I would appreciate a final review + merge!

ulion:
I encountered similar but a little different issues. here is the case:
<text>|abc</text><inline void/><text>|</text>
If select bc+inline void, it can be deleted, but if select all (abc+inline void), the inline void is not deleted due to the unhangRange method skipped the last text, then skipped the inline void since it does not match Text.isText, finally after unhang it got only 'abc'. That is definitely incorrect unhang result. despite start.offset check in unhangRange is no reason. the inline void should not be removed from selection, nor the inline void be counted as anything hanging.
So, what's proper fix for this?

Issue #3927, opened 2020-10-22
'split_node' op: can't exclude node properties


bryanph:
#### Do you want to request a _feature_ or report a _bug_?
In `Editor.transform` a `split_node` operation always results in a spread of all object properties of the node that is being split into the newly inserted node. Because of that there is no way to exclude properties. The properties are already included in the `properties` property so there seems to be no need for this. 

https://github.com/ianstormtaylor/slate/blob/0bbe121d76c5c2313d939de8a7ebed3bfd37f62d/packages/slate/src/interfaces/editor.ts#L1484-L1486

#### What's the current behavior?
Node properties are always included into the newly inserted node for a `split_node` operation

#### What's the expected behavior?
Only merge properties from the `properties` object into the new node for a `split_node` operation

BrentFarese:
I see your point.  But I am not so sure it is a `split_node` op if theoretically only some properties of the node being split are merged in.  Is this something that is better left to the dev - _e.g.,_ if you want to omit properties, omit them after the split?  If you can elaborate on your actual use case, that might help a bit.  I am just somewhat wary of removing this as a safeguard since then we have to be sure **_every_** call to `split_node` has a full set of properties/the correct properties.

bryanph:
My use-case is setting a `uuid` on every element and so if a block is split I want to assign a new `uuid` to the split block. My first idea was to override `apply` and remove the `uuid` from `properties` in the `split_node` op but because in `apply()` the properties of the actual node are also merged in, the `properties` property basically has no effect. I would expect it to respect the `properties` property and just merge those properties. That is currently not the case. I solve this differently now by overriding the `insertBreak` command and doing the required work there. Because under the hood for `split_node` already all properties of the node are set on `properties` I don't see a need to merge in the node properties at `apply` time as well.

TheSpyder:
> My use-case is setting a `uuid` on every element and so if a block is split I want to assign a new `uuid` to the split block.

This might not matter to you now, but please note that if you ever turn on real-time collaboration it is critical that `split_node` properties contain the exact set of attributes from the node that was split - no more, no less. Applying your own hand-crafted `split_node` operations that do not do this will lead to edge case convergence failures.

The correct way to approach this from an RTC perspective is to apply a normal `split_node` and then use a `set_node` to adjust the UUID on the newly split element.


> I am just somewhat wary of removing this as a safeguard since then we have to be sure **_every_** call to `split_node` has a full set of properties/the correct properties

[...]

PR #4004, opened 2020-11-30
Don't clone nodes when applying a split_node operation

TheSpyder:

Fixing a bug

#### What's the new behavior?

Slate no longer clones nodes when applying a `split_node` operation. This clone is unnecessary, and it causes a problem in real-time collaboration (I'll explain at the end, it's hard to make a video so I'll try explaining it conceptually).

When Slate constructs a `split_node` operation, it ensures the operation contains all properties except `text` and `children`:
https://github.com/ianstormtaylor/slate/blob/b1f291ef88d6d0ae921e61690e3661a4962db5e9/packages%2Fslate%2Fsrc%2Ftransforms%2Fnode.ts#L634-L640

This change leverages that guarantee by not cloning the existing node when applying a `split_node` operation. I added tests for the existing behaviour to ensure it did not break, and also tests that are resolved by the change.

- [x] The linter passes with `yarn lint`.

### Detailed explanation
Given this document structure:
```
<element data>
  some text
</element>
<element>
  <cursor />
</element>
```

Picture two users collaborating on this document, both have their cursor in the marked position. One presses backspace to merge the blocks, the other presses enter to split the second block. Following through what happens for each user:
#### Enter case
The element is split leaving three elements. Note the `split_node` operation contains an empty `properties` object because it applied to the second element.
```
<element data>
  some text
</element>
<element>
</element>
<element>
  <cursor />
</element>
```
#### Backspace case
A merge operation is generated (and a few other things to normalize the document) leaving a single element with a `data` attribute.
```
<element data>
  some text<cursor />
</element>
```

In order to converge these two documents, the transformed operations need to do the following:
* In the enter case, the incoming merge is applied as is (resulting in an identical document)
* In the backspace case, the incoming split must be applied to the single remaining element - but recall the split operation contains no attributes. The result is:
```
<element data>
  some text<cursor />
</element>
<element data>
</element>
```

Due to cloning the first node to apply the split the second case results in a data attribute on the second element and the documents have not converged. The only way to converge under these restrictions is to "consume" the user who pressed enter and leave a single element for both users. This is not a good result. Removing the clone is an easy fix and should not impact general use without collaboration.

### DIFF
--- a/packages/slate/src/transforms/general.ts
+++ b/packages/slate/src/transforms/general.ts
@@ -248,7 +248,6 @@ export const GeneralTransforms: GeneralTransforms = {
           const after = node.text.slice(position)
           node.text = before
           newNode = {
-            ...node,
             ...(properties as Partial<Text>),
             text: after,
           }
@@ -258,7 +257,6 @@ export const GeneralTransforms: GeneralTransforms = {
           node.children = before
 
           newNode = {
-            ...node,
             ...(properties as Partial<Element>),
             children: after,
           }
--- a/packages/slate/test/operations/split_node/element-empty-properties.tsx
+++ b/packages/slate/test/operations/split_node/element-empty-properties.tsx
@@ -1,0 +1,30 @@
+/** @jsx jsx */
+import { jsx } from '../..'
+
+export const input = (
+  <editor>
+    <element data>
+      before text
+      <inline>hyperlink</inline>
+      after text
+    </element>
+  </editor>
+)
+export const operations = [
+  {
+    type: 'split_node',
+    path: [0],
+    position: 1,
+    properties: {},
+  },
+]
+export const output = (
+  <editor>
+    <element data>before text</element>
+    <element>
+      <text />
+      <inline>hyperlink</inline>
+      after text
+    </element>
+  </editor>
+)
--- a/packages/slate/test/operations/split_node/element.tsx
+++ b/packages/slate/test/operations/split_node/element.tsx
@@ -1,0 +1,32 @@
+/** @jsx jsx */
+import { jsx } from '../..'
+
+export const input = (
+  <editor>
+    <element data>
+      before text
+      <inline>hyperlink</inline>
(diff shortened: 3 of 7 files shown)


mwood23 (approved):
Applied these changes locally to test and everything appeared to work the same!

bryanph:
This also resolves https://github.com/ianstormtaylor/slate/issues/3927, and might be worth mentioning that it is a breaking change.

TheSpyder:
In what way is this a breaking change? Existing `split_node` operations will still work, and manually generating new ones with undefined values will also be fine. Anyone inspecting the contents of the operation might have an issue but that seems like an extremely small use case.

ianstormtaylor:
@TheSpyder good catch!

ianstormtaylor (approved):

bryanph:
can this be merged?

TheSpyder:
I've been running a copy of Slate with the patch applied since I logged it. Would be nice to see it merged.

BrentFarese:
> I've been running a copy of Slate with the patch applied since I logged it. Would be nice to see it merged.

This seems like a fairly good/non-controversial PR to me.  Any "gotchas" here @TheSpyder?  It seems that we are just allowing the operation to specify what `properties` are carried over to the node that is split, which allows more flexibility and seems good IMO.  And, things like `Transforms.splitNodes`, as far as I know, always include the `properties` from the node being split in the `split_node` operation already, so this should work just the same for most users that use `Transforms.splitNodes` and don't consume/use the `split_node` operation directly?

Looks good to me and glad to merge if there are no major concerns.

TheSpyder:
Correct, if the Slate API is used nothing should change. The only developers impacted by this are creating split node operations manually, either for custom reasons (as in that other comment you linked to) or through collaboration.

I don’t have any concerns about that other ticket, perhaps this change opens up the possibility to misuse split node operations but I think that’s an acceptable risk.

PR #4007, opened 2020-12-02
Smoothly scroll into view if needed 

maccman:
Smoothly scroll into view if needed, and and scroll to the nearest node (rather than the center)

Fixing a bug.


ianstormtaylor:
Hey @maccman thank you for taking the time to submit this! Sorry for the delay. I'm going to close in favor of https://github.com/ianstormtaylor/slate/pull/4037 which will be a more generic way of letting people override the scrolling behavior. Hope that makes sense.

PR #4011, opened 2020-12-08
Fix broken attribute references

BobElDevil:

Fixing a bug

#### What's the new behavior?

Properly accessing properties with hyphens in them

It looks like when the syntax highlighting change originally went in during #3762, the attribute access was typed as `leaf.attr-name`, and then the linter/prettier came and added whitespace to make it `leaf.attr - name`, which is incorrect, although it doesn't 'fail' because there happens to be the global `name` variable


jason-krypton (approved):

ianstormtaylor:
Thanks @BobElDevil!

Issue #3956, opened 2020-10-30
Make getDirtyPaths a function that can be customized by Slate users


BrentFarese:
Make `DIRTY_PATHS` great again.  The idea would be to allow the Slate developer to override `getDirtyPaths` with their own function so that the developer can specify which paths/nodes (s)he wants normalized during the normalization process.  It would be an "opt-out" of Slate's default calculation of `DIRTY_PATHS`.

_Originally posted by @BrentFarese in https://github.com/ianstormtaylor/slate/issues/3758#issuecomment-663727566_

BrentFarese:
Closed by #4012.

Issue #3758, opened 2020-07-01
Editor.wrapNodes does NOT trigger normalize node on wrapped Leaf


flavordaaave:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

1.) Wrap some text with a mark within e.g. a link
2.) `normalizeNode` is executed for the `link` node up
=> The wrapped lead is NOT normalized
3.) When typing within the wrapped node, the normalizer is finally triggered and normalization is applied.

![CxXknLFuXX](https://user-images.githubusercontent.com/5305150/86286427-b7475900-bbe6-11ea-81e1-854fda2acecc.gif)

[Codesandbox](https://codesandbox.io/s/slate-wrapnodes-not-normalizing-leaf-qvgql?file=/src/examples/links.tsx:1212-1249)

Slate: 0.58.3
Browser: Chrome
OS: Mac

#### What's the expected behavior?

When wrapping nodes, alls nodes underneath the wrapper should be normalized

flavordaaave:
Changing an element's type using `Editor.setNodes` also does NOT trigger normalization...

BrentFarese:
The real issue here is that no `DIRTY_PATHS` are calculated for the operations you mention for nodes below where the operation occurred.  This is by design though, I think.  The basic assumption is that the node inside the wrapping has not changed so why does it need to be normalized?

I am not sure I agree with that either but that is the reasoning...

A better solution for Slate users might be to make `getDirtyPaths` a function that is overridable, just like `editor.normalizeNode`.  Then, if you wanted to opt-out of Slate's behavior that is already backed in, you could by providing a `getDirtyPaths` function override that includes the paths you want to normalize, and a custom function in `editor.normalizeNode` to normalize those nodes.

Do you want to take a stab at a PR?

BrentFarese:
@flavordaaave going to close this as #4012 now allows users to customize `getDirtyPaths`.  So, if you need the wrapped node to be normalized in your particular implementation, then you can add it in by customizing `getDirtyPaths`.

Since I can't say for certain in **_all_** cases that all descendants of a wrapped node should be normalized, I'd rather leave it to the particular implementation then add it to core.  If wrong and many users have this same need, glad to add to core's `getDirtyPaths` function as a default.  Thanks!

Fixed by #4012.

PR #4012, opened 2020-12-09
getDirtyPaths can now be customized by Slate users

tomdng:

Adding a new feature as suggested with issue #3956.

#### What's the new behavior?

I moved the `getDirtyPaths()` function from an external function in `create-editor.ts` into the editor object as `editor.getDirtyPaths()`, so you can override Slate's default calculation of dirty paths for an operation. It's meant to be used in conjunction with `editor.normalizeNode()` where you might want to check additional paths due to custom normalization constraints as show by this issue #3758.

To use this, you would create a plugin to override Slate's default behavior. It actually works _very_ similarly to creating custom normalization behavior for `editor.normalizeNode()`. There will be an example in the nextsection.

Most of the customizable slate functions like `editor.normalizeNode()` are created with the `editor` instance in `create-editor.ts`. `getDirtyPaths` was actually also inside the same file, but outside of the editor object as a standalone function. This prevented the function from being customizable via the plugin system unlike `editor.normalizeNode()`. By moving this function into the `editor` instance, it's behavior is the same, but Slate users now get the ability to customize that function if they have additional normalization rules that require different dirty paths to be calculated for an operation.

To demonstrate an example, here's the default behavior of `getDirtyPaths()` as indicated by issue #3758. This example has custom normalization that shouldn't allow bold links, but normalization doesn't trigger initially because the relevant operations don't calculate dirty paths underneath where the operation occurred. However, if you insert a space inside the link, this new operation will indeed give a dirty path to the node that we wanted to nomarlize from the tart. For more detail, [this comment gives a much better explanation](https://github.com/ianstormtaylor/slate/issues/3758#issuecomment-663727566).

Default dirty path calculation example

And here is an example with this new customizable `editor.getDirtyPaths()` where I override the dirty path calculation for the relevant operation (which ended up being `move_node`). Since I can now use a plugin to override `editor.getDirtyPaths()` the normalization runs as soon as you wrap the link because I was able to return the dirty path for the node below the operation.

<details>
<summary>Example code for `editor.getDirtyPaths()` override</summary>

You can use this new `withDirtyPaths()` function as a plugin like
```ts
const editor = useMemo(
    () =>
      withLinks(
        withNormalization(
          withDirtyPaths(withReact(createEditor()))
        )
      ),
    []
  )
```

```ts
const withDirtyPaths = editor => {
  const { getDirtyPaths } = editor

  editor.getDirtyPaths = op => {
    if (op.type === 'move_node') {
      const { path, newPath, node } = op

      if (Path.equals(path, newPath)) {
        return []
      }

      const oldAncestors: Path[] = []
      const newAncestors: Path[] = []

      for (const ancestor of Path.ancestors(path)) {
        const p = Path.transform(ancestor, op)
        oldAncestors.push(p!)
      }

      for (const ancestor of Path.ancestors(newPath)) {
        const p = Path.transform(ancestor, op)
        newAncestors.push(p!)
      }

      const ancestors = [...oldAncestors, ...newAncestors]

      // Since editor.wrapNodes() performs an insert and then a move function
      // to do the actual wrapping, we need to get the path of the node
      // after moving the node. However, the newPath isn't the true path
      // because the newPath of the node will be different after the operation
      // GeneralTransforms.transform() likely has a more robust implementation
      // compared to this example.
      let truePath: Path = []

      if (newPath.length - path.length === 1 && Path.isBefore(path, newPath)) {
        const pathEnd = path.length - 1
        if (newPath[pathEnd] - path[pathEnd] === 1) {
          truePath = [...newPath]
          truePath[pathEnd] = path[pathEnd]
        }
      }

      return [...ancestors, truePath]
    }

    return getDirtyPaths(op)
  }

  return editor
}
```

</details>

link-bold-normalize

Fixes: #3956 
Fixes: #3758

### DIFF
--- a/packages/slate/src/create-editor.ts
+++ b/packages/slate/src/create-editor.ts
@@ -57,7 +57,7 @@ export const createEditor = (): Editor => {
       }
 
       const oldDirtyPaths = DIRTY_PATHS.get(editor) || []
-      const newDirtyPaths = getDirtyPaths(op)
+      const newDirtyPaths = editor.getDirtyPaths(op)
 
       for (const path of oldDirtyPaths) {
         const newPath = Path.transform(path, op)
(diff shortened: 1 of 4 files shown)


timbuckley:
Looks good. Is there any change in the body of `getDirtyPaths`, or was it a straight copy into the editor instance?

tomdng:
> Looks good. Is there any change in the body of `getDirtyPaths`, or was it a straight copy into the editor instance?

The body of `getDirtyPaths` should be a straight copy. I spent most of my time making sure that I could actually override it and fix one of the user issues I linked above. Otherwise it's a super simple change for such a verbose PR 🤐

bryanph:
Did this get reverted? Can't find it here https://github.com/ianstormtaylor/slate/blob/3ad3aaccad5b3aae335eea21e03866e2c266f5a6/packages/slate/src/create-editor.ts#L295

PR #4015, opened 2020-12-11
Feature/improve ts type defs

BrentFarese:

TS enhancement (maybe).

#### What's the new behavior?

By simply unwrapping the functions inside the various modules (started with `Transforms` as an example), we can export functions under the same namespace as now (e.g., `Transforms`) while satisfying Typescript.

This has the advantage of being able to remove the duplicitous Typescript interfaces in each module.

Described above.

#### Comments

@ianstormtaylor I had to bump TS to use [export * as X syntax](https://www.typescriptlang.org/docs/handbook/modules.html#export-all-as-x) and some `es-lint` plugins were yelling at me.  I am not so familiar with the build setup or the `es-lint` rules but the 1 rule that was throwing an error I set to `warn` instead.

I'm marking this PR as a draft so folks can review and see if we feel it's an improvement.  If it is, I can do the same in the `interfaces` and other files while still keeping how we call the API (e.g., `Transforms.setNodes`, etc.) the same.

Reviewers: @thesunny, @ianstormtaylor


ianstormtaylor (approved):

ianstormtaylor:
Looks like a big improvement to not have to keep that stuff in sync, nice find @BrentFarese!

As for `eslint`, feel free to disable whatever needed to get it to work. I think most of the config might have been added before `prettier`, so I'm not even sure if most of it is necessary any more.

Looks like `delete -> deleteContent` may be the only breaking change so far? Just good to keep track of in case there are others in the course of converting interfaces, etc. Sad to lose the `Transforms.delete` syntax but definitely worth it for the gain here 😄

vpontis:
@BrentFarese does it make sense to pull the `prettier` changes into a separate PR? It looks like most of the lines of code changed in this PR are due to prettier changes and not actual meaningful code changes.

ianstormtaylor:
I'm going to try renaming to `master` to `main` again. This is the only PR that still targets `main`, so there's a chance it might get auto-closed by GitHub when I make the change. If so, please feel free to re-open it again since it seems useful from a TypeScript perspective!

PR #4017, opened 2020-12-12
docs: Update document in Chinese

jacob-lcs:

I recently studied Slatejs, but found that its Chinese documentation was badly behind the current version. So I translate according to the latest version of the document, and will update according to the official document.

#### What's the new behavior?

The translated documents are up to date and will be updated later according to the official documents to facilitate Chinese developers.


jacob-lcs:
@thesunny @ianstormtaylor

Issue #4018, opened 2020-12-12
Long placeholder values overflow editor


Glazy:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

![Overflowing text on Slate.js placeholder](https://user-images.githubusercontent.com/8593744/101994041-21e0e280-3cb7-11eb-9bfa-e1126b399c4d.png)

https://codesandbox.io/s/slate-reproductions-forked-ci6kl?file=/index.js

Slate: 0.59.0
Browser: Chrome & Firefox
OS: Linux

#### What's the expected behavior?

Placeholder text should wrap if it is longer than the editor width.

Glazy:
After a little more investigation, this issue occurs due to the following styles on the placeholder `<span>`:

```css
width: 0px;
white-space: nowrap;
```

These are set so that the actual editor `<div>` can sit "underneath" the placeholder content.

Issue #2922, opened 2019-07-11
Placeholder with long text overflows block width


danieltorrer:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
If you add a placeholder with a large string of text, it will go beyond the block width. The width of it should be tied to the block properties. By the way its built it not possible to edit the size via CSS styles.

Fiddle: https://jsfiddle.net/76pukr98/5/
I set the editor to a fixed width. You can type text to see where the line breaks and compare it to the placeholder width.

#### What's the expected behavior?

Placeholder text breaks to match block size

ihd-dev:
You could try style it as follows, i.e. add the style:

> PlaceholderPlugin({
      placeholder:
        'Lorem ipsum dolor sit amet consectetur adipisicing elit. Corrupti assumenda exercitationem saepe iste et? Asperiores facere amet distinctio assumenda eius error, recusandae ut vel atque optio qui ducimus praesentium animi.',
      when: 'isEmpty',
      style: {
      	display: 'inline',
        whiteSpace: 'normal'
      }
    })

danieltorrer:
@ihd-dev  Thanks for your response. I already tried that, you can see it here: https://jsfiddle.net/zLpx4cn8/1/
The issue with this solution is that the cursor is positioned at the end of the dom node.

I think one solution can be to play around with absolute position and adding on top of it the actual input field with a higer z-index. But the current structure of the plugin doesn't allow that AFAIK.

johannesbraeunig:
What about removing the placeholder if the block is selected? I think having `with: 0px` is exactly the "hack" for not displaying the cursor at the end of the line, which is  an issue for longer placeholders.

ianstormtaylor:
I believe that this may be fixed by https://github.com/ianstormtaylor/slate/pull/3093, which has changed a *lot* of the logic in Slate and `slate-react` especially. I'm going to close this out, but as always, feel free to open a new issue if it persists for you. Thanks for understanding.

Glazy:
This issue isn't fixed as of v0.59.0, opened a new issue @ #4018

PR #4019, opened 2020-12-12
Add support for longer placeholder values

Glazy:

Fixing a bug

#### What's the new behavior?

Instead of long placeholder values overflowing the editor bounds horizontally, we now wrap the text and have the editor grow in height.

When the `Leaf` component is rendered, grab the height of the placeholder and the editor. Then, in a `useEffect` dynamically set the editor `minHeight` to that of the placeholder.

Also removed the, now unneeded, styles from the placeholder that were causing the horizontal overlap.

Fixes: #4018, #2922


zaguiini-remo:
Would be nice to have this merged. Is it possible, @ianstormtaylor?

ianstormtaylor:
Thanks @Glazy!

PR #4022, opened 2020-12-15
Mark transform type explicitly

BobElDevil:

Fixing a bug

#### What's the new behavior?

Transforms methods respect custom types

`Transforms` is not an actual interface, and instead it combines objects that each define their own interface. These interfaces all refer to the 'ExtendedTypes', however since the Transform object generates an inferred interface during package generation, it doesn't have any extended types, resulting in an interface that uses the base types.

By explicitly combining these interfaces in the declaration we let typescript keep the extended types in the declaration for us. 

Reviewers: @BrentFarese

### DIFF
--- a/packages/slate/src/transforms/index.ts
+++ b/packages/slate/src/transforms/index.ts
@@ -3,7 +3,10 @@ import { NodeTransforms } from './node'
 import { SelectionTransforms } from './selection'
 import { TextTransforms } from './text'
 
-export const Transforms = {
+export const Transforms: GeneralTransforms &
+  NodeTransforms &
+  SelectionTransforms &
+  TextTransforms = {
   ...GeneralTransforms,
   ...NodeTransforms,
   ...SelectionTransforms,


vpontis:
This looks cool! I wonder if it'll fix some of my typing issues :) Thanks @BobElDevil!

Lmk if there is some way I should try and test this against my local code.

BrentFarese:
Thanks @BobElDevil!  So this fixes TS's type inference incorrectly?  When doing the new "custom types", we struggled with how TS was inferring types incorrectly and the declaration files being spit out of TS were incorrect.  It seems we **_almost_** got it right by providing a defined interface for each `Transform` object, but must've missed this piece of the puzzle.  The declaration files look good now?

Separately, I want to figure out a way to get rid of the large "interface" types in each `Transform` file while still pleasing TS, if we can, b/c it's a lot of code duplication.  #4015 is me taking a first crack at doing that by making each `Transform` file just a series of functions vs. an object and then combining all modules together in a better manner that TS may like better.

BobElDevil:
@BrentFarese Yes this should fix type inference for the transforms at least. I'm not sure if there's more places that needs some sort of fix like this, hard to know until you try to use a function and it doesn't like your types. Would probably have to look at all the '*.d.ts' files generated in the 'dist' folder, and make sure there's nothing exported that uses only the 'Base*' types

I am concerned that something like 4015 will regress that. The one problem with using the extended types instead of generics is that any type inference needs to occur _only_ in the consumer of the library compilation. Any type inference that occurs during the slate compilation step is going to receive an interface that doesn't respect extensions. So everything must be given an explicit type instead of relying on inference.

On the other hand since you're directly exporting the functions in 4015, which have the types specified, it might be alright.

BrentFarese:
> On the other hand since you're directly exporting the functions in 4015, which have the types specified, it might be alright.

That's the thought.  I think TS inference gets messed up because everything is within larger objects.  Anyways, when I get some time I'll work on that branch, compile the types, and check the `d.ts` file for the output to see if it fixes everything and removes a lot of duplicate types.

xiaoyu-tamu:
Can someone release this patch please?

Issue #3920, opened 2020-10-13
scroll to caret  Incorrect


usecodelee:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?
![我的影片](https://user-images.githubusercontent.com/32923083/102490184-73b7ad00-40a9-11eb-97e4-9768f829661a.gif)

![image](https://user-images.githubusercontent.com/32923083/102489585-a4e3ad80-40a8-11eb-9cb7-c16f0d2380f5.png)

When I type, the scroll bar scrolls to the middle

Slate: 0.59.0
Browser: Chrome
OS: Windows/MacOS

#### What's the expected behavior?

When I type, the scroll bar will not scroll to the middle

By reading the source code, I located the reason：
![image](https://user-images.githubusercontent.com/32923083/102490051-4bc84980-40a9-11eb-8612-3e5699ef7493.png)

The content in the red box caused this defect

IsaacMSchultz:
It looks like this was fixed as part of #3866. I tried replicating the issue and it seems to focus correctly for me. The code you have highlighted in the red box is no longer there in the current codebase, and was replaced in #3866.

usecodelee:
> current

thanks

usecodelee:
@IsaacMSchultz 
I updated the issue

dylans:
Fixed in #4037

PR #4023, opened 2020-12-17
fix scroll to caret (#3920)

usecodelee:
fix a bug

#### What's the new behavior?

Can now scroll correctly

Rewrite the scroll function
我的影片2

Fixes: #
Reviewers: @


vpontis:
@usecodelee why do we need this change? What is the bug before that was going on that this fixes? 

It looks like the current code uses `scrollIntoViewIfNeeded` which is more native than this PR which means it should (hopefully) consistently work across different browsers.

usecodelee:
@vpontis 
## issue: #3920 
## before fix:

When we add scroll to editor, 
Will cause scrolling problems, Did not scroll to the caret position
image
我的影片
This question can be seen on the official website：https://www.slatejs.org/examples/richtext

## reason
the current code uses `scrollIntoViewIfNeeded`, 
It will scroll to the middle of the `span`. 
This problem occurs when we continuously input a relatively high `span`

## How to fix
See code

## after fix:
everything is fine, 
When I input content, I can scroll to the caret position normally.
我的影片2

vpontis:
Hmm, so I still don't know what issue you are describing. You could try using Loom.com to record a video with sound. From your GIFs, I have no idea what you are doing and what the behavior is that you expect

usecodelee:
## step 1
open https://www.slatejs.org/examples/richtext
## step 2
modify the properties to make the editor appear scroll bars:
image
## step 3
paste this text into the editor:
`1fhdkshfkshfdlksafhdsklfhdksdhfksaldfhsklahfksdhfkshfdkfhksdhfkshdfksahfkshdjkfhweuywqeoituyeowifqyhdskjhfksahfksdfhskfhuiweqytowiuqyruifohdskhfskahfkjhfjksdhfksdhfskafhskladfhdklsafhkasdfhksahdfksdhjkfsadlfhksladfhksjadhfksdhfklsadfhkslafhklsadfhkdshfdshjfuiqwytowiuqyruifohdskhfskahfkjhfjksdhfksdhfskafhskladfhdklsafhkasdfhksahdfksdhjkfsadlfhksladfhksjadhfksdhfklsadfhkslafhklsadfhkdshfdshjfuiqweuyrtioewqhfhdskfhksadffjkdshfklajhfksdhfkjhfuewlihuffnjdsjfskjanvmzbvkjfhwalhfuelwhfkdshfkfkdjsahjfklfhsklajfhjklsdjhfksytowiuqyruifohdskhfskahfkjhfjksdhfksdhfskafhskladfhdklsafhkasdfhksahdfksdhjkfsadlfhksladfhksjadhfksdhfklsadfhkslafhklsadfhkdshfdshjfuiqweuyrtioewqhfhdskfhksadffjkdshfklajhfksdhfkjhfuewlihuffnjdsjfskjanvmzbvkjfhwalhfuelwhfkdshfkfkdjsahjfklfhsklajfhjklsdjhfkseuyrtioewqhfhdskfhksadffjkdshfklajhfksdhfkjhfuewlihuffnjdsjfskjanvmzbvkjfhwalhfuelwhfkdshfkfkdjsahjfklfhsklajfhjklsdjhfksajfdhjksladfhksldfhksdhfksdhfjksadhfkjshffkjdsahjfdjkshajfkhkdhfksfhksdhfksjahfkdshfkhdksfhskadfhksdfhkshfksdhfkjshdfkshfuewiqyioufhdksjhfskahfksadhfsjakfhklsadfhksafhkslafhksadfhklashfk1234`
## step 4
try to continue typing.

you will see this bug(#3920).

usecodelee:
Shouldn’t I click merge, right? I’m still learning how to use pull request😭

vpontis:
Ah yea, I can see the bug you are describing now. I can record a video later to make it more clear.

RavenColEvol:
How can we have a local fix for this ? Do we have any estimated time till when this fix will get merged in slate ?

packages/slate-react/src/components/editable.tsx
```diff
@@ -218,7 +214,31 @@ export const Editable = (props: EditableProps) => {
...
+      const { selection } = editor
+      if (!selection) return
+      if (!Range.isCollapsed(selection)) return
+      const domSelection = ReactEditor.toDOMRange(editor, selection)
+      if (!domSelection) return
+      const selectionRect = domSelection.getBoundingClientRect()
+      const editorRect = editorEl.getBoundingClientRect()
+      const leafRect = leafEl.getBoundingClientRect()
+      if (
+        selectionRect.bottom - editorRect.top > editorRect.height - 20 ||
```
jacob-lcs, line 229: What does this `20` mean

jacob-lcs:
自动滚动不只是上下，还要考虑左右

packages/slate-react/src/components/editable.tsx
```diff
@@ -11,7 +11,6 @@ import {
 } from 'slate'
 import { HistoryEditor } from 'slate-history'
 import throttle from 'lodash/throttle'
-import scrollIntoView from 'scroll-into-view-if-needed'
```
williamstein, line 14: Since you're no longer using the scroll-into-view-if-needed package, maybe it should also be deleted from package.json?     
https://github.com/ianstormtaylor/slate/blob/ffb6d79b66ba89db7f0687889853011a00944015/packages/slate-react/package.json#L23

By the way, I personally think this is a valuable PR that fixes a serious bug.

dylans (changes requested):
The landing of #4037 may change the need or approach for this one, so please review in context of that update.

This also needs a rebase and a changeset (if this PR is still needed). Thanks!

dylans:
I think #4037 fixes this as [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) supports going to a block end also. Feel free to reopen if you still have this issue and I've missed something here. Thanks!

PR #4027, opened 2020-12-24
Fix useEditor not inside Slate error caused by introduce useSlateStatic

ulion:
Fix ```The `useEditor` hook must be used inside the <Slate> component's context.``` error since useSlateStatic introduced.

#### What's the new behavior?
Slate will still work with useEditor

During introduce useSlateStatic, the EditorContext inside use-editor.tsx is no longer referenced and set editor instance to it.
But for old code which still use useEditor, it will get null editor value by the useEditor call. So this PR change this behaviour by share use the same EditorContext between useSlateStatic and useEditor, so they both works now.

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate-react/src/hooks/use-editor.tsx
+++ b/packages/slate-react/src/hooks/use-editor.tsx
@@ -1,16 +1,10 @@
-import { createContext, useContext } from 'react'
+import { useContext } from 'react'
 
-import { ReactEditor } from '../plugin/react-editor'
-
-/**
- * A React context for sharing the editor object.
- * @deprecated Use useSlateStatic instead.
- */
-
-export const EditorContext = createContext<ReactEditor | null>(null)
+import { EditorContext } from './use-slate-static'
 
 /**
  * Get the current editor object from the React context.
+ * @deprecated Use useSlateStatic instead.
  */
 
 export const useEditor = () => {


ulion:
@songeunyou please check this, thank you.

ulion:
or @timbuckley please. you did one useEditor fix, but need do more.

songeunyou:
@ulion sorry, didn't see this mention earlier! Looks good to me so far, thanks for adding the fix. The Netlify check fails with the error `Cannot find module '@babel/helper-call-delegate'`—this package seems to no longer be in the repo, so this needs to be resolved before merging.

Additionally, I'm not a maintainer so I am unfortunately unable to approve/merge changes to this repo.

packages/slate-react/src/hooks/use-editor.tsx
```diff
@@ -1,16 +1,12 @@
...
-import { ReactEditor } from '../plugin/react-editor'
+import { EditorContext } from './use-slate-static'
 
-/**
- * A React context for sharing the editor object.
- * @deprecated Use useSlateStatic instead.
- */
-
-export const EditorContext = createContext<ReactEditor | null>(null)
+export { EditorContext }
```
BrentFarese, line 5: Why exporting `EditorContext` again?  We're importing it already from `'./use-slate-static`.  Otherwise, looks ok.

ulion: make sense, no need export again. will fix in 1 minute.

BrentFarese: Looks good!

PR #4028, opened 2020-12-25
Fix version links in README

alllex:
Current version links are broken, because they point to non-existing paths.

Fixing typo

#### What's the new behavior?

Fixed links in README


ianstormtaylor:
Thanks @alllex!

Issue #3007, opened 2019-09-08
remove the `data` dictionaries


ianstormtaylor:
#### Do you want to request a _feature_ or report a _bug_?

Debt / improvement.

#### What's the current behavior?

Right now many objects in Slate have a `data` dictionary where you can store userland metadata that Slate doesn't need to concern itself with. This works fine for now, and with Immutable.js it's the only reasonable way to do it.

However, once we switch to plain JSON objects, and are using an interface-based API, I think it will be better to allow userland to store metadata at the top level of any objects. For example:

```js
{
  type: 'link',
  nodes: [...],
  url: 'https://example.com',
}
```

Instead of needing to nest it inside `data.url`, the `url` property can live at the top level. And the object still implements the `Element` interface so everything works as expected.

CameronAckermanSEL:
Panicked when I read "remove data dictionaries" but the alternative sounds good.

ssalka:
I think merging user metadata with internal properties will be problematic - what if my metadata has e.g.`type` or `nodes` fields?

ianstormtaylor:
@ssalka good question! That is definitely the tradeoff here. I don't take it lightly, but I think as I've been working through some of the changes for #2495 I've slightly changed how I think about Slate's data model. I'll explain how my thoughts have evolved since the original decision to keep them separate and you can see what you think...

---

Before, we've had three domain layers:

```js
// This layer is the top one, and used by Slate to store the
// information it cares about.
{
  object: 'block',
  type: 'link',
  nodes: [...],
  // This layer is for the developer using Slate to add any of
  // their domain-specific information.
  data: {
    url: 'https://exmaple.com',
    external: true,
    ...
    // This layer would be if a developer wanted to allow their
    // **end users** to have a namespace where they can keep
    // their own metadata. For example, if you allowed for users
    // to configure their editor differently with a UI.
    properties: {
      ...
    }
  }
}
```

Up to now this has worked, since keeping them separate is guaranteed not to clash—and that is important. And for Immutable.js it's actually required, because it can't handle dynamic properties.

But it's not without tradeoffs. 

It adds complexity to have a nested `data` object, and the API for retrieving things from the nested object becomes awkward. 

[...]

PR #4042, opened 2021-01-13
Allow nodes to hold `null` properties

aglundahl:

Improving the API.

#### What's the new behavior?

Currently, Transforms.setNodes equates `null` with `undefined`, so any property set to `null` will be removed from the node.

This change preserves `null` properties and ensures that only `undefined` properties are removed. The reason why we want this behavior is because we want the type for a property to be explicitly `prop: number | null` and not `prop?: number | null`. We need the nil value to be explicit, because we work a lot with partial nodes (property diffs) and it also works better with JSON serialization.

I haven't found the motivation behind the initial design, but maybe there's a good reason to keep the current behavior. There has been [some discussions](https://github.com/ianstormtaylor/slate/pull/3353) about removing `Transforms.unsetNodes`, though, which obviously collides with this change.

Instead of a loose equality check for `value == null` (which includes both `null` and `undefined`), I strictly check for `value === undefined` before removing the property.

Reviewers: @thesunny @CameronAckermanSEL @ianstormtaylor @BrentFarese


BrentFarese (approved):

ianstormtaylor (changes requested):
The reason for equating `null` and `undefined` was that JSON strips undefined values during serialization, so for OT-style editors we needed to have a way to declare a property should be unset. Instead of opting for an additional `unset_node` operation, we went with `set_node` with `null`. Here's the thread with the original discussion: https://github.com/ianstormtaylor/slate/issues/3007#issuecomment-541747562

I think it would be nice to be able to have `null` at the top-level, but not that this would require creating an `unset_node` operation.

aglundahl:
@ianstormtaylor Ah, I didn't think about the serializability of operations. We don't have that use case (yet). Thanks for the added context!

I think having an `unset_node` operation would improve the API, and I'd be happy to make the required changes if you also think it's an improvement? It's definitely a breaking change, though, so we'd have to plan the release of it accordingly.

ianstormtaylor:
@aglundahl thank you, adding `unset_node` sounds good to me! Be careful to make it so that `unset_node` is invertible, so it'll need to contain enough information to re-set the properties like the others.

TheSpyder:
Adding a new operation is a huge undertaking for collaboration integrations. I'm not saying we can't do it, but we need to be really careful.

Likewise a change to make `null` values not mean "remove property" in `set_node` would break a lot of things. I don't really understand the use case for allowing null values?

aglundahl:
@TheSpyder I'm not using Slate right now, and I'm afraid I don't have time to finish this, so I'm okay with closing this. Unless @BrentFarese or @timbuckley have any objections?

TheSpyder:
All good. We’re clearing out old PRs; this can be reopened (or resubmitted) if it needs to be later.

Issue #4034, opened 2021-01-03
Error when deselecting or pasting: Cannot add property 0, object is not extensible


niieani:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Regression in 0.60.1 (from 0.60.0).

Reproduction:

1. Select any text
2. Either: Copy & paste the same text or copy & unselect it
3. Error: Cannot add property 0, object is not extensible

The error will continue happening for any action in the editor.

Sandbox: https://codesandbox.io/s/slate-0601-bug-usbf4

If you downgrade Slate to 0.60.0, the error goes away.

https://user-images.githubusercontent.com/563469/103475325-230aaa00-4d61-11eb-87e1-109fda8716b6.mov

Slate: 0.60.1
Browser: Chrome 88.0.4324.50 and Safari 14.0 (15610.1.28.1.9, 15610).
OS: Mac

#### What's the expected behavior?

No error is shown.

jacob-lcs:
I fixed the problem
https://github.com/ianstormtaylor/slate/pull/4035

Q-Qplus1s:
@jacob-lcs 
Cannot undo after pasting text

jacob-lcs:
This is the [preview page](https://deploy-preview-4035--slatejs.netlify.app/examples/richtext), can you reproduce this bug? @Q-Qplus1s

Q-Qplus1s:
@jacob-lcs 能否加下微信qiezhd，我们有个 slate.js 技术交流群，同时想问下你用的哪个版本的 slate，我 fork 了 0.60.1 版本，在上面改的

BrentFarese:
Fixed by #3850.

PR #4043, opened 2021-01-13
Prevent immer from freeze operations.

ulion:
Fixing a bug.

#### What's the new behavior?
After copy (editor.getFragment), the editor.operations is not frozen (then no crash due to any update). if in case you actually using immer 7+ though Slate using v5 but if the client code has other dependencies which using higher version of immer, finally the Slate could actually calling immer 7+. 

Like https://github.com/ianstormtaylor/slate/pull/3850/files#r480954559 suggested, but without upgrade the immer version in package.json

Fixes: #4034
Reviewers: @BrentFarese

### DIFF
--- a/packages/slate/src/interfaces/node.ts
+++ b/packages/slate/src/interfaces/node.ts
@@ -292,7 +292,7 @@ export const Node: NodeInterface = {
       )
     }
 
-    const newRoot = produce(root, r => {
+    const newRoot = produce({ children: root.children }, r => {
       const [start, end] = Range.edges(range)
       const nodeEntries = Node.nodes(r, {
         reverse: true,


BrentFarese:
Is Immer freezing `editor.operations` caused by the bump in Immer in `60.1` from `5.0.0` to `5.3.6`?  Do you know @ulion?

ulion:
> Is Immer freezing `editor.operations` caused by the bump in Immer in `60.1` from `5.0.0` to `5.3.6`? Do you know @ulion? Also, I'm probably inclined to merge #3850 as this is a duplicate?

should be immer 7.0 issue. since other packages in my project use immer 7.0, I think sometimes no matter which version slate using, it is effected.

ulion:
well, I am not sure, since I only have a mixed immer 7 env. if you have pure 0.60.1 test env, simply copy then click anywhere will tell us whether it's effected.

BrentFarese:
> well, I am not sure, since I only have a mixed immer 7 env. if you have pure 0.60.1 test env, simply copy then click anywhere will tell us whether it's effected.

K.  Going to close this as it's a duplicate of #3850 and will merge #3850 once the PR owner answers a question(s) I had.  Will get this fix + the Immer version bump in though soon!  Thanks.

PR #4048, opened 2021-01-19
Fix normalizeDOMPoint to do better job.

ulion:
Fix a bug

#### What's the new behavior?

normalizeDOMPoint now get the editable dom point more near to the input non-editable dom point.

```
<div id="a">
  <p>some text</p>
  <div id="b" contenteditable="false">...</div>
</div>
```
It is about the triple click select. If use triple click first <p> content, it will finally result select nothing. which is because, when Editable internally calls ReactEditor.toSlateRange(window.getSelection()) in this case, the end point is at the 'b' div, which internally call ReactEditor.toSlatePoint(div 'a', 1), which will resolve to the [] position in the <p>[]some text</p>, but what we want is <p>some text[]</p> in this case. The problem is caused by the normalizeDOMPoint method, for this case, it internally call getEditableChild with direction 'forward', but it will finally return the p node, the p node is what we want, but since it's in the front of our input dom 'b', the finally result should be at end of node p, but the code in normalizeDOMPoint failed to do that.
This PR will compare getEditableChild returned node index with the input one, to decide locate to node's beginning or end. It finally fixed the triple click can not select all issue on my case.

Fixes: #
Reviewers: @BrentFarese

### DIFF
--- a/packages/slate-react/src/utils/dom.ts
+++ b/packages/slate-react/src/utils/dom.ts
@@ -107,16 +107,21 @@ export const normalizeDOMPoint = (domPoint: DOMPoint): DOMPoint => {
   // If it's an element node, its offset refers to the index of its children
   // including comment nodes, so try to find the right text child node.
   if (isDOMElement(node) && node.childNodes.length) {
-    const isLast = offset === node.childNodes.length
-    const direction = isLast ? 'backward' : 'forward'
-    const index = isLast ? offset - 1 : offset
-    node = getEditableChild(node, index, direction)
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Thanks for this @ulion! I looks like there's a merge conflict, can you resolve it? I'm happy to merge when that's fixed.

ulion:
@ianstormtaylor Rebased on master.

ianstormtaylor:
Thanks @ulion!

PR #4054, opened 2021-01-22
Adds path parameter to NodeMatch

reinvanimschoot:

Feature

#### What's the new behavior?

The NodeMatch function provided in several functions on the  `Editor` and `Transform` now takes a second parameter which is the path of the node provided as the first parameter.

This is a non breaking change.

In the editor.ts, the `Editor.levels` and `Editor.nodes` functions are the only two that explicitly call `match`, so these have been updated to account for the `nodePath` as the second parameter.

In transforms/node.ts, all match functionality is deferred to one of the two functions above but they had a different type, not `NodeMatch<T>` but `(node: node) => boolean`. This has now been updated as well.

Fixes: #3967
Reviewers: @BrentFarese

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -476,7 +476,7 @@ export const Editor: EditorInterface = {
    */
 
   hasBlocks(editor: Editor, element: Element): boolean {
-    return element.children.some(n => Editor.isBlock(editor, n))
+    return element.children.some((n) => Editor.isBlock(editor, n))
   },
 
   /**
@@ -485,7 +485,7 @@ export const Editor: EditorInterface = {
 
   hasInlines(editor: Editor, element: Element): boolean {
     return element.children.some(
-      n => Text.isText(n) || Editor.isInline(editor, n)
+      (n) => Text.isText(n) || Editor.isInline(editor, n)
     )
   },
 
@@ -494,7 +494,7 @@ export const Editor: EditorInterface = {
    */
 
   hasTexts(editor: Editor, element: Element): boolean {
-    return element.children.every(n => Text.isText(n))
+    return element.children.every((n) => Text.isText(n))
   },
 
   /**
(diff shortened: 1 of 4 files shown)


packages/slate/src/interfaces/editor.ts
```diff
@@ -1575,6 +1575,6 @@ export const Editor: EditorInterface = {
  * A helper type for narrowing matched nodes with a predicate.
  */
 
-type NodeMatch<T extends Node> =
-  | ((node: Node) => node is T)
-  | ((node: Node) => boolean)
+export type NodeMatch<T extends Node> =
```
BrentFarese, line 1578: Will this type break existing implementations that have a `NodeMatch` w/o a `path` parameter?  Does everyone now need to update their `match` functions?  I think no but figured I'd double-check...

BrentFarese (approved):
I think this looks good.  @thesunny or @CameronAckermanSEL care to comment?

BrentFarese:
Note: This PR overlaps with #3968 and after merging we can close #3968.

packages/slate/src/interfaces/editor.ts
reinvanimschoot: It shouldn't, adding a second possible parameter like this does not necessarily mean you have to pass a second argument every time. Like when you call a `forEach`, you can call it with the element and index or just the element.

But if you really want me to double check, I can. Is there a simple way to make this version of the code run in my own project?

BrentFarese: No it should be fine there's ton's of `match` functions in Slate core with just `node` so TS would've thrown an error in Slate core is there was an issue..

reinvanimschoot:
@BrentFarese what are the following steps for this to become merged if I may ask?

BrentFarese:
> @BrentFarese what are the following steps for this to become merged if I may ask?

I would like at least some time for @thesunny or @CameronAckermanSEL or another maintainer(s) to review, then I can merge this in.  It's not a controversial change and seems good, just want an additional set of eyes.

reinvanimschoot:
I understand that a second pair of eyes is desired before merging but I'm a bit woried that this PR will suffer the same fate as the countless other PR's that are still open on this project. I fixed this code specifically because we have an urgent need for it in our own project. So I really hope another maintainer can take a look soon.

ianstormtaylor:
@BrentFarese it looks good to me. (But I haven't tested it locally.)

PR #4061, opened 2021-01-28
fix: `Operation.isOperationList`  just check value[0]

jacob-lcs:

fix a bug

#### What's the new behavior?

old code
```javascript
// packages/slate/src/interfaces/operation.ts
isOperationList(value: any): value is Operation[] {
    return (
      Array.isArray(value) &&
      (value.length === 0 || Operation.isOperation(value[0]))
    )
}
```
The old code just Determine if `value[0]` is a `Operation` Object

```javascript
 // packages/slate/src/interfaces/operation.ts
 isOperationList(value: any): value is Operation[] {
    if (!Array.isArray(value)) {
      return false
    }
    for (const valueItem of value) {
      if (!Operation.isOperation(valueItem)) {
        return false
      }
    }
    return true
},
```

Reviewers: @thesunny

### DIFF
--- a/packages/slate/src/interfaces/operation.ts
+++ b/packages/slate/src/interfaces/operation.ts
@@ -218,10 +218,15 @@ export const Operation: OperationInterface = {
    */
 
   isOperationList(value: any): value is Operation[] {
-    return (
-      Array.isArray(value) &&
-      (value.length === 0 || Operation.isOperation(value[0]))
-    )
+    if (!Array.isArray(value)) {
+      return false
+    }
+    for (const valueItem of value) {
+      if (!Operation.isOperation(valueItem)) {
+        return false
+      }
+    }
+    return true
   },
 
   /**
--- a/packages/slate/src/interfaces/operation.ts
+++ b/packages/slate/src/interfaces/operation.ts
@@ -218,15 +218,9 @@ export const Operation: OperationInterface = {
    */
 
   isOperationList(value: any): value is Operation[] {
-    if (!Array.isArray(value)) {
-      return false
-    }
-    for (const valueItem of value) {
-      if (!Operation.isOperation(valueItem)) {
-        return false
-      }
-    }
-    return true
+    return (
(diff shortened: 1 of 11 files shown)


packages/slate/src/interfaces/operation.ts
```diff
@@ -218,10 +218,15 @@ export const Operation: OperationInterface = {
...
-    )
+    if (!Array.isArray(value)) {
+      return false
+    }
+    for (const valueItem of value) {
+      if (!Operation.isOperation(valueItem)) {
+        return false
+      }
+    }
+    return true
```
BrentFarese, line 229: Why not something like this?

```suggestion
    Array.isArray(value) && value.every(val => Operation..isOperation(val));
```

jacob-lcs: 👍

jacob-lcs:
+ `Element. isElementList`
+ `Node. isNodeList`
+ `Operation.isOperationList`
+ `Text.isTextList`
All four have the same problem, and I've added test cases for them. Please review again, Thx! @BrentFarese

site/tsconfig.json
```diff
@@ -1,30 +1,20 @@
...
-  "exclude": [
-    "node_modules"
-  ],
-  "include": [
-    "next-env.d.ts",
-    "**/*.ts",
-    "**/*.tsx"
-  ]
-}
+{
```
BrentFarese, line 1: Are there any changes here or is this just Prettier?

BrentFarese (approved):
Looks good!  If you can reply on the TS config question, then this might be good to merge.  Thanks.

site/tsconfig.json
jacob-lcs: yes

jacob-lcs: > Are there any changes here or is this just Prettier?

The file contents are the same as before, I ran the command `yarn fix`

BrentFarese: Thanks!

PR #4065, opened 2021-02-01
Fix hook name `useSlate` -> `useState`

danseethaler:

Simple docs fix.


BrentFarese:
This is not a typo.  `useSlate` is a custom hook that is part of the codebase.  Thank you for looking tho!

Issue #3786, opened 2020-07-14
Proposal for Sponsored Open Source Android Support


thesunny:
I am considering a sponsored version of an Open Source Android Plugin for Slate for USD $25,000.

1. This is an alternative to the proposal for a Paid Android Plugin (also proposed by me) for USD $500 per license as described here https://github.com/ianstormtaylor/slate/issues/3573

2. The Android Plugin will be released as open source under the same license as Slate and will support at least the latest version of Android with similar features and limitations as the 0.47x Android Support 

3. Building Android support is a lot of work. We would love to be able to contribute for free (like Ian!), but the cost is high to burden alone. We contributed Android Support for Slate 0.47 which was working although not perfect and that cost about USD $40,000. That doesn’t include research we did prior to starting or the devices we had to purchase to test on.

4. It's uncertain how much it will cost to complete the 0.50x version of Slate Android. It may cost more or it may cost less. Either way, we are committing to completing it if we raise the funds. If it costs less, we will not be returning unused funds; if it costs more, we will cover the difference. Note that the 0.50x port is not trivial as another user worked for 3 months on it and it was never completed.

5. For clarity, we already spent USD $40,000, so even if the port costs $0, there will be no profit from this.

The port will be completed in partnership with Wayne Leroux with assistance from me. Wayne has built an impressive collaborative editor based on Slate (https://github.com/wleroux/slate-collaborative) so I have confidence in his ability to complete this project. He has also contributed other open source projects (https://github.com/wleroux?tab=repositories). I expect him to be able to complete the project; however, if for any reason he cannot, I will either complete it myself or will return the funds. Wayne and I intend to work together on a different Slate related project which would benefit from Android support.

If we are unable to raise the funds through sponsorship, we will revert to the paid model at USD $500 per license. For clarity, if we don't raise the funds through sponsorship and use a paid model but later receive the equivalent revenue in licensing fees from the paid model (USD $25,000 or 50 licenses), we will not then open source the code.

I recognize that this is a non-trivial amount to raise. I expect that companies that can afford to do so, will need to contribute more in order to meet this fundraising goal. To put the amount into perspective, this is the equivalent of contracting a senior developer for about 6 weeks. 3 organizations, contributing the equivalent of a contract valued at 2 weeks worth of work will reach the sponsorship goal (about USD $8000 each) or 50 smaller contributors paying what would be the licensing fee of USD $500.

Update: Now Live!

[...]

PR #4071, opened 2021-02-08
[WIP] Introduce Mutation Observer reconciliation strategy to support Android

clauderic:
This PR is adding a new _feature_, and is a work in progress. The code is still fairly scrappy, and should be reviewed as a proof-of-concept, which is why the PR is opened as a draft. In the future, if it moves forward, this PR should be broken down into smaller PRs.

I'm aware that there is currently [already an initiative in progress to bring Android support to Slate](https://github.com/ianstormtaylor/slate/issues/3786). Some of the code around text-diffing in this PR is based on that work.

The intent of the PR is to spark discussion in the hopes of helping to accelerate development of Android support for Slate, and move to developing the Android compatibility layer in the open so that it is open to community feedback.

#### What's the new behavior?

It introduces changes to support [different input reconciliation strategies that can be swapped out depending on the navigator and operating system.
](https://github.com/ianstormtaylor/slate/pull/4071/commits/0b42a32e8b7138d69cfb5885ba9449c3969815e5)
It also introduces a work in progress [Mutation Observer]() based [reconciliation layer to bring Android support to Slate](https://github.com/ianstormtaylor/slate/pull/4071/commits/ee42f5cc997998c8ebc04d940ad899fc501a05e1).

https://user-images.githubusercontent.com/1416436/107285339-fe26e800-6a2c-11eb-878b-eb67d91c44af.mp4

**Things that are currently working:**
- Distinguish between mutations that originate from user-input vs other mutations
- Update text (with some issues)
- Replace selection (with some issues)
- Delete selection
- Insert line break
- Delete backwards

**List of known issues (non-exhaustive):**
- Issues updating cursor position when inserting identical text (for example, inserting the letter `e` multiple times in the word `text` when the cursor is after `t`: `te|eext`)
- Copy/paste should not be tracked as user input by the MutationObserver layer, but currently is
- Need to ignore mutations that happen within void nodes (`contenteditable="false"`)
- Handle `readOnly` prop of editor
- Issues initially updating selection position on some Android devices
- Potentially need to batch observed mutations and flush manually, calling [`observer.takeRecords()`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/takeRecords) to process mutations that have not been processed by the callback yet

Feedback is welcome, and please report all issues so we can discuss them and find solutions. This PR is a rough first draft and there are likely many issues with it.

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -552,7 +552,8 @@ export const Editable = (props: EditableProps) => {
             if (
               !readOnly &&
               hasEditableTarget(editor, event.target) &&
-              !isEventHandled(event, attributes.onKeyDown)
+              !isEventHandled(event, attributes.onKeyDown) &&
+              !isEventHandled(event, inputStrategy.attributes.onKeyDown)
             ) {
               const { nativeEvent } = event
               const { selection } = editor
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -14,14 +14,15 @@ import throttle from 'lodash/throttle'
 import scrollIntoView from 'scroll-into-view-if-needed'
 
 import Children from './children'
+import { useInputStrategy } from '../input-strategy'
+import { hasEditableTarget, isEventHandled } from '../utils/helpers'
 import Hotkeys from '../utils/hotkeys'
 import {
+  HAS_BEFORE_INPUT_SUPPORT,
   IS_FIREFOX,
   IS_SAFARI,
-  IS_EDGE_LEGACY,
-  IS_CHROME_LEGACY,
 } from '../utils/environment'
-import { ReactEditor } from '..'
+import { ReactEditor } from '../plugin/react-editor'
 import { ReadOnlyContext } from '../hooks/use-read-only'
 import { useSlate } from '../hooks/use-slate'
 import { useIsomorphicLayoutEffect } from '../hooks/use-isomorphic-layout-effect'
@@ -31,8 +32,6 @@ import {
   DOMRange,
   isDOMElement,
   isDOMNode,
-  isDOMText,
-  DOMStaticRange,
   isPlainTextOnlyPaste,
 } from '../utils/dom'
 import {
@@ -43,14 +42,6 @@ import {
   IS_FOCUSED,
   PLACEHOLDER_SYMBOL,
 } from '../utils/weak-maps'
-
-// COMPAT: Firefox/Edge Legacy don't support the `beforeinput` event
-// Chrome Legacy doesn't support `beforeinput` correctly
-const HAS_BEFORE_INPUT_SUPPORT = !(
-  IS_FIREFOX ||
-  IS_EDGE_LEGACY ||
-  IS_CHROME_LEGACY
-)
 
 /**
  * `RenderElementProps` are passed to the `renderElement` handler.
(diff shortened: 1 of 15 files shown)


thesunny:
Hi @clauderic, thanks for looking over the Android PR and providing suggestions.

I don't have the bandwidth to dive into this Draft PR but did want to continue the conversation you started in #4200 about using a hooks approach. The general idea of moving all the logic into a hook is really interesting.

So first some context. I can't speak for @wleroux, however, one of the challenges with getting Android working for me was that the last 10-20% is tricky/finicky. It doesn't seem more efficient to restart that work, due to having to find and resolve the edge cases in multiple devices and Android versions (and not having one fix create a bug in another).

My impression, overall, is that we don't want to stop the process but perhaps there is a way we can incorporate this approach moving forward. I feel like the first step is to release a usable version.

Some questions I have that I think will help us evaluate the hooks approach are to understand where this PR is at and to understand your process with it:

*   Have you tried running the manual Android test suite against this draft PR from the 0.47.x branch? The test suite was created in response to bugs that were encountered during development. So if there was a bug, actions in the test suite were created to replicate the bug. Going through the test suite could tell us how far away this Draft PR is from being a working strategy as it is. It would be information to know at which stage this is at.
*   In the development process, which versions of Android/which devices were tested against? One of the challenges was getting 0.47.x in a state where a fix for one version, didn't create a bug in another.
*   Are there any issues that you think will be a challenge when using the same `Editable` for Android and non-Android? I feel like one of the issues with trying to get 0.47.x working was that it was difficult to separate what Android code specifically was doing when intermixed with the rest of the code. This might be different in 0.50+ as the code is cleaner. I'd like to hear the why of your choice including the benefits and particularly the challenges.
*   Finally, I wonder that if using hooks is a better long-term route, that the right path is to deliver the 4200 PR and then try to migrate it to use hooks. Not saying that is the right approach, but under the assumption that we want to use hooks, I'd prefer to start with a working version and convert it. For some additional information, Wayne initially started work on Android with a ground up rewrite of Android support for 0.50+, but due to the edge cases, he went back and did a port instead. We might find the same issue again.

Also, if you need some devices to work with, I think we'd be in a position where we can send some to you so you can do more testing. One of the challenges with Android is that I've found that I needed to keep several devices on hand to make sure a change didn't break one of them. Assuming you continue to have a high interest in contributing, I would like to make sure you have the right equipment available to you to help it along (if you don't already).

clauderic:
@thesunny this PR should be looked at as a proof of concept / roadmap towards implementing swappable input reconciliation layers in Slate without forking the `Editable.tsx` component.

> Are there any issues that you think will be a challenge when using the same Editable for Android and non-Android? I feel like one of the issues with trying to get 0.47.x working was that it was difficult to separate what Android code specifically was doing when intermixed with the rest of the code. This might be different in 0.50+ as the code is cleaner. I'd like to hear the why of your choice including the benefits and particularly the challenges.

This shouldn't be an issue, because in this PR the default input reconciliation of Slate is also extracted so that it doesn't run on Android.

> That the right path is to deliver the 4200 PR and then try to migrate it to use hooks.

I don't think that's a good idea. There are hundreds of PRs open for Slate, and you'd need to remind everyone to make changes for both forks of the `Editable` component for a sustained period of time.

clauderic:
Closing this PR in favour of https://github.com/ianstormtaylor/slate/pull/4224. Adding a mutation observer layer can happen in a follow-up PR

PR #4072, opened 2021-02-10
Memoize Node.isNodeList and Editor.isEditor

mitchellhamilton:

It's fixing a performance problem.

#### What's the new behavior?

Becasue of #4061, any call to Editor.isEditor/Element.isElement or etc. on the editor instance or an element with lots of children causes that entire tree to be walked which can be quite expensive.

Here's typing on the rich text example with ~800 root level nodes before:
Before Node.isNodeList memoization on rich text example

After:
After Node.isNodeList memoization on rich text example

It memoizes `Node.isNodeList` with a `WeakMap`. I'm happy to memoize `Element.isElement` and `Editor.isEditor` as well if people want that, I just did `Node.isNodeList` because it catches the bulk of the cost. It's worth mentioning that there's the potential for this to be wrong if a node list is a node list at one point and then it's mutated and it's no longer a node list, I feel like that's not a huge issue because these things shouldn't really be mutated but just wanted to raise it.

Fixes: #
Reviewers: @

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -273,6 +273,8 @@ export interface EditorInterface {
   withoutNormalizing: (editor: Editor, fn: () => void) => void
 }
 
+const IS_EDITOR_CACHE = new WeakMap<object, boolean>()
+
 export const Editor: EditorInterface = {
   /**
    * Get the ancestor above a location in the document.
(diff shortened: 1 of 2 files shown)


ianstormtaylor:
@mitchellhamilton good call! 

I feel like all those `is*` methods could definitely be cached too. Could merge this first to see if it causes any weirdness and then follow up with caching the others. It does seem super rare that something would be cease to be a valid `NodeList` (if anything it would likely be a bug in the userland code).

ianstormtaylor (approved):

mitchellhamilton:
Just cached Editor.isEditor as well because I saw that was taking a lot of time while pasting a lot of content since the operations list was large. Happy to add the others if these cause no problems.

bryanph:
How about having some kind of `entityType` property on these interfaces with `isX` methods?

TheSpyder:
My analysis pointed to `is-plain-object` as the main source of performance issues; reducing calls to it helps, but it's a safety net I don't think we need anymore. The types should already be able to guarantee that everything that needs to be an object is an object.

I've been tempted to locally patch `is-plain-object` to be no-op and see how much that helps performance but I haven't had time.

ianstormtaylor:
Thanks @mitchellhamilton!

PR #4073, opened 2021-02-11
Fix for Editor.positions not working correctly in all cases

spocke:
#### What's the new behavior?

Here is an example of the bug on the official https://www.slatejs.org/examples/links demo page:
link_caret_issue

When arrowing right over over the last character it skips to the next line. The caret is supposed to be placed after the `!` in this case.

There is an end case in the code where it will break the advance iteration of the current text node if the current string buffer is empty. But when the buffer is empty there might still be valid positions to return. The distance state determines if the previous advance overflowed the text node length and by how much so this change checks if the string is empty and that the distance is null meaning it will continue to advance and yield positions for the remaining distance.

This PR adds a few unit tests that previously was not passing. It also adds a few tests for things that doesn't seem covered but related to this change just to make sure it doesn't break those cases.

Fixes: #
Reviewers: @


ianstormtaylor:
Hey @spocke this looks like a good call.

Can you add tests for the `word/line` units? I'm curious to see if it still properly jumps over whole words/lines despite the inline being in the middle of the word. Thanks!

spocke:
Yes, my initial fix here was incorrect for the word and line units. I think it's just a matter of not stopping the advance from happening if there is still things in distance to advance. I changed the code and added two test for fragmented words and lines and it seems to work.

ianstormtaylor:
@spocke beautiful, thank you!

ianstormtaylor (approved):

jacob-lcs:
Hi~Can this PR be merged?

PR #4078, opened 2021-02-17
set_node operations did not invert correctly after serialization

TheSpyder:

Fixing a bug. Set node operations are not serializable as currently designed, which makes them difficult to use for collaboration.

When calling `Transforms.setNodes`, adding a property that doesn't already exist on the node sets the old value to undefined:

```
{
  type: 'set_node',
  path: [ 0 ],
  properties: { key: undefined },
  newProperties: { key: true }
}
```

This is not serializable (the `properties` object will be empty) resulting in an invalid operation when transmitted across a collaboration connection. This error does not manifest unless the operation is inverted on the other side of the connection (an uncommon but necessary requirement).

#### What's the new behavior?

With my change these operations use null:
```
{
  type: 'set_node',
  path: [ 0 ],
  properties: { key: null },
  newProperties: { key: true }
}
```

I have included two tests:
* After JSON stringify/parse, a `set_node` operation object is deep strict equal
* After JSON stringify/parse, a `set_node` operation can be used to invert the action it just performed

These are general properties every operation should be able to meet but I didn't want to add tests beyond the scope of this PR (however I could, we have similar tests in our collaboration codebase that verify this property).

One line change. `properties[k] = node[k]` becomes `properties[k] = node[k] ?? null`.


TheSpyder:
As an alternative to this, applying a `set_node` operation could compare `properties` and `newProperties` to determine what to do (missing attributes in `newProperties` would be removed) instead of just relying on the new value being loosely equal to `null` to remove attributes:
https://github.com/ianstormtaylor/slate/blob/4e667b91099bdfadb7e2b15175055183be38dc4c/packages/slate/src/transforms/general.ts#L200-L204

TheSpyder:
Sure, I can try that. I'm on leave until the end of next week but I'll give it a go soon.

TheSpyder:
@ianstormtaylor that worked really well! I ended up omitting undefined properties completely, which I think is a better result all around.

ianstormtaylor:
@TheSpyder nice, that's awesome! Thanks for taking the time to update it.

TheSpyder:
I missed one code path in the changes I made to avoid `null` in `set_node`, and I didn't have time to start integrating this update with OT in my app until today so I only just hit the bug. Removing `set_node` null checks in my app now leads to nodes having attribute value `null`.

This is because `Transforms.unsetNodes` uses null values:
https://github.com/ianstormtaylor/slate/blob/737aaa9cde2d4a2d6d64b83256aa5d9d1b5ce720/packages/slate/src/transforms/node.ts#L802-L804

Which means that the way this code is now written, it needs to not add to `newProperties` when `props[k]` is null:
https://github.com/ianstormtaylor/slate/blob/737aaa9cde2d4a2d6d64b83256aa5d9d1b5ce720/packages/slate/src/transforms/node.ts#L637-L641

I'll work on a new PR tomorrow.

TheSpyder:
I actually implemented most of the change that same day, but then forgot about it for 3 months. Oops. PR #4428 now open.

PR #4080, opened 2021-02-17
Fix decorations not rendering because of memoization

### LINKED ISSUE #3751: Decorate Nested Leaf Unsuccessful

Bug 🐛

##### Context
There are 3 leaf nodes and a React state `hovering` which is being updated. It seems like leaves are not re-rendering on React state change, or that we are losing leaf nodes or they are not being recognized as leaf nodes. This may be some optimization or normalization under the hood?

However, I have managed it to work as expected in a very specific circumstance, IFF `onChange` is triggered on the target leaf, the selection is in the target leaf node, and React state is updated. Just changing the React state does not trigger it, typing in the leaf node is required.

##### Sandbox of issue reproduction

https://codesandbox.io/s/slate-repo-irmerk-25-06-2020-sq8r0

##### Steps to reproduce
1. Hover over italicized text, note that `25` is not bold
2. Click and type within the italicized text, note that `25` is not bold
3. Click and type within `25`, note that `25` is not bold
4. Click in `25`
5. Hover over italicized text
6. Type anything
7. Note the bold

Slate: **0.58.0**
Browser: **Chrome**
OS: **Mac**

React `hovering` state change should re-render all 3 leaves. It should decorate the middle (`variable`) leaf node.

taneliang:

Fixes #3751

#### What's the new behavior?

Repro of old issue: https://codesandbox.io/s/slate-3751-repro-xte8r

I've modified the plaintext example to demo the fix; we should probably revert/delete that commit (5a03575) before merging.

2 things were changed:

1. Copied `isRangeListEqual` from `element.tsx` into `text.tsx`. I'm not sure whether duplicating the code is a great idea, but it seems like `isRangeListEqual` is only meant to be used with arrays of decorations.
2. Changed `Range.equals` to do a deep equal of the remaining properties in the range object. A few potential issues with this:
    1. I'm not sure if we want `Range.equals` to care about other props, but functions like `intersection` forward props so there's at least some implicit support for additional props
    2. `Range.equals` is now a potentially expensive operation
    3. slate now has lodash as a dependency

Fixes: #3751

### DIFF
--- a/packages/slate/test/interfaces/Range/equals/equal.tsx
+++ b/packages/slate/test/interfaces/Range/equals/equal.tsx
@@ -10,6 +10,7 @@ export const input = {
       path: [0, 1],
       offset: 0,
     },
+    placeholder: 1,
   },
   another: {
     anchor: {
@@ -20,6 +21,7 @@ export const input = {
       path: [0, 1],
       offset: 0,
     },
+    placeholder: 1,
   },
 }
 export const test = ({ range, another }) => {
--- a/packages/slate/test/interfaces/Range/equals/not-equal.tsx
+++ b/packages/slate/test/interfaces/Range/equals/not-equal.tsx
@@ -10,6 +10,7 @@ export const input = {
       path: [0, 4],
       offset: 7,
     },
+    placeholder: 1,
   },
   another: {
     anchor: {
@@ -20,6 +21,7 @@ export const input = {
       path: [0, 1],
       offset: 0,
     },
+    placeholder: 0,
   },
 }
 export const test = ({ range, another }) => {
--- a/site/examples/plaintext.tsx
+++ b/site/examples/plaintext.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo } from 'react'
+import React, { useState, useMemo, useCallback } from 'react'
 import { Node, createEditor } from 'slate'
 import { Slate, Editable, withReact } from 'slate-react'
 import { withHistory } from 'slate-history'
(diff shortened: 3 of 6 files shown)


TheSpyder:
Does it need to be deep equal with `lodash`? Could it just be flat equal of the top-level attributes as other similar comparisons do (e.g. [text nodes](https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/interfaces/text.ts#L31-L59))?

packages/slate/test/interfaces/Range/equals/not-equal.tsx
```diff
@@ -20,6 +21,7 @@ export const input = {
       path: [0, 1],
       offset: 0,
     },
+    placeholder: 0,
```
TheSpyder, line 24: You'll need a dedicated test for extra attributes as `anchor` and `focus` are already not-equal in this test

taneliang:
Thanks! I didn't realize there was a similar comparison in text nodes. I'll change it to that for consistency

TheSpyder:
Ian has been fairly adamant that attributes in the Slate model are flat, rather than adding the complexity (and potential performance cost) of nested object support :)

taneliang:
Right, that makes sense!

packages/slate/test/interfaces/Range/equals/equal.tsx
```diff
@@ -20,6 +21,7 @@ export const input = {
       path: [0, 1],
       offset: 0,
     },
+    placeholder: 1,
```
spocke, line 24: Maybe call this `custom` or `foo` or something placeholder seems specific to your use case. So that it's obvious that it's custom data or better yet split it out into it's own `equals-user-marks.tsx` test file. That way both the just the anchor/focus comparison is tested and the one that has extra custom properties on it.

packages/slate/test/interfaces/Range/equals/other-props-not-equal.tsx
```diff
@@ -0,0 +1,30 @@
+import { Range } from 'slate'
+
+export const input = {
```
spocke, line 3: Maybe also add a test when the ranges has a different number of keys they should also not equal and on both sides. Just because it's working now it might not work tomorrow if someone tries to optimize that equals code or something.

site/examples/plaintext.tsx
```diff
@@ -1,23 +1,29 @@
...
   const editor = useMemo(() => withHistory(withReact(createEditor())), [])
+  const [placeholder, setPlaceholder] = useState(1)
+  const newPlaceholder = useCallback(() => setPlaceholder(p => p + 1), [])
+
   return (
-    <Slate editor={editor} value={value} onChange={value => setValue(value)}>
-      <Editable placeholder="Enter some plain text..." />
-    </Slate>
+    <>
+      <button onClick={newPlaceholder}>New Placeholder</button>
```
spocke, line 14: This should probably be it's own example if it even makes sense. It's more of a demo of a bug fix can't be that common to change the placeholder.

taneliang: Yeah, my intention was to revert all the changes to this example before merging this PR - I just wanted a quick and easy way to demo this PR, and I don't think we'll actually want to have a changing placeholder example on master

spocke: Oops I should have read the description more closely.

taneliang: No worries! I probably should’ve made it more prominent

ulion:
I would choose to use lodash deep compare to replace Range.equals and keep Range.equals not change.
Since the Range.equals change is not trivial, it indeed should only care about what 'Range' care, not other properties, even if you want to change Range.equals, should add the 3rd parameter like strict with default value false to keep the default behaviour not change.

Issue #4076, opened 2021-02-13
Features for Search Highlighting


Lavcodes:
#### Do you want to request a _feature_ or report a _bug_?
Feature

#### What's the current behavior?

- The **Search Highlighting** feature is case sensitive, which in most cases the user may not prefer. 

- Also the text highlight colour is quite subtle and hard to see (however this may just be a personal preference). An additional feature allowing the user to choose the color of text highlight may be desired (or probably just changing the current highlight to a slightly darker shade of yellow (as in most browsers).

I have attached screenshots for the same below. 
_Case Sensitive_
![CaseSensitive](https://user-images.githubusercontent.com/59924327/107849760-70c80700-6e23-11eb-9253-0fdc96064353.png)
_Existing colour of highlight_
![ColourHighlight](https://user-images.githubusercontent.com/59924327/107849763-77567e80-6e23-11eb-9d57-afd2e3be93a5.png)

Slate: 0.XX.X
Browser: Chrome / Safari / Firefox / Edge
OS: Mac / Windows / Linux / iOS / Android

#### What's the expected behavior?
_Standard chrome results for the same_
![StandardImplementation](https://user-images.githubusercontent.com/59924327/107849766-7aea0580-6e23-11eb-9f01-082780edfd0e.png)

I would love to add these mini functionalities, if approved.

ianstormtaylor:
Those sound like great fixes, thanks @Lavcodes!

Lavcodes:
I was trying to add these features but I'm not able to figure out how to send the color input by the user to the function leaf.
I am making these changes in examples>search-highlighting.tsx
Could someone please help out. My color-picker is in search-highlighting function. I am using hooks to setColor value.

PR #4086, opened 2021-02-20
Making search-highlighting case insensitive

Lavcodes:
**Description**
#4076  Makes Search Highlight Case insensitive. I've used case insensitive RegExp str.split() to achieve that. 

**Issue**
#4076 

**Example**
These are the search results before changes
BeforeChanges

These are the search results after changes
AfterChanges

**Context**
I've modified text.split(search) method. I've made use of case insensitive RegExp. I've included additional checks to make sure the input string that begins or ends with '\' is handled as that was throwing errors due to how RegExps work.
**Checks**
- [ YES ] The new code matches the existing patterns and styles.
- [ YES ] The tests pass with `yarn test`.
- [ YES ] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [ YES ] The relevant examples still work. (Run examples with `yarn start`.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,5 +92,8 @@
     "slate-react": "*",
     "source-map-loader": "^0.2.4",
     "typescript": "^3.7.2"
+  },
+  "dependencies": {
+    "react-pick-color": "^1.0.1"
   }
 }
--- a/site/examples/search-highlighting.tsx
+++ b/site/examples/search-highlighting.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
 import React, { useState, useCallback, useMemo } from 'react'
 import { Slate, Editable, withReact } from 'slate-react'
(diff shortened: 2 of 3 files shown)


ianstormtaylor:
Hey @Lavcodes thanks for taking a stab at this! It looks like there is some extra cruft still left in the PR (imports that aren't used, dependencies that are added but not used, prettier being disabled, etc.) Can you clean these things up? If so I'm happy to merge.

dylans (changes requested):
This PR needs to be updated to only include the relevant changes. Let us know if you have time or if we should close it out. Thanks.

dylans:
Given the merge conflicts and extraneous changes in this PR, and lack of activity, I'm going to close it out. Feel free to raise a new PR that addresses these issues.

Issue #4070, opened 2021-02-07
Export FLUSHING weak map for some cases


Xheldon:
#### Do you want to request a _feature_ or report a _bug_?

_feature_

#### What's the current behavior?

I want log history change, but I can't find the right time point. 

There is a fact that one (user) operations may include many `apply` function invoke, like press enter key in paragraph or something else, but I can't find the right chance to record the multiple && consecutive operations.

#### What's the expected behavior?

The expected `FLUSHING` usage should like:

```js
import {FLUSHING} from 'slate';

const {apply} = editor;
let oldState = editor.children;
editor.apply = op => {
    // other code log `op` ...
    apply(op);
    if (FLUSHING.get(editor)) { // this means that the operation with multiple apply function invoke is finished, so I can log the change 
      log(editor.operations);
      const diff = jsondiff(oldState, editor.children);
      log(diff);
      oldState = editor.children;
    }
    // other code...
}
```

I read the slate source code and know that `editor.apply` function have this FLUSHING weak map, but it's not export. If need a PR, I'm very glad to contribute, Thanks!

@ianstormtaylor

BrentFarese:
Yes please submit a PR.  Or, to export things that are not exported by Slate, we have just used [patch-package](https://www.npmjs.com/package/patch-package) and it works very well.  Do you want to just do that instead vs. waiting for a release?  If so, feel free to close the issue.  Thanks!

Xheldon:
No longer need this PR, my needs may be too niche and have been solved by a local modification of the source code, thanks!

PR #4090, opened 2021-02-22
Export FLUSHING & other weak maps for some cases

Xheldon:
**Description**
I want log history change, but I can't find the right time point.

There is a fact that one (user) operations may include many apply function invoke, like press enter key in paragraph or something else, but I can't find the right chance to record the multiple && consecutive operations.

**Issue**
Fixes:  https://github.com/ianstormtaylor/slate/issues/4070

**Example**
The expected FLUSHING or other weak-maps which used in slate inner, usage should like:

```js
import {FLUSHING} from 'slate';

const {apply} = editor;
let oldState = editor.children;
editor.apply = op => {
    // other code log `op` ...
    apply(op);
    if (FLUSHING.get(editor)) { // this means that the operation with multiple apply function invoke is finished, so I can log the change 
      log(editor.operations);
      const diff = jsondiff(oldState, editor.children);
      log(diff);
      oldState = editor.children;
    }
    // other code...
}
```

**Context**
I read the slate source code and know that editor.apply function have this FLUSHING weak map, but it's not export

**Checks**
- [ x] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)


ianstormtaylor:
Hey @Xheldon thanks for this. I see what you mean, although I'd rather see this exposed as something like an `Editor.isFlushing` method instead of exposing the weak maps directly. Can you make that change? Thanks!

Xheldon:
No longer need this PR, my needs may be too niche and have been solved by a local modification of the source code, thanks!

PR #4091, opened 2021-02-23
Proposal for Remove useCallback from walkthroughs documentation

pukingrainbows:
I have seen in the walkthrough the use of `useCallback` but I don't think are need it, and add a small complexity while understanding the concept, if we moved out the function outside of the component will be stable and will not require to maintain either referential identity or calculated complex logic. 

If people think this is a good idea I can go and update all documentation to remove unnecessary `useCallback` from the example.

unless I'm missing something.

**Description**
Improving documentation

**Issue**
Removing little complexity from the example.

**Example**
I don't think the use of useCallback is necessary on the examples for the walkthroughs 

**Context**

**Checks**

### DIFF
--- a/docs/walkthroughs/04-applying-custom-formatting.md
+++ b/docs/walkthroughs/04-applying-custom-formatting.md
@@ -7,6 +7,15 @@ In this guide, we'll show you how to add custom formatting options, like **bold*
 So we start with our app from earlier:
 
 ```jsx
+function renderElement(props) {
+  switch (props.element.type) {
+    case 'code':
+      return <CodeElement {...props} />
+    default:
+      return <DefaultElement {...props} />
+  }
+})
+
 const App = () => {
   const editor = useMemo(() => withReact(createEditor()), [])
   const [value, setValue] = useState([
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Hey @pukingrainbows that change sounds good to me. Can you use the `const = ` syntax if you make that change everywhere? Thanks!

docs/walkthroughs/04-applying-custom-formatting.md
```diff
@@ -7,6 +7,15 @@ In this guide, we'll show you how to add custom formatting options, like **bold*
 So we start with our app from earlier:
 
 ```jsx
+function renderElement(props) {
```
dylans, line 10:
```suggestion
const renderElement = (props) => {
```

dylans (changes requested):

dylans (approved):
As noted in my comment, I'm happy to approve a PR for this. Given the time since this was opened, I'm going to go ahead and land this change and close the PR. We'll accept PRs for any other similar changes in a more timely manner.

Issue #4098, opened 2021-02-26
Slate-react throwing error after using Babel with loose mode


dchikulaev:
**Description**
Can't edit text with slate-react after using Babel with loose mode. Throwing error "Cannot read property 'removeChild' of undefined".

**Recording**

https://user-images.githubusercontent.com/79687911/109305880-336e6b00-7860-11eb-82b2-0b4123f50673.mp4

**Sandbox**
https://babeljs.io/repl#?browsers=defaults%2C%20ie%2011%2C%20not%20ie_mob%2011%2C%20chrome%20%3E%2049&build=&builtIns=false&spec=false&loose=true&code_lz=MYewdgzgLgBAhjAvDA2gOg6MUCm2JoCOArjgE4CeAyjgDY7BQhkCCttAFAOQoAmcUOAFoItATiEAvciCEB3AJa8oACwC6XAJQAaGBjRZc-IqUo16jZm049DeXLwWCARvUQAzOLQg4NmtUA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.13.7&externalPlugins=

**Steps**
To reproduce the behavior:
1. Build React project with editor.
2. Click on Textarea of editor.
3. See error "Cannot read property 'removeChild' of undefined".

**Expectation**
I can edit text.

**Environment**
- Slate Version: [0.59]
- React: [16.9.0]
- Babel: [7.12.3]

PR #4099, opened 2021-02-26
Fixed converting querySelectorAll results to array for Babel loose mode.

dchikulaev:
**Description**
Slate-react support of Babel loose mode.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4098

**Example**

https://user-images.githubusercontent.com/79687911/109308385-c4931100-7863-11eb-8252-c0433b3baf39.mp4

https://user-images.githubusercontent.com/79687911/109310679-ac70c100-7866-11eb-935c-ba665e21a40f.mp4

**Context**
After using Babel in ReactEditor from slate-react querySelectorAll does not actually return an array and concat just adds as one element instead of expanding everything. Babel convert original code:
`var removals = [...contents.querySelectorAll('[data-slate-zero-width]'), ...contents.querySelectorAll('[contenteditable=false]')];`
into
`var removals = [].concat(contents.querySelectorAll('[data-slate-zero-width]'), contents.querySelectorAll('[contenteditable=false]'));`
so as result `removals` has 2 items of NodeList. 

Solution is explicitly converting of NodeList to array before concat.

**Checks**

### DIFF
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -433,8 +433,12 @@ export const ReactEditor = {
         range.setEnd(nearestNode, nearestOffset)
         const contents = range.cloneContents()
         const removals = [
-          ...contents.querySelectorAll('[data-slate-zero-width]'),
-          ...contents.querySelectorAll('[contenteditable=false]'),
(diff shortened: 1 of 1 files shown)


ianstormtaylor:
Beautiful, thank you @dchikulaev!

PR #4107, opened 2021-03-02
fix: cut do not remove void element when only select void element

githoniel:
**Description**
cut do not remove void element when only select void element

**Issue**
Fixes:  no

**Example**

see context

**Context**

`onCut`  event only deleteFragment when Range.isExpanded and will do nothing when select only one void element.

this pr will remove void element when Range.isCollapsed and selection is in a void element

**Checks**

there is no tests for `slate-react`


williamstein:
For what it is worth, I tested this and the behavior seems right to me.

ianstormtaylor:
Thanks @githoniel!

Issue #3966, opened 2020-11-02
Traversing children in renderElement


Mangatt:
#### Do you want to request a _feature_ or report a _bug_?

Feature

#### What's the current behavior?

Children is always single component, not collection.

Slate: 0.59.0

#### What's the expected behavior?

It would be nice to be able to traverse children provided in `renderElement` functions (with [React.Children](https://reactjs.org/docs/react-api.html#reactchildren)).

That way you can for example visually group/separate elements without actually changing document structure.

Mangatt:
Alternatively, there could be `renderChildren` prop that would be used in `Children.jsx`, although traversing children directly with React is more by-the-book approach.

Mangatt:
Any thoughts on this? This could greatly simplify creating plugins that need enforced structure like lists or tables.

ianstormtaylor:
This would require someone researching whether it's possible. `React.Children` isn't an abstraction that always works.

My guess is that this won't really be possible since Slate expects a specific structure to the children, so allowing that structure to be dynamically changed wouldn't work. For lists/tables the structure would likely need to be enforced in a schema instead.

Mangatt:
I'm aware of that, but React Components can have any internal HTML structure and Slate still works. This makes me believe that it should be possible, but I don't know Slate internals that well.

> If children is a Fragment it will be treated as a single child and not traversed.

Theoretically it should be matter just of changing `<React.Fragment />` into simple `array`. Good thing is it isn't event a breaking change.

Mangatt:
Ok, I've just tested this and Slate appears to be just fine with structure changes on the fly.

Little bit problematic is rather fact that rendered children can't be encapsuled in `Children` component, but have to be rendered one by one in parent, otherwise React sees just single `Children` component while traversing with `React.Children.map`.

Simple solution would be to change `Children` component into plain function `getChildren` or hook `useChildren` that would return array with children to parent component. It does not change anything nor it is breaking change since `Children` is not public endpoint.

If maintainers are ok with this, I can push PR.

ianstormtaylor:
@Mangatt nice, a `useChildren` hook sounds like a good way to do it. A PR would be great.

Can you explain the structure you'd like to impose by iterating children, or provide any pseudo-code snippets of the kind of functionality you're trying to achieve? (Often helps in PRs for people to understand what you're asking for.)

Mangatt:
Great, I'll do that next week.

[...]

PR #4108, opened 2021-03-02
Use children

Mangatt:
**Description**
This is tiny change from `Children` component to `useChildren` hook. This is not breaking because `Children` was not public endpoint. This change allows traversing children with `React.Children`.

**Issue**
Fixes: #3966 

**Checks**

(Other changes were performed by` yarn fix`, only relevant file changes are in `use-children.tsx`, `editable.tsx` and `element.tsx`.)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-core.md
+++ b/.github/ISSUE_TEMPLATE/bug-core.md
@@ -1,4 +1,5 @@
 ---
+
 name: "\U0001F6A8 Bug: Core"
 about: A bug that occurs in Slate's core logic, on all platforms
 title: ''
--- a/.github/ISSUE_TEMPLATE/bug-core.md
+++ b/.github/ISSUE_TEMPLATE/bug-core.md
@@ -4,10 +4,7 @@ about: A bug that occurs in Slate's core logic, on all platforms
 title: ''
 labels: bug
 assignees: ''
-
----
-
-**Description**
+---**Description**
 A clear and concise description of what the bug is.
 
 **Recording**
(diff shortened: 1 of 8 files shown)


ianstormtaylor:
Beautiful PR, thanks @Mangatt!

ianstormtaylor (approved):

Mangatt:
Thanks, @ianstormtaylor , glad to do my part!

Mangatt:
Do you think @thesunny that this could be merged into next `next` / `0.60.5` update? As mentioned, it does not change anything for anyone, but opens large possibilities for new plugins. And it's already approved by Ian.

thesunny:
I'm going to wrap up TypeScript within the next week. I think we can have more regular merges soon afterwards.

You may need to check the plugin against the TypeScript changes but assuming that's done, I don't see a reason why it couldn't be added relatively quickly. I'm hoping we can get everything out of `@next` released into `@latest` releases as well.

Mangatt:
Do you think that can be merged anytime soon, @thesunny ?

thesunny:
Hi @Mangatt,

I have no problem with merging the code but for some reason it seems to be trying to change all of Ian's `.github` templates. Do you have any idea what's going on there?

If we can fix this, I can merge.

ianstormtaylor:
Might be a Prettier issue and we need to add the templates to `.prettierignore`? Not sure.

thesunny:
I added the templates to `.prettierignore` yesterday so that I could add prettier back to linting.

I think you're right though and Mangatt merged before that was in place.

I wonder if we could manually copy and paste those templates into this PR and the issue would go away? I'm afraid I'm one of those people that don't really know how git works underneath so not sure if that is the right way to fix it.

thesunny:
Merged this PR with reverted `.github` files in this PR #4152

PR #4109, opened 2021-03-03
added support for deleting selection properties by setting them to null

spocke:
**Description**
This adds support for removing custom selection properties using `Transforms.setSelection` we use the editor.selection object to store implementation specific selection states. Without this change it's impossible to remove that state once it's set.

**Context**
This will update or delete the properties on editor.selection when the `set_selection` is applied instead of just adding. It also clones the newProperties if the current selection is null this is because immer will Object.freeze the original input selection range since that was set as a direct reference on the editor. It's written to be small not efficient so it does both delete/set in for loop even when it goes from null to selection but I think that transition is rare, it's far more common that new properties are being set on an existing selection.

**Checks**

### DIFF
--- a/packages/slate/src/transforms/general.ts
+++ b/packages/slate/src/transforms/general.ts
@@ -212,18 +212,32 @@ export const GeneralTransforms: GeneralTransforms = {
 
         if (newProperties == null) {
           selection = newProperties
-        } else if (selection == null) {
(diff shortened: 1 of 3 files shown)


ianstormtaylor:
Thanks @spocke! Interesting use case and great PR.

PR #4113, opened 2021-03-05
[WIP] Made changes to unhangRange function

ridhambhat:
**Description**
When a selection is made on browsers like Firefox which share its edges with edges of a node using double click, its range leaked onto the adjacent nodes as well like leaking to previous text node with offset equal to the length of node value or leaking to next node with offset=0.

**Issue**
This fixes the [issue #3923](https://github.com/ianstormtaylor/slate/issues/3923)

**Example**
Old behavior:
</img>
New behavior:
</img>

**Context**
In the unhangRange function which is a part of editor interface, the old check used a condition ```(start.offset!=0)```
This did not check the particular condition when ```start.offset ===start.length``` i.e. the range leaks to the end of start node and doesn't actually hold any text/element of start node. This causes the range to not be identified in the first place and then when a property is set, it sets the property to the nodes where the range leaked as well (which were both adjacent nodes in above example).

To correct this, I first included the condition ```!(startPointAtEndOfNode || endPointAtStartOfNode)```, which says that we cannot confirm an un-hung range if the start point is at the end of the first node or the end point is at the start of a node.

Now the logic which was previously used for choosing the previous text node if in case it leaked from the end, I have implemented a similar one for the start.

Then I have just used this unhangRange function in ```toSlateRange``` as as well to clip the selection as soon as it is made in react.

**Checks**

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -1505,21 +1505,32 @@ export const Editor: EditorInterface = {
   ): Range {
     const { voids = false } = options
     let [start, end] = Range.edges(range)
+    const startPointAtEndOfNode = Editor.isEnd(editor, start, start.path);
+    const endPointAtStartOfNode = Editor.isStart(editor, end, end.path);
 
     // PERF: exit early if we can guarantee that the range isn't hanging.
-    if (start.offset !== 0 || end.offset !== 0 || Range.isCollapsed(range)) {
+    if (!(startPointAtEndOfNode || endPointAtStartOfNode) || Range.isCollapsed(range)) {
       return range
     }
 
+    const startBlock = Editor.above(editor, {
+      at: start,
+      match: n => Editor.isBlock(editor, n),
+    })
+    const startBlockPath = startBlock ? startBlock[1] : []
     const endBlock = Editor.above(editor, {
       at: end,
       match: n => Editor.isBlock(editor, n),
     })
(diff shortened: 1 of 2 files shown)


packages/slate/src/interfaces/editor.ts
```diff
@@ -1505,21 +1505,32 @@ export const Editor: EditorInterface = {
   ): Range {
     const { voids = false } = options
     let [start, end] = Range.edges(range)
+    const startPointAtEndOfNode = Editor.isEnd(editor, start, start.path);
+    const endPointAtStartOfNode = Editor.isStart(editor, end, end.path);
```
timbuckley, line 1509: Small nitpick: Keep styling consistent with Eslint. If you use VSCode, you could maybe use an Eslint plugin to autoformat on save.
```suggestion
    const startPointAtEndOfNode = Editor.isEnd(editor, start, start.path)
    const endPointAtStartOfNode = Editor.isStart(editor, end, end.path)
```

ridhambhat: got it.. will correct that in next commit.

ridhambhat:
Changes in [unhang function](https://github.com/ianstormtaylor/slate/blob/1a4c67f5fad60b3a8df94c9a29568c02bfa1c18e/packages/slate/src/interfaces/editor.ts#L1522) would result in change in functionality in some ways which is not desired.

PR #4114, opened 2021-03-07
Add Deno SSR support

oyzhen:
**Description**
Suppresses warnings on SSR when using useLayoutEffect by falling back to useEffect when running on Deno.

Enhanced for #3262

**Checks**
- [*] The new code matches the existing patterns and styles.
- [*] The tests pass with `yarn test`.
- [*] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [*] The relevant examples still work. (Run examples with `yarn start`.)


ianstormtaylor:
@oyzhen can you explain why Deno needs to be handled specially?

oyzhen:
`Deno` has its own [`window`](https://doc.deno.land/builtin/stable#Window), but doesn't support HTML DOM or `document`. 
While using `slate` in `Deno` SSR (e.g. via [`aleph`](https://alephjs.org/)), React will warn about usage of `useLayoutEffect` on server. Falling back to `useLayout` in `Deno` SSR just works.

packages/slate-react/src/hooks/use-isomorphic-layout-effect.ts
```diff
@@ -4,4 +4,6 @@ import { useLayoutEffect, useEffect } from 'react'
  * Prevent warning on SSR by falling back to useEffect when window is not defined
  */
 export const useIsomorphicLayoutEffect =
-  typeof window !== 'undefined' ? useLayoutEffect : useEffect
+  typeof window !== 'undefined' && typeof Deno === 'undefined'
```
clauderic, line 7: Checking for `Deno` shouldn't be needed. Try to align this conditional with the same check React does internally:
https://github.com/facebook/react/blob/master/packages/shared/ExecutionEnvironment.js

oyzhen: Good idea.

clauderic (approved):
This PR should be merged, it is the correct solution. Perhaps you can rename your PR to better reflect the changes being made @oyzhen

ianstormtaylor:
Thanks @oyzhen!

merko30:
Hi, although it's not related to Deno, I'm receiving the error regarding `useLayoutEffect`. By the way, I'm using Next.

Issue #3185, opened 2019-12-01
Firefox does not fire `beforeinput` events


ianstormtaylor:
Firefox does not currently support the `beforeinput` event, which Slate uses under the covers to be notified whenever the user triggers an input into the editor (inserts, deletes, cuts, pastes, drops, etc.). These `beforeinput` events are cancellable and we use them to then do the Slate-equivalent of whatever the event was.

Right now we hack around this for Firefox and handle some of the common input cases. However, there's only so much we can do. And input events like `formatBold`, or `insertFromComposition` are hard to polyfill. (There might be ways to improve our existing ones, if you see them please PR!)

This is also the reason we disable `autoCorrect` and `spellCheck` in Firefox on the `contenteditable` element, since we have no easy way of determining when those changes occur.

Here are the bug reports… (please vote for and follow them!)

- [x] https://bugzilla.mozilla.org/show_bug.cgi?id=970802
- [ ] https://bugzilla.mozilla.org/show_bug.cgi?id=1305387
- [x] https://bugzilla.mozilla.org/show_bug.cgi?id=1449831

Once these are supported Firefox will get improved editing support (autocorrect, spellcheck, IME composition, B/U/I toolbars, etc.). And we'll be able to reduce bundle size for everyone by removing the current Firefox-specific hacks.

minwe:
Duplicated text inserted when using IME in Firefox:

```
Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:70.0) Gecko/20100101 Firefox/70.0
```

![ime-firefox70](https://user-images.githubusercontent.com/4370770/69947766-4f824000-1529-11ea-845d-2b978dbd8146.gif)

It seems text inserted both `onCompositionEnd` and `onBeforeInput`:

![image](https://user-images.githubusercontent.com/4370770/69947815-69bc1e00-1529-11ea-80a6-6bfcbedd9f8b.png)

![image](https://user-images.githubusercontent.com/4370770/69947862-80627500-1529-11ea-9ea2-bc03968984f5.png)

It's better to use **feature detection** but not **UA detection**.

fletcher91:
https://bugzilla.mozilla.org/show_bug.cgi?id=970802 and https://bugzilla.mozilla.org/show_bug.cgi?id=1449831 have been marked as fixed :+1:

ClubicTech:
[...]

Issue #3800, opened 2020-07-24
Slate onChange is not called in IE11, selection doesn't affect state too. Rich Text Example doesn't work fully.


vshmelev:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?
Trying to run Rich Text example in Internet Explorer 11. Added necessary polyfills:

"babel-polyfill": "^6.26.0",
"element-closest-polyfill": "^1.0.0",

And was able to open the example script in IE11 with no errors. However seems like editing doesn't work properly, at least:

1. Slate onChange is not invoked
2. Selection change doesn't affect the state, as a result toolbar buttons are not enabled and formatting not applied.

![mq0AS7rx5K](https://user-images.githubusercontent.com/6876128/88389834-ed2fc580-cdbf-11ea-9bbd-1d8778a51975.gif)

The same works just fine in the Chrome.

After quick debugging I noticed that in editable.tsx const HAS_BEFORE_INPUT_SUPPORT includes FF/Edge Legacy and Chrome legacy only, so it will use native beforeinput event which IE11 definitely doesn't support. However even if I amend user agent from Emulation tab of IE Dev Tools and use FF for example to force using React onBeforeInput it still doesn't work so probably the issue is something else.

Slate: 0.58.3
Browser: Internet Explorer 11 Version 11.388.19041.0
OS: Windows 10 Pro

#### What's the expected behavior?
At least such major feature as text editing/formatting must work in IE.

albertoblaz:
Having the same problem here

baluubas:
For anyone else ending up here I've gotten IE11 working by adding additional DOM polyfills (on top the ECMAScript polyfills):

- Element.closest
- Node.parentElement (IE only have partial support on some node types)
- Selection.setBaseAndExtent
- Node.contains (IE only have partial support on some node types)
- NodeList.foreach

cayuu:
@baluubas thank you :pray: and this will be amazing if I can get it to work, but I'm stuck on sourcing all the polyfills you mention.

The following are available through polyfill.io service: 
https://polyfill.io/v3/polyfill.min.js?features=Element.prototype.closest%2CNode.prototype.contains%2CNodeList.prototype.forEach

- [Element.closest](https://github.com/Financial-Times/polyfill-library/tree/master/polyfills/Element/prototype/closest)
- [Node.contains](https://github.com/Financial-Times/polyfill-library/tree/master/polyfills/Node/prototype/contains)
[...]

PR #4118, opened 2021-03-09
Make HAS_BEFORE_INPUT_SUPPORT flag more generic.

kamilkazmierczak:
**Description**
HAS_BEFORE_INPUT_SUPPORT flag is currently resolved based on type of browser but should be detected for every browser by itself.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/3185, https://github.com/ianstormtaylor/slate/issues/3800

**Context**
It's impossible to use slate on IE11 because this flag currently does not include IE11, also it is not proper way of detecting whether feature is enabled in browser or not. I left IS_CHROME_LEGACY because of existing comment saying that feature is available but is not working properly.
Detection mechanism was taken from https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/getTargetRanges#feature_detection

**Checks**


ianstormtaylor:
Thanks @kamilkazmierczak!

Issue #3748, opened 2020-06-24
Transforms.wrapNodes and Transfroms.unwrapNodes causing the entire parent node to re-render


zbeyens:
### Description

Transforms.wrapNodes and Transfroms.unwrapNodes causing the entire parent node to re-render

### How to reproduce

https://codesandbox.io/s/slate-plugins-reproductions-bug-177-vo206

Add list items right above embed and now add nested list items. embed component gets rerendered and causes another network request.

### Expected behavior
If the URL for embed is not changed component should not be re-rendered so we can avoid network request Also, re-render impacts UX.

### Workaround

The current workaround is to avoid using these transforms.

More informations there https://github.com/udecode/slate-plugins/issues/177

It may be related to https://github.com/ianstormtaylor/slate/issues/3740

PR #4120, opened 2021-03-11
Fix move_node caused full re-render.

ulion:
**Description**
when apply move_node operation, will trigger a fully re-render, because the parents' key is not restored. (There is a TODO comment in current code)

**Issue**
May Fixes: #3740 #3748

**Context**
This is a minimized fix, which does exactly same as those for insert_node/remove_node/merge_node/split_node, except we found the move_node common parents of source and target.

**Checks**


ianstormtaylor:
Thank you @ulion!

Issue #4124, opened 2021-03-16
data-gramm not actually disabling Grammarly


koralarts:
**Description**
The Grammarly extension is still enabled even though the `data-gramm` attribute is set to `false`.

**Steps**
To reproduce the behavior:
1. Install the Grammarly extension
2. Go to: [https://www.slatejs.org/examples/richtext](https://www.slatejs.org/examples/richtext)
3. Notice the Grammarly extension enabled on the Editor

**Expectation**
The Grammarly extension should not be enabled in the Editor when the `data-gramm` attribute is set to `false`

**Environment**
- Slate Version: v0.60+
- Operating System: Mac / Windows
- Browser: Chrome

**Context**
The issue seems to stem from the `data-slate-editor` attribute that's added. Having said attribute makes Grammarly ignore the `data-gramm` attribute. 

This is a Codepen that simulates the issue outside of Slate but with the relevant attributes ([https://codepen.io/koralarts/pen/MWbRRwM](https://codepen.io/koralarts/pen/MWbRRwM)).

ianstormtaylor:
Does the Grammarly extension being enabled still result in broken editing experience?

It sounds like they are purposely ignoring their own attribute.

koralarts:
The editing itself isn't broken but more of having the `data-gramm` attribute doesn't work in combination with `data-slate-editor`. So if you want to disable Grammarly which `data-gramm="false"` should do based on https://github.com/ianstormtaylor/slate/issues/733, it will still be enabled.

It does seem like Grammarly only looks for the existence `data-gramm` to determine whether or not to disable the extension but having `data-slate-editor` seem to negate that.

ianstormtaylor:
I think the question to answer is whether editing is broken in any way (not just a quick pass). It sounds like Grammarly is purposely circumventing the attribute. 

If editing is totally fine these days, we can remove the attribute by default. If not, we can escalate.

e1himself:
> I think the question to answer is whether editing is broken in any way (not just a quick pass). It sounds like Grammarly is purposely circumventing the attribute.
> If editing is totally fine these days, we can remove the attribute by default. If not, we can escalate.

Editing does work fine indeed. Thanks to how Grammarly team reworked the underlining logic: https://www.grammarly.com/blog/engineering/making-grammarly-feel-native-on-every-website/

But applying corrections from Grammarly doesn't seem to work at the moment: #4579.

e1himself:
Related issue: #733

PR #4126, opened 2021-03-16
Rename data-slate-editor to not interfere with data-gramm

koralarts:
Rename data-slate-editor to data-slate-react-editor because it inteferes with the data-gramm attribute which should disable the Grammarly extension

**Description**
When the Grammarly extension is installed, it does not get disabled even if `data-gramm` is set to `false`.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4124

**Example**

Default:
Screen Shot 2021-03-16 at 2 55 30 PM

When `data-gramm={undefined}`:
Screen Shot 2021-03-16 at 2 55 07 PM

**Context**
The issue stems from `data-slate-editor` which seems to interfere with Grammarly's logic to determine whether the extension is enabled or disabled for an editor.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -491,7 +491,7 @@ export const Editable = (props: EditableProps) => {
         autoCapitalize={
           !HAS_BEFORE_INPUT_SUPPORT ? undefined : attributes.autoCapitalize
         }
-        data-slate-editor
+        data-slate-react-editor
         data-slate-node="value"
         contentEditable={readOnly ? undefined : true}
         suppressContentEditableWarning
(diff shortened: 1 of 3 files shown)


clauderic:
@koralarts it sounds like the developers at Grammarly intentionally added `data-slate-editor` to a list of elements where the extension should not be applied, presumably because of issues at the time (which may or may not still be relevant today).

Based on that, I don't think renaming `data-slate-editor` to `data-slate-react-editor` is the right solution. Rather, you should try to get in touch with the development team at Grammarly to inquire about why they made that decision.

koralarts:
@clauderic Ok. I'll contact them. Thanks.

ianstormtaylor:
@koralarts do you have an update here?

PR #4132, opened 2021-03-20
Make onDomSelectionChange triggered after onClick.

ulion:
**Description**
onDomSelectionChange is called before the same click event triggered and processed, which may cause another round (if some element check useSelected and then render or by css/class name display differently) rerender which may change the dom and then cause the click event got unexpected event.target.

**Issue**
Fixes:

**Example**
No simple example, but if we do not delay onDomSelectionChange process, it will make the onClick event not reliable and got wrong event target and unexpected result.

**Context**
since the onDomSelectionChange does throttle with 100ms, and which is not time critical, but onClick is. So we first make sure onClick get correct event, then as throttled delay 100ms to check the dom selection change, which is harmless.

**Checks**
- [ ] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.) (Original code does not pass this, so leave it as it was)

### DIFF
--- a/.changeset/cold-beans-give.md
+++ b/.changeset/cold-beans-give.md
@@ -1,6 +1,6 @@
 ---
-"slate-react": patch
-"slate": patch
+'slate-react': patch
+'slate': patch
 ---
 
 Make onDomSelectionChange trigger after onClick.
--- a/.changeset/cold-beans-give.md
+++ b/.changeset/cold-beans-give.md
@@ -1,0 +1,6 @@
+---
+"slate-react": patch
+"slate": patch
+---
+
+Make onDomSelectionChange trigger after onClick.
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -511,6 +511,11 @@ export const Editable = (props: EditableProps) => {
     [readOnly]
   )
 
+  const scheduleOnDOMSelectionChange = useCallback(
+    () => setTimeout(onDOMSelectionChange),
+    [onDOMSelectionChange]
+  )
+
   // Attach a native DOM event handler for `selectionchange`, because React's
   // built-in `onSelect` handler doesn't fire for all selection changes. It's a
   // leaky polyfill that only fires on keypresses or clicks. Instead, we want to
@@ -518,15 +523,18 @@ export const Editable = (props: EditableProps) => {
   // https://github.com/facebook/react/issues/5785
   useIsomorphicLayoutEffect(() => {
     const window = ReactEditor.getWindow(editor)
-    window.document.addEventListener('selectionchange', onDOMSelectionChange)
+    window.document.addEventListener(
+      'selectionchange',
+      scheduleOnDOMSelectionChange
+    )
 
     return () => {
       window.document.removeEventListener(
         'selectionchange',
-        onDOMSelectionChange
+        scheduleOnDOMSelectionChange
       )
     }
-  }, [onDOMSelectionChange])
+  }, [scheduleOnDOMSelectionChange])
 
   const decorations = decorate([editor, []])
 


williamstein:
> which is harmless.

Personal minor concern -- I just tried changing the 100ms to something larger, e.g., 5000ms, then play around with slate.  Sometimes when I click on a point and start typing, my typing appears in the wrong place.  This makes me worried that the same thing would randomly happen after this PR is applied sometimes to very quick people.

ulion:
Surely, 5000ms is bad. since that will cause lost selection and dom sync. 100ms is the one which already be applied with the throttle side effect. The problem is, whether we can figure a way to trigger it just right after the onClick event, but not before the corresponding onClick.
If the leading:true cause it run in next tick rather than run at the moment, it would be ideal. But according source code of debounce, it does not.
So, I made an additional commit, which change to wrap a setTimeout inside throttle. so the first trigger will run right after one tick of the trigger, which does still fix my issue, and onClick is processed before the onDomSelectionChange, and reduced the trigger delay from 100ms to 1 tick, and it will keep keep the onClick handler caused change and onDomSelectionChange caused change in one render loop, since the render is scheduled right after these event handlers (include our setTimeout one) if there is any change.

dylans (changes requested):
@ulion it looks like this one needs lint/prettier run and then I'll review.

ulion:
more than that, I found my setTimeout break the throttle then cause some performance issue. will renew this PR to fix the issue.

dylans (changes requested):
Thanks @ulion, this PR is much clearer now. Could you add a quick changeset and then I'll merge?

ulion:
> Thanks @ulion, this PR is much clearer now. Could you add a quick changeset and then I'll merge?

no idea about the changeset. can you do that?

dylans:
> > Thanks @ulion, this PR is much clearer now. Could you add a quick changeset and then I'll merge?
> 
> no idea about the changeset. can you do that?

Sure, I will later today or tomorrow. To create a changeset, all you do is run yarn changeset and follow the prompts (which package is changing, is it a minor or patch change, what changed). Then you add, commit, and push it. This allows us to automate the releases, publish to npm @next after each PR, etc.

ulion:
> > > Thanks @ulion, this PR is much clearer now. Could you add a quick changeset and then I'll merge?
> > 
> > 
> > no idea about the changeset. can you do that?
> 
> Sure, I will later today or tomorrow. To create a changeset, all you do is run yarn changeset and follow the prompts (which package is changing, is it a minor or patch change, what changed). Then you add, commit, and push it. This allows us to automate the releases, publish to npm @next after each PR, etc.

:), I still don't have clear idea how to do this by my own, e.g. the 'slate' package is not involved in this PR but also marked as patch?

dylans (approved):

dylans:
> > > > Thanks @ulion, this PR is much clearer now. Could you add a quick changeset and then I'll merge?
> > > 
> > > 
> > > no idea about the changeset. can you do that?
> > 
> > 
> > Sure, I will later today or tomorrow. To create a changeset, all you do is run yarn changeset and follow the prompts (which package is changing, is it a minor or patch change, what changed). Then you add, commit, and push it. This allows us to automate the releases, publish to npm @next after each PR, etc.
> 
> :), I still don't have clear idea how to do this by my own, e.g. the 'slate' package is not involved in this PR but also marked as patch?

I probably made a mistake with this one. In this case it doesn’t matter too much as Slate 0.66 is going to be the next release as we have some minor level changes already committed.

dylans:
@ulion I received some feedback from https://github.com/zhugexinxin that this may have caused an infinite loop issue but not I'm not seeing their post. @zhugexinxin was there an issue with this PR?

ulion:
> @ulion I received some feedback from https://github.com/zhugexinxin that this may have caused an infinite loop issue but not I'm not seeing their post. @zhugexinxin was there an issue with this PR?

Old version of this PR may have issues like that since the setTimeout invalided the throttled. This new implement fixed that issue and clearly just delayed the call, no bad side-effect like the old version.

djibarian:
I'm experiencing a strange problem and I think is related to this change:

I have a modal dialog which saves the selection when opens, and restores it on closing. After restoring the selection it runs an `onClose()` function which modifies the editor contents (using `unwrapNodes()` in this case). The `editor.selection` is properly set before calling `unwrapNodes()` and the transform is properly done (and the selection updated accordingly). But afterwards, the selection goes back to the original value, which is now wrong.

I've experimented a few things, and finally noticed that I have to wait two ticks (a `setTimeout(0)` inside of a `setTimeout(0)`) before executing the unwrap, for the selection to not being restored afterwards. A tick is not enough.

Reading this thread and its debouncing/throttling soap opera, and specially @williamstein's comment about strange behaviours with long delays, it looks like this scenario of restoring the selection and performing a transform immediately afterwards has not been taken into account. I suspect it's an uncanceled debouncing problem.

@ulion It would be good if you can confirm what I'm saying, and of course I would love to know if there is any other (less hacky) way of fixing this problem... Or if I'm totally wrong and it's only my faulty code.

ulion:
> I'm experiencing a strange problem and I think is related to this change:
> 
> I have a modal dialog which saves the selection when opens, and restores it on closing. After restoring the selection it runs an `onClose()` function which modifies the editor contents (using `unwrapNodes()` in this case). The `editor.selection` is properly set before calling `unwrapNodes()` and the transform is properly done (and the selection updated accordingly). But afterwards, the selection goes back to the original value, which is now wrong.
> 
> I've experimented a few things, and finally noticed that I have to wait two ticks (a `setTimeout(0)` inside of a `setTimeout(0)`) before executing the unwrap, for the selection to not being restored afterwards. A tick is not enough.
> 
> Reading this thread and its debouncing/throttling soap opera, and specially @williamstein's comment about strange behaviours with long delays, it looks like this scenario of restoring the selection and performing a transform immediately afterwards has not been taken into account. I suspect it's an uncanceled debouncing problem.
> 
> @ulion It would be good if you can confirm what I'm saying, and of course I would love to know if there is any other (less hacky) way of fixing this problem... Or if I'm totally wrong and it's only my faulty code.

@djibarian How did you save and restore the selection? I would use a rangeRef to save it, and "restore" it by Transforms.select method, it should works unless you have other different setup than I think.

djibarian:
That modal is independent of the editor, so I can't use a rangeRef. What the modal does is the following:

On opening:

```
selectionRange.current = selection.getRangeAt(0);
selection.removeAllRanges();
```

(removing all ranges wouldn't be necessary, but if the dialog focuses in any input/textarea the selection will be cleared anyway).

On closing:

```
const selection = window.getSelection();
selection.removeAllRanges();
selection.addRange(selectionRange.current);
```

Prior to the timeouts I tried `Transform.select` as you suggested, but it didn't work.

```
const selection = window.getSelection();
if(!selection.rangeCount)
    Transforms.deselect(editor);
else
    Transforms.select(
        editor,
        ReactEditor.toSlateRange(
            editor,
            selection.getRangeAt(0),
            {}
        )
    );
```

ulion:
You have store the previous rangeRef before the modal visible, via Editor.rangeRef(editor, editor.selection) if there is editor.selection. then when modal close, newRange = rangeRef.unref(), newRange && Transforms.select(editor, newRange)
I would do it this way. what stopped you exactly not using rangeRef?

djibarian:
I don't get what would be the difference. In my tests, it doesn't matter if you set the selection explicitly via `Transforms.select` or `selection.addRange`. In both cases the `editor.selection` seems to properly update.

The problem is that in the next tick, the `editor.selection` seems to update again, back to the original value. So if you make any transform between the first selection set AND the next tick, the correction applied to the selection by the transform is lost.

ulion:
I will set break point and try to figure out why the selection is reseted, editor selection can update dom selection, dom selection can also update editor selection. in your case, "back to the original value", what caused it, I would trace it down, and eliminate the source.

djibarian:
As I said before, it looks like an uncanceled debounce/throttle in `onDOMSelectionChange` or `scheduleOnDOMSelectionChange`. Having a quick look at the code, both functions are memoized so if any of their dependencies change, the functions will be recreated and any previous in-progress calls won't cancel.

This is specially true for `onDOMSelectionChange`, which depends on `editor`. If while throttling the editor changes, the function will be recreated and even if you try to `cancel()` it, you will cancel the new instance, not the old one. And because the old one is tied to the old `editor` reference, makes sense to think that the selection will be reverted "back to the original value", prior to the editor changes.

ulion:
I don't know your case, normally editor won't change, always be the same editor.
As the first part of your guess, onDOMSelectionChange only reflect current dom selection to editor.selection, will this harm?
and why before 1 tick the dom selection changed (from where to where?), can this change be eliminated?

Issue #3447, opened 2020-01-20
Decorators do not work if renderLeaf function is memoized


ryanmitts:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

If the `renderLeaf` method is memoized, then decorators will not be rendered.

https://codesandbox.io/s/slate-decorator-usecallback-bug-fv18m

Slate: 0.57.0
Browser: Chrome / Safari / Firefox / Edge
OS: Mac / Windows / Linux / iOS / Android

#### What's the expected behavior?

The decorators are rendered.

#### Notes

This appears to be because Text elements do not take into account the decorator list to determine if they should re-render or not. The current search highlight example is making use of the fact that `renderLeaf` is changing every render.

mpkelly:
Odd one as my own code looks like:

```
  const renderLeaf = useCallback(
    (props: RenderLeafProps) => handleRenderLeaf(props, plugins),
    []
  );

  const decorate = useCallback(
    (entry: NodeEntry) => handleDecorate(entry, plugins, editor),
    []
  );
```
And decorate is working fine. I am not doing a search but I am doing on-the-fly code highlighting which works without any issue. 

The sandbox does work if you use `[search]` instead of `[]` as the dependency arg for the `useCallback` call for `renderLeaf`, though my understanding is that an empty array should be fine here, too. Using `[search]` will at least avoid the unnecessary re-renders that would happen without memoizing.

bartzy:
@mpkelly are you using 0.57.1, or a version from master?
I get the same behavior as @ryanmitts

mpkelly:
Hi @bartzy, I am using `0.57.1` My code is actually on Github now. See [here](https://github.com/mpkelly/react-editor-kit/blob/master/packages/editor-kit/src/editor/Editor.tsx) and [here](https://github.com/mpkelly/react-editor-kit/blob/2c4aae42af99894ebd4ac325596e11f9c195fcce/packages/editor-kit/src/features/code/CodeHighlighterPlugin.tsx#L66).

bartzy:
That's really weird. As soon as I add useCallback to renderLeaf,
decorations stop updating. Any idea what can cause that?

bartzy:
@mpkelly Is it possible that your leaf/element component(s) are just getting rerendered on every change for some other reason, and that's why decoration works?

mpkelly:
[...]

PR #4137, opened 2021-03-22
Fix text not rerender by decorations change

ulion:
**Description**
like #4080 tried to fixed, the decorations change does not trigger Text node re-render (because of memo compare not consider decorations).
This PR fix it just straightforward, do lodash isEqual compare between decorations items.

After fix this, I found it still does not work, due to the other bug from Leaf memo, which compare leaf without compare the text field, but, with slate-collaborative as example, we use decorate to render remote caret, when user click the other place of same text but different position, the memo of Leaf does not re-render the updated Leaf (with text updated), because Text.matches does not compare text field, while the text prop transferred to Leaf is the common 'text value' before splitted by the caret (decorate) which does not change.

**Issue**
Fixes: #3751 #3447

**Example**
To make the remote caret (decorations) work, slate-collaborative have to update renderLeaf method each time when any decoration changes, which caused many unnecessary re-render. With this PR, only the updated (effected) text and leaf will be re-rendered.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/leaf.tsx
+++ b/packages/slate-react/src/components/leaf.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { Element, Text } from 'slate'
+import isEqual from 'lodash/isEqual'
 import String from './string'
 import { PLACEHOLDER_SYMBOL } from '../utils/weak-maps'
 import { RenderLeafProps } from './editable'
@@ -70,7 +71,7 @@ const MemoizedLeaf = React.memo(Leaf, (prev, next) => {
     next.isLast === prev.isLast &&
     next.renderLeaf === prev.renderLeaf &&
     next.text === prev.text &&
-    Text.matches(next.leaf, prev.leaf)
+    isEqual(next.leaf, prev.leaf)
   )
 })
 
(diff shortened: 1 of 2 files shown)


williamstein:
Big +1 to this PR.   I made similar (but much less efficient) fixes to my copy of react-slate, when implementing realtime collaboration as well.  In particular, I also implemented other user cursors using decorations and hit this issue.

FezVrasta:
+1! We are using this fix with patch-package and it'd be great to have this merged.

ianstormtaylor:
I believe this is solved by merging https://github.com/ianstormtaylor/slate/pull/3437/files but if that is not the case please feel free to comment and I can re-open. Thanks!

HerbiePorter:
I think this may still be an issue @ianstormtaylor.  

In this example, I'm trying to decorate the selection on blur so that it still _appears_ selected but the decoration only applies to the leaf nodes on the next render (triggered by editing the editor content) even though the decorate function has run and been applied onBlur.  If this doesn't sound like the same issue I'm happy to open a new issue.

http://g.recordit.co/9zkYvOCN1G.gif

PR #4138, opened 2021-03-22
Decorate re-render optimize to not re-render all elements.

ulion:
**Description**
This PR will largely increase the performance of re-render caused by decorate update, in my case, avg 100+ms -> 30+ms.
You probably need first fix decorate related bug by #4137, then apply this optimize PR. Even though, this PR can be applied without that fix.

**Issue**
Fixes: 

**Context**
When decorate function updated, currently all Element get re-rendered, since memo compare failed, but the Element just for pass the decorate to Children, the only component which use decorate is Editable and Children. Well, then let's use context provider to supply the decorate to Children, and the performance is hugely improved by this.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -498,6 +498,7 @@ export const Editable = (props: EditableProps) => {
 
   return (
     <ReadOnlyContext.Provider value={readOnly}>
+      <DecorateContext.Provider value={decorate}>
       <Component
         // COMPAT: The Grammarly Chrome extension works by changing the DOM
         // out from under `contenteditable` elements, which leads to weird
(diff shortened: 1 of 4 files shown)


williamstein:
Good idea!  I was looking at this same code and wondering what the right react way to solve this problem is. Context!

williamstein:
To whatever slate dev review this -- I'm not a slate dev or committer, but I'm maintaining a heavily modified fork of the slate-react part of slate-react [here](https://github.com/sagemathinc/cocalc/tree/master/src/smc-webapp/frame-editors/markdown-editor/slate/slate-react).    I read through this PR and applied it to my code and it seems to work well and make sense.

bartzy:
I'm not sure I'm following if #4137 should be merge before this PR?

ulion:
> I'm not sure I'm following if #4137 should be merge before this PR?

#4137 is a bug fix, this PR is a performance enhancement, they are not relied each other. but they are all about decorate.

ianstormtaylor:
@ulion wow, thank you for this. Great call!

I've merged a few other pieces and this has conflicts now, if you're able to resolve them I'd love to merge this. Thank you!

ulion:
Rebased on master. moved the provider one level up since Children is no longer a component, and useContext won't work without a component wrapper.

ianstormtaylor:
Thanks @ulion!

PR #4148, opened 2021-03-29
Fix unhangRange condition and make sure result not out of initial range.

ulion:
**Description**
This is a minor fix for unhangRange, one for no matter whether start.offset is zero, should always do unhangRange if end.offset is zero and not collapsed, one for make sure the result range is still inside the original range.

**Issue**
Fixes: 

**Checks**

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -1507,7 +1507,7 @@ export const Editor: EditorInterface = {
     let [start, end] = Range.edges(range)
 
     // PERF: exit early if we can guarantee that the range isn't hanging.
-    if (start.offset !== 0 || end.offset !== 0 || Range.isCollapsed(range)) {
+    if (end.offset !== 0 || Range.isCollapsed(range)) {
       return range
     }
 
@@ -1535,6 +1535,9 @@ export const Editor: EditorInterface = {
         end = { path, offset: node.text.length }
         break
       }
+    }
+    if (Point.isBefore(end, start)) {
+      end = start
     }
 
     return { anchor: start, focus: end }


ulion:
This does fail a few tests, all of them are matching unhang with start.offset non-zero case, but they should all be unhanged. Since if you move the start anchor to offset 0, it does, but start offset zero make not any sense and does not matter with whether unhang should be done. Or, any one can explain why we should not do unhang when start.offset is not zero?

ulion:
Beyond above, I also an additional fix for unhangRange with inline void, since inline void should not be skipped for unhang, but I'm not sure, so only post it here for who is interested.

```
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -1523,10 +1523,15 @@ export const Editor: EditorInterface = {

     for (const [node, path] of Editor.nodes(editor, {
       at: before,
-      match: Text.isText,
+      match: n => Text.isText(n) || editor.isInline(n) && editor.isVoid(n),
       reverse: true,
       voids,
     })) {
+      if (!Text.isText(node)) {
+        // no unhang for inline void
+        end = Editor.start(editor, path)
+        break
+      }
       if (skip) {
         skip = false
         continue
```

ianstormtaylor:
@ulion can you add tests for this? Not exactly sure what the case is that we're missing right now. Thanks!

ulion:
> @ulion can you add tests for this? Not exactly sure what the case is that we're missing right now. Thanks!

The main issue of current PR is that which break 5 existed tests:
slate
   transforms
     delete
        selection
          1) block-across-nested
          2) block-depths-nested
          3) block-join-edges
          4) block-join-inline
          5) block-join-nested

So, is the test incorrect, or the PR? What does "unhang" for? Why anchor(the start point) not at offset 0 do not do unhang? (or why only anchor at offset 0 do unhang)?

As the inline void (the additional fix for), a simple test for delete could be (no matter focus at the beginning of 2nd block or the end of the first block):

```
export const input = (
  <editor>
    <block>
      <anchor />
      one
      <inlinevoid />
    </block>
    <block>
      <focus />
      three
    </block>
  </editor>
)
export const output = (
  <editor>
    <block>
      <cursor />
    </block>
    <block>three</block>
  </editor>
)
```

ianstormtaylor:
@ulion ah I see what you mean. 

Yeah, "unhang" is tricky and not standardized at all. If I remember correctly, it's trying to prevent though is when you've selected an entire block **and** the beginning of the next one. This happens most commonly when triple-clicking. In those cases, we want to convert to just selecting the entire block.

But those tests cases you mentioned where it's between the blocks are correct.

TheSpyder:
I was literally just about to post a PR with a similar change (although I hadn't considered the `end < start` condition, that's a good catch).

I have a bunch of tests in my branch, @ulion do you mind if we go with my PR instead? It has better justification for the change.

TheSpyder:
> As the inline void (the additional fix for), a simple test for delete could be (no matter focus at the beginning of 2nd block or the end of the first block):
> 
> ```
> export const input = (
>   <editor>
>     <block>
>       <anchor />
>       one
>       <inlinevoid />
>     </block>
>     <block>
>       <focus />
>       three
>     </block>
>   </editor>
> )
> export const output = (
>   <editor>
>     <block>
>       <cursor />
>     </block>
>     <block>three</block>
>   </editor>
> )
> ```

That test already works because normalization will add an empty text node after the void. But I think you're right, unhang should not skip over void nodes. I'm adding that to my branch.

ulion:
@ianstormtaylor Then I know why you check the start.offset, then it did not target any selection which end.offset is zero.
It is tricky. we need decide what exactly it is targeting. Maybe, just put unhangRange out as other methods which can be modified by the outside editor wrappers, so user can adjust the behaviour to do what they care about. Then just supply a simple default clearly defined version?
If we only want to handle the triple click case, then start.offset zero is not enough, it also includes some cases not caused by triple click. If generally we do not trust the end offset zero case, then this PR was trying to work with that case.
@TheSpyder I think we need first clearly define what unhangRange does and for. then we know how to implement it correctly.

ulion:
> > As the inline void (the additional fix for), a simple test for delete could be (no matter focus at the beginning of 2nd block or the end of the first block):
> > ```
> > export const input = (
> >   <editor>
> >     <block>
> >       <anchor />
> >       one
> >       <inlinevoid />
> >     </block>
> >     <block>
> >       <focus />
> >       three
> >     </block>
> >   </editor>
> > )
> > export const output = (
> >   <editor>
> >     <block>
> >       <cursor />
> >     </block>
> >     <block>three</block>
> >   </editor>
> > )
> > ```
> 
> That test already works because normalization will add an empty text node after the void. But I think you're right, unhang should not skip over void nodes. I'm adding that to my branch.

empty text will be skipped by current implement, inline void normally has empty text, not to mention the nodes looping only hits text and the voids parameter may cause it do not look into the inline void so even the text are not looped.

TheSpyder:
ah sorry I didn't see the email notifications before I opened my PR.

You're right, that test didn't already work I just tried it after I added your inline void change to my branch. I'll add that test as well.

TheSpyder:
> You're right, that test didn't already work I just tried it after I added your inline void change to my branch. I'll add that test as well.

I need to think before I post 😂 

Empty text nodes are not skipped when doing an unhang across blocks, because the condition is `node.text !== '' || Path.isBefore(path, blockPath)`. An empty text node at the end of a block passes the second check.

I'm still adding your test, because it's important to know it works.

dylans:
> @TheSpyder I think we need first clearly define what unhangRange does and for. then we know how to implement it correctly.

There was a recent discussion in Slack about the intent of unhangRange, adding here for context:

Jason:
Why does unhangRange only unhang the end of a range and not the start as well? Trying to tell if this is a bug or some logic I'm not following. If the start of the range is at the last index of a text shouldn't unhangRange advance that to the next node?

TheSpyder :
There's been a bit of discussion about unhang range in PRs recently :thinking_face: I have an old one that I need to fix up, and https://github.com/ianstormtaylor/slate/pull/4522 was logged last week

ianstormtaylor:
I don’t believe it’s a bug. It was designed to handle ranges that result from triple-clicks.

ianstormtaylor:
Not that there might not be improvements, or that there might not be some other better way to solve for triple-clicks, or that there might be a better name for it

Jason:
ianstormtaylor When you say triple-click handling you mean the fact that the browser selection by default gets placed from start of line to start of next line? In that cause yes the end of the selection is where the problem is.
If looked at as a more generic implementation it seems like unhanging the start in cases where you have {text: 'foo|'}, {text: 'bar|', bold: true} and it makes sense to normalize the selection to {text: 'foo'}, {text: '|bar|', bold: true} which helps in some cases by avoiding splits that immediately get merged back, etc.

ianstormtaylor:
Yup, that’s exactly right. It was created for the triple-click case. I agree that something to handle the start-edged cases would be nice too. But folks in GitHub discussions have been kind of assuming that the code meant to be written for both, and that it’s a bug to fix, which isn’t true. I’m not sure if there are gotchas to expanding the scope to handle both edges

dylans:
I think this has been addressed by other recent PRs. If not, please re-open.

PR #4149, opened 2021-03-30
[docs] Prune import that is never used

mariusbutuc:
**Description**

In the Serializing docs, for the `serialize` function for HTML example,
we import `Node` but never use it in the scope of the example. 

Remove it to eliminate the possibility for confusion.

**Checks**

### DIFF
--- a/docs/concepts/09-serializing.md
+++ b/docs/concepts/09-serializing.md
@@ -53,7 +53,7 @@ For example, here's a similar `serialize` function for HTML:
 
 ```js
 import escapeHtml from 'escape-html'
-import { Node, Text } from 'slate'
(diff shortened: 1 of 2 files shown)


mariusbutuc:
Failed for a file outside of the scope of this PR:

```sh
Checking formatting...
docs/concepts/11-typescript.md
Code style issues found in the above file(s). Forgot to run Prettier?
```

For the sake of CI/tests passing on this PR, will add that fix (remove the duplicated white line) as a separate commit in the meantime. If that's unnecessary, having it as a separate commit will make it easier to be removed. 👍🏼

thesunny:
Thanks @mariusbutuc

Issue #4055, opened 2021-01-23
Spell check correction does not trigger a change event


edfxphn:
#### Do you want to request a _feature_ or report a _bug_?

I would say this is a bug (but may be an incorrect usage).

#### What's the current behavior?

I am new to Slate. I implemented it with an autosave feature. I searched in the existing issues, but do not find the same issue.

When fixing an error through  the spell check of Firefox, the word is fixed, but the "onChange" event is not triggered. If I type a new key, the onChange is triggered, and the error is backed again.

I tried with autoCorrect, autoCorrect="true", ... but nothings working.

This behavior is reproductible in the samples.

![Example](http://g.recordit.co/4rodUkRtDM.gif)

Slate: 0.59.0
Browser: Firefox 84.0.2
OS: macOS 11.1

#### What's the expected behavior?

Errors are fixed in the underlying model, and change event is triggered.

BrentFarese:
How does Firefox spellcheck work?  Does it just update the DOM?  If Slate's model is not updated, `onChange` will not fire.

edfxphn:
I don't know. It seems only a "input" event is fired which is ignored by Slate.
PS: it's not specific to Firefox ; it is the same with all navigators.

stefanoverna:
having the same issue here. happy to contribute with a PR if someone can point me in the right direction!

mattwalters:
I'm having the same problem. For some reason this seems to have only started for us in the last few days. Does anyone have any insight here. Echoing @stefanoverna here... happy to contribute a PR if someone can point me in the right direction.

mdmag:
I have the same issue.

mattwalters:
@mdmag FWIW, Updating to the latest version solved my problem.

mdmag:
@mattwalters, I will update from version 0.57. 
So, I assume that there are a lot of breaking changes?

mcturner1995:
Having the same issue. Has anyone found a solution?

mattwalters:
@mcturner1995 What version are you using?

mcturner1995:
Here are my versions from my package.json:
```
    "slate": "^0.59.0",
    "slate-history": "^0.59.0",
    "slate-hyperscript": "^0.59.0",
    "slate-react": "^0.59.0",
```

dylans:
[...]

PR #4150, opened 2021-03-30
replace IS_FIREFOX with IS_FIREFOX_LEGACY in HAS_BEFORE_INPUT_SUPPORT

nivekithan:
**Description**

Firefox did not support `onBeforeInput` event so code used  `React` leaky `SyntheticEvent` as a fallback. Which leads certain bugs like

- #4055 

But as of `Firefox/87` `onBeforeInput` is supported [mdn](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforeinput_event#browser_compatibility)

**Issue**

- #3185 
- #4055 

**Context**
So instead of checking if the browser is firefox we will check if the browser is  firefox and version is less than 87. If thats true only then we will fallback to `React SyntheticEvent`

**Checks**


ianstormtaylor:
Thanks @nivekithan!

TheSpyder:
#4118 is probably a better fix, rather than version detection?

Also fixes #3185

Issue #4142, opened 2021-03-26
remove normalizeDOMPoint in DOM selection sync


githoniel:
**Problem**

We have been plagued by the problem of IME duplicate input/crash for a long time such as #3914, #4060.  after investigation I found they are all caused by selection sync after investigation.

We will sync selection in `React.useLayoutEffect` and  we will exit when DOM selection is equal to `editor.selection`.

```ts
   if (
      hasDomSelection &&
      hasDomSelectionInEditor &&
      selection &&
      Range.equals(ReactEditor.toSlateRange(editor, domSelection), selection)
    ) {
      return
    }
```

`ReactEditor.toSlateRange` is the problem. It will call `normalizeDOMPoint(domPoint)` to normalize a DOM point so that it always refers to a text node as comment says.

think of these

```
<span>
  <span contentEditable={false}></span>
  <cursor>
  <span data-slate-node="text">
    <span data-slate-leaf="true" >
      <span data-slate-string="true">xx</span>
    </span>
</span>
```

it is no need to sync selection without IME but if typing `测试` using IME it will result in 

```
<span>
  <span contentEditable={false}></span>
  测试
  <span data-slate-node="text">
    <span data-slate-leaf="true" >
      <span data-slate-string="true">测试<cursor></span>
    </span>
</span>
```

**Solution**

remove `normalizeDOMPoint` and always sync DOM selection in to slate text node.  allow `toSlatePoint` to return null when dom point is not in slate text node.

**Alternatives**

N/A

**Context**

N/A

Issue #4046, opened 2021-01-18
Bug: Undo/Redo shortcuts do not work with Cyrillic keyboard layout


OlegDydy:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Keyboard shortcuts `Ctrl+Z`, `Ctrl+Y` and `Shift+Ctrl+Z` don't work with Russian keyboard layout.

Can be reproduced in example editor: https://www.slatejs.org/examples/richtext
![Screencast](https://user-images.githubusercontent.com/30150630/104891855-82340580-5982-11eb-8d40-b1fa6f0e406c.gif)
In the GIF, I first type in “Hello world” and try to undo the text, then I type in another letter to show that the keyboard layout hasn't changed. Then I add a new line with the text "qwerty" and show that everything is fine with the en-US layout. At the end I undo the "qwerty" and try to repeat with the ru-RU keyboard layout.

Slate: 0.59.0
Browser: Chome/Chromium 87 and Firefox 84
OS: Ubuntu 20.04 and Window 10 2004

#### What's the expected behavior?

The keyboard shortcuts `Ctrl+Z`, `Ctrl+Y` and `Shift+Ctrl+Z` must work regardless of the keyboard layout.

kiyasov:
When will it be fixed?

kiyasov:
@techirdliu

kiyasov:
[techird](https://github.com/techird)

Issue #3914, opened 2020-10-09
Chinese text being doubled (duplicated inserted into the wrong node)


gt2345:
#### Do you want to request a _feature_ or report a _bug_?
_bug_

#### What's the current behavior?

Slate: 0.50.0
Browser: Chrome
OS: Mac

![Screen Shot 2020-10-09 at 4 34 28 PM](https://user-images.githubusercontent.com/40620519/95561783-d5c2b900-0a4d-11eb-8d2b-e1412256e804.png)

When typing in Chinese for the second time (if the editor has just rendered and not been touched, this won't happen. It only happens when I type in something and clear it), text somehow got inserted into both < p data-slate-node="element"> and < span data-slate-string="true"> nodes, which results in text being doubled on screen. At this time, if I press enter to switch to the next line, then everything comes back to normal.
P.S. I already change some code in toSlatePoint function to normally render at this situation, otherwise is will through an error.
Could anyone provide some work around to eliminate text being inserted into < p data-slate-node="element">, or auto trigger pressing enter?

#### What's the expected behavior?
Text should only appear in < span data-slate-string="true">, but not in < p data-slate-node="element">

ls19930101:
不要使用placeholder,单独自己维护一个placeholder可以规避slate对中文输入法的不支持.

emmkong:
@ls19930101 你有具体的实现方式吗？你怎样替代原有的placeholder？

Q-Qplus1s:
same problem
![](https://content.markdowner.net/pub/dGN5q0-EM5eB83)

repeat render when input chinese directly after the decorated text

Q-Qplus1s:
@emmkong 即使不用 placeholder 问题同样存在

Xheldon:
你可能需要在 Editable 中的 compositionStart 中和 End 中做一些处理

Issue #4060, opened 2021-01-28
editor will crash when input Chinese in the end of link element


sennpang:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### Chrome behavior
input **Chinese** at end of **link** element, character will input twice, then move focus to left, the editor crush
![Kapture 2021-01-28 at 12 05 09_Chrome](https://user-images.githubusercontent.com/16861647/106089562-25df9b80-6163-11eb-953b-b99171b6f3e4.gif)

#### Safari behavior
input **Chinese** at end of **link** element, then move focus to left, the editor crush
![Kapture 2021-01-28 at 12 11 09_Safari](https://user-images.githubusercontent.com/16861647/106089564-28da8c00-6163-11eb-9fe0-7ee6ce8e749b.gif)

#### Firefox behavior
input **Chinese** at end of **link** element, then **move focus to link element**, then came back, input Chinese, the editor crush
![Kapture 2021-01-28 at 12 15 07_Firefox](https://user-images.githubusercontent.com/16861647/106089566-29732280-6163-11eb-8105-1b51f2399155.gif)

Slate: 0.XX.X (https://www.slatejs.org/examples/links)
Browser: Chrome / Safari / Firefox
OS: Mac

#### What's the expected behavior?
input Chinese with no errors

ulion:
This only happens in chrome, imo. For now, I only figured out some ugly workaround for this, no idea whether has other side effect. The idea is to document.execCommand("undo") to remove the first unexpected input by ime, in custom onCompositionEnd.

1c7:
Hi, Any update on this?

OldDream:
> This only happens in chrome, imo. For now, I only figured out some ugly workaround for this, no idea whether introduce other side effect. The idea is to document.execCommand("undo") to remove the first unexpected input by ime, in custom onCompositionEnd.


Thanks !
document.execCommand("undo")   helps me out  in 0.81.1. 
When I use sogou input before a inline element, the whole editor crashed.

PR #4157, opened 2021-04-01
feat: sync selection to extactly matched DOM selection

githoniel:
**Description**

We will return if DOM selection is equal to `editor.selection` when we  sync DOM selection state as below

```ts
   if (
      hasDomSelection &&
      hasDomSelectionInEditor &&
      selection &&
      Range.equals(ReactEditor.toSlateRange(editor, domSelection), selection)
    ) {
      return
    }
```

`ReactEditor.toSlateRange`  will call `normalizeDOMPoint(domPoint)` to normalize a DOM point so that it always refers to a text node nearby as comment says.

think of DOM structure below

```
<span>
  <span contentEditable={false}></span>
  <cursor>
  <span data-slate-node="text">
    <span data-slate-leaf="true" >
      <span data-slate-string="true">xx</span>
    </span>
</span>
```

there is no need to sync DOM selection without this PR。but if typing `测试` using IME it will result in 

```
<span>
  <span contentEditable={false}></span>
  测试
  <span data-slate-node="text">
    <span data-slate-leaf="true" >
      <span data-slate-string="true">测试<cursor>xx</span>
    </span>
</span>
```

**Issue**

Fixes: https://github.com/ianstormtaylor/slate/issues/4142

**Example**

```ts
toSlateRange<T extends boolean>(editor: ReactEditor, domRange: DOMRange | DOMStaticRange | DOMSelection, option: {
    extractMatch: T;
}): T extends true ? Range | null : Range
``` 

add `extractMatch` to `ReactEditor.toSlateRange`，it will return `Range | null` when using `extractMatch`

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -177,7 +177,7 @@ export const Editable = (props: EditableProps) => {
     // If the DOM selection is in the editor and the editor selection is already correct, we're done.
     if (hasDomSelection && hasDomSelectionInEditor && selection) {
       const slateRange = ReactEditor.toSlateRange(editor, domSelection, {
-        extractMatch: true,
+        exactMatch: true,
       })
       if (slateRange && Range.equals(slateRange, selection)) {
         return
@@ -190,7 +190,7 @@ export const Editable = (props: EditableProps) => {
     // and thus it doesn't transform selection on its own
     if (selection && !ReactEditor.hasRange(editor, selection)) {
       editor.selection = ReactEditor.toSlateRange(editor, domSelection, {
-        extractMatch: false,
+        exactMatch: false,
       })
       return
     }
@@ -283,7 +283,7 @@ export const Editable = (props: EditableProps) => {
 
           if (targetRange) {
             const range = ReactEditor.toSlateRange(editor, targetRange, {
-              extractMatch: false,
+              exactMatch: false,
             })
 
             if (!selection || !Range.equals(selection, range)) {
@@ -449,7 +449,7 @@ export const Editable = (props: EditableProps) => {
 
         if (anchorNodeSelectable && focusNodeSelectable) {
           const range = ReactEditor.toSlateRange(editor, domSelection, {
-            extractMatch: false,
+            exactMatch: false,
           })
           Transforms.select(editor, range)
         } else {
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -175,13 +175,13 @@ export const Editable = (props: EditableProps) => {
     }
 
     // If the DOM selection is in the editor and the editor selection is already correct, we're done.
-    if (
-      hasDomSelection &&
-      hasDomSelectionInEditor &&
-      selection &&
-      Range.equals(ReactEditor.toSlateRange(editor, domSelection), selection)
-    ) {
-      return
+    if (hasDomSelection && hasDomSelectionInEditor && selection) {
+      const slateRange = ReactEditor.toSlateRange(editor, domSelection, {
+        extractMatch: true,
+      })
+      if (slateRange && Range.equals(slateRange, selection)) {
+        return
+      }
     }
 
     // when <Editable/> is being controlled through external value
@@ -189,7 +189,9 @@ export const Editable = (props: EditableProps) => {
     // but Slate's value is not being updated through any operation
     // and thus it doesn't transform selection on its own
     if (selection && !ReactEditor.hasRange(editor, selection)) {
-      editor.selection = ReactEditor.toSlateRange(editor, domSelection)
+      editor.selection = ReactEditor.toSlateRange(editor, domSelection, {
+        extractMatch: false,
+      })
       return
     }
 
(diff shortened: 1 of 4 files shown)


williamstein:
This looks vastly better than watching the composition event and disabling selection sync temporarily, which I've been doing in my code (and which -- as you point out -- doesn't work well in general).

Important note -- in English there is no word "extractly".  I think you mean "exactly".    Basically, make the change "extract" --> "exact" everywhere in your code, or people might be confused reading this.

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -513,27 +521,29 @@ export const ReactEditor = {
     }
 
     if (!textNode) {
```
williamstein, line 523: I don't think this is correct.  You can't just delete this exception in case extractMatch [sic] is false.  I think the code should be
```
    if (!textNode) {
      if (exactMatch) {
        return null as T extends true ? Point | null : Point
      }
      throw new Error(
        `Cannot resolve a Slate point from DOM point: ${domPoint}`
      );
    }
```
Of course you would only notice this when some other bug messes up the input to this function...

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -513,27 +521,29 @@ export const ReactEditor = {
...
    * Find a Slate range from a DOM range or selection.
    */
 
-  toSlateRange(
+  toSlateRange<T extends boolean>(
     editor: ReactEditor,
-    domRange: DOMRange | DOMStaticRange | DOMSelection
-  ): Range {
+    domRange: DOMRange | DOMStaticRange | DOMSelection,
+    option: {
```
williamstein, line 542: This is inconsistent  -- using an option object versus a parameter for toSlatePoint above.   react-editor.ts has some existing use of this sort of pattern already and it uses an object called `options`.  Thus I would change both here and in  toSlatePoint to use an object called "`options`".

williamstein: And of course it is important for backward compatibility and consistency to not force options to be provided, like earlier in this file we find:
```
    options: { editable?: boolean } = {}
```

packages/slate-react/src/plugin/react-editor.ts
githoniel: you're right,  I forgot to restore the code while all errors that were thrown manually in `slate-react` and replaced them with deselection in my own forked slate.

githoniel:
thank you for preview and sorry for my poor english


> This looks vastly better than watching the composition event and disabling selection sync temporarily, which I've been doing in my code (and which -- as you point out -- doesn't work well in general).

watching the composition event and disabling selection sync temporarily is not enough, the wrong location of the selection is generally caused by mouse clicks, not by input or moving cursor using arrow keys

as below structure, there may be many wrong cursor if you do not set `contentEditable={false}` at root span when you click on it

```jsx
<span>
   <cursor>                                        <----- cursor maybe here
  <span contentEditable={false}></span>
  <cursor>                                         <----- cursor maybe here
  <span data-slate-node="text">
     <cursor>                                      <----- cursor maybe here
    <span data-slate-leaf="true" >    <----- cursor maybe here
      <span data-slate-string="true">xx</span>
    </span>
</span>
```

> Important note -- in English there is no word "extractly". I think you mean "exactly". Basically, make the change "extract" --> "exact" everywhere in your code, or people might be confused reading this.

techird:
This PR does not fix #4046 nor #3914 in my local tests:

https://user-images.githubusercontent.com/1901286/113691738-59104f00-96ff-11eb-9931-7f19b6135ed7.mp4

githoniel:
@techird you  may need #4158  to fix #3914

techird:
@githoniel As this pr aims to fix #4142, which mention issue  #4046 nor #3914 as the problem, I misunderstood this is the solution.

> #### Problem
> We have been plagued by the problem of IME duplicate input/crash for a long time such as #3914, #4060. after investigation I found they are all caused by selection sync after investigation.

So I am a little confuse here, which particular issue is this pr aim to fix? Try to understand the changes, but still got nothing.

githoniel:
> @githoniel As this pr aims to fix #4142, which mention issue #4046 nor #3914 as the problem, I misunderstood this is the solution.
> 
> > #### Problem
> > We have been plagued by the problem of IME duplicate input/crash for a long time such as #3914, #4060. after investigation I found they are all caused by selection sync after investigation.
> 
> So I am a little confuse here, which particular issue is this pr aim to fix? Try to understand the changes, but still got nothing.

they will both result in IME double input but it's not the same. there are two reason: 

1. when you click to locate to cursor, DOM selection may not in the right place that extactly matched editor.selection. this PR will fix this
2. when you input Text with mark,  `Slate`  will insert new node but React will not refresh to remove the input in previous DOM point, #4158 will fix this

ianstormtaylor:
Thanks @githoniel!

beorn:
How about just catching the error from `toSlateRange` the one place you need to, instead of adding the extra parameter to `toSlateRange` and complicating the logic for all the other calls of the function?

packages/slate-react/src/components/editable.tsx
```diff
@@ -444,7 +448,9 @@ export const Editable = (props: EditableProps) => {
           isTargetInsideVoid(editor, focusNode)
 
         if (anchorNodeSelectable && focusNodeSelectable) {
-          const range = ReactEditor.toSlateRange(editor, domSelection)
+          const range = ReactEditor.toSlateRange(editor, domSelection, {
+            exactMatch: false,
```
clauderic, line 452: @githoniel do you remember why this is set to `exactMatch: false`? This cause this bug in Firefox, which is fixed when setting to `exactMatch: true`: https://github.com/ianstormtaylor/slate/issues/4373

githoniel: just because it is `exactMatch: false` before I add `exactMatch` options.  sorry I dont know too...

PR #4158, opened 2021-04-01
fix ime double input with mark

githoniel:
**Description**

this pr will force `String` element refresh it self to avoid ime double input

**Issue**

No

**Example**

xx

**Context**

image

while we should get 

image

this is becasue we memoizedLeaf to avoid refresh. now we will refresh all sibling leaf when a leaf changes

**Checks**

### DIFF
--- a/packages/slate-react/src/components/leaf.tsx
+++ b/packages/slate-react/src/components/leaf.tsx
@@ -43,7 +43,13 @@ const Leaf = (props: {
   }, [placeholderRef])
 
   let children = (
-    <String isLast={isLast} leaf={leaf} parent={parent} text={text} />
+    <String
+      key={Math.random()}
+      isLast={isLast}
+      leaf={leaf}
+      parent={parent}
+      text={text}
+    />
   )
 
   if (leaf[PLACEHOLDER_SYMBOL]) {
--- a/packages/slate-react/src/components/leaf.tsx
+++ b/packages/slate-react/src/components/leaf.tsx
@@ -4,6 +4,9 @@ import String from './string'
 import { PLACEHOLDER_SYMBOL } from '../utils/weak-maps'
 import { RenderLeafProps } from './editable'
 
+// auto-incrementing key for String component, force it refresh to
+// prevent inconsistent rendering by React with IME input
+let keyForString = 0
 /**
  * Individual leaves in a text node with unique formatting.
  */
@@ -44,7 +47,7 @@ const Leaf = (props: {
 
   let children = (
     <String
-      key={Math.random()}
+      key={keyForString++}
       isLast={isLast}
       leaf={leaf}
       parent={parent}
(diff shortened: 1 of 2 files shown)


packages/slate-react/src/components/leaf.tsx
```diff
@@ -43,7 +43,13 @@ const Leaf = (props: {
   }, [placeholderRef])
 
   let children = (
-    <String isLast={isLast} leaf={leaf} parent={parent} text={text} />
+    <String
+      key={Math.random()}
```
ianstormtaylor, line 47: Can you use a number key like `let key = 0` at the beginning of the file, then `key++` here instead? Makes it more standard and easier to augment in the future if we need.

githoniel: OK, I have add comments for it too

packages/slate-react/src/components/leaf.tsx
```diff
@@ -43,7 +46,13 @@ const Leaf = (props: {
   }, [placeholderRef])
 
   let children = (
-    <String isLast={isLast} leaf={leaf} parent={parent} text={text} />
+    <String
+      key={keyForString++}
```
techird, line 50: Increase key on every leaf render may potentionally burden the performance?

githoniel: yes, but I don't find any other better ways

ianstormtaylor:
Thanks @githoniel!

rockettomatooo:
@githoniel I'm running into issues in my editor because of this change. In my case, I'm listening to `click` events in the editor. Now a click event only fires, when `mousedown` and `mouseup` fire on the same node or have a common ancestor. Since the `<String />` gets force-remounted, react seems to remove the DOM-Node the `mousedown` occured on from the DOM so when the `mouseup` occurs, it doesn't have a common ancestor thus the `click` never occurs. 

Now I'm not really familiar with IME so is there a way to conditionalize the key increment?

githoniel:
@rockettomatooo you may try to compare slate nodes to see if they are the same instead of DOM nodes. 

IME input is a hard problem to solve, strictly speaking it's a React Reconciliation's bug. I have not found a better solution for this issue

rockettomatooo:
@githoniel The problem I have is, that the browser does this comparison. What I described is the browsers logic to fire a `click` event. I can try to detect a click on my own but I want to avoid that if possible somehow. I prepared a [small jsfiddle here](https://jsfiddle.net/tu0py2gL/).

Can you maybe explain to me, how the reconciliation bug occurs specifically? 
My guess would be, that the IME directly modifies the HTML and slate lets it do it because it's a composition and it's not done yet. Then, when the composition is done, slate inserts the composition as bold, but whats already in the HTML doesn't get updated because of the bug. This should not be a re-render problem am I correct?

githoniel:
@rockettomatooo Chrome is able to remove composition text correctly when the final input text is on the side of the composition text.

we will create a new node and insert text in it so the the final input text is **not** on the side of the composition text. 

maybe we could  move selection and insert new node  in advance when `onCompositionStart` trigger

PR #4159, opened 2021-04-01
Update 09-serializing.md

dev-cyprium:
This is a typo in the docs. You can't have `children` be a `const` because inside the `if` block you change it. It needs to be defined as let, or else it doesn't work.

**Description**
It's just a simple typo, the documentation is not correct.

**Checks**


ianstormtaylor:
Thanks @dev-cyprium!

PR #4161, opened 2021-04-01
Create .gitattributes and explicitly set the line breaks to be LF

nivekithan:
**Description**
In windows git will convert the LF line breaks to CRLF by default and due to eslint configuration the eslint will throw error. If you are using eslint plugin in vscode this leads to whole screen becoming red.  Which should be avoided.

Thats why we will set the EOL to be LF in `.gitattributes ` so that git wont convert the LF to CRLF.

For reference 

- [eslint]( https://eslint.org/docs/rules/linebreak-style)
- [stackoverflow](https://stackoverflow.com/a/42135910)

**Checks**

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


ianstormtaylor:
Thanks @nivekithan!

Issue #4130, opened 2021-03-20
Caret: <enter> misbehaving on leaf edges


marco-bor:
**Description**
Caret holds previous location when pressing \<enter> on a leaf edge. Ideal behavior would be to go to the start of the new block.

**Recording**
![slate](https://user-images.githubusercontent.com/5829628/111869168-4d423000-897e-11eb-955d-6d8ee6218fe3.gif)

**Sandbox**
[Rich text example](https://www.slatejs.org/examples/richtext)

**Steps**
To reproduce the behavior:
1. Go to [Rich text example](https://www.slatejs.org/examples/richtext)
2. Place caret before/after any leaf ("rich", "much", "<textarea>")
3. Press \<enter>
4. See caret location

**Expectation**
See Draft.js equivalent
![draft](https://user-images.githubusercontent.com/5829628/111869210-867aa000-897e-11eb-836d-4eda8199a7c2.gif)

marco-bor:
Currently using this workaround

``` typescript
import { Editor, Point, Transforms } from 'slate'

export function withFix(editor: Editor) {
    var a = editor
    var { insertBreak } = editor

    a.insertBreak = function () {
        const { selection: beforeSelection } = a

        insertBreak()

        const afterLocation = Editor.after(a, a.selection!!, { unit: "block" })!!
        if (Point.equals(beforeSelection!!.anchor, a.selection!!.anchor)) {
            Transforms.select(a, afterLocation)
        }
    }

    return a
}
```

toboid:
Just retested and this appears to be resolved, can it be closed @thesunny @marco-bor?

PR #4163, opened 2021-04-02
Fix cursor movements when splitting nodes

nivekithan:
**Description**
In `split_node` operation sometimes there will be creation empty text node which then will cleared by default normalization function. 
This leads to bugs regarding cursor positions like 
- #4130 

This pr aims to solve problems like those by not creating those empty text nodes itself

**Issue**
- #4130 

## Before

From [example](https://www.slatejs.org/examples/richtext)

gif

## After

gif

**Context**

[old code](https://github.com/ianstormtaylor/slate/blob/0b0328b2d0a595d48905b41c6afcc4f65bc352c6/packages/slate/src/transforms/general.ts#L261)  will create two new text node even if the text is empty

[new code](https://github.com/nivekithan/slate/blob/acc4051286474a647a02cb58658e3d340da9555e/packages/slate/src/transforms/general.ts#L269) will not add new text node if its empty and will remove the old text node if it became empty

Since we are not always creating text nodes sometime Element node might contain empty `children` array in that case we will create new empty text node and populate with previous text node properties 

I submitting this is as draft since still there are some errors like 

bug

where the `marks` are not transferred properly

And the also some tests are failing I will also take care of that,

Before doing those I just want to know opinions of maintainers since I am changing the behavior of an operation weather it is okay to do that or not

**Checks**

### DIFF
--- a/packages/slate/src/transforms/general.ts
+++ b/packages/slate/src/transforms/general.ts
@@ -265,6 +265,12 @@ export const GeneralTransforms: GeneralTransforms = {
             ...(properties as Partial<Text>),
             text: after,
           }
+
+          if (newNode.text) {
+            parent.children.splice(index + 1, node.text ? 0 : 1, newNode)
+          } else {
+            parent.children.splice(index + 1, node.text ? 0 : 1)
+          }
         } else {
           const before = node.children.slice(0, position)
           const after = node.children.slice(position)
@@ -274,9 +280,8 @@ export const GeneralTransforms: GeneralTransforms = {
             ...(properties as Partial<Element>),
             children: after,
           }
+          parent.children.splice(index + 1, 0, newNode)
         }
-
-        parent.children.splice(index + 1, 0, newNode)
 
         if (selection) {
           for (const [point, key] of Range.points(selection)) {
--- a/packages/slate/src/transforms/general.ts
+++ b/packages/slate/src/transforms/general.ts
@@ -266,15 +266,16 @@ export const GeneralTransforms: GeneralTransforms = {
             text: after,
           }
 
-          if (newNode.text) {
-            parent.children.splice(index + 1, node.text ? 0 : 1, newNode)
-          } else {
-            parent.children.splice(index + 1, node.text ? 0 : 1)
-          }
+          newNode.text && parent.children.splice(index + 1, 0, newNode)
+          !node.text && index && parent.children.splice(index, 1)
         } else {
           const before = node.children.slice(0, position)
           const after = node.children.slice(position)
           node.children = before
+
+          if (after.length === 0) {
+            after.push({ ...before[before.length - 1], text: '' })
+          }
 
           newNode = {
             ...(properties as Partial<Element>),


TheSpyder:
I like the idea, those empty split nodes cause problems and generate unnecessary operation traffic. However please don't mess with the central transform method; if a split node operation is applied it must split as requested or real-time collaboration will break.

There may be subsequent operations applied after the split node that mean the empty text is not normalised away, for example the bug you found with marks not transferring. I can also see pressing enter at the start of a paragraph needing this (split is the best way to do it to retain paragraph and mark attributes).

Instead, I recommend adjusting [`Transforms.splitNodes`](https://github.com/ianstormtaylor/slate/blob/69f8cb146640d554a171010a29581922d55e892c/packages/slate/src/transforms/node.ts#L655) (or the code that calls it, triggering the bugs you found) to not create the operation in the first place.

nivekithan:
Pressing at the start of paragraph wont delete the text node even though it is empty and I am also confident there is some way to make sure that marks attributes get transferred to new node.

But as you said changing the core of transform might lead to some problems for example`Operation.inverse` might not work properly with further changing it.

So I will try to solve the bug in `Transform.splitNodes`, But I do have to share my skepticism over this method since if someone where to call `Editor.apply(editor, {\\ split_node operation} )` manually the bug will stay ( I dont think it is common though)

nivekithan:
I am unable to come up with a solution that solves this bug. 

I will just share ways I tried to solve this, someone might find it useful or not

- Since it is not okay to modify the code in main transform, I tried to modify the code in `Transforms.splitNodes`. My idea is to store `editor.selection` after applying all `split_node` operation then scan the `editor.children` for empty text nodes then remove it by `Transforms.removeNodes` then the set the correct selection by `Transforms.setSelection` . But this has certain problems that is

  - If the users is calling this function inside `Editor.withoutNormalizing` then potentially there could be other empty text nodes which we would have removed which is not an intended effect and there is no way to figure out if we are inside an user created `Editor.withoutNormalizing` since the code is inside the `Editor.withoutNormalizing` which is created by `Transforms.splitNodes`
  
- Second way I thought is when `remove_node` operation is applying we wont change the position of cursor if the operation is due to core normalizing process basically we will have `editor_is_normalizing<Editor, boolean>` week map and we will change normalize code to this

```typescript
 
normalizeNode: (entry: NodeEntry) => {
 
editor_is_normalizing.set(editor, true)
// Normalizing function
editor_is_normalizing.set(editor, false)
}
```
then we wont change the `selection` if the editor is `normalizing`. This also has certain problems that is
 
   -  In real-time collaboration second editor will have no idea if an operation is due to normalizing as result state wont stay consistent among editors, we can try adding information about normalizing to each operation to prevent this error but that means adding extra field to `operation` which does not look right.
  
  
-  My third idea is to just remove the part of `normalizing` process which is removing every single empty `text` nodes. I am not really sure behind the reason why there should not be empty text nodes, but I am sure there is some valid reason so this solution is not going to work

I still think if there is someway to solve this bug then that is to change the core of `spilt_node` operation so that it wont create new text node unnecessarily. 

If there is some other way to solve this bug then I dont know how to do that

nivekithan:
Just for clarification if changing core `split_node` operation is okay then I can come up with a solution that solves this bug

dylans (changes requested):
@nivekithan still planning to work on this. Still blocked by the suggestion for changing split node behavior? Should we close this out or keep it open?

nivekithan:
@dylans I should have closed this pr, #4169 and #4166 long now. Anyways I will be closing those now.

dylans:
@nivekithan thanks!

PR #4166, opened 2021-04-03
respect reverse option in Editor.nodes

nivekithan:
**Description**
As of now `reverse` option in `Editor.nodes` will be ignored if the type of `at` is `Span`, I dont think this is intentional

**Checks**

### DIFF
--- a/.changeset/forty-carrots-look.md
+++ b/.changeset/forty-carrots-look.md
@@ -1,0 +1,5 @@
+---
+'slate': minor
+---
+
+`reverse` option in `Editor.nodes` will not be ignored if the type of `at` is `Span`
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -885,8 +885,8 @@ export const Editor: EditorInterface = {
     let to
 
     if (Span.isSpan(at)) {
-      from = at[0]
-      to = at[1]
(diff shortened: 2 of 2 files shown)


nivekithan:
With the tests failing it seems that, it is intentional ?

nivekithan:
[sandbox](https://codesandbox.io/s/slate-reproductions-forked-s26ys?file=/index.js)

- In above sandox  if you open `console` you will see four `text` nodes. 
- Now if you change `reverse` to `true` then you will see `Editor` obj in `console`

If this is expected result then my pr is wrong. 

But I dont think thats the case as you said `reverse` option will just reverse the output nodes it wont return completely new nodes.

After changing code just like in pr, setting `reverse` to `true` will return 4 `text` nodes which will be in reverse order, which I think is expected result.

thesunny:
Well, completely misread that. Sorry and thanks for the code example! :)

If the test is written incorrectly, please fix the test so that the PR can be merged.

nivekithan:
NP , I should have added the sandox in pr itself to avoid any confusion.

I did take a look at failing tests, some failing tests seems to have no relation to the code I changed, I should go deeper to figure things out. 

I can do that but as of now, I am having exams so only after the exams are finished I can figure out why tests are failing

dylans (changes requested):
@nivekithan let us know if you want to continue working on this PR, or if we should close it out.

PR #4167, opened 2021-04-03
readme: minor fix for updated contributing link

udayanshevade:
**Description**

Wanted to refresh my mind on the contributing guidelines, and saw that the link no longer worked, so just thought to update it really quick.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -154,6 +154,6 @@ Slate's codebase is monorepo managed with [Lerna](https://lerna.js.org/). It con
 
 ### Contributing!
 
-All contributions are super welcome! Check out the [Contributing instructions](./Contributing.md) for more info!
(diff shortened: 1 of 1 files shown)


thesunny:
@udayanshevade That seems like it would be helpful. May I recommend that it points to the online documentation instead as it will be easier to read. It's also consistent with the fact other links point to online URLs as well.

[https://docs.slatejs.org/general/contributing](https://docs.slatejs.org/general/contributing)

With that change, I can merge the PR

udayanshevade:
Sure thing. I thought about whether to point it to the docs but just went with how it had been before. Let me push the link to the site instead.

thesunny:
Thanks! Merged.

TheSpyder:
There are actually two links to contributing, you fixed the one at the end but there's another at the top 🤭

Issue #4038, opened 2021-01-05
Bold/italic/underline can be applied to more than selected text


muster-mark:
#### Do you want to request a _feature_ or report a _bug_?

Report a bug

#### What's the current behavior?
![bug1](https://user-images.githubusercontent.com/12281979/103677914-d00c3f00-4f7a-11eb-9e1b-e70d15cf24ad.gif)

Note that the selection was started by double clicking on the first word.
Not providing a sandbox as the issue is appearing on slate.js website - above gif was recorded on [https://www.slatejs.org/examples/hovering-toolbar](https://www.slatejs.org/examples/hovering-toolbar)
Slate version: whatever version is currently used for examples on www.slatejs.org, presumable the latest
On Firefox 84, Mac OS 11.1. Does not happen in Chrome.

#### What's the expected behavior?

The bold/italic/underline should only be applied the the text that is selected.

BrentFarese:
I think this bug occurs because the [splitting logic](https://github.com/ianstormtaylor/slate/blob/b5859b7e2ef97cdc5d5aaa675b807c4783b2e83c/packages/slate/src/transforms/node.ts#L584-L595) in `setNodes` has a bug.  Specifically, in splitNodes at [these lines](https://github.com/ianstormtaylor/slate/blob/b5859b7e2ef97cdc5d5aaa675b807c4783b2e83c/packages/slate/src/transforms/node.ts#L741-L753) doesn't split a node if the selection is at an edge of the node, even if `split: true`.  You can see in your example that your selection is at the edge of the italic word "shows", then you expand the selection so the anchor is either at the start of "shows" or more likely the end of "This example ".  Since the anchor is at the end of "This example ", Slate will then not split the text between "This example " and the italic "shows".

Can you try your bug if you set `always: true` on `setNodes` and see if the bug goes away?  It would be great if you could submit a PR that improves the splitting logic in `setNodes`!

In our local branch, we always split if the end point of the selection is not at the end of the related node, and always if the start point of the selection is not at the start of the related node.  It looks like this:

```
const atRange = Editor.range(editor, at);
    const rangeCollapsed = Range.isCollapsed(atRange);
    const rangeRef = Editor.rangeRef(editor, atRange, {
      affinity: rangeCollapsed ? 'forward' : 'inward'
    });

    Editor.withoutNormalizing(editor, () => {
      if (rangeCollapsed) {
[...]

PR #4168, opened 2021-04-04
Correcting the splitting logic in setNodes

ridhambhat:
**Description**

This pull request aims to solves a couple of things:
- In setNodes, if splitting has to happen, it should be set to always at :
  - [start](https://github.com/ridhambhat/slate/blob/d808c4ba6f36829064dc2cce42bcb502502343ca/packages/slate/src/transforms/node.ts#L603-L610) if anchor is not at the start of the node
  - [end](https://github.com/ridhambhat/slate/blob/d808c4ba6f36829064dc2cce42bcb502502343ca/packages/slate/src/transforms/node.ts#L595-L602) if focus is not at the end of the node
- In [normalizeNode](https://github.com/ridhambhat/slate/blob/d808c4ba6f36829064dc2cce42bcb502502343ca/packages/slate/src/create-editor.ts#L195), while iterating through children, the variable [prev](https://github.com/ridhambhat/slate/blob/d808c4ba6f36829064dc2cce42bcb502502343ca/packages/slate/src/create-editor.ts#L230) pointed towards wrong node in some cases. Fixed that by getting value of prev from updated node. 

**Issue**

Fixes: [#4038](https://github.com/ianstormtaylor/slate/issues/4038)

**Example**

*Old behavior*
Old behavior
*New behavior*
New Behaviour

**Context**

For the splitting logic, if always is not set for the mentioned conditions, then for the case where anchor is at the end of one node and focus is at the start of another node, the selection bleeds through and properties is applied on nodes where it is not intended. It is safer to split nodes and have empty nodes which will be normalized later.

For the normalization, in particular conditions when a text node was sandwiched between two empty text nodes, then due to the way previous variable was calculated in logic before, it pointed to the wrong node, not taking in account that it could have been merged/deleted. To correct this behavior, now prev is taken from an updated copy of the base node at each iteration.

**Checks**

### DIFF
--- a/.changeset/two-lies-appear.md
+++ b/.changeset/two-lies-appear.md
@@ -1,0 +1,5 @@
+---
+"slate": patch
+---
+
+Fixed a bug in splitting and applying overlapping marks to text nodes.
--- a/.changeset/two-lies-appear.md
+++ b/.changeset/two-lies-appear.md
@@ -1,5 +1,5 @@
 ---
-"slate": patch
+slate: patch
 ---
 
 Fixed a bug in splitting and applying overlapping marks to text nodes.
--- a/packages/slate/src/create-editor.ts
+++ b/packages/slate/src/create-editor.ts
@@ -224,8 +224,10 @@ export const createEditor = (): Editor => {
       let n = 0
 
       for (let i = 0; i < node.children.length; i++, n++) {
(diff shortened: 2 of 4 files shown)


ianstormtaylor:
Thanks @ridhambhat!

PR #4175, opened 2021-04-07
fix(slate-react): use Text.equals to compare leaf

gyh9457:
**Description**
fix `MenoizedLeaf` function

**Issue**

**Example**
```
  prev: { text: 'a' }
  next: { text: 'a', highlight: true }

  matches(text: Text, props: Partial<Text>): boolean {
    for (const key in props) {
      if (key === 'text') {
        continue
      }

      if (!text.hasOwnProperty(key) || text[key] !== props[key]) {
        return false
      }
    }

    return true
  },

  Text.matches(next.leaf, prev.leaf) // compare

  // in the matches function ,props do not include `highlight`. so it return true. but in fact ,it should return false
```

**Checks**


packages/slate-react/src/components/leaf.tsx
```diff
@@ -92,7 +92,7 @@ const MemoizedLeaf = React.memo(Leaf, (prev, next) => {
     next.renderLeaf === prev.renderLeaf &&
     next.text === prev.text &&
     next.leaf.text === prev.leaf.text &&
-    Text.matches(next.leaf, prev.leaf) &&
+    Text.equals(prev.leaf, next.leaf, { loose: true }) &&
```
ianstormtaylor, line 95: Great catch @gyh9457! Would it be possible to remove the previous line now (and the `loose` option) to simplify the logic even further. Instead of:

```
     next.text === prev.text &&
     next.leaf.text === prev.leaf.text &&
     Text.equals(prev.leaf, next.leaf, { loose: true }) &&
```

We'd have:

```
     next.text === prev.text &&
     Text.equals(prev.leaf, next.leaf) &&
```

gyh9457: done   @ianstormtaylor

dylans:
Part of this was handled in #4347.

Need to decide if we can safely also remove `next.leaf.text === prev.leaf.text &&`

packages/slate-react/src/components/leaf.tsx
```diff
@@ -104,8 +104,7 @@ const MemoizedLeaf = React.memo(Leaf, (prev, next) => {
     next.renderLeaf === prev.renderLeaf &&
     next.renderPlaceholder === prev.renderPlaceholder &&
     next.text === prev.text &&
-    next.leaf.text === prev.leaf.text &&
-    Text.equals(next.leaf, prev.leaf) &&
+    Text.equals(prev.leaf, next.leaf) &&
```
dylans, line 107:
```suggestion
    Text.equals(next.leaf, prev.leaf) &&
```

dylans: Changing this order back as it reduces the PR to what has actually changed.

dylans (approved):

PR #4177, opened 2021-04-07
Generic typescript

ianstormtaylor:
**Description**

This is an attempt at making Slate work with TypeScript generics.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/3725
Fixes: https://github.com/ianstormtaylor/slate/issues/3416

**Example**

The best example is looking at the [new TypeScript example](https://github.com/ianstormtaylor/slate/blob/generic-typescript/site/examples/typescript.tsx) on this branch. (Other examples have been purposefully simplified to have very minimal type annotations, to keep them easier to understand.)

**Context**

This works by changing the `Editor` type to be `Editor<V>` where `V` represents the "value" being edited by Slate. In the most generic editor, `V` would be equivalent to `Element[]` (since that is what is accepted as children of the editor). But in a custom editor, you might have `Editor<Array<Paragraph | Quote>>`.

Other `Editor`-and-`Node`-related methods have been also made generic, so for example if you use `Editor.leaf(editor, path)` it knows that the return value is a `Text` node. But more specifically, it knows that it is the text node of the type you've defined in your custom elements (with any marks you've defined).

This replaces the declaration merging approach, and provides some benefits. One of the drawbacks to declaration merging was that it was impossible to know whether you were dealing with an "unknown" or "known" element, since the underlying type was changed. Similarly, having two editors on the page with different schemas wasn't possible to represent. Hopefully this approach with generics will be able to smoothly replace the declaration merging approach. (While being easy to migrate to, since you can pass those same custom element definitions into `Editor` still.)

It's still a work in progress.

**Checks**


packages/slate-history/src/history-editor.ts
```diff
@@ -24,39 +24,39 @@ export const HistoryEditor = {
    * Check if a value is a `HistoryEditor` object.
    */
 
-  isHistoryEditor(value: any): value is HistoryEditor {
+  isHistoryEditor(value: any): value is HistoryEditor<Value> {
```
BrentFarese, line 27: Missed a `V extends Value` type parameter?

```suggestion
  isHistoryEditor<V extends Value>(value: any): value is HistoryEditor<V> {
```

ianstormtaylor: Good question. I'm not really sure how to handle the checking functions in general (like `Element.isElement`, or this one). I'm not sure there's a way to accurately infer `V` in that case without the user passing it in? And even if they pass it in, the function itself isn't verifying that it objects any sort of value?

ChildishForces:
Excited about this branch, it makes a lot more sense than overwriting type definitions. Is there a rough ETA before an @next release?

cmditch:
Also pretty excited about this. Thanks for exploring this!

LoganDark:
I'd switch to Slate in a heartbeat if it didn't require using some weird hacky global override thing -- unfortunately it looks like this PR is now broken. :/

image

packages/slate/src/create-editor.ts
```diff
@@ -139,20 +142,31 @@ export const createEditor = (): Editor => {
...
-    insertFragment: (fragment: Node[]) => {
+    // @ts-ignore
+    insertFragment: (
+      fragment: Array<ElementOf<Editor<V>> | TextOf<Editor<V>>>
+    ) => {
       Transforms.insertFragment(editor, fragment)
     },
 
-    insertNode: (node: Node) => {
+    // @ts-ignore
```
zbeyens, line 163: This would create issues like https://github.com/udecode/plate/issues/936

packages/slate/src/create-editor.ts
```diff
@@ -139,20 +142,31 @@ export const createEditor = (): Editor => {
...
+    insertNode: (
+      node:
+        | ElementOf<Editor<V>>
+        | TextOf<Editor<V>>
+        | Array<ElementOf<Editor<V>> | TextOf<Editor<V>>>
```
zbeyens, line 168: To tackle ts error, we could set:
```suggestion
       VElement<V> | VText<V> | Array<VElement<V> | VText<V>>
```
where
```tsx
export type VElement<V extends Value> = ElementOf<V[0]>;
export type VText<V extends Value> = TextOf<V[0]>;
```
that way, we don't use `Editor<V>` but just `V[0]`

packages/slate/src/create-editor.ts
```diff
@@ -139,20 +142,31 @@ export const createEditor = (): Editor => {
...
+    ) => {
```
zbeyens, line 169: To tackle the circular ts error, we could set:

```suggestion
    insertNode: <EV extends V>(
      node:
        | VElement<V>
        | VText<V>
        | Array<VElement<V> | VText<V>>
    ) => {
```

where
```tsx
export type VElement<V extends Value> = V[number];
export type VText<V extends Value> = Extract<
  V[number]['children'][number],
  TText
>;
```
that way, we don't use `Editor<V>` but just `V`.

Note that `EV` generic is needed to avoid another ts error.

packages/slate/src/interfaces/editor.ts
```diff
@@ -25,23 +29,22 @@ import {
...
+export type Editor<V extends Value> = {
+  children: V
   selection: Selection
   operations: Operation[]
-  marks: Omit<Text, 'text'> | null
+  marks: Record<string, any> | null
+  [key: string]: unknown
 
   // Schema-specific node behaviors.
   isInline: (element: Element) => boolean
```
zbeyens, line 50:
```suggestion
isInline: <EV extends V>(element: VElement<EV>) => boolean
```

zbeyens, line 50:
```suggestion
isInline: <EV extends V>(element: VElement<EV>) => boolean;
    isVoid: <EV extends V>(element: VElement<EV>) => boolean;
    normalizeNode: <EV extends V>(
      entry: TNodeEntry<TEditor<EV> | VElement<EV> | VText<EV>>
    ) => void;
```

zbeyens, line 50:
```tsx
isInline: <EV extends V>(element: VElement<EV>) => boolean;
    isVoid: <EV extends V>(element: VElement<EV>) => boolean;
    normalizeNode: <EV extends V>(
      entry: TNodeEntry<TEditor<EV> | VElement<EV> | VText<EV>>
    ) => void;
```

packages/slate/src/interfaces/editor.ts
```diff
@@ -55,246 +58,28 @@ export interface BaseEditor {
   deleteBackward: (unit: 'character' | 'word' | 'line' | 'block') => void
   deleteForward: (unit: 'character' | 'word' | 'line' | 'block') => void
   deleteFragment: (direction?: 'forward' | 'backward') => void
-  getFragment: () => Descendant[]
+  getFragment: () => Array<Element | Text>
```
zbeyens, line 61:
```suggestion
    getFragment: <EV extends V>() => Array<VElement<EV> | VText<EV>>;
```

packages/slate/src/interfaces/editor.ts
```diff
@@ -55,246 +58,28 @@ export interface BaseEditor {
...
   insertBreak: () => void
-  insertFragment: (fragment: Node[]) => void
-  insertNode: (node: Node) => void
+  insertFragment: (fragment: Array<Element | Text>) => void
+  insertNode: (node: Element | Text | Array<Element | Text>) => void
```
zbeyens, line 64:
```suggestion
insertFragment: <EV extends V>(
      fragment: Array<VElement<EV> | VText<EV>>
    ) => void;
    insertNode: <EV extends V>(
      node: VElement<EV> | VText<EV> | Array<VElement<EV> | VText<EV>>
    ) => void;
```

packages/slate/src/interfaces/editor.ts
zbeyens, line 50:
```tsx
isInline: <EV extends V>(element: VElement<EV>) => boolean;
    isVoid: <EV extends V>(element: VElement<EV>) => boolean;
    normalizeNode: <EV extends V>(
      entry: TNodeEntry<TEditor<EV> | VElement<EV> | VText<EV>>
    ) => void;
```

zbeyens:
I unfortunately don't have the time to complete that PR, so I've forked that PR into Plate. If you're looking to use these types, see [Plate 11](https://github.com/udecode/plate/pull/1517).

maral:
Is there someone kind and brave enough to lift this PR up and continue the work on it? Or is the official approach against this?

dylans:
> Is there someone kind and brave enough to lift this PR up and continue the work on it? Or is the official approach against this?

The official approach is that we definitely want to finish this, but no one has had time to do so. Plate seems to be doing pretty well with bounties attached to work they've not had time to finish, but I haven't wanted to go that route with Slate.

At this point, a good path would probably be to look at what Plate has done, see if it makes sense to wrap this up, and if so land it. More likely there are some things to improve that were uncovered in that work.

After we land this, a decent test suite for slate-react, and some optimizations for Reach 18 and 19, I'm ready to call it 1.0. (Just in time to start thinking about switching to EditContext API if everyone supports it (currently in Chrome, waiting for indication from others): https://w3c.github.io/edit-context/     (this is a substantial change))

Issue #1344, opened 2017-10-30
[Slate-React] Inline placeholder is treated as part of editor content


blakeembrey:
![](https://d26dzxoao6i3hh.cloudfront.net/items/0R0E1k1x3c0x143D001X/Screen%20Recording%202017-10-30%20at%2008.57%20am.gif)

When rendering an inline placeholder, it's treated as part of the document and the cursor always appears after the placeholder (which doesn't feel natural). My thought was maybe this has something to do with rendering and Slate.js measurements including the placeholder for cursor positioning.

```js
    function renderPlaceholder(props) {
      const { document } = props.editor.value

      if (document.text !== '' || document.nodes.size !== 1) return false

      return (
        <span
          contentEditable={false}
          styles={{ pointerEvents: 'none', userSelect: 'none' }}
        >
          {editor.props.placeholder}
        </span>
      )
    }
```

ianstormtaylor:
@blakeembrey thanks for opening this.

The issue is we're kind of constrained by what CSS/HTML allow us to do here. We render the placeholder as a sibling of the node's inner content itself (before the content to be exact). By default, the placeholder has `width: 0; white-space: nowrap;` which ensures that it never messes up the selection, since it takes up no space.

An alternative would be to use `position: absolute`, but this then requires that the block has `position: relative`, which I didn't want to force users to add for the general case. And this also means that the placeholder is taken out of the layout.

I'm not sure there's a way to render `contenteditable={false}` inside the block and have it both take up space _and_ not mess with selection events. I've tried a few different things, but I haven't done a super deep exploration, so I could be wrong. I'd be open to changes to how our default logic works if there's a way to handle the general case better.

blakeembrey:
Based on my limited knowledge of the project and some personal thoughts, here's an approach I would consider taking (not sure if it even matches up with the object model used by Slate.js - if not, I'm happy to create a separate project to demonstrate and flesh out the idea more thoroughly first). 

First, make the placeholder part of the model itself. This could have interesting side-effects later down the line for different use-cases, but if each object in the model manages selection events for itself the "placeholder" object could say "on selection move the selection to the beginning of the element". This would avoid the need entirely needing to have zero width and it'd just be regularly selectable - just the selection event would immediately change (like it is today anyway) to where it says it should go based on selection type.

[...]

PR #4182, opened 2021-04-08
Fix placeholder overflow

Avizura:
**Description**
Fix issue when placeholder horizontally overflows editor bounds.

**Issue**
Fixes: #4018, #2922

**Context**
@Glazy in his [PR](https://github.com/ianstormtaylor/slate/pull/4019) already fixed this issue, however it wasn't correctly [ported to master](https://github.com/Avizura/slate/commit/00393bb18548976d4e79c8bc46d54b2125bb6701). Specifically [this line](https://github.com/ianstormtaylor/slate/pull/4019/files#diff-18ef6af284404b1670857e562c183f2da6774d81bb65ed39d82fe124f769787bL40) wasn't ported for some reason. Also I didn't understand why we should dynamically set the editor `minHeight` and `position: absolute` (it brings vertical overflowing issue) instead of using static positioning.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/leaf.tsx
+++ b/packages/slate-react/src/components/leaf.tsx
@@ -23,25 +23,6 @@ const Leaf = (props: {
     renderLeaf = (props: RenderLeafProps) => <DefaultLeaf {...props} />,
   } = props
 
-  const placeholderRef = useRef<HTMLSpanElement | null>(null)
-
-  useEffect(() => {
-    const placeholderEl = placeholderRef?.current
-    const editorEl = document.querySelector<HTMLDivElement>(
-      '[data-slate-editor="true"]'
-    )
-
-    if (!placeholderEl || !editorEl) {
-      return
-    }
-
-    editorEl.style.minHeight = `${placeholderEl.clientHeight}px`
-
-    return () => {
-      editorEl.style.minHeight = 'auto'
-    }
-  }, [placeholderRef])
-
   let children = (
     <String isLast={isLast} leaf={leaf} parent={parent} text={text} />
   )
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/components/leaf.tsx
```diff
@@ -23,25 +23,6 @@ const Leaf = (props: {
...
-      return
-    }
-
-    editorEl.style.minHeight = `${placeholderEl.clientHeight}px`
-
-    return () => {
-      editorEl.style.minHeight = 'auto'
-    }
-  }, [placeholderRef])
-
```
juliankrispel, line 44: this can't be removed (it's needed to match the height of the editor to the height of the placeholder)

packages/slate-react/src/components/leaf.tsx
```diff
@@ -50,20 +31,17 @@ const Leaf = (props: {
...
             display: 'inline-block',
             width: '100%',
             maxWidth: '100%',
-            whiteSpace: 'nowrap',
             opacity: '0.333',
             userSelect: 'none',
             fontStyle: 'normal',
             fontWeight: 'normal',
             textDecoration: 'none',
-            position: 'absolute',
```
juliankrispel, line 66: removing this breaks the placeholder because it will now be rendered below the first empty line rather than in the same place

juliankrispel (commented):
these changes may address the issues referenced but would also break `placeholder`, even though no tests fail.

packages/slate-react/src/components/leaf.tsx
Avizura: Hi @juliankrispel, could you explain why?
This is demo with current react-slate: https://codesandbox.io/s/clever-pare-ywluq?file=/src/App.js
This is demo with this PR: https://codesandbox.io/s/pensive-noyce-q83gr?file=/src/App.js

Avizura:
Now I see that position absolute was dictated by this https://github.com/ianstormtaylor/slate/issues/1344. Feel free to close then.

Avizura:
#4190

PR #4187, opened 2021-04-10
feat(react-slate): Ability to add multiple placeholders.

xobotyi:
**Description**
As for my case i had a need to add multiple placeholders for a single editable, that are hidden separately as content fills.

**Example**

Above example has such `placeholder` property:
```ts
placeholder={[
  'Summarize your issue.',
  <>
    And describe it in details here.
    <br />
    Try to avoid using languages not supported by interface, it will significantly improve
    and speed up our work.
  </>,
]}
```

**Checks**

### DIFF
--- a/.changeset/nine-llamas-hope.md
+++ b/.changeset/nine-llamas-hope.md
@@ -1,0 +1,14 @@
+---
+'slate-react': minor
+'slate': patch
+---
+
+ReactNode previews and ability to customize it.
+
+- Now previews can be any of ReactNode or ReactNodeArray
+- Leaf component is fully refactored, no effects used anymore, just straight render.
+- Preview styling moved to `DefaultLeaf` component so users now able to change it on their need via `Editable.renderLeaf`.
+- `Editable` component has proper typing now.
+
+BREAKING CHANGE: anyone using `Editable.renderLeaf` and not calling `DefaultLeaf` on placeholder leaves wont have placeholder displayed.
+Sadly this is the only way to allow placeholder customisation.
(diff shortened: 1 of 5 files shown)


ianstormtaylor:
@xobotyi can you see if your use case would be solved by https://github.com/ianstormtaylor/slate/pull/4190 with fragment rendering? It seems like a simpler solution so I'd rather go with it if possible.

xobotyi:
@ianstormtaylor current placeholder will disappead at the moment first character entered. Thus no, it cant be achieved with renderer.

ianstormtaylor:
@xobotyi I believe your case could instead be solved by handling the placeholder logic at the block level? One for the heading and one for the paragraph like with the Forced Layout example?

xobotyi:
@ianstormtaylor and no again, at block level i wont be able to gain access to keypress handlers to move between placholders by pressing enter.
Currently if you press Enter and there is empty placeholder ahead, it will switch to that line instead of splitting current one. I dont know how to manage it outside the slate.

ianstormtaylor:
@xobotyi I'm not sure I understand what you mean by splitting the block. But since placeholders internally are implemented with decorations, I believe you could implement a similar behavior in userland with special decoration logic for your two nodes.

xobotyi:
@ianstormtaylor when enter pressed - handler calls `Transforms.splitNodes` which splits current line in two (i did not told about block splitting).
If such question was only about decoration i wold not change the base components, here it is also about behaviour

dylans (changes requested):
@xobotyi unfortunately this one has fallen a bit behind main.

If you're still interested in landing this, please update against the latest and then I'll review. Apologies that this PR was ignored after your last comment.

dylans:

PR #4190, opened 2021-04-11
Add renderPlaceholder

### LINKED ISSUE #4018: Long placeholder values overflow editor

### LINKED ISSUE #4123: slate-react placeholder unable to change the style
**Problem**
image

the leaf with placeholder use the inline-styles，user cannot change the placeholder style ，Can I make the placeholder render configurable .

Even if I rewrite RenderLeaf, I can't  find the placeholder node by *symbol* tag .

juliankrispel:
**Description**
Adds a `renderPlaceholder` prop to `Editable`

**Issue**
Fixes #4123
Fixes #4018 by removing `whitespace: nowrap`

**Example**
```tsx
<Editable
  renderPlaceholder={props => {
    return (
      <div {...props}>
        <p>Hello there</p>
        <pre>Here's something rendered with renderPlaceholder</pre>
      </div>
    )
  }}
/>
```

image

**Context**

**Checks**


packages/slate-react/src/components/leaf.tsx
```diff
@@ -40,7 +40,7 @@ const Leaf = (props: {
     return () => {
       editorEl.style.minHeight = 'auto'
     }
-  }, [placeholderRef])
+  }, [placeholderRef, leaf])
```
juliankrispel, line 43: Adding this dependency actually executes this hook now when the placeholderRef has a element referenced. This didn't work before

packages/slate-react/src/components/editable.tsx
```diff
@@ -98,6 +114,7 @@ export type EditableProps = {
   style?: React.CSSProperties
   renderElement?: (props: RenderElementProps) => JSX.Element
   renderLeaf?: (props: RenderLeafProps) => JSX.Element
+  renderPlaceholder?: (props: typeof defaultPlaceholderProps) => JSX.Element
```
ianstormtaylor, line 117: Can you refactor this to be declared the same was as `RenderElementProps/RenderLeafProps` for consistency? So people can use those types in their own code if they need to.

packages/slate-react/src/components/editable.tsx
```diff
@@ -473,16 +492,20 @@ export const Editable = (props: EditableProps) => {
 
   const decorations = decorate([editor, []])
 
+  const renderedPlaceholder =
+    renderPlaceholder?.(defaultPlaceholderProps) ||
+    (placeholder && <span {...defaultPlaceholderProps}>{placeholder}</span>)
```
ianstormtaylor, line 497: Can you pull this logic out into a `defaultRenderPlaceholder` function, so that it always exists and we can simplify this to always calling the function here?

packages/slate-react/src/components/leaf.tsx
```diff
@@ -53,17 +53,10 @@ const Leaf = (props: {
           ref={placeholderRef}
           contentEditable={false}
           style={{
+            position: 'absolute',
```
ianstormtaylor, line 56: Can you explain the changes here? Why isn't it structured so that the only thing rendered is the placeholder returned by `renderPlaceholder`? Instead of the wrapping `<span>` that remains here?

juliankrispel: I wrote it this way thinking it'd be safer for the user but I agree with you this is more consistent with the rest of slate :+1:

Addressed

juliankrispel:
thanks for the review @ianstormtaylor - I addressed your feedback :+1:

packages/slate-react/src/components/editable.tsx
```diff
@@ -474,15 +478,15 @@ export const Editable = (props: EditableProps) => {
   const decorations = decorate([editor, []])
 
   if (
-    placeholder &&
+    (renderPlaceholder || placeholder) &&
```
ianstormtaylor, line 481: I think this should still only check for `placeholder` (since the renderer will be defaulted).

More generally, I don't think we're currently solving for the ability to define any arbitrary rendering logic in the `renderPlaceholder` override. If we were (which we could discuss in a separate issue) we'd likely want it to include the check right below this as well in the rendering layer:

```ts
editor.children.length === 1 &&
Array.from(Node.texts(editor)).length === 1 &&
Node.string(editor) === ''
```

Since without that included it wouldn't be very customizable.

packages/slate-react/src/components/leaf.tsx
```diff
@@ -40,34 +58,38 @@ const Leaf = (props: {
...
+        pointerEvents: 'none',
+        width: '100%',
+        maxWidth: '100%',
+        display: 'block',
+        opacity: '0.333',
+        userSelect: 'none',
+        fontStyle: 'normal',
+        fontWeight: 'normal',
+        textDecoration: 'none',
+      } as React.CSSProperties,
```
ianstormtaylor, line 82: And then all this placeholder rendering logic can move out of the `<Leaf>` and be consolidated into the top-level `<Editable>`. 

The only props we'd be passing in to the `renderPlaceholder` would be `children` (which is the `placeholder` prop from the editor) and `ref`. The other things here would just be defined inline in our new `<DefaultPlaceholder>` component.

packages/slate-react/src/components/leaf.tsx
```diff
@@ -40,34 +58,38 @@ const Leaf = (props: {
...
-          }}
-        >
-          {leaf.placeholder}
-        </span>
+        {typeof leaf.placeholder === 'string'
+          ? defaultRenderPlaceholder({
+              ...placeholderProps,
+              children: leaf.placeholder,
+            })
+          : leaf.placeholder(placeholderProps)}
```
ianstormtaylor, line 92: This could then be simplified to just the `leaf.placeholder`.

ianstormtaylor (commented):
Hey @juliankrispel thanks for the update!

I was just going through to review again, and I think I realized that we're trying to solving a few different problems at once here with the same API, and I'm not sure we're doing it right. For a summary of how I'm thinking about this, we have a few problems we're solving for:

- For the most common case, a simple `placeholder="..."` string is needed to mimic the default browser behavior that applies to `<textarea>` elements. We want this to get the same styling that those placeholders do (namely the opacity).

- There's another common use case that is changing the rendering logic for when and where placeholders are rendered. This can be when implementing placeholders in the middle of documents (eg. in a caption block for an image), or when wanting to have multiple placeholders (eg. for a heading and body in a forced layout). I think these cases should be left unsolved by the `placeholder/renderPlaceholder` props, and instead people should use their own custom `decorate` prop instead for maximum control.

- A less common but totally valid case is wanting to render more complex components in place of the default global placeholder. But the issue here is that people may or may not want to render them with the same styles as the default, specifically #4123 doesn't want the opacity styles. Right now we're forcing those styles, but we could tweak the logic to just use a `<DefaultPlaceholder>` similar to how we have a `<DefaultLeaf>`, allowing them to override it with `renderPlaceholder`.

I added a few comments in line to explain what the changes would be to arrive at that. It's a little complex to understand, but the goal is to make `renderPlaceholder` a very simple but fully customizable escape hatch that runs in the default code path.

Let me know if you have questions!

packages/slate-react/src/components/editable.tsx
```diff
@@ -114,6 +116,7 @@ export const Editable = (props: EditableProps) => {
     readOnly = false,
     renderElement,
     renderLeaf,
+    renderPlaceholder,
```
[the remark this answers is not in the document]
juliankrispel: I wanted to just make it a component. The reason I didn't was because I would have to use forwardRef to keep the [resizing logic](https://github.com/ianstormtaylor/slate/pull/4190/files#diff-18ef6af284404b1670857e562c183f2da6774d81bb65ed39d82fe124f769787bR46-R60) intact.

packages/slate-react/src/components/editable.tsx
juliankrispel: I've reverted this so it only checks for placeholder and now instead of using the placeholder variable to pass the render prop I'm passing it through similarly to `renderElement` and `renderLeaf`.

Regarding the second part of your message I'm not entirely sure I follow. Would you like me to include this or shall we deal with it in a separate change?

juliankrispel:
thanks for the tight feedback loop @ianstormtaylor much appreciated 🙏 

I addressed your feedback to the best of my abilities and there are two open points remaining afaik.

Also added an example for custom placeholders.

packages/slate-react/src/components/leaf.tsx
juliankrispel: I've made these changes except I have kept the `contentEditable: false` in the props passed to `renderPlaceholder` because removing this would break the editor and make the placeholder editable

I've removed the style prop as suggested but made it an export that people can use it for convenience (since without it the placeholder render will break the editor like so:

image

Also added an example to show how this can be used.

packages/slate-react/src/components/editable.tsx
```diff
@@ -1049,6 +1053,37 @@ export const Editable = (props: EditableProps) => {
   )
 }
 
+/**
+ * The props that get passed to renderPlaceholder
+ */
+export type RenderPlaceholderProps = {
+  contentEditable: boolean
+  children?: React.ReactNode
+  ref: React.RefObject<any>
```
ianstormtaylor, line 1062: Can you put these `ref` and `contentEditable` keys inside an `attributes` hash? 

The pattern is essentially that there will be `children, attributes, ...stuff` where `attributes` must be mixed on to the component, and `children` is special-cased. (We might want to change this in the future for all of them at once, since it's a little strange, but let's go with consistency for now.)

packages/slate-react/src/components/editable.tsx
```diff
@@ -1049,6 +1053,37 @@ export const Editable = (props: EditableProps) => {
...
+export const defaultPlaceholderStyle: React.CSSProperties = {
+  position: 'absolute',
+  pointerEvents: 'none',
+  width: '100%',
+  maxWidth: '100%',
+  display: 'block',
+  opacity: '0.333',
+  userSelect: 'none',
+  textDecoration: 'none',
+}
```
ianstormtaylor, line 1077: I totally see what you mean now about how these styles are required to get things to work properly, my bad for not understanding before. In that case having it inside `attributes.style` as you did before sounds good to me, and people can cherry-pick if they really need to.

Just since I'd rather not add another export for people to keep track of.

packages/slate-react/src/components/editable.tsx
```diff
@@ -1049,6 +1053,37 @@ export const Editable = (props: EditableProps) => {
...
 }
 
+/**
+ * The props that get passed to renderPlaceholder
+ */
+export type RenderPlaceholderProps = {
+  contentEditable: boolean
+  children?: React.ReactNode
+  ref: React.RefObject<any>
+}
```
ianstormtaylor, line 1063: Can you also add an `attributes['data-slate-placeholder']: true` to that dictionary as well? That will be consistent with the others and allow people to see it clearly in DevTools as well as make simple tweaks to styles with overrides when necessary.

packages/slate-react/src/components/editable.tsx
```diff
@@ -1049,6 +1053,37 @@ export const Editable = (props: EditableProps) => {
...
+  textDecoration: 'none',
+}
+
+/**
+ * The default placeholder render method
+ */
+
+export const defaultRenderPlaceholder = (props: RenderPlaceholderProps) => (
+  <span style={defaultPlaceholderStyle} {...props} />
+)
```
ianstormtaylor, line 1085: I see what you mean about needing `forwardRef` to keep the resize logic. Is there any reason `forwardRef` wasn't working? It doesn't seem bad in my opinion to have a `<DefaultPlaceholder>` component that has ref-forwarding built in. And would stay consistent that way.

And then that `DefaultPlaceholder` component can be exported (like the others), and this little rendering function can stay private.

packages/slate-react/src/components/leaf.tsx
```diff
@@ -12,6 +12,7 @@ const Leaf = (props: {
   isLast: boolean
   leaf: Text
   parent: Element
+  renderPlaceholder: (props: RenderPlaceholderProps) => JSX.Element
```
ianstormtaylor, line 15: I think this new prop needs to be reflected in `MemoizedLeaf` below.

.changeset/few-planets-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+'slate-react': minor
```
ianstormtaylor, line 2: This part is counter-intuitive, but since we're pre-`1.0` we use `patch` for non-breaking releases for now.

.changeset/few-planets-brush.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Added renderPlaceholder prop to Editable component
```
ianstormtaylor, line 5: Can you change this to:

```
Added a `renderPlaceholder` prop to the `<Editable>` component for customizing how placeholders are rendered.
```
> Added a `renderPlaceholder` prop to the `<Editable>` component for customizing how placeholders are rendered.

Just because it will show up in the changelog so I try to keep it nicely written. Usually I just make edits to this on people's PRs by myself, but just putting this comment here since you're a maintainer so you know for the future.

ianstormtaylor (commented):
Thanks @juliankrispel, this is looking great. I realized you were totally right about mixing in the `style` attribute, my bad for arguing against that. I put a few more comments in line for refining. I love the idea for the example of the customized placeholder!

packages/slate-react/src/components/editable.tsx
juliankrispel: I assigned `children` the `any` prop, consistent with the other render prop interfaces, is that what you meant by:

> children is special-cased

packages/slate-react/src/components/editable.tsx
juliankrispel: Not sure what I did the last time but this of course absolutely works, no forwardRef required

packages/slate-react/src/components/leaf.tsx
juliankrispel: good catch!

juliankrispel:
> Thanks @juliankrispel, this is looking great. I realized you were totally right about mixing in the `style` attribute, my bad for arguing against that. I put a few more comments in line for refining. I love the idea for the example of the customized placeholder!

Thank you for the review @ianstormtaylor, I much appreciate the help!

I've addressed all the feedback in your most recent review, cheers

packages/slate-react/src/components/editable.tsx
```diff
@@ -114,6 +115,7 @@ export const Editable = (props: EditableProps) => {
     readOnly = false,
     renderElement,
     renderLeaf,
+    renderPlaceholder = DefaultPlaceholder,
```
ianstormtaylor, line 118: This shouldn't be just the component as default but instead a function that returns the component like:

```tsx
renderPlaceholder = (props) => <DefaultPlaceholder {...props} />
```

packages/slate-react/src/components/leaf.tsx
```diff
@@ -112,3 +107,17 @@ export const DefaultLeaf = (props: RenderLeafProps) => {
...
+const defaultPlaceholderStyle: React.CSSProperties = {
+  position: 'absolute',
+  pointerEvents: 'none',
+  width: '100%',
+  maxWidth: '100%',
+  display: 'block',
+  opacity: '0.333',
+  userSelect: 'none',
+  textDecoration: 'none',
+}
```
ianstormtaylor, line 123: Now that we're not exporting can you inline this so it's easier to find?

ianstormtaylor (commented):
Everything looks great @juliankrispel! Two comments inline and then I'm good to merge.

juliankrispel:
> Everything looks great @juliankrispel! Two comments inline and then I'm good to merge.

thanks, all feedback addressed

ianstormtaylor (approved):

PR #4191, opened 2021-04-11
Move hasRange to core

### LINKED ISSUE #4160: move `hasRange` method to core `Editor`
**Description**
Move the `hasRange` method from the `ReactEditor` to the core `Editor`.

**Context**
This was added in a previous pull request and I didn't realize it was in the wrong place. There's nothing React-specific about the logic so it doesn't make sense for it to be where it is currently.

juliankrispel:
**Description**
Moves hasRange to core Editor, as described in https://github.com/ianstormtaylor/slate/issues/4160#issue-848635767

**Issue**
Fixes #4160

**Example**
n/a

**Context**
Moves `hasRange` from `ReactEditor` to `Editor`.
Non-breaking since ReactEditor inherits from Editor,

**Checks**

### DIFF
--- a/.changeset/dull-kiwis-drive.md
+++ b/.changeset/dull-kiwis-drive.md
@@ -1,0 +1,6 @@
+---
+'slate': minor
+'slate-react': minor
+---
+
+Move hasRange to core
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -188,7 +188,7 @@ export const Editable = (props: EditableProps) => {
     // then its children might just change - DOM responds to it on its own
     // but Slate's value is not being updated through any operation
     // and thus it doesn't transform selection on its own
-    if (selection && !ReactEditor.hasRange(editor, selection)) {
+    if (selection && !Editor.hasRange(editor, selection)) {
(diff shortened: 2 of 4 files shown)


ianstormtaylor:
Thanks for this @juliankrispel! I think this is actually breaking unfortunately because the `ReactEditor` namespace doesn't inherit all of the helper utilities from the `Editor` namespace. 

I've got a related fix in https://github.com/ianstormtaylor/slate/pull/4177 that re-implements `hasRange` as `Range.exists` since I realized while moving it that there's nothing editor-specific about how it was written. I'm going to close in favor of that one, hope that makes sense. (Sorry for not properly linking that issue to give you a heads up that it would be solved!)

juliankrispel:
> Thanks for this @juliankrispel! I think this is actually breaking unfortunately because the `ReactEditor` namespace doesn't inherit all of the helper utilities from the `Editor` namespace.
> 
> I've got a related fix in #4177 that re-implements `hasRange` as `Range.exists` since I realized while moving it that there's nothing editor-specific about how it was written. I'm going to close in favor of that one, hope that makes sense. (Sorry for not properly linking that issue to give you a heads up that it would be solved!)

no worries thanks for clearing that up

PR #4192, opened 2021-04-11
Remove execute file permission from .ts files

beorn:
These .ts files has executable file permission - I just removed the execute bit.


ianstormtaylor:
Thanks @beorn!

PR #4193, opened 2021-04-11
Save some remove_text calls if the text is zero length

beorn:
There's no need to remove_text of 0 length, so prevent that from happening.

It's arguably an optimization, and some plugins that hook into editor.apply may balk this happening.  So, it seems like a good idea to do in general.  If you don't agree with this, then just close this PR.

**Checks**

### DIFF
--- a/packages/slate/src/transforms/general.ts
+++ b/packages/slate/src/transforms/general.ts
@@ -44,6 +44,7 @@ export const GeneralTransforms: GeneralTransforms = {
 
       case 'insert_text': {
         const { path, offset, text } = op
+        if (text.length === 0) break
         const node = Node.leaf(editor, path)
         const before = node.text.slice(0, offset)
         const after = node.text.slice(offset)
@@ -167,7 +168,7 @@ export const GeneralTransforms: GeneralTransforms = {
 
       case 'remove_text': {
         const { path, offset, text } = op
-        if (!text || text.length === 0) break
+        if (text.length === 0) break
         const node = Node.leaf(editor, path)
         const before = node.text.slice(0, offset)
         const after = node.text.slice(offset + text.length)
--- a/packages/slate/src/transforms/general.ts
+++ b/packages/slate/src/transforms/general.ts
@@ -167,6 +167,7 @@ export const GeneralTransforms: GeneralTransforms = {
 
       case 'remove_text': {
         const { path, offset, text } = op
+        if (!text || text.length === 0) break
         const node = Node.leaf(editor, path)
         const before = node.text.slice(0, offset)
         const after = node.text.slice(offset + text.length)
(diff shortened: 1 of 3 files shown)


packages/slate/src/transforms/general.ts
```diff
@@ -167,6 +167,7 @@ export const GeneralTransforms: GeneralTransforms = {
 
       case 'remove_text': {
         const { path, offset, text } = op
+        if (!text || text.length === 0) break
```
ianstormtaylor, line 170: Can you remove the `!text` check, it's unnecessary since `text` seems to always be a string here.

packages/slate/src/transforms/text.ts
```diff
@@ -177,7 +177,8 @@ export const TextTransforms: TextTransforms = {
         const { path } = point
         const { offset } = start
         const text = node.text.slice(offset)
-        editor.apply({ type: 'remove_text', path, offset, text })
+        if (text.length > 0)
+          editor.apply({ type: 'remove_text', path, offset, text })
```
ianstormtaylor, line 181: Can you remove these checks in this file, since the change above makes `editor.apply` a no-op already I don't think we need to guard against it everywhere else?

ianstormtaylor (commented):
Thanks @beorn! Do you have a use case this was inspired by?

packages/slate/src/transforms/general.ts
beorn: Ok, I pushed a change for this.

I also added a similar check for 'insert_text'.

beorn:
> Thanks @beorn! Do you have a use case this was inspired by?

To be honest, it's a year ago, and I believe there were some unfortunate side-effects of 0-length remove_texts and integration with Automerge, but it is a while ago, so I could be wrong.  We could also just close this PR until the need is more clear, if you like.

packages/slate/src/transforms/text.ts
beorn: And/or, leave the checks here and remove them from editor.apply?

The benefit of having them here is that they'll save these no-op editor.apply calls from even happening, also saving any plugins that hook into editor.apply from dealing with them.  I believe my issues from this actually came from plugins not handling it well.  Arguably plugins should just check that text.length > 0 as well, but then that creates a kind-of undocumented protocol for this.  Up to you, lmk :)

ianstormtaylor: Ah interesting, that's a good point. In that case keeping them sounds both like a good idea I agree. Thanks!

ianstormtaylor:
That's a great reason I think, so I'm happy to merge. Thanks for this @beorn!

TheSpyder:
I admit I considered a change like this when I started seeing a lot of zero-length insert and remove text operations in my collaboration property tests. I wasn't convinced removing them was a good idea, so I was planning to leave it until I had 100% of my tests passing.

I'm still not there, but I have updated my copy of Slate and all of my existing tests pass so it looks good to me!

PR #4197, opened 2021-04-13
Updates outdated TypeScript link in first walkthrough document

scotchfield:
**Description**
This PR updates TypeScript link in the `Installing Slate` documentation for new users to a valid reference. Looks like this URL changed in the past, so this is just fixing a 404. Thanks!


ianstormtaylor:
Thanks @scotchfield!

PR #4200, opened 2021-04-14
Feature/android keyboard support

wleroux:
**Description**
This adds a preliminary support for android keyboard support. This is targetted specifically for Chrome on Android 10 and Chrome on Android 11.

**Context**
DOM Mutations are observerd and analyzed to determine the slate operations required to produce those DOM Mutations. While ideally we would capture the composition and keyboard events, there are variations between android versions and there are no governing specification for the expected behaviors. 

An error boundary was added to the android slate editor to capture any exception thrown. On android, the keyboard may mutate the DOM independently from the JavaScript thread. As such, DOM modifications may occur at any given time and cause issues. To resolve these issues, we capture the error and log it. If an error is caught, the state of the editor will be reset and the user will have to redo the operation that caused an issue, however it will not prevent the editor from working.

As this involves a lot of overriding of the DOM events, the default Editable was renamed to DefaultEditable and used by default unless on an Android device.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/android/android-editable.tsx
+++ b/packages/slate-react/src/components/android/android-editable.tsx
@@ -10,7 +10,7 @@ import { Descendant, Editor, Node, Range, Transforms } from 'slate'
 import throttle from 'lodash/throttle'
 import scrollIntoView from 'scroll-into-view-if-needed'
 
-import { ReactEditor } from '../..'
+import { DefaultPlaceholder, ReactEditor } from '../..'
 import { ReadOnlyContext } from '../../hooks/use-read-only'
 import { useSlate } from '../../hooks/use-slate'
 import { useIsomorphicLayoutEffect } from '../../hooks/use-isomorphic-layout-effect'
@@ -38,7 +38,8 @@ import {
   hasEditableTarget,
   isEventHandled,
   isTargetInsideVoid,
-} from '../default-editable'
+} from '../editable'
+import { IS_FIREFOX } from '../../utils/environment'
 
 export const AndroidEditableNoError = (props: EditableProps): JSX.Element => {
   return (
@@ -61,6 +62,7 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
     readOnly = false,
     renderElement,
     renderLeaf,
+    renderPlaceholder = props => <DefaultPlaceholder {...props} />,
     style = {},
     as: Component = 'div',
     ...attributes
@@ -109,7 +111,8 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
   // Whenever the editor updates, make sure the DOM selection state is in sync.
   useIsomorphicLayoutEffect(() => {
     const { selection } = editor
-    const domSelection = window.getSelection()
+    const root = ReactEditor.findDocumentOrShadowRoot(editor)
+    const domSelection = root.getSelection()
 
     if (state.isComposing || !domSelection || !ReactEditor.isFocused(editor)) {
       return
@@ -133,12 +136,23 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
     }
 
     // If the DOM selection is in the editor and the editor selection is already correct, we're done.
-    if (
-      hasDomSelection &&
-      hasDomSelectionInEditor &&
-      selection &&
-      Range.equals(ReactEditor.toSlateRange(editor, domSelection), selection)
-    ) {
+    if (hasDomSelection && hasDomSelectionInEditor && selection) {
+      const slateRange = ReactEditor.toSlateRange(editor, domSelection, {
+        exactMatch: true,
+      })
+      if (slateRange && Range.equals(slateRange, selection)) {
+        return
+      }
+    }
+
+    // when <Editable/> is being controlled through external value
+    // then its children might just change - DOM responds to it on its own
+    // but Slate's value is not being updated through any operation
+    // and thus it doesn't transform selection on its own
+    if (selection && !ReactEditor.hasRange(editor, selection)) {
+      editor.selection = ReactEditor.toSlateRange(editor, domSelection, {
+        exactMatch: false,
+      })
       return
     }
 
@@ -146,42 +160,47 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
     const el = ReactEditor.toDOMNode(editor, editor)
     state.isUpdatingSelection = true
 
-    try {
-      const newDomRange = selection && ReactEditor.toDOMRange(editor, selection)
+    const newDomRange = selection && ReactEditor.toDOMRange(editor, selection)
 
-      if (newDomRange) {
-        if (Range.isBackward(selection!)) {
-          domSelection.setBaseAndExtent(
-            newDomRange.endContainer,
-            newDomRange.endOffset,
-            newDomRange.startContainer,
-            newDomRange.startOffset
-          )
-        } else {
-          domSelection.setBaseAndExtent(
-            newDomRange.startContainer,
-            newDomRange.startOffset,
-            newDomRange.endContainer,
-            newDomRange.endOffset
-          )
-        }
-        const leafEl = newDomRange.startContainer.parentElement!
-        scrollIntoView(leafEl, {
-          scrollMode: 'if-needed',
-          boundary: el,
-        })
+    if (newDomRange) {
+      if (Range.isBackward(selection!)) {
+        domSelection.setBaseAndExtent(
+          newDomRange.endContainer,
+          newDomRange.endOffset,
+          newDomRange.startContainer,
+          newDomRange.startOffset
+        )
       } else {
-        domSelection.removeAllRanges()
+        domSelection.setBaseAndExtent(
+          newDomRange.startContainer,
+          newDomRange.startOffset,
+          newDomRange.endContainer,
+          newDomRange.endOffset
+        )
+      }
+      const leafEl = newDomRange.startContainer.parentElement!
+      leafEl.getBoundingClientRect = newDomRange.getBoundingClientRect.bind(
+        newDomRange
+      )
+      scrollIntoView(leafEl, {
+        scrollMode: 'if-needed',
+        boundary: el,
+      })
+      // @ts-ignore
+      delete leafEl.getBoundingClientRect
+    } else {
+      domSelection.removeAllRanges()
+    }
+
+    setTimeout(() => {
+      // COMPAT: In Firefox, it's not enough to create a range, you also need
+      // to focus the contenteditable element too. (2016/11/16)
+      if (newDomRange && IS_FIREFOX) {
+        el.focus()
       }
 
-      setTimeout(() => {
-        state.isUpdatingSelection = false
-      })
-    } catch (err) {
-      // eslint-disable-next-line no-console
-      console.error(err)
       state.isUpdatingSelection = false
-    }
+    })
   })
 
   useLayoutEffect(() => {
@@ -219,9 +238,10 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
       if (!readOnly && !state.isComposing && !state.isUpdatingSelection) {
         inputManager.onSelect()
 
-        const { activeElement } = window.document
+        const root = ReactEditor.findDocumentOrShadowRoot(editor)
+        const { activeElement } = root
         const el = ReactEditor.toDOMNode(editor, editor)
-        const domSelection = window.getSelection()
+        const domSelection = root.getSelection()
 
         if (activeElement === el) {
           state.latestElement = activeElement
@@ -245,12 +265,10 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
           isTargetInsideVoid(editor, focusNode)
 
         if (anchorNodeSelectable && focusNodeSelectable) {
-          try {
-            const range = ReactEditor.toSlateRange(editor, domSelection)
-            Transforms.select(editor, range)
-          } catch (e) {
-            // expected if the value hasn't been updated yet
-          }
+          const range = ReactEditor.toSlateRange(editor, domSelection, {
+            exactMatch: false,
+          })
+          Transforms.select(editor, range)
         } else {
           Transforms.deselect(editor)
         }
@@ -424,6 +442,7 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
           decorations,
           node: editor,
           renderElement,
+          renderPlaceholder,
           renderLeaf,
           selection: editor.selection,
         })}
--- a/packages/slate-react/src/components/android/android-editable.tsx
+++ b/packages/slate-react/src/components/android/android-editable.tsx
@@ -6,16 +6,7 @@ import React, {
   useRef,
   useState,
 } from 'react'
-import {
-  Descendant,
-  Editor,
-  Element,
-  Node,
-  NodeEntry,
-  Range,
-  Text,
-  Transforms,
-} from 'slate'
+import { Descendant, Editor, Node, Range, Transforms } from 'slate'
 import throttle from 'lodash/throttle'
 import scrollIntoView from 'scroll-into-view-if-needed'
 
@@ -25,11 +16,7 @@ import { useSlate } from '../../hooks/use-slate'
 import { useIsomorphicLayoutEffect } from '../../hooks/use-isomorphic-layout-effect'
 import {
   DOMElement,
-  DOMNode,
-  DOMRange,
-  DOMStaticRange,
   getDefaultView,
-  isDOMNode,
   isPlainTextOnlyPaste,
 } from '../../utils/dom'
 import {
@@ -49,12 +36,9 @@ import useChildren from '../../hooks/use-children'
 import {
   defaultDecorate,
   hasEditableTarget,
-  hasTarget,
   isEventHandled,
   isTargetInsideVoid,
 } from '../default-editable'
-
-const HAS_BEFORE_INPUT_SUPPORT = true
 
 export const AndroidEditableNoError = (props: EditableProps): JSX.Element => {
   return (
--- a/packages/slate-react/src/components/android/android-editable.tsx
+++ b/packages/slate-react/src/components/android/android-editable.tsx
@@ -19,12 +19,6 @@ import {
 import throttle from 'lodash/throttle'
 import scrollIntoView from 'scroll-into-view-if-needed'
 
-import {
-  IS_CHROME_LEGACY,
-  IS_EDGE_LEGACY,
-  IS_FIREFOX,
-  IS_SAFARI,
-} from '../../utils/environment'
 import { ReactEditor } from '../..'
 import { ReadOnlyContext } from '../../hooks/use-read-only'
 import { useSlate } from '../../hooks/use-slate'
@@ -52,14 +46,15 @@ import { AndroidInputManager } from './android-input-manager'
 import { EditableProps } from '../editable'
 import { ErrorBoundary } from './ErrorBoundary'
 import useChildren from '../../hooks/use-children'
+import {
+  defaultDecorate,
+  hasEditableTarget,
+  hasTarget,
+  isEventHandled,
+  isTargetInsideVoid,
+} from '../default-editable'
 
-// COMPAT: Firefox/Edge Legacy don't support the `beforeinput` event
-// Chrome Legacy doesn't support `beforeinput` correctly
-const HAS_BEFORE_INPUT_SUPPORT = !(
-  IS_FIREFOX ||
-  IS_EDGE_LEGACY ||
-  IS_CHROME_LEGACY
-)
+const HAS_BEFORE_INPUT_SUPPORT = true
 
 export const AndroidEditableNoError = (props: EditableProps): JSX.Element => {
   return (
@@ -196,12 +191,6 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
       }
 
       setTimeout(() => {
-        // COMPAT: In Firefox, it's not enough to create a range, you also need
-        // to focus the contenteditable element too. (2016/11/16)
-        if (newDomRange && IS_FIREFOX) {
-          el.focus()
-        }
-
         state.isUpdatingSelection = false
       })
     } catch (err) {
@@ -334,24 +323,12 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
   return (
     <ReadOnlyContext.Provider value={readOnly}>
       <Component
-        // COMPAT: The Grammarly Chrome extension works by changing the DOM
-        // out from under `contenteditable` elements, which leads to weird
-        // behaviors so we have to disable it like editor. (2017/04/24)
-        data-gramm={false}
         key={contentKey}
         role={readOnly ? undefined : 'textbox'}
         {...attributes}
-        // COMPAT: Certain browsers don't support the `beforeinput` event, so we'd
-        // have to use hacks to make these replacement-based features work.
-        spellCheck={
-          !HAS_BEFORE_INPUT_SUPPORT ? undefined : attributes.spellCheck
-        }
-        autoCorrect={
-          !HAS_BEFORE_INPUT_SUPPORT ? undefined : attributes.autoCorrect
-        }
-        autoCapitalize={
-          !HAS_BEFORE_INPUT_SUPPORT ? undefined : attributes.autoCapitalize
-        }
+        spellCheck={attributes.spellCheck}
+        autoCorrect={attributes.autoCorrect}
+        autoCapitalize={attributes.autoCapitalize}
         data-slate-editor
         data-slate-node="value"
         contentEditable={readOnly ? undefined : true}
@@ -367,30 +344,6 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
           // Allow for passed-in styles to override anything.
           ...style,
         }}
-        onBeforeInput={useCallback(
-          (event: React.FormEvent<HTMLDivElement>) => {
-            // COMPAT: Certain browsers don't support the `beforeinput` event, so we
-            // fall back to React's leaky polyfill instead just for it. It
-            // only works for the `insertText` input type.
-            if (
-              !HAS_BEFORE_INPUT_SUPPORT &&
-              !readOnly &&
-              !isEventHandled(event, attributes.onBeforeInput) &&
-              hasEditableTarget(editor, event.target)
-            ) {
-              return
-            }
-          },
-          [readOnly]
-        )}
-        onBlur={useCallback((event: React.FocusEvent<HTMLDivElement>) => {}, [
-          readOnly,
-          attributes.onBlur,
-        ])}
-        onClick={useCallback((event: React.MouseEvent<HTMLDivElement>) => {}, [
-          readOnly,
-          attributes.onClick,
-        ])}
         onCompositionEnd={useCallback(
           (event: React.CompositionEvent<HTMLDivElement>) => {
             if (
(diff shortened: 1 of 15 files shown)


thesunny:
Happy to see this in a PR now @wleroux !

This is a reminder that even in earlier versions of Slate, Android support is not a 1:1 replacement for the existing event handlers. Namely, Android does not send a number of events that some plugins may expect to see and therefore, you cannot intercept them as you normally would. Over time, hopefully, with the community, we can continue to improve Android support where more and more of these cases can be handled naturally.

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,580 @@
...
+
+          Transforms.select(this.editor, range)
+          ReactEditor.focus(this.editor)
+
+          this.clearAction()
+          restoreDOM(this.editor)
+        })
+      }
+    }, 20)
+  }
```
clauderic, line 488:
```suggestion
```

I don't think any of this logic is needed. `onCompositionEnd` typically only fires when the keyboard is closed on Android, by which point we should have already flushed all mutations

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,580 @@
...
+        this.lastDiff !== undefined
+      ) {
+        debug('onSelect:applyDiff', this.lastDiff)
+        this.applyDiff()
+        Transforms.select(this.editor, range)
+
+        this.clearAction()
+        flushController(this.editor)
+        restoreDOM(this.editor)
+      }
```
clauderic, line 563: Why is this needed?

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,580 @@
...
+  /**
+   * Handle `onSelect` event
+   *
+   * Save the selection after a `requestAnimationFrame`
+   *
+   * - If we're not in the middle of flushing mutations
+   * - and cancel save if a mutation runs before the `requestAnimationFrame`
+   */
+
+  onSelect = () => {
```
clauderic, line 499: Why does Android require special `onSelect` handling, and does this logic not belong in `android-editable.tsx`?

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,580 @@
...
+}
+
+function normalizeDOMSelection(selection: Selection) {
+  return {
+    anchorNode: selection.anchorNode,
+    anchorOffset: selection.anchorOffset,
+    focusNode: selection.focusNode,
+    focusOffset: selection.focusOffset,
+  }
+}
```
clauderic, line 578: Seems like this is only used for debugging purposes

```suggestion
```

🔥

packages/slate-react/src/components/android/diff-text.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (offsets == null) return null
+  const insertText = sliceText(next, offsets)
+  const removeText = sliceText(prev, offsets)
+  return {
+    start: offsets.start,
+    end: prev.length - offsets.end,
+    insertText,
+    removeText,
+  }
+}
```
clauderic, line 104: How are you handling cases where the same letter is inserted or deleted multiple times in a row?

Take this example: 
```
hello|o
```

If another `o` is inserted at the selection's anchor in the example above, it should be inserted at the anchor, but using only text diffing, we detect that the character should be inserted after the second `o` instead, which is incorrect:

``` 
helloo[o]|
```

clauderic (commented):
Hey @wleroux, thanks for all your hard work on this 🙏 

Only had time for a quick first pass, I've left a number of questions and suggestions. Will give this a more in-depth review in the coming weeks.

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,580 @@
...
+    debug('resolveDOMNode:diff', this.lastDiff)
+  }
+
+  /**
+   * handle `onCompositionStart`
+   */
+
+  onCompositionStart = () => {
+    debug('onCompositionStart')
+  }
```
clauderic, line 423: How are you distinguishing between mutations that occur based on user input versus other mutations that might occur as a result of other actions such as `Transforms` being called when the user presses buttons in the toolbar of the rich text editor?

packages/slate-react/src/components/android/android-editable.tsx
```diff
@@ -0,0 +1,636 @@
...
+    readOnly = false,
+    renderElement,
+    renderLeaf,
+    style = {},
+    as: Component = 'div',
+    ...attributes
+  } = props
+  const editor = useSlate()
+  const ref = useRef<HTMLDivElement>(null)
+  const inputManager = useMemo(() => new AndroidInputManager(editor), [editor])
```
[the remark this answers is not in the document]
wleroux: Absolutely! As all the reconciliation happens outside of React's lifecycle events, it gets no benefit from hooks. Using hooks would adds a layer of complexity while not improving the performance nor maintainability of the work required.

I could add a new hook that merely manages the android input manager, but that extra structure does not make the code any clearer as the `android-editable` class is all about managing these dependendies.

packages/slate-react/src/components/android/android-input-manager.ts
wleroux: Mutations are not observed when react re-renders the value as we turn off the mutation observations before and reset the observations after the rendering.

packages/slate-react/src/components/android/diff-text.ts
wleroux: The general approach is to determine the difference in the text and apply the changes such that the final value is identical to what the user is seeing.

packages/slate-react/src/components/android/android-editable.tsx
```diff
@@ -0,0 +1,636 @@
...
+                event.preventDefault()
+                const range = ReactEditor.findEventRange(editor, event)
+                const data = event.dataTransfer
+                Transforms.select(editor, range)
+                ReactEditor.insertData(editor, data)
+              }
+            }
+          },
+          [readOnly, attributes.onDrop]
+        )}
```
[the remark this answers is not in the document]
wleroux: Suggested changes applied RE: removing code to support drag-and-drop.

packages/slate-react/src/components/android/android-editable.tsx
```diff
@@ -0,0 +1,636 @@
...
+              const el = ReactEditor.toDOMNode(editor, editor)
+              state.latestElement = window.document.activeElement
+
+              // COMPAT: If the editor has nested editable elements, the focus
+              // can go to them. In Firefox, this must be prevented because it
+              // results in issues with keyboard navigation. (2017/03/30)
+              if (IS_FIREFOX && event.target !== el) {
+                el.focus()
+                return
+              }
```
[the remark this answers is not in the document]
wleroux: Removed any code related to specific environment handling.

packages/slate-react/src/components/android/android-editable.tsx
```diff
@@ -0,0 +1,636 @@
...
+  event: EventType,
+  handler?: (event: EventType) => void
+) => {
+  if (!handler) {
+    return false
+  }
+
+  handler(event)
+  return event.isDefaultPrevented() || event.isPropagationStopped()
+}
```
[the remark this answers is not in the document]
wleroux: Exported these functions from `default-editables.tsx` and removed them from `android-editable.tsx`.

packages/slate-react/src/components/android/android-editable.tsx
```diff
@@ -0,0 +1,636 @@
...
+
+  const prevValue = useRef<Descendant[]>([])
+  if (prevValue.current !== editor.children) {
+    inputManager.onRender()
+    prevValue.current = editor.children
+  }
+
+  useLayoutEffect(() => {
+    inputManager.onDidUpdate()
+  })
```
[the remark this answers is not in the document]
wleroux: Addressed hook concerns in another thread.

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,580 @@
...
+
+function renderSync(editor: ReactEditor, fn: () => void) {
+  try {
+    fn()
+    flushController(editor)
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err)
+  }
+}
```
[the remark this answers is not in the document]
wleroux: The DOM reconciliation happens outside of the normal React Lifecycle and needs to be reconciled with the virtual DOM.

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,580 @@
...
+     * This happens only when using the virtual keyboard. Hitting enter on a
+     * hardware keyboard does not trigger this bug.
+     *
+     * The call to `focus` is required because when we switch examples then
+     * merge a block, we lose focus in Android 9 (possibly others).
+     */
+
+    window.requestAnimationFrame(() => {
+      renderSync(this.editor, () => {
+        this.applyDiff()
```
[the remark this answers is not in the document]
wleroux: Text changes are not instantly applied. If the block had any text changes prior to the merge, they need to be reconciled.

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,580 @@
...
+     * setting it and it is correct, but somewhere Android moves it to the right.
+     *
+     * This happens only when using the virtual keyboard. Hitting enter on a
+     * hardware keyboard does not trigger this bug.
+     *
+     * The call to `focus` is required because when we switch examples then
+     * merge a block, we lose focus in Android 9 (possibly others).
+     */
+
+    window.requestAnimationFrame(() => {
```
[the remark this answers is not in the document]
wleroux: An extended goal is to provide as much broad support for Android as possible, even if the commitment is for Android 10.

packages/slate-react/src/components/android/android-input-manager.ts
wleroux: Text differences are applied at specific events to ensure they don't conflict with the default editor. This is one such event were we reconcile the differences.

packages/slate-react/src/components/android/android-input-manager.ts
wleroux: Will inline it.

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,580 @@
...
+        mutation => mutation.removedNodes.length > 0
+      ).length
+      const addedNodes = mutations.filter(
+        mutation => mutation.addedNodes.length > 0
+      ).length
+
+      if (removedNodes > addedNodes) {
+        this.mergeBlock()
+      } else if (addedNodes > removedNodes) {
+        this.splitBlock()
```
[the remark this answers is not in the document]
wleroux: Can you provide a test case where this code is insufficient?

packages/slate-react/src/components/android/android-input-manager.ts
wleroux: Discussed reconcilation requirements in other threads.

packages/slate-react/src/components/android/android-input-manager.ts
wleroux: Discussed reconciliation requirements in other threads.

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,580 @@
...
+     *
+     * The call to `focus` is required because when we switch examples then
+     * merge a block, we lose focus in Android 9 (possibly others).
+     */
+
+    window.requestAnimationFrame(() => {
+      renderSync(this.editor, () => {
+        this.applyDiff()
+
+        Transforms.select(this.editor, this.lastRange!)
```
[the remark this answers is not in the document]
wleroux: Discussed reconciliation requirements in other threads.

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,580 @@
...
+
+    debug('connect:run')
+
+    this.observer.disconnect()
+    this.observer.observe(rootEl, {
+      childList: true,
+      characterData: true,
+      attributes: true,
+      subtree: true,
+      characterDataOldValue: true,
```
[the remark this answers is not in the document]
wleroux: Removed attributes from observed mutations.

packages/slate-react/src/components/android/android-input-manager.ts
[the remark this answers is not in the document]
clauderic: Certainly, will do next week

packages/slate-react/src/components/android/android-input-manager.ts
clauderic: Again, I don't think this has been discussed or addressed. I'm fairly certain you can delete all of this code.

If you can't, it means the reconciliation approach needs to be fixed elsewhere.

Considering `onCompositionEnd` only fires when the soft keyboard is dismissed on Android, your code needs to handle reconciliation regardless of whether this event is fired or not, otherwise it would mean that you're only reconciling properly when the keyboard is dismissed, which may not always be sufficient.

packages/slate-react/src/components/android/android-input-manager.ts
clauderic: Again, I don't think this has really been answered. What's different about selection handling on Android that requires us having this code?

clauderic:
@wleroux can you please un-resolve the discussions with the exception of the ones where you have removed/moved code? I don't think it's fair to viewers of this PR to mark them as resolved.

packages/slate-react/src/components/android/diff-text.ts
clauderic: Yes but the text is inserted at the wrong position, so the cursor ends up in the wrong position. How is that being addressed?

packages/slate-react/src/components/android/android-editable.tsx
```diff
@@ -0,0 +1,449 @@
...
+import useChildren from '../../hooks/use-children'
+import {
+  defaultDecorate,
+  hasEditableTarget,
+  hasTarget,
+  isEventHandled,
+  isTargetInsideVoid,
+} from '../default-editable'
+
+const HAS_BEFORE_INPUT_SUPPORT = true
```
thesunny, line 57: This looks like it's not necessary in the Android specific editable.

packages/slate-react/src/components/android/android-editable.tsx
```diff
@@ -0,0 +1,449 @@
...
+    () => ({
+      isComposing: false,
+      isUpdatingSelection: false,
+      latestElement: null as DOMElement | null,
+    }),
+    []
+  )
+
+  // Update element-related weak maps with the DOM element ref.
+  useIsomorphicLayoutEffect(() => {
```
thesunny, line 102: Hi @wleroux , there are a number of these `useIsomorphicLayoutEffect` methods that are verbatim from `editable.tsx`.

I wonder if it might be better to take some of these, especially the longer and more complex methods, and turn them into a custom hook that is exported from `editable.tsx`. Maybe this particular one isn't a great example (because it's not that long), but as an example, it could be something like: `useUpdateWeakMaps(editor, window, ref)`.

By having them in a separate `AndroidEditable` still gives us the option to "roll back" just the AndroidEditable version if the default version causes problems specific to Android.

I won't add comments to each one but just as a general idea.

packages/slate-react/src/components/android/android-input-manager.ts
```diff
@@ -0,0 +1,573 @@
...
+
+const debug = (...message: any[]) => {}
+
+/**
+ * Unicode String for a ZERO_WIDTH_SPACE
+ *
+ * @type {String}
+ */
+
+const ZERO_WIDTH_SPACE = String.fromCharCode(65279)
```
thesunny, line 18: I think the ZERO\_WIDTH\_SPACE hack is no longer used with the 0.50+ version of Slate and a search for "65279" doesn't appear to show any hits (at least in GitHub's search).

Maybe check to see if it's required and if not remove related code.

packages/slate-react/src/components/android/diff-text.ts
```diff
@@ -0,0 +1,111 @@
+/**
+ * Returns the number of characters that are the same at the beginning of the
+ * String.
+ *
+ * @param {String} prev  the previous text
+ * @param {String} next  the next text
+ * @returns {Number | null} the offset of the start of the difference; null if there is no difference
```
thesunny, line 7: Maybe remove old pre-typescript commenting style. Only mentioned once but should apply to all methods.

thesunny (commented):
Hey @wleroux , here's a few thoughts on the code.

packages/slate-react/src/components/android/android-editable.tsx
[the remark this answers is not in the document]
wleroux: I believe that the concerns raised in this thread are covered in: https://github.com/ianstormtaylor/slate/pull/4200#discussion_r613709845

packages/slate-react/src/components/android/android-input-manager.ts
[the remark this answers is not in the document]
wleroux: Test case for this is pretty trivial:

> You insert text, then you merge the paragraph after by pressing delete.

I believe that your concern stems from misunderstanding the overall strategy for resolving mutations. We do not convert the mutations to slate operations directly. For DOM structure changes (split/merge), mutations are applied immediately. For text changes (insert,remove), we keep track of the text difference and only apply the text difference when a dom structure changes or window selection changes.

packages/slate-react/src/components/android/android-input-manager.ts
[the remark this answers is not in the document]
wleroux: This was discussed and further elaborated in:

https://github.com/ianstormtaylor/slate/pull/4200#discussion_r618950248

packages/slate-react/src/components/android/android-input-manager.ts
wleroux: This has been discussed and further elaborated in:

https://github.com/ianstormtaylor/slate/pull/4200#discussion_r618950248

packages/slate-react/src/components/android/android-input-manager.ts
wleroux: This has been discussed and further elaborated in:

https://github.com/ianstormtaylor/slate/pull/4200#discussion_r618950248

packages/slate-react/src/components/android/diff-text.ts
wleroux: This has been discussed and further elaborated in:

https://github.com/ianstormtaylor/slate/pull/4200#discussion_r618950248

packages/slate-react/src/components/android/diff-text.ts
wleroux: When I attempted to remove the types from the documentation, the eslint started failing:

>   1:1  error  Missing JSDoc return type                   valid-jsdoc
>   5:4  error  Missing JSDoc parameter type for 'prev'     valid-jsdoc
>   6:4  error  Missing JSDoc parameter type for 'next'     valid-jsdoc

It appears that the parameter type is still required, even though it is redundant in typescript.

thesunny: That's unexpected since there are other files in the repo that don't have them and they don't raise eslint errors. I wonder if it's an all or nothing deal. If there are some JSDoc comments, then the entire file has to be? It does appear that jsdoc does appear in `.eslintrc`. I feel like we can just remove that from the `.eslintrc`. I just did a search through the Slate repo and it is used in exactly one method call which probably was left there unintentially `packages/slate-react/src/utils/lines.ts`

packages/slate-react/src/components/android/android-editable.tsx
```diff
@@ -0,0 +1,636 @@
...
+      <AndroidEditable {...props} />
+    </ErrorBoundary>
+  )
+}
+
+/**
+ * Editable.
+ */
+
+export const AndroidEditable = (props: EditableProps): JSX.Element => {
```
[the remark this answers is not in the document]
wleroux: Ah, I thought you just wanted to leave a note of a previous concern that had already been voiced and discussed.

The alternative design you are proposing was already considered when we discussed the need to integrate the android keyboard support into the core repository instead of the previous plugin design. In that discussion we got approved to embed the android keyboard support as long as we fork the `Editable` component. The approach you're proposing has merits but we will not be pursuing it.

While your concern is still relevant, it was already acknowledged and considered. The particular aspect of the design is not open for discussion.

packages/slate-react/src/components/android/android-editable.tsx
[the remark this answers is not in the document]
wleroux: > @wleroux can you please un-resolve the discussions with the exception of the ones where you have removed/moved code? I don't think it's fair to viewers of this PR to mark them as resolved.

@clauderic Sorry about that, it was my assumption that anyone can unresolve an issue / posting new comments would unresolve that conversation.

wleroux:
> @wleroux can you please un-resolve the discussions with the exception of the ones where you have removed/moved code? I don't think it's fair to viewers of this PR to mark them as resolved.

@clauderic Sorry about that, it was my assumption that anyone can unresolve an issue / posting new comments would unresolve that conversation.

packages/slate-react/src/components/android/diff-text.ts
wleroux: Removed eslint rule and removed jsdoc comments

wleroux: Removed eslint rule and removed jsdoc comments

packages/slate-react/src/components/android/android-input-manager.ts
wleroux: Removed ZERO_WIDTH_SPACE and related code

packages/slate-react/src/components/android/android-editable.tsx
wleroux: Removed dead code.

packages/slate-react/src/components/android/android-editable.tsx
wleroux: `Editable.tsx` is a very frequently changed file and all the effort into making this portion re-usable will have to be re-done every time `Editable.tsx` changes in the main branch.

I would like to deal with this after the pull request is merged. By delaying it until after the merge, the pull request would be a trivial refactoring. Alternatively, we can delay this change until right before this pull request merge. Thoughts?

clauderic:
I've left a lot of disconnected pieces of feedback on this PR, so I'll try to summarize.

### Level of abstraction

Conceptually, I disagree with the overall approach taken in this PR. It's the wrong level of abstraction in my view. 

The `<Editable>` component manages all matters related to make Slate compatible with React. Forking it in its entirety only to support Android isn't the right long term decision. In the short term, perhaps that's an acceptable change, but I do worry that it will bee a decision that has a lot of friction to revert.

Long term, the right abstraction to model this is to abstract all logic related to user input reconciliation out of `Editable` and
into a hook, something like `useInputReconciliation`.

Input reconciliation can further be broken down into different categories of user input that needs to be reconciled with Slate's internal value:

- Selection reconciliation
- Text composition reconciliation
- Drag and drop
- Copy / cut / paste

All four of these types of user input can be further broken down into hooks (`useSelectionReconciliation`, etc.).

The text composition reconciliation layer can further be broken down into different reconciliation strategies that can be conditionally enabled depending on the browser capabilities:

- `beforeinput`-based reconciliation (`useBeforeInput`)
- `MutationObserver`-based reconciliation (`useMutationObserver`)

### Mutation reconciliation

> I believe that your concern stems from misunderstanding the overall strategy for resolving mutations. We do not convert the mutations to slate operations directly. For DOM structure changes (split/merge), mutations are applied immediately. For text changes (insert,remove), we keep track of the text difference and only apply the text difference when a dom structure changes or window selection changes.

I understand the current overall strategy for resolving mutations, but in my view the current strategy is incorrect.

Mutations resulting from user input need to be reconciled as soon as they are flushed, not when the selection changes or when a DOM structure changes.

Input reconciliation layers should be as closely aligned with one another as possible to avoid behavioural differences when authoring functionality on top of Slate.

https://user-images.githubusercontent.com/1416436/115888593-177ee400-a421-11eb-875d-6330facbfd0b.mp4

_Samsung Galaxy S20, Android 10_

The `value` should be updated anytime a new character data mutation is detected to match the behaviour of the `beforeinput` reconciliation strategy. Doing so will avoid behavioural differences that could otherwise break consumer-built functionality on top of Slate.

In my experience this is possible to do, and doing so would greatly simplify the overall complexity of `AndroidInputManager` because you'd only need to process a single set of flushed mutations at a time as opposed to the current approach of delayed batch updates that requires you to check for text diff in selection changes and DOM structure changes.

### Various bugs

Samsung Galaxy S20, Android 10

https://user-images.githubusercontent.com/1416436/115892154-e30d2700-a424-11eb-81e1-b35dbf2310a1.mp4

https://user-images.githubusercontent.com/1416436/115892131-ddafdc80-a424-11eb-872d-11e22c15b774.mp4

https://user-images.githubusercontent.com/1416436/115892189-eacccb80-a424-11eb-8040-c688e59e4785.mp4

https://user-images.githubusercontent.com/1416436/115892210-f0c2ac80-a424-11eb-9edd-eae2408b1347.mp4

packages/slate-react/src/components/android/android-editable.tsx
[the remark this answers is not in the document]
clauderic: Assuming you move forward with having swappable `Editable` components, shared methods should be exported from a different file than `default-editable.tsx` to help with tree shaking (dead code elimination).

packages/slate-react/src/components/android/android-editable.tsx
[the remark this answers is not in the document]
thesunny: Hi @clauderic, I've opened up a separate issue to discuss a hook based Android implementation. I'm going to "resolve" these conversations, because we aren't going to be implementing your proposal in this PR, but it is not meant to mean that the issue of which approach is better is resolved. I've re-posted your summary in the issue here #4223 as well as adding some comments about why we won't be implementing that idea in this particular PR.

packages/slate-react/src/components/android/android-editable.tsx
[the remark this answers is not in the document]
thesunny: Hi @clauderic, I've opened up a separate issue to discuss a hook based Android implementation. I'm going to "resolve" these conversations, because we aren't going to be implementing your proposal in this PR, but it is not meant to mean that the issue of which approach is better is resolved. I've re-posted your summary in the issue here #4223 as well as adding some comments about why we won't be implementing that idea in this particular PR.

thesunny:
Hi @clauderic, I've opened up a separate issue to discuss a hook based Android implementation. I've "resolve"ed the related conversations above, because we aren't going to be implementing your proposal in this PR, but it is not meant to mean that the issue of which approach is better is resolved. I've re-posted your summary in the issue here #4223 as well as adding some comments about why we won't be implementing that idea in this particular PR.

packages/slate-react/src/components/android/android-editable.tsx
thesunny: Yep, that's fine to defer IMO as long as we do it.

clauderic:
Updated my comment to only keep the comments on the mutation reconciliation which are relevant to this PR, we can discuss level of abstraction in the other issue 👍

thesunny:
Hi @wleroux, sorry, I didn't think of this earlier but I think one thing we should change is that we should not rename the current `editable.tsx` to `default-editable.tsx`. The reason is I think any PR that targets `editable.tsx` will become broken without an easy path to merge. Instead, I recommend we do something like introduce an `editable-triage.tsx` (or something like that) and have it defer to the original `editable.tsx` or `android-editable.tsx`. Any PRs that target `editable.tsx` should work as before.

Maybe this is the reason why you're having to rebase often as well? I can see the rename causing problems in a lot of PRs unless we make this change.

wleroux:
> Updated my comment to only keep the comments related to the mutation reconciliation layer, which are relevant to this PR, we can discuss level of abstraction in the other issue 👍

The reason that we delay the resolution of text mutations is that they disrupt the keyboard manager on the device. There were two issues around this area that I was working on at the time: 1st was keyboard would be closed when the dom structure changes and the 2nd would be some input issues with auto-suggestions (e.g.: 't' auto-completed to 'test' would end with the final result of 'ttest'). By delaying the mutation resolution until specific changes (e.g.: dom structure changes or selection changes), these issue were avoided. Otherwise, I agree with the premise: We want to minimize the time between mutation resolution.

wleroux:
> Hi @wleroux, sorry, I didn't think of this earlier but I think one thing we should change is that we should not rename the current `editable.tsx` to `default-editable.tsx`. The reason is I think any PR that targets `editable.tsx` will become broken without an easy path to merge. Instead, I recommend we do something like introduce an `editable-triage.tsx` (or something like that) and have it defer to the original `editable.tsx` or `android-editable.tsx`. Any PRs that target `editable.tsx` should work as before.
> 
> Maybe this is the reason why you're having to rebase often as well? I can see the rename causing problems in a lot of PRs unless we make this change.

Agreed and implemented. To avoid issues down-the-line, I've rebased the code such that the rename never happened; this should avoid any issues with other merge requests... However to avoid issues with `editable.tsx`, the control for selecting the android version is now in `packages\slate-react\src\index.ts`.

PR #4203, opened 2021-04-15
Allow unhang when the range start offset is not zero

TheSpyder:
**Description**

The `unhangRange` implementation produces inconsistent user actions. It has an early exit if the start of the selection is not at the start of a node, but that seems like an arbitrary decision.

The result from a user perspective is that if they get into a range hang situation, whether the following paragraph is included in their next action is seemingly random. 

**Issue**
Fixes: #4148 (I've also incorporated the inline void change)

**Example**

* `<p>[ab</p><p>]c</p>` becomes `<p>[ab]</p><p>c</p>` ✅ 
* `<p>a<bold>[b</bold></p><p>]c</p>` becomes `<p>a<bold>[b]</bold></p><p>c</p>` ✅ 
* `<p>a[b</p><p>]c</p>` is not adjusted ❌ 

**Context**
I've added a bunch of tests for the existing behaviour, the test that fails without my fix is `unhang-text-middle.tsx`.

I'm not 100% sure if this is a good idea, because the selection delete tests depend on `<p>ab[</p><p>]c</p>` which my change now unhangs to `<p>ab[]</p><p>c</p>`. Adding extra text to the tests allows them to pass but I'm concerned it might have changed the intention of the tests.

**Checks**

### DIFF
--- a/.changeset/good-hairs-allow.md
+++ b/.changeset/good-hairs-allow.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Selection across the end of a paragraph now more consistently excludes the following paragraph from modifications
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -1648,16 +1648,16 @@ export const Editor: EditorInterface = {
 
     for (const [node, path] of Editor.nodes(editor, {
       at: before,
-      match: Text.isText,
+      match: n => Text.isText(n) || (editor.isInline(n) && editor.isVoid(n)),
       reverse: true,
       voids,
     })) {
       if (skip) {
         skip = false
         continue
-      }
-
-      if (node.text !== '' || Path.isBefore(path, blockPath)) {
+      } else if (!Text.isText(node)) {
+        break
+      } else if (node.text !== '' || Path.isBefore(path, blockPath)) {
         end = { path, offset: node.text.length }
         break
       }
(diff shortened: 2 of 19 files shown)


ianstormtaylor:
I'm not really sure this is what we want, because it seems like it makes it not possible to use the cursor to move to the next block and then press `delete` to merge the blocks. 

For example I believe it would make these two behaviors not possible:

TheSpyder:
Delete doesn't use `unhangRange`, at least not in the same way. It calls `Transforms.delete` which passes a _point_ to `Transorms.mergeNodes`. Try your examples with bold; I don't think the result is what users would expect if they accidentally create a selection like this.

I thought that was the point of unhang - to correct scenarios where the user might accidentally create a hanging range. Creating a hanging selection intentionally seems unlikely, even if as experienced editor developers we know how to do it :)

TheSpyder:
Since my aim was to fix `Transforms.setNodes` in hang edge cases I'll see if I can make an adjustment that only impacts it. I hate adding options, but an option to adjust whether whole-node selection is required to perform an unhang might be the best way forward.

TheSpyder:
I was halfway through a fix and forgot about the branch. This push only updates the branch to latest main, it's not ready for re-review yet.

dylans:
Still relevant or should we close this out?

TheSpyder:
We’re using this code in production and are quite happy with it, I just haven’t been able to schedule time to finish up the PR and make it part of slate core.

I would still like to, I think the usability improvements are worth it.

TheSpyder:
Not likely to happen now. Sorry for leaving it open so long.

Issue #4196, opened 2021-04-13
Integration tests for slate


juliankrispel:
**Problem**

Slate doesn't currently have integration tests. Because of this:

- Accepting changes is high risk.
- Reviewing changes requires thorough manual testing.

**Solution**

Introduce automated integration tests (ideally via gh-actions to be consistent) using cypress

### Cypress
- Great developer ergonomics and debuggability (GUI for stepping through tests, automatic screenshots and video recording)
- Designed to make it easy to avoid flake
- Supports chromium browsers (Edge, Chrome) as well as Firefox

**Alternatives**

### Puppeteer
- Less opinionated than cypress but more configurable
- Supports chromium browsers only (Edge, Firefox, Chrome)

### Webdriver aka selenium
- Has support for all modern browsers
- Third-party support for mobile browsers (look at appium for example)
- Poor developer ergonomics

**Context**

I believe adding good integration tests (for slate-react) could significantly increase confidence and reduce the maintenance burden.

Although there are many options to choose from for integration testing I strongly believe that optimizing for developer ergonomics is the key to success here. Let's make it as easy as possible to increase coverage and debug failing tests.

juliankrispel:
Full disclosure: I've been using both puppeteer and cypress to test slate-editors on a daily basis at work and would like to use what I've learned to help slate.js. I think we could quite easily catch bugs such as #4170 for example by smoke testing the examples.

To keep it small as a first pr I was thinking one could install cypress (or another framework) and write a simple test.

wdyt @ianstormtaylor - would love to get your input

ianstormtaylor:
@juliankrispel that sounds amazing, and would be immensely appreciated. I think the first PR being a single test sounds good too, will give an idea for what the setup looks like. Maybe it can just be writing a single character in an editor? Not sure—trust your decision on whatever it is!

TheSpyder:
> * Supports chromium browsers (Edge, Firefox, Chrome)

nitpick: firefox is not a chromium browser 😂 

Any automated testing is better than no testing, but browser-specific issues will still need to be manually tested.

juliankrispel:
> nitpick: firefox is not a chromium browser 😂

You're right, corrected. However - [Firefox is supported by cypress](https://www.cypress.io/blog/2020/02/06/introducing-firefox-and-edge-support-in-cypress-4-0/) hence browser specific issues could indeed be tested in cypress.

PR #4206, opened 2021-04-15
4196 integration tests with cypress

juliankrispel:
**Description**
- [x] Adds boilerplate and configuration for integration tests via cypress.
- [x] Runs integration tests on ci

🤔  It's tempting to add a bunch of tests from the start but I'd like to keep this pr small so it's frictionless to review 

**Issue**
Fixes: #4196 

**Example**
image

image

**Checks**


docs/general/contributing.md
```diff
@@ -87,6 +87,10 @@ yarn lint
 
 This will catch TypeScript, Prettier, and Eslint errors.
 
+## Running integration tests
+
+To run integrations with [cypress](https://github.com/cypress-io/cypress), first run `yarn start` to run the examples website, then run `yarn cypress:open` to open the cypress GUI.
```
timbuckley, line 92:
```suggestion
To run integrations with [cypress](https://github.com/cypress-io/cypress), first run `yarn start` to run the examples website, then run `yarn cypress:open` in a separate session to open the cypress GUI.
```

.gitignore
```diff
@@ -10,3 +10,8 @@ node_modules/
 packages/*/yarn.lock
 site/out/
 tmp/
+cypress/screenshots
+cypress/videos
+cypress/fixtures
+cypress/plugins
+.DS_Store
```
timbuckley, line 17:
```suggestion
.DS_Store

```

.github/workflows/ci.yml
```diff
@@ -25,6 +26,6 @@ jobs:
           node-version: 12
 
       - name: Run ${{ matrix.command }}
-        run: yarn && yarn build:rollup && yarn ${{ matrix.command }}
+        run: yarn && yarn build && yarn ${{ matrix.command }}
         env:
           CI: true
```
timbuckley, line 31: Bring `env: CI: true` up to the root of the `ci.yml` please. Makes this step a tiny bit cleaner.

cypress.json
```diff
@@ -0,0 +1,7 @@
+{
+  "projectId": "slate-react",
```
timbuckley, line 2: Is slate react or slate the appropriate name here?
```suggestion
  "projectId": "slate",
```

cypress/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "moduleResolution": "node" /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */,
+    "types": ["cypress", "@testing-library/cypress"],
+    "allowSyntheticDefaultImports": true /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */,
+    "esModuleInterop": true /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */,
+    "resolveJsonModule": true,
+    "skipLibCheck": true /* Skip type checking of declaration files. */,
+    "isolatedModules": false,
+    "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */
+  }
+}
```
timbuckley, line 27: Can we define these options in the root `tsconfig.json`, and then just `extend` it here?
For example, in my project's `cypress/`, we have:
```json
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "noEmit": false,
    "esModuleInterop": true
  },
  "include": ["../global.d.ts", "../node_modules/cypress", "../node_modules/cypress-xpath", "*/*.ts"]
}

```

timbuckley (approved):
Awesome, I love Cypress, can't wait til we use it here!

cypress/tsconfig.json
juliankrispel: That's what I wanted to do. But currently there seem to be separate tsconfigs for every package. Just trying to be consistent.

cypress/support/commands.ts
```diff
@@ -0,0 +1,15 @@
...
+
+declare global {
+  namespace Cypress {
+    interface Chainable<Subject> {
+      goTo(uri: string): void
+    }
+  }
+}
+
+Cypress.Commands.add('goTo', (uri: string) => {
```
beorn, line 11: Instead of creating a new `goTo` command, I think you may be able to just set Cypress' `baseUrl` via an environment variable and then just `cy.visit('/examples/richtest')`, as per https://docs.cypress.io/guides/guides/environment-variables#Setting

cypress.json
```diff
@@ -0,0 +1,7 @@
+{
+  "projectId": "slate",
+  "env": {
+    "SITE_URL": "localhost:3000"
```
beorn, line 4: Instead of defining a `$SITE_URL` and a `goTo` command that works on that, perhaps it would be possible to set Cypress' `baseUrl` via an environment variable and just use `cy.visit('/example/app')` - just prepend `CYPRESS_BASE_URL=https://localhost:3000` to the cypress app command in `package.json` or elsewhere, or add it to where environment variables are defined.  See https://docs.cypress.io/guides/guides/environment-variables#Setting

beorn (changes requested):

cypress/support/commands.ts
juliankrispel: great suggestion thank you! Addressed

Issue #3808, opened 2020-07-31
Transform.setNodes improvement and bug


BrentFarese:
#### Do you want to request a _feature_ or report a _bug_?

Both.

#### What's the current behavior?

Transforms.setNodes currently takes `Partial<Node>` as its second parameter for `props`.  This means you can only override the props on the nodes you are setting and, for example, you cannot base the new properties on the nodes you are setting off of the already-existing properties on those nodes.

Additionally, Transforms.setNodes has a `split` option, but the `split` doesn't occur in a couple edge cases that should likely be considered a bug.

The split bug occurs due to [these lines in Transforms.setNodes](https://github.com/ianstormtaylor/slate/blob/a180cc53000417f9055b7b87ce793216759870ee/packages/slate/src/transforms/node.ts#L471-L492).  Basically, if you look at Transforms.splitNodes, it will NOT split the relevant node if the split point is "at an edge" of the node.  This makes sense for single splits, but not for an expanded range.  For an expanded range, you always want to split if the start point is not at the start of its relevant node, and if the end point is not at the end of its relevant node.  If the start point of the expanded range is at the start of its relevant node, no split is needed since you already capture the full node anyways.  Same for if the end point is at the end of its relevant node.

Transforms.setNodes will not split an expanded range if the start/end points are at ANY edge of their respective nodes, which is not correct and results in setting props on nodes that was not intended.

Slate: 0.58.4
Browser: Chrome
OS: Mac

#### What's the expected behavior?

Transforms.setNodes `props` parameter should accept a set of props as `Partial<Node>` OR a function that takes the existing props for the relevant node and returns a new set of props to attach to the node.

Also, the splitting bug should be fixed.

adisen:
Hello I would like to take a look at solving this issue, can you assign it to me @BrentFarese @timbuckley

adisen:
Is it possible I get a short clip showing the reproduction of this feature?

iSplasher:
![recording](https://user-images.githubusercontent.com/9113348/110389671-21ad7480-8065-11eb-865e-629e9097bef6.gif)

Maybe this will work?

PR #4207, opened 2021-04-15
Added functionality for setNode to take a function as props argument

adisen:
**Description**
This PR allows Transforms.setNodes to take either a Partial as it second argument to props (which we have before) which overrides the props on the node or a function that takes the existing props for the relevant node and returns a new set of props to attach to the node.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/3808

**Checks**

### DIFF
--- a/packages/slate/src/interfaces/node.ts
+++ b/packages/slate/src/interfaces/node.ts
@@ -635,3 +635,7 @@ export type NodeProps =
   | Omit<Editor, 'children'>
   | Omit<Element, 'children'>
   | Omit<Text, 'text'>
+
+// export type NodeProps<T extends Node> = T extends Element
+//   ? Omit<T, 'children'>
+//   : Omit<T, 'text'>
--- a/packages/slate/src/interfaces/node.ts
+++ b/packages/slate/src/interfaces/node.ts
@@ -631,11 +631,6 @@ export type NodeEntry<T extends Node = Node> = [T, Path]
 /**
  * Convenience type for returning the props of a node.
  */
-export type NodeProps =
-  | Omit<Editor, 'children'>
-  | Omit<Element, 'children'>
-  | Omit<Text, 'text'>
-
-// export type NodeProps<T extends Node> = T extends Element
-//   ? Omit<T, 'children'>
-//   : Omit<T, 'text'>
+export type NodeProps<T extends Node = Node> = T extends Element
+  ? Omit<T, 'children'>
+  : Omit<T, 'text'>
--- a/packages/slate/src/transforms/node.ts
+++ b/packages/slate/src/transforms/node.ts
@@ -12,6 +12,7 @@ import {
   Ancestor,
 } from '..'
 import { NodeMatch } from '../interfaces/editor'
+import { NodeProps } from '../interfaces/node'
 
 export interface NodeTransforms {
   insertNodes: <T extends Node>(
@@ -67,7 +68,7 @@ export interface NodeTransforms {
   ) => void
   setNodes: <T extends Node>(
     editor: Editor,
-    props: Partial<Node>,
+    props: Partial<T> | ((existingProps: NodeProps) => Partial<T>),
     options?: {
       at?: Location
       match?: NodeMatch<T>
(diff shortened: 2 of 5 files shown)


adisen:
@BrentFarese check here. The NodeProps type you asked me to adjust didn't work.

TheSpyder:
This doesn't look like it fixes #3808 completely, there's a secondary split issue mentioned?

..... oh, it seems the split was fixed in #4168 🤔 so maybe that's ok

ianstormtaylor:
Hey @adisen there's an issue with the TypeScript linting here. Are you able to fix it? If not, can you make your branch editable so that I can make changes to the PR? Thanks!

TheSpyder:
It's the change from `Partial<Node>` to `Partial<T extends Node>`. I've found that when `T` is defined as something like `ImageElement | LinkElement | VideoElement`, TypeScript only checks the first of the `T` definitions when it's used inside a wider multi-type definition such as `Partial<T> | ((existingProps: NodeProps) => Partial<T>)`.

Hence the error that 'link' is not assignable to 'image'.

Switching `setNodes` to `Partial<Node> | ((existingProps: NodeProps) => Partial<Node>)` should fix it.

BrentFarese:
> It's the change from `Partial<Node>` to `Partial<T extends Node>`. I've found that when `T` is defined as something like `ImageElement | LinkElement | VideoElement`, TypeScript only checks the first of the `T` definitions when it's used inside a wider multi-type definition such as `Partial<T> | ((existingProps: NodeProps) => Partial<T>)`.
> 
> Hence the error that 'link' is not assignable to 'image'.
> 
> Switching `setNodes` to `Partial<Node> | ((existingProps: NodeProps) => Partial<Node>)` should fix it.

If there is any way to preserve the generic type, that might be preferred.  What we want is for the callback to `setNodes` to be aware of the specific `NodeProps` available to it.  Makes for a much better TS experience as now I can know what exactly the shape of `existingProps` is.  Maybe there is some way to solve the type issue w/o removing the generic.

TheSpyder:
Perhaps it could be `Partial<Node> | ((existingProps: NodeProps) => Partial<T>)` but I think it would just lead to the same issue, only for people who use the function.

Specifying a concrete `T` in the calling type (change `const newProperties: Partial<SlateElement>` to `const newProperties: Partial<LinkElement>` in `embeds.tsx` line 73) also resolves this but that's a breaking change.

ianstormtaylor:
I think going without the generic for now is best, because I found that there were some weird gotchas in converting to generics in my other PR. Especially in cases like this where the generic was trying to be used for `match` as well.

absolutejam:
Is the scope of this issue limited to still returning the same node type - ie. `'T => 'T` or would it facilitate changing node type?

I'm just coming back to a side-project that uses Slate and I've written an F# wrapper for (most of) the Slate API, but one of the things that causes me no end of issues is that `setNodes` changes the shape of the node instead of straight up replacing the node. 

Ideally - for use in TypeScript or F# - I'd want to completely transform from `ParagraphNode` to `TitleNode`, even if that means creating a new object, but I don't know if this would play havoc with how Slate manages its internal state.

TheSpyder:
> Is the scope of this issue limited to still returning the same node type - ie. `(T) => T` - or would it facilitate changing node type?

`setNodes` can change [anything except the `children` or `text` properties](https://github.com/ianstormtaylor/slate/blob/95f402c59414331b2eeca9a19bd2c73c0ab6cd6c/packages/slate/src/transforms/node.ts#L632-L642). So changing between block and inline types is allowed but it can't do to or from text node.

> I'm just coming back to a side-project that uses Slate and I've written an F# wrapper for (most of) the Slate API

Oh interesting! My team have created bindings in ReScript. Is yours open source? Might be fun to compare notes 🤔 


> one of the things that causes me no end of issues is that `setNodes` changes the shape of the node instead of straight up replacing the node.
> 
> I find myself often having to mutate the object and manually juggle types, but realistically - for use in TypeScript or F# - I'd want to completely transform from `ParagraphNode` to `TitleNode`, even if that means creating a new object (ensuring there are no unaccounted-for properties), but I don't know if this would play havoc with how Slate manages its internal state.

`setNodes` is designed for modifying attributes. Technically it creates a new object to do this, since the Slate model is immutable, but the apparent effect is that it's mutating the node.


> EDIT: Or maybe I'm just handling this wrong and should be using another pattern (`unwrap` and `wrap`?)

Possibly. Wrap and unwrap will work, it's just a little more effort.

dylans:
@adisen are you still interested in this PR? If so, it needs a rebase and there's some feedback to address. Thanks!

dylans:
Closing due to inactivity. Feel free to reopen or re-raise if there's interest.

PR #4211, opened 2021-04-19
Collapse expanded selection before move word hotkey

clauderic:
**Description**
This PR fixes the `moveWordBackward` / `moveWordForward` hotkey handling of `slate-react`. 

**Issue**
I could not find any existing related open issues.

The  `moveWordBackward` / `moveWordForward` hotkeys need to collapse expanded selections before moving the selection by word. The current behaviour is not aligned with how selection is moved in unmanaged `contenteditable` elements nor is it aligned with how other text editors work.

**Example**

_Before:_

https://user-images.githubusercontent.com/1416436/115253692-9eb71980-a0fa-11eb-83a8-e20eca5f4a06.mp4

_After:_

https://user-images.githubusercontent.com/1416436/116178116-fa793800-a6e2-11eb-9cbc-c6520d25407d.mp4

Slate above, native Textarea element below for reference

**Checks**

### DIFF
--- a/.changeset/move-word-collapse.md
+++ b/.changeset/move-word-collapse.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Collapse expanded selection before handling `moveWordBackward` (`alt + left`) and `moveWordForward` (`alt + right`) hotkeys.
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -912,12 +912,22 @@ export const Editable = (props: EditableProps) => {
 
                 if (Hotkeys.isMoveWordBackward(nativeEvent)) {
                   event.preventDefault()
+
+                  if (selection && Range.isExpanded(selection)) {
+                    Transforms.collapse(editor, { edge: 'focus' })
+                  }
+
                   Transforms.move(editor, { unit: 'word', reverse: !isRTL })
                   return
                 }
 
                 if (Hotkeys.isMoveWordForward(nativeEvent)) {
                   event.preventDefault()
+
+                  if (selection && Range.isExpanded(selection)) {
+                    Transforms.collapse(editor, { edge: 'focus' })
+                  }
+
                   Transforms.move(editor, { unit: 'word', reverse: isRTL })
                   return
                 }


BrentFarese (approved):
Looks good to me.

packages/slate-react/src/components/editable.tsx
```diff
@@ -912,13 +912,25 @@ export const Editable = (props: EditableProps) => {
 
                 if (Hotkeys.isMoveWordBackward(nativeEvent)) {
                   event.preventDefault()
-                  Transforms.move(editor, { unit: 'word', reverse: !isRTL })
+
+                  if (selection && Range.isCollapsed(selection)) {
+                    Transforms.move(editor, { unit: 'word', reverse: !isRTL })
+                  } else {
+                    Transforms.collapse(editor, { edge: 'start' })
+                  }
```
TheSpyder, line 920: I like this change, but a simple collapse isn't _quite_ how the range selection scenario works, at least on macOS. It collapses and then moves.

```suggestion
                  if (selection && !Range.isCollapsed(selection)) {
                    Transforms.collapse(editor, { edge: 'start' })
                  }
                  
                  Transforms.move(editor, { unit: 'word', reverse: !isRTL })
```

Some apps collapse to the focus rather than the start, which is how I prefer to do it, but there is no consistency even amongst Apple apps so that's more of a personal choice.

clauderic: I agree, there is definitely inconsistency across apps and platforms around this. I don't think that's quite correct however.

In plain text inputs on macOS:
- When moving towards the anchor, the selection is first collapsed at the anchor.
- When moving towards the focus, the selection is collapsed at the focus and moved to the next word.

https://user-images.githubusercontent.com/1416436/116172295-76ba4e00-a6d8-11eb-9679-bfa8078f35c2.mp4

I originally took my cues on this from Google Docs, which to me feels like the more predictable behavior, but I generally think Slate shouldn't be opinionated about these types of things so it's probably best to mimic the OS behaviour so happy to update this PR to match that behavior if others agree

TheSpyder: > * When moving towards the anchor, the selection is first collapsed at the anchor.

Single word selections can be misleading; try selecting 3 words and see what happens. My read of your video is that it is collapsing to the focus and then moving a word, but it isn't totally clear with just a single word.

Selecting multiple words left-to-right, then pressing alt+left, puts the cursor in the middle of the range that was selected.

clauderic: You're right 😄 Will update accordingly

clauderic: https://user-images.githubusercontent.com/1416436/116178140-06fd9080-a6e3-11eb-9bdf-523ec6282f26.mp4

[Updated commit](https://github.com/ianstormtaylor/slate/pull/4211/commits/5077c85e4309fc0fa7347c700e13036dec6dce27)

.changeset/move-word-collapse.md
```diff
@@ -0,0 +1,6 @@
+---
+'slate-react': patch
+'slate': patch
```
juliankrispel, line 3: 🤔 afaik this patch only affects slate-react?

juliankrispel (commented):
Really liking this 🎉 nicely done. It'd be great if a test could be added as well 🙏

clauderic:
Could we get this PR merged? 

I don't have bandwidth to add cypress tests to this PR at the moment, but I don't think that should stop this PR from being merged. As outlined in the PR description, this entire code path is untested and it would take more effort than I can dedicate to this at the moment to make sure it's thoroughly tested.

BrentFarese:
This seems good to me.  In general, I think we should be writing more tests for slate-react but appreciate not everyone has the bandwidth to do it.  It'd also be helpful to have more Cypress scaffolding in place so it's easier to write tests, but that will come in time.

Issue #4170, opened 2021-04-05
slate-react 0.62.0 - Chromium iframe crash


Mangatt:
**Description**
When you try to use `[Editable]` inside `iframe`, your app will crash with `Unable to find DocumentOrShadowRoot for editor element: [object HTMLDivElement]` error. Version `0.61.3` do not have this issue.

**Recording**
Not necessary, just open sandbox.

**Sandbox**
https://codesandbox.io/s/slate-reproductions-forked-rzk5c?file=/index.js

**Steps**
To reproduce the behavior:
1. See sandbox with error
2. Remove `TextEditor` from iframe and everything works just fine

**Expectation**
Editable should work in every `window` context, as it used to work.

**Environment**
- Slate Version: 0.62.0
- Operating System: Win10
- Browser: Any Chromium based (works just fine in FF)

codercodingthecode:
I have the exact same error with Editable under a chrome extension I'm building with slate.

karudo:
https://www.slatejs.org/examples/iframe
Iframe example does not work for me in Chrome 89.0.4389.114

dusty-phillips:
I think this is the underlying issue behind some odd failures in my cypress tests.

PR #4219, opened 2021-04-23
Fix iframe crash

juliankrispel:
**Description**
Fixes iframe crashes

Context: The exception is thrown if an editor root element does not inherit from `Document`. However an document inside iframe will never inherit the same prototype as `window.document`, this change prevents triggering this exception if we are inside an iframe

**Issue**
Fixes: #4170

**Example**
Currently the iframe example crashes with the exception `Unable to find DocumentOrShadowRoot for editor element ...`
image

With this change it works again:
iframe-slate-fix

**Checks**

### DIFF
--- a/.changeset/thick-apricots-fly.md
+++ b/.changeset/thick-apricots-fly.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fixes error thrown when Editor is rendered inside iframe
--- a/.changeset/thick-apricots-fly.md
+++ b/.changeset/thick-apricots-fly.md
@@ -2,4 +2,4 @@
 'slate-react': patch
 ---
 
-Fixes error thrown when Editor is rendered inside iframe
+Fixes error that occurs when Editor is rendered inside iframe
--- a/cypress/integration/iframe.ts
+++ b/cypress/integration/iframe.ts
@@ -1,0 +1,31 @@
+// Taken from https://www.cypress.io/blog/2020/02/12/working-with-iframes-in-cypress/
+const getIframeDocument = () => {
+  return cy
+    .get('iframe')
+    .its('0.contentDocument')
+    .should('exist')
+}
+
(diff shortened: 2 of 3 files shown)


juliankrispel:
🤔 One question that nags me is [why do we throw an exception](https://github.com/ianstormtaylor/slate/pull/3749/files#r619027341) in `findDocumentOrShadowRoot` in the first place?
Wouldn't it be safer to remove the exception entirely or would this break something else and if so how can we test this. 

Copying in @davidruisinger and @ianstormtaylor maybe you have some context I am missing

juliankrispel:
- [x] Added smoke test for iframe example so that this issue doesn't regress

clauderic:
I'm assuming this issue only happens when the editor is rendered within a portal DOM element that belongs to an iframe? Would have been good to clarify.

juliankrispel:
> I'm assuming this issue only happens when the editor is rendered within a portal DOM element that is in an iframe? Would have been good to clarify 🙏

That is one of the issues yes (who knows what other usecases there are that fall into this category). In this case it's affecting the iframe example so https://www.slatejs.org/examples/iframe doesn't currently work (until this change is deployed).

Nauss:
I was facing this issue in a different context.
I'm building an electron app where the user is able to create a report with rich texts (slate editors in a page). The user can then print his/her report. To do that I use an external child window created with `window.open` where I render the slate editors through a Portal.
This PR fixes the problem but while trying to fix this bug myself (before seeing this pull request), I tried the following:
```
var thisWindow = ReactEditor.getWindow(editor);
if (!(root instanceof thisWindow.Document || root instanceof thisWindow.ShadowRoot)) throw new Error("Unable to find DocumentOrShadowRoot for editor element: ".concat(el));
```
This works because `root` is still a `Document` but from the child window, not the parent.
Not sure if this would work with an iframe...

PR #4224, opened 2021-04-24
Move input reconciliation logic out of `editable.tsx` into composable layers under `useInputReconciler` hook

clauderic:
**Description**
This PR moves all of the user input reconciliation logic out of `editable.tsx` in order to have a better separation of concerns between all the different layers of input reconciliation, as well as carve a path forward towards having swappable composition reconciliation layers (`useBeforeInputReconciler` vs `useMutationObserverReconciler`).

This PR is a refactor, so there should be zero differences in behaviour between this PR and the `main` branch. 

**Context**
I've broken down user input reconciliation into the following layers:

- **Composition** (which in the future can be further broken down into different composition reconciliation strategies)
- **Clipboard** (copy, cut, paste)
- **Drag and drop**
- **Focus**
- **Selection**

Screen Shot 2021-04-24 at 1 07 29 PM

The boundaries between `selection` and `focus` are somewhat blurry, I hesitated between combining them or not. Decided to split them out but I could be convinced otherwise.

The main downside of merging this PR is that it touches a lot of files that are commonly modified in PRs, so those will have to be rebased. It also means that this PR will need to be rebased anytime the `editable.tsx` file changes, so I'd love to get eyes on this PR before other major changes are made to `editable.tsx.`

Long term, I strongly believe that this approach will lead to a better separation of concerns, which should help with testability, maintainability and also enable swapping only certain aspects of input reconciliation (for example, swapping the composition layer on Android).

**Checks**


ianstormtaylor:
@clauderic can you elaborate on what the goals for this refactoring are? Or what new things it unlocks? I'd love some more context before I review. What types of customization benefits from this, whether it's specific to Android or whether it has benefits for other uses cases, etc. Thanks!

clauderic:
Hey @ianstormtaylor, the main motivation behind this refactor is to unlock having interchangeable layers of input reconciliation.

The primary use-case for now is to be able to build conditionally swappable composition reconciliation strategies depending on the browser capabilities (`beforeinput` vs `MutationObserver`), without having to duplicate the logic that is now encapsulated in the other reconciliation layers.

Breaking down the logic that currently lives in `editable.tsx` allows to create a better separation of concerns between the different aspects of input management that need to be reconciled with Slate's internal model, which should also allow better maintainability, test ability and documentation of these layers.

In the future, you could also choose to assemble these input reconciliation layers differently; for example, the drag and drop layer doesn't need to be included on mobile devices since it builds on the HTML5 Drag & Drop API, which isn't supported on mobile devices.

thesunny:
Hi @clauderic 

I think we are a lot closer that we thought WRT to Android (at least with respect to this step). I'm glad you posted the PR because that's similar to what I proposed to Wayne as a way to reduce code duplication. So seeing what you were thinking helps. I agree that re-using code and placing them in hooks is a good idea.

I suppose the primary difference is that I still see separate `Editable`s as a good way to separate the `beforeinput` approach and the `mutable` approach. My thought is that we could import the correct hooks into the appropriate version of the `Editable`. The reason I like this is that it feels less magical because it uses one less level of indirection. I also like it (personally) because it's still not clear to me that there aren't other things that we may want in \`Editable\` that's specific to just the Android input method and separate `Editables` give us a direct route to make those changes instead of going through the abstractions. It's like the difference between the plugin system in 0.47x vs the "just use JavaScript" approach in 0.50+. It's not exactly the same, but generally I feel like it's simpler.

Now all this said, and acknowledging that I generally think this is a useful abstraction (and like I said, even suggested it in Wayne's PR), I still do wonder if we can reduce by another layer of abstraction. Like I'm not sure if I like importing `handlers` which is a combined set of handlers because it's very opaque. I get the sense when reading the code that you really can't tell where things are coming from and you end up doing a "look up the tree" type of search.

For example, let's take the file `use-clipboard-reconciler`. In the PR, it is combined into a single object and exported. Looking at the code (at least without TypeScript) it's difficult to see what's coming out of there. I wonder if it wouldn't work better if we just exported each method individually without any combining:

```typescript
export const useOnCopy = useCallback(...)
export const useOnCut = useCallback(...)
export const useOnPaste = useCallback(...)
```

One thing I like about that is it's really explicit what's happening here in terms of what's coming out. Then we import it into both our respective `EditableDefault` and `EditableAndroid` it's also very clear.

```typescript
import {useOnCopy, useOnCut, useOnPaste} from 'use-clipboard-reconciler'

// pseudo-code Default Editable
const onCopy = useOnCopy()
const onCut = useOnCut()
const onPaste = useOnPaste()

return <Component onCopy={onCopy} onCut={onCut} onPaste={onPaste} />
```

Some benefits of this as well because it removes a level of abstraction is that it can be easier to work with. For example, let's say that it turns out Android has funny behavior with `onPaste`. We can easily swap out just that one function.

```typescript
import {useOnCopy, useOnCut, useOnPasteAndroid} from 'use-clipboard-reconciler'

// pseudo-code Android Editable
const onCopy = useOnCopy()
const onCut = useOnCut()
const onPaste = useOnPasteAndroid()

return <Component onCopy={onCopy} onCut={onCut} onPaste={onPaste} />
```

It's still another level of indirection/abstraction but it's a little lighter. I feel like generally, Slate 0.50+ has been moving to more direct code and for me, I've been finding it a lot easier to comprehend and manage.

What do you think with respect to keeping hooks to reduce code duplication but combining them in the Editable instead of in an all-in-one hook? Also, I'd like to hear @ianstormtaylor thoughts on this option.

clauderic:
Hey @thesunny, thanks for your feedback 👍 

> I suppose the primary difference is that I still see separate Editables as a good way to separate the beforeinput approach and the mutable approach. [...] I also like it (personally) because it's still not clear to me that there aren't other things that we may want in `Editable` that's specific to just the Android input method and separate Editables give us a direct route to make those changes instead of going through the abstractions

What concrete examples of this do you have? 

Looking at what's now left in `editable.tsx`, I can't think of anything that would need to be different enough on Android to warrant forking that component in its entirety.  


> ```tsx
> import {useOnCopy, useOnCut, useOnPaste} from 'use-clipboard-reconciler'
> 
> // pseudo-code Default Editable
> const onCopy = useOnCopy()
> const onCut = useOnCut()
> const onPaste = useOnPaste()
> ```

While this pseudo-code looks like it leads to a nice separation of concerns, in practice, most input reconciliation layers need to have a some level of configuration, as well as shared state, such that breaking them out into separate methods that are imported would be repetitive and impractical. 

> ```tsx
> import {useOnCopy, useOnCut, useOnPaste} from '../input-reconciler/use-clipboard-reconciler'
> 
> // pseudo-code
> const state = useReconcilerState();
> const onCopy = useOnCopy(editor, {
>   node: nodeRef,
>   state, 
>   handler: attributes.onCopy, 
>   options: {readOnly},
> })
> const onCut = useOnCut(editor, {
>   node: nodeRef,
>   state, 
>   handler: attributes.onCut, 
>   options: {readOnly},
> })
> const onPaste = useOnPaste(editor, {
>   node: nodeRef,
>   state, 
>   handler: attributes.onPaste, 
>   options: {readOnly},
> })
>
>  <Component {...onCopy.handlers} {...onCut.handlers} {...onPaste.handlers} />
> ```

You could however remove the `useInputReconciler` abstraction and move the logic that is encapsulated in that abstraction over to `Editable`. 

> ```tsx
> import {useCopyPaste} from '../input-reconciler/use-clipboard-reconciler'
> import {useDragDrop} from '../input-reconciler/use-drag-drop-reconciler'
> /* ... */
> 
> // pseudo-code
> const state = useReconcilerState();
> const copyPaste = useOnCopyPaste(editor, {
>   node: nodeRef,
>   state, 
>   handlers: [attributes.onCopy, attributes.onCut, attributes.onPaste], 
>   options: {readOnly},
> })
> const dragDrop = useDragDrop(editor, {
>   node: nodeRef,
>   state, 
>   handlers: [attributes.onDragStart, attributes.onDragOver, attributes.onDrop], 
>   options: {readOnly},
> })
>
>  <Component {...copyPaste.handlers} {...dragDrop.handlers} />
> ```

You would essentially have to copy over the entirety of the code that lives under that hook, which doesn't feel like an improvement to me. 

Personally I still think that all input reconciliation logic should be grouped and extracted out of `Editable` to separate concerns, but happy to consider other approaches around that particular aspect of the PR if we can think of better abstractions.

thesunny:
I think this is something I'd like @ianstormtaylor to weigh in on again.

I think you are probably correct in that there is nothing else that needs to be in `Editable` that couldn't be expressed in hooks based on our recent conversations that we had elsewhere. This could also be expressed with shared methods across multiple `Editable`'s which would remove the code duplication so it feels like it's a matter of aesthetics.

For me, I do find that using a single hook obfuscates the code somewhat and I feel like that goes against some of the recent directions that Ian has been taking the code like the removal of a formal plugin structure and instead using plain JavaScript (TypeScript). I could personally live with the obfuscation, but I think that is because I am involved in the process so I know what's going on. If I was reading the code for the first time, having to dive into a single hook to see how everything wires up underneath would be harder to understand IMO.

All this said, this seems like a decision for Ian.

ianstormtaylor:
Hey @clauderic thanks for taking the time to write up this potential direction.

Overall I'm okay with this approach, although I say that with some hesitation because it will add complexity to Slate and specifically to a part of Slate that is very hard to test and contribute to.

That said, I'd like to see some changes to this PR to reduce some of the layers of abstraction/indirection to make this split as simple to follow as possible. It feels a little over-abstracted right now (kind of how I felt about Draft.js before writing Slate) and I'd like to get that fixed before merging. From a quick review that would be:

- Eliminating the `index` files with single exports (since they only add indirection).
- Eliminating the single-file folders (since they only add indirection, everything should be in one folder).
- Moving/renaming the `utils/helpers` exports (since it's just a grab bag).
- Un-nesting `reconciler.handlers` (since there are no other properties).
- Removing a lot of the extra LOC in `useInputReconciler`.
- Removing `input-reconciler/types` (and just define inline instead).
- Ensuring all files are kebab cased (to match the codebase).
- Removing the `extends ReconcilerArguments` stuff (seemed like none extended it?).
- Removing the term "reconciler" altogether (and just call these "handlers").

I'd be happy to do a more in-depth review once the overall picture is simpler. Let me know what you think of that, and if there are any pieces that don't make sense when taking the future mutation observer bits into account.

Thanks!

dylans (changes requested):
Marking this one as needing changes per the previous feedback. Also will need a rebase against main.

dylans:
I'm going to close this one out due to inactivity and the list of feedback that would need to be addressed. Feel free to reopen it if I got this wrong, or create a new PR if that's more appropriate.

clauderic:
@dylans I'm planning to open a new PR that addresses the feedback. This PR was quite stale and would have been difficult to rebase anyway.

PR #4227, opened 2021-04-26
Fix addMark/removeMark caused unexpected onChange.

ulion:
**Description**
Collaborate editor relies on onChange and editor.operations, but in editor.addMark and editor.removeMark, onChange is called directly without check it was already scheduled by editor.apply method. which will cause onChange with same group or part of same group of editor.operations be called twice, then cause problem with collaborate editors.

**Issue**
Fixes: (link to issue)

**Context**
This PR simply check whether there was already a scheduled onChange, if so, do not call onChange in addMark and removeMark.

**Checks**

### DIFF
--- a/.changeset/three-parrots-remember.md
+++ b/.changeset/three-parrots-remember.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Fixed a bug that would allow multiple changes to be scheduled at the same time.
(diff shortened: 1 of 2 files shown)


TheSpyder (approved):
Looks good to me, nice catch!

This will still lead to possibly unexpected change events (when `editor.operations` is an empty array) but at least that shouldn't cause any problems.

ianstormtaylor:
@ulion can you add a changeset please? Thanks. (Or allow edits to your branch so that I can do it myself.)

ulion:
> @ulion can you add a changeset please? Thanks. (Or allow edits to your branch so that I can do it myself.)

invited you with write access.

ianstormtaylor:
Thanks @ulion!

PR #4230, opened 2021-04-30
Added insert_node path checking to match how other operations are applied

TheSpyder:
**Description**
Applying an `insert_node` operation now throws an exception if the path to insert is invalid. Previously it only threw an exception if the parent path doesn't exist.

**Context**
This came up while developing our collaborative editor based on Slate. Allowing invalid `insert_node` operations means a mistake in the transforms is not picked up immediately; it only causes a failure when a subsequent operation depends on the invalid insert. This makes bugs harder to track down, and is inconsistent with how every other operation is applied.

Other operation apply cases throw an exception when the operation is invalid - either directly or implicitly by calling `Node.get`. `insert_node`, however, only used `Path.parent` and then `splice` directly on the children array. Splice does not throw an exception for invalid indexes.

While writing the test, I found two places a try/finally was necessary to ensure the editor state was still valid after an exception:
* In the general transform method , if an exception was triggered `editor.children` remained in draft status.
* In `Editor.withoutNormalizing()`, if an exception was triggered the document would never again be normalized.

This is arguably a bug fix patch level change, but it could trigger new exceptions for some use cases so I've made it a minor.

**Checks**

### DIFF
--- a/.changeset/bright-pants-grin.md
+++ b/.changeset/bright-pants-grin.md
@@ -1,0 +1,5 @@
+---
+'slate': minor
+---
+
+Applying invalid `insert_node` operations will now throw an exception for all invalid paths, not just invalid parent paths.
--- a/.changeset/sixty-glasses-beg.md
+++ b/.changeset/sixty-glasses-beg.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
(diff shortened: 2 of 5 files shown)


ianstormtaylor:
Thanks @TheSpyder!

Issue #4136, opened 2021-03-22
Crash if you enter multibyte characters.


fs-kanayama:
**Description**
Deletes all characters and crashes when entering multibyte characters.

**Recording**
![screen-recording (1)](https://user-images.githubusercontent.com/1792852/111935230-6249cb00-8b06-11eb-88c3-84334a970998.gif)

**Sandbox**
Example
https://www.slatejs.org/examples/plaintext

Following the same steps, CodeSandbox crashes in the same way.
https://codesandbox.io/s/slate-reproductions-c7gyg

**Steps**
To reproduce the behavior:
1. Change IME Japanese
2. Delete all characters with Backspace
3. Keep the focus and enter the text.
4. Press Enter

```
react_devtools_backend.js:2430 Error: Cannot resolve a Slate point from DOM point: [object Text],4
    at Object.toSlatePoint ([example]-6c432d8938a9c9ec3d0c.js:1)
    at Object.toSlateRange ([example]-6c432d8938a9c9ec3d0c.js:1)
    at [example]-6c432d8938a9c9ec3d0c.js:1
    at uo (framework.c4060028918b378fc130.js:1)
    at co (framework.c4060028918b378fc130.js:1)
    at Cu (framework.c4060028918b378fc130.js:1)
    at UdKW.t.unstable_runWithPriority (framework.c4060028918b378fc130.js:1)
    at $l (framework.c4060028918b378fc130.js:1)
    at Su (framework.c4060028918b378fc130.js:1)
    at du (framework.c4060028918b378fc130.js:1)
```

**Environment**
- Operating System: Windows 10
- Browser: Chrome

fs-kanayama:
This is not reproduced when placeholder is not used.
I think it is a placeholder-related bug.

pjhoogendijk:
this also happens when entering characters like ä via mac (alt+u and then a)

h-sakano:
The same problem occurs when multibyte characters are entered immediately after an inline element, as shown below.

![output](https://user-images.githubusercontent.com/26650149/116025298-1e109400-a68b-11eb-9605-1c74285b9dfe.gif)

PR #4234, opened 2021-05-03
[WIP]fix(slate-react): prevent a crash when typing in IME mode after an inline element

h-sakano:
**Description**

When text is entered without IME mode, the `insertText` function will adjust the insertion point of the text after the inline element.

However, in IME mode, the text is inserted before the process is executed, as shown in the image below, so the text is inserted in the wrong place.

Text is inserted in the wrong place

To fix the above problem, the listener of the beforeInput event will now adjust the text insertion point after the appropriate inline element.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4136#issuecomment-826478544

**Checks**

### DIFF
--- a/.changeset/short-flowers-flash.md
+++ b/.changeset/short-flowers-flash.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+Prevent a crash when typing in IME mode after an inline element.
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -267,12 +267,31 @@ export const Editable = (props: EditableProps) => {
         const { inputType: type } = event
         const data = (event as any).dataTransfer || event.data || undefined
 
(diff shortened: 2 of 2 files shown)


h-sakano:
I found a bug, so I added "WIP" to the title

h-sakano:
I have found that the solution in this pull request does not solve the problem of text being inserted in the wrong place, so I close it.
If a solution is found, I will create the pull request again.

githoniel:
@h-sakano I think this is bug of React Reconciliation. the link element did re-render but React considers it unnecessary to update the DOM. I don't find any generic solutions but we can fix it by adding a key to force update DOM like below

```jsx
const Element = ({ attributes, children, element }) => {
  switch (element.type) {
    case 'link':
      return (
        <a {...attributes} href={element.url} key={Math.random()}>
          {children}
        </a>
      )
    default:
  }
}
```

PR #4237, opened 2021-05-04
Update IS_FIREFOX_LEGACY flag to prevent FF crash

### LINKED ISSUE #3855: Bug: Entering an emoji causes the editor crashing in Firefox

The editor behaves weirdly and crashes when you either:
1. Replace all the content with an emoji, The editor crashes.
2. Insert an emoji (first) when there is no content, delete it, and insert an emoji (second).
2.1 You will see the second emoji appears twice. The second one cannot be deleted.
2.2 If you delete the first one, and insert another emoji, it crashes.

iShot2020-09-03-11 33 41
iShot2020-09-03-11 37 31

Slate: > 0.58.3
Browser: Only Firefox 
OS: Mac OS Catalina

The emoji should not appear twice. The emoji should work as other texts.

dylans:
#4118 and #4150 introduced IS_FIREFOX_LEGACY, but this block of code was not updated to reflect that. This breaks insertion of emoji and unicode characters in slate-react. This change fixes #3855.

**Description**
As noted in #3855, Firefox crashes when inserting emoji using the emoji keyboard.

**Issue**
Fixes: #3855

**Example**
This fixes an obviously overlooked update/typo. The example in #3855 shows the issue before this change. With this fix, the editor behaves as expected (doesn't crash when inserting an emoji)

**Context**
Correct a typo

**Checks**

### DIFF
--- a/.changeset/forty-mails-roll.md
+++ b/.changeset/forty-mails-roll.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fixed text insertion logic to prevent crashing in newer Firefox versions.
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -17,7 +17,7 @@ import scrollIntoView from 'scroll-into-view-if-needed'
 import useChildren from '../hooks/use-children'
 import Hotkeys from '../utils/hotkeys'
 import {
-  IS_FIREFOX,
+  IS_FIREFOX_LEGACY,
   IS_SAFARI,
   IS_EDGE_LEGACY,
   IS_CHROME_LEGACY,
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -653,7 +653,7 @@ export const Editable = (props: EditableProps) => {
                 // aren't correct and never fire the "insertFromComposition"
                 // type that we need. So instead, insert whenever a composition
                 // ends since it will already have been committed to the DOM.
-                if (!IS_SAFARI && !IS_FIREFOX && event.data) {
+                if (!IS_SAFARI && !IS_FIREFOX_LEGACY && event.data) {
                   Editor.insertText(editor, event.data)
                 }
               }
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -17,6 +17,7 @@ import scrollIntoView from 'scroll-into-view-if-needed'
 import useChildren from '../hooks/use-children'
 import Hotkeys from '../utils/hotkeys'
 import {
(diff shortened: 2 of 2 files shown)


TheSpyder:
Changes in 2021 can't cause a bug logged in September 2020. In addition, the original report was on Firefox 80 and `IS_FIREFOX_LEGACY` matches before 87 suggesting versions in between would still break with this change.

I tracked down where the `IS_FIREFOX` check was added - #3202. Around the time of Firefox 70. It seems weird that Firefox would do `beforeinput` correctly and then a year later change it to match chrome's behaviour months before `beforeinput` support was properly released.

Have you tested this with regular composition input?

dylans:
@TheSpyder, I agree that it's peculiar and it took a lot of poring through the codebase to try to understand what's going on, and I still don't really know if I understand this one.

> Changes in 2021 can't cause a bug logged in September 2020. In addition, the original report was on Firefox 80 and `IS_FIREFOX_LEGACY` matches before 87 suggesting versions in between would still break with this change.

We just upgraded from 0.59.x to the latest Slate and just recently upgraded to Firefox 88, which perhaps makes sense, given that we were running code prior to the change in #3202?

> I tracked down where the `IS_FIREFOX` check was added - #3202. Around the time of Firefox 70. It seems weird that Firefox would do `beforeinput` correctly and then a year later change it to match chrome's behaviour months before `beforeinput` support was properly released.

I don't have the history to know if it was broken or not with the various Firefox releases over the past year. I don't know the details of the report for #3855 beyond it sounding like the same issue we observe.
 
> Have you tested this with regular composition input?

I'm not 100% sure I understand the term "regular composition input" though I can guess.

What I observe is that normal typing of text is fine (no crashing or errors in calculating offsets), copy/pasting of emojis from other content is fine, but using the macOS emoji keyboard to insert an emoji causes behavior reported in #3855 with the latest Slate and Firefox, and modifying the line in this PR prevents the crash we were observing. If this is not what you mean, please explain the term and I'll reply with more detail.

TheSpyder:
By “regular composition input” I mean a full IME where typing multiple keys create one character. That’s what was being fixed in 3202. It’s probable that the macos emoji picker uses similar composition events to insert as an IME, but I’d like to be sure this change doesn’t cause a regression.

You don’t have to know another language to use an IME, just enable it as a keyboard option (Chinese is my usual go-to for this) and hit a few keys until you have less visible characters than you typed.

As for the Firefox version, it might be nice to have it fixed on more versions but I guess most Slate users on Firefox will have upgraded to at least 87 by now.

dylans:
Thanks @TheSpyder for the explanation.

I can confirm that this patch works when using the Chinese-simplified Pinyin keyboard, and is broken without this patch, in the same way as with emojis.

ianstormtaylor:
Thanks @dylans!

Issue #4222, opened 2021-04-23
D'n'D. When dragging, the text jumps to the beginning of the editor.


fredure:
**Description**
If you select text at the beginning or in the middle of the editor, drag it to the end, the selected text moves to the beginning of the editor.

**Recording**
![Apr-23-2021 18-29-51](https://user-images.githubusercontent.com/8024592/115894426-effc3b80-a461-11eb-8d77-864c7e91ae5d.gif)

**Sandbox**
https://www.slatejs.org/examples/plaintext

**Steps**
To reproduce the behavior:
1. Select text not at the beginning of the editor
2. Move selected text to the end of the editor
3. See error

**Expectation**
The selected text is expected to move to the specified location

**Environment**
Slate Version: 0.62.0.
Operating System: all.
Browser: Google Chrome 88.0.4324.192.

juliankrispel:
Pretty sure this is a regression since I've seen this work previously, when we fix this let's add a test to make sure it doesn't happen again.

clauderic:
Solved by #4238

PR #4238, opened 2021-05-04
Fix duplicated content and other bugs related to drag and drop handling

clauderic:
**Description**
This PR fixes a number of different issues that happen today with drag and drop. 
It's loosely based off of this abandonned PR: https://github.com/ianstormtaylor/slate/pull/3562

**Issue**
Fixes: 
https://github.com/ianstormtaylor/slate/issues/4222
https://github.com/ianstormtaylor/slate/issues/3522
https://github.com/ianstormtaylor/slate/issues/3137
https://github.com/ianstormtaylor/slate/issues/2076

**Examples**

https://user-images.githubusercontent.com/1416436/117067981-b1d00900-acf8-11eb-9271-df09e50ed963.mp4

https://user-images.githubusercontent.com/1416436/117067990-b4326300-acf8-11eb-838f-feb8b49c23f3.mp4

https://user-images.githubusercontent.com/1416436/117202852-51a29b00-adbc-11eb-8f0e-b9f20718ee33.mp4

https://user-images.githubusercontent.com/1416436/117068035-c14f5200-acf8-11eb-9818-6bfaeb6c4c0d.mp4

https://user-images.githubusercontent.com/1416436/117068045-c4e2d900-acf8-11eb-9ba7-ebf41c1d9d76.mp4

https://user-images.githubusercontent.com/1416436/117068049-c7453300-acf8-11eb-890d-fae171d71774.mp4

https://user-images.githubusercontent.com/1416436/117068075-cc09e700-acf8-11eb-9dad-b1cf6dd74d3b.mp4

**Context**

The main things this PR does:

- Keeps track of whether the drag start event happened internally
- If the drag start event happened internally, the original dragged range needs to be deleted before inserting a new fragment
- It's better to use `insertFragment` and call `preventDefault()` in the `drop` event rather than letting the `beforeinput` event go through. This strategy works consistently across all browsers and would also work with other input reconciliation strategies once Android support lands.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -130,6 +130,7 @@ export const Editable = (props: EditableProps) => {
   const state = useMemo(
     () => ({
       isComposing: false,
+      isDraggingInternally: false,
       isUpdatingSelection: false,
       latestElement: null as DOMElement | null,
     }),
@@ -423,7 +424,12 @@ export const Editable = (props: EditableProps) => {
   // while a selection is being dragged.
   const onDOMSelectionChange = useCallback(
     throttle(() => {
-      if (!readOnly && !state.isComposing && !state.isUpdatingSelection) {
+      if (
+        !readOnly &&
+        !state.isComposing &&
+        !state.isUpdatingSelection &&
+        !state.isDraggingInternally
+      ) {
         const root = ReactEditor.findDocumentOrShadowRoot(editor)
         const { activeElement } = root
         const el = ReactEditor.toDOMNode(editor, editor)
@@ -748,7 +754,9 @@ export const Editable = (props: EditableProps) => {
               ) {
                 const node = ReactEditor.toSlateNode(editor, event.target)
                 const path = ReactEditor.findPath(editor, node)
-                const voidMatch = Editor.void(editor, { at: path })
+                const voidMatch =
+                  Editor.isVoid(editor, node) ||
+                  Editor.void(editor, { at: path, voids: true })
 
                 // If starting a drag on a void node, make sure it is selected
                 // so that it shows up in the selection's fragment.
@@ -756,6 +764,8 @@ export const Editable = (props: EditableProps) => {
                   const range = Editor.range(editor, path)
                   Transforms.select(editor, range)
                 }
+
+                state.isDraggingInternally = true
 
                 ReactEditor.setFragmentData(editor, event.dataTransfer)
               }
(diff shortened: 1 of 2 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -423,7 +424,12 @@ export const Editable = (props: EditableProps) => {
   // while a selection is being dragged.
   const onDOMSelectionChange = useCallback(
     throttle(() => {
-      if (!readOnly && !state.isComposing && !state.isUpdatingSelection) {
+      if (
+        !readOnly &&
+        !state.isComposing &&
+        !state.isUpdatingSelection &&
+        !state.isDraggingInternally
```
clauderic, line 431: Noticed issues in Firefox when trying to reconcile the selection while dragging

packages/slate-react/src/components/editable.tsx
```diff
@@ -748,7 +754,9 @@ export const Editable = (props: EditableProps) => {
               ) {
                 const node = ReactEditor.toSlateNode(editor, event.target)
                 const path = ReactEditor.findPath(editor, node)
-                const voidMatch = Editor.void(editor, { at: path })
+                const voidMatch =
+                  Editor.isVoid(editor, node) ||
+                  Editor.void(editor, { at: path, voids: true })
```
clauderic, line 759: Not sure why `Editor.void` doesn't check the current path, seems it only checks `above` the current path which is a bit counter-intuitive, so we also need to also if the node at the current path is void.

packages/slate-react/src/components/editable.tsx
```diff
@@ -765,28 +775,58 @@ export const Editable = (props: EditableProps) => {
...
-                // https://bugs.chromium.org/p/chromium/issues/detail?id=1028668
-                if (
-                  !HAS_BEFORE_INPUT_SUPPORT ||
-                  (!IS_SAFARI && event.dataTransfer.files.length > 0)
-                ) {
-                  event.preventDefault()
-                  const range = ReactEditor.findEventRange(editor, event)
-                  const data = event.dataTransfer
-                  Transforms.select(editor, range)
-                  ReactEditor.insertData(editor, data)
```
clauderic, line 784: This logic should run on all browsers, not just the ones that don't have `beforeinput` support. Deferring to `beforeinput` causes most of the text-insertion related bugs that this PR fixes. This change is aligned with how copy paste is handled in Slate.

ianstormtaylor:
Thanks @clauderic!

PR #4188, opened 2021-04-11
Ensure round-trip tests don't mutate output before checking against it => reveals 4 tests are broken

beorn:
[ This was previously reported as PR #3718 last June, but never merged.  I closed that PR in favor of this. ]

Several round-trip tests, tests that check that the output is the same as the input, set `output = input`, which means both input and output point to the same object.  By the looks of it, when the input is modified the output is as well, making the tests no-ops.

We need to do deep copy instead of shallow/reference copy, and when this is done by this PR, 4 tests, that by the looks of it previously should have failed, now start to fail:

- slate-history / undo / delete_backward / block-text
- slate-history / undo / delete_backward / custom-prop
- slate-history / undo / delete_backward / inline-across
- slate / interfaces / Node / get / root

This PR converts `const output = input` to `const output = cloneDeep(input)` AND adds a `const skip = true` for the tests that are broken and needs to be fixed.  An alternative might be to, in the test system, automatically do a deepClone of output or input if they are found to be reference to the same object (or in all cases), ensuring that they can be mutated safely; but this is a bit automagical perhaps.

I don't know what the correct procedure is to report this as it's adding failing tests instead of fixing them, and I don't pretend to be able to fix all of the below failing tests in this PR - I just want to report that the tests themselves are broken, and so there may be bugs lurking there.

**Checks**
- [X] The new code matches the existing patterns and styles.
- [X] The tests pass with `yarn test`. => **4 TESTS ARE BROKEN AND SET TO skip=true**
- [X] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.) => many seem to be broken in node_moduels
- [X] The relevant examples still work. (Run examples with `yarn start`.)

# Tests that fail (and are now skipped)

- slate-history / undo / delete_backward / block-text
- slate-history / undo / delete_backward / custom-prop
- slate-history / undo / delete_backward / inline-across
- slate / interfaces / Node / get / root

Here's the full output:
```
  908 passing (7s)
  9 pending
  4 failing

  1) slate-history
       undo
         delete_backward
           block-text :

      AssertionError [ERR_ASSERTION]: Expected values to be loosely deep-equal:

{
  anchor: {
    offset: 3,
    path: [
      0,
      0
    ]
  },
  focus: {
    offset: 3,
    path: [
      0,
      0
    ]
  }
}

should loosely deep-equal

{
  anchor: {
    offset: 2,
    path: [
      0,
      0
    ]
  },
  focus: {
    offset: 2,
    path: [
      0,
      0
[...]

PR #4245, opened 2021-05-06
replace lodash depency in slate

JonasKruckenberg:
**Description**
This PR removes the dependency on `loadsh/omit` reducing bundle size by ~50Kb

> Edit: The scope of this has expanded to replace all usages of lodash, see below

**Example**
Before
Bildschirmfoto 2021-05-06 um 13 20 47

After
Bildschirmfoto 2021-05-06 um 13 18 46

**Checks**

### DIFF
--- a/packages/slate/src/interfaces/text.ts
+++ b/packages/slate/src/interfaces/text.ts
@@ -36,15 +36,15 @@ export const Text: TextInterface = {
   ): boolean {
     const { loose = false } = options
 
-    function omit(obj: Record<any, any>, key: string) {
-      const { [key]: _, ...rest } = obj
+    function omitText(obj: Record<any, any>) {
+      const { text, ...rest } = obj
 
       return rest
     }
 
     return isEqual(
-      loose ? omit(text, 'text') : text,
-      loose ? omit(another, 'text') : another
+      loose ? omitText(text) : text,
+      loose ? omitText(another) : another
     )
   },
 
--- a/packages/slate/src/interfaces/text.ts
+++ b/packages/slate/src/interfaces/text.ts
@@ -1,5 +1,5 @@
 import isPlainObject from 'is-plain-object'
-import isEqual from 'lodash/isEqual'
+import isEqual from 'fast-deep-equal'
 import { Range } from '..'
 import { ExtendedType } from './custom-types'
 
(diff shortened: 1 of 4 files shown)


packages/slate/src/interfaces/text.ts
```diff
@@ -37,6 +36,12 @@ export const Text: TextInterface = {
   ): boolean {
     const { loose = false } = options
 
+    function omit(obj: Record<any, any>, key: string) {
+      const { [key]: _, ...rest } = obj
```
TheSpyder, line 40: `omit` is only used one way, with `text` as the key, so perhaps we should just make it specific for this case?

TheSpyder (commented):
For the record, `omit` was only recently added in #3532. I was just making plans to measure how much that merge bloated the bundle size, so thank you for this! Is the total `lodash` size in your screenshot just for Slate? 50kb for `isEqual` is also unacceptable IMO but one step at a time.

JonasKruckenberg:
> Is the total `lodash` size in your screenshot just for Slate?

Yes all functions are either imported by other lodash functions or by slate

> Having the remaining 50kb just for `isEqual` is also unacceptable IMO but one step at a time.

Thinking about that, does the equality function need to check nested objects? I mean since it's equality on leave nodes and those shouldn't really have nested stuff anyway right?

Because if we don't allow nested keys the equality check is trivial and doesn't need isEqual too 🤔

packages/slate/src/interfaces/text.ts
JonasKruckenberg: you're right, addressed in the lates commit

JonasKruckenberg:
Alright, I replaced the lodash/isEqual function with fast-deep-equal. This allows us to drop loads entirely 👍🏻

TheSpyder (approved):
LGTM. Can you please also update Slate's `package.json`? Remove `lodash`, add `fast-deep-equal`?

TheSpyder:
correction: lodash needs to remain as a dev dependency after #4188 was merged.

JonasKruckenberg:
> LGTM. Can you please also update Slate's `package.json`? Remove `lodash`, add `fast-deep-equal`?

Yep you're right I forgot to commit the change ^^. loads is now a dev-rep

TheSpyder:
> I don't really know what changes to add, should this be released as a patch? probably right?

Yeah, just a patch change. Keeping it brief is fine, just a short note that lodash is gone and why.

JonasKruckenberg:
Yep added the changeset, wasn't feeling very inspired when I wrote it though xD

TheSpyder:
eh what you wrote is fine XD

PR #4250, opened 2021-05-07
Fix the problem of text being inserted in the wrong place in IME mode

h-sakano:
**Description**

In IME mode, if you try to type text right after an inline element, it will be inserted in the wrong place.
(A string will be inserted directly as a child of `data-slate-node="element"`)

before
Text is inserted in the wrong place

Therefore, I fixed it so that text inserted in the wrong place will be removed.

after

**Checks**

### DIFF
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -23,6 +23,7 @@ import {
   isDOMSelection,
   normalizeDOMPoint,
   hasShadowRoot,
+  isDOMText,
 } from '../utils/dom'
 import { IS_CHROME } from '../utils/environment'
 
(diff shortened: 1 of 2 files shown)


baijiahei:
Editor will crash in Safari

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -463,6 +464,15 @@ export const ReactEditor = {
       let leafNode = parentNode.closest('[data-slate-leaf]')
       let domNode: DOMElement | null = null
 
+      // The parent of TEXT_NODE must have `date-slate-string`.
+      if (
+        !voidNode &&
+        isDOMText(nearestNode) &&
+        !parentNode.hasAttribute('data-slate-string')
+      ) {
+        parentNode.removeChild(nearestNode)
```
dylans, line 473: I'm not sure I think the right answer is to simply remove the just inserted text. Would it not be better to wrap it in a valid parent with a transform/normalization rule instead?

dylans:
We've had a number of other recent fixes for IME insertion. I'm closing this one as I don't think the proposed solution is correct. Feel free to reopen if you disagree @h-sakano.

PR #4253, opened 2021-05-10
Skip wrapping nodes when the only match is an editor

TheSpyder:
**Description**
`Transforms.wrapNodes` attempts to detect when zero nodes match, but it missed an edge case where one node matches - the editor. The function very quickly throws `Error: Cannot get the parent path of the root path []` when this happens.

**Example**
The test shows the failure condition. The editor I'm working on supports `contenteditable=false` sections in document content, and while we are able to handle most of the filtering outside of slate this is one case that is difficult to handle without a lot of extra code.

**Context**
There are a few other ways I can think of to fix this - such as wrapping the supplied `match` function to enforce `!Editor.isEditor` - but this seemed like the least invasive way to deal with it.

**Checks**

### DIFF
--- a/packages/slate/src/transforms/node.ts
+++ b/packages/slate/src/transforms/node.ts
@@ -940,11 +940,11 @@ export const NodeTransforms: NodeTransforms = {
         if (matches.length > 0) {
           const [first] = matches
           const last = matches[matches.length - 1]
-          const [firstNode, firstPath] = first
-          const [lastNode, lastPath] = last
+          const [, firstPath] = first
+          const [, lastPath] = last
 
-          if (firstNode === lastNode && Editor.isEditor(firstNode)) {
-            // if the only matching node is an editor, no nodes matched
+          if (firstPath.length === 0 && lastPath.length === 0) {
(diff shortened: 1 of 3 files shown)


packages/slate/src/transforms/node.ts
```diff
@@ -940,8 +940,14 @@ export const NodeTransforms: NodeTransforms = {
         if (matches.length > 0) {
           const [first] = matches
           const last = matches[matches.length - 1]
-          const [, firstPath] = first
-          const [, lastPath] = last
+          const [firstNode, firstPath] = first
+          const [lastNode, lastPath] = last
+
+          if (firstNode === lastNode && Editor.isEditor(firstNode)) {
```
spocke, line 946: Wouldn't this just be `if (firstPath.length === 0 && lastPath.length === 0)`? Since the only way you would ever get editor in a nodePath pair is if the path is an empty path. Would give a slight performance boost since isEditor checks nodes etc. Would be more hard to read though so maybe a comment would make it clearer.

spocke (approved):

dylans (approved):

PR #4257, opened 2021-05-13
Revert "Temporarily revert Android commits in order to create a release of Slate focused solely on Android (#4256)"

clauderic:
This reverts commit 4f502119841d013ea2749e91f4bd1f793dc7f98e.

There's a few issues that slipped by with the initial merge of the Android PR, we're hoping to fix those in this PR before merging them.

Non exhaustive list of things to fix before merging:

- [x] `android-editable.tsx` was out of sync with `editable.tsx`, I had to [manually rebase it](https://github.com/ianstormtaylor/slate/pull/4257/commits/68a5f8dafbf05ca8c2b21c2c504ce6362fb33926). In the future, we should consider [reducing duplication as much as possible to help with maintenance](https://github.com/ianstormtaylor/slate/pull/4224)
- [x] ErrorBoundary component doesn't seem to be working, needs investigation. 
  - Update: it's not catching errors because most errors that happen during reconciliation of mutations happen outside of the render cycle, and [therefore cannot be caught by an ErrorBoundary](https://reactjs.org/docs/error-boundaries.html):
Screen Shot 2021-05-12 at 9 46 47 PM
- [x] [Couple lines of dead code in `android-editable.tsx`, and some logic that's specific to Firefox](https://github.com/ianstormtaylor/slate/pull/4257/commits/06af88a5a463c9f52c47a3f8667c4f6bea32107b)
- [x] Selection reconciliation is broken at the moment, because the selection isn't being reconciled until `onCompositionEnd` fires, which doesn't fire reliably on Android; it only fires when the keyboard is dismissed or when the a collapsed selection is expanded, which is causing a lot of issues
  - [ ] Update: this is mostly fixed now, but there are still some issues, will investigate further
- [ ] Temporary regression: composition using IMEs on Android is not working. This happens because mutations are now handled synchronously. Planning to fix that in a follow up as it requires changes to the memoization of the `leaf.tsx` and `text.tsx` component. The composition gets interrupted because the `leaf` node increments a `key` when it re-renders, which removes the node from the DOM and re-appends a new one, which interrupts the composition.

cc @thesunny


thesunny:
Thanks @clauderic for the additional investigation and work on fixing the selection issues!

thesunny:
I've said this in our several chats, but felt like I should post publicly as well. Thank you for all the excellent work you've put into fixing these issues.

clauderic:
### Current state
This PR should now be ready to be merged. It is the result of combining the changes that @wleroux and @thesunny have worked on [in this original PR](https://github.com/ianstormtaylor/slate/pull/4255) with some of the learnings we've had internally at Shopify while working on an Android plugin for Slate&nbsp;0.5.x.

Bugs should be expected; translating mutations into a set of operations that need to be reconciled onto the Slate model is not an absolute science, and requires a lot of guesswork and handling of edge cases. There are still edge cases that aren't being handled.

### Supported versions of Android

This PR is primarily targeted at Android devices running <kbd>Android 10</kbd> and <kbd>Android 11</kbd>

I've also tested against Android 9 and 8, and though a lot of things do work, those versions definitely have more bugs.

https://user-images.githubusercontent.com/1416436/118306628-ae3a4000-b4b7-11eb-8985-86b625c7fafb.mp4


https://user-images.githubusercontent.com/1416436/118306641-b1353080-b4b7-11eb-85d9-b22b7269074e.mp4


https://user-images.githubusercontent.com/1416436/118306644-b3978a80-b4b7-11eb-91f8-106f665edd28.mp4


https://user-images.githubusercontent.com/1416436/118306654-b5f9e480-b4b7-11eb-908f-c15e6ba58d3a.mp4

### Next (in follow-up PRs):
- Fixing composition using IME keyboards (requires updates to `leaf.tsx`  and `text.tsx` memoization logic)
- Making `restoreDOM` more granular such that it only restores a portion of the DOM, the current approach restores the entire document which is safe, but expensive. It causes interruptions on Google devices when holding the backspace button.

### Someday
- Finding strategies to reduce duplication between `android-editable.tsx` and `editable.tsx`
- With some work, Android 11 could potentially be supported using `beforeinput` events, such that only devices using Android 10 or below would need to use the `MutationObserver` solution.

thesunny (approved):
Thanks @wleroux and @clauderic . Excited to see this available for people to start using.

For clarity to others, the PR should not affect non-Android platforms so it should be a positive, even if there are some bugs.

circlingthesun:
According to the changelog this PR should be included in slate-react 0.64.0. On NPM the latest tag for slate-react points to 0.63.0, despite 0.64.0 being released. Is that intentional?

clauderic:
@circlingthesun definitely not intentional. Could be an issue with how changesets is configured, will investigate further, thanks for flagging this.

clauderic:
After investigating, it seems like this is indeed an issue with the changesets config, it looks like `0.64.0` was published by the `dev` workflow rather than the `latest` workflow:

Screen Shot 2021-05-20 at 11 39 14 AM

Screen Shot 2021-05-20 at 11 39 39 AM

gfohl:
Hello!  Thank you so much for the work on this, were you able to figure out the release issues?

clauderic:
@gfohl I opened a PR here to fix the release workflow https://github.com/ianstormtaylor/slate/pull/4274, waiting to hear back from @ianstormtaylor.

Even though version `0.64.0` is not tagged as `latest` on npm, you can still install it manually (for example, by running `yarn add slate-react@0.64.0` or by updating your `package.json` to `"slate-react": "^0.64.0"`)

gfohl:
Are there any usage changes needed for this release?  Running into a couple TS errors

PR #4276, opened 2021-05-21
Remove fast-deep-equal

TheSpyder:
This swaps to a custom slate-specific object deep equality check - based on the original `Text.equals` code -  that only supports nested objects, not arbitrary keys.

**Description**
Across various discussions in #3532 and #4245 the text equality method was swapped out to support nested objects. In the end, using a library-based equality check causes a regression in an undocumented feature; text nodes with marks set to undefined should be able to merge with text nodes that don't include that mark.

This PR creates a custom deep equality method to restore that behaviour while retaining the nested object comparisons. I've also refactored it into `utils` so we can look at using it in other parts of the code.

**Checks**

### DIFF
--- a/packages/slate/src/utils/deep-equal.ts
+++ b/packages/slate/src/utils/deep-equal.ts
@@ -1,0 +1,37 @@
+import isPlainObject from 'is-plain-object'
+
+/*
+  Custom deep equal comparison for Slate nodes.
+
+  We don't need general purpose deep equality;
+  Slate only suports plain values and nested objects.
+
+  Slate objects are designed to be serialised, so
+  missing keys are deliberately normalised to undefined.
+ */
+export const isDeepEqual = (
+  text: Record<string, any>,
+  another: Record<string, any>
+): boolean => {
+  for (const key in text) {
+    if (isPlainObject(text[key])) {
+      return isDeepEqual(text[key], another[key])
+    } else if (text[key] !== another[key]) {
+      return false
+    }
+  }
+
+  /*
+    Deep object equality is only necessary in one direction.
+
+    In the reverse direction we are only looking for missing keys.
+  */
+
+  for (const key in another) {
+    if (text[key] !== another[key]) {
+      return false
+    }
+  }
+
+  return true
+}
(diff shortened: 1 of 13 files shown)


spocke (approved):

packages/slate/src/utils/deep-equal.ts
```diff
@@ -0,0 +1,37 @@
+import isPlainObject from 'is-plain-object'
+
+/*
+  Custom deep equal comparison for Slate nodes.
+
+  We don't need general purpose deep equality;
+  Slate only suports plain values and nested objects.
```
clauderic, line 7: Is this documented somewhere? I only quickly looked at the docs, but didn't find anything relevant other than:

> Text nodes too can contain any custom properties you want

At the moment Slate supports merging of text nodes that have identical arrays, for example, these text nodes would be merged today.

```ts
{
  text: 'Hello',
  marks: ['bold, 'italic'],
},
{
  text: ' world',
  marks: ['bold, 'italic'],
}
```

Your change would be a breaking change for consumers that do this, which is why I'm wondering if we ever documented this.

TheSpyder: "at the moment" is relative. It's only been this way since March, when #3532 swapped this algorithm for lodash. This PR simply reverts that change and makes the old algorithm recursive (to solve the nested object requirement).

clauderic: That's a fair point, still technically a breaking change though, so perhaps best categorized as `minor`?

Could also be a good opportunity to improve documentation around this if you're up for it.

spocke: Adding array equals makes sense. The deep equals that this PR is replacing here does things we don't need though like comparing dates or regexp objects. It should be documented that the valid data types for these properties are json serializable things not other complex JS object types.

TheSpyder:
While adding Array support I realised there was a bug in my algorithm, that was lucky!

dylans (approved):

clauderic (approved):
Looks great 👍

PR #4278, opened 2021-05-21
Remove use-decorate because not work in some case(#4277)

liangxiwei:
**Description**
the useDecorate not work with editor's children
**Issue**
Fixes: (link to issue)
#4277

**Example**
no
**Context**
no
**Checks**
don't need?


liangxiwei:
I use some code to workaround with it.
`function decorate(nodeEntry: NodeEntry): Range[] {
    let decorateRange: any[] = []
    const [node,] = nodeEntry
    if (Editor.isEditor(node)) {
        const editor = node
        for (let i = 0; i < node.children.length; i++) {
            const p = path.concat(i)
            const n = node.children[i] as Descendant
            const ds = decorate([n, p])
            Array.prototype.push.apply(decorateRange, ds)
        }
        return decorateRange;
    } else {
        for (let plugin of plugins.values()) {
            if (plugin.decorate) {
                const ds = plugin.decorate(nodeEntry)
                Array.prototype.push.apply(decorateRange, ds)
            }
        }
    }
    return decorateRange
}
export { decorate }`

PR #4279, opened 2021-05-21
Remove use-decorate because not work in some case(#4277)

liangxiwei:
**Description**
useDecarote behavior not right with editor'children.
**Issue**
Fixes: (link to issue)
#4277
**Example**
I also use this code to workround it.
```
function decorate(nodeEntry: NodeEntry): Range[] {
    let decorateRange: any[] = []
    const [node,] = nodeEntry
    if (Editor.isEditor(node)) {
        const editor = node
       //editor's children can not call this method in source code,so we call ourself.
        for (let i = 0; i < node.children.length; i++) {
            const p = path.concat(i)
            const n = node.children[i] as Descendant
            const ds = decorate([n, p])
            Array.prototype.push.apply(decorateRange, ds)
        }
        return decorateRange;
    } else {
        for (let plugin of plugins.values()) {
            if (plugin.decorate) {
                const ds = plugin.decorate(nodeEntry)
                Array.prototype.push.apply(decorateRange, ds)
            }
        }
    }
    return decorateRange
}
export { decorate }
```

**Context**

**Checks**
no need


dylans (changes requested):
There's been a fair amount of work done on Children and decoration that has landed recently. It's unclear if this PR remains relevant, but it currently needs a rebase and should be checked against recent updates. @liangxiwei let me know if you want to update this, or if we should close it out due to other recent fixes in slate-react.

liangxiwei:
this PR is for https://github.com/ianstormtaylor/slate/issues/4277  
but the issue was fixed

Issue #4228, opened 2021-04-26
Cursor position is inconsistent when deleting line


churichard:
**Description**
When a line contains plaintext and it is deleted, the cursor remains on the empty line.

However, if a line contains marks or other inline formatting, then the cursor returns to the previous line.

**Recording**
![56d32da4-4318-4e10-96e1-5e78b4394728](https://user-images.githubusercontent.com/4218237/116151010-ee688880-a698-11eb-8f4f-757f7c937020.gif)

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/richtext
2. Delete the second line (which contains marks). Observe that the cursor returns to the previous line.
3. Now try deleting the fourth line (which contains only plaintext). Observe that the cursor remains on the same line.

**Expectation**
I would expect that the cursor position would be consistent between the two scenarios (ideally, that the cursor would remain on the same line).

**Environment**
- Slate Version: 0.62
- Operating System: Windows
- Browser: Chrome

liangxiwei:
if delete 'blod' string,it will ok

liangxiwei:
there has tree node in selection and finally it will call normalizeNode. And then call mergeNodes,And removeNode will be call....so the node is removed by normalizeNode.   I want to fixed but I dont know how...

https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/transforms/node.ts#L429

PR #4287, opened 2021-05-25
Do not remove empty node in merge operation if it is first children i…

liangxiwei:
…n parent(#4228)

**Description**
see issue.

**Issue**
Fixes: (#4228)

**Example**
in richtext example.select the line and delete it. see issue video pls.

**Context**
one line has bold node or other node,not just one textnode.

**Checks**

### DIFF
--- a/.changeset/merge-nodes-issue
+++ b/.changeset/merge-nodes-issue
@@ -1,0 +1,5 @@
+---
+"slate": patch
+---
+
+Do not remove empty node in merge operation if it is first children in it's parent.
--- a/.changeset/merge-nodes-issue
+++ b/.changeset/merge-nodes-issue
@@ -2,4 +2,4 @@
 "slate": patch
 ---
 
-Do not remove empty node in merge operation if it is first children in it's parent.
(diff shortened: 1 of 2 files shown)


dylans (approved):

.changeset/merge-nodes-issue
```diff
@@ -0,0 +1,5 @@
+---
+"slate": patch
+---
+
+Do not remove empty node in merge operation if it is first children in it's parent.
```
dylans, line 5:
```suggestion
Do not remove empty node in merge operation if it is first children in its parent.
```

TheSpyder:
The changeset is wrong - it was not created with `yarn changeset add`, so it's not a random name and doesn't have a `.md` extension. The bot flagged this as not having a changeset @dylans so please keep an eye on that next time :)

As for the change itself... I would've liked to see a video demonstrating the issue before accepting this. I'm not entirely convinced this is the right solution; the goal is to avoid losing the formatting of the second node (which merge is intended to do) when the first node isn't likely to have useful formatting to keep. This generally provides a better user experience.

PR #4294, opened 2021-05-26
Revert to using matrix for release workflow

clauderic:
**Description**
Setting max parallel number of jobs to 1 and updated the job order so that the `latest` job runs before the `dev` job.

The `yarn changeset publish` command deletes files in the .changesets folder, so it's not safe to run two changeset steps in the same job.
Screen Shot 2021-05-26 at 10 14 41 AM

Which means that by the time the `snapshot` release command runs, there are no changesets found and it exits early:
Screen Shot 2021-05-26 at 10 14 52 AM

**Issue**
Follows-up on #4291 and #4274

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,14 @@ on:
 
 jobs:
   release:
+    name: ${{ matrix.channel }}
     runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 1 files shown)


clauderic:
Confirmed that this is now finally working as intended

Screen Shot 2021-05-26 at 10 40 01 AM

PR #4296, opened 2021-05-27
Fix mergeNodes moving node into parent sibling

kellyjosephprice:
**Description**
This PR changes the point resolution of `remove_node` ops, such that if the previous point doesn't share a common ancestor and the next point is available, prefer the next one.

**Example**
see test cases

**Context**
I ran into this behavior when trying to write helpers that preserve cursor position. I insert empty text nodes like: `{ text: '', anchor: true }`. If a change deleted from the beginning of a line, I noticed the cursor would end up in the prior line.

**Checks**

### DIFF
--- a/packages/slate/src/transforms/general.ts
+++ b/packages/slate/src/transforms/general.ts
@@ -150,7 +150,10 @@ const applyToDraft = (editor: Editor, selection: Selection, op: Operation) => {
               }
             }
 
-            if (prev) {
+            const preferNext =
+              prev && next && Path.common(prev[1], path).length === 0
+
+            if (prev && !preferNext) {
               point.path = prev[1]
               point.offset = prev[0].text.length
             } else if (next) {
--- a/packages/slate/src/transforms/general.ts
+++ b/packages/slate/src/transforms/general.ts
@@ -150,8 +150,16 @@ const applyToDraft = (editor: Editor, selection: Selection, op: Operation) => {
               }
             }
 
-            const preferNext =
-              prev && next && Path.common(prev[1], path).length === 0
+            let preferNext = false
+            if (prev && next) {
+              if (Path.equals(next[1], path)) {
+                preferNext = !Path.hasPrevious(next[1])
+              } else {
+                preferNext =
+                  Path.common(prev[1], path).length <
+                  Path.common(next[1], path).length
+              }
+            }
 
             if (prev && !preferNext) {
               point.path = prev[1]
--- a/packages/slate/src/transforms/general.ts
+++ b/packages/slate/src/transforms/general.ts
@@ -125,12 +125,7 @@ const applyToDraft = (editor: Editor, selection: Selection, op: Operation) => {
 
     case 'remove_node': {
       const { path } = op
-      const index = path[path.length - 1]
-      const parent = Node.parent(editor, path)
-      parent.children.splice(index, 1)
 
-      // Transform all of the points in the value, but if the point was in the
-      // node that was removed we need to update the range or remove it.
       if (selection) {
         for (const [point, key] of Range.points(selection)) {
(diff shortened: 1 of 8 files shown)


dylans (approved):

packages/slate/src/transforms/general.ts
```diff
@@ -150,7 +150,10 @@ const applyToDraft = (editor: Editor, selection: Selection, op: Operation) => {
               }
             }
 
-            if (prev) {
+            const preferNext =
+              prev && next && Path.common(prev[1], path).length === 0
```
jaked, line 154: this seems too conservative; it works only when there is no common ancestor. So e.g. if you wrap another `block` around the test cases they fail; it doesn't seem like this is what you are trying to achieve.

I tried to generalize `preferNext` in a few different ways to produce the desired behavior deeper in the tree, but it caused some other tests to fail for various reasons.

I wonder if this is too low-level a place to implement selection behavior. Higher-level functions have more information about what's happening and can do a better job of leaving the selection in the expected place.

Maybe you could try to fix it in `mergeNodes`, since that's the original problem you're trying to solve?

dylans (changes requested):
@kellyjosephprice , please see feedback from @jaked when you have time. Thanks.

packages/slate/src/transforms/general.ts
kellyjosephprice: Thanks for the excellent feedback. I'll give that a go.

> I tried to generalize preferNext in a few different ways to produce the desired behavior deeper in the tree, but it caused some other tests to fail for various reasons.

I spent a minute trying to do that too. :grimacing: IIRC, some of the tests that failed as a result, seemed like 'correct' behavior. But I figured the case I was trying to fix is not 'correct'.

TheSpyder: Could you change it to prefer a sibling path? Something like

```
if (prev && Path.isSibling(prev, path)) {
...
} else if (next && Path.isSibling(next, path)) {
...
}
```
And maybe fall back to the existing behaviour if neither are siblings.

jaked: Thanks @TheSpyder for the suggestion.

Here is another attempt:
```suggestion
            let preferNext = false;
            if (prev && next) {
              if (Path.equals(next[1], path)) {
                preferNext = !Path.hasPrevious(next[1]);
              } else {
                preferNext = Path.common(prev[1], path).length < Path.common(next[1], path).length;
              }
            }
```
This passes the existing and new tests.

The main idea is to prefer the option with the longer common path, so closer to the removed node in the tree. (This generalizes the `isSibling` idea—a non-sibling has a shorter common path).

However we have to special-case the situation where `next[1]` is the same as `path` (this arises because we recompute the selection *after* removing the node, so `next[1]` is the same as `path` whenever the deleted node has following siblings). In this case, `next` always has the longer common path, which breaks some existing tests. Instead, prefer `next` only when it has no previous sibling (so `prev` is not a sibling).

This logic is hard to understand (and I'm not sure it covers every case). It might help to rewrite the function so that the selection is recomputed before actually removing the node—then we could just compare the common paths, without the special case.

By the way: I said in an earlier comment I thought this should be handled at a higher level, but after digging into it more (I ran into this bug in my own app) I think it makes sense to try to improve the default low-level behavior here. The reason is that the bug often shows up due to normalization, so it would be necessary to handle it in lots of higher-level places.

kellyjosephprice: I added some tests for the nested case. I made one attempt at factoring out the special-case (by calculating the position before hand), but I reverted it when the tests failed. Let me know if you think its worth it to try again.

dylans (approved):

PR #4299, opened 2021-05-28
Allow control of Slate's event handler execution in custom event handler API

georgberecz:
**Description**
From the [Slate Slack channel](https://slate-js.slack.com/archives/CC58ZGGU9/p1622124885032100):
We noticed that our own custom dnd behavior broke with the upgrade to 0.63 due to the changes introduced in this [PR](https://github.com/ianstormtaylor/slate/pull/4238). For some context: We use react-dnd to enable users, among other things, to drag Nodes next to each other to automatically set up column layouts. Thus, while we want to use Slate's internal dnd logic in some cases, we need to prevent the default logic in others.
The root cause of the issue lies within the `state.isDraggingInternally` property that was introduced and that is set in the `onDragStart` handler. Unfortunately, it is not possible to overwrite `onDragStart` and prevent the default logic on the Editable component without having to call `preventDefault` or `stopPropgation` on the event (which in turn would break react-dnd behavior). Otherwise, Slate's default handler is still executed.
It is limiting and confusing that the differentiator whether or not an event was handled is solely based on isDefaultPrevented or isPropagationStopped.
This PR introduces an escape hatch for custom event handlers to specify whether or not the event should be treated as handled and, therefore, further processed by Slate's own handler.

**Issue**
Fixes: issue described above

**Example**
```javascript
const onClick = useCallback(event => {
  // Implement custom event logic...

  // Not returning a value resembles the current state where isEventHandled depends on
  // preventDefault and stopPropgation
}, []);

const onDrop = useCallback(event => {
  // Implement custom event logic...

  // No matter the status of the event, treat it as being handled by returning true here
  // -> Slate will skip its own event handler
  return true;
}, []);

const onDragStart = useCallback(event => {
  // Implement custom event logic...

  // No matter the status of the event, treat event as *not* being handled by returning false
  // -> Slate will continue with its own handler
  return false;
}, []);

<Editable
    renderElement={renderElement}
    renderLeaf={renderLeaf}
    onClick={onClick}
    onDrop={onDrop}
    onDragStart={onDragStart
    {/*...*/}
/>
```

**Context**
This PR suggests the following change: Optionally, all custom event handlers can return a boolean to control the execution of Slate's default handler logic:
`void` - Currently the only supported option and there is no change in this existing behavior. (thus, no breaking change for existing Slate editors)
`true` - Explicitly treat event as handled -> Do not execute Slate's default event handler
`false` - Explicitly do not treat event as handled -> Execute Slate's default event handler

The boolean flags will take precedence over the existing `preventDefault` and `stopPropgation` checks and will, therefore, allow for more flexibility in adjusting Slate's behavior.

**Checks**

### DIFF
--- a/.changeset/chilly-gifts-dance.md
+++ b/.changeset/chilly-gifts-dance.md
@@ -1,5 +1,5 @@
 ---
-'slate-react': patch
+'slate-react': minor
 ---
 
 Allow custom event handlers on Editable component to return boolean flag to specify whether the event can be treated as being handled
--- a/.changeset/chilly-gifts-dance.md
+++ b/.changeset/chilly-gifts-dance.md
@@ -1,5 +1,0 @@
----
-'slate-react': minor
----
-
-Allow custom event handlers on Editable component to return boolean flag to specify whether the event can be treated as being handled
--- a/.changeset/chilly-gifts-dance.md
+++ b/.changeset/chilly-gifts-dance.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Allow custom event handlers on Editable component to return boolean flag to specify whether the event can be treated as being handled
--- a/docs/libraries/slate-react.md
+++ b/docs/libraries/slate-react.md
@@ -20,9 +20,11 @@ The main Slate editor.
 
 #### Event Handling
 
-By default, the `Editable` component comes with a set of event handlers that realize typical rich-text editing behavior (e.g., among others `Editable` implements its own `onCopy`, `onPaste`, `onDrop`, and `onKeyDown`). In some cases you may want to extend or overwrite Slate's default behavior, which can be done by passing your own event handler to the `Editable` component.
+By default, the `Editable` component comes with a set of event handlers that handle typical rich-text editing behaviors (for example, it implements its own `onCopy`, `onPaste`, `onDrop`, and `onKeyDown` handlers). 
 
-Your custom event handler can control whether or not Slate will execute its own event handler after yours depending on the return value of your event handler as described below.
+In some cases you may want to extend or override Slate's default behavior, which can be done by passing your own event handler(s) to the `Editable` component.
+
+Your custom event handler can control whether or not Slate should execute its own event handling for a given event after your handler runs depending on the return value of your event handler as described below.
 
 ```javascript
 const onClick = useCallback(event => {
--- a/docs/libraries/slate-react.md
+++ b/docs/libraries/slate-react.md
@@ -26,36 +26,42 @@ In some cases you may want to extend or override Slate's default behavior, which
 
 Your custom event handler can control whether or not Slate should execute its own event handling for a given event after your handler runs depending on the return value of your event handler as described below.
 
-```javascript
-const onClick = event => {
-  // Implement custom event logic...
+```jsx
+import {Editable} from 'slate-react';
 
-  // When no value is returned, Slate will execute its own event handler when neither
-  // isDefaultPrevented nor isPropagationStopped was set on the event
-};
+function MyEditor() {
+  const onClick = event => {
+    // Implement custom event logic...
 
-const onDrop = event => {
-  // Implement custom event logic...
+    // When no value is returned, Slate will execute its own event handler when
+    // neither isDefaultPrevented nor isPropagationStopped was set on the event
+  };
 
-  // No matter the state of the event, treat it as being handled by returning true here
-  // -> Slate will skip its own event handler
-  return true;
-};
+  const onDrop = event => {
+    // Implement custom event logic...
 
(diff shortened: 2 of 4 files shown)


.changeset/chilly-gifts-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+'slate-react': patch
```
clauderic, line 2: I think this should be a `minor` bump as it introduces new functionality rather than just a fix

packages/slate-react/src/components/editable.tsx
```diff
@@ -1191,14 +1191,21 @@ export const isEventHandled = <
...
+  const shouldTreatEventAsHandled = handler(event)
 
-  handler(event)
-  return event.isDefaultPrevented() || event.isPropagationStopped()
+  return (
+    (shouldTreatEventAsHandled === true ||
+      event.isDefaultPrevented() ||
+      event.isPropagationStopped()) &&
+    shouldTreatEventAsHandled !== false
+  )
```
clauderic, line 1208: This code is a bit hard to follow, could we break it down?
```suggestion
  if (shouldTreatEventAsHandled != null) {
    return shouldTreatEventAsHandled
  }
  
  return event.isDefaultPrevented() || event.isPropagationStopped()
```

clauderic (commented):
Looks good overall, just a small suggestion. 

Could you check if documentation needs to be updated around this change?

packages/slate-react/src/components/editable.tsx
georgberecz: Nice! Yeah, that's much better than what I suggested.

georgberecz:
> Looks good overall, just a small suggestion.
> 
> Could you check if documentation needs to be updated around this change?

Thanks a lot for the review @clauderic. There is not much documentation on the `Editable` component so far. I took a stab at extending the content with an event handling section that describes the behavior introduced with this PR. Let me know if that's what you had in mind.

docs/libraries/slate-react.md
```diff
@@ -18,6 +18,44 @@ React components for rendering Slate editors
 
 The main Slate editor.
 
+#### Event Handling
+
+By default, the `Editable` component comes with a set of event handlers that realize typical rich-text editing behavior (e.g., among others `Editable` implements its own `onCopy`, `onPaste`, `onDrop`, and `onKeyDown`). In some cases you may want to extend or overwrite Slate's default behavior, which can be done by passing your own event handler to the `Editable` component.
```
clauderic, line 23:
```suggestion
By default, the `Editable` component comes with a set of event handlers that handle typical rich-text editing behaviors (for example, it implements its own `onCopy`, `onPaste`, `onDrop`, and `onKeyDown` handlers). 

In some cases you may want to extend or override Slate's default behavior, which can be done by passing your own event handler(s) to the `Editable` component.
```

docs/libraries/slate-react.md
```diff
@@ -18,6 +18,44 @@ React components for rendering Slate editors
...
+
+Your custom event handler can control whether or not Slate will execute its own event handler after yours depending on the return value of your event handler as described below.
```
clauderic, line 25:
```suggestion
Your custom event handler can control whether or not Slate should execute its own event handling for a given event after your handler runs depending on the return value of your event handler as described below.
```

docs/libraries/slate-react.md
```diff
@@ -18,6 +18,44 @@ React components for rendering Slate editors
...
+
+```javascript
+const onClick = useCallback(event => {
```
clauderic, line 28: Could we remove the `useCallback` from here and below? It's not specifically relevant to what is being explained

clauderic (commented):
This is looking great, thanks for adding documentation around this! Just a few final thoughts

packages/slate-react/src/components/editable.tsx
```diff
@@ -1191,13 +1191,19 @@ export const isEventHandled = <
...
+  // shall be treated as being handled or not.
+  const shouldTreatEventAsHandled = handler(event)
 
-  handler(event)
+  if (shouldTreatEventAsHandled != null) {
+    return shouldTreatEventAsHandled
+  }
+  
   return event.isDefaultPrevented() || event.isPropagationStopped()
 }
```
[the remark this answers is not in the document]
georgberecz: Ha, very good point! I initially thought it shouldn't, but after having another look at the code I feel like it actually should behave the same way to stay consistent and not confuse people. I will adjust it.

georgberecz:
> This is looking great, thanks for adding documentation around this! Just a few final thoughts

Hey mate, thanks for the second round of feedback. I updated the PR according to your suggestions.

clauderic (approved):
Thanks for your contribution 🙏

Issue #4293, opened 2021-05-26
Backwards text typing on empty, nested, editable void Slate Editor


davidruisinger:
**Description**
When there is NO text, yet within a nested, editoable void Slate Editor (see [editable voids example](https://www.slatejs.org/examples/editable-voids), starting to type some text leads to backwards typing.

**Recording**
![khkYcJF2rp](https://user-images.githubusercontent.com/5305150/119672292-c3f62080-be3a-11eb-98a6-226d59e13524.gif)

**Sandbox**
Can be reproduced in the [editable voids example](https://www.slatejs.org/examples/editable-voids).

**Steps**
To reproduce the behavior:
1. Go to [editable voids example](https://www.slatejs.org/examples/editable-voids)
2. Delete all text within the nested Slate Editor
3. Start typing
4. See the first letter beeing shifted to the very beginning

**Expectation**
Text should be added as it is typed

**Environment**
- Slate & Slate-React Version: 0.63.0
- Operating System: MacOS
- Browser: Chrome & Safari

**Context**
This might be connected to #4275

moxxuk:
Can this ticket please be re-opened, as the fix was reverted?

angjiang:
this is also extremely problematic for us. having nested editable voids is a big part of our app but this bug makes the nested editor unusable. is there anything we can do try to normalize or force the selection while the root is being addressed?

nalband:
This is causing major issues for our product too. It totally breaks one of our core integrations.

PR #4304, opened 2021-05-31
Do NOT use exact match when updating dom selection

davidruisinger:
**Description**
When there is NO text, yet within a nested, editable void Slate Editor (see [editable voids example](https://www.slatejs.org/examples/editable-voids), starting to type some text leads to backwards typing.

**Issue**
Fixes: #4293 

**Example**
Before:
khkYcJF2rp

After:
ayoBo1dHIj

**Context**
TBD

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -170,7 +170,7 @@ export const Editable = (props: EditableProps) => {
     // If the DOM selection is in the editor and the editor selection is already correct, we're done.
     if (hasDomSelection && hasDomSelectionInEditor && selection) {
       const slateRange = ReactEditor.toSlateRange(editor, domSelection, {
-        exactMatch: true,
+        exactMatch: false,
       })
       if (slateRange && Range.equals(slateRange, selection)) {
         return


githoniel:
I think this a `ReactEditor.toSlateRange` bug.

after typing when text in empty, DOM Selection is 

image

But `ReactEditor.toSlateRange` convert it to `offset: 1`  while it should be offset: 0`

davidruisinger:
> thanks for metion
> 
> I think this is `ReactEditor.toSlateRange` problem.
> 
> after typing when text in empty, DOM Selection is
> 
> image
> 
> `ReactEditor.toSlateRange` convert it to `offset: 1` while it should be `offset: 0`, so Slate point to start after selection sync. I think We should fix it.

I tried to debug the `toSlateRange` method and I the problem seems to be within `toSlatePoint`. There is a difference between "normal" nodes and void nodes when setting the offset. 
In "normal" nodes, the offset is calculated by the length of the text of a `DocumentFragment` that is received through a created range (see [here](https://github.com/ianstormtaylor/slate/blob/f283a708fd4b7c555e1c20176b7079d54728c1c5/packages/slate-react/src/plugin/react-editor.ts#L493)).
In void nodes, the length is also taken from a text length, but this time the `domNode` itself is used (see [here](https://github.com/ianstormtaylor/slate/blob/f283a708fd4b7c555e1c20176b7079d54728c1c5/packages/slate-react/src/plugin/react-editor.ts#L506)).

Unfortunately I wasn't able to find a fix, yet. I'd really appreciate if someone else could have a look at this as well.

hueyhe:
I also ran into this issue. After debugging the `toSlatePoint`, I think the problem is caused by the unexpected element type of the anchor/focus node of the DOM selection.

[Here](https://github.com/ianstormtaylor/slate/blob/ea5e3e4ebf948264809b740f02e0fdf4bd5fead6/packages/slate-react/src/plugin/react-editor.ts#L457) in `toSlatePoint`, the `nearestNode`, which refers to the `domPoint` if exact match is used, is expected to be an html text node. However, the anchor/focus node of the DOM selection we got from [here](https://github.com/ianstormtaylor/slate/blob/ea5e3e4ebf948264809b740f02e0fdf4bd5fead6/packages/slate-react/src/components/editable.tsx#L147) isn't always a text node.

In the case below, the anchor node of the DOM selection is the editable DOM element rather than the html text node. So `toSlatePoint` can't find the corresponding slate point of the anchor node.


https://user-images.githubusercontent.com/13460383/123504840-38b4b880-d68e-11eb-873a-27ff42ac0bb9.mov

The DOM selection after deletion:

Screen Shot 2021-06-26 at 2 41 06 PM
Screen Shot 2021-06-26 at 2 46 06 PM

Things work fine outside the editable void because the leaf node will always be undefined and won't be corrected by the [void node](https://github.com/ianstormtaylor/slate/blob/ea5e3e4ebf948264809b740f02e0fdf4bd5fead6/packages/slate-react/src/plugin/react-editor.ts#L495). So the DOM selection is always updated by the slate selection since it's [out of sync](https://github.com/ianstormtaylor/slate/blob/ea5e3e4ebf948264809b740f02e0fdf4bd5fead6/packages/slate-react/src/components/editable.tsx#L191).

githoniel:
I recovered from the busy and I will fix this later

dylans (approved):

githoniel:
I will look this later.  I think we should fix `toSlateRange` instead

zbeyens:
This cursor bug makes the nested editors unusable in production, I would be in favor to merge this until we find a fix inside `toSlateRange`.

dylans:
Accepting and landing this for now, until we can come up with something better.

Issue #4330, opened 2021-06-09
Mention is not being deleted when focused


golota60:
**Description**

Focused mention is not being deleted when hitting backspace/delete button. 

This is an extension of #3456 . The issue described there also happens on safari, but the PR that closed it fixes it only on Chrome

**Recording**

![safari-mention-before](https://user-images.githubusercontent.com/37072867/121380619-502d3b00-c945-11eb-8e25-73684f201cd7.gif)

**Sandbox**
I think there's no need to add a sandbox link as it is fully reproducable in the hosted mention demo. https://www.slatejs.org/examples/mentions . 

**Steps**
To reproduce the behavior:
1. Open Safari
2. Go to https://www.slatejs.org/examples/mentions
3. Click on a mention(or focus on it using keyboard)
4. Click delete/backspace
5. Mention is not being deleted

**Expectation**
Mention should be deleted

**Environment**
- Slate Version: v0.65.0
- Operating System: macOS Catalina 10.15.7
- Browser: Safari Version 14.0.3 (15610.4.3.1.7, 15610)

PR #4331, opened 2021-06-09
fix: fix deletion in safari when inline node is selected

golota60:
**Description**

This PR fixes the deletion of inline void nodes on Safari. It's basically an extension of #4307, so it applies also in Safari.

**Issue**
Fixes: #4330 

**Example**
safari-mention-before
safari-mention-after

**Checks**


clauderic (approved):
Thanks for your contribution, this looks good 👍

PR #4347, opened 2021-06-22
Re-render leaf when new properties were added to it

aiwenar:
**Description**

Currently memoization uses `Text.matches` to compare new and previous leaf. This means that any new property added to a leaf — for example a decoration — will not cause the leaf to be re-rendered.

**Example**

Our editor supports a “tracking changes” mode. Among other things this means that when you delete text it's not removed from the editor, but marked as deleted. The problem occurs when user tries to remove an entire paragraph:

Peek 2021-06-22 15-21

while paragraph is marked as removed, this is not reflected in the UI. This is because the decoration spans whole paragraph, and so the only change is addition of a new property to the leaf, which is not considered a change by `Text.matches`. When using `Text.equals` this works fine:

Peek 2021-06-22 15-28

Note that I've tested this change on 0.59, but it should work the same on master, as leaf memoization hasn't changed. This is why I haven't checked _The relevant examples still work_ (I did run cypress tests though, and they pass).

**Checks**

### DIFF
--- a/.changeset/sixty-ties-fetch.md
+++ b/.changeset/sixty-ties-fetch.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Re-render leaf when new properties were added to it
--- a/packages/slate-react/src/components/leaf.tsx
+++ b/packages/slate-react/src/components/leaf.tsx
@@ -105,7 +105,7 @@ const MemoizedLeaf = React.memo(Leaf, (prev, next) => {
     next.renderPlaceholder === prev.renderPlaceholder &&
     next.text === prev.text &&
     next.leaf.text === prev.leaf.text &&
-    Text.matches(next.leaf, prev.leaf) &&
+    Text.equals(next.leaf, prev.leaf) &&
     next.leaf[PLACEHOLDER_SYMBOL] === prev.leaf[PLACEHOLDER_SYMBOL]
   )
 })


rockettomatooo:
Thank you @aiwenar! This is a problem we're running into as well. 

@cameracker @thesunny this is a really easy fix for a problem thats causing a lot of trouble. Is there a way we can get this into a next version fast? 🙈

rockettomatooo (approved):

thesunny:
@rockettomatooo I can do the merge but I think first you need to add a changeset so that it will automatically deploy. Ian added this recently but it's not documented (probably an important thing to add to the Slate to do list).

I found a link here [https://github.com/atlassian/changesets](https://github.com/atlassian/changesets) that Ian had posted about adding changesets.

Please take care to prep this properly. I know this has nothing to do with you, but I somehow end up being the first to do something related to work from someone else and the result is I need to spend a lot of time getting it working! Has happened on a few occasions. I understand that things can go wrong unexpectedly, but I'd appreciate the time to make sure everything is prepped properly so I can just merge and, hopefully, everything works.

The change itself looks pretty straightforward so more about doing the due diligence on the changeset. Thank you.

rockettomatooo:
@aiwenar I tried adding a changeset to your branch but I don't have permission to push my commit. Can you add it instead? It's pretty easy. As described in the link in the todo-list, just run `yarn changeset` and follow the prompts. A patch-bump should be sufficient. 
This would help us a lot!

Thanks in advance :)

aiwenar:
I've added the commit message as a patch changeset. If you would prefer it to be worded differently do let me know.

thesunny:
@aiwenar thank you!

Issue #4327, opened 2021-06-08
Proposal: Add `Element.isElementType` method


thesunny:
**Problem**
This boilerplate is common...

```ts
if (Element.isElement(node) && node.type === "paragraph") {
  // ...
}
```

It is necessary because TypeScript needs to narrow to an `Element` before safely accessing the `node.type` property.

Because it is so common, we can simplify code and reduce the amount of code by including a method that does this for us.

**Solution**
Create an `Element.isElementType` method.

```ts
if (Element.isElementType(node, 'paragraph')) {
  // ...
}
```

**Alternatives**

The alternatives are:
- Write two tests as in the "Problem" section like `Element.isElement(node) && node.type === type)`
- Each person implements their own version of the code and then imports it

**Context**
This code snippet is common and part of our tutorials.

It's a short method that will not add much to the code size and be used many times saving more than it adds.

PR #4349, opened 2021-06-23
feature(slate): Add isElementType utility function to Element Interface

imdbsd:
**Description**
Trying to add isElementType function support like what suggested in the proposal

**Issue**
#4327 

**Example**

https://user-images.githubusercontent.com/16449020/123110300-87840780-d46e-11eb-965e-3e1c9d4666bd.mp4

**Context**
Not sure if this is the correct implementation like what suggested in proposal, since node.type will has type `string` instead of the string literal like implemented in CustomType. I hope there is a suggestion about how to make it become string literal

**Checks**

### DIFF
--- a/packages/slate/src/interfaces/element.ts
+++ b/packages/slate/src/interfaces/element.ts
@@ -18,7 +18,22 @@ export interface ElementInterface {
   isElement: (value: any) => value is Element
   isElementList: (value: any) => value is Element[]
   isElementProps: (props: any) => props is Partial<Element>
+  isElementType: (
+    value: any,
+    type: string
+  ) => value is Element & { type: string }
   matches: (element: Element, props: Partial<Element>) => boolean
+}
+
+/**
+ * Shared the function with isElementType utility
+ */
+const isElement = (value: any): value is Element => {
+  return (
+    isPlainObject(value) &&
+    Node.isNodeList(value.children) &&
+    !Editor.isEditor(value)
+  )
 }
 
 export const Element: ElementInterface = {
@@ -34,14 +49,7 @@ export const Element: ElementInterface = {
    * Check if a value implements the `Element` interface.
    */
 
-  isElement(value: any): value is Element {
-    return (
-      isPlainObject(value) &&
-      Node.isNodeList(value.children) &&
-      !Editor.isEditor(value)
-    )
-  },
-
+  isElement,
   /**
    * Check if a value is an array of `Element` objects.
    */
@@ -56,6 +64,20 @@ export const Element: ElementInterface = {
 
   isElementProps(props: any): props is Partial<Element> {
     return (props as Partial<Element>).children !== undefined
+  },
+
+  /**
+   * Check if a value implements the `Element` interface and has type key
+   */
+
+  isElementType: (
+    value: any,
+    type: string
+  ): value is Element & { type: string } => {
+    if (value?.type === type && isElement(value)) {
+      return true
+    }
+    return false
   },
 
   /**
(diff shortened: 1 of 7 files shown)


imdbsd:
Also, because im share the isElement type function to it, im kinda confuse where to put the function. Under interface/shared.tsx maybe?

packages/slate/src/interfaces/element.ts
```diff
@@ -18,9 +18,24 @@ export interface ElementInterface {
   isElement: (value: any) => value is Element
   isElementList: (value: any) => value is Element[]
   isElementProps: (props: any) => props is Partial<Element>
+  isElementType: (
+    value: any,
+    type: string
+  ) => value is Element & { type: string }
```
BrentFarese, line 24: I would suggest maybe using a Generic type for this:

```suggestion
  isElementType: <T extends Element>(
    value: any,
    discriminantVal: string,
    discriminantKey?: string
  ) => value is T
```

packages/slate/src/interfaces/element.ts
```diff
@@ -58,6 +66,20 @@ export const Element: ElementInterface = {
...
+
+  isElementType: (
+    value: any,
+    type: string
+  ): value is Element & { type: string } => {
+    if (value?.type === type && isElement(value)) {
+      return true
+    }
+    return false
+  },
```
BrentFarese, line 81: Maybe do something like the below to make the func more flexible?

```suggestion
  isElementType: (
    value: any,
    discriminantVal: string,
    discriminantKey: string = 'type'
  ): value is Element & { type: string } => {
    return isElement(value) && value[discriminantKey] === discriminantVal;
  },
```

BrentFarese: The point being that not everyone uses `type` (although most people do) so maybe we make the func flexible to other [discriminants](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions).

imdbsd: haaa i see, now im understand why `type` is not in Element type

packages/slate/src/interfaces/element.ts
imdbsd: implemented, and also im adding default Extend value to `Element & {type: string}`

imdbsd:
Hi @BrentFarese thank you for reviewing this PR, im already implemented your suggestion. Please review it again.

imdbsd:
> Do you need to add `{ type: string }` here?

packages/slate/test/interfaces/Element/isElement/elementType.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+export const input = {
+  type: 'paragraph',
+  children: [{ text: '' }],
+}
+export const test = value => {
+  return Element.isElementType(value, 'heading-large')
+}
+
+export const output = true
```
BrentFarese, line 11: Shouldn't this return `false` since `'heading-large'` is not `'paragraph`'?

BrentFarese: We should probably add tests for `true` and `false` cases.

BrentFarese:
> ~im thinking about using `isElementType` without the discriminant key and the user doesn't need to add the generic for accessing the `element.type`.~
> since the purpose is to safely access `element.type` as default key, im adding `{ type: string}` as the default extends, do you have any recomendation or suggestion ?

I like what you did in your recent commit by making it the default type, but it's not required to have a `type` key.

packages/slate/test/interfaces/Element/isElement/elementType.tsx
imdbsd: My bad. Updated and adding some test for the discriminant key usage

BrentFarese (approved):
Looks good to me.

horacioh:
do we need to add documentation for this method right? cc @imdbsd

PR #4361, opened 2021-07-04
Added test for 'move-node' operation

anishagg17:
**Description**
Added test for the `move-node` operation when `path` is not equal to the `new path`.

**Checks**
~- [ ] You've [added a changeset](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md) if changing functionality. (Add one with `yarn changeset add`.)~

### DIFF
--- a/packages/slate/test/operations/move_node/path-not-equals-new-path.tsx
+++ b/packages/slate/test/operations/move_node/path-not-equals-new-path.tsx
@@ -1,0 +1,22 @@
+/** @jsx jsx */
+import { jsx } from 'slate-hyperscript'
+
+export const input = (
+  <editor>
+    <element>1</element>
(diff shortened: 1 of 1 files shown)


BrentFarese:
@anishagg17 is this test case in response to an issue/bug or just something you noticed we were lacking in test coverage?

anishagg17:
yes, @BrentFarese the codeBase was lacking in test coverage.

cboppert (approved):
Looks good to me, is there a slightly more complex case we want to test at all? Or some sort of failure case?

I'm not sure. Will leave up to @BrentFarese or @timbuckley for final approval (or another maintainer a bit more active)

timbuckley (approved):
Looks good to me

Issue #4195, opened 2021-04-12
Inconsistent cursor position after return key pressed


juliankrispel:
**Description**
By default, pressing `Enter` anywhere in a document inserts a line break and moves the cursor to the beginning of that newly inserted line.

However, doing so at the edge of a leaf node does not follow the default behaviour.

**Recording**
![linebreak-at-edge](https://user-images.githubusercontent.com/1188186/114385908-783c3000-9b88-11eb-8699-e979c1142e27.gif)

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/richtext
2. Place cursor at the end of the word "rich" (in bold, first line of text)
3. Press enter
4. Notice that the cursor remains on the same line rather than moving to the next line

**Expectation**
Placement of the cursor should be consistent regardless of where it is positioned before pressing Enter.

**Environment**
- Slate Version: Tested in 0.59 - 0.62
- Operating System: macOs
- Browser: Chrome, Safari & Firefox 

**Context**
Add any other context about the problem here. (The fastest way to have an issue fixed is to create a pull request with working, tested code and we'll help merge it. Slate is solving a pretty complex problem, and we can't do it without active contributors, so thank you so much for your help!)

baijiahei:
#4163

PR #4362, opened 2021-07-04
Replaced empty text with null character

anishagg17:
**Description**
Earlier the cursor was unable to move to empty text, now it moves to the newly created `null character`.

**Issue**
Fixes: #4195

**Example**

### Before

114385908-783c3000-9b88-11eb-8699-e979c1142e27

### After

ezgif com-gif-maker

**Checks**

### DIFF
--- a/packages/slate/src/transforms/general.ts
+++ b/packages/slate/src/transforms/general.ts
@@ -272,7 +272,7 @@ const applyToDraft = (editor: Editor, selection: Selection, op: Operation) => {
         node.text = before
         newNode = {
           ...(properties as Partial<Text>),
-          text: after,
+          text: after || '\0',
         }
       } else {
         const before = node.children.slice(0, position)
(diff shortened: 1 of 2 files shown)


packages/slate/src/transforms/general.ts
```diff
@@ -272,7 +272,7 @@ const applyToDraft = (editor: Editor, selection: Selection, op: Operation) => {
         node.text = before
         newNode = {
           ...(properties as Partial<Text>),
-          text: after,
+          text: after || '\0',
```
BrentFarese, line 275: I'm not sure we should be modifying core operations of Slate to fix this bug.

TheSpyder:
In general null characters are a bad idea in content, at best zero-width spaces should be used and even then they're only really useful for rendering.

I can no longer reproduce this in the Slate demos; it looks like enter at the end of a text node no longer splits the text node. I'm not sure if that's a core change to avoid the split or a UI change to adjust the cursor position where enter is pressed.

dylans:
Closing due to lack of interest. Happy to reopen it if there are strong arguments for this change.

Issue #4338, opened 2021-06-14
`useSelected` does not work for non-void Elements.


kellyjosephprice:
**Description**
`useSelected` works for text nodes, and non-void elements. It does not appear to update non-void elements, ie. paragraphs.

**Recording**
![use-selected](https://user-images.githubusercontent.com/451488/121966244-213b0e80-cd23-11eb-94f4-5c0f998c0767.gif)

**Sandbox**
https://jsfiddle.net/4c21gd5v/22/

**Steps**
To reproduce the behavior:
1. Go to [sandbox](https://jsfiddle.net/4c21gd5v/22/)
2. Open console
3. Select multiple lines of paragraph
4. See that neither the console nor the styling get updated with the selection

**Expectation**
I would expect `useSelected` to work for any element in the doc. Looking at [the code](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/element.tsx#L134-L139), it looks like `selection` is set and sent to the `SelectedContext.Provider` correctly:

```jsx
  return (
    <SelectedContext.Provider value={!!selection}>
      {renderElement({ attributes, children, element })}
    </SelectedContext.Provider>
  )
```

**Environment**
- Slate Version: 0.62.0
- Operating System: Linux
- Browser: Chrome

**Context**

samarsault:
From the [code](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/element.tsx#L134-L139), It seems like the Element does not have access to the SelectedContext, but all elements inside it do.

One fix maybe to wrap the SelectedContext around the Element as following as in https://github.com/ianstormtaylor/slate/pull/4365 :
```jsx
<SelectedContext.Provider value={!!sel}>
  <ElementComponent
      ...
  />
</SelectedContext.Provider>
```

kellyjosephprice:
> From the [code](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/element.tsx#L134-L139), It seems like the Element does not have access to the SelectedContext, but all elements inside it do.

I'm very confused. That's the same code I linked to, and it looks to me like the call to `renderElement` **is** wrappend in a `SelectedContext.Provider`?

> One fix maybe to wrap the SelectedContext around the Element as following as in #4365 :

But your PR seems to fix it, so that's cool. Thanks!

PR #4365, opened 2021-07-06
Capture element selections through useSelected

samarsault:
**Description**
Element selections were not captured by the `useSelected` hook

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4338

**Example (Before)**

Borrowed from https://github.com/ianstormtaylor/slate/issues/4338.
Code - https://jsfiddle.net/4c21gd5v/22/

before-selection

**Example (After)**

after-selection

**Context**
From the [code](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/element.tsx#L134-L139), It seems like the Element does not have access to the SelectedContext, but all elements inside it do. So instead of inside the Element, I wrapped the SelectionContext around the element as following:

```jsx
<SelectedContext.Provider value={!!sel}>
  <ElementComponent
      ...
  />
</SelectedContext.Provider>
```

**Checks**


cboppert (approved):
Looks good to me, will leave up to @brentfarese or @timbuckley for a final approval

timbuckley (approved):

dylans (changes requested):
Apologies @samarsault , there's a merge conflict from a PR we landed yesterday. Could you please rebase when you have a moment. Also if you could create a changeset (see the comment in the PR) then I'll go ahead and merge into main. Thanks!

dylans (approved):

Issue #3750, opened 2020-06-25
Autoscrolling not working


svenadlung:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

Related to https://github.com/ianstormtaylor/slate/issues/3677 it seems that autoscrolling doesn't work anymore since https://github.com/ianstormtaylor/slate/commit/1652ecab52088e64ffbaa130fa4049f60cdce57e (in frames?).

Please see sandbox: https://jsfiddle.net/zobnxmtL/1/

Place the cursor at the end of the visible text and write so that the cursor leaves the visible area. 

![Jun-25-2020 11-05-38](https://user-images.githubusercontent.com/11420688/85690419-d66c5500-b6d3-11ea-8c04-a9c25c99be3f.gif)

Slate: 0.58.3
Browser: Chrome
OS: Mac 

#### What's the expected behavior?

Automatic scrolling :)

thesunny:
I have the same issue and checked a few other browsers:

- Works on Safari OSX
- Works on Safari iOS
- Scroll is broken on Chrome OSX
- Scroll is broken on Firefox OSX

tmathew0309:
Works if you set the height on the Editable and then add a `scroll:auto`. Problem is if you want your scrollable container to be something separate from the Editor in which case I think https://github.com/ianstormtaylor/slate/commit/1652ecab52088e64ffbaa130fa4049f60cdce57e doesn't work.

thesunny:
Auto scrolling does not work on the sample rich text editor https://www.slatejs.org/examples/richtext which is a typical use case where the scrollable container doesn't have a height set on it.

This `scrollIntoView` call works in Chrome for scrolling the element into view.

```js
element.scrollIntoView({ behavior: "smooth", block: "nearest" }
```

This is a poor quality but working fix I made in case you need it before a fix goes into Slate proper:

```ts
      <Editable
        onSelect={(e) => {
          /**
           * Chrome doesn't scroll at bottom of the page. This fixes that.
           */
          if (!(window as any).chrome) return
          if (editor.selection == null) return
          try {
            /**
             * Need a try/catch because sometimes you get an error like:
             *
             * Error: Cannot resolve a DOM node from Slate node: {"type":"p","children":[{"text":"","by":-1,"at":-1}]}
             */
            const domPoint = ReactEditor.toDOMPoint(
              editor,
              editor.selection.focus
            )
            const node = domPoint[0]
            if (node == null) return
[...]

PR #4369, opened 2021-07-09
Fix scroll into view to scroll parent scrollables

thesunny:
**Description**
When typing at the bottom of the Editor, the editor will always scroll into view; however, the parent scrollable containers won't scroll into view.

**Issue**
Noticed this issue independently but I think it may resolve the below issue as well.

Fixes:
https://github.com/ianstormtaylor/slate/issues/3750

**Example**
See https://github.com/ianstormtaylor/slate/issues/3750

**Context**
This is a very small change that removes the `boundary` option of `scrollIntoVIew` which limits the scrolling to the editor.

```ts
      scrollIntoView(leafEl, {
        scrollMode: 'if-needed',
        // boundary: el,
      })
```

Note: I added an examples to the `examples` folder to see this in action. I kept it there so that if there were issues with the solution, the framework to look at the issue doesn't have to be recreated; however, eventually, I think it can be removed if we are satisfied with the solution.

**Checks**

### DIFF
--- a/.changeset/poor-moose-approve.md
+++ b/.changeset/poor-moose-approve.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Scroll when inserting new text will now scroll parent scrollables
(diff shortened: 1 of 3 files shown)


karthikcodes6:
Somehow this is breaking here - https://github.com/udecode/plate/issues/1046

dylans:
@thesunny scrolling with Slate has been somewhat unreliable since this change. Do you believe we should revert or are there additional things we should fix on top of this original PR?

PR #4374, opened 2021-07-13
make editor.selection avaliable when editor is contenteditable false

Xheldon:
Developer needs to access the selection in read-only mode in some cases. For example, when implementing the comment feature, the readonly mode can add comment in document, and I need to know the location of the selection in Slate in order to do the corresponding highlighting, referencing and location recording. But the current design, in read-only mode, editor.selection is always null.

And more, my understanding of read-only mode would be to allow only the editor API to modify the document while contenteditable="false" (for instance: in collaborative read-only mode), and not allow user modifications via events such as keydown, beforeinput, input, drag-and-drop input, etc. 

But I can't hide the keyboard on mobile if I prevent these events while keeping contenteditable="true" to achieve a fake "read-only".

I've tested this pr in my own project and in the readonly code in the example and it works fine.

@ianstormtaylor  Any idea?

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -513,7 +513,12 @@ export const Editable = (props: EditableProps) => {
           }
           data-slate-editor
           data-slate-node="value"
-          contentEditable={readOnly ? undefined : true}
+          // explicitly set this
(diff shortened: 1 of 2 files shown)


Xheldon:
Sorry I missed some code, will resubmit later

PR #4375, opened 2021-07-14
Feat: make editor.selection avaliable when editor is contenteditable false

Xheldon:
Developer needs to access the selection in read-only mode in some cases. For example, when implementing the comment feature, the readonly mode can add comment in document, and I need to know the location of the selection in Slate in order to do the corresponding highlighting, referencing and location recording. But the current design, in read-only mode, editor.selection is always null.

And more, my understanding of read-only mode would be to allow only the editor API to modify the document while contenteditable="false" (for instance: in collaborative read-only mode), and not allow user modifications via events such as keydown, beforeinput, input, drag-and-drop input, etc.

I can also simulate read-only by disabling the event when contenteditable="true", but then the keyboard will always appear on mobile.

I've tested this pr in my own project and in the readonly code in the example and it works fine.

@ianstormtaylor  any idea?

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -513,7 +513,12 @@ export const Editable = (props: EditableProps) => {
           }
           data-slate-editor
           data-slate-node="value"
-          contentEditable={readOnly ? undefined : true}
+          // explicitly set this
+          contentEditable={!readOnly}
+          // in some cases, if use a decoration which range are relate with selection to decorate a text node,
+          // then you will select the whole text node when you select part of text
+          // and this magic zIndex="-1" will fix it
+          zindex={-1}
           suppressContentEditableWarning
           ref={ref}
           style={{
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -414,7 +414,6 @@ export const Editable = (props: EditableProps) => {
   const onDOMSelectionChange = useCallback(
     throttle(() => {
       if (
-        !readOnly &&
         !state.isComposing &&
         !state.isUpdatingSelection &&
         !state.isDraggingInternally
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -223,9 +223,12 @@ export const ReactEditor = {
 
     return (
       targetEl.closest(`[data-slate-editor]`) === editorEl &&
-      (!editable ||
-        targetEl.isContentEditable ||
-        !!targetEl.getAttribute('data-slate-zero-width'))
+      (!editable || targetEl.isContentEditable
+        ? true
+        : (typeof targetEl.isContentEditable === 'boolean' && // isContentEditable only on HTMLElement, and other node will be undefined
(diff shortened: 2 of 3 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -513,7 +512,12 @@ export const Editable = (props: EditableProps) => {
           }
           data-slate-editor
           data-slate-node="value"
-          contentEditable={readOnly ? undefined : true}
+          // explicitly set this
+          contentEditable={!readOnly}
+          // in some cases, if use a decoration which range are relate with selection to decorate a text node,
```
dylans, line 517: I'm not able to fully understand this comment... could you please clean up the grammar, and maybe provide a link to a resource or example that explains the use of zindex=-1?

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -223,9 +223,12 @@ export const ReactEditor = {
 
     return (
       targetEl.closest(`[data-slate-editor]`) === editorEl &&
-      (!editable ||
-        targetEl.isContentEditable ||
-        !!targetEl.getAttribute('data-slate-zero-width'))
+      (!editable || targetEl.isContentEditable
+        ? true
+        : (typeof targetEl.isContentEditable === 'boolean' && // isContentEditable only on HTMLElement, and other node will be undefined
```
dylans, line 228: Please fix the grammar of this comment as I'm not quite following the explanation. Maybe run it through a grammar checker if that helps.

dylans (changes requested):

dylans (changes requested):

YaoKaiLun:
Is there any progress?

packages/slate-react/src/components/editable.tsx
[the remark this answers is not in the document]
YaoKaiLun: @Xheldon Could you follow the suggestions to modify so than the PR can be merged?

YaoKaiLun:
I really need this feature. Is there anything I can do to help? @dylans

dylans:
> I really need this feature. Is there anything I can do to help? @dylans

You could checkout the PR and raise a new PR that extends this one. That way the original person raising it retains author credit and we can land it.

packages/slate-react/src/components/editable.tsx
```diff
@@ -414,7 +414,6 @@ export const Editable = (props: EditableProps) => {
   const onDOMSelectionChange = useCallback(
     throttle(() => {
       if (
-        !readOnly &&
```
AsortKeven, line 417: @dylans In readOnly mode， if I select a void element, I'll get an error:  "Cannot resolve a Slate point from DOM point" caused by remove this line

AsortKeven: @Xheldon

Xheldon: You may need check which API you use that cause this error, sometimes the selection should be check before you use it, because it can be null sometimes @AsortKeven

dylans: Ok I’ll look into this tonight.

AsortKeven: Thanks，I just clone the project，then use the ”metions exmaple“，just change it's Readonly to false，it can reproduce this problem

PR #4383, opened 2021-07-20
added cypress test for check-list page

anishagg17:
**Description**
 
Added cypress test for check-list page.

**Checks**


anishagg17:
@timbuckley , @BrentFarese please feel free to review the PR.

cypress/integration/checkLists.ts
```diff
@@ -0,0 +1,31 @@
+describe('On check-lists example', () => {
+  it('checks the bullet when clicked', () => {
+    cy.visit('examples/check-lists')
```
timbuckley, line 3: Put `visit`s in a `beforeEach`. Every test that will be written `checkList` example will be on that page - even though there's exactly one test for now.

site/examples/check-lists.tsx
```diff
@@ -176,7 +176,7 @@ const CheckListItemElement = ({ attributes, children, element }) => {
         className={css`
           flex: 1;
           opacity: ${checked ? 0.666 : 1};
-          text-decoration: ${checked ? 'none' : 'line-through'};
+          text-decoration: ${!checked ? 'none' : 'line-through'};
```
timbuckley, line 179: Looks like you already found a bug with your test lol :)

cypress/integration/checkLists.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    cy.get('div[data-slate-node="element"]').should('have.length', 6)
```
timbuckley, line 5: You re-use the string `'div[data-slate-node="element"]'` a few times, so put it in a variable with a decent name so it's clearer what's going on in the test
```suggestion
    const slateNodeElement = 'div[data-slate-node="element"]'

    cy.get(slateNodeElement).should('have.length', 6)
```

cypress/integration/checkLists.ts
```diff
@@ -0,0 +1,31 @@
+describe('On check-lists example', () => {
```
timbuckley, line 1: Could you and @samarsault put all these tests in something like `integration/examples/`? There will be non-example-based cypress tests in the future probably, and this groups them nicely.

```suggestion
describe('Check-lists example', () => {
```

cypress/integration/checkLists.ts
```diff
@@ -0,0 +1,31 @@
...
+    cy.get('div[data-slate-node="element"]').should('have.length', 6)
+
+    cy.get('div[data-slate-node="element"]')
+      .eq(3)
+      .should('contain', 'Criss-cross!')
+      .find('span')
+      .eq(1)
+      .should('have.css', 'text-decoration-line', 'line-through')
+
+    // uncheck the checkbox and then test if text is not crossed.
```
timbuckley, line 14: Good comment on a chain of functionality that's otherwise difficult to understand.
```suggestion
    // Unchecking the checkboxes should un-cross the corresponding text.
```

site/examples/check-lists.tsx
anishagg17: yeah,

timbuckley (approved):

PR #4388, opened 2021-07-22
Update forced-layout example

LukasMurdock:
Change the forced-layout example to enforce layout to an explicit block index to allow for block types beyond 'paragraph'.

**Description**
Building from the current forced-layout example seems to prevent all types other than 'paragraph' from being used. The solution implemented here is to enforce the layout by an explicit block index. This way will not enforce subsequent blocks to be type 'paragraph.'

**Checks**
- Newbie dev, the `enforceType()` implementation smells a bit and likely needs to be refactored.
- edit: Also, I do not know what the Typescript implementation looks like

### DIFF
--- a/site/examples/forced-layout.tsx
+++ b/site/examples/forced-layout.tsx
@@ -32,6 +32,7 @@ const withLayout = editor => {
       }
 
       for (const [child, childPath] of Node.children(editor, path)) {
+        let type: string;
         const slateIndex = childPath[0];
 
         switch (slateIndex) {
--- a/site/examples/forced-layout.tsx
+++ b/site/examples/forced-layout.tsx
@@ -32,11 +32,25 @@ const withLayout = editor => {
       }
 
       for (const [child, childPath] of Node.children(editor, path)) {
-        const type = childPath[0] === 0 ? 'title' : 'paragraph'
+        const slateIndex = childPath[0];
 
-        if (SlateElement.isElement(child) && child.type !== type) {
-          const newProperties: Partial<SlateElement> = { type }
-          Transforms.setNodes(editor, newProperties, { at: childPath })
+        switch (slateIndex) {
+          case 0:
+            type = 'title';
+            enforceType(type);
+            break;
+          case 1:
+            type = 'paragraph';
+            enforceType(type);
+          default:
+            break;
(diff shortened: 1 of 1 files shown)


site/examples/forced-layout.tsx
```diff
@@ -32,11 +32,25 @@ const withLayout = editor => {
...
       for (const [child, childPath] of Node.children(editor, path)) {
-        const type = childPath[0] === 0 ? 'title' : 'paragraph'
+        const slateIndex = childPath[0];
 
-        if (SlateElement.isElement(child) && child.type !== type) {
-          const newProperties: Partial<SlateElement> = { type }
-          Transforms.setNodes(editor, newProperties, { at: childPath })
+        switch (slateIndex) {
+          case 0:
+            type = 'title';
```
dylans, line 39: type was undeclared in this PR, so this needs to either remain declared or defined with const or let.

site/examples/forced-layout.tsx
```diff
@@ -32,11 +32,25 @@ const withLayout = editor => {
...
+            enforceType(type);
+            break;
+          case 1:
+            type = 'paragraph';
```
dylans, line 43: type was undeclared in this PR, so this needs to either remain declared or defined with const or let.

dylans (changes requested):

site/examples/forced-layout.tsx
```diff
@@ -32,11 +32,25 @@ const withLayout = editor => {
       }
 
       for (const [child, childPath] of Node.children(editor, path)) {
-        const type = childPath[0] === 0 ? 'title' : 'paragraph'
+        const slateIndex = childPath[0];
```
dylans, line 35:
```suggestion
        let type: string;
        const slateIndex = childPath[0];
```

dylans (changes requested):
Suggestion to fix undeclared variable.

dylans (approved):

dylans (changes requested):
This change breaks one of the ESLint rules about where functions can be defined.

dylans:
Superseded by #4463 with these changes and an update to fix the linting rules. Thanks!

PR #4389, opened 2021-07-22
Cypress tests for examples

samarsault:
**Description**
Add cypress tests for the following examples:
- mentions
- search highlighting
- rich text
- table
- placeholder
- plain text

**Issue**
NA

**Checks**


site/examples/mentions.tsx
```diff
@@ -181,6 +182,7 @@ const Mention = ({ attributes, children, element }) => {
     <span
       {...attributes}
       contentEditable={false}
+      data-test={`mention-${element.character.replace(' ', '-')}`}
```
timbuckley, line 185: `data-cy` might be a better name for this, as its only use is to be used by cypress tests specifically
```suggestion
      data-cy={`mention-${element.character.replace(' ', '-')}`}
```

samarsault: Done

cypress/integration/mentions.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  it('shows list of mentions', () => {
+    cy.findByRole('textbox')
+      .type('{movetoend}')
+      .type(' @ma')
+      .get('[data-cy=mentions-portal]')
+      .should('exist')
+  })
+
+  it('insert on enter from list', () => {
```
timbuckley, line 20: Just keeping the grammar consistent with other tests :)
```suggestion
  it('inserts on enter from list', () => {
```

cypress/integration/search-highlighting.ts
```diff
@@ -0,0 +1,7 @@
+describe('search highlighting', () => {
+  it('text searched is highlighted', () => {
```
timbuckley, line 2:
```suggestion
  it('highlights the searched text', () => {
```

cypress/integration/tables.ts
```diff
@@ -0,0 +1,9 @@
+describe('table example', () => {
+  it('table tag rendered', () => {
+    cy.visit('examples/tables')
```
timbuckley, line 3: Put this in a `beforeEach`, as every test in this file will be for the tables example (even though there's exactly one for now. Do the same with all `visit`s in a test (unless you have to `visit` multiple pages in a given test for some reason.)

cypress/integration/search-highlighting.ts
```diff
@@ -0,0 +1,7 @@
...
+    cy.visit('examples/search-highlighting')
+    cy.get('input[type="search"]').type('text')
+    cy.get('[data-cy="search-highlighted"]').should('have.length', 2)
```
timbuckley, line 5: Put these strings in a variable to make it more obvious to readers what the test is doing. Something like...

```ts
const searchInputField = 'input[type="search"]'
const highlightedText = `[data-cy="search-highlighted"]`
```

cypress/integration/mentions.ts
```diff
@@ -0,0 +1,28 @@
+describe('mentions example', () => {
+  before(() => cy.visit('examples/mentions'))
```
anishagg17, line 2: Interesting, what's the difference between `before()` and `beforeEach()`?

timbuckley: `before` runs one time at the beginning of the entire suite (the whole `mentions.ts`). `beforeEach` runs right before every `it` block.

For that reason, I believe `beforeEach` should be used here, because I would assume `visit`ing a page again between one test and another resets the state of the page - which should make the state of each test clean (and not depend on how the previous test left things)

```suggestion
  beforeEach(() => cy.visit('examples/mentions'))
```

timbuckley (approved):
Awesome, thanks Samarjeet!

PR #4411, opened 2021-08-04
Do not needlessly JSON.stringify() the whole editor

wereHamster:
The `value` and `editor` passed to the `<Slate>` component can be big. The `useMemo()` which creates the context is invoked whenever the value changes. This means the code would JSON.stringify() both the value and editor (which through .children also contains the value), just for the `invariant()` message which may not be used most of the time.

A lazy version of `invariant` would be nice to have (eg. `invariant(false, () => "<expensive string builder>")`). But to be honest I'd prefer dumping the object into the console via console.log/info/error, because then I can actually see and inspect it.

### DIFF
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -22,12 +22,15 @@ export const Slate = (props: {
   const [key, setKey] = useState(0)
   const context: [ReactEditor] = useMemo(() => {
     if (!Node.isNodeList(value)) {
-      throw new Error(`[Slate] value is invalid! Expected a list of elements` + 
-        `but got: ${JSON.stringify(value)}`)
+      throw new Error(
+        `[Slate] value is invalid! Expected a list of elements` +
+          `but got: ${JSON.stringify(value)}`
+      )
     }
     if (!Editor.isEditor(editor)) {
-      throw new Error(`[Slate] editor is invalid! you passed:` +
-      `${JSON.stringify(editor)}`)
+      throw new Error(
+        `[Slate] editor is invalid! you passed:` + `${JSON.stringify(editor)}`
+      )
     }
     editor.children = value
     Object.assign(editor, rest)
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -22,10 +22,12 @@ export const Slate = (props: {
   const [key, setKey] = useState(0)
   const context: [ReactEditor] = useMemo(() => {
     if (!Node.isNodeList(value)) {
-      throw new Error(`[Slate] value is invalid! Expected a list of elements but got: ${JSON.stringify(value)}`)
+      throw new Error(`[Slate] value is invalid! Expected a list of elements` + 
+        `but got: ${JSON.stringify(value)}`)
     }
     if (!Editor.isEditor(editor)) {
-      throw new Error(`[Slate] editor is invalid! you passed: ${JSON.stringify(editor)}`)
+      throw new Error(`[Slate] editor is invalid! you passed:` +
+      `${JSON.stringify(editor)}`)
     }
     editor.children = value
     Object.assign(editor, rest)
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -1,7 +1,5 @@
 import React, { useMemo, useState, useCallback, useEffect } from 'react'
 import { Editor, Node, Element, Descendant } from 'slate'
-import invariant from 'tiny-invariant'
-
 import { ReactEditor } from '../plugin/react-editor'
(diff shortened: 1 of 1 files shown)


jaked:
fwiw this was a pretty recent addition in #3326

jaked:
but it does seem like the original change there https://github.com/ianstormtaylor/slate/pull/3326/commits/6e7c3f6c806d1d7883d325cd40716411b727322f was better in that it didn't stringify the objects until after testing the invariant.

and I feel like `tiny-invariant` is not doing anything useful here, and this is the only place in the codebase it's used.

bryanph:
@wereHamster might be nice to log the editor object as you mentioned

thesunny:
This can be fixed without changing the error or the causing a performance penalty by taking this:

```typescript
    invariant(
      Node.isNodeList(value),
      `[Slate] value is invalid! Expected a list of elements but got: ${JSON.stringify(
        value
      )}`
    )
    invariant(
      Editor.isEditor(editor),
      `[Slate] editor is invalid! you passed: ${JSON.stringify(editor)}`
    )
```

and replacing it with:

```typescript
    const isNodeList = Node.isNodeList(value)
    invariant(
      isNodeList,
      isNodeList ? `[Slate] value is invalid! Expected a list of elements but got: ${JSON.stringify(
        value
      )}` : ''
    )
    const isEditor = Editor.isEditor(editor)
    invariant(
      isEditor,
      isEditor ? `[Slate] editor is invalid! you passed: ${JSON.stringify(editor)}` : ''
    )
```

I'm not in actually work mode but this should satisfy all parties I think

thesunny:
Note: I think the better long term solution might be to replace `tiny-invariant` (it's just a few lines of code https://github.com/alexreardon/tiny-invariant/blob/master/src/tiny-invariant.ts) with a version that accepts a `string` or `() => string` as the second argument and only executes the function to get the string on a failure.

dylans:
@thesunny combining the things you wrote, maybe something like this:

```ts
export default function invariant(
  condition: any,
  message: string | () => string,
): asserts condition {
  if (condition) {
    return;
  }
  // Condition not passed
  throw new Error(message);
}

const isNodeList = Node.isNodeList(value)
invariant(
  isNodeList,
  // peformance optimization
  () => isNodeList ? `[Slate] value is invalid! Expected a list of elements but got: ${JSON.stringify(
    value
  )}` : ''
)

const isEditor = Editor.isEditor(editor)
invariant(
  isEditor,
  // peformance optimization
  () => isEditor ? `[Slate] editor is invalid! you passed: ${JSON.stringify(editor)}` : ''
)
```

I do see @wereHamster's point about wanting to also log or error the editor state to the console. So maybe something like this that logs (in this case I chose to log for non-production environments:

```ts
const isProduction: boolean = process?.env?.NODE_ENV === 'production';
// Throw an error if the condition fails
export default function invariant(
  condition: any,
  message: string | () => string,
  value: any,
): asserts condition {
  if (condition) {
    return;
  }
  // Condition not passed
  if (!production) {
    console.error(message, value);
  }
  throw new Error(message);
}

const isNodeList = Node.isNodeList(value)
invariant(
  isNodeList,
  // peformance optimization
  () => isNodeList ? `[Slate] value is invalid! Expected a list of elements but got: ${JSON.stringify(
    value
  )}` : ''
)

const isEditor = Editor.isEditor(editor)
invariant(
  isEditor,
  // peformance optimization
  () => isEditor ? `[Slate] editor is invalid! you passed: ${JSON.stringify(editor)}` : ''
)
```

Or some other way to decide if you prefer to log as well as throw?

TheSpyder:
```
// peformance optimization
() => isNodeList ? `[Slate] value is invalid! Expected a list of elements but got: ${JSON.stringify(
  value
)}` : ''
```

This is unnecessary - the method will only be called if the condition fails, so there's no performance to optimise.

I think what we do depends on if we see more use cases for the invariant method.

If there aren't any, maybe we return to the original style of code. Instead of a generic `invariant` method that we only use here just make a specific validation method for the two `if` conditions (passing editor and value to it).

thesunny:
Didn't realize that this is the only place `tiny-invariant` is used. In this scenario, I'd agree with @TheSpyder and remove the dependency on it since we are hacking around it anyway and do something like this:

```typescript
if (!Node.isNodeList(value)) {
  throw new Error(`[Slate] value is invalid! Expected a list of elements but got: ${JSON.stringify(value)}`)
}
if (!Editor.isEditor(editor)) {
  throw new Error(`[Slate] editor is invalid! you passed: ${JSON.stringify(editor)}`)
}
```

dylans:
@wereHamster let us know if you're happy with the proposed alternative by @thesunny . If you want to update your PR and add a changeset that would be amazing. If not, I'll do that on sometime on Monday. Thanks!

Issue #4059, opened 2021-01-27
Committing .yarnrc.yml or move to yarn 2 (pnp)


bryanph:
#### Do you want to request a _feature_ or report a _bug_?
Feature

#### What's the expected behavior?
I'm running the repository as a git submodule in a monorepo which uses yarn 2 (the only way to make this kind of setup work currently is with yarn 2 afaik because of the changed node_modules resolution behavior). 

In order to make sure slate builds properly I fix the yarn version of slate to what's in package.json (1.23.0). When I do that yarn generates a bunch of files like `.yarnrc.yml` and a `.yarn` folder which includes the binary for the yarn release mentioned in `.yarnrc.yml`. The recommended way in yarn 2 currently is to commit the binary along with a .yarnrc.yml file (for example in babel they do that here https://github.com/babel/babel/tree/main/.yarn/releases).

It would be very useful for me (and I'm sure others) to commit these files into the main repository here so that this kind of setup is supported out of the box. Alternatively we can go even further and migrate to yarn 2, projects like babel and jest are already using it and imo it is a great improvement over the non-strict behavior of node's resolution strategy. Is is also possible to use yarn 2 and still use the node_modules strategy and slowly migrate later on.

Looking for some opinions before I make a PR

TheSpyder:
Checking in yarn binaries seems to be the general recommendation now, including when migrating a project that can't run in PnP mode to Yarn 2.

PR #4417, opened 2021-08-06
Migrate yarn to berry v3

bryanph:
**Description**
Migrate to yarn v3 berry

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4059

**Context**
Yarn 2+ uses a different module resolution strategy that is much more strict in how dependencies are listed. Previously, if a dependency is not listed but incidendally included by another package, no error was thrown. Now, all dependencies must be explicitly listed in `package.json`. This makes managing dependencies generally a lot less brittle.

Note I had to fix up some dependencies from remote packages that are still on relatively old versions using `packageExtensions`. Most of these are fixed upstream in the respective packages in later versions so they can be removed once we upgrade these packages.

**Checks**


bryanph:
Looks like the CI might also need some changes to succeed? It's showing `The lockfile would have been modified by this install, which is explicitly forbidden.`

.yarnrc.yml
```diff
@@ -0,0 +1,15 @@
+packageExtensions:
+  react-error-boundary@*:
+    dependencies:
+      prop-types: '*'
```
dylans, line 4:
```suggestion
      prop-types: "*"
```
for consistency

packages/slate-react/src/components/android/use-android-input-manager.ts
```diff
@@ -17,7 +17,7 @@ export function useAndroidInputManager(node: RefObject<HTMLElement>) {
   const editor = useSlateStatic()
   const [inputManager] = useState(() => new AndroidInputManager(editor))
   const { receivedUserInput, onUserInput } = useTrackUserInput()
-  const timeoutId = useRef<NodeJS.Timeout | null>(null)
+  const timeoutId = useRef<ReturnType<typeof setTimeout> | null>(null)
```
dylans, line 20: Please verify this is intentional? Might be better to raise this as a separate PR.

dylans (changes requested):
Looking great. I have a couple of minor questions/suggestions.

packages/slate-react/src/components/android/use-android-input-manager.ts
bryanph: I was getting a `Cannot find namespace 'NodeJS'` error when running the build with yarn 3. It's trying to resolve `NodeJS` to a package it seems and can't find it. After some googling this seemed a good way to set the type for a `setTimeout` as the `NodeJS.Timeout` type is used in node environments but this is browser code.

dylans: Makes sense, thanks! It still might be better in a separate PR since it's not really part of yarn 3, up to you.

bryanph: Sounds good to me, made a separate PR here https://github.com/ianstormtaylor/slate/pull/4419

dylans (changes requested):
Rebase due to splitting the setTimeout type definition and then this looks good to me.

bryanph:
@dylans rebased :+1:

dylans (approved):
Ok, this generally looks fine to me. I'd prefer at least one other slate maintainer would take a look as it is a bigger change and I'm the newest maintainer. :)

package.json
```diff
@@ -8,7 +8,7 @@
     "build": "yarn build:rollup && yarn build:next",
     "build:next": "cd ./site && next build && next export",
     "build:rollup": "rollup --config ./config/rollup/rollup.config.js",
-    "clean": "rimraf ./node_modules ./packages/*/{dist,lib,node_modules} ./site/{.next,out}",
+    "clean": "bash -c 'rimraf ./packages/*/{dist,lib,node_modules} ./site/{.next,out}'",
```
TheSpyder, line 11: What about windows users? I guess this probably already didn't work for them unless they're in WSL.

TheSpyder (approved):
The changes are all devDependencies so it should be a safe change as long as it compiles (which CI says it does).

package.json
dylans: I had the same thought, but came to conclusion that either work on Windows without WSL.

cameracker: rimraf is a cross platform compatible rm -rf utility library, so this should have been working on windows unless there's something I'm missing.

dylans, line 11: As @cameracker reminded me, rimraf adds cross-platform support, and switch this to base breaks Windows users. Please restore this line. Thanks!

dylans (changes requested):

package.json
TheSpyder: Rimraf isn’t the problem here, it’s the bash syntax to select which folders to remove. This already didn’t work on Windows command line.

DamareYoh: I'll appeal to a bit of authority here and say that I originally wrote this command on a windows box 😆  so I'm pretty confident the old version of it works. 

`rimraf` is built on top of a `glob` parser that has direct support for this notation https://github.com/isaacs/node-glob#glob-primer

cameracker: I'll appeal to a bit of authority here and say that I originally wrote this command on a windows box 😆 so I'm pretty confident the old version of it works.

rimraf is built on top of a glob parser that has direct support for this notation https://github.com/isaacs/node-glob#glob-primer

TheSpyder: Oh 🤔 I guess it makes sense glob would support the shell syntax for this as well. Carry on 😆

merceyz: Assuming this was done because Yarn threw an error:
You need to quote the glob pattern to avoid glob expansion

bryanph: Ah i didn't realize rimraf uses its own glob parser, makes sense. Thanks, will fix it up.

bryanph:
I'm having a hard time getting eslint to resolve `slate` types correctly after rebasing. Anyone's got any ideas what might be happening?

packages/slate-history/package.json
```diff
@@ -17,11 +17,14 @@
     "is-plain-object": "^3.0.0"
   },
   "devDependencies": {
+    "@babel/runtime": "^7.7.4",
+    "lodash": "^4.17.21",
     "slate": "^0.65.3",
-    "slate-hyperscript": "^0.62.0"
+    "slate-hyperscript": "^0.62.0",
```
nemanja-tosic, line 23:
```suggestion
    "slate-hyperscript": "workspace:*",
```

packages/slate-history/package.json
```diff
@@ -17,11 +17,14 @@
     "is-plain-object": "^3.0.0"
   },
   "devDependencies": {
+    "@babel/runtime": "^7.7.4",
+    "lodash": "^4.17.21",
     "slate": "^0.65.3",
```
nemanja-tosic, line 22:
```suggestion
    "slate": "workspace:*",
```

packages/slate-hyperscript/package.json
```diff
@@ -17,10 +17,12 @@
     "is-plain-object": "^3.0.0"
   },
   "devDependencies": {
-    "slate": "^0.65.3"
+    "@babel/runtime": "^7.7.4",
+    "slate": "^0.65.3",
```
nemanja-tosic, line 21:
```suggestion
    "slate": "workspace:*",
```

packages/slate-react/package.json
```diff
@@ -24,15 +24,21 @@
...
   },
   "devDependencies": {
+    "@babel/runtime": "^7.7.4",
+    "@types/react": "^16.9.13",
+    "@types/react-dom": "^16.9.4",
     "jsdom": "^16.6.0",
+    "react": ">=16.8.0",
+    "react-dom": ">=16.8.0",
     "react-test-renderer": ">=16.8.0",
     "slate": "^0.65.3",
```
nemanja-tosic, line 34:
```suggestion
    "slate": "workspace:*",
```

packages/slate-react/package.json
```diff
@@ -24,15 +24,21 @@
...
-    "slate-hyperscript": "^0.62.0"
+    "slate-hyperscript": "^0.62.0",
```
nemanja-tosic, line 35:
```suggestion
    "slate-hyperscript": "workspace:*",
```

packages/slate/package.json
```diff
@@ -21,7 +21,10 @@
     "tiny-warning": "^1.0.3"
   },
   "devDependencies": {
-    "lodash": "^4.17.21"
+    "@babel/runtime": "^7.7.4",
+    "lodash": "^4.17.21",
+    "slate-hyperscript": "^0.62.0",
```
nemanja-tosic, line 26:
```suggestion
    "slate-hyperscript": "workspace:*",
```

nemanja-tosic (commented):
Since yarn workspaces are used, shouldn't all intra-project dependencies use the workspace protocol?

The suggestions i made here do not include the yarn.lock file, so a new lock file should be generated if this is accepted (might be simpler to just set the dependencies locally).

nemanja-tosic:
> I'm having a hard time getting eslint to resolve `slate` types correctly after rebasing (`import/named` is causing errors). Anyone's got any ideas what might be happening?

@bryanph happy to take a look at it if you were to commit the problematic rebased version.

bryanph:
@nemanja-tosic I rebased again and pushed the changes. Your suggestion makes sense, I did this initially but I reverted the change because I wasn't sure whether this would cause problems when publishing the package as they are all published separately. What do you think?

merceyz:
> Since yarn workspaces are used, shouldn't all intra-project dependencies use the workspace protocol?

Note that lerna doesn't support those yet https://github.com/lerna/lerna/pull/2450

.yarnrc.yml
```diff
@@ -0,0 +1,15 @@
+packageExtensions:
+  react-error-boundary@*:
+    dependencies:
+      prop-types: "*"
+  eslint-module-utils@*:
+    dependencies:
+      eslint-import-resolver-node: "*"
+  debug@*:
+    peerDependencies:
+      supports-color: "*"
```
merceyz, line 10: Yarn applies this one by default https://github.com/yarnpkg/berry/blob/6618f693adff26152d8e400c9dcc98b0f8ff00cc/packages/plugin-compat/sources/extensions.ts#L50-L55

package.json
```diff
@@ -63,7 +63,7 @@
     "babel-plugin-dev-expression": "^0.2.2",
     "babel-plugin-module-resolver": "^3.1.1",
     "cross-env": "6.0.3",
```
merceyz, line 65: Since Yarn uses a portable shell you shouldn't need `cross-env` anymore
https://github.com/yarnpkg/berry/tree/2036672b8c3c51bf0a9ba2552a967c17fc95d6fe/packages/yarnpkg-shell

bryanph:
Next.js also broken for me, will look into that tomorrow

TheSpyder:
I don't think we use lerna for publishing, in fact in another ticket I speculated we could probably delete it.

@bryanph yarn automatically swaps in a valid version when publishing (as long as yarn is used to publish) so it should be safe.

Looking at the docs, I think we want `workspace:^` not `*`
https://yarnpkg.com/features/workspaces#workspace-ranges-workspace

merceyz:
> Looking at the docs, I think we want workspace:^ not *

Note that was released in `3.1.0-rc.1`
```sh
yarn set version 3.1.0-rc.1
```

nemanja-tosic:
Curious about the benefits of using `workspace:^` as compared to `workspace:*` - are there any benefits to using `workspace:^`?

TheSpyder:
It controls what the published package depends on. The details are explained in the docs. Using star depends on an exact version, caret adds the caret as we use today.

Since only changed packages are published during a release the caret is essential. I guess tilde would also be acceptable?

nemanja-tosic:
They are dev dependencies so it's a moot point, i was just curious. I suppose if consistency is what is looked for (import all packages the same way), `workspace:^` is fine i suppose.

@merceyz i see you're the author of a PR for a similar change @ https://github.com/yarnpkg/berry/pull/2763, will yarn update the way they link packages, or in other words, any insider tips? :)

TheSpyder:
They may only be dev dependencies but that's how Slate is used by Slate integrators. It would still break a lot of people if we switched to exact references and didn't start releasing every package with every release.

package.json
bryanph: good one, thanks!

bryanph:
Are we ok with using release candidate versions of yarn to support this `^` notation? We can also choose to do this in a future follow-up whenever `3.1.0` is properly released?

.gitignore
```diff
@@ -15,3 +15,11 @@ cypress/videos
...
 cypress/plugins
 .DS_Store
+
+# Recommendation from https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored (not using Zero-installs)
+.yarn/*
+!.yarn/releases
+!.yarn/plugins
+!.yarn/sdks
+!.yarn/versions
+.pnp.*
```
nemanja-tosic, line 25: @bryanph as far as i can tell, the project is using zero-installs, but .gitignore is setup for non-zero installs. In case PnP was the preferred implementation, here's the updated .gitignore.

```suggestion
# Recommendation from https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored (using Zero-installs)
.yarn/*
!.yarn/cache
!.yarn/patches
!.yarn/plugins
!.yarn/releases
!.yarn/sdks
!.yarn/versions
```

merceyz:
> Will yarn update the way they link packages, or in other words, any insider tips? :)
https://github.com/ianstormtaylor/slate/pull/4417#issuecomment-903346131

We don't have any plans to add new linkers at the moment, just PnP and node_modules for now.

nemanja-tosic:
> > Will yarn update the way they link packages, or in other words, any insider tips? :)
> > [#4417 (comment)](https://github.com/ianstormtaylor/slate/pull/4417#issuecomment-903346131)
> 
> We don't have any plans to add new linkers at the moment, just PnP and node_modules for now.

My question had to do with the protocol for dev deps, IOW `workspace:*` vs `workspace:^`, as that was what the discussion was about and the linked PR used `workspace:*` for dev deps.

I don't want to restart that discussion at this point, whatever the team decides is fine with me.

merceyz:
> My question had to do with the protocol for dev deps, IOW workspace:* vs workspace:^, as that was what the discussion was about and the linked PR used workspace:* for dev deps.

Ah, that's not how I parsed that question. No plans for any change to the `devDependencies` though we'll probably switch to `workspace:^` in `dependencies` to avoid merge conflicts after releases.

.gitignore
bryanph: what do you mean we use zero-installs right now? that's a yarn berry feature :thinking:. We don't commit dependencies right now so I assumed that's what we wanted to continue doing.

bryanph:
Last blocker is making integration tests pass, some of them started failing :(

.gitignore
nemanja-tosic: Fair point, I've been thinking about pnp.

echarles:
> Also, next.js works again when upgrading to version 10, pushed those changes as well.

Just tried it with latest yarn (npm install -g yarn). `yarn install` works fine.  `yarn build` throws the following exception

```
ModuleNotFoundError: Module not found: Error: Can't resolve '/Users/echarles/datalayer/repos/slate/node_modules/next/node_modules/@babel/runtime/regenerator' in '/Users/echarles/datalayer/repos/slate/.yarn/__virtual__/next-virtual-4bb5462f15/0/cache/next-npm-10.2.3-e1de0387ed-715c83439f.zip/node_modules/next/dist/pages'


> Build error occurred
Error: > Build failed because of webpack errors
    at /Users/echarles/datalayer/repos/slate/.yarn/__virtual__/next-virtual-4bb5462f15/0/cache/next-npm-10.2.3-e1de0387ed-715c83439f.zip/node_modules/next/dist/build/index.js:17:924
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async Span.traceAsyncFn (/Users/echarles/datalayer/repos/slate/.yarn/__virtual__/next-virtual-4bb5462f15/0/cache/next-npm-10.2.3-e1de0387ed-715c83439f.zip/node_modules/next/dist/telemetry/trace/trace.js:6:584)
```

merceyz:
I just tested with a fresh clone and `yarn build` works, I'm guessing you need to clear your build cache

echarles:
> I just tested with a fresh clone and yarn build works, I'm guessing you need to clear your build cache

Thx for the quick feedback. I had cleared the repos .yarn folder. I have also now remove `~/.yarn`. I still hit the same error. Let me try tomorrow with a fresh env.

merceyz:
I meant the cache `next` creates at `site/.next`, not removing Yarn itself

echarles:
Works great after removing `site/.next` (and sure, I meant I had removed `.yarn/cache` folder, not yarn itself).

👍

dylans:
@bryanph and perhaps others, with the upgrade, I'm struggling to release the latest version of slate. If you look at NNxAMtk2epHFbbGt , you'll see that we're getting the warning: `8:12:35 AM: ➤ YN0028: │ The lockfile would have been modified by this install, which is explicitly forbidden.`

It's unclear to me if our config is wrong, or if we need to change something in our netlify environment, or something else. If anyone has any time to look at this that would be super helpful. Thanks!

echarles:
When I run 'yarn install` on a clean main branch, the `yarn.lock` is updated. Maybe that diff should be committed, and then the release be tried again?

```
git status
On branch main
Your branch is up to date with 'origin/main'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   yarn.lock
```

merceyz:
https://github.com/ianstormtaylor/slate/pull/4436 was merged without updating the lockfile, updating https://github.com/ianstormtaylor/slate/blob/2a346ded9dde640c62308bc1f575b69d64c8902b/.github/workflows/release.yml#L43 to also run `yarn install && git add .` and setting the env variable `ENABLE_IMMUTABLE_INSTALLS=false` should fix it

TheSpyder:
We were hoping to leave immutable install enabled, is that not possible in a build that also updates the version number?

or should we install _then_ update the number?

merceyz:
Only need to disable it for for the step that changes the versions

dylans:
ok, so something like this? I'm not clear what the grammar is under the `with` block

```
# https://github.com/changesets/action
      - name: Create release pull request or Publish to npm
        if: matrix.channel == 'latest'
        uses: changesets/action@master
        with:
          version: yarn changeset version  // or something like version: yarn changeset version && yarn install && git add . ?
          install: yarn install  
          add: git add .
          publish: yarn changeset publish
        env:
          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
          ENABLE_IMMUTABLE_INSTALLS: false
```

merceyz:
The `env` part looks correct but I meant to add `&& yarn install && git add .` to the line I linked to

dylans:
Updating with the correct key now.

The work from before seems to have fixed things for one release, but now it looks like the action to generate PRs with new updates is failing due to 'No commits between main and changeset-release/main', see https://github.com/ianstormtaylor/slate/runs/3590475761?check_suite_focus=true

I'll push the fix with the correct key.

merceyz:
It complains on https://github.com/ianstormtaylor/slate/runs/3590475761#step:6:21, try moving `changeset version && yarn install && git add .` to a script in `package.json#scripts` and call that instead

dylans:
> It complains on https://github.com/ianstormtaylor/slate/runs/3590475761#step:6:21, try moving `changeset version && yarn install && git add .` to a script in `package.json#scripts` and call that instead
> 
> https://github.com/changesets/action/tree/c3d88279fd9abe11410de073005e34863f735b1c#with-version-script

Thanks, giving this a try in #4516 , I appreciate the feedback and second set of eyes!

PR #4420, opened 2021-08-06
added cypress tests

anishagg17:
**Description**

Added cypress tests for :
 - editable voids
 - embeds
 - forced layout
 - hovering toolbar
 - huge document
 - images
 - markdown preview

**Checks**


cypress/integration/examples/huge-document.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  beforeEach(() => {
+    cy.visit('examples/huge-document')
+  })
+
+  it('contains image', () => {
+    elements.forEach(({ tag, count }) => {
+      cy.get(tag).should('have.length', count)
+    })
+  })
```
timbuckley, line 15: In this case, `elements` would only be relevant to the test block
```suggestion
  beforeEach(() => {
    cy.visit('examples/huge-document')
  })

  it('contains image', () => {
    const elements = [
      { tag: 'h1', count: 100 },
      { tag: 'p', count: 700 },
    ]

    elements.forEach(({ tag, count }) => {
      cy.get(tag).should('have.length', count)
    })
  })
```

timbuckley (approved):
Looks good!

cypress/integration/examples/huge-document.ts
anishagg17: But it can be used in future in different tests too.

PR #4424, opened 2021-08-08
Update 01-interfaces.md

PulkitSinghDev:
**Description**
Just made some changes in grammer of the text

### DIFF
--- a/docs/concepts/01-interfaces.md
+++ b/docs/concepts/01-interfaces.md
@@ -12,7 +12,7 @@ Which means it must have a `text` property with a string of content.
 
 But **any** other custom properties are also allowed, and completely up to you. This lets you tailor your data to your specific domain and use case, adding whatever formatting logic you'd like, without Slate getting in the way.
 
-This interface-based approach separates Slate from most other richtext editors which require you to work with their hand-rolled "model" classes, and makes it much easier to reason about. It also means that it avoids startup time penalties related to "initializing" the data model.
+This interface-based approach separates Slate from most other rich text editors which require you to work with their hand-rolled "model" classes and makes it much easier to reason about. It also means that it avoids startup time penalties related to "initializing" the data model.
 
 ## Custom Properties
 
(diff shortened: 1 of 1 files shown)


docs/concepts/01-interfaces.md
```diff
@@ -24,9 +24,9 @@ interface Element {
 }
 ```
 
-This is a very permissive interface. All it requires is that the `children` property be defined containing the element's child nodes.
+This is a very permissive interface. All it requires is that the `children` property is defined containing the element's child nodes.
```
dylans, line 27:
```suggestion
This is a very permissive interface. All it requires is that the `children` property gets defined containing the element's child nodes.
```

docs/concepts/01-interfaces.md
```diff
@@ -41,13 +41,13 @@ const link = {
 }
 ```
 
-The `type` and `url` properties there are your own custom API. Slate sees that they exist, but it doesn't ever use them for anything. However, when it goes to render a link element you'll receive an object with the custom properties attached, so that you can render it as:
+The `type` and `url` properties there are your custom API. Slate sees that they exist, but it doesn't ever use them for anything. However, when it goes to render a link element you'll receive an object with the custom properties attached so that you can render it as:
```
dylans, line 44:
```suggestion
The `type` and `url` properties are your custom API. Slate sees that they exist, but doesn't use them. However, when Slate renders a link element, you'll receive an object with the custom properties attached so that you can render it as:
```

docs/concepts/01-interfaces.md
```diff
@@ -41,13 +41,13 @@ const link = {
...
 
 ```jsx
 <a href={element.url}>{element.children}</a>
 ```
 
-When getting started with Slate, it's important to understand all of the interfaces it defines. There are a handful of them that are discussed in each of the guides.
+When getting started with Slate, it's important to understand all of the interfaces it defines. There is a handful of them that are discussed in each of the guides.
```
dylans, line 50:
```suggestion
When getting started with Slate, it's important to understand all of the interfaces it defines. There are a handful of interfaces that are discussed in each of the guides.
```

I think are is more accurate... easier to see when changing the word them to interfaces

docs/concepts/01-interfaces.md
```diff
@@ -79,11 +79,11 @@ if (Range.isCollapsed(range)) {
 }
 ```
 
-There are lots of helper functions available for all of the common use cases when working with the different interfaces. When getting started it pays to read through them all, because you can often simplify complex logic into just a handful of lines of code with them.
+There are lots of helper functions available for all of the common use cases when working with different interfaces. When getting started it pays to read through them all because you can often simplify complex logic into just a handful of lines of code with them.
```
dylans, line 82:
```suggestion
There are many helper functions available for all common use cases when working with different interfaces. When getting started it helps to read through all of them so you can often simplify complex logic into just a handful of lines of code.
```

dylans (changes requested):
Nice improvements, inspired me to make a few additions as well.

dylans (approved):

Issue #4267, opened 2021-05-18
slate-react 0.62.0+ - Crash if initial render's root node is not the document (reakit Portal)


JackRobards:
**Description**
Since the change in slate v0.62.0 from [this PR](https://github.com/ianstormtaylor/slate/pull/3749), the editor will cause a crash if the root node for it's initial render is not the document. Version `0.61.3` does not have this issue.

Notably, this happens if trying to use the slate editor in conjunction with the `reakit` `Portal`, or a similarly built component. Right now, the latest version of slate's editor does not work with the `reakit` `Portal` out of the box and trying to use them together causes a crash. In essence the problem is that if for some reason a component renders the editor before it is a child of the document, then it causes a crash.

**Recording**
Shouldn't be needed, open sandbox to see error.

**Sandbox**
https://codesandbox.io/s/slate-reproduction-reakit-portal-fk002

**Steps**
To reproduce the behavior:
1. Click on either the "Toggle Basic Portal" or "Toggle Reakit Portal" buttons.
2. See `Unable to find DocumentOrShadowRoot for editor element: [object HTMLDivElement]` error.
3. If you change the versions of `slate` and `slate-react` back to 0.61.3, then there is no error when you click either button.

**Expectation**
Ideally there would be no error thrown, similar to how it worked in previous versions. I'm not exactly sure what this error is trying to protect against though, so it may be needed. Any insight on the purpose of the thrown error though would be appreciated!

**Environment**
- Slate Version: [e.g. 0.59] 0.62.0+
- Operating System: [e.g. iOS] Mac
- Browser: [e.g. Chrome, Safari] Chrome, Firefox, Safari

**Context**
The sandbox includes a simplified version of the [reakit portal implementation](https://github.com/reakit/reakit/blob/master/packages/reakit/src/Portal/Portal.tsx) to hopefully make it a bit easier to debug and see what is going on. Basically, the portal is created using `createPortal` and then appended to the document after that. So, for the initial render it's root node is not the document and then the error is thrown causing a crash.

This issue can be worked around by delaying the rendering of the slate editor inside of the `Portal` until it's finished mounting completely for anyone that does encounter something similar.

cmditch:
Also seeing this error using Slate wrapped in a custom element within an Elm app.

hanford:
I'm also running into this exception when I have a slate editor on a page and navigate away from that page while the editor is focused

mfranzs:
[...]

PR #4427, opened 2021-08-09
Update `findDocumentOrShadowRoot` to return the root node instead of throwing unnecessarily

ben10code:
**Description**
In Slate v0.62, a new util method `getDocumentOrShadowRoot` was introduced. This latter returns the `ShadowRoot` or `Document` node in which the editor is rendered but **throws** when not found. 

`Unable to find DocumentOrShadowRoot for editor element: [object HTMLDivElement]`

This seems to be causing issues for consumers who render the editor within React portals. The issue arises at unmount time, because `getRootNode` returns the now-detached portal node. 

Version `0.61.3` does not have this issue and seems to be the go-to for most right now.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4267

**Example**
In this first example, the editor is rendered inside a modal that uses React portal. When first opening the modal, it works fine - but closing the modal (unmounting) and reopening will crash because `root` isn't an instance of `Document` or `ShadowRoot` anymore.

https://codesandbox.io/s/slate-reproduction-reakit-portal-forked-yeykl?file=/index.js:902-907

In this second example, the portal is created using `createPortal` and then appended to the document later. In the initial render, the root node is not the `Document` and throws. (Example by JackRobards)

https://codesandbox.io/s/slate-reproduction-reakit-portal-fk002

**Context**
This pull request is straightforward and shouldn't introduce unwanted behaviours. Instead of throwing when the root isn't an instance of `Document` or `ShadowRoot`, I simply return `undefined`. In places where the root node is used, I simply check whether it exists before performing operations. 

I've been trying to gather context surrounding the reasons why we throw, but couldn't find anything. Feel free to share if you have context 🙏 ! 

**Checks**

### DIFF
--- a/packages/slate-react/src/components/android/android-editable.tsx
+++ b/packages/slate-react/src/components/android/android-editable.tsx
@@ -237,40 +237,43 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
           !inputManager.isReconciling.current
         ) {
           const root = ReactEditor.findDocumentOrShadowRoot(editor)
-          if (root !== undefined) {
-            const { activeElement } = root
-            const el = ReactEditor.toDOMNode(editor, editor)
-            const domSelection = root.getSelection()
 
-            if (activeElement === el) {
-              state.latestElement = activeElement
-              IS_FOCUSED.set(editor, true)
-            } else {
-              IS_FOCUSED.delete(editor)
-            }
+          if (root === undefined) {
+            return
+          }
 
-            if (!domSelection) {
-              return Transforms.deselect(editor)
-            }
+          const { activeElement } = root
+          const el = ReactEditor.toDOMNode(editor, editor)
+          const domSelection = root.getSelection()
 
-            const { anchorNode, focusNode } = domSelection
+          if (activeElement === el) {
+            state.latestElement = activeElement
+            IS_FOCUSED.set(editor, true)
+          } else {
+            IS_FOCUSED.delete(editor)
+          }
 
-            const anchorNodeSelectable =
-              hasEditableTarget(editor, anchorNode) ||
-              isTargetInsideVoid(editor, anchorNode)
+          if (!domSelection) {
+            return Transforms.deselect(editor)
+          }
 
-            const focusNodeSelectable =
-              hasEditableTarget(editor, focusNode) ||
-              isTargetInsideVoid(editor, focusNode)
+          const { anchorNode, focusNode } = domSelection
 
-            if (anchorNodeSelectable && focusNodeSelectable) {
-              const range = ReactEditor.toSlateRange(editor, domSelection, {
-                exactMatch: false,
-              })
-              Transforms.select(editor, range)
-            } else {
-              Transforms.deselect(editor)
-            }
+          const anchorNodeSelectable =
+            hasEditableTarget(editor, anchorNode) ||
+            isTargetInsideVoid(editor, anchorNode)
+
+          const focusNodeSelectable =
+            hasEditableTarget(editor, focusNode) ||
+            isTargetInsideVoid(editor, focusNode)
+
+          if (anchorNodeSelectable && focusNodeSelectable) {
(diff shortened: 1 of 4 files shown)


dylans (commented):
Thank you @ben10code . This has very much been a problem for me as well in production and I was just about to tackle it so I really appreciate it. And the second commit really helped make this easier to review so thanks for that as well.

The general approach looks fine to me. I'm going to ask others on the Slate team for a quick review and hopefully we can land this quickly.

Also if you could add a changeset that would be great (helps us automate the release notes). You can run yarn changeset or in your branch and it will help you generate one. Thanks!

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -111,9 +113,7 @@ export const ReactEditor = {
     if (el.ownerDocument !== document) return el.ownerDocument
 
     if (!(root instanceof Document || root instanceof ShadowRoot))
-      throw new Error(
-        `Unable to find DocumentOrShadowRoot for editor element: ${el}`
-      )
+      return undefined
```
dylans, line 116: Looking back at https://github.com/ianstormtaylor/slate/pull/3749/files , I think this may just suppress things, rather than restoring support for window/window.document. Perhaps better would be:

```suggestion
      return ReactEditor.getWindow(editor).document
```

And then we wouldn't need the change to support `| undefined` throughout? Unless I'm missing something (e.g. if window wasn't actually working previously).

packages/slate-react/src/components/android/android-editable.tsx
```diff
@@ -96,7 +96,7 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
       // Make sure the DOM selection state is in sync.
       const { selection } = editor
       const root = ReactEditor.findDocumentOrShadowRoot(editor)
-      const domSelection = root.getSelection()
+      const domSelection = root ? root.getSelection() : null
```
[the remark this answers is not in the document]
dylans: Slate has started to use nullish coalescing and optional chaining so we should use them where appropriate. Assuming we don't go with my other suggestion.

packages/slate-react/src/plugin/react-editor.ts
TheSpyder: This is the correct fix given what used to happen before #3749, and after it's merged every other change (to deal with potentially undefined return value) can be reverted.

I also suggest mentioning in the changeset that this fixes a regression caused by #3749.

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -111,9 +111,7 @@ export const ReactEditor = {
     if (el.ownerDocument !== document) return el.ownerDocument
 
     if (!(root instanceof Document || root instanceof ShadowRoot))
-      throw new Error(
-        `Unable to find DocumentOrShadowRoot for editor element: ${el}`
-      )
+      return ReactEditor.getWindow(editor).document
```
clauderic, line 114: I think this be simplified to:

```suggestion
      return root.ownerDocument
```

Though to be honest, I think the entire method could be re-written, the logic is hard to follow currently.

Something like:

```ts
findDocumentOrShadowRoot(editor: ReactEditor): Document | ShadowRoot {
  const el = ReactEditor.toDOMNode(editor, editor)
  const root = el.getRootNode()

  if (root instanceof Document || root instanceof ShadowRoot) {
    return root
  }
  
  return root.ownerDocument
}
```

packages/slate-react/src/plugin/react-editor.ts
ben10code: Good to go now!

packages/slate-react/src/plugin/react-editor.ts
ben10code: Updated, indeed much cleaner! Also tested affected examples to ensure they still work :)

dylans (approved):

clauderic (approved):

PR #4423, opened 2021-08-07
prevent code highlighting test failure due to timeout

dylans:
**Description**

The introduction of a test in #4409 was not allowing sufficient time for the test to succeed. This test was causing failures in unrelated PRs such as #4037 and #3888.

**Issue**
Fixes: Failing test due to insufficient default timeout.

**Checks**
- [x] The new code matches the existing patterns and styles.
- [x] The tests pass with `yarn test`.
- [x] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [x] The relevant examples still work. (Run examples with `yarn start`.)
- [ ] You've [added a changeset](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md) if changing functionality. (Add one with `yarn changeset add`.)

No changeset as this lives outside the packages covered by changesets.

PR #4428, opened 2021-08-09
Don't set null in set_node newProperties when using unsetNodes

TheSpyder:
**Description**
Having `null` values in `set_node` operations is unnecessary, thanks to #4078, and adds complexity to collaboration transforms.

**Issue**
Follow up to #4078

**Context**
After #4078 was submitted @ianstormtaylor suggested omitting removed keys, rather than setting them to `null`, which was a much cleaner solution. This tidies up the last place that null values might be set.

**Checks**

### DIFF
--- a/.changeset/spotty-squids-walk.md
+++ b/.changeset/spotty-squids-walk.md
@@ -1,0 +1,5 @@
+---
+'slate': minor
+---
+
+Don't set `null` in `set_node`'s `newProperties` object when using `Transforms.unsetNodes()`
--- a/packages/slate/src/transforms/node.ts
+++ b/packages/slate/src/transforms/node.ts
@@ -632,19 +632,23 @@ export const NodeTransforms: NodeTransforms = {
           continue
         }
 
+        let hasChanges = false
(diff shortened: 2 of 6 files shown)


dylans (approved):
looks reasonable to me.

TheSpyder:
The CI failure seems to be an inconsistent test, it was [recently bumped to a 10s timeout](https://github.com/ianstormtaylor/slate/pull/4423) but it seems to not be enough sometimes. I suspect it's actually failing for some unexpected reason if 10 seconds isn't enough.

TheSpyder:
Given that the idea was pre-approved by Ian in #4078, I'll merge it now - easy to revert if other maintainers spot problems later.

dylans:
> Given that the idea was pre-approved by Ian in #4078, I'll merge it now - easy to revert if other maintainers spot problems later.

The failing integration test is not the result of your change, so feel free to squash & merge.

TheSpyder:
Holding for a potential patch release before this minor bump

Issue #4432, opened 2021-08-10
[android]: clipboard data application/x-slate-fragment is missing when paste on android


imdbsd:
**Description**
Im doing a copy-paste / cut-paste inline element or a text that have mark. When listening on `onCopy` or `onCut` props, event.clipboardData has `application/x-slate-fragment`. But on `onPaste` props, the `application/x-slate-fragment` is missing.

**Recording**
![bug-copy-paste-slate](https://user-images.githubusercontent.com/16449020/128813453-e71f3c2d-fab7-47c0-acd7-85b8c9277e0d.gif)

**Sandbox**
it was replicated on slate rich text example https://www.slatejs.org/examples/richtext by accessing it from android device or emulator

**Steps**
To reproduce the behavior:
1. Go to 'https://www.slatejs.org/examples/richtext'
2. select any inline element or text that has formatting
3. press copy or cut
4. paste the clipboard on editable

**Expectation**
what paste to the editable must been same with what the user copy from the slate data

**Environment**
- Slate Version: 
  - slate: 0.63.0
  - slate-react: 0.65.2
- Operating System: Android 9 - 11
- Browser: Chrome

**Context**
im still not sure why the paste event is not having the `application/x-slate-fragment` data, and im land on this question in [stackoverflow](https://stackoverflow.com/questions/32282198/clipboarddata-not-working-when-copying-text-to-content-editable-element-in-chrom). checking that prosemirror dooesnt have this issue but editorjs also facing this issue. Im thingking a workaround like having a ref to hold the same clipboardData like in onCopy and onCut. so onPaste can compare from native clipboardData or that ref

imdbsd:
if anyone land on this issue, my current workaround is listening on slate-react insertData and applied it there like this
```
const catchSlateFragment = /data-slate-fragment="(.+?)"/m;
export const getClipboardData = (dataTransfer: DataTransfer): DataTransfer => {
  /*
  * COMPAT: work  around for catch android paste buffer until 
  * this get merged https://github.com/ianstormtaylor/slate/pull/4433 or other fix applied
  */
  if (!dataTransfer.getData('application/x-slate-fragment')) {
    const htmlData = dataTransfer.getData('text/html');
    const [, fragment] = htmlData.match(catchSlateFragment) || [];
    if (fragment) {
      const clipboardData = new DataTransfer();
      dataTransfer.types.forEach(type => {
        clipboardData.setData(type, dataTransfer.getData(type));
      });
[...]

PR #4433, opened 2021-08-10
fix(android): Get fragment from text/html when application/x-slate-fragment is missing on copy-paste slate fragment in android

imdbsd:
**Description**
`application/x-slate-fragment` is missing on paste event clipbooard data. So im thingking about adding a clipboard ref for android editable to holding what user copy or cut. Then on paste event,  editor can compare the ref with the even clipboard data

**Issue**
Fixes: #4432 

**Example**
fix-bug-copy-paste-slate

**Context**
This will hold what user copy or cut in a ref, so paste event can compare it with the event clipboardData. Use it by pass the native event clipboard.

**Checks**


dylans (commented):
The change seems ok to me (I've recently spent some time working with DataTransfer). I think it would be beneficial to document why this exists, and why Android's paste buffer is excluding our attempt to add `application/x-slate-fragment`

imdbsd:
hai @dylans thanks for the feedback. im just checking it again, and make a tweak for condition props onPaste was call preventDefault. So `application/x-slate-fragment` still accessible from the props

imdbsd:
im just realize, that the cliboard data `text/html` has `data-slate-fragment` in the attribute that same as `application/x-slate-fragment`. Will change my PR to get the fragment from there instead

imdbsd:
Im just update it to check the fragment in `text/html` data and pass it to the event. Please kindly rereview it again

imdbsd:
updated fix result

fix-get-slate-fragment

packages/slate-react/src/utils/dom.ts
```diff
@@ -233,3 +233,20 @@ export const getPlainText = (domNode: DOMNode) => {
 
   return text
 }
+
+const catchSlateFragment = /data-slate-fragment="(.+?)"/m
+export const getClipboardData = (dataTransfer: DataTransfer): DataTransfer => {
+  if (!dataTransfer.getData('application/x-slate-fragment')) {
+    const htmlData = dataTransfer.getData('text/html')
```
dylans, line 240: Just so I'm clear, I think what this is doing is:

* On paste into a slate document, check to see if `application/x-slate-fragment` exists
* If true, return the dataTransfer and proceed
* If false, see if the html contains a `data-slate-fragment`.
* If that is true, convert that to a slate fragment and add it to the clipboardData and return that in place of the original DataTransfer and return. My question here is, this is now returning data only on `x-slate-fragment` instead of retaining whatever values were set for `text/html`, `text/plain`, etc., so this would likely break someone who is parsing both slate data and say plain text?
* If false, return the original dataTransfer and proceed

So I think we need to make sure that anything we return here is additive to the DataTransfer rather than just returning x-slate-fragment data. Unless I'm misunderstanding.

dylans (changes requested):

packages/slate-react/src/utils/dom.ts
dylans: Clearly I missed

```dataTransfer.types.forEach(type => {
        clipboardData.setData(type, dataTransfer.getData(type));
      });```

in my previous review. So I think we're good.

dylans (approved):

PR #4441, opened 2021-08-11
Comment out currently flaky integration test

dylans:
Even with a longer timeout, this test is intermittently failing, causing most PRs to show as failing.

**Description**
One of the recent integration tests is failing intermittently. I'm commenting it out until someone has time to address it as it's causing most PRs to fail their tests.

**Checks**

No changeset as it's just commenting out a test.

### DIFF
--- a/cypress/integration/examples/code-highlighting.ts
+++ b/cypress/integration/examples/code-highlighting.ts
@@ -34,7 +34,7 @@ describe('code highlighting', () => {
         .eq(0)
         .find(leafNode)
         .eq(0)
-        .should('contain', 'const')
+        //.should('contain', 'const')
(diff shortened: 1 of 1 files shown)


cypress/integration/examples/code-highlighting.ts
```diff
@@ -34,7 +34,7 @@ describe('code highlighting', () => {
         .eq(0)
         .find(leafNode)
         .eq(0)
-        .should('contain', 'const')
+        //.should('contain', 'const')
```
dylans, line 37:
```suggestion
        // test is failing in CI, but is not actually due to breaking behavior
        //.should('contain', 'const')
```

PR #4451, opened 2021-08-13
fix: insert new node in advance to avoid ime double input with editor marks

githoniel:
**Description**

When the user enters text using the input method, the browser will trigger `composition` event sequentially and insert  composition text in the DOM. At the end of composition, the browser will remove the previously inserted composition text and insert the final IME output text.

But the problem is, if we type with editor mark, slate will create a new node and insert the final text into the new node, which will cause the position of the inserted final text is different from the position of the the previously inserted composition text, which will cause the browser can't remove the inserted composition text correctly, and React can't detect the change too.

This will result in inconsistent data structure and DOM rendering and slate will crash when select text that shouldn't exist

**Issue**

Fixes:  #4158 #3888

**Example**

xx

**Context**

I used to force refresh all relevant string component to let `React` correct the wrong DOM node, but it led to more problems, such as the bug mentioned in [Issue] (https://github.com/ianstormtaylor/slate/pull/4158#issuecomment-886820923) that could not trigger the click event and this is hacky and looks stupid

Now I have a new way to implement it.  In `onCompositionStart`  event, we will move selection and insert a new node with mark in advance of `Editor.insertText`. To avoid nodes being deleted by `normalize` due to empty text, I set Text to '\uFEFF'  which will not be displayed in browser and remove it in `onCompositionEnd`  event, and in order to avoid unnecessary removal operations,  a new state `state.hasInsertPrefixInCompositon`  has been added.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -131,6 +131,7 @@ export const Editable = (props: EditableProps) => {
   const state = useMemo(
     () => ({
       isComposing: false,
+      hasInsertPrefixInCompositon: false,
       isDraggingInternally: false,
       isUpdatingSelection: false,
       latestElement: null as DOMElement | null,
@@ -717,6 +718,23 @@ export const Editable = (props: EditableProps) => {
                 if (!IS_SAFARI && !IS_FIREFOX_LEGACY && !IS_IOS && event.data) {
                   Editor.insertText(editor, event.data)
                 }
+
+                if (editor.selection && Range.isCollapsed(editor.selection)) {
+                  const leafPath = editor.selection.anchor.path
+                  const currentTextNode = Node.leaf(editor, leafPath)
+                  if (state.hasInsertPrefixInCompositon) {
+                    state.hasInsertPrefixInCompositon = false
+                    Editor.withoutNormalizing(editor, () => {
+                      // remove prefix added in onCompositionStart
+                      const text = currentTextNode.text.replace(/^\uFEFF/, '')
+                      Transforms.delete(editor, {
+                        distance: currentTextNode.text.length,
+                        reverse: true,
+                      })
+                      Transforms.insertText(editor, text)
+                    })
+                  }
+                }
               }
             },
             [attributes.onCompositionEnd]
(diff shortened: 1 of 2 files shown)


dylans (commented):
🎉 Thanks @githoniel , I'll review this over the weekend.

dylans:
@githoniel any thoughts on whether this might also resolve #3162?

githoniel:
@dylans no  #3162 exist even without IME

just go example https://www.slatejs.org/examples/search-highlighting

1. input editable in search input
2. input anything after editable highlight

I used to skip this problem in my own editor by  disable highlight when any editor value change when I using decorator a half year ago

packages/slate-react/src/components/editable.tsx
```diff
@@ -739,9 +757,42 @@ export const Editable = (props: EditableProps) => {
...
                 !isEventHandled(event, attributes.onCompositionStart)
               ) {
-                const { selection } = editor
-                if (selection && Range.isExpanded(selection)) {
-                  Editor.deleteFragment(editor)
+                const { selection, marks } = editor
+                if (selection) {
+                  if (Range.isExpanded(selection)) {
+                    Editor.deleteFragment(editor)
+                  } else {
```
dylans, line 764: I think in general the Slate preference would be to add a return statement inside the preceding if block rather than using an else.

packages/slate-react/src/components/editable.tsx
```diff
@@ -739,9 +757,42 @@ export const Editable = (props: EditableProps) => {
...
+                        const point = Editor.after(editor, inlinePath)!
+                        Transforms.setSelection(editor, {
+                          anchor: point,
+                          focus: point,
+                        })
+                      }
+                    }
+                    // insert new node in advance to ensure composition text will insert
+                    // along with final input text
+                    // and add prefix text to avoid this node removed by normalize
```
dylans, line 781:
```suggestion
                    // add Unicode BOM prefix to avoid normalize removing this node
```

packages/slate-react/src/components/editable.tsx
```diff
@@ -717,6 +718,23 @@ export const Editable = (props: EditableProps) => {
...
                   Editor.insertText(editor, event.data)
                 }
+
+                if (editor.selection && Range.isCollapsed(editor.selection)) {
+                  const leafPath = editor.selection.anchor.path
+                  const currentTextNode = Node.leaf(editor, leafPath)
+                  if (state.hasInsertPrefixInCompositon) {
+                    state.hasInsertPrefixInCompositon = false
+                    Editor.withoutNormalizing(editor, () => {
+                      // remove prefix added in onCompositionStart
```
dylans, line 728:
```suggestion
                      // remove Unicode BOM prefix added in `onCompositionStart`
```

dylans (changes requested):
minor suggestions, then I think we can land this. Thanks @githoniel

dylans (approved):

Jokcy:
Greate job, a nice solution which I never thought 👍

Issue #4217, opened 2021-04-21
MacOS. Safari. Line breaks disappear when inserting paragraphs


fredure:
**Description**
When pasting several paragraphs from the buffer, line breaks disappear.

**Recording**
![Apr-21-2021 14-51-19](https://user-images.githubusercontent.com/8024592/115549406-4c6c2900-a2b1-11eb-87f5-95e95c6ef0e6.gif)

**Sandbox**
https://www.slatejs.org/examples/richtext

**Steps**
To reproduce the behavior:
1. Enter 2 lines of text
2. Select and cut text
3. Paste text

**Expectation**
Expect behavior similar to google chrome - line breaks should persist.

**Environment**
- Slate Version: 0.62.0
- Operating System: macOS BigSur 11.2.2 (20D80)
- Browser: Safari 14.0.3 (16610.4.3.1.4)

andrew-culver:
Confirm this behavior.

Slate Version: 0.59.0
Operating System: macOS Catalina 10.15.7
Browser: Safari 14.0.3

simonuvarov:
Can confirm.
Slate 0.62.1
macOS Big Sur 11.2.3
Safari 14.03

fredure:
Any styles are not saved. Safari probably doesn't support custom types  in dataTransfer.

simonuvarov:
> Any styles are not saved. Safari probably doesn't support custom types in dataTransfer.

Do styles apply to new lines, too? Looks more like `\n` or `\r\n` characters are processed incorrectly.

fredure:
Looks line a take https://github.com/ianstormtaylor/slate/issues/3589.

fredure:
fixed

Nikhil22:
This is till an issue for me. Line breaks not honored when pasting into Safari.

Issue #3589, opened 2020-04-06
Copying and pasting decorated text and elements doesn't work on Safari


bibixx:
#### Do you want to request a _feature_ or report a _bug_?
A bug

#### What's the current behavior?
When trying to copy and paste decorated text or elements from slate input the pasted value is always a string. The issue happens both in my application as well as in the [Slate examples](https://www.slatejs.org/examples/mentions)

Correct me if I'm wrong but I couldn't find any issue that addresses this bug (or maybe Safari's limitations?). It's strange though as it should affect quite a lot of users

**Decorators**
![ezgif-7-3c9347953fd7](https://user-images.githubusercontent.com/14880213/78546179-4538c100-77fd-11ea-8933-1770901a2d48.gif)

**Elements**
![ezgif-7-5ddb2d700fff](https://user-images.githubusercontent.com/14880213/78546131-34884b00-77fd-11ea-9f4a-ce59e7b116a8.gif)

Slate: 0.57.1
Browser: Safari 13.0.5 (15608.5.11)
OS: MacOS 10.15.3 (19D76)

#### What's the expected behavior?
It should work the same as in the other browsers – the decorated text / elements should be pasted as they were copied

williamstein:
I happened to be reimplementing the clipboard for slate in https://cocalc.com, and noticed this issue and have some remarks that might be useful to whoever fixes it for slate (not me).  In the file slate-react/plugin/with-react.ts try changing insertData as given below.   Anyway, reading through relevant code, there is a sort of half-way implementation of paste, which is there for no apparent reason and not used.   Maybe it was used in a previous version of slate and was unfinished in the rewrite, due to chrome/firefox letting you set any data on the DataTransfer object, and not realizing that Safari evidently does not allow doing this.      

This code below is just an ugly hack, and definitely not what I'm actually doing in my code.  But if you try it out, it'll mostly work and from there you can maybe do something much better than silently failing completely on Safari (or maybe other browsers too).     Incidentally, searching around suggests that Safari revamped their DataTransfer implementation in response to a ton of security concerns about a year ago, so maybe slate paste used to work in safari, but broke about a year ago...

```
  e.insertData = (data: DataTransfer) => {
    let fragment = data.getData("application/x-slate-fragment");

    if (fragment == null) {
      // On Safari (probably for security reasons?), the
      // application/x-slate-fragment data is not set.
      // My guess is this is why the html is also modified
      // even though it is never used in the upstream code!
[...]

PR #4454, opened 2021-08-13
Read data-slate-fragment when application/x-slate-fragment is missing

imdbsd:
**Description**
From what im understand, the mutation should be detected from beforeInput event and only using onPaste as the fallback. But im found out, window.DataTransfer is not holding the slate-fragment data. For now, im only found out it missing when using safari on mac (safari version 14).

**Issue**
#3589, #4217

**Example**
fix-slate-fragment

**Context**
When try to debug this, im found out the application/x-slate-fragment is exist in onPaste clipboard but not in window data transfer that used in beforeInput event. Since `data-slate-fragment` also hold the same data,  i think its better to check to it as the fallback when `application/x-slate-fragment ` is missing. This also what i did in #4433 for android. Im using safari 14 and macos catalina.

**Checks**

### DIFF
--- a/.changeset/nervous-planes-wonder.md
+++ b/.changeset/nervous-planes-wonder.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fix to read fragment from data-slate-fragment when application/x-slate-fragment is missing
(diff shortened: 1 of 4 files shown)


dylans (approved):
Great timing, I was just starting to look into what needed to get done to make this work on Safari!

PR #4455, opened 2021-08-16
Fix(editable component): reselect the range created by triple click

### LINKED ISSUE #4329: Triple-click selection and Selection API issue
**Description**
Hello! We are currently facing a bug with triple-click selection. When we triple-click on a word, the expectation is that the whole text block gets selected, a paragraph for example. The problem here is that for 2 of the 3 browsers I tested (it happens on Chrome and Edge, not on Firefox), Slate will also select the following (or part of the following) sibling node.

You can see recordings for both behaviours, the first for Chrome (bug) and the second on Firefox (no bug).

**Recording**

https://user-images.githubusercontent.com/29894873/121357512-29193e00-c932-11eb-91d9-54da429cea47.mov

https://user-images.githubusercontent.com/29894873/121357551-2e768880-c932-11eb-8bc7-62b05be9e5a2.mov

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/images
2. Triple-click on the first paragraph
3. On Chrome the image gets selected as well, not on Firefox

**Expectation**
The expectation would be to have the selection only on the text block that was actually targeted, and not the following one.

**Environment**
- Slate Version: latest
- Operating System: macOS 10.15.7
- Browser: Chrome, Edge

**Context**
From my understanding, there is two things that feel are part of the issue.

1. The first might be because of the way `offsets` are considered within a selection. 
At this line: 
https://github.com/ianstormtaylor/slate/blob/e042ebd4a2539aeabf0996e8f1a1dad1be5860c0/packages/slate-react/src/utils/dom.ts#L110
`offset` is compared to `childNodes.length`
But according to the [Selection API](https://developer.mozilla.org/en-US/docs/Web/API/Selection/focusOffset) the offset represents the number of characters selected, not the number of child nodes, so this is not comparing the same things at all.

2. The second  thing is, if the value of `focusOffset` is 0, it means there is nothing selected in the `focusNode` and therefore, this node should be removed from the selection range. Currently, maybe because of the first point, Slate keeps digging into this `focusNode` to find a selectable child and include it in the range. On top of it, if `focusNode` is a `void` element, it can create more problems, as Slate is not supposed to handle the content of a `void` element.

bytrangle:
**Description**
When triple clicking inside the editable component of Slate, it doesn't select the correct block. This poses two problems when Slate is to be used as a rich text editor:
- Give false visual feedback, e.g. the wrong toolbar button will be highlighted which leads to confusion
- Unable to apply new style to the selected block only.

This PR tries to select the correct block that users triple click on.

**Issue**
Fixes #4329 

**Example**

Before the fix:

triple-click-before

After the fix:

triple-click-after

**Context**

Given a node at path [1].

```js
const editor = {
  children: [
    { // first node
    }
    {
      type: 'paragraph', // the node that receives triple click
      children: [
        { text: "Since it's rich text, you can do things like turn a selection of text " }, // path: [1, 0]
        { text: "bold", bold: true }, // path: [1, 1]
        { text: ", or add a semantically rendered block quote in the middle of the page, like this:" // path: [1, 2]
      ]
    }
  ]
}
```
Suppose we triple click just before the letter "S" of the paragraph.

- Inside the `onClick` event check if there is triple click with `event.detail === 3`.
- By design, the node that users click on will be the text node at path [1, 0]. Call `Editor.above()` to retrieve the closest element ancestor of that text node.
- If the element exists, get all the children of the element. Create an array of their respective path. It should be
[[1, 0], [1, 1], [1, 2]]
- Create a range whose anchor is the first child and focus is the third child. Its anchor offset is 0 (because the selection starts from the very start of the block) and focus offset is the end point of the third child (82) in this case.
- Calling `Transforms.select(range)` to select that range.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -624,52 +624,19 @@ export const Editable = (props: EditableProps) => {
               ) {
                 const node = ReactEditor.toSlateNode(editor, event.target)
                 const path = ReactEditor.findPath(editor, node)
-                // Check if user clicks thrice
-                if (event.detail === 3) {
-                  // Check the closest element block above the selection
-                  const elem = Editor.above(editor, {
-                    at: path,
-                    match: n => Editor.isBlock(editor, n),
-                  })
-                  if (elem) {
-                    // Get all the children of the element above
-                    const children = Node.children(editor, elem[1])
-                    const childrenPath = []
-                    // Loop through the generator of element children and create a list of their respective paths.
-                    for (const [, childPath] of children) {
-                      childrenPath.push(childPath)
-                    }
-                    // Triple-click selection is equal to click and drag from the begining to the end of a block.
-                    // Its anchor will be the first child and focus being the last child of the block.
-                    const anchorPath = childrenPath[0]
-                    const focusPath = childrenPath[childrenPath.length - 1]
-                    // In a triple-click selection, the anchor offset will always be 0 and focus offset will be the end point of the last child of the block.
-                    const { offset: focusOffset } = Editor.end(
-                      editor,
-                      focusPath
-                    )
-                    const range = Editor.range(
-                      editor,
-                      { path: anchorPath, offset: 0 },
-                      { path: focusPath, offset: focusOffset }
-                    )
-                    Transforms.select(editor, range)
-                  }
-                } else {
-                  const start = Editor.start(editor, path)
-                  const end = Editor.end(editor, path)
+                const start = Editor.start(editor, path)
+                const end = Editor.end(editor, path)
 
-                  const startVoid = Editor.void(editor, { at: start })
-                  const endVoid = Editor.void(editor, { at: end })
+                const startVoid = Editor.void(editor, { at: start })
+                const endVoid = Editor.void(editor, { at: end })
 
-                  if (
-                    startVoid &&
-                    endVoid &&
-                    Path.equals(startVoid[1], endVoid[1])
-                  ) {
-                    const range = Editor.range(editor, start)
-                    Transforms.select(editor, range)
-                  }
+                if (
+                  startVoid &&
+                  endVoid &&
+                  Path.equals(startVoid[1], endVoid[1])
+                ) {
+                  const range = Editor.range(editor, start)
+                  Transforms.select(editor, range)
                 }
               }
             },
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -624,19 +624,52 @@ export const Editable = (props: EditableProps) => {
               ) {
                 const node = ReactEditor.toSlateNode(editor, event.target)
                 const path = ReactEditor.findPath(editor, node)
-                const start = Editor.start(editor, path)
-                const end = Editor.end(editor, path)
+                // Check if user clicks thrice
+                if (event.detail === 3) {
+                  // Check the closest element block above the selection
+                  const elem = Editor.above(editor, {
+                    at: path,
+                    match: n => Editor.isBlock(editor, n),
+                  })
+                  if (elem) {
+                    // Get all the children of the element above
+                    const children = Node.children(editor, elem[1])
+                    const childrenPath = []
+                    // Loop through the generator of element children and create a list of their respective paths.
+                    for (const [, childPath] of children) {
+                      childrenPath.push(childPath)
+                    }
+                    // Triple-click selection is equal to click and drag from the begining to the end of a block.
+                    // Its anchor will be the first child and focus being the last child of the block.
+                    const anchorPath = childrenPath[0]
+                    const focusPath = childrenPath[childrenPath.length - 1]
+                    // In a triple-click selection, the anchor offset will always be 0 and focus offset will be the end point of the last child of the block.
+                    const { offset: focusOffset } = Editor.end(
+                      editor,
+                      focusPath
+                    )
+                    const range = Editor.range(
+                      editor,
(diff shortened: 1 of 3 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -624,19 +624,52 @@ export const Editable = (props: EditableProps) => {
               ) {
                 const node = ReactEditor.toSlateNode(editor, event.target)
                 const path = ReactEditor.findPath(editor, node)
-                const start = Editor.start(editor, path)
-                const end = Editor.end(editor, path)
+                // Check if user clicks thrice
+                if (event.detail === 3) {
+                  // Check the closest element block above the selection
+                  const elem = Editor.above(editor, {
```
dylans, line 630: Might be clearer if it was `const block` rather than `const elem`, since it's not returning a match if it's not a block?

packages/slate-react/src/components/editable.tsx
```diff
@@ -624,19 +624,52 @@ export const Editable = (props: EditableProps) => {
...
+                    at: path,
+                    match: n => Editor.isBlock(editor, n),
+                  })
+                  if (elem) {
+                    // Get all the children of the element above
```
dylans, line 635:
```suggestion
                    // Get all the children of the block element above the selection
```

packages/slate-react/src/components/editable.tsx
```diff
@@ -624,19 +624,52 @@ export const Editable = (props: EditableProps) => {
...
+                    const children = Node.children(editor, elem[1])
+                    const childrenPath = []
+                    // Loop through the generator of element children and create a list of their respective paths.
```
dylans, line 638: Please run the linter on this file, this comment and a few below likely exceed the line length rules.

dylans (changes requested):
Thanks, this change seems reasonable to me. I've made a few minor suggestions.

bytrangle:
> Thanks, this change seems reasonable to me. I've made a few minor suggestions.

Thank you, I will fix them tomorrow.

But I've noticed that when I triple click on the second paragraph of the **Rich Text** example, the Quote Block button is highlighted for a split second and then it's not highlighted.

I think it's because the `useLayoutEffect` is synchronous and run before the `onClick` event handler. The former will try to find the selected DOM range (incorrectly), which makes the Quote Block highlighted, then the `onClick` event handler kicks in.

Do you experience the same?

bytrangle:
Hello @dylans ,
I've discovered that when I triple click on a block, the `Editor.nodes()` method is supposed to grab a generator of nodes at the current selection, but somehow it always includes the node below the clicked node.

next-sibling-node-selected-when-triple-click

I've traced the problem down to Slate's [`transform` operator](https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/transforms/general.ts#L318).

This is what happens on the third click:

third-click-transform

Between `createDraft` and `finishDraft`, the `selection` object has been reassigned:

```
transform(editor, op): {
  // ...
  let selection = editor.selection && createDraft(editor.selection)
  try {
    selection = applyToDraft(editor, selection, op)
  } finally {
      editor.children = finishDraft(editor.children)
      if (selection) {
        editor.selection = isDraft(selection)
          ? (finishDraft(selection) as Range)
          : selection
    }
  }
}
```
I think the culprit is probably the helper function `applyToDraft()`, so please allow me a few more days to figure it out.

dylans:
@bytrangle thanks, I appreciate you digging into this further.

xenostar:
Just wanted to chime in and say I had some time to test this fix today and it is looking really good. This problem has been plaguing Slate for along time and I'm really looking forward to this fix!

packages/slate-react/src/components/editable.tsx
```diff
@@ -624,19 +624,52 @@ export const Editable = (props: EditableProps) => {
               ) {
                 const node = ReactEditor.toSlateNode(editor, event.target)
                 const path = ReactEditor.findPath(editor, node)
-                const start = Editor.start(editor, path)
-                const end = Editor.end(editor, path)
+                // Check if user clicks thrice
+                if (event.detail === 3) {
```
xenostar, line 628: Have you thought about setting this to `event.detail >= 3` instead? I noticed when trying a similar approach to this fix locally that if you spam click many times, `event.detail` continues to increment so this wouldn't be fired once you go past 3 clicks.

dylans:
@bytrangle to clarify, is this really only an issue in Chrome/Chromium-based browsers? I ask as it is a very different PR from your original. Thanks!

dylans:
> @dylans Thank you, and actually that's a good call because I've written an e2e test for this triple-click behavior.
> 
> It passed, but it opened up a torrent of Typescript warnings. I've tried to fix it as I can, but since I've never used Typescript and just learn as I go, please take a look and tell me what you think.

Your TS seems reasonable. Maybe there are ways to make it less messy, but I think it's fine. Landing this now.

dylans:
@bytrangle This appears to have caused a regression in Chrome... when you triple check on a link with marks (e.g. bold or italic) and press the backspace/delete key, it only deletes back to the point of the most recent mark.

The example in examples/richtext , triple click on the first line and press backspace to reproduce.

TheSpyder:
Looking at the stack trace generated by #4490 the use of `!` on values that might be null is to blame. I haven't been paying enough attention to PRs, sorry, I would've flagged that as a concern.

bytrangle:
@dylans  You mean this?

https://user-images.githubusercontent.com/28924121/132483273-c2630240-1b14-43a8-aab1-750fac7edc10.mp4

PR #4458, opened 2021-08-18
Normalization: Nonempty to empty to inline should remove empty node

taj-codaio:
**Description**
As the code exists today, if you have an empty text node before an inline node, the empty text node will not be removed if it has a nonempty node before it with differing attributes.

The issue is that we consider pairs of text nodes, and generally only remove the earlier node if that node is empty. The only time we remove the current node is if it is the last node in the block. This fails to behave correctly when the current node is the last node before an inline object, and leaves the empty text node lingering.

This PR aims to fix that edge case.

**Example**
This is just a copy/paste of the test case, but the test really exemplifies this better than anything else I can think of:

As things stand today, this input:
```
<editor>
  <block>
    <text>not empty</text>
    <text a/>
    <inline>inline</inline>
    <text />
  </block>
</editor>
```
Will leave the `<text a/>` node lingering after normalization.

With the fix, the normalization will remove the `<text a/>` node and produce:
```
<editor>
  <block>
    <text>not empty</text>
    <inline>inline</inline>
    <text />
  </block>
</editor>
```

**Checks**

### DIFF
--- a/.changeset/seven-fishes-check.md
+++ b/.changeset/seven-fishes-check.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Normalization now removes empty text nodes after nonempty nodes with differing styles, but before inline nodes.
--- a/packages/slate/src/create-editor.ts
+++ b/packages/slate/src/create-editor.ts
@@ -273,7 +273,7 @@ export const createEditor = (): Editor => {
                 voids: true,
               })
               n--
-            } else if (isLast && child.text === '') {
+            } else if (child.text === '') {
               Transforms.removeNodes(editor, {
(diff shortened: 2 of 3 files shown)


taj-codaio:
Oh, my apologies - I thought lint had completed, but I was incorrect. I need to make a lint fix.

taj-codaio:
Lint is fixed now.

dylans (approved):
Thanks for the PR @taj-codaio . It's certainly a simple change, and all the tests still pass, so I'm inclined to approve and land this. I'll leave it open for a day or two in case others have feedback that differs from mine.

jason-codaio (approved):

Issue #3816, opened 2020-08-06
Selection is not placed correctly after inserting mutliple nodes with Transforms.insertNodes


aglundahl:
#### Do you want to request a _feature_ or report a _bug_?

Bug.

#### What's the current behavior?

The following code will place the cursor between `foo` and `bar`.
 
```javascript
const editor = (
  <editor>
    <block>
      <text />
      <cursor />
    </block>
  </editor>
)
Transforms.insertNodes(editor, [<text>foo</text>, <text>bar</text>], { select: true });
```

Sandbox: https://codesandbox.io/s/slate-reproductions-9eeek?file=/index.js

Slate: 0.58.4

#### What's the expected behavior?

When inserting multiple nodes with `Transforms.insertNodes`, the selection should be placed at the end of the last inserted node.

PR #4482, opened 2021-08-31
fix cursor not correct when insert multiple nodes using insertNodes

Jokcy:
**Description**
When using `Transforms.insertNodes` to insert multiple nodes, the cursor will be placed only after the first inserted node. This mainly because we only move `at` to next path once, key code below:
```js
at = isAtEnd ? Path.next(path) : path
// ...
const point = Editor.end(editor, at)

if (point) {
  Transforms.select(editor, point)
}
```
The `Path.next` will only invoke once which means the cursor only move after one node.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/3816

**Example**
before
old

after
Aug-31-2021 21-38-11

**Context**
I just move the cursor every time a node is inserted, and because we go to next at first, so we need to go back at last.

Also I added two test cases for this situation, both cases are failed at the previous version.

**Checks**

Actually I can build the site and server it under `http-server`, but when I run `yarn start` I can not view anything in my browser, it keep loading forever, I don't have time to figure the reason now, but I suppose I can build and server it success then it means *The relevant examples still work*.

### DIFF
--- a/.changeset/neat-flies-eat.md
+++ b/.changeset/neat-flies-eat.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Fix cursor not correct issue after insert multiple nodes with `Transform.insertNodes`
--- a/packages/slate/src/transforms/node.ts
+++ b/packages/slate/src/transforms/node.ts
@@ -228,7 +228,9 @@ export const NodeTransforms: NodeTransforms = {
         const path = parentPath.concat(index)
         index++
         editor.apply({ type: 'insert_node', path, node })
+        at = Path.next(at)
       }
+      at = Path.previous(at)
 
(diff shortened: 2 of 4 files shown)


dylans (approved):

TheSpyder:
There's an easier way to do this, and `insertNodes` already uses it when deleting selected nodes:
https://github.com/ianstormtaylor/slate/blob/dd752df11dc90da6bd6add88d1cfa6f00f03912b/packages/slate/src/transforms/node.ts#L184-L186

Point refs (in fact all refs) are automatically updated in response to applied operations.

The code works, so having this already merged is fine, but might be nice as a cleanup task at some point.

Issue #4483, opened 2021-08-31
feat(slate-react): more flexible / performant dynamic decorations via small API change


jaked:
**Problem**
With the existing decorations mechanism, it is not easy to make decorations dynamic, so they're recomputed based on external state. Currently the only way to do it is to change the `decorate` function, which causes every element in the document to re-render. (There was a recent change #4138 which improved this slightly by keeping the `decorate` function in a context rather than passing it down as a prop, but still every `Element` component accesses the context.)

For example: In [my application](https://github.com/jaked/programmable-matter) I use decorations for error highlighting in code blocks. A change in one code block can affect highlighting in another block, so I need to redecorate on every edit to a code block. Changing the `decorate` function on even a moderately-sized document is very slow.

**Solution**
A small change to the `renderElement` API can greatly improve the situation: pass a new `renderChildren` argument, which takes a `decorations` argument. These decorations are added to any existing decorations when rendering the children.

Now in `renderElement`, instead of e.g.
```ts
const renderElement = (props: RenderElementProps) => {
  switch (props.element.type) {
    case 'code':
      return <code>{props.children}</code>
    ...
```
we can write
```ts
const renderElement = (props: RenderElementProps) => {
  switch (props.element.type) {
    case 'code':
      const decorations = ... // syntax highlighting
      return <code>{props.renderChildren({ decorations })}</code>
    ...
```
We can use whatever React state mechanism we like to trigger re-rendering / redecoration per node, rather than globally.

As a side benefit: with the existing decoration mechanism, the `decorate` function is run on the immediate children of the editor on every edit (in `useChildren` in `Editable`), even when the function is not changed. This doesn't cause the elements to be re-rendered (because `Element` is memoized on the decorations list), but it can be expensive in itself (in my test case it was ~20ms). If we compute decorations per node in `renderElement`, they are not recomputed on every edit.

I prototyped this change in my app and it makes an **enormous** difference—in a moderately-sized document, editing a code block drops from ~200ms to ~10ms.

**Alternatives**
I experimented with rendering the decorations manually in `renderElement`. Unfortunately I wasn't able to get this to work because Slate needs `Text` nodes to be rendered with the `Text` component (to populate various `WeakMap`s used to map DOM elements to Slate nodes), and this component is not exported. Even if this did work, the decorations mechanism is nicer—it requires less code duplication, exposes less of Slate's internals, and implements cross-node decorations already.

[...]

PR #4484, opened 2021-08-31
add `renderChildren` callback to `renderElement`

jaked:
Add a `renderChildren` argument to `renderElement` that takes decorations to be applied to the element's children, so
dynamic decorations may be rendered without a global re-render.

**Description**
With the existing decorations mechanism, it is not easy to make decorations dynamic, so they're recomputed based on external state. Currently the only way to do it is to change the `decorate` function, which causes every element in the document to re-render.

A small change to the `renderElement` API can greatly improve the situation: pass a new `renderChildren` argument, which takes a `decorations` argument. These decorations are added to any existing decorations when rendering the children.

(This is a first cut PR for discussion.)

**Issue**
Fixes: #4483

**Example**
```ts
const renderElement = (props: RenderElementProps) => {
  switch (props.element.type) {
    case 'code':
      const decorations = ... // syntax highlighting
      return <code>{props.renderChildren({ decorations })}</code>
    ...
```

**Context**
See #4483 for some alternatives.

**Checks**

### DIFF
--- a/site/examples/richtext-with-code-highlighting-renderChildren.tsx
+++ b/site/examples/richtext-with-code-highlighting-renderChildren.tsx
@@ -1,0 +1,405 @@
+import Prism from 'prismjs'
+import 'prismjs/components/prism-python'
+import 'prismjs/components/prism-php'
+import 'prismjs/components/prism-sql'
+import 'prismjs/components/prism-java'
+import React, { useCallback, useMemo, useState } from 'react'
+import faker from 'faker'
+import isHotkey from 'is-hotkey'
+import {
+  Editable,
+  withReact,
+  useSlate,
+  RenderTextProps,
+  Slate,
+} from 'slate-react'
+import {
+  Editor,
+  Transforms,
+  createEditor,
+  Descendant,
+  Element as SlateElement,
+  Text,
+} from 'slate'
+import { withHistory } from 'slate-history'
+import { css } from 'emotion'
+
+import { Button, Icon, Toolbar } from '../components'
+
+const HOTKEYS = {
+  'mod+b': 'bold',
+  'mod+i': 'italic',
+  'mod+u': 'underline',
+  'mod+`': 'code',
+}
+
+const LIST_TYPES = ['numbered-list', 'bulleted-list']
+
+const HEADINGS = 100
+const PARAGRAPHS = 7
+const initialValue: Descendant[] = []
+
+for (let h = 0; h < HEADINGS; h++) {
+  initialValue.push({
+    type: 'heading',
+    children: [{ text: faker.lorem.sentence() }],
+  })
+
+  for (let p = 0; p < PARAGRAPHS; p++) {
+    initialValue.push({
+      type: 'paragraph',
+      children: [{ text: faker.lorem.paragraph() }],
+    })
+  }
+}
+
+const LanguageContext = React.createContext('html')
+
+const RichTextWithCodeHighlightingRenderChildrenExample = () => {
+  const [value, setValue] = useState<Descendant[]>(initialValue)
+  const [language, setLanguage] = useState('html')
+  const renderElement = useCallback(props => <Element {...props} />, [])
+  const renderLeaf = useCallback(props => <Leaf {...props} />, [])
+  const editor = useMemo(() => withHistory(withReact(createEditor())), [])
+
+  const renderText = useCallback((props: RenderTextProps) => {
+    const { attributes, renderChildren, path, text } = props
+    const decorations = []
+    if ('code' in text) {
+      const language = React.useContext(LanguageContext)
+      const tokens = Prism.tokenize(text.text, Prism.languages[language])
+      let start = 0
+
+      for (const token of tokens) {
+        const length = getLength(token)
+        const end = start + length
+
+        if (typeof token !== 'string') {
+          decorations.push({
+            [token.type]: true,
+            anchor: { path, offset: start },
+            focus: { path, offset: end },
+          })
+        }
+
+        start = end
+      }
+    }
+    return <span {...attributes}>{renderChildren({ decorations })}</span>
+  }, [])
+
+  return (
+    <Slate editor={editor} value={value} onChange={value => setValue(value)}>
+      <div
+        contentEditable={false}
+        style={{ position: 'relative', top: '5px', right: '5px' }}
+      >
+        <h3>
+          Select a language
+          <select
+            value={language}
+            style={{ float: 'right' }}
+            onChange={e => setLanguage(e.target.value)}
+          >
+            <option value="js">JavaScript</option>
+            <option value="css">CSS</option>
+            <option value="html">HTML</option>
+            <option value="python">Python</option>
+            <option value="sql">SQL</option>
+            <option value="java">Java</option>
+            <option value="php">PHP</option>
+          </select>
+        </h3>
+      </div>
+      <Toolbar>
+        <MarkButton format="bold" icon="format_bold" />
+        <MarkButton format="italic" icon="format_italic" />
+        <MarkButton format="underline" icon="format_underlined" />
+        <MarkButton format="code" icon="code" />
+        <BlockButton format="heading-one" icon="looks_one" />
+        <BlockButton format="heading-two" icon="looks_two" />
+        <BlockButton format="block-quote" icon="format_quote" />
+        <BlockButton format="numbered-list" icon="format_list_numbered" />
+        <BlockButton format="bulleted-list" icon="format_list_bulleted" />
+      </Toolbar>
+      <LanguageContext.Provider value={language}>
+        <Editable
+          renderElement={renderElement}
+          renderText={renderText}
+          renderLeaf={renderLeaf}
+          placeholder="Enter some rich text…"
+          spellCheck
+          autoFocus
+          onKeyDown={event => {
+            for (const hotkey in HOTKEYS) {
+              if (isHotkey(hotkey, event as any)) {
+                event.preventDefault()
+                const mark = HOTKEYS[hotkey]
+                toggleMark(editor, mark)
+              }
+            }
+          }}
+        />
+      </LanguageContext.Provider>
+    </Slate>
+  )
+}
+
+const toggleBlock = (editor, format) => {
+  const isActive = isBlockActive(editor, format)
+  const isList = LIST_TYPES.includes(format)
+
+  Transforms.unwrapNodes(editor, {
+    match: n =>
+      LIST_TYPES.includes(
+        !Editor.isEditor(n) && SlateElement.isElement(n) && n.type
+      ),
+    split: true,
+  })
+  const newProperties: Partial<SlateElement> = {
+    type: isActive ? 'paragraph' : isList ? 'list-item' : format,
+  }
+  Transforms.setNodes(editor, newProperties)
+
+  if (!isActive && isList) {
+    const block = { type: format, children: [] }
+    Transforms.wrapNodes(editor, block)
+  }
+}
+
+const toggleMark = (editor, format) => {
+  const isActive = isMarkActive(editor, format)
+
+  if (isActive) {
+    Editor.removeMark(editor, format)
+  } else {
+    Editor.addMark(editor, format, true)
+  }
+}
+
+const isBlockActive = (editor, format) => {
+  const [match] = Editor.nodes(editor, {
+    match: n =>
+      !Editor.isEditor(n) && SlateElement.isElement(n) && n.type === format,
+  })
+
+  return !!match
+}
+
+const isMarkActive = (editor, format) => {
+  const marks = Editor.marks(editor)
+  return marks ? marks[format] === true : false
+}
+
+const Element = ({ attributes, children, element }) => {
+  switch (element.type) {
+    case 'block-quote':
+      return <blockquote {...attributes}>{children}</blockquote>
+    case 'bulleted-list':
+      return <ul {...attributes}>{children}</ul>
+    case 'heading-one':
+      return <h1 {...attributes}>{children}</h1>
+    case 'heading-two':
+      return <h2 {...attributes}>{children}</h2>
+    case 'list-item':
+      return <li {...attributes}>{children}</li>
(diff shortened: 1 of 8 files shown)


ulion:
the children was already calculated before renderElement, and the new renderChildren will do it again, not good enough.

would you consider add optional renderChildren to Editable like renderElement, and use it to calculate children with additional props you need? so still use one children (and only calculated once) but customizable by user?

on the other hand, I'm not in slate maintain team, it's just my personal comment for this.

echarles:
I don't have enough slate internal knowledge to comment on @ulion feedback.

Something that would help to figure out performance gain would be to add a `Fast Code Highlighting` example aside the current on https://github.com/ianstormtaylor/slate/blob/e51566ada84cfa107c445cc6f3908e78c18656b6/site/examples/code-highlighting.tsx.

Pasting the source of this github page with HTML highlighting takes numerous seconds. One should be able to measure as a user the gain with the current option vs the gain with the option @ulion proposes.

Happy to measure the gain when the example is available and report here.

jaked:
> the children was already calculated before renderElement, and the new renderChildren will do it again, not good enough.

Yes, I mentioned in the issue that we could address this by making `children` a lazily-initialized value. What do you think of that approach? I'll take a crack at that and add it to the PR.

> would you consider add optional renderChildren to Editable like renderElement, and use it to calculate children with additional props you need? so still use one children (and only calculated once) but customizable by user?

This seems difficult to use—it would require that `renderElement` and `renderChildren` each dispatch on the node type, and they couldn't share data between them.

On the other hand—what you're suggesting is not really different from the existing `decorate` callback on `Editable`, since `decorations` is the only extra prop passed to `renderChildren`. Maybe it would be possible to use React hooks inside of `decorate` to achieve dynamic decorations, without changing the API? (I think it might require changing where `decorate` gets called though; right now it's called for each node during the render of the node's parent; it should be called during the render of that node, so hooks are set on the correct component.)

ulion:
> > the children was already calculated before renderElement, and the new renderChildren will do it again, not good enough.
> 
> Yes, I mentioned in the issue that we could address this by making `children` a lazily-initialized value. What do you think of that approach? I'll take a crack at that and add it to the PR.
> 
> > would you consider add optional renderChildren to Editable like renderElement, and use it to calculate children with additional props you need? so still use one children (and only calculated once) but customizable by user?
> 
> This seems difficult to use—it would require that `renderElement` and `renderChildren` each dispatch on the node type, and they couldn't share data between them.
> 
> On the other hand—what you're suggesting is not really different from the existing `decorate` callback on `Editable`, since `decorations` is the only extra prop passed to `renderChildren`. Maybe it would be possible to use React hooks inside of `decorate` to achieve dynamic decorations, without changing the API? (I think it might require changing where `decorate` gets called though; right now it's called for each node during the render of the node's parent; it should be called during the render of that node, so hooks are set on the correct component.)

As your last comment, I can share that personally I am using react-tracked, by which I can control which exactly node be re-rendered. but, it maybe not worth to inject it into slate officially.

jaked:
> Something that would help to figure out performance gain would be to add a `Fast Code Highlighting` example aside the current on https://github.com/ianstormtaylor/slate/blob/e51566ada84cfa107c445cc6f3908e78c18656b6/site/examples/code-highlighting.tsx.
> 
> Pasting the source of this github page with HTML highlighting takes numerous seconds. One should be able to measure as a user the gain with the current option vs the gain with the option @ulion proposes.

Unfortunately this change is not a general improvement in render time for decorations; it just avoid unnecessary re-rendering in cases where decorations are dynamically updated.

Out of curiosity I tried profiling pasting the source of this page into the highlighting example. Basically all the time is spent manipulating the Slate tree, not in rendering. I think it would be well worth optimizing this! But in a separate issue / PR.

echarles:
> Unfortunately this change is not a general improvement in render time for decorations; it just avoid unnecessary re-rendering in cases where decorations are dynamically updated.

I guess we have to distinguish 2 case:

1. The rendering and decoration of the initial value (or of the pasted value).
2. The update of the decorations based on an incremental value change.

Maybe this PR help the incremental change. An new example would really help to showcase the intent of `renderChild` 

> Out of curiosity I tried profiling pasting the source of this page into the highlighting example. Basically all the time is spent manipulating the Slate tree, not in rendering. I think it would be well worth optimizing this! But in a separate issue / PR.

I am still catching up, but will also profile and think more on optimization as soon as I can.

echarles:
Playing a bit more on https://www.slatejs.org/examples/code-highlighting

1. Pasting the source of this pages takes +/- 20 seconds on my laptop (UI is like freezed, and then everything shows up after 20 seconds).
2. Once shows, I can type new html source which is highlight without delay.- I guess the `decorate` function is only triggered for the new Element.
3. Changing the language in the top right dropdown changes the colors very fast (nearly instantaneous), with the token remaining unchanged - sounds to me like it is more a CSS update rather that a full decoration.

Food for thoughts:

- Monaco does show the text without highlight, and then only highlight (visible when pasting a big html source in foo.html in Visual Studio Code)
- CodeMirror.next does incremental parsing, see example https://codemirror.net/6/examples/million, with the incremental lezer parser https://github.com/lezer-parser
- React 18 has plans for performance https://reactjs.org/blog/2021/06/08/the-plan-for-react-18.html (automatic batching and new startTransition API).
- Can service worker help https://developers.google.com/web/fundamentals/primers/service-workers?
- I have implemented performance improvement for JupyterLab with Intersection Observers https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API. Probably not to be put in slate core, but worth considering using slate to optimize decorations. BTW I have never used https://github.com/thebuilder/react-intersection-observer but there maybe good ideas to pull from there.

echarles:
Forgot to mention "A proof-of-concept code editor with syntax highlighting that uses" https://github.com/datavis-tech/codearea

Sorry, my comment does not help reviewing this PR, but it looks to me that this PR aims getting better performance for decorations, so worth considering any other approach in parrallel.

echarles:
I have tested the `renderChildren` approach for the highlight example (see the implementation below of the renderElement which decorates each descendants, with the Editable not using the global decorate method). Pasting the source of this page takes +/- 30 seconds in both cases (the global decorate or the renderChildren).

```
  const renderElement = useCallback((props: RenderElementProps) => {
    const { children, renderChildren, path, element, attributes } = props
    const decorations = []
    const [[node, _]] = Node.descendants(element)
    if (Text.isText(node)) {
      const tokens = Prism.tokenize(node.text, Prism.languages[language])
      let start = 0
      for (const token of tokens) {
        const length = getLength(token)
        const end = start + length
        if (typeof token !== 'string') {
          decorations.push({
            [token.type]: true,
            anchor: { path, offset: start },
            focus: { path, offset: end },
          })
        }
        start = end
      }
    }
    return <p>{renderChildren({ decorations })}</p>
  }, [])

```

jaked:
> I have tested the `renderChildren` approach for the highlight example (see the implementation below of the renderElement which decorates each descendants, with the Editable not using the global decorate method). Pasting the source of this page takes +/- 30 seconds in both cases (the global decorate or the renderChildren).

Yeah, again, this change isn't going to help the from-scratch case; it just gives a way to re-render only affected nodes when decorations need to change. I will cook up an example where this makes a difference and add it to the branch. (I am still trying to figure out if there is a way to achieve the goal without hairing up the interface of `renderElement`.)

Also, as I mentioned above, the cost of pasting a big text is mostly in the Slate tree manipulation during the paste, not in decoration or rendering. To see this, remove the `decorate` argument to `Editable` and try the experiment again.

echarles:
> Yeah, again, this change isn't going to help the from-scratch case; it just gives a way to re-render only affected nodes when decorations need to change.

Got it now. Thx for helping understand things here. I didn't expect wrapping in a the render element in a <Suspense> would have helped, and indeed, it did not help...

> I will cook up an example where this makes a difference and add it to the branch.

That would be great!

> (I am still trying to figure out if there is a way to achieve the goal without hairing up the interface of renderElement.)

The new API change is understandable to me. Just not backwards compatible (for typescript developers).

BTW on https://github.com/ianstormtaylor/slate/issues/4483#issue-984187882 you say `I think the approach of passing decorations to renderChildren is better overall than a global decorate function, and in some future revision it would make sense to remove decorate entirely.`- Doesn't the search case need that global decorate function, in which case we would end-up having both ways: global or child specific?

> Also, as I mentioned above, the cost of pasting a big text is mostly in the Slate tree manipulation during the paste, not in decoration or rendering. To see this, remove the decorate argument to Editable and try the experiment again.

Without the call to `decorate`, the paste remain slow which tells me that the latency is not at the decorate level, but somewhere else. Still it may be at the rendering side (or not), which a profiling could tell (still need to do that).

> Out of curiosity I tried profiling pasting the source of this page into the highlighting example. Basically all the time is spent manipulating the Slate tree, not in rendering. I think it would be well worth optimizing this! But in a separate issue / PR.

Pasting the large html in the plaintext example is also slow (takes the same time).

However, when I use the "huge document" example and I set as initialValue that exact same large piece of html, the display. is instantaneous.

I am wondering if the slowness would be more on the data transfert (copy/paste) or at the reception of that data?

jaked:
> BTW on [#4483 (comment)](https://github.com/ianstormtaylor/slate/issues/4483#issue-984187882) you say `I think the approach of passing decorations to renderChildren is better overall than a global decorate function, and in some future revision it would make sense to remove decorate entirely.`- Doesn't the search case need that global decorate function, in which case we would end-up having both ways: global or child specific?

Yeah, it's a good point, sometimes you do want global decoration. I guess that was a hasty comment on my part—I was just thinking that it would be better not to have two ways to do it. Still thinking about the right API here.

> Pasting the large html in the plaintext example is also slow (takes the same time).
> 
> However, when I use the "huge document" example and I set as initialValue that exact same large piece of html, the display. is instantaneous.

This makes sense to me—the `initialValue` already contains the Slate tree so there is no work to construct it; what you're seeing there is just rendering time. You could do something similar with the highlighting example (set an `initialValue` containing the tree of the pasted text) to see just the decoration / rendering time.

I didn't dig into the profile much, but it seemed like it had to do with splitting up the paste into paragraphs and inserting each paragraph then normalizing. I think we should move this discussion to a separate issue though!

jaked:
I added two examples combining the rich text editor, code highlighting, and huge document examples. One uses the existing API, the other uses the `renderChildren` API. Also I made the existing `children` into a lazy value so we don't duplicate work.

Through the render profiler in the React developer tools extension you can see that the `renderChildren` API avoids re-rendering non-code nodes when you switch code highlighting languages, because only code nodes depend on the language variable (used via a React context). And it's overall a lot faster (~25ms instead of ~80ms).

I don't totally understand the profile results though, they don't seem to show every re-rendered component so I'm not totally clear where the time is going.

Here's the profile showing that all the `Element` nodes re-render when `decorate` changes in the existing API:
image

Here's the profile showing that only code nodes re-render using the `renderChildren` API:
image

site/examples/richtext-with-code-highlighting.tsx
```diff
@@ -0,0 +1,398 @@
...
+  })
+
+  for (let p = 0; p < PARAGRAPHS; p++) {
+    initialValue.push({
+      type: 'paragraph',
+      children: [{ text: faker.lorem.paragraph() }],
+    })
+  }
+}
+
```
echarles, line 49:
```suggestion
const CODE_BLOCKS = 1000
const initialValue: Descendant[] = []

initialValue.push({
  type: 'paragraph',
  children: [
    {
      text: `<!doctype html>
    <meta charset="utf8">
    <body>`,
      code: true,
    },
  ],
})

for (let h = 0; h < CODE_BLOCKS; h++) {
  initialValue.push({
    type: 'paragraph',
    children: [
      {
        text: `          <p>This is a HTML placeholder paragraph.</p>
        <hr>
        <img src="./slate.svg">`,
        code: true,
      },
    ],
  })
}

initialValue.push({
  type: 'paragraph',
  children: [
    {
      text: `</body>`,
    },
  ],
})

```

site/examples/richtext-with-code-highlighting-renderChildren.tsx
```diff
@@ -0,0 +1,405 @@
...
+  })
+
+  for (let p = 0; p < PARAGRAPHS; p++) {
+    initialValue.push({
+      type: 'paragraph',
+      children: [{ text: faker.lorem.paragraph() }],
+    })
+  }
+}
+
```
echarles, line 55:
```suggestion
const CODE_BLOCKS = 1000
const initialValue: Descendant[] = []

initialValue.push({
  type: 'paragraph',
  children: [
    {
      text: `<!doctype html>
    <meta charset="utf8">
    <body>`,
      code: true,
    },
  ],
})

for (let h = 0; h < CODE_BLOCKS; h++) {
  initialValue.push({
    type: 'paragraph',
    children: [
      {
        text: `          <p>This is a HTML placeholder paragraph.</p>
        <hr>
        <img src="./slate.svg">`,
        code: true,
      },
    ],
  })
}

initialValue.push({
  type: 'paragraph',
  children: [
    {
      text: `</body>`,
    },
  ],
})

```

echarles:
Thx @jaked. I have also done some profiling with your 2 new examples:

On initial load, no real gain, which is expected.

- RichTextWithCodeHighlightingExample (global decorate): 170 ms
- RichTextWithCodeHighlightingRenderChildrenExample (renderChildren): 150 ms

On language onChange, I see the same results as you have reported.

- RichTextWithCodeHighlightingExample (global decorate): 40 ms
- RichTextWithCodeHighlightingRenderChildrenExample (renderChildren): 10ms

I have suggested other initialValue which are html code, but not sure if it makes sense, as in that case, the difference between the 2 examples will not be shown (all children would have to be rerendered).

I am not sure for the interpretation of the profiles. I see a pattern where the elements are rendered at the begging, then the elements are not rendrered, to end with a sections where elements are rendered again. I guess I would first need to exercise a bit more with profiling with standard case, before trying to understand the decorate case.

dylans:
@jaked still thinking of exploring this one or should I close it out?

jaked:
I'm not going to be able to work on this. thanks @dylans

PR #4488, opened 2021-09-03
call `decorate` for a node inside render of node's component

jaked:
**Description**
One way to implement dynamic decorations is to pass a different
`decorate` function on every decoration change. But this is very
expensive, because it triggers a render of every `Element`.

Another way to implement dynamic decorations is to call a React hook
inside the `decorate` function, so rendering is triggered when the
hook fires. By selectively calling the hook, you can arrange to render
only components that actually need to be updated.

However, in the existing code, the `decorate` call for a node runs in
the render of the node's parent component, so more of the tree is
rendered than necessary when a hook fires. This change moves the
`decorate` call for a node into the component for that node, so only
affected components are rendered.

**Issue**
Fixes: #4483

**Example**
Suppose you want to do syntax highlighting with themes. You can write a `decorate` function like so:
```ts
function decorate(([node, path])) => {
  if (Element.isElement(node) && node.type === 'code') {
    const theme = React.useContext(ThemeContext);
    return highlight(node, theme);
  }
}
```
Now when `ThemeContext` changes, the hook fires and `code` nodes are re-rendered, but non-`code` nodes are not re-rendered.

**Context**
See description. I made another attempt to address this problem in #4484, but I think this way is better because it doesn't change the API, and keeps decorations separate from other rendering.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/element.tsx
+++ b/packages/slate-react/src/components/element.tsx
@@ -1,9 +1,10 @@
 import React, { useRef } from 'react'
 import getDirection from 'direction'
-import { Editor, Node, Range, NodeEntry, Element as SlateElement } from 'slate'
+import { Editor, Node, Path, Range, Element as SlateElement } from 'slate'
 
 import Text from './text'
 import useChildren from '../hooks/use-children'
+import { useDecorate } from '../hooks/use-decorate'
 import { ReactEditor, useSlateStatic, useReadOnly } from '..'
 import { useIsomorphicLayoutEffect } from '../hooks/use-isomorphic-layout-effect'
 import {
@@ -27,6 +28,7 @@ import {
 const Element = (props: {
   decorations: Range[]
   element: SlateElement
+  path: Path
   renderElement?: (props: RenderElementProps) => JSX.Element
   renderPlaceholder: (props: RenderPlaceholderProps) => JSX.Element
   renderLeaf?: (props: RenderLeafProps) => JSX.Element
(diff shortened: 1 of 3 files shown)


echarles:
I have tried to paste a medium html document in the code-highlighting example https://www.slatejs.org/examples/code-highlighting which has taken 17 seconds.

Then I have pulled, built and run this branch locally, and the same document pasting the the code-higlighting example has taken exactly the same time (17 seconds). 

Is anything needed to update the example?

echarles:
I guess the unchanged performance was expected as this PR targets rendering only what needs to be re-rendered. In the test I have done, everything needs to be rendered in both cases.

jaked:
turns out this idea doesn't quite work, I didn't notice at first because I was just looking at a narrow performance test.

First: it's not possible to call a hook in `decorate` in the existing code, because `decorate` is called from `useChildren` inside the parent component. So if the decorated things in the parent's children change (e.g. you add a new decorated node), that changes the hooks called for the parent component, which breaks the Rules of Hooks.

Second: with this change (move the `decorate` call into `Element` / `Text`), calling a hook in `decorate` shouldn't break the Rules of Hooks because hooks calls should match up with components. However, Slate is very careful to maintain React component keys across operations (see `apply` in `with-react.ts`), so even when an operation changes the type of a node (with `set_node`) it gets the same React key. That means that the React component is updated, not unmounted / mounted; so if a node changes from a non-decorated type to a decorated type, again the hooks for the component change.

One way to fix this would be to drop the React key mapping when a node is changed with `set_node`, but I'm not sure if that is important for performance. It does seem like if you change the `type` of a node, it's not surprising if it causes a remount of the whole subtree—it is likely to do so anyway because `renderElement` functions usually dispatch to another component based on the `type`. But other changes made by `set_node` might not necessarily entail a remount, so for performance it might be better to keep the key mapping.

dylans:
This one has gone a bit stale as well @jaked ... please update or close when time permits. Thanks.

jaked:
I'm not going to be able to work on this. thanks @dylans

Issue #4496, opened 2021-09-08
on Chrome, shift-arrow down then delete doesn't delete text


jaked:
**Description**
If you position the cursor at the beginning of a paragraph, shift-arrow down, then press delete, the text should be deleted. On HEAD on Chrome, instead the cursor moves to the previous line and the text is not deleted.

I bisected the issue to https://github.com/ianstormtaylor/slate/commit/2d1aaafa from #4455. Not sure what the problem is. I see that the change affects Chrome only, and I tested on Safari; the bug is not present.

**Recording**
Bug at https://github.com/ianstormtaylor/slate/commit/2d1aaafa :

![Bug](http://g.recordit.co/MfP3aKW9He.gif)

Correct behavior at previous commit https://github.com/ianstormtaylor/slate/commit/8eb1abac :

![Correct](http://g.recordit.co/nB93ldYtDn.gif)

**Sandbox**
https://www.slatejs.org/examples/richtext

**Steps**
To reproduce the behavior:
1. position the cursor at the beginning of a paragraph
2. shift-arrow down to select paragraph
3. press delete
4. see that the cursor moves to the previous line and the text is not deleted

**Expectation**
Text should be deleted.

**Environment**
- Slate Version: HEAD
- Operating System: MacOS
- Browser: Chrome 93.0.4577.63

TheSpyder:
Duplicate of #4492

PR #4498, opened 2021-09-08
Fix: regression caused by triple click fix

### LINKED ISSUE #4490: Crash during mouse selection on Chrome over multiple nodes
**Description**
When selecting across Nodes in Chrome, the editor crashes out.

**Recording**
slate_crash

**Sandbox**
https://www.slatejs.org/examples/editable-voids 

**Steps**
To reproduce the behavior:

- Going to the example: https://www.slatejs.org/examples/editable-voids in Google Chrome ( 92.0.4515.159 ). 
- Enter text in name and hand radio ('test' and left)
- Select from top to bottom using mouse click and drag
- Big red face is shown

**Expectation**
No crash.

**Environment**
- Slate Version: slate@0.65.3
- Operating System: Mac Big Sur
- Browser: Chrome 92.0.4515 

**Context**
I think this is something to do with selecting void nodes. I came across it when selecting large pieces of text and then applying formatting to it. I don't think the formatting has anything to do with it. It happens randomly in my code, but it is very consistent in the official slate example above.

bytrangle:
fixes #4490 

**Description**
Introducing fix for triple-click selection in this [PR](https://github.com/ianstormtaylor/slate/pull/4455#issuecomment-913821241) caused some regressions that I didn't expect.

This new PR will alleviate that.

**Issue**
Some of the issue caused by the triple click fix include, but not limited to:
- Crash when selecting void nodes #4490 
- Shift + down arrow then delete text doesn't delete it #4496 

**Context**
There are two problems with my triple-click fix:
- The condition for checking if an action is a triple-click is too broad.
- The logic for resetting `focusNode` of a selection is not fail-proof. The building block of the content component in a Slate editor are HTML element with attribute `data-slate-node="element"`. They may contain child elements `data-slate-string`, but not always. The current logic assumes that Element always contains String elements, which leads to the crash when selecting void nodes.

**Checks**
- [] The tests pass with `yarn test`.

### DIFF
--- a/.changeset/smart-pears-beam.md
+++ b/.changeset/smart-pears-beam.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fix: regression caused by triple click fix
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -590,11 +590,15 @@ export const ReactEditor = {
             )
             const focusElement = tripleClickedBlock!.lastElementChild
             // Get the element node that holds the focus text node
+            // Not every Slate element node contains a child node with `data-slate-string`,
+            // such as void nodes, so the result below could be null.
             const innermostFocusElement = focusElement!.querySelector(
               '[data-slate-string]'
             )
-            const lastTextNode = innermostFocusElement!.childNodes[0]
-            focusNode = lastTextNode
+            if (innermostFocusElement) {
+              const lastTextNode = innermostFocusElement.childNodes[0]
+              focusNode = lastTextNode
+            }
           }
         }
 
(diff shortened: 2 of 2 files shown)


bytrangle:
Currently, two integration tests failed, one for Forced Layout and the other for Rich Texts. However, they all pass on my part so can anyone check where the problems come from?

dylans (approved):

bytrangle:
Thank you @dylans for the edit. Are the failed tests acceptable? They passed on my local machine.

dylans:
> Thank you @dylans for the edit. Are the failed tests acceptable? They passed on my local machine before the addition of changeset.

They appear to be timeouts in the CI/CD testing environment, so I’m not going to block based on that.

TheSpyder:
Thanks for doing that quickly after it was discovered! I have a small request.

This is a fix for a change that wasn't released, so it doesn't need a changeset. The original fix did need a changeset but one wasn't added. As such, your one entry in the changelog now looks a bit weird.

Would you mind creating a new PR that removes this changeset and adds one for the #4455 fix? I would do it myself but then you'd lose credit.

Issue #4492, opened 2021-09-06
Deleting a selection only deletes the last element


nemanja-tosic:
**Description**

Given an element with multiple text nodes and the entire element was selected, when deleting the selection only the last node gets removed.

**Recording**

![Peek 2021-09-06 21-39](https://user-images.githubusercontent.com/10130001/132256443-14256b63-825f-40e7-a282-ee281c530be8.gif)

Note: I am hitting backspace to delete the selection.

**Expectation**

The entire selection should be deleted.

**Environment**

The issue is reproducible on Chrome on https://www.slatejs.org/examples/richtext.

**Context**

By bisecting from the last know good (0.65.3) to main, [the triple click PR](https://github.com/ianstormtaylor/slate/commit/2d1aaafa0b1947fada70fdb672fafeb45a008609) appears to be the culprit.

tomhicks:
Hey, I just stumbled across something similar.

The same behaviour is exhibited when selecting text and cutting/copying multiple lines of text. When you paste, you realise it's only cut/copied the last line.

jaked:
this is broken in a different way now—if you shift-arrow down two paragraphs then hit delete, only the first paragraph is deleted:

current main (b10849182086699d4bb18209a37ea6247f712bd0):

![Bug](http://g.recordit.co/y25Z88QDyQ.gif)

correct behavior (8eb1abac prior to #4455):

![Correct](http://g.recordit.co/MSZz8IejcP.gif)

jaked:
spent some time investigating, here are some findings:

* the code as written assumes the selection has only one block, which is of course not true in general (see previous comment)
* Safari has the same behavior as Chrome when triple-clicking (selection hangs to following element)
* all of Safari, Firefox, and Chrome (on Mac at least) place the selection hanging to the following element if you select with shift-arrow-down, giving the "false toolbar button highlighted" issue reported in #4329

I think that the original Slate behavior is correct, and that the bug is actually in the toolbar highlighting example code—it should unhang the selection before checking for marks. It's unfortunate that Firefox doesn't do the same thing as Chrome/Safari/Edge on triple-click (non-hanging vs. hanging selection) but it seems pretty minor. 

[...]

PR #4512, opened 2021-09-13
fix(react-editor): reset focus offset when triple clicking

bytrangle:
**Description**

In my triple click fix, I reset the `focusNode` but forgot to reset the related `focusOffset` value. As a result, the Slate range computed from DOM selection is still not correct and cause a number of issues when users try to transform a block by applying formatting or deleting. 

**Issue**
Fixes: #4492 

**Example**
Before the fix

https://user-images.githubusercontent.com/28924121/133071690-27d21192-b86c-4b6a-9741-0f2f1ff53218.mp4

After the fix

- Selecting a block by clicking and dragging, then hitting Backspace should remove that block.

https://user-images.githubusercontent.com/28924121/133098079-78fdccd1-3eef-42a3-b852-ca0ba55453d9.mp4

- Selecting a block through clicking and hitting Shift + Arrow down, then hitting Backspace should remove that block.

https://user-images.githubusercontent.com/28924121/133098140-5d2963af-0014-4501-a117-b4b6d71f1d91.mp4

- Triple click at the beginning of a Mark should delete the whole block that contains that Mark.

https://user-images.githubusercontent.com/28924121/133098182-1198893c-aa2c-4291-b777-ab5c44f06643.mp4

**Context**
Here's what I did:
- Refactor the condition for checking for triple clicking or Shift + down arrow. Chrome seems to have issue with selection that is not clicking and dragging.
- Rewrite the code for getting `focusNode` to avoid error thrown when accessing property of a potentially null value.
- Add logic for getting the DOM Text node of the `focusNode` object.

**Checks**

### DIFF
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -572,36 +572,52 @@ export const ReactEditor = {
         // beginning of the next block
         if (
           IS_CHROME &&
-          anchorNode?.nodeType !== focusNode?.nodeType &&
+          anchorNode &&
+          focusNode &&
+          anchorNode.nodeType !== focusNode.nodeType &&
           domRange.anchorOffset === 0 &&
           domRange.focusOffset === 0 &&
-          focusNode?.nodeValue == null
+          focusNode.nodeValue == null
         ) {
           // If an anchorNode is an element node when triple clicked, then the focusNode
           //  should also be the same as anchorNode when triple clicked.
-          if (anchorNode!.nodeType === 1) {
-            focusNode = anchorNode
-          } else {
-            // Otherwise, anchorNode is a text node and we need to
-            // - climb up the DOM tree to get the farthest element node that receives
-            // triple click. It should have atribute 'data-slate-node' = "element"
-            // - get the last child of that element node
-            // - climb down the DOM tree to get the text node of the last child
-            // - this is also the end of the selection aka the focusNode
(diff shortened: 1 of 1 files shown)


packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -572,36 +572,52 @@ export const ReactEditor = {
...
-            // Get the element node that holds the focus text node
-            // Not every Slate element node contains a child node with `data-slate-string`,
-            // such as void nodes, so the result below could be null.
-            const innermostFocusElement = focusElement!.querySelector(
-              '[data-slate-string]'
-            )
-            if (innermostFocusElement) {
-              const lastTextNode = innermostFocusElement.childNodes[0]
-              focusNode = lastTextNode
+          // if (anchorNode!.nodeType === 1) {
```
dylans, line 584: Just remove the commented out code.

dylans (changes requested):

packages/slate-react/src/plugin/react-editor.ts
bytrangle: I'm not sure what you meant. Which part is the comment-out code?

dylans (approved):

dylans:
Sigh, I merged without a changeset. opening a PR for the changeset now.

bytrangle:
I forgot to ask a noob question: When should I add a changeset, and when should I not? I thought this change is a correction for my flawed triple click fix, so I didn't think of adding a changeset.

Issue #3605, opened 2020-04-11
Select all then delete leaves empty blocks


nikglavin:
#### Do you want to request a _feature_ or report a _bug_?
Bug

#### What's the current behavior?
Using the [plaintext](https://www.slatejs.org/examples/plaintext) example 

1.  Enter a line of text
2. Press return 3 or 4 times
3. Ctrl + A (select all), delete

Observe that the placeholder is not displayed and there are empty blocks still in the editor. To get the placeholder to display again you must press delete a few times (equal to the number of empty blocks) then type a character and press delete again (only removes the last empty not when a character has been removed).

![slate](https://user-images.githubusercontent.com/2267259/79033786-5e25e580-7bf4-11ea-8639-06d036e86877.gif)

Slate: 0.XX.X (Production demo)
Browser: Chrome
OS: Mac

#### What's the expected behavior?
Select all +  delete removes all nodes returning the editor to the placeholder.

CameronAckermanSEL:
Hi @nikglavin , I went to go check this out (on my windows box so not full OS parity here) and I couldn't reproduce the behavior. Instead, I saw that the delete worked but it left the selection expanded over the placeholder.

If you have a couple of moments, it'd help if you could double check that the bug is still there (or not) on the examples.

Thanks!

nikglavin:
Hi @CameronAckermanSEL the issue still exists. I think what you experienced is the issue. When there are empty blocks in the editor they are not being removed with: select all -> delete

ktrieu:
I can also reproduce this issue on the latest [richtext demo](https://www.slatejs.org/examples/richtext).

I've attached [a video.](https://streamable.com/i1q42e)

The leftover second block can be normalized away, but the placeholder still doesn't show up even with only one block remaining.

As far as I can tell, [the conditions](https://github.com/ianstormtaylor/slate/blob/97dbab13a88cbeb7b9d41287fc7e6a2086211f1e/packages/slate-react/src/components/editable.tsx#L435) for adding the placeholder seem to be satisfied in the broken state, but for some reason the placeholder is still not showing up.

pkeuter:
I can also easily reproduce it on the demo. 

If it helps: the same issue also occurs when a void node is the last item. This item will not be selected and thus will not be deleted when pressing `mod+a`.

emcmanus:
This is because `editor.deleteFragment` unhangs the selection before deletion. The last line is blank, so the block is excluded from the range and can't be deleted.

[...]

PR #4522, opened 2021-09-14
Fix unhangRange to respect voids

steve-codaio:
**Description**
When dealing with void elements, unhangRange would treat a range in a void the same as any other element. Conceptually since a void only has that single point to represent content, it shouldn't be treated as hanging in that case. 

In addition, this updated the method to fix two other issues that cropped up when writing the tests for it.
1. Drop the `voids` option entirely, and just perform the above behavior. The reason for that change is that it's somewhat unclear what this option even means for 'unhanging' a range. `voids` is a property of what is traversed/fetched _at_ a location, while the range itself _is_ a location. 

    For example if I pass `voids: false`, but the range I pass in has focus sitting inside a void, what should happen? The range isn't considered 'hanging' by the new definition above, but should we still move the focus because it's in a void? 

    This is also supported by the fact that:
     - A previous version I made of this change (See just the first commit) tried to respect the flag, however this pre-existing [test](https://github.com/ianstormtaylor/slate/blob/main/packages/slate/test/transforms/delete/voids-false/block-hanging-into.tsx#L15) has a range ending in a void, and the expected output is actually that the range does _not_ get modified. This test only passed because we were passing `voids: false` to our traversal and also skipping the first node due to the `skip = true` flag, effectively skipping 2 nodes instead of just 1. Fixing that 'bug' caused the test to fail and not have the desired behavior. 
     - We only ever actually forwarded the 'voids' option to this method in one place, most locations that call it do not pass a voids option, despite having an upstream param for it. 
2. Remove the 'start !== 0' optimization, since the code only unhangs the 'end', it was actually excluding ranges that were hanging, because the start was in the middle of a node.

**Example**

If you had the following situation today
```
<block>
  <text>Some text<anchor></text>
  <inline void><text/></inline void>
  <text><focus></text>
</block>

```
and wanted to perform an operation that included the void node but did not include the trailing empty text, (i.e., unhang the range), that would not be possible since we would skip over the empty text 

**Checks**

### DIFF
--- a/.changeset/metal-tomatoes-beg.md
+++ b/.changeset/metal-tomatoes-beg.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+unhangRange treats void elements as having content
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -1631,9 +1631,15 @@ export const Editor: EditorInterface = {
   ): Range {
     const { voids = false } = options
     let [start, end] = Range.edges(range)
+    const endAtVoid = Boolean(Editor.void(editor, { at: end.path }))
 
     // PERF: exit early if we can guarantee that the range isn't hanging.
-    if (start.offset !== 0 || end.offset !== 0 || Range.isCollapsed(range)) {
+    if (
+      start.offset !== 0 ||
+      end.offset !== 0 ||
+      (voids && endAtVoid) ||
+      Range.isCollapsed(range)
+    ) {
       return range
     }
 
@@ -1644,7 +1650,9 @@ export const Editor: EditorInterface = {
     const blockPath = endBlock ? endBlock[1] : []
     const first = Editor.start(editor, [])
     const before = { anchor: first, focus: end }
-    let skip = true
+    // If the end is in a void and we're not traversing voids, no need to skip
+    // as the end point will implicitly be skipped
+    let skip = voids || !endAtVoid
 
     for (const [node, path] of Editor.nodes(editor, {
       at: before,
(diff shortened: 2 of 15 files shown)


dylans:
Thanks @steve-codaio! I'm going to leave this open for a few days for others on the team to review and give feedback as there may be scenarios you and I are not considering (but none are immediately coming to mind for me a very quick cursory review).

steve-codaio:
Not sure why that integration test failed. Just ran it locally and everything passed fine

dylans:
> Not sure why that integration test failed. Just ran it locally and everything passed fine

They're a bit flaky in the CI environment and time out sometimes, I don't see anything concerning with the integration tests and your changes.

TheSpyder:
I need to check how this relates to #4203 which I'm hoping to get back to soon. At the time @ianstormtaylor was concerned about changes to unhang that impacted default delete selection behaviour.

steve-codaio:
@TheSpyder ah yeah it looks like a slightly different approach to the problem, as I was also trying to solve the inline void. For the tests that started failing around deletion, I 'fixed' them differently. I added `hanging: true` to the delete calls. I believe the tests themselves weren't quite testing the right thing, in that they were behaving as if hanging were true, but it only happened to work due to the `start !== 0` check. 

That fact actually relates to the issue brought up here #3605, where it's somewhat unexpected that deletion unhangs selection (especially when whether it unhangs or not depends on if your selection happens to be at the start of a text or not). 

The scenarios worried about in your PR would work fine with `hanging: true`. I think the early bailout for startOffset !== 0 is effectively working in tandem with the unexpected unhanging during deletion, to inadvertently get the desired behavior. For example in our codebase we actually have a wrapper api that always does deletes with `hanging: true` for this reason.

dylans (approved):

dylans:
Apologies that this one has been sitting open for a while @ steve-codaio . I'm going to get more feedback from the team to make sure this is the right way to proceed and have a decision shortly.

dylans:
There have been some recent updates that may have addressed this separately. Please review the latest release and re-open if this is still relevant. Thanks.

PR #4523, opened 2021-09-14
Fix setNodes when called with 'split' and a collapsed range

steve-codaio:
**Description**
When calling 'setNodes' with a collapsed range, and `split: true`, what should happen is nothing, (`split` means we want to style whatever is in between the two points, which will be nothing), however we were seeing a full node get styled as if a split never occurred. Tracked it down to two issues:

- When a range is already collapsed and the affinity was 'inward' (the default), it was possible to end up going too far 'inward' and have the anchor push past the focus. This change detects that case and just leaves it collapsed, preferring whichever direction the anchor would be moving. 
- setNodes uses this transform as part of a rangeRef to track edits while it splits the nodes at the provided range. Because the range was collapsed the rangeRef would end up outside the target of the split, and set properties on the whole node, even with the above fix to 'inward' affinity because we don't have enough context to know which direction to move in the collapsed case for the two splits. Since it should be a noop anyway for non-empty text nodes, just early return when we detect that case

Technically only the 2nd fix is needed to prevent setNodes from doing the wrong thing, but the range transform seemed incorrect either way

**Checks**

### DIFF
--- a/.changeset/sharp-donkeys-sparkle.md
+++ b/.changeset/sharp-donkeys-sparkle.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Fix setNodes when called with 'split' and a collapsed range
--- a/packages/slate/src/interfaces/range.ts
+++ b/packages/slate/src/interfaces/range.ts
@@ -222,12 +222,16 @@ export const Range: RangeInterface = {
       let affinityFocus: 'forward' | 'backward' | null
 
       if (affinity === 'inward') {
+        // If the range is collapsed, make sure to use the same affinity to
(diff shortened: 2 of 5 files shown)


steve-codaio:
Update: This has an issue, when the collapsed range is on an empty text the updates should be able to go through just fine (since they won't be normalized away). Marked as WIP for now, will update with new changes and updated tests tomorrow

steve-codaio:
Changed the code to only do the early return on collapsed ranges in non-empty text nodes rather than unconditionally for collapsed ranges

dylans (approved):

Issue #4527, opened 2021-09-15
Text is always being inserted, even when using auto markdown


churichard:
**Description**
Slate seems to always be inserting text on every key press, even when you're not explicitly calling `editor.insertText`. This is especially a problem if you're using Markdown Shortcuts: the last key you type will always be inserted into the editor. If you continue typing, that piece of text will be deleted and overwritten with whatever you type.

This bug does not happen on 0.65.3.

**Recording**
https://user-images.githubusercontent.com/4218237/133479278-1761fd91-6119-46db-8a03-1a84b3bed359.mp4

Basically what's happening is that there is an extra space being inserted after typing `# ` for the Heading 1 auto markdown shortcut. You can tell by inspecting the DOM or by seeing that the cursor is misaligned with the other lines. It's even more noticeable if your trigger character is not a space, because then it will insert that trigger character directly.

**Sandbox**
https://www.slatejs.org/examples/markdown-shortcuts

**Steps**
To reproduce the behavior:
1. Type the auto markdown shortcut for Heading 1: `# `
2. Observe that a space is inserted even though it's not supposed to

**Expectation**
Text should only be inserted when you're calling `editor.insertText`.

**Environment**
- Slate Version: 0.66.1
- Operating System: Windows
- Browser: Chrome

nemanja-tosic:
Tracked the issue down to https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/editable.tsx#L314 as part of #3888

dylans:
@churichard could you please try out #4529 and verify this works for you as well? Thanks!

churichard:
> @churichard could you please try out #4529 and verify this works for you as well? Thanks!

I don't know how to try it out in my own codebase, but I can verify that it works on the Markdown Shortcuts example on the Slate website. So likely should be fine for me too 👍

PR #4529, opened 2021-09-15
Fix erroneous text after native insert

nemanja-tosic:
**Description**

The code was backing out of a native insert if there are ops other than insert_text, but at that point the decision to preventDefault was already made. The PR takes into account the flush as well, and does a prevent default in case a flush has occured.

**Issue**

Fixes: #4527

**Notes**

Refactored asNative to record previous state and reset it in a finally clause. A variable is set, so the function should clean up after itself, to put it tersely. Also, captured the previous state of AS_NATIVE, as to account for potential nesting of the calls.

**Checks**

### DIFF
--- a/packages/slate-react/src/utils/native.ts
+++ b/packages/slate-react/src/utils/native.ts
@@ -10,10 +10,25 @@ export const NATIVE_OPERATIONS: WeakMap<Editor, Operation[]> = new WeakMap()
  * @param {Editor} editor - Editor on which the operations are being applied
  * @param {callback} fn - Function containing .exec calls which will be queued as native
  */
-export const asNative = (editor: Editor, fn: () => void) => {
+export const asNative = (
+  editor: Editor,
+  fn: () => void,
+  { onFlushed }: { onFlushed?: () => void } = {}
+) => {
+  const isNative = AS_NATIVE.get(editor)
+
   AS_NATIVE.set(editor, true)
-  fn()
-  AS_NATIVE.set(editor, false)
+  try {
+    fn()
+  } finally {
+    if (isNative !== undefined) {
+      AS_NATIVE.set(editor, isNative)
+    }
+  }
+
+  if (!NATIVE_OPERATIONS.get(editor)) {
+    onFlushed?.()
+  }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/slate-react/src/utils/native.ts
```diff
@@ -10,10 +10,25 @@ export const NATIVE_OPERATIONS: WeakMap<Editor, Operation[]> = new WeakMap()
...
-  AS_NATIVE.set(editor, false)
+  try {
+    fn()
+  } finally {
+    if (isNative !== undefined) {
+      AS_NATIVE.set(editor, isNative)
+    }
+  }
+
+  if (!NATIVE_OPERATIONS.get(editor)) {
```
nemanja-tosic, line 29: The important bit, everything else is a refactor (boy scout rule).

packages/slate-react/src/utils/native.ts
```diff
@@ -10,10 +10,25 @@ export const NATIVE_OPERATIONS: WeakMap<Editor, Operation[]> = new WeakMap()
...
+    onFlushed?.()
+  }
```
nemanja-tosic, line 31: The important bit, everything else is a refactor (boy scout rule).

dylans (approved):

churichard (approved):
Would be great to merge this in 😄

clauderic:
Hey there, looks like there was a regression with deletion of selected fragments in Firefox. I've traced the regression back to this PR:


https://user-images.githubusercontent.com/1416436/134735852-8c7e82e7-0977-45bd-850a-a792022414e4.mp4


This behaviour can be replicated in all versions of Firefox that I tested (I checked on Firefox 92, 91, and 90).

nemanja-tosic:
Hi @clauderic. Are you sure this PR was the cause and not #3888? This PR was trying to fix a bug introduced by #3888.

clauderic:
Hey @nemanja-tosic, I'm fairly certain it was this PR yes. The previous commits right up until this PR work fine in Firefox:


https://user-images.githubusercontent.com/1416436/134739272-22f17cc3-5e12-40d1-ae81-bc268b0ff158.mp4

This isn't to say that your PR doesn't indeed fix a bug introduced by #3888, but it seems to have caused other regressions.

nemanja-tosic:
That's ok, I'll take a look. There are several problems due to #3888 right now which I'm looking at as well. Thanks!

Issue #3575, opened 2020-04-02
Delayed setState(value) results in Cannot resolve a DOM point from Slate point


FezVrasta:
#### What's the current behavior?

Repro: https://codesandbox.io/s/slate-reproductions-1sbiz

![Kapture 2020-04-02 at 17 24 18](https://user-images.githubusercontent.com/5382443/78267072-d90d4300-7506-11ea-9f44-701ed4b500bc.gif)

Slate: 0.57.1
Browser: Chrome
OS: Mac

#### What's the expected behavior?

The above repro shows how a delayed `setState(value)` call systematically results in the error `Cannot resolve a DOM point from Slate point`.

In my case, I'm using Apollo to keep some text in sync with the server, I have optimistic updates in place, but it's still not quick enough to not trigger this error.

I also encountered the same issue while trying to sync with some quite large (and badly written) Redux store.

CameronAckermanSEL:
Thanks for posting this issue.

I think this is caused by the selection state of the editor briefly being in a state that isn't possible to achieve in its value state because they are being applied out of sync.

So, understanding that the onChange here is contrived and not going to usually include setTimeout (?), my recommendation is to ensure that your onChange loop is _as tight as possible_ and if you must really do something that takes a meaningful amount of time, subscribe to onChange only to get updates and don't pass that value back to slate so it can continue to control its own internal value

One "fix" for this bug is to move the selection state back into the value state. However, that's a non trivial change and wont be done any time soon, if at all. It was a deliberate decision to keep the selection state internal. 

But a lot of people have these issues so maybe over time it'll be enough to justify that transition

agnislav:
Hi guys.

I'm pretty new with Slate (a few weeks), but the current behavior looks reasonable for me. I'd be grateful if you can explain what's wrong with it.

So, how I see it:
- slate editor object contains the state of the editor
- `editor.children` contains value while `editor.selection` contains the current state of selection
- `editor.children` is exposed outside via `value` and `onChange` `<Editor>` props and not updated by slate itself
- `editor.selection` is purely internal stuff, which is clear as the selection can be made over a couple of nodes.
- this means that a developer is responsible to keep editor value in immediate sync with the slate
- also, this means that if a developer changes the value from outside, he's responsible to update the selection as well.

Would be great to hear which of the points above aren't correct and why.

phamstack:
Same issue (Chrome Latest). Typing too fast (intentionally) will cause the problem above.

[...]

PR #4540, opened 2021-09-20
Only use Slate Provider's value prop as initial state

bryanph:
**Description**
As is made clear by the following issue https://github.com/ianstormtaylor/slate/issues/3575, the `value` prop that must be passed to the `Slate` provider is kind of a hoax because in reality this prop has to be exactly what has been passed to `onChange`. This changes the `Slate` provider to relax this requirement. The `value` prop is now only used on the initial render to initialize the `value` prop. On subsequent changes it is ignored. 

This also paves the way to rename `value` to `initialValue` as the prop is really intended to be.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/3575

**Checks**


dylans (changes requested):
This needs a changeset, and it needs review by a few others as I don't have the history of experience to know with certainty if this change is ok (it seems ok).

bryanph:
Not sure whether to use a `minor` or `patch` for this one :thinking:. Settled with `minor` for now.

Ghost---Shadow:
The documentation needs to be updated as well.
https://docs.slatejs.org/walkthroughs/06-saving-to-a-database

hassenc:
> I'm curious whether this misses some subtlety in the rendering code that I'm unaware of. Please do educate me if so 

onChange should not be the only way of changing the Component's value. If you rely on a database to render your content, you'll also want to pass the new value using the value prop. In that case you don't want to rely on a local state that is changed with onChange.

An example would be when you need to display someone else changes to the value.

dylans:
> This is a breaking change. It should be a major bump.

Pre 1.0 beta so not a major bump.

> > I'm curious whether this misses some subtlety in the rendering code that I'm unaware of. Please do educate me if so
> 
> onChange should not be the only way of changing the Component's value. If you rely on a database to render your content, you'll also want to pass the new value using the value prop. In that case you don't want to rely on a local state that is changed with onChange.
> 
> An example would be when you need to display someone else changes to the value.

How do you reconcile that with the discussion at https://github.com/ianstormtaylor/slate/issues/3575#issuecomment-923241634 ?

Ghost---Shadow:
I am ok with either way. But the documentation over here  https://docs.slatejs.org/walkthroughs/06-saving-to-a-database should be updated.

bryanph:
Besides the docs, the examples should probably also be updated to reflect this change which means just passing `initialValue` directly to the `value` prop instead of the `value` state that is currently kept in all examples.

philicious:
I also think the documentation could need some update regarding how to update the editors value.

I had a rather common use-case: async loading contents from my backend. Therefore couldnt set it as init value in `useState`. (docs only talk about synchronously loading from a local storage database.)
I spent quite some frustrating time playing around, reading  and code docs and eventually stumbling over this discussion.

Docs left the impression to me that you can use `value={content}` together with `useState` to update the editors contents any time.

bryanph:
@philicious agreed about the docs, although for async loading you can simply defer rendering the editor until the content is fetched.

PinoFlores:
So at the end of all how clients can update the slate value?

bsides:
> @philicious agreed about the docs, although for async loading you can simply defer rendering the editor until the content is fetched.

Works for the first load of the editor, but what if you want to return to the original state of the editor after changing the values?

For example:

- The UI has a form with an Edit button in the end.
- If the user clicks the Edit button, the UI hides it and then shows a Cancel and Save button.
- The initial value is loaded from a database and then the Slate component is rendered in the first time the page mounts, as suggested above.
- The user then changes the value of the slate textarea.
- Then the user press cancel.
- The expected behavior was to use the initial value that came from the database but it never happens because we lost it and we couldn't change the state externally... again.

How do you tackle this problem? Is it still with `editor.children`?

philicious:
@bsides I would (try to) save the original content in state, then use `editor.children = originalContent` in Cancel button click handler

bsides:
> @bsides I would (try to) save the original content in state, then use `editor.children = originalContent` in Cancel button click handler

Since the original content is external, that's saving props as state, which is an anti pattern. I'm updating to `editor.children` for now but considering a better alternative.

philicious:
@bsides ye in case you consider the content being updated from other places, you need to fetch it again and not use state. or use state + employ mutex locking in your content DB to only allow single edit instance

bsides:
The thing is, this is a pretty trivial thing for a form to do: to be trully controllable in case we have external data being pulled/pushed asynchronously. Right now it's really not a good implementation because it's trying to do 2 things and failing in both.

lraccomando:
Would like to +1 on the need to making this clearer in the documentation. For me anyway this is a very common use-case and directly setting `editor.children` is relatively unintuitive given the design patterns used in other React inputs. Not at all the end of the world, but an explicit note and example in the documentation could save a lot of headaches.

macrozone:
Is there a chance that this can be reverted? I think it was never justified in the first place, but seeing the problems it created just confirms this.

mrcoles:
Just dropping a note for anyone using slate-history…

Setting `editor.children` doesn’t save the change to your history, e.g., when your editor is this:

 ```ts
const editor = useMemo(() => withHistory(withReact(createEditor())), []);
```

I see in @KarimNahas example https://github.com/ianstormtaylor/slate/pull/4540#issuecomment-951903419 that we can instead use `editor.apply` to make changes that will be remembered.

I ran into an additional problem getting the cursor focus to remain on undo though. I tried adding `Transforms.select` and also `Transforms.setSelection` and I was seeing the `selectionBefore` value in `editor.history.undos[0]`, but when I did undo, the cursor focus would get lost. As a workaround, instead of the node operations, I used "remove_text" and "insert_text" operations and that seems to work (also, I’m using `Element.isElement` to parse editor.children safely).

xc:
@SantosOMartinez , this is a nice idea, but unfortunately it doesn't work in my case. 

My problem: 
- using editor.children under root will lose selection
- using useEffect(with editor.children inside) will result in rendering issue(not rendered up-to-date when external changes, always missing on character, so I need to use key for this - that's crazy).

Using the way Santos suggested will result in so many renderings for same content.

Really hope this can be solved. What I'm doing now is a parameter (props:{useEffect?:boolean}) set useEffect and direct editor.children(need to make sure the useEffect:true will be no external change on the content) - kind of ugly...

PR #4555, opened 2021-09-28
Allow passing custom editor creator to slate-hyperscript `createEditor`

bryanph:
**Description**
Just like in slate's core, I use `slate-hyperscript` to perform tests for my particular editor. This change makes it easier to create your own testing setup by allowing you to pass your own slate `Editor` instance to `slate-hyperscript`'s `createEditor`. This way you can simply reuse the setup code that's in there without needing to duplicate that code in your own setup.

**Issue**
Fixes: (link to issue)

**Example**
`createHyperscript` would then look like this
```
const jsx = createHyperscript({
  creators: {
    editor: createEditor(aFunctionThatReturnsAnEditorObject)
  },
  elements: {
    block: { type: 'block' },
    inline: { type: 'inline' }
  }
})
```
`createEditor` then takes care of setting the `anchor` and `focus` properly given the JSX.

**Checks**


dylans (approved):
Please add a changeset when you get a chance, thanks!

bryanph:
@dylans done!

Issue #4524, opened 2021-09-15
Bug: If the cursor is at the end of an inline, Slate moves it outside of the inline before inserting text


jameshfisher:
**Description**
"If the cursor is at the end of an inline, Slate moves it outside of the inline before inserting text." [This is explicitly coded behavior.](https://github.com/ianstormtaylor/slate/blob/95c759a19c1e057bbc99148867298a73b014831d/packages/slate/src/create-editor.ts#L165-L184) I nevertheless claim that this is a bug, or at least a misfeature in core.

**Recording**

https://user-images.githubusercontent.com/166966/133413152-4e8a6ae0-ad89-4405-92e1-072875f5ad88.mp4

**Sandbox**
https://codepen.io/jameshfisher/pen/xxrXvVO (basically a default Slate app, with an inline `code` element)

**Steps**
To reproduce the behavior:
1. Go to https://codepen.io/jameshfisher/pen/xxrXvVO
2. Put the cursor at the end of the inline code element
3. Type something

**Expectation**
If the cursor is at the end of an inline, text should be inserted at the end of the inline. This expectation comes from:

1. The fact that the cursor is _inside_ the inline. For the user, the blinking cursor is visually _inside_ the inline, not outside it. For the developer, the cursor (i.e. `editor.selection`) is _inside_ the inline, not outside it. The _definition_ of "the cursor" is "this is where your text will go when you type". To break that behavior is really jarring.
1. Slate's principle that "all of its logic is implemented with a series of plugins". IMO, if the current behavior is actually desirable in some circumstance, it absolutely shouldn't be core; it should be exposed as something like `withMoveCursorOutOfInline(editor)`. It's harder and less elegant for me to _remove_ the core behavior with my own plugin.
1. Comparison with other editors. Default `contenteditable` inserts the text at the end of the inline, as expected. Medium inserts the text at the end of the inline, as expected. (Though note that Notion has the same bug as Slate, and as a user, it's REALLY annoying.)
1. Usability: how else is the user supposed to enter text at the end of the inline ..? The only way seems to be to insert text _before_ the end of the inline, and then delete the rest of the inline. This is obviously horrible.

**Environment**
- Slate Version: 0.66
- Operating System: presumably all (tested macOS, Windows)
- Browser: presumably all (tested Chrome, Firefox)

**Context**

* This was brought up by @jarosik10 in https://github.com/ianstormtaylor/slate/issues/4074, but was less explicit about the problem.
[...]

Issue #3253, opened 2019-12-05
Links example: text added after link gets included in link


dmarkow:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

In the Links example, when the last word is a link, it becomes impossible to add non-link text after it without hitting Enter and starting a new paragraph first.

![ScreenFlow](https://user-images.githubusercontent.com/8009/70268015-382f9680-1765-11ea-98f4-62975c73cded.gif)

Slate: 0.51.0
Browser: Chrome 
OS: Mac

#### What's the expected behavior?

Placing the cursor immediately after the last character in a link and typing should result in non-link text (this is how most rich editors handle it)

ianstormtaylor:
This is a known bug right now, but solvable by: https://github.com/ianstormtaylor/slate/issues/3151

dmarkow:
Fixed in #3260

PR #4578, opened 2021-10-08
Allow typing at the end of an inline element

jameshfisher:
**Description**

Steps to reproduce the buggy behavior:

* Create a page with an inline element, or go to https://codepen.io/jameshfisher/pen/xxrXvVO
* Put the cursor at the end of the inline code element
* Type something

Expected behavior: If the cursor is at the end of an inline, text should be inserted at the end of the inline.

Actual behavior: Slate moves the cursor outside the inline before inserting text.

This current behavior is explicitly coded and was added in https://github.com/ianstormtaylor/slate/pull/3260. I nevertheless claim that this is a bug, or at least a misfeature in core. My expected behavior comes from:

* The fact that the cursor is inside the inline. For the user, the blinking cursor is visually inside the inline, not outside it. For the developer, the cursor (i.e. `editor.selection`) is inside the inline, not outside it. The definition of "the cursor" is "this is where your text will go when you type". To break that behavior is really jarring.
* Slate's principle that "all of its logic is implemented with a series of plugins". If the current behavior is actually desirable in some circumstance, it should be in a plugin, not core. It's harder and less elegant for me to remove the core behavior with my own plugin.
* Comparison with other editors. The following editors all insert text at the end of the inline, as expected: default contenteditable, Medium, Coda.io, Google Docs, OneNote, Evernote, QuillJS, TinyMCE, EditorJS, ProseMirror. Two editors with the current buggy behavior are Notion and Dropbox Paper, but I find it hard to imagine that their current behavior on this issue is actually designed, and not just accidental.
* Usability: how else is the user supposed to enter text at the end of the inline ..? The only way seems to be to insert text before the end of the inline, and then delete the rest of the inline. This is obviously horrible.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4524

**Example**

https://user-images.githubusercontent.com/166966/133413152-4e8a6ae0-ad89-4405-92e1-072875f5ad88.mp4

**Checks**


jaked:
The current behavior was added explicitly in #3260 to fix #3253, and there's still a test for it in https://github.com/ianstormtaylor/slate/blob/main/packages/slate/test/transforms/insertText/selection/inline-end.tsx

There's also some code in `Editable` that seems to be supposed to match the code you've removed here: https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/editable.tsx#L297

I think the fix in #3260 was a mistake. The issue reported in #3253 is not correct; it is possible to navigate past the end of an inline. If it's the last element in the document, you can arrow down (or drag the cursor down) to move outside the inline; or if it's not the last element, you can move one position past the end of the inline, then back again. This works in all the browsers I tested (Chrome, Safari, Firefox, all on Mac). To see this, try moving the cursor in this [Slate Explorer playground](https://slate-explorer.glitch.me/#eyJpbnB1dCI6IjxlZGl0b3I+XG4gIDxwPlxuICAgIDxpbmxpbmU+c29tZSB0ZXh0PC9pbmxpbmU+PGN1cnNvci8+XG4gIDwvcD5cbjwvZWRpdG9yPiIsInNsYXRlIjpbeyJ0eXBlIjoicCIsImNoaWxkcmVuIjpbeyJ0ZXh0IjoiIn0seyJ0eXBlIjoiaW5saW5lIiwiY2hpbGRyZW4iOlt7InRleHQiOiJzb21lIHRleHQifV19LHsidGV4dCI6IiJ9XX1dLCJ0cmFuc2Zvcm0iOiIiLCJzaG93SGVscCI6ZmFsc2V9) and see where the `<cursor/>` ends up.

jameshfisher:
@jaked so at the end of all that, I've left in the special-case code, and instead changed its comment - the real reason it exists is that native Chrome has (rather unbelievably) the very same bug we're trying to remove from Slate: it doesn't allow typing at the end of an `<a>` element.

(Honestly, the #3888 change really scares me, considering how buggy and inconsistent the native contenteditable implementations are.)

jameshfisher:
> > The integration tests keep failing, but it looks like a flaky test to me.
> 
> Yes, I wouldn't worry about that too much.

Okay, great. I think this PR is ready for review then.

> > (Honestly, the #3888 change really scares me, considering how buggy and inconsistent the native contenteditable implementations are.)
> 
> Me as well. I've been debating whether we should revert it or try to work through the various issues it has introduced.

I really, really, _really_ appreciate the simplicity of "just re-render every time". If the goal is to make spellcheck less flickery, it feels like there must be a less radical solution (e.g. opening bug reports with buggy browser vendors, or somehow disabling spellcheck for ~1 second after each editor change). It also looks the native events doesn't help the "support IME" goal, since it currently only works on characters `a-z`. But I don't understand these issues well enough yet. May be worth opening a new issue to track the various issues with native character insertion ...

dylans:
@jameshfisher could you please look at main after the landing of #4605 and then make sure this works fine on top of that. I'm going to release 0.67.0 now, but happy to do a 0.67.1 once this change has been verified on top of that. Thanks!

dylans:
> This was perfect timing since I'm currently working on fixing this problem for inline nodes in our text editor.
> 
> Unfortunately, the new behavior is also undesirable since it forces you to hit enter and backtrack like @jaked mentioned. In my opinion, this is really no better than forcing a user to backtrack in the middle of an inline node to add the text they wanted and then delete the irrelevant bits.
> 
> Would it be possible to introduce a new placeholder node type in addition to the empty text node (`{ text: '' }`) that sits directly adjacent to it?
> 
> ```tsx
> const tree = {
>   type: 'paragraph',
>   children: [
>     { text: '...' },
>     {
>       type: 'hyperlink',
>       href: 'google.com',
>       children: [...]
>     },
>     { text: 'some unique placeholder' }, 
>     { text: '' }
>   ]
> }
> ```
> 
> That way, pressing the right arrow key somehow positions the selection after that placeholder node.
> 
> Maybe Slate could detect the right arrow key at the end of an inline and insert a space after it, but not part of the inline?
> 
> Just throwing some ideas out there.

It is something I do pretty regularly in our codebase and also in Plate (though this example inserts an extra empty text node) e.g. https://github.com/udecode/plate/blob/a504a1a91a6195a6e8d8c4d7ce5ec0844e9be733/packages/elements/mention/src/getMentionOnSelectItem.ts . 

Historically this is something slate has not tried to do, but I think it's worth trying to solve.

jameshfisher:
I've opened a PR https://github.com/ianstormtaylor/slate/pull/4615 which contributes an official example of a custom inline that should clearly work with "offset" behavior: an "editable button". It shows the several hoops that I've had to jump through to get something like this working. I think we can use this official example as a medium for discussing Slate's default behavior, and how developers can customize it.

Issue #4583, opened 2021-10-09
Setting selection from contentEditable:false element causes crash


jameshfisher:
**Description**
When clicking some text in a `contentEditable:false` element, if the handler for this sets the selection, Slate crashes. Slate tries to find a Slate range for the text that was clicked on, but there is no such range, because the text is inside a `contentEditable:false` element.

**Recording**

https://user-images.githubusercontent.com/166966/136656787-2e45ff8d-fcde-48d7-b03b-2b294a4f5bf9.mp4

**Sandbox**
https://codepen.io/jameshfisher/pen/xxLxwLv

**Steps**
To reproduce the behavior:
1. Go to https://codepen.io/jameshfisher/pen/xxLxwLv
2. Click outside the editor area, so that the editor is defocussed
3. Click on the "Select first para" button

**Expectation**
The selection is set to the first paragraph of the document, because the listener for that button calls `Slate.Transforms.select(editor, [0])`.

**Actual behavior**

Slate crashes with:

```
slate-react.js:3722 Uncaught Error: Cannot resolve a Slate point from DOM point: [object Text],8
    at Object.toSlatePoint (slate-react.js:3722)
    at Object.toSlateRange (slate-react.js:3824)
    at slate-react.js:2172
```

**Environment**
- Slate Version: 0.66.5
- Operating System: macOS
- Browser: Chrome, Firefox

**Context**
Slate seems to be making a bad assumption that the current DOM selection necessarily corresponds to a Slate range, but this is not the case if the user just clicked into an element with `contentEditable: false`.

I can fix this crash by [changing `exactMatch: false` to `exactMatch: true` on this line](https://github.com/ianstormtaylor/slate/blob/7ba486aa397411a3e83ab636b0982167d95319c0/packages/slate-react/src/components/editable.tsx#L185). Everything seems to work as expected after this change. I will submit a PR for this.

PR #4584, opened 2021-10-09
Fix bug: setting selection from contentEditable:false element causes crash

jameshfisher:
Fixes https://github.com/ianstormtaylor/slate/issues/4583

**Description**
When clicking some text in a `contentEditable:false` element, if the handler for this sets the selection, Slate crashes. Slate tries to find a Slate range for the text that was clicked on, but there is no such range, because the text is inside a `contentEditable:false` element. Slate seems to be making a bad assumption that the current DOM selection necessarily corresponds to a Slate range, but this is not the case if the user just clicked into an element with `contentEditable: false`.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4583

**Example**

https://user-images.githubusercontent.com/166966/136656787-2e45ff8d-fcde-48d7-b03b-2b294a4f5bf9.mp4

**Context**
To fix this, I changed `exactMatch: false` to `exactMatch: true`, which seems to mean "fail gracefully if there is no exact match". (This is the exact opposite of what I expect `exactMatch` to mean, but it seems to work.)

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -183,7 +183,10 @@ export const Editable = (props: EditableProps) => {
     if (hasDomSelection && hasDomSelectionInEditor && selection) {
       const slateRange = ReactEditor.toSlateRange(editor, domSelection, {
         exactMatch: false,
-        suppressThrow: false,
+
+        // domSelection is not necessarily a valid Slate range
+        // (e.g. when clicking on contentEditable:false element)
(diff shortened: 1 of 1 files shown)


dylans:
The opposite of this was just changed in a very recent release as noted in #4304. We can revert, but we'll need to reopen the discussion in #4304 and #4293 and come up with a better solution.

jameshfisher:
@dylans great, thanks! 😊

Issue #4530, opened 2021-09-17
Undo/redo regression in slate-react > 0.65.3


z0al:
**Description**

Using `slate-react` > v0.65.3 breaks the undo/redo functionality. Here are two things I noticed:

* Undo (CTRL+Z) changes a single character at a time compared to "normal" undo (in v0.65.3)  that undoes the whole word if written fast enough.
* Redo (CTRL-Y) also redoes a single character at a time but also broken results sometimes. See the recording below.

**Recording**

Tested on https://www.slatejs.org/examples/richtext

https://user-images.githubusercontent.com/12673605/133743140-3d4c0ffd-0e10-4b32-a161-1629f0f29de1.mp4

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/richtext
2. Type `some text`
3. Press undo (CTRL+Z) a few times
4. Press redo (CTRL+Y) a few times

**Expectation**

* Undo should undo words if written fast enough
* Redo should not break the text

**Environment**
- Slate Version: 0.66.1
- Operating System: Ubuntu Linux
- Browser: Brave

nemanja-tosic:
Looking into it. I see no correlation between the speed of typing and what gets undone in the previous version. Here's a video where you can see how slowly i type and still all three a's get removed. Only when i hit backspace and continue typing does undo behave differently. If i were to type a long enough sentence without anything but adding text (and i would as i'm an infallible typist :imp:) and hit undo the whole sentence would be gone.

![Peek 2021-09-24 13-39](https://user-images.githubusercontent.com/10130001/134668718-df23c16a-11d0-46c1-8b9a-d7a26fdc8e84.gif)

Tracked down the change in behavior to #3888.

nemanja-tosic:
The problem is that withReact, after #3888, affects how ops are applied. If it is in native mode, then it gathers operations until onInput is invoked, and then flushes. Changing the example so that `withReact` is called before `withHistory` fixes the issue:

```diff
diff --git a/site/examples/richtext.tsx b/site/examples/richtext.tsx
index 243a182b..98044bf2 100644
--- a/site/examples/richtext.tsx
+++ b/site/examples/richtext.tsx
@@ -25,7 +25,7 @@ const RichTextExample = () => {
   const [value, setValue] = useState<Descendant[]>(initialValue)
   const renderElement = useCallback(props => <Element {...props} />, [])
   const renderLeaf = useCallback(props => <Leaf {...props} />, [])
[...]

PR #4604, opened 2021-10-15
in examples, move withReact in front of withHistory

jaked:
**Description**
PR #3888 defers some operations in `editor.apply` via `withReact`. This interacts badly with the `editor.apply` in `withHistory` if `withHistory` is applied outside `withReact`. This change moves `withReact` outside `withHistory` in all the site examples.

For most examples I put `withReact` outside all the other plugins, but for a few (like the images example) the `withReact` override of `insertData` conflicts with the `insertData` in the examples plugin, so I put `withReact` inside the example's plugin.

**Issue**
Fixes: #4530

**Example**
Current behavior:

New behavior:

**Checks**

I didn't add a changeset because this isn't code that goes in released packages.


dylans:
Reverting this change in favor of #4605.

Issue #4541, opened 2021-09-23
editor.selection not updating after insertText


codewitch:
**Description**

Prior to 0.66.0 editor.selection was updating after insertText to account for the next text. However, any text entry after the first does not update the selection and the offset stays at the previous. 
 

**Recording**
I tested on the Link slate example by adding `console.log(editor.selection)` before and after [this line](https://github.com/ianstormtaylor/slate/blob/main/site/examples/links.tsx#L49).
Below is the console logs after inserting 3 `f`s
<img width="1440" alt="Screen Shot 2021-09-22 at 6 26 11 PM" src="https://user-images.githubusercontent.com/5115188/134442411-735c7587-7dea-4835-b0a3-6c4e5134beca.png">

**Sandbox**
I made a simplified sandbox of it showing the console logs of the editor selection
https://codesandbox.io/s/slate-reproductions-forked-39mx5?file=/index.js

**Steps**
To reproduce the behavior:
1. go to above sandbox 
2. insert text in slate editor
3. see the console and see that the editor selection stays the same.
4. change slate versions to 0.65 and repeat
5. see that editor selection updates

**Expectation**
I expected the editor selection to update after `insertText` because that it how it used to work before the update.

**Environment**
- Slate Version: 0.66
- Operating System: macOS
- Browser: Chrome

nemanja-tosic:
Introduced by #3888 - if only insert_text ops are present, editor state lags until `onInput` is called.

codewitch:
Was that and intentional change or a bug that will get fixed?

If it is an intentional change, how do we consistent editor.selection state in `insertText`?

> Introduced by #3888 - if only insert_text ops are present, editor state lags until `onInput` is called.

nemanja-tosic:
I don't think it was intentional, and i see how it can be problematic. I was reporting my findings to start a conversation around it.

codewitch:
Thanks! Was just making sure. What is the process of getting this fixed? Does it need more thumbs ups?

dylans:
@codewitch Thanks. @nemanja-tosic and I have been discussing it and trying to come up with a good solution and we haven't come up with something easy yet or we would have already fixed it. We should also see if @bkrausz has thoughts.

Mostly we just need time to figure out a good solution that hopefully doesn't break something else along the way.

NicerDicerPro:
Hi @dylans,
im currently facing the same issue in version 0.86.0.

Are there any updates on this? 
Should this already be fixed?

PR #4605, opened 2021-10-16
defer native events within Editable to avoid bugs with Editor

jaked:
**Description**
In #3888 some `insertText` events are allowed to run their default native behavior, in order to avoid re-rendering DOM nodes (re-rendering defeats some browser behaviors—in particular, showing spellchecking squigglies only when typing has stopped). The corresponding Slate `insertText` operation is deferred until the `onInput` handler; after the Slate operation runs, the DOM already matches the new Slate state, so React does not re-render the node.

In the original #3888 code, this is achieved by setting a global flag `AS_NATIVE` in `onDOMBeforeInput`, then running `editor.insertText`. The `withReact` plugin overrides `editor.apply` to add Slate ops to a queue when `AS_NATIVE` is set; then this queue is flushed in the `onInput` handler. However, this interacts badly with other plugins that expect ops to be applied immediately (see #4530, #4541).

To achieve the desired result, it's not necessary to defer individual Slate ops; it's enough to defer the whole `insertText` call. 
This can be done entirely in `Editable`, without involving the `Editor` implementation, so that's what this change does. Note that it doesn't try to bail out when `insertText` is overridden (as #3888 does in `editor.apply`), but I think this is fine—in that case a re-render is triggered, so the effect of the native event is overwritten.

(I guess there are possible overrides of `insertText` where this is not true—one which skips some inserts, or re-renders a different part of the tree. If this is a valid use case then maybe a deeper change is needed.)

**Issue**
Fixes: #3888 #4530 #4541

**Example**
Old behavior showing spellcheck squigglies (current code with `native` always `false`):

New behavior showing proper spellcheck squigglies

Old behavior showing wrong undo behavior:

New behavior showing correct undo behavior:

**Context**
I am not totally sure why it's necessary to defer the Slate operation. If my explanation above about how #3888 works is correct, then it shouldn't really be necessary—the React redraw is already deferred with `ReactDOM.unstable_batchedUpdates` in the `onChange` from `withReact`.

~~I think maybe it's necessary to wait for `onDOMSelectionChange` to fire in order to update Slate's view of the selection to match the changed selection due to the native operation, or else the Slate operation can fail because the selection is wrong. So waiting until `onInput` gives `onDOMSelectionChange` a chance to run first. I haven't investigated this theory however.~~

On second thought this theory doesn't make sense—we want the Slate selection to be in the original spot before running the Slate operation. Still, if we don't `preventDefault` and also don't defer the Slate operation, we get a `Cannot resolve a DOM point from Slate point`, so somehow the selections are out of sync.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -49,7 +49,7 @@ import {
   EDITOR_TO_WINDOW,
 } from '../utils/weak-maps'
 
-type DeferredOperation = () => void;
+type DeferredOperation = () => void
 
 const Children = (props: Parameters<typeof useChildren>[0]) => (
   <React.Fragment>{useChildren(props)}</React.Fragment>
@@ -435,8 +435,9 @@ export const Editable = (props: EditableProps) => {
               // Only insertText operations use the native functionality, for now.
               // Potentially expand to single character deletes, as well.
               if (native) {
-                deferredOperations.current.push(() => Editor.insertText(editor, data)
-                })
+                deferredOperations.current.push(() =>
+                  Editor.insertText(editor, data)
+                )
               } else {
                 Editor.insertText(editor, data)
               }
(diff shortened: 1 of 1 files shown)


echarles:
@jaked Did a clean build, and now undo/redo works great with this PR. Sorry for the confusion here.

Regarding my 2nd point for the issue reported on slack https://slate-js.slack.com/archives/C1RH7AXSS/p1634146964395500, I guess this is out of scope of this PR. Just wanted to give it a try to see if it would bring changes.

Issue #4613, opened 2021-10-20
A more extensible insertData


echarles:
**Problem**

When implementing a custom copy/paste, I would like to be able to reuse as much as possible the existing implementations instead of having to copy/paste the existing ones, with the risk that the implementation diverge.

echarles:
Taking back here comment posted by @dylans on https://github.com/ianstormtaylor/slate/pull/4614#issuecomment-947943736


> mimetype is good also, though keep in mind that many of these operations will have multiple mimetypes so it needs to be easy to prioritize which mimetype wins.

I guess plugins will register mimetype handlers, that will bel be applied in the order that have been registered (first registered, first applied).

> In my experience working on slate, I generally prefer this order: slate > html > plain text (I actually have separate handlers for csv, markdown, and true plain text, but they report the same mimetype, so mimetype may not be enough).

Got it. I need to POC the API and open a PR for that taking your input into account. It will be easier to discuss on code

> Beyond that, we also find scenarios where we should ignore a type based on the paste target. For example, when we paste into a code-block, we generally only want slate ast or plain text, not html, csv, or markdown handling.

OK, will also try to take that into account in the API

PR #4614, opened 2021-10-20
Add insertFragmentData and insertTextData to the ReactEditor API

echarles:
**Description**

This PR split the existing `insertData` in two separated methods that can be overriden. One for the `insertFragmentData`, the other one for the `insertTextData`.

Actually it does not do the intended job which is that we want something more generic.

A way to move forward would be to create a registry when one could register per `mimetype` the actions to be taken.

Looking forward to feedback on this case.

**Issue**

Related to https://github.com/ianstormtaylor/slate/issues/4613

**Checks**

### DIFF
--- a/docs/libraries/slate-react.md
+++ b/docs/libraries/slate-react.md
@@ -138,7 +138,15 @@ Check if a DOM node is within the editor.
 
 ### `insertData(editor: ReactEditor, data: DataTransfer)`
 
-Insert data from a `DataTransfer` into the editor.
+Insert data from a `DataTransfer` into the editor. This is a proxy method to call in this order `insertFragmentData(editor: ReactEditor, data: DataTransfer)` and then `insertTextData(editor: ReactEditor, data: DataTransfer)`.
+
+### `insertFragmentData(editor: ReactEditor, data: DataTransfer)`
+
+Insert fragment data from a `DataTransfer` into the editor.
+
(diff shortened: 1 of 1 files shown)


dylans:
I approve what is there, but agree we need more than this to make it useful.

mimetype is good also, though keep in mind that many of these operations will have multiple mimetypes so it needs to be easy to prioritize which mimetype wins.

In my experience working on slate, I generally prefer this order: slate > html > plain text (I actually have separate handlers for csv, markdown, and true plain text, but they report the same mimetype, so mimetype may not be enough).

Beyond that, we also find scenarios where we should ignore a type based on the paste target. For example, when we paste into a code-block, we generally only want slate ast or plain text, not html, csv, or markdown handling.

echarles:
@dylans Thx for the review. I have updated the description so this only focusses on the method split. The larger factory can be tackled in other PR. Quick questions:

- Do we need a change set.
- Do we need to update the docs so that the 2 new methods becomes part of the public facing API?

dylans:
> @dylans Thx for the review. I have updated the description so this only focusses on the method split. The larger factory can be tackled in other PR. Quick questions:
> 
>     * Do we need a change set?

Yes, anything that changes the codebase (not docs or examples) should have a changeset. I would list this as patch since it's refactoring rather than additive.

>     * Do we need to update the docs so that the 2 new methods becomes part of the public facing API?

Ideally yes. Probably in docs/libraries/slate-react.md

PR #4615, opened 2021-10-21
Official custom inlines example

jameshfisher:
**Description**
This generalizes the "links" example to an "inlines" example, adding a new example of an inline: an "editable button".

Firstly, this is important to demonstrate that Slate really does allow _custom_ elements, and not just "standard" ones like links that you'll find in any editor.

Secondly, it's important to show an example of an inline where "offset" movement should be used. With links, it's arguable that the cursor positions `<link>foo<cursor/></link>` and `<link>foo</link><cursor/>` should be considered the same, because they display in the same position. But with the editable button, the cursor is clearly in a different position, and so offset movement should be used.

**Example**

https://user-images.githubusercontent.com/166966/138306658-0627023b-e417-4139-8bff-ce6e830d8292.mp4

**Checks**

### DIFF
--- a/cypress/integration/examples/inlines.ts
+++ b/cypress/integration/examples/inlines.ts
@@ -6,6 +6,6 @@ describe('Inlines example', () => {
   it('contains link', () => {
     cy.findByRole('textbox')
       .find('a')
-      .should('contain.text', 'hyperlinks')
+      .should('contain.text', 'hyperlink')
   })
 })


imdbsd:
is this https://github.com/ianstormtaylor/slate/pull/4615/files#diff-98b609559db12cb7cc755fd41741b9c72affed934a10cd75db684d5455f3868cR62 workaround on `onKeyDown` necessary now? Last version i use was around 0.5x.0 slate and slate-react. When having only a single inline element on editor, i can easily move to the last and first text node with arrow key. 

When im updating to the latest im cannot moving the cursor again to the first and last text element with cursor again. It fix when im copying the workaoround.

bryanph:
Thanks for this example! It has been incredibly useful.

@imdbsd yes you still need it

Issue #4240, opened 2021-05-05
"Cannot find a descendant at path [0,1] in node" when i try remove inline void element by click on it


slimklim:
**Description**
There is more than one bad magic behavior, but I describe the most important for users.
And I believe fix will help for more than 1 problem.
So. We have a custom element Tag - it is an inline void element. A tag has a cross icon to remove a tag by click. Editor has content: [tag] [text] [tag]. When I remove the first tag by click on the cross icon - focus sets on the next tag. But it should be at the start of [text] (where the first tag was before removing).
For clarity, I use different border colors of a tag when it is focused\unfocused.
You can see other wrong things on the recording. Or just try it yourself in the sandbox, behavior doesn't meet expectations

**Recording**
https://user-images.githubusercontent.com/20947663/117173161-2a0bfc80-add5-11eb-9e43-63c46e0fe4bf.mov

**Sandbox**
https://codesandbox.io/s/slate-reproductions-forked-o9fh7?file=/index.js
I set packages versions as I use in my project, but there is still a problem with the iframe and slate v0.62 https://github.com/ianstormtaylor/slate/issues/4170.
So the record was made in an external window.

**Steps**
To reproduce the behavior:
1. Go to the sandbox
2. Click on the cross icon in the first tag
3. Look where is the focus (this is one trouble)
4. Try to remove the second tag by click until you see an error (and another one trouble)

**Expectation**
A tag must be removable by a cross icon click.
Focus (cursor) must stay on that place, where the removed tag was.
The console must have no errors.

**Environment**
- Slate Version: 0.62
- Operating System: macOS
- Browser: Chrome

liangxiwei:
it look like editor.selecton update not right.you can use setTimeOut about 200ms to workaround this bug.

slimklim:
Thanks! This really helps, but I use 10 ms instead of 200.
However, it is just a workaround. I hope the root cause will be found and the bug will be fixed.

VictorBaron:
In case anyone stumbles upon this:
You can also stop the event propagation, to avoid editor's selection that you probably don't want.

jameshfisher:
[Here's a much smaller reproduction of the bug](https://codepen.io/jameshfisher/pen/ExXoVKe).

Here's what's happening, and how we could fix it.

1. A Slate node `n` is rendered to a DOM element `el` with an `onClick` event that deletes `n` from the Slate document.
[...]

PR #4616, opened 2021-10-21
Fix crash when a void node deletes itself on click

jameshfisher:
**Description**

[Here's a reproduction of the bug that this fixes](https://codepen.io/jameshfisher/pen/ExXoVKe). Here's what's happening:

1. A Slate node `n` is rendered to a DOM element `el` with an `onClick` event that deletes `n` from the Slate document.
2. `el` is clicked, the `onClick` handler runs, and the node `n` is deleted.
3. The `click` event then propagates to [this top-level `Component.onClick` handler](https://github.com/ianstormtaylor/slate/blob/48b7129447347c9cf7a0535026287896ef59779b/packages/slate-react/src/components/editable.tsx#L694-L696), which wants to implement "if the user clicked on a void node, select it".
4. This `Component.onClick` wants to know whether we clicked on a void node, so it consults a reverse mapping from DOM elements back to Slate nodes, looking up the clicked DOM element `el`.
5. But there is no longer a valid Slate node corresponding to `el`, because its custom `onClick` handler has run already, which deleted its own Slate node (and in general, could have arbitrarily changed the Slate document).

This PR fixes the issue by adding some checks that the node is still at that path, before selecting that node.

This PR also adds an 'image delete' feature to the examples. My immediate motivation is to demonstrate the bug that this fixes. But this is also a very common editor feature, and I think it's valuable to show how to achieve it.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4240

**Example**

https://user-images.githubusercontent.com/166966/138331410-74463935-6c3d-4d77-9a9b-d77a4275a7af.mp4

**Checks**

### DIFF
--- a/site/examples/mentions.tsx
+++ b/site/examples/mentions.tsx
@@ -8,7 +8,6 @@ import {
   withReact,
   useSelected,
   useFocused,
-  useEditor,
 } from 'slate-react'
 
 import { Portal } from '../components'
@@ -177,9 +176,6 @@ const Element = props => {
 }
 
 const Mention = ({ attributes, children, element }) => {
-  const editor = useEditor()
-  const path = ReactEditor.findPath(editor, element)
-
   const selected = useSelected()
   const focused = useFocused()
   return (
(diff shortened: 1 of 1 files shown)


jameshfisher:
(An alternative "fix" would be to remove the feature that selects void nodes when they're clicked. It's not clear to me that this is actually useful, and devs can always add this behavior themselves.)

Issue #4626, opened 2021-10-25
insertNodes with select:true puts the cursor *before* the new node


jameshfisher:
**Description**
A call to `Slate.Transforms.insertNodes(..., { select: true })` should always put the cursor inside the new node. But under some conditions (minimal sandbox reproduction below), the cursor is placed BEFORE the new node. Reverting https://github.com/ianstormtaylor/slate/pull/4304 fixes this bug.

**Recording**

https://user-images.githubusercontent.com/166966/138719381-5fbf12d1-76a3-4109-816d-969ff8806c78.mp4

**Sandbox**
https://codepen.io/jameshfisher/pen/gOxmWoJ

**Steps**
To reproduce the behavior:
1. Go to https://codepen.io/jameshfisher/pen/gOxmWoJ
2. Put your cursor in the pink box.
3. Hit return. This removes the pink box, inserts a new one in its place, and selects the new one. But from slate-react@0.66.4, the cursor is put BEFORE the new box.

**Expectation**
A call to `Slate.Transforms.insertNodes(..., { select: true })` should put the cursor inside the new node.

But with slate-react@0.66.4 and above, the cursor is placed BEFORE the new node.

**Environment**
- Slate Version: 0.66.4 and above (i.e. any version with https://github.com/ianstormtaylor/slate/pull/4304; reverting this one line fixes the bug)
- Operating System: macOS
- Browser: Chrome, Firefox, Safari

**Context**

The feature I'm implementing is a Notion-style "commander", where you type something like "/math", hit return, and it executes some command, like inserting an inline math element.

This is probably related to https://github.com/ianstormtaylor/slate/issues/4601, another bug caused by https://github.com/ianstormtaylor/slate/pull/4304

ricardo:
This issue looks related: https://github.com/ianstormtaylor/slate/issues/4642

Issue #4601, opened 2021-10-14
Cursor goes back to the start of the editor after inserting a bullet or number list on Safari


georgesbiaux:
**Description**
Cursor goes back to the start of the editor after inserting a bullet or number list on Safari

**Recording**

https://user-images.githubusercontent.com/22441392/137275736-b5eee0a3-5419-42a2-bf8d-95b5017f81a0.mp4

**Sandbox**
Reproductible on the demo here: https://www.slatejs.org/examples/richtext

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/richtext
2. Click on the insert a bullet list button
3. See cursor go back to start of editor

**Expectation**
Cursor stay at the new list point

**Environment**
- Slate Version: 0.66.5
- Operating System: macOS
- Browser: Safari
- TypeScript Version: 4.4.3

**Context**
Problem was not occurring when slate was in version 0.63.0

georgesbiaux:
@dylans I checkout to the `slate-react@0.67.0` branch, and the problem is still there 😕:

https://user-images.githubusercontent.com/22441392/137927537-a8d2a302-aa3d-4011-a2d8-6dfde091852d.mp4

jameshfisher:
I might take a look at this today. I'm now facing a different bug due to exactMatch:false. There was talk in #4304 of a better solution; I'll try to understand it.

Issue #4602, opened 2021-10-15
Spellcheck On Android: Cursor Position Jumps To Corrected Letter & Inserts Space


robin-macpherson:
**Description**

Firstly, THANK YOU SO MUCH to all those who have contributed to Android support improvements. It is finally usable for us with the most recent releases, but with a small new issue introduced:

As shown in the recording from one of our users, when applying a spellcheck correction on an Android device, the cursor position jumps to the letter that was changed and inserts a space right after it.

**Recording**

https://user-images.githubusercontent.com/34203886/137451687-d4590147-a46b-46ae-82b6-aeca83364581.mp4

**Sandbox**
This should be reproducible on the Slate website: https://www.slatejs.org/examples/richtext

In case this is helpful, our Android test user reported this on our [test environment](https://journaly-git-update-slate-journaly.vercel.app/) (a simple sign up and go to "New Post" to find the editor) for [this PR](https://github.com/Journaly/journaly/pull/670) which simply updates all Slate packages to latest.

**Steps**
To reproduce the behavior:
1. On an Android device, type in the editor and make a spelling mistake
2. Click on the correct suggestion from to correct the spelling
3. Observe behaviour shown in the recording

**Expectation**
See above.

**Environment**
- Slate Version: 0.66.5
- Operating System: Android
- Browser: Chrome

**Context**
Until now we've had major IME and Android support issues where every word was doubled upon input, and deleting texts would insert more words with the cursor jumping all over the place. So this is still a HUGE improvement and finally makes the experience at least usable, albeit frustrating for Android users that are using spellcheck.

robin-macpherson:
Hi @clauderic ! I hope you don't mind me tagging you but I know you've been a huge wealth of knowledge and contribution to this project and especially issues around Android support. I'd be more than happy to try and investigate this but honestly I have zero experience or expertise with Android development - is there any chance that on first glance at this issue, you have an immediate idea of where the problem is likely occurring to point me in the right direction?

Thanks so much for any time or thoughts you might have for this, ami!

robin-macpherson:
@Slapbox I saw you were having a quick look at this, I was just wondering if you have any initial ideas as to what could be causing this or where I could start digging around? I wish I was more familiar with addressing Android-specific issues. No worries at all if you don't have time to get back to me, I just wanted to see if there's anything I can do to support moving this forward in any way.

Thanks for all that you do 🙏🏼

Slapbox:
[...]

PR #4627, opened 2021-10-25
Revert "Do NOT use exact match when updating dom selection"

jameshfisher:
**Description**
The change to `exactMatch: false` in https://github.com/ianstormtaylor/slate/pull/4304 was intended to fix https://github.com/ianstormtaylor/slate/issues/4293, a bug where "backwards typing" happened in nested editors.

But this change has introduced at least two new bugs:

- https://github.com/ianstormtaylor/slate/issues/4601
- https://github.com/ianstormtaylor/slate/issues/4626

These are (IMO) worse than the original "backwards typing" bug.

**Issue**
- https://github.com/ianstormtaylor/slate/issues/4601
- https://github.com/ianstormtaylor/slate/issues/4626

**Context**
From discussion in https://github.com/ianstormtaylor/slate/pull/4304, the true underlying bug is in `ReactEditor.toSlateRange`. I'll attempt to fix this underlying bug instead.

**Checks**

### DIFF
--- a/.changeset/fair-bats-rescue.md
+++ b/.changeset/fair-bats-rescue.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fixed issues where cursor jumps to wrong location


jameshfisher:
@robin-macpherson thanks although all I did was revert previous work 😅  I can't say much about #4602, although my first question would be: does your Android bug still happen with slate-react@0.68.1 and above? If so, it's probably a separate bug ...

robin-macpherson:
@jameshfisher ah, apologies for not fully understanding that 🤦🏻‍♂️ I just got the results from our user who does Android testing for us and it appears the bug does indeed still happen. I'll keep digging around, thanks for replying here!

Issue #4579, opened 2021-10-08
Grammarly autocorrections don't work


churichard:
**Description**
Grammarly autocorrections don't seem to work anymore.

From looking at [this PR](https://github.com/ianstormtaylor/slate/pull/2855), it seems like they did indeed work in a previous version of Slate. And I seem to recall them working not too long ago as well, so this is likely a regression.

Now when you select a correction, the text is inserted at the cursor position and does not overwrite the misspelled word.

**Recording**
![grammarly](https://user-images.githubusercontent.com/4218237/136622334-9ac0c734-3201-4eab-8794-31152b334551.gif)

**Sandbox**
https://www.slatejs.org/examples/richtext

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/richtext
2. Install + enable Grammarly on the website
3. Type a misspelled word
4. Hover over the misspelled word and select a correction
5. Observe that the correction is inserted at the cursor and does not overwrite the misspelled word

**Expectation**
Autocorrections should overwrite the misspelled word.

**Environment**
- Slate Version: 0.66.6
- Operating System: Windows
- Browser: Chrome

dylans:
This looks like the same issue as the currently broken application of suggestions from spell checking.

PR #4650, opened 2021-11-09
Do not try to disable Grammarly in Slate

e1himself:
**Description**

Stop _trying_ to disable Grammarly extension from working in Slate editors.

**Issue**
Fixes: #4124 

**Example**

Peek 2021-11-09 18-19

**Context**

The Grammarly extension is not causing DOM issues anymore, as they have reworked their highlighting logic: 
https://www.grammarly.com/blog/engineering/making-grammarly-feel-native-on-every-website/

Also the `data-gramm` attribute (introduced in #733) is no longer working actually (see #4124), as the Grammarly team changed the name.

Though applying Grammarly suggestions is still broken: #4579 (but it does not break the editor, so it may be fine).

**Checks**

### DIFF
--- a/.changeset/wet-lobsters-travel.md
+++ b/.changeset/wet-lobsters-travel.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Do not disable Grammarly extension in Slate editors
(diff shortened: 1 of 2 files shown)


dylans (approved):
Thanks, it does look Grammarly has resolved the issues with their approach.

Issue #4647, opened 2021-11-09
Selection and scroll not working


karthikcodes6:
**Description**

When I try to select multiple blocks it's not smooth and again it's repeating the top section. 

**Recording**
https://user-images.githubusercontent.com/24937683/140888687-a10f7b3e-9321-4c41-b520-7512a8e0177b.mov

**Steps**
To reproduce the behavior:
1. Go to 'https://www.slatejs.org/examples/huge-document'
2. Select and multiple blocks and scroll to the bottom
3. Observe the scrolling behaviour

**Expectation**
The scrolling should be smooth

**Environment**
- Slate Version: latest
- Operating System: iOS
- Browser: Chrome
- TypeScript Version: [e.g. 3.9.7 - required only if it's a TypeScript issue]

karthikcodes6:
It seems like this [PR](https://github.com/ianstormtaylor/slate/pull/4369) is the root cause of this issue.

PR #4652, opened 2021-11-10
fix: default scroll selection

karthikcodes6:
**Description**
Disabled the  auto scroll behaviour when it's on selection

**Issue**
Fixes: (link to issue)
https://github.com/ianstormtaylor/slate/issues/4647

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -1326,15 +1326,19 @@ const defaultScrollSelectionIntoView = (
   editor: ReactEditor,
   domRange: DOMRange
 ) => {
-  if (editor.selection?.anchor?.path[0] !== editor.selection?.focus?.path[0]) {
-    return
+  // This was affecting the selecting multi blocks and dragging behaviour so, enabled only
+  // if the selection is collapsed.
+  if (
+    !editor.selection ||
+    (editor.selection && Range.isCollapsed(editor.selection))
+  ) {
+    const leafEl = domRange.startContainer.parentElement!
+    leafEl.getBoundingClientRect = domRange.getBoundingClientRect.bind(domRange)
+    scrollIntoView(leafEl, {
+      scrollMode: 'if-needed',
+    })
+    delete leafEl.getBoundingClientRect
   }
-  const leafEl = domRange.startContainer.parentElement!
-  leafEl.getBoundingClientRect = domRange.getBoundingClientRect.bind(domRange)
-  scrollIntoView(leafEl, {
-    scrollMode: 'if-needed',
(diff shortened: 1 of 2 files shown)


karthikcodes6:
Let me know if there is any change required. I know we can pass custom function to override this function. But it's better to provide the default behaviour correctly within slate itself.

dylans (approved):

dylans (changes requested):
Hit approve prematurely... in general I think you're on the right track, but there's likely a better way to do this. Just landed after an overnight flight so I'll respond with feedback once my brain is working fully. :)

packages/slate-react/src/components/editable.tsx
```diff
@@ -1326,6 +1326,9 @@ const defaultScrollSelectionIntoView = (
   editor: ReactEditor,
   domRange: DOMRange
 ) => {
+  if (editor.selection?.anchor?.path[0] !== editor.selection?.focus?.path[0]) {
```
nemanja-tosic, line 1329: Not sure if the check is for an expanded range, if so `Range.isExpanded` should be used. If not, a variable should be extracted so that the if is a bit more readable.

karthikcodes6: I forgot that we have helper methods.

karthikcodes6:
@dylans any update on this PR?

packages/slate-react/src/components/editable.tsx
```diff
@@ -1326,12 +1326,19 @@ const defaultScrollSelectionIntoView = (
   editor: ReactEditor,
   domRange: DOMRange
 ) => {
-  const leafEl = domRange.startContainer.parentElement!
-  leafEl.getBoundingClientRect = domRange.getBoundingClientRect.bind(domRange)
-  scrollIntoView(leafEl, {
-    scrollMode: 'if-needed',
-  })
-  delete leafEl.getBoundingClientRect
+  // This was affecting the selecting multi blocks and dragging behaviour so, enabled only
```
dylans, line 1329:
```suggestion
  // This was affecting the selection of multiple blocks and dragging behavior, so enabled only
```

dylans (changes requested):
Overall I think this change is fine (see my grammar suggestion), but I'm not sure if it fully fixes the issue as there might be other scenarios leading to this issue. But let's go ahead and land this and see if there are remaining scenarios that break scrolling.

dylans:
P.S. This one needs a changeset as well, thanks!

karthikcodes6:
I added changeset, thanks again.

dylans (approved):

PR #4653, opened 2021-11-11
Adds clarification & examples to demystify Transforms.

DougReeder:
**Description**
This expands the (somewhat terse) documentation

### DIFF
--- a/docs/api/transforms.md
+++ b/docs/api/transforms.md
@@ -1,4 +1,4 @@
-# Transforms
+# Transforms API
 
 Transforms are helper functions operating on the document. They can be used in defining your own commands.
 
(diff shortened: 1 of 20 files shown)


docs/api/transforms.md
```diff
@@ -1,4 +1,4 @@
-# Transforms
+# Transforms API
```
dylans, line 1: I don't think we call the other API sections as API, so I would either change all of them or revert this.

docs/concepts/04-transforms.md
```diff
@@ -4,6 +4,40 @@ Slate's data structure is immutable, so you can't modify or delete nodes directl
...
+  match: node =>
+    !Editor.isEditor(node) &&
+    node.children?.every(child => Editor.isBlock(editor, child)),
+  mode: 'all', // also the Editor's children
+})
+```
+
+Non-standard operations (or debugging/tracing which Nodes will be affected by a set of NodeOptions) may require using
+`Editor.nodes` to create a JavaScript Iterator of NodeEntries and a for..of loop to act.
+Here's code to replace all image elements with their alt text:
```
dylans, line 22:
```suggestion
For example, to replace all image elements with their alt text:
```

dylans (changes requested):
A couple of small suggestions, and then I think this is good to land. Thanks @DougReeder !

docs/api/transforms.md
DougReeder: I've added "API" to all of the API documents, to distinguish them from the Concepts documents.

dylans (approved):

Issue #4660, opened 2021-11-16
Could not copy content out of editable in iframe


leoc4e:
**Description**
Could not copy content out of editable in iframe

**Recording**
![capture](https://user-images.githubusercontent.com/20469933/141957902-7bd719e9-2bc8-461b-bd02-bf4acf102ca6.gif)

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/iframe
2. Copy the content using keyboard shortcut or mouse
3. Copy the content to anywhere only support clipboard data `text/plain` would not work

**Expectation**
Should be able to copy the text to anywhere

**Environment**
- Slate Version: 0.70
- Operating System: window
- Browser: Chrome

**Context**
PR:

PR #4661, opened 2021-11-16
fix: use ownerDocument to create element

leoc4e:
**Description**
Fixed the issue that copying from editable inside iframe does not work
https://github.com/leoc4e/slate/blob/d06706c9e15bbbdd7cdd9a1bbb38c87d37c85ea1/packages/slate-react/src/utils/dom.ts#L66-L70

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4660

**Example**
capture

**Context**
The issues is due to `getPlainText` is returning empty string in iframe.
https://github.com/leoc4e/slate/blob/272ad7ffddae8d1163e395a44dd59f66e2b0f1b0/packages/slate-react/src/plugin/with-react.ts#L191-L192

This is because `isDomNode` would check if the element is an instance of the `ownerDocument.defaultValue.Node`
(https://github.com/leoc4e/slate/blob/d06706c9e15bbbdd7cdd9a1bbb38c87d37c85ea1/packages/slate-react/src/utils/dom.ts#L66-L70)
While slate is using default document when creating the element.
This PR change to use the `ownerDocument` instead

**Checks**
-  [x] The tests pass with `yarn test`.

### DIFF
--- a/.changeset/warm-dolphins-roll.md
+++ b/.changeset/warm-dolphins-roll.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+use ownerDocument to create element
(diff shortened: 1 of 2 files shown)


dylans (changes requested):
Please add a changeset and then I'll land this. Thanks for the contribution @leoc4e .

The instructions are pretty straightforward for changesets, but if you have any questions just let me know.

dylans (approved):

PR #4662, opened 2021-11-16
fix: isBlockActive should use Array.from()

greypants:
**Description**
The [richtext.tsx](https://github.com/ianstormtaylor/slate/blob/main/site/examples/richtext.tsx#L99) example `isBlockActive`  was not working for me in my environment because `Editor.nodes` returns a Generator, not an Array. So `isBlockActive` always returned false. Wrapping it in `Array.from` fixes the example. 

**Issue**
Fixes: #4663

**Example**
| Before | After |
| -- | -- | 
| before | after | 
| `isBlockActive` always returns `false` | `isBlockActive` properly toggles |

**Checks**

### DIFF
--- a/site/examples/richtext.tsx
+++ b/site/examples/richtext.tsx
@@ -96,11 +96,11 @@ const isBlockActive = (editor, format) => {
   const { selection } = editor
   if (!selection) return false
 
-  const [match] = Editor.nodes(editor, {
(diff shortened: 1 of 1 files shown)


dylans (changes requested):
Please run prettier on your PR and then I'll merge it. Thanks @greypants.

PR #4664, opened 2021-11-17
A start to rework decorations into slate core

bryanph:
**Description**
Rework decorations

Instead of having decorations in react land, make them a slate core feature.


bryanph:
oops didn't mean to open this PR here

PR #4669, opened 2021-11-19
fix: flush selection change on before input

BitPhinix:
**Description**
Fixes a regression introduced by https://github.com/ianstormtaylor/slate/pull/4132.

Some IMEs/Chrome extensions like e.g. Grammarly set the selection immediately before triggering a `beforeinput` expecting the change to be applied to the immediately before set selection. Defering this selection update causes the change to be applied to an outdated selection.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4579

**Example**

Before:
Screen Recording 2021-11-19 at 13 13 16

After:
Screen Recording 2021-11-19 at 13 12 07

**Checks**

### DIFF
--- a/.changeset/chilled-bears-lick.md
+++ b/.changeset/chilled-bears-lick.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Flush scheduleOnDOMSelectionChange on beforeinput
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -10,6 +10,7 @@ import {
   Path,
 } from 'slate'
 import getDirection from 'direction'
+import debounce from 'lodash/debounce'
 import throttle from 'lodash/throttle'
 import scrollIntoView from 'scroll-into-view-if-needed'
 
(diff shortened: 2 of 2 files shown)


dylans (commented):
@BitPhinix thanks, overall this looks good! Our integration test for triple-clicking seems to be failing. These tests are pretty flaky in general, but if you could give a quick test to make sure that this change doesn't break triple-click behavior that would be greatly appreciated.

BitPhinix:
Just checked, seems to be working just fine. I think it's just flaky ci.

dylans (approved):

dylans:
@BitPhinix did you by chance check this on macOS? I'm having intermittent failures with the IME keyboard not showing up until I defocus slate. Trying to track down if it's this change or #4671, but I'm guessing it's this one unfortunately.

BitPhinix:
Could you send a video of the issue? I tried writing with several IMEs (Japanese etc.) and it seemed to be working just fine

dylans:
I really don't know what to say, yesterday I could not get the IME to pull up reliably on slate documents, today things are fine. No reboot, etc. So I'm guessing things are fine.

PR #4672, opened 2021-11-19
Fix deep-equals for array properties

VictorBaron:
**Description**
deepEquals was always returning true if any array properties were equals.

**Issue**
Fixes: (link to issue)

**Example**
Code example :

```ts
const A = {
  array: [1],
  otherProp: true
};
const B = {
  array: [1],
  otherProp: false
};
deepEquals(A, B); // Should return false... but was returning true.
```

**Context**

**Checks**

### DIFF
--- a/.changeset/merge-nodes-issue
+++ b/.changeset/merge-nodes-issue
@@ -1,5 +1,0 @@
----
-"slate": patch
----
-
-Do not remove empty node in merge operation if it is first children in its parent.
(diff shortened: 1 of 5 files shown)


dylans (approved):
Thanks, I will merge this after we do our next patch release.

VictorBaron:
Cool ! Thanks for the quick review !
Not gonna lie this fixes a bug on my side as well.

PR #4674, opened 2021-11-20
Fix logic error in isDeepEqual

SmilinBrian:
**Description**

If an object being tested had an array property that matched, the test was cut short and true returned. If subsequent properties did not match, the result would be wrong.

**Example**
If a developer used a Mark or something where the value was an array, the `isDeepEqual()` utility function would incorrectly return `true` as soon as it found that the array property matched the other object's property.

**Checks**

### DIFF
--- a/.changeset/slow-mangos-know.md
+++ b/.changeset/slow-mangos-know.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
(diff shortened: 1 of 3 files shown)


SmilinBrian:
Oh wow, #4672 beat me to it! Well done @VictorBaron !

Issue #4657, opened 2021-11-13
Warning: Extra attributes from the server: spellcheck,autocorrect,autocapitalize with Next


tiavina-mika:
**Description**

**Recording**
![image](https://user-images.githubusercontent.com/42656064/141642185-ae339232-4cbe-424c-b894-a41fb4349351.png)

**Sandbox**
[https://codesandbox.io/s/determined-architecture-06e8p](https://codesandbox.io/s/determined-architecture-06e8p)
**Steps**
To reproduce the behavior:
1. Go to first loaded page (home)
2. Open the btowser console
3. See warning

**Expectation**
Should be no warning

**Environment**
- Slate Version: 0.70
- Operating System: Windows 10
- Browser: Chrome
- Next.js: 12.0

**Context**
The server rendered all attributes from the server, those attributes may not be used when rendered in the client side

matthewkeil:
I was able to solve this with 
```
<Editable
          className={styles.editable}
          spellCheck={false}
          autoCorrect="false"
          autoCapitalize="false"
        />
```

tiavina-mika:
thanks, it's working

matthewkeil:
created PR https://github.com/ianstormtaylor/slate/pull/4682

Issue #4318, opened 2021-06-03
Server side rendering results in spellCheck, autoCorrect, and autoCapitalize mismatch


daniel-keller:
**Description**
Using SlateJS in server side rendering (in this case through NextJS) results in mismatching spellCheck, autoCorrect, and autoCapitalize between server and client. The result is these features fail to work in slate editors that are rendered server side. SlateJS's [Editable](https://github.com/ianstormtaylor/slate/blob/411e5a193bd639fb743c2253a5f5e43a5949b100/packages/slate-react/src/components/editable.tsx) component sets these properties based on `HAS_BEFORE_INPUT_SUPPORT` which (I assume) isn't available server side. 

**Recording**
<img width="1431" alt="Screen Shot 2021-06-03 at 11 04 24 AM" src="https://user-images.githubusercontent.com/4036797/120667451-80fc0480-c45b-11eb-87af-f97f9cea8707.png">

**Sandbox**
https://codesandbox.io/s/github/daniel-keller/nextjs-slatejs-spellcheck-bug

**Steps**
To reproduce the behavior:
1. Server side render the Editable component with spellCheck, autoCorrecct, or autoCapitalize set to true (or 'true' for spellCheck)

**Expectation**
spellcheck, autocorrect, and autocapitalize have worked in server side rendering in previous versions of SlateJS (0.58.0). 

**Environment**
- Slate Version: 0.65.0
- Operating System: iOS
- Browser: Chrome, Safari, Firefox
- TypeScript Version: N/A

cloudre01:
facing the same issue here. Using the version 0.58.x solved this issue when using on Nextjs.

matthewkeil:
I solved this on slate@0.70.0 using.

```typescript
<Editable
  className={styles.editable}
  spellCheck={false}
  autoCorrect="false"
  autoCapitalize="false"
/>
```

matthewkeil:
added PR https://github.com/ianstormtaylor/slate/pull/4682

PR #4682, opened 2021-11-22
fix ssr prop mismatch

matthewkeil:
**Description**
Fixes 2 issue for Next.js (or any SSR method).

1) Under SSR the Editable component requires three props to be added to avoid a warning in the console.  This is because defaults are added to the component to deal with `HAS_BEFORE_INPUT_SUPPORT.` They are added because the check fails server side and the warning arrises because the check works client side, resulting in the prop mismatch.  The result is to force SSR users to add the following to avoid the warning.
```
<Editable
  spellCheck={false}
  autoCorrect="false"
  autoCapitalize="false"
>
```
Screen Shot 2021-11-21 at 9 04 40 PM

2) SSR users are not able to use `spellCheck, autoCorrect or autoCapitalize.` If the developer would like to opt-in to these features they can set them to true and the library will verify compatibility.  However it is not possible to SSR render the application because `HAS_BEFORE_INPUT_SUPPORT` is false on the server which overrides the develop props. If a developer/consumer would like to attempt to use this feature, `HAS_BEFORE_INPUT_SUPPORT` results in false on the server and true on the browser which throws a warning for prop mismatch.
```
<Editable
  spellCheck={true}
  autoCorrect="true"
  autoCapitalize="true"
>
```
Screen Shot 2021-11-21 at 9 04 16 PM

**Issue**
Fixes: 
[Issue 4318](https://github.com/ianstormtaylor/slate/issues/4318)
[Issue 4657](https://github.com/ianstormtaylor/slate/issues/4657)

**Example**
Screen Shot 2021-11-21 at 9 02 29 PM
Screen Shot 2021-11-21 at 9 03 35 PM
Screen Shot 2021-11-21 at 9 35 21 PM
Screen Shot 2021-11-21 at 9 04 16 PM
Screen Shot 2021-11-21 at 9 04 40 PM

**Context**
Only made change to how props are added to the component, did not change value check `HAS_BEFORE_INPUT_SUPPORT` which is used in other places and could potentially change how other parts of the render happen.

Made this change to support developer ergonomics for SSR.  No longer need to added the spellCheck, auotCorrect and autoCapitalize props to the component.  Also made change to support users as [ 94.03% of users](https://caniuse.com/?search=beforeinput) will miss out on the opportunity to use spellCheck, autoCorrect, autoCapitalize with a text editing library when used under SSR situations.  Only drawback/tradeoff for this change is 6% of users will see a prop mismatch warning under SSR conditions (similar to issue 1) as the server will add the positive props and the client will not pass the `HAS_BEFORE_INPUT_SUPPORT` check.

**Checks**

### DIFF
--- a/.changeset/strong-jeans-train.md
+++ b/.changeset/strong-jeans-train.md
@@ -1,0 +1,12 @@
+---
+'slate-react': minor
+---
+
+Support SSR for autoCorrect, spellCheck and autoCapitalize.
+Fixes prop mismatch between server and client.
+Removes the need to add
+<Editable
+  spellCheck={false}
+  autoCorrect="false"
+  autoCapitalize="false"
+/>
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -24,6 +24,7 @@ import {
   IS_FIREFOX_LEGACY,
   IS_QQBROWSER,
   IS_SAFARI,
+  CAN_USE_DOM,
 } from '../utils/environment'
 import { ReactEditor } from '..'
 import { ReadOnlyContext } from '../hooks/use-read-only'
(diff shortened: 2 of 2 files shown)


dylans (changes requested):
Thanks @matthewkeil for the thorough explanation and clear PR. If you could please add a changeset (there's a link in the PR), I'll get this landed and include it in the next release. I would choose minor as the option for the impact of this change.

matthewkeil:
@dylans thanks!  Changeset included.  Just found this package this weekend and am incorporating it into a new project I'm starting.  So far so good.

dylans (approved):

PR #4688, opened 2021-11-24
Adding update on library import to avoid error message

YoannBuzenet:
The error message ' The package "emotion" has been replaced by "@emotion/styled" ' appears with the old one. With this, it compiles successfully and shows no errors (I'm on next js12).

Great lib, thanks!

**Description**
error message pops when copying the code from [components used in rich text](https://github.com/ianstormtaylor/slate/blob/main/site/components.tsx) example page.

**Issue**
Fixes: update the import on emotion library

**Context**
I'm on nextjs 12, I don't know if it matters.

Thanks for your work, I hope I'll really help later

### DIFF
--- a/site/components.tsx
+++ b/site/components.tsx
@@ -1,6 +1,6 @@
 import React, { Ref, PropsWithChildren } from 'react'
 import ReactDOM from 'react-dom'
-import { css, cx } from @emotion/css';
+import { css, cx } from '@emotion/css';
 
 interface BaseProps {
   className: string
--- a/site/components.tsx
+++ b/site/components.tsx
@@ -1,6 +1,6 @@
 import React, { Ref, PropsWithChildren } from 'react'
 import ReactDOM from 'react-dom'
-import { css, cx } from "@emotion/css";
+import { css, cx } from @emotion/css';
 
 interface BaseProps {
   className: string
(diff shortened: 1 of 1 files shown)


dylans (approved):

YoannBuzenet:
Did it miss because of missing types ? I did it without typescript.

I could have a look at which types it needs ?

site/components.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { Ref, PropsWithChildren } from 'react'
 import ReactDOM from 'react-dom'
-import { cx, css } from 'emotion'
+import { css, cx } from "@emotion/css";
```
dylans, line 3:
```suggestion
import { css, cx } from @emotion/css';
```

I think it was failing because the change violated prettier's rules.

dylans (changes requested):

site/components.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { Ref, PropsWithChildren } from 'react'
 import ReactDOM from 'react-dom'
-import { cx, css } from 'emotion'
+import { css, cx } from @emotion/css';
```
dylans, line 3:
```suggestion
import { css, cx } from '@emotion/css';
```

dylans (changes requested):
This is likely failing because it's missing an update to the dependency in package.json and yarn. Please fix and update with a complete PR and then I'll review it. Thanks!

dylans:
Superseded by #4783

PR #4695, opened 2021-11-27
Update editable.tsx: auto-focus fix for end of line with several chil…

GrinchakAndrew:
Solution resolves cursor position in the Editor with autofocus set to true, when cursor pos is 0 instead of end of line.


packages/slate-react/src/components/editable.tsx
```diff
@@ -250,10 +250,20 @@ export const Editable = (props: EditableProps) => {
...
   // needs to be manually focused.
   useEffect(() => {
-    if (ref.current && autoFocus) {
-      ref.current.focus()
-    }
-  }, [autoFocus])
+      if (ref.current && autoFocus) {
+        if(window.getSelection && document.createRange) {
+          const range = document.createRange();
+          node.focus();
```
akonsu, line 256: what is `node`?

GrinchakAndrew: @akonsu Sorry, node was a reference to ref.current in my other sandbox. Will correct it now.

dylans (changes requested):
Looks reasonable. Please add a changeset when you have a moment and then we can land this.

packages/slate-react/src/components/editable.tsx
```diff
@@ -250,10 +250,20 @@ export const Editable = (props: EditableProps) => {
...
   // needs to be manually focused.
   useEffect(() => {
-    if (ref.current && autoFocus) {
-      ref.current.focus()
-    }
-  }, [autoFocus])
+      if (ref.current && autoFocus) {
+        if(window.getSelection && document.createRange) {
+          const range = document.createRange();
+          ref.current.focus();
```
akonsu, line 256: Is it essential to call `ref.current.focus()` here and not at the end of this `if` clause? Maybe we can move this call out of the `if` statement?

packages/slate-react/src/components/editable.tsx
```diff
@@ -250,10 +250,20 @@ export const Editable = (props: EditableProps) => {
   // The autoFocus TextareaHTMLAttribute doesn't do anything on a div, so it
   // needs to be manually focused.
   useEffect(() => {
-    if (ref.current && autoFocus) {
-      ref.current.focus()
-    }
-  }, [autoFocus])
+      if (ref.current && autoFocus) {
+        if(window.getSelection && document.createRange) {
```
akonsu, line 254: Please prettify the diff. This `if` needs a space after it.

dylans:
@GrinchakAndrew any chance you have time to wrap this one up. It looks like it needs a linting fix and there are a couple of comments to address. Thanks!

GrinchakAndrew:
This has been already merged, and I think we addressed this with the husky
precommit hook (as per linting) that automates these little things.

packages/slate-react/src/components/editable.tsx
dylans, line 254:
```suggestion
        if (window.getSelection && document.createRange) {
```

dylans:
> > packages/slate-react/src/components/editable.tsx
> 
> Hello, i am `afkb` for the time being. No need to run linting on the whole project. Pls., add it to one line of code.

Unfortunately because you created your PR from your main branch it's not easy to clone your PR locally. But I'll make it work.

dylans (approved):

dylans (changes requested):
This change is causing a regression in the markdown-shortcuts example and is thus failing our integrations tests. Will look when I get a chance, unless someone else has time to track this down before I do.

dylans:
> Is there a recommended node version or caveat to getting it to run locally? (Using node v`14.17.6` and OS `Ubuntu 20.04`)

Hi @JustinBeaudry, thanks, I hope @GrinchakAndrew is ok.

With respect to running locally, they run for me with 14.15.5 on macOS, but there's nothing special about that version. If they won't run locally you can commit to the PR which will trigger a re-run which is not ideal but better than nothing. Feel free to ping me on Slack ( https://join.slack.com/t/slate-js/shared_invite/zt-f8t986ip-7dA1DyiqPpzootz1snKXkw ) if you're stuck.

dylans:
My attempt at using GitHub's tooling to resolve merge conflicts failed badly here. I will look at this later.

dylans:

PR #4702, opened 2021-11-30
Fix CJK IME(like Chinese, Japanese) double input

ttitoo:
**Description**
* Resolve double input when using CJK IME
* adapt UC mobile which will input nothing after selected from the candidates

**Issue**
Fixes: #3914

**Example**
**Before**

**After**

**Context**
CJK IME, like the Mac OS/iOS native ones, will trigger events(compositionstart, compositionend, insertFromComposition), which will invoke "insertText" multiple times if the conditions are not considered.
This PR will skip some insert actions to avoid strange behaviors, like double input, no content after selection

**Checks**

### DIFF
--- a/.changeset/large-melons-warn.md
+++ b/.changeset/large-melons-warn.md
@@ -2,4 +2,4 @@
 'slate-react': minor
 ---
 
-Fix CJK IME(like Chinese, Japanese) double input
+Fix CJK IME (e.g. Chinese or Japanese) double input
--- a/.changeset/large-melons-warn.md
+++ b/.changeset/large-melons-warn.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+Fix CJK IME(like Chinese, Japanese) double input
--- a/packages/slate-react/src/utils/environment.ts
+++ b/packages/slate-react/src/utils/environment.ts
@@ -47,6 +47,10 @@ export const IS_QQBROWSER =
 export const IS_UC_MOBILE =
   typeof navigator !== 'undefined' && /.*UCBrowser/.test(navigator.userAgent)
 
+// Wechat browser
+export const IS_WECHATBROWSER =
+  typeof navigator !== 'undefined' && /.*Wechat/.test(navigator.userAgent)
+
 // Check if DOM is available as React does internally.
 // https://github.com/facebook/react/blob/master/packages/shared/ExecutionEnvironment.js
 export const CAN_USE_DOM = !!(
--- a/packages/slate-react/src/utils/environment.ts
+++ b/packages/slate-react/src/utils/environment.ts
@@ -43,6 +43,10 @@ export const IS_FIREFOX_LEGACY =
 export const IS_QQBROWSER =
   typeof navigator !== 'undefined' && /.*QQBrowser/.test(navigator.userAgent)
 
+// UC mobile browser
+export const IS_UC_MOBILE =
+  typeof navigator !== 'undefined' && /.*UCBrowser/.test(navigator.userAgent)
+
 // Check if DOM is available as React does internally.
 // https://github.com/facebook/react/blob/master/packages/shared/ExecutionEnvironment.js
 export const CAN_USE_DOM = !!(
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -25,6 +25,7 @@ import {
   IS_QQBROWSER,
   IS_SAFARI,
   IS_UC_MOBILE,
+  IS_WECHATBROWSER,
   CAN_USE_DOM,
 } from '../utils/environment'
 import { ReactEditor } from '..'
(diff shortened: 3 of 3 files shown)


363676727:
> **Description**
> 
> * Resolve double input when using CJK IME
> * adapt UC mobile which will input nothing after selected from the candidates
> 
> **Issue** Fixes: #3914
> 
> **Example** **Before** 
> 
> **After** 
> 
> **Context** CJK IME, like the Mac OS/iOS native ones, will trigger events(compositionstart, compositionend, insertFromComposition), which will invoke "insertText" multiple times if the conditions are not considered. This PR will skip some insert actions to avoid strange behaviors, like double input, no content after selection
> 
> **Checks**
> 
> * [x]  The new code matches the existing patterns and styles.
> * [x]  The tests pass with `yarn test`.
> * [x]  The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
> * [x]  The relevant examples still work. (Run examples with `yarn start`.)
> * [x]  You've [added a changeset](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md) if changing functionality. (Add one with `yarn changeset add`.)

It seems to modify the original input logic here

ttitoo:
> image
> 
> It seems to modify the original input logic here

The original logic is working in most cases, just add some exceptions

363676727:
> The original logic is working in most cases, just add some exceptions

I think it means the same code

if (type === 'insertFromComposition') {
  // COMPAT: in Safari, `compositionend` is dispatched after the
  // `beforeinput` for "insertFromComposition". But if we wait for it
  // then we will abort because we're still composing and the selection
  // won't be updated properly.
  // https://www.w3.org/TR/input-events-2/
  state.isComposing && setIsComposing(false)
  state.isComposing = false

  if (!IS_UC_MOBILE) {
    // browsers except UC mobile do nothing to avoid duplicated insertion
    return
  }
}

ttitoo:
> > The original logic is working in most cases, just add some exceptions
> 
> I think it means the same code
> 
> ```
> if (type === 'insertFromComposition') {
>   // COMPAT: in Safari, `compositionend` is dispatched after the
>   // `beforeinput` for "insertFromComposition". But if we wait for it
>   // then we will abort because we're still composing and the selection
>   // won't be updated properly.
>   // https://www.w3.org/TR/input-events-2/
>   state.isComposing && setIsComposing(false)
>   state.isComposing = false
> 
>   if (!IS_UC_MOBILE) {
>     // browsers except UC mobile do nothing to avoid duplicated insertion
>     return
>   }
> }
> ```

Yes, but in the original `state.isComposing && setIsComposing(false)` has a check statement for isComposing, I am not sure if `state.isComposing` is always true.

363676727:
> 

Would it be more reasonable to deal with onCompositionEnd?

```
if (
    !IS_SAFARI &&
    !IS_FIREFOX_LEGACY &&
    !IS_IOS &&
    !IS_QQBROWSER &&
    !IS_UC_MOBILE &&
    event.data
  ) {
    Editor.insertText(editor, event.data)
  }
```

ttitoo:
I think both are ok, personal prefer readable

zhugexinxin:
> I think both are ok, personal prefer readable

Browser differences should be unified and resolved in onCompositionEnd to be more intuitive
you can try the codes
```
const IS_WECHAT = typeof navigator !== 'undefined' && /.*Wechat/.test(navigator.userAgent)

if (
  !IS_SAFARI &&
  !IS_FIREFOX_LEGACY &&
  !IS_IOS &&
  !IS_QQBROWSER &&
  !IS_UC_MOBILE &&
  !IS_WECHAT &&
  event.data
) {
  Editor.insertText(editor, event.data)
}
```

.changeset/large-melons-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+Fix CJK IME(like Chinese, Japanese) double input
```
dylans, line 5:
```suggestion
Fix CJK IME (e.g. Chinese or Japanese) double input
```

dylans (changes requested):
I believe this is generally fine. Do you want to add the wechat browser also per the suggested feedback/comment?

Suggested a small change to make the changelog a bit easier to parse.

.changeset/large-melons-warn.md
zhugexinxin: Now the purpose of processing in beforeinput is for the processing of WeChat browser, So just use it for onCompositionEnd processing, the two are duplicates
```
if (type === 'insertFromComposition') {
  //.....
}
```
insertFromComposition is to adapt to the second level of input that has been implemented, so there should not be redundant processing, and user agent should not be used to judge the browser, this is not a long-term consideration.

zhugexinxin: I contacted ttitoo and confirmed this issue

ttitoo:
Revert to a more specific way to deal with different browsers, but maybe there would be another PRs when new/uncovered browsers  come to the stage

dylans (approved):

samkk:
"slate": "^0.72.3",
"slate-history": "^0.66.0",
"slate-react": "^0.72.1",


There is the same problem, still not fixed, Chinese directly exploded in Firefox
Is there a solution?

ttitoo:
> "slate": "^0.72.3", "slate-history": "^0.66.0", "slate-react": "^0.72.1",
> 
> There is the same problem, still not fixed, Chinese directly exploded in Firefox Is there a solution?

OS？
Test on Mac OSX, Firefox(91.0.2), native Chinese input/Squirrel, it is working

bingluen:
https://user-images.githubusercontent.com/6832868/170954162-ec267ece-5d5f-4bba-a9d7-cbfa8cb37a0e.mov

Still broken at  some situation on Firefox 100.0.2.

Error message:
```
Cannot resolve a DOM point from Slate point: {"path":[0,0],"offset":7}
```

Is there exists some issue when slate updating DOM data ?

packages/slate-react/src/utils/environment.ts
```diff
@@ -43,6 +43,14 @@ export const IS_FIREFOX_LEGACY =
 export const IS_QQBROWSER =
   typeof navigator !== 'undefined' && /.*QQBrowser/.test(navigator.userAgent)
 
+// UC mobile browser
+export const IS_UC_MOBILE =
+  typeof navigator !== 'undefined' && /.*UCBrowser/.test(navigator.userAgent)
+
+// Wechat browser
+export const IS_WECHATBROWSER =
+  typeof navigator !== 'undefined' && /.*Wechat/.test(navigator.userAgent)
```
mengxingshike2012, line 52: /.*Wechat/i is better  becouse of  Enterprise WeChat.

dylans: Please open a new PR, as conversations in closed PRs generally don't get noticed. Thanks!

PR #4706, opened 2021-12-02
Update android restoreDOM to allow partial dom restoring

BitPhinix:
**Description**
The current android restoreDOM implementation always restores the entire editor. This has a drastic negative impact on editor performance.

**Context**
This PR updates the approach we are using to restore the dom to only include the minimal amount of nodes required for react to be able to reliably restore the dom to the desired state. It does so by listening for user-caused mutations, finding the closest know parent node, and only restoring mutated nodes on `restoreDOM`

**Checks**


dylans (approved):
Thanks @BitPhinix !

thesunny:
Awesome @BitPhinix !

So good to see more updates on Android that I'm not involved in! Thank you so much for taking the time to dive into Android and contribute. The hope has always been to have more people understanding and working on Android.

Issue #4694, opened 2021-11-26
Drag & drop doesn't create a new history batch, undo deletes text


djagya:
**Description**

When using Slate with `slate-history`, you get a list of `undo` batches. The problem: When you drag & drop a text/element, that action doesn't create a new history batch, instead writing it in the current one. If you then "undo" the drag & drop, it will undo everything in that batch -- deleting a text/element, or un-applying marks, etc.

Drag & drop should create a new batch in the `undo` list.

**Recording**

On recording I write some text (one batch in `undo`), then drag & drop it to the first paragraph, then "undo" - the text gets deleted. I then "redo" - and the text appears where I dropped it, no at the original place.

![Nov-25-2021 17-32-55](https://user-images.githubusercontent.com/2600431/143531953-44f053fb-374d-4d38-b16b-f2a545f0456b.gif)

**Sandbox**
https://codesandbox.io/s/slate-history-dnd-bug-repro-45qd3

**Steps**
To reproduce the behavior:
1. Go to https://codesandbox.io/s/slate-history-dnd-bug-repro-45qd3
2. Create a new paragraph in the editor
3. Type som text
4. Select it, long-press on selection, drag & drop it to the first paragraph
5. Now undo (Cmd + z on Mac)
6. See the text you typed gets deleted

**Expectation**

Drag & drop action should create a new batch in the history's `undo` list. Undoing drag & drop should return the element back to the original place. Only the second "undo" should delete a text/element.

**Environment**
- Slate Version: 71.0, Slate history: 0.66.0
- Operating System: MacOS
- Browser: Chrome 96
- TypeScript Version: [e.g. 3.9.7 - required only if it's a TypeScript issue]

**Context**

I would be glad to contribute and create a PR, but a pointer to where potentially the problem lies would be appreciated! IIUC there's nothing wrong in the history plugin, it's something we need to fix around how Slate transforms nodes when drag & dropping.

imdbsd:
im also facing this, when try to debug, found out the history was splitted between delete and inserted data that has been dropped. i think it because this line https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/editable.tsx#L965.

Not sure if it should use `moveNodes` transformation instead delete then insert it, but i cannot figure out how to move the dragged data with moveNodes. And not sure why it not merged in one history on history-plugin here https://github.com/ianstormtaylor/slate/blob/main/packages/slate-history/src/with-history.ts#L82

[...]

Issue #4559, opened 2021-09-29
deleteFragment undo should select re-inserted fragment


bryanph:
**Problem**
Currently when selecting text, deleting it and then undoing, the selection is moved to the end of the re-inserted text. The common behavior on many rich text editors is instead to re-select the deleted text.

**Solution**
Re-select the inserted text if it results from an undo

**Context**
**Before**:

![abc](https://user-images.githubusercontent.com/3816712/135244238-a1e7aaa0-3cc7-46d6-a20b-c27016c15d35.gif)
**Expected outcome**:
![image](https://user-images.githubusercontent.com/3816712/135244055-17297c2b-656c-4a84-a9c8-913ed2693d1d.png)

Issue #3756, opened 2020-06-30
slate-history does not correctly undo selection movement


TheSpyder:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?

After performing actions that modify a ranged selection, undo does not restore the selection correctly. In my replication case, selecting across paragraphs and then deleting does not undo correctly. The replication instructions are in the sandbox (and recorded in the gif).
https://codesandbox.io/s/undo-selection-bug-replication-case-wtkbr
![undo selection bug](https://user-images.githubusercontent.com/298292/86175375-8f181580-bb66-11ea-9195-926cdecb5584.gif)
Slate: 0.58.3
Browser: Chrome / Safari / Firefox
OS: Mac

#### What's the expected behavior?

Undo should restore the cursor exactly. I found a workaround - more on that along with a codesandbox link after the gif.
![undo selection expected behaviour](https://user-images.githubusercontent.com/298292/86175453-b969d300-bb66-11ea-8235-2196450938e2.gif)

The problem is that in scenarios such as this delete, Slate generates partial `set_selection` operations (similar to the issue raised in #3705). These are stateful operations making them very sensitive to the order they are applied.

[My workaround sandbox](https://codesandbox.io/s/undo-selection-bug-workaround-pjsf0) logs the order of operations in the history batch, and it's quite weird:
```
[
  {"type":"remove_text","path":[0,0],"offset":5,"text":" line"},
  {"type":"remove_text","path":[1,0],"offset":0,"text":"seco"},
  {"type":"merge_node","path":[1],"position":1,"target":null,"properties":{}},
  {"type":"set_selection","properties":{"anchor":{"path":[0,0],"offset":5}},"newProperties":{"anchor":{"path":[0,1],"offset":0}}},
  {"type":"merge_node","path":[0,1],"position":5,"target":null,"properties":{}}
]
```
Why is the `set_selection` in between the two `merge_node` operations? Perhaps only Ian knows for sure. Regardless, when undo is performed - which inverts these operations and applies them in reverse order - the partial selection update doesn't work. The actual work performed during the undo is:
- split the text node at `[0,0]` offset `5`
- move the anchor back to the end of the split text node (focus remains at the start of the new text node created by the split)
- split the block at `[0]` putting one text node in each block
- insert the removed text in the second block
- insert the removed text in the first block

[...]

Issue #3534, opened 2020-03-04
Selection is broken after undoing


hadrysmateusz:
#### Do you want to request a _feature_ or report a _bug_?

Bug

### Steps to reproduce
- select text spanning multiple block elements ( with both ends in the middle of the line)
- press `enter` or `backspace` or `delete`
- press `ctrl + z` to undo
- the selection now starts at the end of the top-most block instead of where it was originally (you can verify that by starting to type)

[Sandbox](https://codesandbox.io/s/slate-reproductions-y4brr)

![selection issue gif](https://user-images.githubusercontent.com/26979471/75919887-a536f300-5e5e-11ea-8851-0709fd59beac.gif)

Firefox 73.0.1
Chromium 80.0.3987.122

Slate: 0.57.1 (It also happens in 0.54.2 which is in the sandbox)
Browser: Chrome / Firefox / (Haven't tested it in other browsers yet)
OS: Linux

#### What's the expected behavior?

The selection should be what it was before inserting the line break / deleting the text

skogsmaskin:
Possibly related https://github.com/ianstormtaylor/slate/issues/3499

PR #4717, opened 2021-12-05
Change how slate-history handles selection undo

bryanph:
**Description**
This is a proposal, curious what you all think.

As @TheSpyder mentioned in https://github.com/ianstormtaylor/slate/issues/3756#issuecomment-737022068, often times normalization logic can trigger `set_selection` operations that modify the selection state in the middle of an operation stack. Also, the `set_selection` operation often only contains the `anchor` OR the `focus` property because partial updates are allowed. As the current handling of the `set_selection` operation in `slate-history` overrides the previous op, this information is often lost as you apply more operations and undo them.

Instead of using operations to reset selection, this change stores a separate `selectionBefore` property as part of the undo batch that contains the selection as it was before the operations were applied. This works around some of the complexities with undoing that were well described by @TheSpyder in https://github.com/ianstormtaylor/slate/issues/3756 by explicitly storing the selection state as part of the undo. This change also means that `set_selection` operations are no longer stored in the history operations stack. Instead a full selection is included with every undo batch.

With the limited testing that I have done, the results seem to be what I would expect from an undo. Could add the same behavior for redo as well perhaps as in some cases I've noticed it doesn't reset the selection quite well (you can notice in the video when I redo the mark change that it doesn't reselect the remarked selection after changing the selection). Besides mark changes, I couldn't find another redo case that doesn't set the selection properly as besides the mark operations the selection is normally collapsed when redoing. EDIT: fixed this case in https://github.com/ianstormtaylor/slate/pull/4717/commits/a7385110db58e85f94057a42f9bbef743cfc472e

**Issue**
Fixes:  #4694, https://github.com/ianstormtaylor/slate/issues/4559, https://github.com/ianstormtaylor/slate/issues/3756, https://github.com/ianstormtaylor/slate/issues/3534, probably more

**Example**
The cases that were mentioned in https://github.com/ianstormtaylor/slate/issues/3756 work well from my testing

https://user-images.githubusercontent.com/3816712/144755757-e73d1d29-8189-4f3d-80e0-52fcb3ddf749.mp4

**Checks**

### DIFF
--- a/.changeset/cuddly-kiwis-work.md
+++ b/.changeset/cuddly-kiwis-work.md
@@ -1,0 +1,5 @@
+---
+'slate-history': minor
+---
+
+Changes how selections are stored in the history resulting in more consistent results
--- a/packages/slate-history/src/history-editor.ts
+++ b/packages/slate-history/src/history-editor.ts
@@ -9,7 +9,6 @@ export const HISTORY = new WeakMap<Editor, History>()
 export const SAVING = new WeakMap<Editor, boolean | undefined>()
 export const MERGING = new WeakMap<Editor, boolean | undefined>()
 
-
 /**
  * `HistoryEditor` contains helpers for history-enabled editors.
  */
--- a/packages/slate-history/src/history-editor.ts
+++ b/packages/slate-history/src/history-editor.ts
@@ -9,6 +9,7 @@ export const HISTORY = new WeakMap<Editor, History>()
 export const SAVING = new WeakMap<Editor, boolean | undefined>()
 export const MERGING = new WeakMap<Editor, boolean | undefined>()
 
+
 /**
  * `HistoryEditor` contains helpers for history-enabled editors.
  */
--- a/packages/slate-history/src/history.ts
+++ b/packages/slate-history/src/history.ts
@@ -1,14 +1,19 @@
 import { isPlainObject } from 'is-plain-object'
-import { Operation } from 'slate'
+import { Operation, Range } from 'slate'
 
 /**
  * `History` objects hold all of the operations that are applied to a value, so
  * they can be undone or redone as necessary.
  */
 
+interface Batch {
+    operations: Operation[];
+    selectionBefore: Range | null
+}
+
 export interface History {
-  redos: Operation[][]
-  undos: Operation[][]
+  redos: Batch[]
+  undos: Batch[]
 }
 
 export const History = {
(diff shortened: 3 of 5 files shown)


dylans (commented):
It's really interesting and could be a nice simplification, but I don't feel qualified to make that call.

I would love to hear feedback from @BitPhinix @nemanja-tosic @TheSpyder and others that have managed collab editing with Slate.

TheSpyder:
I saw this and meant to reply. Sorry I forgot for over two months. The way I'm doing undo (I don't use `slate-history`) isn't far off this, and I admit the idea isn't something I've considered.

One problem I see is that while this does move selection storage out to a referenced attached to the undo batch, it doesn't remove selection operations from the batch. So the overall undo behaviour might still be weird; perhaps filter out selection operations from the batch.

It also doesn't help me directly as my undo implementation runs through `onChange` to make it easier to tie in to collab features 😂 but this might be a reason to change that (this technique can only work through overriding `apply` to store selection before anything changes).

JosNun:
Fwiw, this would also fix the exception that can be caused in #4842

bryanph:
I've been using this change for a while now and haven't found any issues with it so far. @TheSpyder `set_selection` ops are no longer saved with this change, see https://github.com/ianstormtaylor/slate/blob/3f65bf09abf5b6ee416abd9ff37237110b787603/packages/slate-history/src/with-history.ts#L148-L154

TheSpyder:
> I've been using this approach for a while now and haven't found any issues with it so far. @TheSpyder `set_selection` ops are no longer saved with this change, see
> 
> https://github.com/ianstormtaylor/slate/blob/3f65bf09abf5b6ee416abd9ff37237110b787603/packages/slate-history/src/with-history.ts#L148-L154

TheSpyder:
> `set_selection` ops are no longer saved with this change

Right, sorry, I keep forgetting this is much easier to do with `apply`. I swapped mine to `onChange` as I mentioned and `set_selection` can appear anywhere in a change array. Removing them would be a bit more complicated.

I will need to give some thought to whether I want to change my history implementation. The beauty of `onChange` is Slate guarantees the document will be normalised when it fires. I've tried to avoid overriding `apply` for that reason.

BitPhinix:
I don't really see an issue with this approach as well. It's along the same lines what slate-yjs does under the hood and it seems to be working fine. Take my opinion with a grain of salt though since I don't actively use slate-history.

bryanph:
@dylans I think this can be merged, it doesn't really affect any crdt implementations because those will have their own custom history code anyway.

dylans:
@bryanph great! could you add a quick changeset and then I'll land it.

bryanph:
@dylans done

PR #4720, opened 2021-12-07
Add origin event type to setFragmentData to be able to distinguish copy, cut and drag

bryanph:
**Description**
Currently in order to know whether `setFragmentData` was called as a result of a copy, cut or drag; you have to overwrite all of the `onCut`, `onCopy`, `onDrag` methods and call your own function with the original event. This change adds a single property to `setFragmentData` called `originEvent` that can be set to `copy`, `cut`, `drag`. This allows you to set some extra data (or emit data) on the `DataTransfer` object based on whether `setFragmentData` was called as part of a drag, copy or cut event without having to modify a lot of code. We could potentially do the same thing for `insertData` as the same thing applies there but for `paste` and `drop`.

**Issue**
Fixes: (link to issue)

**Checks**

### DIFF
--- a/.changeset/brave-emus-refuse.md
+++ b/.changeset/brave-emus-refuse.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Add origin event type to setFragmentData to be able to distinguish copy, cut and drag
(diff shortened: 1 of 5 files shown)


dylans (changes requested):
I like this and would find it useful.

My primary suggestion would be to perhaps have a default and/or make it optional so that we don't break users migrating forward if possible.

Also I would say this should be a minor change (major in post 1.0 semver) rather than a patch change in its current form (though maybe it is patch level if we can address the above).

bryanph:
@dylans Hmm yeah good point making it optional is probably fine

dylans (approved):

Issue #4726, opened 2021-12-09
"Cannot resolve from DOM point" error on onDomSelectionChange for readonly void elements


ahoisl:
**Description**
The "Cannot resolve a Slate point from DOM point" error is thrown in the onDomSelectionChange event handler for readonly editors when void elements are to be selected, e.g. when clicking next to an image.

In general, this is just a problem because of selection causing the "Cannot resolve" error and there may be a more general and better approach to fix as outlined in #4643 

**Recording**
![void_error_select](https://user-images.githubusercontent.com/32290337/145485418-47802e87-b2bd-480a-8a63-552a87d20fd7.gif)

**Sandbox**
Easy to create with samples by just modifying 1 line:
* Use the "Images" example in the slate repository
* Mark the Editable `readonly` in `images.tsx:33`

**Steps**
To reproduce the behavior:
1. Use the "Images" example in the Slate repository, modified to be readonly (as explained above)
2. Go to the readonly "Images" example 
2. Click next to the first image, i.e. not directly on it, but slightly right of it (the error also happens if you select a part of the text AND the image)
3. See the dev console for the error

**Expectation**
No error, also no selection or anything.

**Environment**
- Slate Version: 0.72.0
- Operating System: Windows
- Browser: Firefox, Chrome

**Context**
See #4727 for a potential fix

PR #4727, opened 2021-12-09
Fix "Cannot resolve from DOM point" error on onDomSelectionChange for readonly void elements

ahoisl:
**Description**
Fixes "Cannot resolve from DOM point" error on onDomSelectionChange for readonly void elements (see #4726). The code tries to update selection and falsely elects void elements in a readonly editor.

**Issue**
Fixes: #4726 

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -289,11 +289,11 @@ export const Editable = (props: EditableProps) => {
 
         const anchorNodeSelectable =
           hasEditableTarget(editor, anchorNode) ||
-          isTargetInsideVoid(editor, anchorNode)
+          isTargetInsideNonReadonlyVoid(editor, anchorNode)
 
         const focusNodeSelectable =
           hasEditableTarget(editor, focusNode) ||
-          isTargetInsideVoid(editor, focusNode)
+          isTargetInsideNonReadonlyVoid(editor, focusNode)
 
         if (anchorNodeSelectable && focusNodeSelectable) {
           const range = ReactEditor.toSlateRange(editor, domSelection, {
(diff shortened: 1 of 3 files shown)


dylans (approved):

packages/slate-react/src/components/editable.tsx
```diff
@@ -289,11 +289,11 @@ export const Editable = (props: EditableProps) => {
 
         const anchorNodeSelectable =
           hasEditableTarget(editor, anchorNode) ||
-          isTargetInsideVoid(editor, anchorNode)
+          isTargetInsideNonReadonlyVoid(editor, anchorNode)
```
heyitsaamir, line 292: @ahoisl 👋🏽 I ran into this issue today. I'm wondering if the `||` needs to be `&&` instead.

ahoisl: Are you sure it's the same issue? "Cannot resolve from DOM point" is a very common slate error.

I'm pretty sure the boolean operation (`||`) is correct in this case. "void" nodes can usually be selected, just not when they are readonly and that's the only thing this fix is really checking for.

But there might be other issues or use cases that probably need to be considered?

heyitsaamir: I actually think the issue is with the implementation of `isTargetInsideNonReadonlyVoid`. 

https://github.com/ahoisl/slate/blob/43f84e489f2a5012acb3a7c3e296ad7d4a645ce8/packages/slate-react/src/components/editable.tsx#L1414..1422. 

It's actually implemented as "!isReadOnly || isVoid" instead of "!isReadOnly && isVoid". The double negative is confusing. I think that's where the confusion comes from. I think [this](https://github.com/ianstormtaylor/slate/pull/4887) fixes it

ahoisl: The fix in this PR #4727 was really just intended to avoid the "Cannot resolve from DOM point" error in readonly void nodes. I tried to change code as minimal as possible. The `onDOMSelectionChange` already had the `|| isTargetInsideVoid` in there, I just made sure it's not doing anything on readonly void nodes.

Your PR #4887 now changes the behavior to not step into the if-clause and call `toSlateRange` for non-readonly void nodes. I'm not sure if this is correct/intended or not.

PR #4731, opened 2021-12-13
Docs: Clarifies when to use Editor.insertNode vs. Transforms.insertNodes

DougReeder:
**Description**
Clarifies when to use Editor.insertNode vs. Transforms.insertNodes

### DIFF
--- a/docs/api/nodes/editor.md
+++ b/docs/api/nodes/editor.md
@@ -268,19 +268,19 @@ Insert a block break at the current selection.
 
 #### `Editor.insertFragment(editor: Editor, fragment: Node[]) => void`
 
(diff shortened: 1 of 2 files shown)


VictorBaron (approved):

DougReeder:
If someone can clarify when to use `Transforms.insertFragment` and when to use `Transforms.insertNodes` that would be helpful.

dylans:
> If someone can clarify when to use `Transforms.insertFragment` and when to use `Transforms.insertNodes` that would be helpful.

@DougReeder I'm going to land the PR as-is, and this can be addressed with another update.

DougReeder:
That's cool. Thanks!

Issue #4320, opened 2021-06-05
Blinking spelling highlight


fredure:
**Description**
When you enter text with spelling errors, the highlighting of words blinks

**Recording**
https://user-images.githubusercontent.com/8024592/120894087-58bd0300-c61f-11eb-9c5d-f7f0229aa10e.mp4

**Sandbox**
https://www.slatejs.org/examples/richtext

**Steps**
To reproduce the behavior:
1.  Type misspelled word
2. See error

**Expectation**
Spelling highlighting is expected not to blink. For example, TinyMCE and DraftJS no blinking.

**Environment**
Slate Version: 0.62.0.
Operating System: all.
Browser: Google Chrome 91.0.4472.77, (x86_64).

jaked:
this should be fixed by #3888 / #4605, give it a try with the just-released version 0.67.0

juliankrispel:
This has regressed by the looks of it, therefore reopening it.
Here's a loom reproducing the issue https://www.loom.com/share/879b476cc884477597801d7a58819483

Trying to debug currently why `TextString` is re-rendering even though the updated content is no different from what's in the dom. If anyone has any suggestions please let me know.

fredure:
With Cyrillic letters the issue is repeated.

PR #4733, opened 2021-12-15
Optimize TextString rendering to support browser/OS text features, eg fix native spellcheck

Schipy:
**Description**
This fix is the last or at least the next piece on the journey to [Use native character insertion to fix browser/OS text features](https://github.com/ianstormtaylor/slate/pull/3888/files) With all the good progress so far, native spellcheck as you type is still broken on practically all browsers (no spellcheck underlining on Safari & FF and blinking underlines on Chrome).

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4320
Related: https://github.com/ianstormtaylor/slate/pull/3888, https://github.com/ianstormtaylor/slate/issues/2051

**Example**
Safari, before (spellcheck prevented)
video of Safari, before (spellcheck prevented)

Chrome, before (spellcheck blinks)
Chrome-with-react-text

**Context**
The reason for this is that the continuously changing text will rerender the virtual DOM on each keystroke, which in turn updates the DOM TextNode, which will opt out browsers from checking spelling in that - now programatically updated, not user typed - text node.

Essentially now that data flow is two-way (simple characters are inserted natively, and propagated to slate, while complex characters and events are caught and happen in slate and propagated to native DOM), we need to tweak the leaf text node rendering to reflect this and instead of using the one way dataflow of virtual DOM, implement the leaf TextNode rendering as an Effect that does proper diffing between the actual DOM and the virtual DOM, and not based on diffing between older and newer version of the virtual DOM. We would like to replace the text content in a slate span exactly if and only if its current content does not match our current virtual DOM (and for natively inserted characters, it will already match and need no interference from code)

Safari, after
video of Safari, after

Chrome, after
video of Chrome, after

While this change may seem like quite fundamental and risky, I think it's actually quite locally confined: the `TextString` component behaves almost identical as before.

- it is rendered by React exactly on the same occasions when the previous version rendered (its props did not change)
- the result of the render will exactly be the same for the new and old version (both the traditional update via the virtual DOM, and the effect will achieve the same thing, namely that the span will contain the textContent)

A note on `null` handling: it may be unnecessary, but I added an extra measure to handle the extreme case if the incoming text is null/undefined, and also thereby preserving the previous render behavior in the `TextString` component (which also did not fail anddid not output any text node children for the span, should this happen).

Note 2: even with this fix spellcheck will still opt-out (or blink on Chrome) for all characters and events that are not natively handled (eg punctuation, backspace, shift+characters etc). Will have to see, this may be acceptable tradeoff, or if not, likely further steps need to be taken to handle as much natively as possible.

**Checks**
- [ ] functionality did not change, likely needs a `patch` changeset in `slate-react` package?

### DIFF
--- a/packages/slate-react/src/components/string.tsx
+++ b/packages/slate-react/src/components/string.tsx
@@ -71,7 +71,7 @@ const TextString = (props: { text: string; isTrailing?: boolean }) => {
   // useLayoutEffect: updating our span before browser paint
   useLayoutEffect(() => {
     // null coalescing text to make sure we're not outputing "null" as a string in the extreme case it is nullish at runtime
-    const textWithTrailing = (text ?? '') + (isTrailing ? '\n' : '')
+    const textWithTrailing = `${text ?? ''}${isTrailing ? '\n' : ''}`
 
     if (ref.current && ref.current.textContent !== textWithTrailing) {
       ref.current.textContent = textWithTrailing
--- a/packages/slate-react/src/components/string.tsx
+++ b/packages/slate-react/src/components/string.tsx
@@ -70,12 +70,8 @@ const TextString = (props: { text: string; isTrailing?: boolean }) => {
 
   // useLayoutEffect: updating our span before browser paint
   useLayoutEffect(() => {
-    let textWithTrailing: string | null = text + (isTrailing ? '\n' : '')
-
-    // making sure we're not outputing "null" in the extreme case the text is nullish at runtime
-    if (text == null) {
-      textWithTrailing = isTrailing ? '\n' : null
-    }
+    // null coalescing text to make sure we're not outputing "null" as a string in the extreme case it is nullish at runtime
+    const textWithTrailing = (text ?? '') + (isTrailing ? '\n' : '')
 
     if (ref.current && ref.current.textContent !== textWithTrailing) {
       ref.current.textContent = textWithTrailing
--- a/packages/slate-react/src/components/string.tsx
+++ b/packages/slate-react/src/components/string.tsx
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react'
+import React, { useRef, useLayoutEffect } from 'react'
 import { Editor, Text, Path, Element, Node } from 'slate'
 
 import { ReactEditor, useSlateStatic } from '..'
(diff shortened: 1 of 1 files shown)


dylans (commented):
Thanks @Schipy, looks promising. I need a day or two to think through this one and to give others a change to give feedback, before we would land it.

packages/slate-react/src/components/string.tsx
```diff
@@ -59,21 +59,34 @@ const TextString = (props: { text: string; isTrailing?: boolean }) => {
...
-    </span>
-  )
+  // useLayoutEffect: updating our span before browser paint
+  useLayoutEffect(() => {
+    let textWithTrailing: string | null = text + (isTrailing ? '\n' : '')
+
+    // making sure we're not outputing "null" in the extreme case the text is nullish at runtime
+    if (text == null) {
+      textWithTrailing = isTrailing ? '\n' : null
+    }
```
nemanja-tosic, line 78: The comment states we want to make sure we're not outputting null, but we do have a null. Are the comment and code out of sync or am i reading it wrong?

Schipy: Sorry, I meant in the comment to avoid outputting "null" as a string, eg "null\n" as a result of `null + '\n'`.
So the intent is to set actual null in case `text == null && !isTrailing`.
Actually, we could also set empty string in this case instead of null, as it seems to two is equivalent (both will actually result in no childNodes under the span and `textContent ===  ''`).
I will clarify the comment, and simplify this also in the code.

nemanja-tosic: You kind of addressed both things i wanted to ask there, looks good!

packages/slate-react/src/components/string.tsx
```diff
@@ -59,21 +59,30 @@ const TextString = (props: { text: string; isTrailing?: boolean }) => {
...
-  return (
-    <span data-slate-string ref={ref} key={forceUpdateCount.current}>
-      {text}
-      {isTrailing ? '\n' : null}
-    </span>
-  )
+  // useLayoutEffect: updating our span before browser paint
+  useLayoutEffect(() => {
+    // null coalescing text to make sure we're not outputing "null" as a string in the extreme case it is nullish at runtime
+    const textWithTrailing = (text ?? '') + (isTrailing ? '\n' : '')
```
nemanja-tosic, line 74: For your consideration, maybe we do interpolation here instead?

```suggestion
    const textWithTrailing = `${text ?? ''}${isTrailing ? '\n' : ''}`
```

I don't think it's better or worse, just more ES6-ish. Not a factor for merging the PR though.

nemanja-tosic (approved):
Love the fix, the spellchecker "lag" was just annoying...

hueyhe:
This fix breaks server-side rendering. All text won't get rendered on ssr, because the `useLayoutEffect` hook won't fire until the JavaScript is downloaded. Should use `useIsomorphicLayoutEffect` instead.

Issue #4719, opened 2021-12-07
Can't select images by clicking on them on Android


AkromDev:
**Description**
It is not possible to select images by clicking on them. We have to move cursor to images manually to select images. I believe this issue is general to other void types as well but I tested on images only.

**Recording**

https://user-images.githubusercontent.com/41626115/144971348-854d10a2-e753-48a0-8f52-832711542d59.mp4

**Sandbox**
https://www.slatejs.org/examples/images

**Steps to reproduce the behavior**
Go to examples. Click on image on android, it does not get selected.

**Expectation**
Clicking on images should select it on android as well just like IOS and web version.

**Environment**
Whatever is on example website

Issue #4715, opened 2021-12-05
Quick typing on android


YuriyMorkovnik:
**Description**
Quick typing causes missing some symbols

**Recording**

https://user-images.githubusercontent.com/33699475/144739634-77392f65-d46a-4168-9dcb-3e7270f79b92.mp4

**Sandbox**
There is just a copy of an official demo in sandbox
https://codesandbox.io/s/agitated-http-goi7b?file=/src/App.js

**Steps**
To reproduce the behavior:
1. Type words quickly
2. See error

**Expectation**
See correct pressed symbols when you're typing quickly

**Environment**

- Slate Version: 0.71
- Operating System: Android 10
- Browser: Chrome

garymillergit:
I am also experiencing this issue in my react project, i.e. not just a sandbox. I'm hopeful for a solution.

skogsmaskin:
Seems to be an issue with Swiftkey keyboard. I can't reproduce it on Gboard.

YuriyMorkovnik:
Updated slate to v0.80. Seems this issue is fixed. Thanks all!

Issue #4714, opened 2021-12-05
Long press backspace


YuriyMorkovnik:
**Description**
Long press a backspace button causes a concatenation a part of deleting word. Also you can't hold a backspace to delete a part of a word because it stops you after the first symbol.

**Recording**

https://user-images.githubusercontent.com/33699475/144739309-ba79b1da-db2c-413c-9c03-94edbf658886.mp4

**Sandbox**
There is just a copy of a official demo in sandbox
https://codesandbox.io/s/agitated-http-goi7b?file=/src/App.js

**Steps**
To reproduce the behavior:
1. Write a text
2. Hold a backspace button
3. See error

**Expectation**
Smooth deleting of symbols while you’re holding a backspace button

**Environment**
- Slate Version: 0.71
- Operating System: Android 10
- Browser: Chrome

AkromDev:
I had the same issue on galaxy s8 but did not face the issue on s21 and galaxy a21. Maybe It seems that editor has some performance issues on slower/older phones and require more CPU/RAM power. I am not getting this issue anymore on latest examples. The backward delete seems to have been improved as well. My assumption is that https://github.com/ianstormtaylor/slate/pull/4706 has improved editor performance for android and some issues resulting from poor performance are now fixed.

AkromDev:
I noticed it is happening near images while testing but it improved compared to slate@0.71.0

TrySpace:
Seems related to this: https://github.com/ianstormtaylor/slate/issues/4715

YuriyMorkovnik:
Updated slate to v0.80. Seems this issue is fixed. Thanks all!

Issue #4705, opened 2021-12-02
Android device issues include IME problems and inability to focus on void element.


YasinChan:
**Description**
I found here '[export const Editable = IS_ANDROID ? AndroidEditable : DefaultEditable](https://github.com/ianstormtaylor/slate/blob/fece5c157099b5ddef9f6dc8d070f2fdb3724b90/packages/slate-react/src/index.ts#L7)'. What is the reason for splitting `AndroidEditable` and `DefaultEditable`? Some of the issues I found on Android device included that i can't focus on Void Element and had some problem with IME.

**Recording**

https://user-images.githubusercontent.com/22654945/144382771-27b54d50-5cd8-4220-965e-3f6e2afe6a03.mp4

**Steps**
1. go to https://www.slatejs.org/examples/images
2. use Chrome DevTools.
3. select one Android device.

**Expectation**
Perhaps can provide an initialization options to control whether or not use `AndroidEditable` in Android device.

**Environment**
- Slate Version: slate@0.71.0 slate-react@0.71.0
- Operating System: Android

**Context**
Similar to #4400

YasinChan:
In addition, I tried to force the use of `DefaultEditable` on Android device, and it worked fine.

YasinChan:
> Perhaps can provide an initialization options to control whether or not use AndroidEditable in Android device.

My fault, i can use `DefaultEditable` instead of `Editable` by `import { DefaultEditable } from 'slate-react';`

see [here](https://github.com/ianstormtaylor/slate/blob/fece5c157099b5ddef9f6dc8d070f2fdb3724b90/packages/slate-react/src/index.ts#L9)

PR #4734, opened 2021-12-15
fix: [AndroidEditor] Solve input association problems and add click events.

YasinChan:
**Description**
I asked some questions about Android devices earlier in #4705 . I tried to switch to `DefaultEditable` but that didn't resolve my questions. After i read source code about `AndroidEditable`.  I may sum up a few questions and try to resolve them.

Q:

1. As we can see in these issues #4715 #4714 or the first video, using the typing with association function will trigger an unexpected insertText operation .
2. We can't click block element on Android which also mentioned in #4719 .

A:

1. `MutationObserver` is used in `AndroidEditable` to monitor DOM changes. Whenever we input a character, `MutationObserver` will be triggered and `insertText` into the editor. However, in the keyboard association state, performing a delete operation or inputting characters may trigger character association, and then unexpected characters are added between triggering the onDOMBeforeInput event and triggering the MutationObserver event. I haven't figured out the reason yet, but I try to solve it in another way. The association function will trigger the `composition` event, so I attempt to record the characters that originally wanted to insertText in the MutationObserver event in the `isComposing` , and then `insertText` in the `onCompositionend` event.
2. Synchronize the click event from `Editable`. By the way, i also synchronize some other code from `Editable` like #4669 .

**Example**
before

https://user-images.githubusercontent.com/22654945/146237301-d3277a50-cf8d-4e5c-adb2-71efa254b8cc.mp4

after

https://user-images.githubusercontent.com/22654945/146237493-d6e66b7a-cede-4e65-af7a-ba9ee4b8e423.mp4

Image can be clicked.

https://user-images.githubusercontent.com/22654945/146238103-a3670a5d-840f-41ad-8538-50bada9696b4.mp4

**Context**
There are still some minor problems. When the above method is triggered, the mark setting cannot take effect immediately, but it takes effect after `onCompositionend`.

https://user-images.githubusercontent.com/22654945/146238981-d4f00790-d02a-45d5-8524-ef4bd72ed908.mp4

**Checks**

### DIFF
--- a/.changeset/quick-feet-melt.md
+++ b/.changeset/quick-feet-melt.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+[AndroidEditor] Solve input association problems and add click events.
--- a/packages/slate-react/src/components/android/android-editable.tsx
+++ b/packages/slate-react/src/components/android/android-editable.tsx
@@ -1,6 +1,7 @@
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
-import { Descendant, Editor, Element, Node, Range, Transforms } from 'slate'
+import { Editor, Element, Node, Range, Transforms, Path, Text } from 'slate'
 import throttle from 'lodash/throttle'
+import debounce from 'lodash/debounce'
 import scrollIntoView from 'scroll-into-view-if-needed'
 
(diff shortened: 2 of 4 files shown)


dylans (approved):
Amazing, thanks for tackling this @YasinChan  !

PR #4735, opened 2021-12-16
Avoid transforming dirty paths if the op won't touch them

steve-codaio:
**Description**
While large bulk edits are not a common scenario, we noticed that the dirty path tracking during batches has an O(n^2) behavior in relation to the number of ops. This is because as every op gets applied, we loop over all previous dirty paths and transform them. This implements an optimization where we can bypass the no-op transform entirely if we know the op won't affect paths at all.

In a simple test where I had a slate with 1000 texts and I applied an 'insert_text' op appending a character to each text, I saw about a 50% time savings (went from ~2s down to ~1s)

**Checks**

### DIFF
--- a/packages/slate/src/create-editor.ts
+++ b/packages/slate/src/create-editor.ts
@@ -13,7 +13,7 @@ import {
   Text,
   Transforms,
 } from './'
-import { DIRTY_PATHS, FLUSHING } from './utils/weak-maps'
+import { DIRTY_PATHS, DIRTY_PATH_KEYS, FLUSHING } from './utils/weak-maps'
 
 /**
  * Create a new Slate `Editor` object.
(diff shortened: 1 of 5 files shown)


packages/slate/src/create-editor.ts
```diff
@@ -57,13 +57,17 @@ export const createEditor = (): Editor => {
       }
 
       const oldDirtyPaths = DIRTY_PATHS.get(editor) || []
-      const newDirtyPaths = getDirtyPaths(op)
-
+      const canTransformPath = Path.operationCanTransformPath(op)
       for (const path of oldDirtyPaths) {
-        const newPath = Path.transform(path, op)
-        add(newPath)
+        if (canTransformPath) {
```
jason-codaio, line 62: I was hoping we could optimize this more but I realized we have to rebuild the set for de-duping each time. Maybe the key set should be persisted with the DIRTY_PATHS so this doesn't need to be reconstructed each time. This code path can be so hot on large edits.

steve-codaio: I can investigate storing the path keys alongside the dirty paths and seeing what additional savings we can get

steve-codaio: Updated to reuse the array/set entirely if the op won't touch them. Shaved an additional ~10% of time off the test mentioned in the summary

dylans (approved):
Thanks, this looks promising.

I'll plan to land it in a couple of days unless anyone objects.

PR #4737, opened 2021-12-17
Documents use of withoutNormalizing in conjunction with normalization

DougReeder:
**Description**
Cues programmers that sequences of Transforms may need to be wrapped in `Editor.withoutNormalizing` if the node tree should _not_ be normalized between Transforms.


dylans (approved):
Thanks @DougReeder!

PR #4749, opened 2021-12-27
bugfix: On Preact, sometimes DOM & Slate selection are mismatching

Jabher:
**Description**
Sadly, I cannot create stand-alone reproduction environment, only happens in real world for me.

When switching between multiple editors quickly, there is around 5% chance of following error (location differs):
```
Uncaught (in promise) Error: Cannot resolve a DOM point from Slate point: {"path":[1,0],"offset":6}
```

which is caused by this line:
https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/editable.tsx#L217

No idea why this only happens on real-world, but I was not able to reproduce this behavior on some demo stage.
However, the environment is the following:
- Slate 0.72.3, Slate-React 0.72.1
- Plate 7.x
- Preact 10.5.4
- 2 or more editors are present on a same page

Due to fact it is not happening every time, looks like this is race-condition problem and incompatibility between slate and (possibly) preact. However, this logic fully fixes everything.

**Checks**

### DIFF
--- a/.changeset/shiny-eagles-swim.md
+++ b/.changeset/shiny-eagles-swim.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fix "cannot resolve DOM point" error when switching between multiple errors
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -214,7 +214,10 @@ export const Editable = (props: EditableProps) => {
     // Otherwise the DOM selection is out of sync, so update it.
     state.isUpdatingSelection = true
 
-    const newDomRange = selection && ReactEditor.toDOMRange(editor, selection)
+    const newDomRange =
+      selection &&
+      hasDomSelectionInEditor &&
+      ReactEditor.toDOMRange(editor, selection)
 
     if (newDomRange) {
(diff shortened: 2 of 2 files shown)


dylans (commented):
Seems like a reasonable check that won't slow things down or cause harm. Let me know when it's ready for a real review and remember to add a changeset when you're ready.

Jabher:
Sure! just wanted to check every thing possible

Jabher:
@dylans Hi! Everything is ready, tests are fine... sadly I was not able to figure out how to make stable test that reproduces this behavior. However, all the tests are running fine

dylans (approved):
Thanks @Jabher . I'm thinking this may also help fix #4789 and other similar issues.

Jabher:
was happy to help!

BitPhinix:
Sadly this change causes issues for us, checking if the selection is set and in the editor at this stage results in slate-react not updating the selection if it isn't already inside the editor. So focusing and selecting something without having a dom selection (or selection inside the editor) doesn't update/set the dom selection.

BitPhinix:
@dylans could you revert it? There has to be another way to solve this issue, but this pr breaks basic slate functionality.

BitPhinix:
Awesome, thanks so much 👍

Issue #4709, opened 2021-12-03
editor.insertText never gets called inside plugins on android


fridaystreet:
**Description**
if you create a withMyPlugin and have it override the editor.insertText function, the function is never called when inserting text on an android device

**Steps**
create a plugin and override the editor.insertText function, console.log the output, it never gets called on android

**Expectation**
the editor.insertText function is called like other functions when user enters text on andorid

**Environment**
- Slate Version: latest
- Operating System: android
- Browser: Chrome

fridaystreet:
sorry not sure why the cross platform label was added, I did choose bug for specific platform

alessiogaldy:
I took a look at this issue, and if I change this line https://github.com/ianstormtaylor/slate/blob/735d2d047aa5a7ded7d1180fcbda6008f58943a8/packages/slate-react/src/components/android/android-editable.tsx#L563 to call `Editor.insertText`, the plugins work again.
I am unsure if this is the correct fix because I don't know when `Editor.insertText` should be used instead of `Transforms.insertText`. The regular `Editable` component uses both.
I hope this helps narrow down the issue 🤞

dylans:
> sorry not sure why the cross platform label was added, I did choose bug for specific platform

fwiw, I think the label gets applied when something breaks cross-platform compatibility.

dylans:
> I took a look at this issue, and if I change this line
> 
> https://github.com/ianstormtaylor/slate/blob/735d2d047aa5a7ded7d1180fcbda6008f58943a8/packages/slate-react/src/components/android/android-editable.tsx#L563
> to call `Editor.insertText`, the plugins work again.
> I am unsure if this is the correct fix because I don't know when `Editor.insertText` should be used instead of `Transforms.insertText`. The regular `Editable` component uses both.
> I hope this helps narrow down the issue 🤞

`Editor.insertText` simply calls this: https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/create-editor.ts#L169-L182 which checks for a selection and if the selection contains marks. If no selection it does nothing, if selection contains marks, it calls `Transforms.insertNodes`, or if no marks, it calls `Transforms.insertText`.

So yes, overriding `Editor.insertText` will get ignored since the underlying code calls `Transforms.insertText` in both the Android and non-Android Editable code.

[...]

PR #4753, opened 2021-12-30
Fix editor.insertText never gets called inside plugins on Android

alessiogaldy:
**Description**
Call `editor.insertText` instead of `Transforms.insertText` to allow overriding behavior in plugins.

**Issue**
Fixes: #4709
I was able to verify that this now works correctly on my Android device, but I am not sure how to add a test for this. 

**Checks**

### DIFF
--- a/.changeset/breezy-ears-happen.md
+++ b/.changeset/breezy-ears-happen.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fix "editor.insertText never gets called inside plugins on android"
--- a/packages/slate-react/src/components/android/android-editable.tsx
+++ b/packages/slate-react/src/components/android/android-editable.tsx
@@ -560,9 +560,7 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
                       })
                       editor.marks = null
                     } else {
-                      Transforms.insertText(editor, text, {
-                        at,
-                      })
+                      Editor.insertText(editor, text)
                     }
                   })
                 }, RESOLVE_DELAY)
--- a/packages/slate-react/src/components/android/android-input-manager.ts
+++ b/packages/slate-react/src/components/android/android-input-manager.ts
@@ -134,9 +134,7 @@ export class AndroidInputManager {
         })
         this.editor.marks = null
       } else {
-        Transforms.insertText(this.editor, text, {
-          at,
-        })
+        Editor.insertText(this.editor, text)
       }
     })
   }
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -793,7 +793,7 @@ export const Editable = (props: EditableProps) => {
                         distance: currentTextNode.text.length,
                         reverse: true,
                       })
-                      Transforms.insertText(editor, text)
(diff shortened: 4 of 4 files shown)


dylans (approved):
Thanks @alessiogaldy , I will land this sometime in the next few days unless there are objections.

TheSpyder:
This code was added in #4734 and appears to very specifically _not_ call `Editor.insertText` in order to adjust the `at` position. I don't have an Android device so I don't really understand the scenario where this might happen. We probably need @YasinChan's input here.

YasinChan:
Hi @TheSpyder~ 
[Code in android-editable.tsx](https://github.com/ianstormtaylor/slate/blob/cc9cba017a3abc998de06cc0a3d15a6a0bd47100/packages/slate-react/src/components/android/android-editable.tsx#L547) is mine, but it came from [Code in android-input-manager.ts](https://github.com/ianstormtaylor/slate/blob/cc9cba017a3abc998de06cc0a3d15a6a0bd47100/packages/slate-react/src/components/android/android-input-manager.ts#L125). So I didn't pay much attention to the difference between `Editor.insertText` and `Transforms.insertText` at that time.

TheSpyder:
Ah, my apologies.

It looks like _that_ code originally came from #4257:
https://github.com/ianstormtaylor/slate/pull/4257/commits/a93c555ca84ef0dd3eafc8c95603640eaf58a867

@clauderic, are you able to provide any insights here?

alessiogaldy:
I’ll test this again later today but I suspect that this caused a regression and now the text is inserted at the wrong position

alessiogaldy:
I checked, and it looks like setting `at` is needed because this function adjusts the location before inserting the text:
https://github.com/ianstormtaylor/slate/blob/1b0e7c6b928865cb4fd656b6f922e30fbe72d77a/packages/slate-react/src/components/android/diff-text.ts#L161
I was able to obtain the same behavior by replacing the `Transforms.insertText` call with:
```Typescript
Transforms.setSelection(editor, at)
Editor.insertText(editor, text)
```
I am not sure if this is a good idea, but it seems to work.
I think a cleaner solution would be to allow for an optional `to` parameter to be passed to `Editor.insertText` so that it can be passed down to `Transforms.insertText`. Unfortunately that would require all existing plugins to pass the new argument down the chain.
It might make sense to change every function of the `Editor` interface to accept two arguments, the editor and an object with all other parameters. This would make adding arguments a non-breaking change in most cases because each plugin will either override the behavior or pass the object down the chain.

When testing on my phone I also noticed that when keyboard inputs are delayed, which seems to be needed to deal with edge cases on Android, the cursor sometimes jumps back and forth. 
This is easier to reproduce by typing special chars like `@` `$`.
I can reproduce this issue on the examples page so it's not a regression caused by this change (AFAIK it's not live yet).

I am happy to open a PR to add the `Transforms.setSelection` call and to disable the input delayed logic.

alessiogaldy:
Reading the [ProseMirror source code](https://github.com/ProseMirror/prosemirror-view/blob/master/src/domobserver.js) that is referenced in `android-input-manager.ts` it looks like the delay is only used for Internet Explorer <= 11  so it should be safe to remove it.

Issue #4754, opened 2021-12-30
useFocused hook displays old value


jhurwitz:
**Description**
The results of the `useFocused` hook are unexpected. I dug into the root cause, and I believe it's from #3988. I have a fix forthcoming, so I'll explain more in my PR and will let the reproduction steps below serve as documentation of the issue.

**Recording**
* If you don't type any text in the editor, and just click in and out of it repeatedly, after the first click, `useFocused` always indicates the opposite of whether the editor is focused.
![GIF 12-30-21 at 3 21 23 PM](https://user-images.githubusercontent.com/591990/147794082-7ba00b02-a960-48e2-aec7-778081d363de.gif)
* If you type text into the editor, and then click in and out of it repeatedly, `useFocused` almost always indicates that the editor is focused even when it's not (except for brief flashes of not focused when you first focus).
![GIF 12-30-21 at 3 22 04 PM](https://user-images.githubusercontent.com/591990/147794083-194ea265-d86e-4408-a9e7-459275d67da8.gif)

**Sandbox**
The two GIFs above are from [this code sandbox](https://codesandbox.io/s/stupefied-moore-qsmjh?file=/src/Editor.js).

**Steps**
To reproduce the behavior:
1. Go to the code sandbox
2. Click into the editor
3. Click out of the editor
4. Click back into the editor

**Expectation**
Expected: you should see "not focused" after step 3, and "focused" after step 4. The actual behavior is flipped.

**Environment**
- Slate Version: 0.72.3
- Operating System: macOS 12
- Browser: I have reproed in both Chrome 96 and Safari 15.1

**Context**
I'll create a PR soon.

adri1wald:
I've done a bit of experimentation as I've run into the same issue with my editor, and noticed that this error only happens on react version 17. If you use react and react-dom 16 in [this code sandbox](https://codesandbox.io/s/stupefied-moore-qsmjh?file=/src/Editor.js) you'll see the inverted behaviour goes away

PR #4755, opened 2021-12-30
run focus event listener after existing onFocus handlers

jhurwitz:
**Description**
As described in the issue, the `useFocused` hook was returning incorrect/unexpected values. I debugged and realized it was displaying old values because of the order in which event listeners were running.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4754

**Example**
Here I repro the exact two behaviors I described in the linked issue, but with this fix in place. (View the GIFs in the issue to see what they looked like before this fix.)
GIF 12-30-21 at 3 49 48 PM
GIF 12-30-21 at 3 50 21 PM

**Context**
Based on my testing, I think there was a race condition between the focus/blur event listeners in `components/slate.tsx` (added in #3987) and the onFocus/onBlur handlers in `components/editable.tsx`. The `IS_FOCUSED` map was eventually consistent/had the correct values after all handlers ran, but the event listeners in `slate.tsx` were running before the handlers in `editable.tsx` and therefore the `isFocused` hook was displaying the old values, from _before_ the focus/blur handlers ran.

Adding a `setTimeout` with a delay of 0 cause the event listeners in `slate.tsx` to run after the handlers in `editable.tsx`, thereby fixing the issue.

**Checks**

### DIFF
--- a/.changeset/stale-poems-itch.md
+++ b/.changeset/stale-poems-itch.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+fix useFocused hook
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -57,13 +57,21 @@ export const Slate = (props: {
   })
 
   useIsomorphicLayoutEffect(() => {
-    const fn = () => setIsFocused(ReactEditor.isFocused(editor))
+    const fn = () => {
+      window.setTimeout(() => {
+        setIsFocused(ReactEditor.isFocused(editor))
+      }, 0)
+    }
     document.addEventListener('focus', fn, true)
     return () => document.removeEventListener('focus', fn, true)
   }, [])
 
   useIsomorphicLayoutEffect(() => {
-    const fn = () => setIsFocused(ReactEditor.isFocused(editor))
+    const fn = () => {
+      window.setTimeout(() => {
+        setIsFocused(ReactEditor.isFocused(editor))
+      }, 0)
+    }
     document.addEventListener('blur', fn, true)
     return () => document.removeEventListener('blur', fn, true)
   }, [])
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -58,7 +58,7 @@ export const Slate = (props: {
 
   useIsomorphicLayoutEffect(() => {
     const fn = () => {
-      window.setTimeout(() => {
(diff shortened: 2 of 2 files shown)


jhurwitz:
I believe the integration test failure is a flaky test. That test passes locally, and it's failing on this line which suggests to me that maybe the wait just isn't long enough?
```
    cy.findByRole('textbox')
      // need wait() here otherwise the slate component is not fully mounted yet sometimes
      .wait(1000)
      .type(...
```

dylans (approved):

dylans (commented):
Actually @jhurwitz , I wonder if we really want to use `window.setTimeout` inside `useIsomorphicLayoutEffect` instead of some other timeout mechanism that doesn't rely on the presence of the `window` object?

jhurwitz:
> Actually @jhurwitz , I wonder if we really want to use `window.setTimeout` inside `useIsomorphicLayoutEffect` instead of some other timeout mechanism that doesn't rely on the presence of the `window` object?

Ah, good call. Would changing to just `setTimeout` work? That should call the same method in the browser, and seems to also work in Node/server-side environments where there is no `window` object.

363676727:
I use the newest recently appeared.

It seems that the react on your demo is not the same.

Kapture 2022-01-06 at 12 14 01

jhurwitz:
> I use the newest recently appeared.
> It seems that the react on your demo is not the same.
> Or is it the browser difference?
> Kapture 2022-01-06 at 12 14 01

Thanks for the comment. Can you please give more details?

Are you running this out of the [Codesandbox](https://codesandbox.io/s/stupefied-moore-qsmjh?file=/src/Editor.js) or locally? I suspect you might be running it locally, because the gray background in your screen capture does not appear in the Codesandbox.

And just to confirm -- are you testing on the `main` branch of `slate-react`, or on the branch containing this PR?

> I use the newest recently appeared.

The newest what, exactly?

> It seems that the react on your demo is not the same.

Are you saying you're using a different version of React.js (if so, which version? the codesandbox uses 17.0.2, the latest version), or are you saying that the reaction of the useFocused hook is different?

> Or is it the browser difference?

What browser and OS are you using?

bryanph:
This change caused a regression for me when changing selection state between two editors, it resets the selection to the last selection position instead of the new one.

bryanph:
This causes a regression for me somehow when selecting between two different editors. It moves the cursor back to the previous selection instead. Trying to investigate why exactly but the timeout here seems to have something to do with it.

Issue #4285, opened 2021-05-24
Slate does not work in Firefox Iframes


maclockard:
**Description**

When rendering a slate editor inside of an Iframe using firefox, the editor does not render the initial content and is not intractable. There are no errors in the console AFAIK.

Possibly related to https://github.com/ianstormtaylor/slate/issues/4170?

**Recording**

Firefox is on the left, chrome is on the right:

https://user-images.githubusercontent.com/7308388/119389084-fdcef780-bc7f-11eb-9661-1c4452934a33.mov

**Sandbox**

This repros in the official slate examples: https://www.slatejs.org/examples/iframe

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/iframe in Firefox
2. Try to use the editor

You will see that the editor does not render the initial text and that you can't interact with it. 

**Expectation**

The editor renders the initial content and is intractable in Firefox. This currently works in Chrome.

**Environment**
- Slate Version: 0.63.0
- Operating System: macOS
- Browser: Firefox 88.0.1

ottworks:
You can get the editor to work by opening the side menu and closing it again.

PR #4758, opened 2022-01-02
iframe example - use an onLoad callback instead of the ref

adrid:
**Description**
Makes the iframe example work on the Firefox and keeps it working on other browsers. It changes using the ref on the iframe to an `onLoad` event solution

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4285

**Example**
https://www.slatejs.org/examples/iframe on the Firefox

**Context**
Firefox resets the iframe content on the load event. I've tested on some older versions of the browser and it keeps happening too. Here is an issue on the react repo https://github.com/facebook/react/issues/22847

**Checks**

### DIFF
--- a/site/examples/iframe.tsx
+++ b/site/examples/iframe.tsx
@@ -104,15 +104,17 @@ const MarkButton = ({ format, icon }) => {
   )
 }
 
-const IFrame = ({ children, ...props }) => {
-  const [contentRef, setContentRef] = useState(null)
-  const mountNode =
-    contentRef &&
-    contentRef.contentWindow &&
-    contentRef.contentWindow.document.body
+const IFrame = ({ children, onLoad = null, ...props }) => {
+  const [iframeBody, setIframeBody] = useState(null)
+
+  const handleLoad = e => {
+    onLoad && onLoad(e)
+    !e.defaultPrevented && setIframeBody(e.target.contentDocument.body)
+  }
+
   return (
-    <iframe {...props} ref={setContentRef}>
-      {mountNode && createPortal(React.Children.only(children), mountNode)}
+    <iframe srcDoc={`<!DOCTYPE html>`} {...props} onLoad={handleLoad}>
+      {iframeBody && createPortal(children, iframeBody)}
     </iframe>
   )
 }
--- a/site/examples/iframe.tsx
+++ b/site/examples/iframe.tsx
@@ -104,14 +104,11 @@ const MarkButton = ({ format, icon }) => {
   )
 }
 
-const IFrame = ({ children, onLoad = null, ...props }) => {
+const IFrame = ({ children, ...props }) => {
   const [iframeBody, setIframeBody] = useState(null)
-
   const handleLoad = e => {
-    onLoad && onLoad(e)
-    !e.defaultPrevented && setIframeBody(e.target.contentDocument.body)
+    setIframeBody(e.target.contentDocument.body)
   }
-
   return (
     <iframe srcDoc={`<!DOCTYPE html>`} {...props} onLoad={handleLoad}>
       {iframeBody && createPortal(children, iframeBody)}
--- a/cypress/integration/examples/iframe.ts
+++ b/cypress/integration/examples/iframe.ts
@@ -12,6 +12,7 @@ const getIframeBody = () => {
       .its('body')
       // automatically retries until body is loaded
       .should('not.be.undefined')
+      .should('not.be.null')
       .then(cy.wrap)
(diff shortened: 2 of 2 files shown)


site/examples/iframe.tsx
```diff
@@ -104,15 +104,17 @@ const MarkButton = ({ format, icon }) => {
...
-  const [contentRef, setContentRef] = useState(null)
-  const mountNode =
-    contentRef &&
-    contentRef.contentWindow &&
-    contentRef.contentWindow.document.body
+const IFrame = ({ children, onLoad = null, ...props }) => {
+  const [iframeBody, setIframeBody] = useState(null)
+
+  const handleLoad = e => {
+    onLoad && onLoad(e)
```
nemanja-tosic, line 111:
```suggestion
    onLoad?.(e)
```

site/examples/iframe.tsx
```diff
@@ -104,15 +104,17 @@ const MarkButton = ({ format, icon }) => {
   )
 }
 
-const IFrame = ({ children, ...props }) => {
-  const [contentRef, setContentRef] = useState(null)
-  const mountNode =
-    contentRef &&
-    contentRef.contentWindow &&
-    contentRef.contentWindow.document.body
+const IFrame = ({ children, onLoad = null, ...props }) => {
```
nemanja-tosic, line 107: Any specific reason to assign null to `onLoad`, instead of leaving it undefined?

```suggestion
const IFrame = ({ children, onLoad, ...props }) => {
```

adrid: Thanks for pointing that out. I was experimenting with solutions and I had to left out it there. And actually that whole `onLoad` is not needed there... I will remove it

site/examples/iframe.tsx
adrid: I've removed that part completely as not needed

nemanja-tosic:
Just to make sure, it would not be possible to pass in `onLoad` load anymore. Is that intentional?

adrid:
> Just to make sure, it would not be possible to pass in `onLoad` load anymore. Is that intentional?

Yes. I don't think this is needed for this example.

dylans:
@adrid, this change has broken our integration test. Any chance you have time to update the test to reflect the change in the example?

```Running:  examples/iframe.ts                                                             (8 of 22)


  iframe editor
    1) should be editable


  0 passing (10s)
  1 failing

  1) iframe editor
       should be editable:
     TestingLibraryElementError: Timed out retrying after 4000ms: Unable to find an accessible element with the role "textbox"

There are no accessible roles. But there might be some inaccessible roles. If you wish to access them, then set the `hidden` option to `true`. Learn more about this here: https://testing-library.com/docs/dom-testing-library/api-queries#byrole
      at Object.getElementError (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts:1764:17)
      at eval (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts:2729:25)
      at eval (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts:2701:24)
      at eval (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts:2754:25)
      at baseCommandImpl (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts:477:16)
      at commandImpl (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts:480:40)
      at getValue (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts:504:23)
      at resolveValue (http://localhost:3000/__cypress/tests?p=cypress/support/index.ts:544:35)




  (Results)

  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Tests:        1                                                                                │
  │ Passing:      0                                                                                │
  │ Failing:      1                                                                                │
  │ Pending:      0                                                                                │
  │ Skipped:      0                                                                                │
  │ Screenshots:  1                                                                                │
  │ Video:        true                                                                             │
  │ Duration:     10 seconds                                                                       │
  │ Spec Ran:     examples/iframe.ts                                                               │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘

```

adrid:
@dylans when I created the pull request it was not passing this test and I noticed that later - the same thing what you pointing out. 

I already fixed it here in the next commit https://github.com/ianstormtaylor/slate/pull/4758/commits/87a860462051fbd45583e7404790727570b8ec29

I've tested on the main branch and seems to be okay so maybe you referring to some check before I fixed that?

Issue #4765, opened 2022-01-04
Slate-react is an uncontrolled component in v0.67+


antsago:
**Description**
If you initiate a value change from outside of slate by calling setValue (e.g. from within a setTimeout), the value does not change, making the slate component behave as if it was an uncontrolled component. I've reproed it with slate-react 0.67.0 and later, but found it working with slate-react 0.66.7 and prior. 

**Recording**

https://user-images.githubusercontent.com/3484531/148126679-2567696e-1f99-4117-9ee4-5df27e4cb28e.mov

**Sandbox**
https://codesandbox.io/s/slate-reproductions-forked-4vb05

**Steps**
To reproduce the behavior:
1. Refresh the page (from the sandbox)
2. Wait for two seconds
3. See the text fail to change but the corresponding message appear in the console

**Expectation**
The value the editor displays should change to match what the value passed to it.

**Environment**
- Slate Version: 0.67.0-0.72.3
- Operating System: macOS
- Browser: Chrome

**Context**
The idea is to change the editor text from external (non-slate) events; the use of setTimeout is just a simple way of imitating that, not the actual use-case. I've also managed to get around it by wrapping Slate and Editor (and the createEditor useMemo) within another component and forcing it to re-render by giving it a key and changing it every time.

dylans:
See the discussion in https://github.com/ianstormtaylor/slate/pull/4540 for why this was changed.

antsago:
@dylans thanks for pointing it out! I'll close this issue now and sorry I couldn't find the previous issues (of which there are quite a lot!). From what I read, rather than setValue, we should use something like: 
```javascript
  const externalSetValue = useCallback(
    (newValue) => {
      editor.children = newValue
      setValue(newValue)
    },
    [editor]
  )
```

Should I send a pr updating the documentation with that snippet?

dylans:
No worries. A PR updating the docs would be amazing, thanks!

PR #4768, opened 2022-01-05
Update docs to reflect Slate is an uncontrolled component

antsago:
**Description**
Updated docs to explain that Slate provider is now an uncontrolled component, to match the changes introduced in this pr: https://github.com/ianstormtaylor/slate/pull/4540. Thanks to @KarimNahas for his [resetNodes example](https://github.com/ianstormtaylor/slate/pull/4540#issuecomment-951903419)

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4765

**Example**
N/A

**Context**
N/A

**Checks**
- [x] This is a pure docs change

### DIFF
--- a/docs/walkthroughs/01-installing-slate.md
+++ b/docs/walkthroughs/01-installing-slate.md
@@ -110,7 +110,9 @@ const App = () => {
 }
 ```
 
-You can think of the `<Slate>` component as providing a "controlled" context to every component underneath it.
+You can think of the `<Slate>` component as providing a context to every component underneath it.
+
+> As of v0.67 the Slate Provider's "value" prop is now only used as initial state for editor.children. If your code relies on replacing editor.children you should do so by replacing it directly instead of relying on the "value" prop to do this for you. See [here](https://github.com/ianstormtaylor/slate/pull/4540) for a more in-depth discussion.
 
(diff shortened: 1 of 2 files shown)


docs/walkthroughs/01-installing-slate.md
```diff
@@ -110,7 +110,9 @@ const App = () => {
 }
 ```
 
-You can think of the `<Slate>` component as providing a "controlled" context to every component underneath it.
+You can think of the `<Slate>` component as providing a context to every component underneath it.
+
+> As of v0.67 the Slate Provider's "value" prop is now only used as initial state for editor.children. If your code relies on replacing editor.children you should do so by replacing it directly instead of relying on the "value" prop to do this for you. See [here](https://github.com/ianstormtaylor/slate/pull/4540) for a more in-depth discussion.
```
dylans, line 115:
```suggestion
> As of v0.67 the Slate Provider's "value" prop is now only used as initial state for editor.children. If your code relies on replacing editor.children you should do so by replacing it directly instead of relying on the "value" prop to do this for you. See [Slate PR 4540](https://github.com/ianstormtaylor/slate/pull/4540) for a more in-depth discussion.
```

dylans (changes requested):
One small suggestion as I loathe linking to the word here. Thanks @antsago for taking the time to contribute this update!

antsago:
Good suggestion! Thanks to yourself and @ianstormtaylor for sharing this library with everyone :D

dylans (approved):

PR #4769, opened 2022-01-05
Warn when normalization removes node

jameshfisher:
**Description**
Slate requires the invariant that children are all blocks or all inlines. It enforces this in default normalization by removing children. When such a node is removed, it's almost certainly due to a programming error: the developer needs to fix their application to ensure it maintains this invariant. But currently Slate does not tell the developer this.

I made such a programming error, and spent a long time debugging nodes that mysteriously went missing. I would have fixed it within 30 seconds if Slate had warned me when it detected this error.

**Context**
Note I have used `console.warn` despite the eslint rule. As far as I can see, Slate has no other facility for runtime warnings.

**Checks**

### DIFF
--- a/.changeset/wild-penguins-shout.md
+++ b/.changeset/wild-penguins-shout.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Warn when normalization removes node
--- a/packages/slate/src/create-editor.ts
+++ b/packages/slate/src/create-editor.ts
@@ -227,6 +227,14 @@ export const createEditor = (): Editor => {
         // other inline nodes, or parent blocks that only contain inlines and
         // text.
         if (isInlineOrText !== shouldHaveInlines) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            `Removing ${
+              isInlineOrText ? 'inline' : 'block'
+            } node at path ${path.concat(n)} because parent expects ${
+              shouldHaveInlines ? 'inline' : 'block'
+            } children`
+          )
           Transforms.removeNodes(editor, { at: path.concat(n), voids: true })
           n--
         } else if (Element.isElement(child)) {


dylans (approved):
Excellent suggestion, I've been a victim of this as well.

TheSpyder:
I disagree with how this is implemented. Strongly. Slate core does not emit warnings; that lint rule is there for a reason.

The core is used by a variety of editors, not just `slate-react`, and even though this might be caused by programmer error and good to highlight we can't know that it is appropriate to log a console message.

In particular, collaborative editors can end up in all sorts of weird states that require normalization to fix. This is normal behaviour, I would like to see this reverted (or I can do it) and we look for another way to handle the scenario. Some vague ideas:
- sample code that, during normalization, warns if `editor.apply` is called with a `remove_node` operation (admittedly this would not be able to offer information about why it was removed)
- start building a system for runtime warning messages, perhaps using weak references so they don't cause memory leaks
- some sort of "development" mode to help ensure warnings are not shipped to production

dylans:
@TheSpyder ok, no problem, reverting now. I would vote for option 2 if feasible.

jameshfisher:
> Slate core does not emit warnings; that lint rule is there for a reason.

@TheSpyder what's the reason, more explicitly? That you don't want warnings emitted in production?

The point about collaborative editing is interesting - I'm not dealing with those kinds of problems.

For context, [here's how React does it](https://reactjs.org/docs/codebase-overview.html#warnings-and-invariants). I think this resembles @TheSpyder's option 3. It seems to work well. Perhaps we can use/copy/steal it.

TheSpyder:
> @TheSpyder what's the reason, more explicitly? That you don't want warnings emitted in production?

Pretty much. Slate is a framework to build editors, and I don’t like frameworks that emit console logs.

> For context, [here's how React does it](https://reactjs.org/docs/codebase-overview.html#warnings-and-invariants). I think this resembles @TheSpyder's option 3. It seems to work well. Perhaps we can use/copy/steal it.

React also has separate production and development builds. I’m not sure we want to go that far.

I guess the `process.env.NODE_ENV` approach is fairly common? We might need to be careful about introducing a check for that though.

Issue #4513, opened 2021-09-13
onCopy not firing for a collapsed selection using React 17


bryanph:
**Description**
See title and codesandbox, been trying to debug this in a react example without luck so it does seem a bug in slate although it only happens when using react 17 and only in firefox

**Sandbox**
https://codesandbox.io/s/slate-react-17-oncopy-bug-t54vc?file=/index.js
(make sure to open the example in a new window in firefox

**Steps**
1. Go to above sandbox in firefox
2. Open the example in a new window
3. Do ctrl-c without selecting text
4. You don't see the console.log

**Expectation**
onCopy is triggered

**Environment**
- Slate Version: current master
- Browser: Firefox

dylans:
The code for onCopy is pretty terse:

https://github.com/ianstormtaylor/slate/blob/48b7129447347c9cf7a0535026287896ef59779b/packages/slate-react/src/components/editable.tsx#L816-L827

My best guess is that the following is true in FF but also in other browsers:

```ts
if (
  hasEditableTarget(editor, event.target) &&
  !isEventHandled(event, attributes.onCopy)
) {
```

which then prevents the default event, and then calls setFragmentData, but if the selection is empty, setFragment data returns:

https://github.com/ianstormtaylor/slate/blob/48b7129447347c9cf7a0535026287896ef59779b/packages/slate-react/src/plugin/with-react.ts#L134-L138

bryanph:
@dylans thanks for looking into it :smile:. If I add an event listener breakpoint for the `copy` event firefox it doesn't get triggered as well though so it seems to me like it doesn't get triggered at all. I've been trying to reproduce this independently from slate without success though (just react 17 and some content editable code) so I might be missing something here. As for the `setFragment` there, the selection isn't null in the case I was trying to describe above but rather collapsed, so it shouldn't halt there.

skogsmaskin:
I'm experiencing the same. What is interesting is that it works in the images example (https://www.slatejs.org/examples/images), but not when I try to do the same in my code.

If you only select a void block in Firefox, `onCopy` and `onCut` and `onCopy` events doesn't trig at all, as @bryanph says. 

However if I press `arrowDown`, it will trig the events (though the editor slate selection object is still the same).

I suspect it is related to there not being a real DOM-range set and Firefox thinks nothing is selected and this will not fire the events.

I see we have some special Firefox code hacks here and there (`el.focus()`), could it be related? Maybe some special treatment for FF is missing?

[...]

PR #4775, opened 2022-01-07
Fix broken single-void-selections in Firefox

skogsmaskin:
**Description**
Firefox will not trig `insertData` methods when selecting single void nodes.
Make sure that the range has 0-1 (not 1-1) offset so something is selected if this is the case.
This obviously works fine in other browsers, but Firefox seems to need a real range here.

**Issue**
Fixes https://github.com/ianstormtaylor/slate/issues/4513

**Example**

**Context**

**Checks**


skogsmaskin:
Found one side-effect. Arrowing up and down between void blocks now takes two key-presses.

I think it's because we don't handle up/down events in a consistent way, but leaves that up to the default browser behaviour.

Firefox seems to handle this a bit differently than the other browsers.

I've made a new commit that handles these events over void blocks with custom code.

dylans (approved):
Looks great, thanks for fixing this! If you have time please add a changeset. I’ll leave it open for a couple of days for feedback and then land it unless there are objections.

skogsmaskin:
@dylans - Updated the first commit to be even more agressive, checking for focus on a void block all together. Now it's working with selection spans as well.

Will add changeset. Thanks for reviewing!

bryanph:
I think it would be a good idea to first come up with a reproducible example that consistently demonstrates this issue. I haven't been able to yet (my failed attempts in https://github.com/ianstormtaylor/slate/issues/4513#issuecomment-1007435032) and since the void image example works fine we might be looking at the wrong source for this? I don't think it's general behavior in firefox that a collapsed selection does not trigger an oncopy event, I've been able to confirm that much in the codesandbox at least.

skogsmaskin:
Extend selection up/down should also be handled in a consistent way, added that.

@bryanph - please note that the missing clipboard events in FF is only an issue when focus is at a void node.

dylans:
> I've made this sandbox that illustrates the problem:
> 
> https://codesandbox.io/s/slate-react-17-oncopy-bug-not-working-forked-0uyde?file=/index.js
> 
> I see there are some issues still which the PR doesn't solve when testing against the above code example. I think this needs to be investigated further.

https://codesandbox.io/s/slate-react-17-oncopy-bug-not-working-forked-qg78q?file=/index.js has a couple small fixes to your sandbox.

I'll leave this open for now given your comment on needing to investigate this further.

skogsmaskin:
@dylans - sorry I forgot to save it! Not it is there with the latest `slate-react `

skogsmaskin:
If one make a shift-arrow-right when a void (only) is selected it will fire onCut events. Presumably the selection in slate is the same.

bryanph:
Updated your example a bit to remove some errors https://codesandbox.io/s/slate-react-17-oncopy-bug-not-working-forked-mow9x?file=/index.js.

That seems like a decent reproduction, I'll try and see if the same happens in plain contenteditable

bryanph:
In this very simple contenteditable example I also can't get `onCut` to trigger: https://codesandbox.io/s/dazzling-elbakyan-8qgtt?file=/src/App.js:94-204

skogsmaskin:
Put this back to draft. Not sure if we should try to work around this or wait for React to fix the bug. So far there's not response in the issue that @bryanph filed with them.

bryanph:
I can imagine it will take a while to fix on the react side, also who knows it might be some firefox quirk that is hard to resolve. So overall I'd be a proponent of a workaround. Would be nice to get a response from the react team on what the underlying cause might be though, can maybe bump the thread once in a while.

Thinking about the issue a little more I just realized that in react 17 the event handlers are no longer attached to the document root. So I tried to see if it worked if you attach an event to the document root and voila it seems to work, see https://codesandbox.io/s/contenteditable-cut-react-18-working-drqnl?file=/src/App.tsx:179-188

So I guess all we have to do is attach a `cut` listener to `document` that doesn't do anything, in that case the `onCut` event is also fired on the `Editable` div

dylans:
I'd suggest that we fix per @bryanph's most recent note and keep an eye on the React issue.

skogsmaskin:
Yes, this seems like the way to go! Thanks @bryanph!

dylans:
@bryanph Now that we require React 18, perhaps this can be resolved?

dylans:

PR #4779, opened 2022-01-08
Fix Android editor.insertText regression

alessiogaldy:
**Description**
- Remove logic to delay the handling of text insertion
- Call `Transforms.setSelection` before `Editor.insertText` to adjust position

**Issue**
Fixes: #4709
Based on discussion from #4753 

**Checks**

### DIFF
--- a/.changeset/dry-worms-happen.md
+++ b/.changeset/dry-worms-happen.md
@@ -1,0 +1,8 @@
+---
+'slate-react': patch
+---
+
+Android editable updates
+
+- Remove logic to delay handling of text insertion
+- Call Transforms.setSelection before Editor.insertText to adjust position
--- a/packages/slate-react/src/components/android/android-editable.tsx
+++ b/packages/slate-react/src/components/android/android-editable.tsx
@@ -24,11 +24,7 @@ import {
   IS_READ_ONLY,
   NODE_TO_ELEMENT,
   PLACEHOLDER_SYMBOL,
-  IS_COMPOSING,
-  IS_ON_COMPOSITION_END,
-  EDITOR_ON_COMPOSITION_TEXT,
 } from '../../utils/weak-maps'
-import { normalizeTextInsertionRange } from './diff-text'
 
 import { EditableProps, hasTarget } from '../editable'
 import useChildren from '../../hooks/use-children'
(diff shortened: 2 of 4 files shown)


dylans (commented):
Ideally @YasinChan could give this a review since they added some of these changes recently. I want to make sure we're not breaking things that were fixed in #4734.

YasinChan:
Hi @alessiogaldy @dylans ~ I had tested this PR and so far I think it is a better way!

However, I found a problem during testing. In some IMEs such as the sougou pinyin IME,  when typing after a line break, the cursor will go to the beginning of the line when entering the third character. After debugging, it was found that when using these IMEs, the 
https://github.com/ianstormtaylor/slate/blob/e3a325f8a6f2009afd5f8363ace1a4849e1fc4ae/packages/slate-react/src/components/string.tsx#L36
was not successfully triggered when the first character was entered at the beginning of the line after the line break, so the zero-width components were not replaced, which eventually led to the problem as shown in the video below.

I'll try to fix this after this pr merged.


https://user-images.githubusercontent.com/22654945/148875999-1d46ec59-80ed-4912-9c09-d5d0a2843250.mp4



https://user-images.githubusercontent.com/22654945/148875949-9b378251-60a4-42db-a24a-324dac106024.mp4

dylans:
> I'll try to fix this after this pr merged.

@YasinChan it's merged, I'll hold off on a release until you get a chance to raise a PR. Thanks!

YasinChan:
Hi @alessiogaldy @dylans ~ I tried to solve the problems mentioned above, but seems to have some difficulties. Could you give me some advice? 

By the way I checkout to this pr #4753 and test it, the regression you mentioned in https://github.com/ianstormtaylor/slate/pull/4753#issuecomment-1007965338 seems to be the problem shown in this video? 

https://user-images.githubusercontent.com/22654945/149158168-0dbe6dfd-1504-4eaf-92c3-fd00b84dfed1.mp4

I attempt to change [this code](https://github.com/alessiogaldy/slate/blob/5880553fcb7b85652cebed7b25a985ef931427ad/packages/slate-react/src/components/android/android-input-manager.ts#L128-L138) to
```
Transforms.setSelection(this.editor, at)
Editor.insertText(this.editor, text)
```
and everything seems to be all right. I think it may be not necessary to remove the other logic in this PR ? 🤔

alessiogaldy:
@YasinChan correct, I removed the delay logic and the issue in the video was resolved.
I'll try to add back the delay with the change you suggested to check if this solves the issue on my side.
Thank you!

alessiogaldy:
@YasinChan I tried the changes on my side and the cursor no longer jumps around.
Let me know if this works on for you as well.
Thank you!

PR #4790, opened 2022-01-14
Merge layout effects for adding focus and blur event listeners

davbrito:
**Description**

Use a single layout effect for subscribing to the focus and blur listeners. It is currently made in two separate effects.

**Issue**

**Example**

**Context**
**Checks**

### DIFF
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -1,11 +1,11 @@
-import React, { useMemo, useState, useCallback, useEffect } from 'react'
-import { Editor, Node, Element, Descendant } from 'slate'
-import { ReactEditor } from '../plugin/react-editor'
-import { FocusedContext } from '../hooks/use-focused'
-import { EditorContext } from '../hooks/use-slate-static'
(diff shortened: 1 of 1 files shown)


chase:
Looks like you need to run `yarn fix` since unrelated code seems to have been updated with the wrong Prettier rules

davbrito:
@chase It's ready

chase:
@davbrito I'm not a maintainer :sweat_smile: Just noticed you didn't run the steps required in the checklist for a PR.

dylans (commented):
The change looks fine. I would like to better understand the pros and cons of this change (if any beyond more concise code).

Or if there's a resource out there that explains why focus and blur should live inside the same layout effect, a link would be great.

davbrito:
Hi @dylans. Technically there is no difference in functionality between using two effects and using one. But IMO there is no need for having to register one additional effect if you can use the one that you already have and that does the same thing.

Another advantage is having a more compact code and to reuse the listener function. So if you need to change something there, you just need to change it on a single place.

dylans (approved):

PR #4804, opened 2022-01-26
update insertText logic when selection is not collapsed

suilang:
**Description**
When I select a paragraph with different styles of text, for example, bold and plain, and then insert the letter q. What I want is to insert a non-bold letter.

**Example**
this is old code, the q is bold
chrome-capture

this is new code, the q is normal
chrome-capture1

**Context**

chrome-capture 2

Normally, when we insert the character q between unbold and bold, it is unbold. the style use position of anchor. but when selection is not collapsed, selection be set by focus or end of range.

The processing logic is inconsistent for different actions. also, the cursor should be placed at the starting position. If the deleted area has elements that cannot be deleted, like table, the cursor position will be incorrect

**Checks**
- [X] You've [added a changeset]

### DIFF
--- a/.changeset/silly-flowers-worry.md
+++ b/.changeset/silly-flowers-worry.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+update insertText logic when selection is not collapsed
--- a/packages/slate/src/transforms/text.ts
+++ b/packages/slate/src/transforms/text.ts
@@ -483,12 +483,11 @@ export const TextTransforms: TextTransforms = {
         if (Range.isCollapsed(at)) {
           at = at.anchor
         } else {
-          const start = Range.start(at)
-
-          if (!voids && Editor.void(editor, { at: start })) {
+          const end = Range.end(at)
+          if (!voids && Editor.void(editor, { at: end })) {
             return
           }
-
+          const start = Range.start(at)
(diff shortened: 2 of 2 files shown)


dylans (commented):
Thanks @suilang . I'm not sure if it matters here, but have you checked selecting in reverse to make sure we don't break due to the direction of the selection (some parts of the codebase are good about reversing the selection and some are not).

suilang:
> 

You mean selecting text from back to front? In there i use `Range.start` to get start point, it make sure that start is always brefore end point。

dylans (approved):

imbolo:
@suilang @dylans
After this change，it throws an error when I select two lines and input any character.

https://user-images.githubusercontent.com/3530602/153788940-0ca8cf69-918c-4088-8042-5b14dc048fcb.mov

cmditch:
Also experiencing errors with this in production. We've reverted back to 0.72.3.

dylans:
Reverted as noted above. Hopefully we can find a better solution for the issue identified in this original PR without breaking other behavior.

suilang:
chrome-capture

dylans:
@suilang it's ok, rich text editing is not easy. Feel free to raise a new PR with tests and we'll evaluate it.

Issue #4801, opened 2022-01-23
applying Grammarly suggestions when the editor is not in focus is broken (Chrome)


Schipy:
**Description**
Applying a suggestion via Grammarly extension, when the text editor is not in focus, is broken and will insert replaced text to the front of the text instead of the proper place.

It happens with Grammarly extension for Chrome, but appears to be working properly under Safari. Also applying native spellcheck suggestion when text is not in focus - on browsers where this is allowed, eg Firefox, and not Chrome - seems to work.

Unfortunately even with the reduced scope, this issue makes Grammarly unusable under Chrome, which is a blocker issue for production usage of the library in certain contexts.

**Recording**
![grammarly_chrome](https://user-images.githubusercontent.com/2234706/150681550-4d271246-f146-4254-b651-20bf00cfd258.gif)

**Sandbox**
https://www.slatejs.org/examples/richtext

**Steps**
To reproduce the behavior: as per short recording. Important detail is that the issue happens when the text editor is **not** in focus when the replacement is made.

**Expectation**
Text suggestion should be replaced as per intention of the extension.

**Environment**
- Slate Version: latest
- Operating System: Mac
- Browser: Chrome

**Context**
May be related: https://github.com/ianstormtaylor/slate/issues/4579

PR #4809, opened 2022-01-26
fix: flush onDOMSelectionChange on onDOMBeforeInput

BitPhinix:
**Description**
Fixes the same fundamental issue as #4669 did, only triggered in a slightly different way. If the editor isn't focused the first select causes the 2nd one that Grammarly causes to be delayed by the `onDOMSelectionChangeonDOMSelectionChange` throttle.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4801

**Example**
Before:
before

After:
after

**Checks**

### DIFF
--- a/.changeset/wise-kings-serve.md
+++ b/.changeset/wise-kings-serve.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Flush onDOMSelectionChange on onDOMBeforeInput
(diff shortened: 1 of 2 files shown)


BitPhinix:
Integration tests pass locally, something seems to be up with ci. Maybe just needs a re-run.

dylans (approved):

PR #4813, opened 2022-01-27
fix: toSlatePoint suppressThrow leaf without text node

BitPhinix:
**Description**
`toSlatePoint` throws while using `suppressThrow` if the selection is inside a node with a `data-slate-leaf` attribute with no text child (due to calling `range.setStart` with `textNode` being `null`). This can happen when e.g. rendering a slate 0.47 editor inside a void element.

This pr adds a simple check to see if `textNode` is set.

**Checks**

### DIFF
--- a/.changeset/kind-olives-add.md
+++ b/.changeset/kind-olives-add.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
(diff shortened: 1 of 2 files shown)


BitPhinix:
Ci passes locally

dylans (approved):

PR #4814, opened 2022-01-28
Fix failing unhangRange test case with inlines

bryanph:
**Description**

Adding this here for now to get some help on trying to fix this particular bug and can then modify it into an actual fix later on.

Take a look at the test that I added. It fails right now.

It returns:
```
{
  anchor: { path: [ 1, 0 ], offset: 0 },
  focus: { path: [ 0, 0 ], offset: 12 }
}
```

which is clearly not what we want. Adding this here for now so we can explore why this happens.

@e1himself you've looked at this code recently (thanks for adding the tests! They're a big help.). Do you have some idea what might be going wrong here?

**Checks**

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -1656,7 +1656,7 @@ export const Editor: EditorInterface = {
       match: n => Editor.isBlock(editor, n),
     })
     const blockPath = endBlock ? endBlock[1] : []
-    const first = Editor.start(editor, [])
+    const first = Editor.start(editor, start)
     const before = { anchor: first, focus: end }
     let skip = true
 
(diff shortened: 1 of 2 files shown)


bryanph:
An update: I changed a line that seems to make the tests pass. (see commit)
image

It started scanning at the beginning of the editor, now it starts at the beginning of the cursor. That seems to make the tests pass but I don't quite understand it was starting at the beginning of the editor there. Is there something I might be missing here?

dylans:
> An update: I changed a line that seems to make the tests pass. (see commit) image
> 
> It started scanning at the beginning of the editor, now it starts at the beginning of the cursor. That seems to make the tests pass but I don't quite understand why it was starting at the beginning of the editor there. Is there something I might be missing here?

It looks like the empty array was probably causing scanning of the entire editor iirc. Thanks for the fix.

Issue #4818, opened 2022-02-02
State update occurs when unmounting


ugaya40:
**Description**
Hi, I have a problem that is difficult to reproduce with minimal code.
However, what is happening is very clear.

The phenomenon is that after the slate-react/Slate component is unmounted, a focus-related state update occurs and I get a react warning.
<img width="839" alt="warning" src="https://user-images.githubusercontent.com/1430166/152118620-7c7c4ef6-6644-4d9d-b054-f43af5944f2b.png">

Also, the cause is clear.
   https://github.com/ianstormtaylor/slate/blob/4b97197f9ae6e0ca453c47d363c121a9dbbeb904/packages/slate-react/src/components/slate.tsx#L60-L64
In the above line, the presence of setTimeout may cause the state to be updated even after unmounting. (This is what is happening in my relatively complex project).

I think the fix is to determine if it has been unmounted before setIfFocused.

It may be inappropriate to raise this issue in such an incomplete state, but I would appreciate any feedback.

PR #4819, opened 2022-02-04
fix not update state when unmounted

ugaya40:
**Description**
Changed so that the state of the slate component is not updated when unmounting.

**Issue**
Fixes: #4818 

**Checks**

### DIFF
--- a/.changeset/hot-beers-poke.md
+++ b/.changeset/hot-beers-poke.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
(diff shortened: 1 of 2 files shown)


dylans (approved):
Thanks, looks reasonable. Will land it once the tests pass.

dylans (commented):
Ok just needs a change set and then we can land this.

ugaya40:
> 

I'll try.

ugaya40:
Added changeset.

Issue #4836, opened 2022-02-13
Bug: `setNodes(editor, { foo: null })` removes the property `foo`


jameshfisher:
**Description**
`Transforms.setNodes(editor, { foo: null })` removes the property `foo` instead of setting its value to `null`.

**Recording**
(n/a: not a UI/React bug)

**Sandbox**
https://codesandbox.io/s/slate-bug-setnodes-with-value-null-removes-property-97rok

**Steps**
To reproduce the behavior:
1. Go to https://codesandbox.io/s/slate-bug-setnodes-with-value-null-removes-property-97rok
2. Click on "Set element.foo = null"
3. See `element` no longer has the property `foo`

**Expectation**
The element has the property `foo` with the value `null`.

**Environment**
- Slate Version: 0.72.8
- Operating System: Windows
- Browser: Chrome, Firefox

**Context**
I would potentially accept this behavior with the value `undefined`, but definitely not with `null`.

jameshfisher:
I now see there is a test case that asserts this buggy behavior:

https://github.com/ianstormtaylor/slate/blob/8eb1abac87809e650507248806e9ee7585cf334f/packages/slate/test/operations/set_node/remove-null.tsx#L1-L29

I disagree with this test case. This says it's testing that "newProperties should omit removed values", but `{foo: null}` is not removing `foo`; it's setting it to `null`.

jameshfisher:
Principle: "Slate's core logic assumes very little about the schema of the data you'll be editing". If it assumes that properties are not `null`, that should be listed as one of its assumptions.

The docs also talk repeatedly about serializing to JSON. Slate should at least accept JSON data types, of which `null` is one. (But `undefined` is not, which is one reason I would accept the current behaviour with `undefined`.)

jameshfisher:
This behavior seems to date back at least to

https://github.com/ianstormtaylor/slate/pull/3235/files#diff-7638f987cb9760bd6d58c4c20af87d611ef823d2c77a9aba5a39876b2acf6a06R231-R232

where @ianstormtaylor made the `set_node` operation remove the property if its value is set to `null`. I would say, if we want this design, we should be using `undefined` as the tombstone instead of `null`. However, I see in https://github.com/ianstormtaylor/slate/pull/4078 by @TheSpyder that we avoid `undefined` in operations, so that operations are JSON-serializable. I assume this JSON-serializable requirement comes from operations being used in collaboration/synchronization protocols.

[...]

PR #4838, opened 2022-02-14
Document additional schema constraints due to design of operations

jameshfisher:
**Description**
Slate imposes additional constraints on the schema that are not discussed. In particular, it assumes properties are JSON-serializable, and does not allow properties to be `null`. Developers need to know these constraints before they design their schema. In my case, I designed a schema that allowed `null`, and now have live documents with this value -- I would not have to fix this if I knew the constraints up front.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4836


jameshfisher:
I'm aware that these constraints are not actively enforced by built-in normalizations, but I still think this is the most appropriate place for these docs: it's the one place where the docs describe schema constraints. (Also these constraints could, or should, be actively enforced by built-in normalizations.)

TheSpyder (approved):
Thanks for this. I'm sorry you ended up with live `null` value docs in the meantime.

We might also need a note in the `setNodes` documentation about it 🤔

dylans:
Going to land what is here, thanks! Feel free to add a separate PR for the feedback, "We might also need a note in the setNodes documentation about it 🤔"

PR #4840, opened 2022-02-15
render void spacer in readonly mode

BitPhinix:
**Description**
Currently finding the dom range of a void node in read-only crashes the editor since slate doesn't render the void spacer in read-only.

This pr simply adjusts the Element component so that it is rendered even in read-only.

> ⚠️  I'm not sure why this wasn't the case before, so there might be something I'm missing here. 

**Checks**

### DIFF
--- a/.changeset/hip-pigs-film.md
+++ b/.changeset/hip-pigs-film.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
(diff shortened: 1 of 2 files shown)


dylans:
> I'm not sure why this wasn't the case before, so there might be something I'm missing here.

I'm guessing oversight rather than intentional. I'll land this now, thanks.

PR #4841, opened 2022-02-15
Feature/slate selectors

Fibs7000:
**Description**
As we have been using slate as a company to quite a large editor we ran into several rerendering issues for editor menus.

This PR solves the rendering issues in a redux selector inspired style, so every component which has been using useSlate (which of course every toggle button or any other component which needs the current editor state) has to use, will now use the useSlateSelector hook.

### Example:

lets define Formatting.isBoldActive as an function which checks if the current selection is in a text node which has a bold style

#### Before
```typescript
const editor = useSlate();
const isBoldActive = Formatting.isBoldActive(editor);
```
This will rerender on every keystroke or selection navigation.

#### After
```typescript
const isBoldActive = useSlateSelector(Formatting.isBoldActive); 
// or useSlateSelector(editor => Formatting.isBoldActive(editor))
```
This will rerender ONLY if the state changes (eg. from false to true).

The hook will also accept a second argument, a compare function to deal with objects and arrays

### Visual Example
Have a look at the right toolbar - this is what I mean with many rerenders... also performance of the editor is pretty slow at that stage. In the after example you can clearly see how often this will rerender.

**Before:**
chrome-capture-2022-1-15 (2)

**After:**
chrome-capture-2022-1-15 (1)

**Checks**

### DIFF
--- a/.changeset/serious-papayas-smoke.md
+++ b/.changeset/serious-papayas-smoke.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+Added redux-style useSlateSelector to improve and prevent unneccessary rerendering with the useSlate hook
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -4,6 +4,10 @@ import { ReactEditor } from '../plugin/react-editor'
 import { FocusedContext } from '../hooks/use-focused'
 import { EditorContext } from '../hooks/use-slate-static'
 import { SlateContext } from '../hooks/use-slate'
+import {
+  getSelectorContext,
+  SlateSelectorContext,
+} from '../hooks/use-slate-selector'
 import { EDITOR_TO_ON_CHANGE } from '../utils/weak-maps'
 import { useIsomorphicLayoutEffect } from '../hooks/use-isomorphic-layout-effect'
 
@@ -38,9 +42,15 @@ export const Slate = (props: {
     return [editor]
   })
 
+  const {
+    selectorContext,
+    onChange: handleSelectorChange,
+  } = getSelectorContext(editor)
+
   const onContextChange = useCallback(() => {
     onChange(editor.children)
     setContext([editor])
+    handleSelectorChange(editor)
   }, [onChange])
 
   EDITOR_TO_ON_CHANGE.set(editor, onContextChange)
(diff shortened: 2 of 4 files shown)


Fibs7000:
Ok so I really don't know why this test is failing... did not change anything which could potentially break this test.

Would someone be so kind and give me some feedback on this PR?

dylans:
> Ok so I really don't know why this test is failing... did not change anything which could potentially break this test.
> 
> Would someone be so kind and give me some feedback on this PR?

I don't think the test is failing from this change. Integrations tests are flaky at times with the GH CI/CD environment.

Your proposed change looks very promising, I need some time to digest it and generally like to leave things open for at least a few days to give others in the community time to review and give feedback.

dylans:
Given the lack of feedback, I'm going to land it. If it causes significant issues we may need to revert it, but I don't want to get in the way of this type of improvement.

davbrito:
I think this could be vastly simplified by using the `useSyncExternalStore` hook.

zhugexinxin:
> 我认为这可以通过使用`useSyncExternalStore`钩子来大大简化。

That's a good idea

bryanph:
This is as very nice addition (haven't really tested it tho). As mentioned above it would be nice indeed if we could use something provided directly by the react team if possible to make it less complex and hence error prone.

davbrito:
> This is as very nice addition (haven't really tested it tho). As mentioned above it would be nice indeed if we could use something provided directly by the react team if possible to make it less complex and hence error prone.

There is currently a release candidate for the `use-sync-external-store` package. Which is a shim for that hook. That could be used for this.

Fibs7000:
> I think this could be vastly simplified by using the `useSyncExternalStore` hook.

Thats a great Idea! Didn't know that hook til now. Will try if it can replace the complex callback and sync logic

zhugexinxin:
> > I think this could be vastly simplified by using the `useSyncExternalStore` hook.
> 
> Thats a great Idea! Didn't know that hook til now. Will try if it can replace the complex callback and sync logic

Looking forward to it.

davbrito:
> > > I think this could be vastly simplified by using the `useSyncExternalStore` hook.
> > 
> > 
> > Thats a great Idea! Didn't know that hook til now. Will try if it can replace the complex callback and sync logic
> 
> Looking forward to it.

This is the official shim package https://www.npmjs.com/package/use-sync-external-store
I think the RC version is stable enough to be used here.

PR #4848, opened 2022-02-21
fix point transform for insert_text to account for affinity

rockettomatooo:
**Description**
When one creates a RangeRef with affinity `outward` and the ref becomes collapsed, it can't be expanded again with an `insert_text` operation (although it should IMO). 

**Example**
Lets take an example string (and not care about paths for simplicity sake)
```
This is a test
```
The RangeRef is the following
```ts
{ 
  affinity: 'outward',
  current: { 
    anchor: { offset: 4 }, 
    focus: { offset: 4 } 
  },
}
```

Now if the user types something at offset 4 (right after the first word) the RangeRef right now changes to the following
```ts
{ 
  affinity: 'outward',
  current: { 
    anchor: { offset: 5 }, 
    focus: { offset: 5 } 
  },
}
```
Although I naively expected it to be this.
```ts
{ 
  affinity: 'outward',
  current: { 
    anchor: { offset: 4 }, 
    focus: { offset: 5 } 
  },
}
```

**Context**

If you take a collapsed RangeRef with affinity `outward` and apply an `insert_text` operation at the offset of where the Range is the following happens: 
Because the collapsed ranges count as forward pointing ranges, the `Range.transform()` will determine the anchors' affinity as `backward`. Therefor it will run the `Point.transform()` on the anchor with that affinity. But, because the `Point.transform()` doesn't account for the affinity, the offset of the anchor gets incremented regardless. The same happens with the focus thus the Range will stay collapsed.

**Checks**

### DIFF
--- a/.changeset/fast-doors-peel.md
+++ b/.changeset/fast-doors-peel.md
@@ -1,0 +1,5 @@
+---
+'slate': minor
+---
+
+fix point transform for insert_text operations to account for affinity
--- a/packages/slate/src/interfaces/point.ts
+++ b/packages/slate/src/interfaces/point.ts
@@ -110,7 +110,11 @@ export const Point: PointInterface = {
         }
 
         case 'insert_text': {
-          if (Path.equals(op.path, path) && op.offset <= offset) {
+          if (
+            Path.equals(op.path, path) &&
+            (op.offset < offset ||
+              (op.offset === offset && affinity === 'forward'))
+          ) {
             p.offset += op.text.length
           }
 
--- a/packages/slate/test/interfaces/Point/transform/backward-insert-text-after-point.tsx
+++ b/packages/slate/test/interfaces/Point/transform/backward-insert-text-after-point.tsx
@@ -1,0 +1,24 @@
+import { Point } from 'slate'
+
+export const input = {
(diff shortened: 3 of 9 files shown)


rockettomatooo:
@BrentFarese @dylans I'm not sure if I also need to touch `remove_text` operations. The use case would be when you remove something where the focus is and you want it to stay rather than shrink (if that makes any sense). I can add this as well. 
But I don't think its as straight forward and I'm not sure I understand the contents of the `remove_text` operation enough. From what it looks like, the `offset` for `remove_text` operation is the new offset (meaning the offset after the contents was removed) whereas in an `insert_text` operation, the `offset` is the offset before the text was inserted.  
Right now, the `remove_text` operation checks for `op.offset <= offset` but I'm not sure this is actually correct. Shouldn't it be `op.offset + op.text.length <= offset`? wdyt?

rockettomatooo:
As for the failing integration test: The iframe example seems to be failing because it can't find an element with `role="textbox"` within the iframe (in time). This doesn't seem to be related to the changes in this PR. I ran it locally a couple of times and those integration tests seem a bit flaky. Any ideas?

dylans (approved):

PR #4856, opened 2022-02-27
DOCS: improved deserialization example

janpaepke:
**Description**
The initial motivation for this was that the example did not provide details on how to deserialize marks.

Additionally I feel it's okay to return successfully serialized child nodes from unrecognized elements (default fallback of element switch), rather than just its text-content.
This also required the child map to be flattened.

While I was at it, I also replaced the numbers in the `nodeType` checks with their respective Constants (`Node.TEXT_NODE` and `Node.ELEMENT_NODE`), which makes the code easier to grasp for a reader.

**Context**
It took me quite a while how to figure out how to properly deserialize marks. An example is included for serialization (strong) but not for deserialization.

**Disclaimer**
While I have tested this in my actual code, the original is written in typescript.
So I may have made an error when converting it.
If you feel this is not an improvement, feel free to close.

### DIFF
--- a/docs/concepts/10-serializing.md
+++ b/docs/concepts/10-serializing.md
@@ -168,16 +168,18 @@ const deserialize = (el, markAttributes = {}) => {
   } else if (el.nodeType !== Node.ELEMENT_NODE) {
     return null
   }
-  
-  const nodeAttributes = { ...markAttributes };
-  
+
+  const nodeAttributes = { ...markAttributes }
+
   // define attibutes for text nodes
   switch (el.nodeName) {
     case 'strong':
-      nodeAttributes.bold = true;
+      nodeAttributes.bold = true
   }
-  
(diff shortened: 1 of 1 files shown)


dylans:
I’ll review asap

dylans (commented):
This looks fine to me. The automated tools are complaining about the formatting, so please run prettier/lint and push an update and then I'll land it. Thanks @janpaepke !

janpaepke:
Some might say using prettier rule enforcing for JS in markdown might be overdoing it, but rules are rules, I guess. :)

side note: I use npm over yarn, which breaks the pre-commit hook.

dylans:
> Some might say using prettier rule enforcing for JS in markdown might be overdoing it, but rules are rules, I guess. :)

I had the same thought the first time I raised a PR, but I'm used to it now. 😂

dylans (approved):

PR #4873, opened 2022-03-08
Add soft break as a separate overridable editor method

bryanph:
**Description**
A different behavior for inserting a soft break with `shift+enter` is quite common in rich text editors. Right now you have to do this in `onKeyDown` which is not so nice. This adds a separate `insertSoftBreak` method on the editor instance that gets called when a soft break is inserted. I made sure to maintain the current default behavior for backwards compatibility (it just splits the block). But at least you can easily overwrite it now.

If you rely on overwriting `editor.insertBreak` for extra behavior for soft breaks this might be a breaking change for you and you should overwrite `editor.insertSoftBreak` instead.

**Issue**
Fixes: (link to issue)

**Checks**

### DIFF
--- a/.changeset/fair-forks-hope.md
+++ b/.changeset/fair-forks-hope.md
@@ -1,0 +1,8 @@
+---
+'slate': minor
+'slate-react': minor
+---
+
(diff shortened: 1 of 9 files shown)


dylans (commented):
overall I like it, but I'm going to leave it open for a few days for others to give feedback.

Issue #4886, opened 2022-03-12
"Cannot resolve from DOM point" error on onDomSelectionChange for readonly void elements


heyitsaamir:
**Description**
If you click / select a void element when the editor is in readonly mode, it throws a "Cannot resolve from DOM point" error.

**Recording**

https://user-images.githubusercontent.com/48929123/158026541-719c5c25-dde7-495c-907b-ab5ae2748d79.mov

**Sandbox**
https://codesandbox.io/s/slate-reproductions-forked-g486e7?file=/index.js

**Steps**
To reproduce the behavior:
1. Create a new type of element that's void.
2. Put your editor in reaadonly mode
3. Click on the read only void element.

**Expectation**
Shouldn't crash.

**Environment**
- Slate Version: [e.g. 0.71.0]
- Slate React: 0.72.8
- Operating System: [macOS]
- Browser: [Firefox]

**Context**
I have a local fix in draft - https://github.com/ianstormtaylor/slate/pull/4887
More context: I took the image example and tried to reproduce it here: https://codesandbox.io/s/slate-reproductions-forked-g486e7?file=/index.js. It doesn't. The reason being that the image example has the `contentEditable` set to true. So I guess, the question then becomes, should all `void` elements set `contentEditable` to false?

ahoisl:
I just tried to reproduce your problem to see how it differs from my original problem with readonly void element (#4726) and noticed that the error also happens if the editor is _not_ readonly:

https://codesandbox.io/s/slate-reproductions-forked-k8hnlz

heyitsaamir:
> I just tried to reproduce your problem to see how it differs from my original problem with readonly void element (#4726) and noticed that the error also happens if the editor is _not_ readonly:
> 
> https://codesandbox.io/s/slate-reproductions-forked-k8hnlz

Thanks @ahoisl ! Seems like if we set `contentEditable` to false, then the problem goes away. (Added this in the original issue). I wonder if that's the recommended approach then?

cmmartin:
@heyitsaamir I believe it is the recommended approach. IIRC there was a Firefox bug that required this previously anyway. You can see it's been done in the examples [here](https://github.com/ianstormtaylor/slate/blob/main/site/examples/editable-voids.tsx#L71) and [here](https://github.com/ianstormtaylor/slate/blob/main/site/examples/images.tsx#L102)

ahoisl:
[...]

PR #3649, opened 2020-05-04
Fixing selection across a void node (i.e. mention or image) with the mouse

jerkly:
Related pull #3572 
#### Is this adding or improving a _feature_ or fixing a _bug_?
fixing a bug

#### What's the new behavior?
screencast-localhost_3000-2020 05 04-05_53_32
screencast-localhost_3000-2020 05 04-05_52_52

#### Have you checked that...?

- [x] The new code matches the existing patterns and styles.
- [x] The tests pass with `yarn test`.
- [x] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [x] The relevant examples still work. (Run examples with `yarn start`.)

#### Does this fix any issues or need any specific reviewers?

Fixes: #3647 #3648
Reviewers: @

Issue #3648, opened 2020-05-04
Selection via mouse across void nodes not worked


jerkly:
#### Do you want to request a _feature_ or report a _bug_?
bug

#### What's the current behavior?

Selection with mouse are collapsed, when hitting void node's edge.

Mention example: 
https://www.slatejs.org/examples/mentions
![screencast-www slatejs org-2020 05 04-05_35_00](https://user-images.githubusercontent.com/8193821/80932811-c8800f00-8dc9-11ea-9ecb-d147a5a2285f.gif)

Image example:
https://www.slatejs.org/examples/images
![screencast-www slatejs org-2020 05 04-05_43_31](https://user-images.githubusercontent.com/8193821/80932968-6a076080-8dca-11ea-8f56-60b76467ff80.gif)

Slate: 0.XX.X
Browser: Chrome / Safari / Firefox / Edge
OS: Mac / Windows / Linux / iOS / Android

#### What's the expected behavior?
Selection is not collapsed, when selecting via mouse across void nodes

jerkly:
Selection with keyboard arrows works as expected.

knubie:
Was this fixed in #3594? (Potential duplicate of #1969)

jerkly:
@knubie, #3594 is not fixed selection in all cases (only selection via keyboard). Bug are still present in Slate 0.57.2.
Look at the mention and image example https://www.slatejs.org/examples/mentions and try to select across voids via mouse.

knubie:
> Bug are still present in Slate 0.57.2.

#3594 Is not included in 57.2. Could you try `master`?

jerkly:
> > Bug are still present in Slate 0.57.2.
> 
> #3594 Is not included in 57.2. Could you try `master`?
Master checked again. Voids with extra content (i.e mention or embed) selectable only via keyboard.
With mouse selection there are two cases:
- selection via `Shift + mouse_click` working
- selection via `mouse_move` not working

jerkly:
fixed in #3649

PR #4887, opened 2022-03-12
Fix Cannot find DOM for readonly void elements

heyitsaamir:
**Description**
Modified the logic for the editor to gracefully handle selections inside void elements of readonly editors.
Changed the `isTargetInsideNonReadonlyVoid` to be `isTargetInsideReadonlyVoid` because double negatives are confusing :)

PENDING: Tests based on feedback from @dylans.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4886

**Example**
Before:

https://user-images.githubusercontent.com/48929123/158206992-bb0348d6-ae4f-4881-beb1-2acb16f5c84a.mov

After:

https://user-images.githubusercontent.com/48929123/158207019-d0d1bbbd-0ed3-43de-9909-900a7ecdc569.mov

**Context**
The logic previously was:
```
const anchorNodeSelectable =
          hasEditableTarget(editor, anchorNode) ||
          isTargetInsideNonReadonlyVoid(editor, anchorNode)
```
The node selection wouldn't go inside the second if-statement because the first clause would return true (`hasEditableTarget` checks if this node target is in the same editor, which is true). The second clause also needs to be checked as well to see if the target is inside a readonly-void element. If it is, then we should avoid selection.

I'm not sure if this is the right long term approach. Void elements should be selectable too, but the behavior should probably be up to the element-renderer. 

**Checks**

### DIFF
--- a/.changeset/twenty-moles-breathe.md
+++ b/.changeset/twenty-moles-breathe.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+Fixes graceful handling of selecting void slements inside readonly editors
--- a/packages/slate-react/src/components/android/android-editable.tsx
+++ b/packages/slate-react/src/components/android/android-editable.tsx
@@ -30,7 +30,7 @@ import {
 } from '../../utils/weak-maps'
 import { normalizeTextInsertionRange } from './diff-text'
 
-import { EditableProps, hasTarget } from '../editable'
+import { EditableProps, hasTarget, isTargetInsideReadonlyVoid } from '../editable'
 import useChildren from '../../hooks/use-children'
 import {
   defaultDecorate,
@@ -239,12 +239,12 @@ export const AndroidEditable = (props: EditableProps): JSX.Element => {
           const { anchorNode, focusNode } = domSelection
 
           const anchorNodeSelectable =
-            hasEditableTarget(editor, anchorNode) ||
-            isTargetInsideNonReadonlyVoid(editor, anchorNode)
+            hasEditableTarget(editor, anchorNode) &&
+            !isTargetInsideReadonlyVoid(editor, anchorNode)
 
           const focusNodeSelectable =
-            hasEditableTarget(editor, focusNode) ||
-            isTargetInsideNonReadonlyVoid(editor, focusNode)
+            hasEditableTarget(editor, focusNode) &&
+            !isTargetInsideReadonlyVoid(editor, focusNode)
 
           if (anchorNodeSelectable && focusNodeSelectable) {
             const range = ReactEditor.toSlateRange(editor, domSelection, {
--- a/packages/slate-react/src/components/android/android-editable.tsx
+++ b/packages/slate-react/src/components/android/android-editable.tsx
@@ -30,14 +30,14 @@ import {
 } from '../../utils/weak-maps'
 import { normalizeTextInsertionRange } from './diff-text'
 
-import { EditableProps, hasTarget, isTargetInsideReadonlyVoid } from '../editable'
+import { EditableProps, hasTarget } from '../editable'
 import useChildren from '../../hooks/use-children'
 import {
   defaultDecorate,
   hasEditableTarget,
   isEventHandled,
   isDOMEventHandled,
-  isTargetInsideNonReadonlyVoid,
+  isTargetInsideReadonlyVoid,
 } from '../editable'
 
 import { useAndroidInputManager } from './use-android-input-manager'
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -287,12 +287,12 @@ export const Editable = (props: EditableProps) => {
         const { anchorNode, focusNode } = domSelection
 
         const anchorNodeSelectable =
-          hasEditableTarget(editor, anchorNode) ||
-          isTargetInsideNonReadonlyVoid(editor, anchorNode)
+          hasEditableTarget(editor, anchorNode) &&
+          !isTargetInsideReadonlyVoid(editor, anchorNode)
 
         const focusNodeSelectable =
-          hasEditableTarget(editor, focusNode) ||
-          isTargetInsideNonReadonlyVoid(editor, focusNode)
+          hasEditableTarget(editor, focusNode) &&
+          !isTargetInsideReadonlyVoid(editor, focusNode)
 
         if (anchorNodeSelectable && focusNodeSelectable) {
           const range = ReactEditor.toSlateRange(editor, domSelection, {
@@ -1419,19 +1419,17 @@ export const hasEditableTarget = (
 }
 
 /**
- * Check if the target is inside void and in an non-readonly editor.
+ * Check if the target is inside void and in an readonly editor.
  */
-
-export const isTargetInsideNonReadonlyVoid = (
+export const isTargetInsideReadonlyVoid = (
   editor: ReactEditor,
   target: EventTarget | null
 ): boolean => {
-  if (IS_READ_ONLY.get(editor)) return false
-
   const slateNode =
     hasTarget(editor, target) && ReactEditor.toSlateNode(editor, target)
-  return Editor.isVoid(editor, slateNode)
+  return IS_READ_ONLY.get(editor) && Editor.isVoid(editor, slateNode)
 }
+
 
 /**
  * Check if an event is overrided by a handler.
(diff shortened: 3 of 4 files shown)


dylans (commented):
Makes sense. We’re just starting to add tests for slate-react so it would be great if you can add a test (I won’t block for that, just want to start building up a decent test suite). Thanks.

heyitsaamir:
> Makes sense. We’re just starting to add tests for slate-react so it would be great if you can add a test (I won’t block for that, just want to start building up a decent test suite). Thanks.

Yup will do. I noticed that we use `react-test-renderer` to add unit tests. I've found using [react-testing-library](https://testing-library.com/docs/react-testing-library/example-intro) to be way more intuitive and simpler to build tests. Obviously not saying we need to add that in for this, but would be happy to add that in in a separate PR if you think it's a good idea.

@dylans

dylans:
> > Makes sense. We’re just starting to add tests for slate-react so it would be great if you can add a test (I won’t block for that, just want to start building up a decent test suite). Thanks.
> 
> Yup will do. I noticed that we use `react-test-renderer` to add unit tests. I've found using [react-testing-library](https://testing-library.com/docs/react-testing-library/example-intro) to be way more intuitive and simpler to build tests. Obviously not saying we need to add that in for this, but would be happy to add that in in a separate PR if you think it's a good idea.
> 
> EDIT: Tried to do this, and I found it hard to atleast create a reasonable unit test that involves clicking in the void elements. I can do it with cypress, but I don't know if we want to modify examples and make the e2e tests more sophisticated (or they should just be simple tests as they are). I think the react-testing-library should help here. LMK what you think.
> 
> @dylans

Let me think about it a bit and ask a few of the other maintainers for their thoughts. There is https://github.com/mwood23/slate-test-utils which leverages react-testing-library, but there has been reasonable reluctance to stray too far from the approach used in slate core.

heyitsaamir:
> > > Makes sense. We’re just starting to add tests for slate-react so it would be great if you can add a test (I won’t block for that, just want to start building up a decent test suite). Thanks.
> > 
> > 
> > Yup will do. I noticed that we use `react-test-renderer` to add unit tests. I've found using [react-testing-library](https://testing-library.com/docs/react-testing-library/example-intro) to be way more intuitive and simpler to build tests. Obviously not saying we need to add that in for this, but would be happy to add that in in a separate PR if you think it's a good idea.
> > EDIT: Tried to do this, and I found it hard to atleast create a reasonable unit test that involves clicking in the void elements. I can do it with cypress, but I don't know if we want to modify examples and make the e2e tests more sophisticated (or they should just be simple tests as they are). I think the react-testing-library should help here. LMK what you think.
> > @dylans
> 
> Let me think about it a bit and ask a few of the other maintainers for their thoughts. There is https://github.com/mwood23/slate-test-utils which leverages react-testing-library, but there has been reasonable reluctance to stray too far from the approach used in slate core.

Makes sense. It's definitely non-trivial to test the behavior right now given the existing framework in the repo.

ahoisl:
Just to mention: The logic with void elements changed in this PR was already there before my changes for readonly editors in #4727. So I guessed back then that some selection for void elements was indeed intended and just avoided selection when the editor is readonly.

In other words: I think, your PR changes behavior not only for readonly editors, but also for void elements in editable editors. Is this intended?

Doing some git blaming, the initial code comes from PR #3649, fixing issue #3648. For testing, I removed the `isTargetInsideNonReadonlyVoid` locally and selection across void elements seems to work just fine - and also my use cases keep on working without errors. So it might be that the selection issue was fixed by something else and the code involving void elements can be removed altogether.

That said, even with a good set of automated tests I would be hesitating to remove some code I have no clue about, even worse without tests... 🤞

heyitsaamir:
Thank you for looking at this with a keen eye!

> Just to mention: The logic with void elements changed in this PR was already there before my changes for readonly editors in #4727. So I guessed back then that some selection for void elements was indeed intended and decided to just avoided selection when the editor is readonly.

That's true and it makes sense that you should be able to select elements when you're editing them.

> In other words: I think, your PR changes behavior not only for readonly editors, but also for void elements in editable editors. Is this intended?

Hm, I don't think so. If the editor is in readonly mode, that clause will return false and that clause should be ignored, right?

```
hasEditableTarget(editor, anchorNode) &&
          !isTargetInsideReadonlyVoid(editor, anchorNode)
```

If `isTargetInsideReadonlyVoid` is false, then it should have no effect in the above statement, right?

> Doing some git blaming, the initial code comes from PR #3649, fixing issue #3648. For testing, I removed the `isTargetInsideNonReadonlyVoid` locally and selection across void elements seems to work just fine - and also my use cases keep on working without errors. So it might be that the selection issue was fixed by something else and the code involving void elements can be removed altogether.

Hm, I don't see that locally. This is what I see without the `isTargetInsideReadonlyVoid` clause. (This is on safari, but same behavior on FF)

https://user-images.githubusercontent.com/48929123/158307514-659beb3c-8557-4bac-a33d-d5e7eedc2f1d.mov

@ahoisl

ahoisl:
> > In other words: I think, your PR changes behavior not only for readonly editors, but also for void elements in editable editors. Is this intended?
> 
> Hm, I don't think so. If the editor is in readonly mode, that clause will return false and that clause should be ignored, right?
> 
> ```
> hasEditableTarget(editor, anchorNode) &&
>           !isTargetInsideReadonlyVoid(editor, anchorNode)
> ```
> 
> If `isTargetInsideReadonlyVoid` is false (which it should be for all non-readonly editors), then it should have no effect in the above statement, right?

Assume that there is a void element where `hasEditableTarget` returns false. If the editor is readonly, then your change does not make difference. However, if the editor is editable (not readonly), then before your changes the result of `isTargetInsideNonReadonlyVoid` would have been true, consequently the result of the boolean operation (=`anchorNodeSelectable` and `focusNodeSelectable`) would have been true and the `if` clause would have been executed. With your changes, the `if` clause is not executed any more.

I'm just saying that your changes cause a change in behavior also for non-readonly/editable editors, not just for readonly ones and I'm unsure if this is intended.

From your issue description, I thought that your changes should only affect readonly editors.

I guess that your void element has an editable target, causing the first part of the boolean operation to always be true. If you really want your changes to only affect readonly editors, then I think some change like the following (pseudo-code) would be needed, where `isTargetInsideVoid` would not check readonly any more:
`const anchorNodeSelectable = readonly ? hasEditableTarget && !isTargetInsideVoid : hasEditableTarget || isTargetInsideVoid`

> Hm, I don't see that locally. This is what I see without the `isTargetInsideReadonlyVoid` clause. (This is on safari, but same behavior on FF)

I only tried out my use case with image void elements, so your use case may still fail and that's a good reason to keep as much use cases unchanged as possible, I think - we don't know all the use cases the existing code is necessary for 😄

heyitsaamir:
> Assume that there is a void element where `hasEditableTarget` returns false. If the editor is readonly, then your change does not make a difference. However, if the editor is editable (not readonly), then before your changes the result of `isTargetInsideNonReadonlyVoid` would have been true, consequently the result of the boolean operation (=`anchorNodeSelectable` and `focusNodeSelectable`) would have been true and the `if` clause would have been executed. With your changes, the `if` clause is not executed any more.

Ah that's true, but I guess it's unexpected for `hasEditableTarget` to be false when the click target is inside the editor, right? It's very possible that I'm misunderstanding the code :). But you're right, it does change the behavior in that scenario. 



> guess that your void element has an editable target, causing the first part of the boolean operation to always be true. If you really want your changes to only affect readonly editors, then I think some change like the following (pseudo-code) would be needed, where `isTargetInsideVoid` would not check readonly any more:
> `const anchorNodeSelectable = readonly ? hasEditableTarget && !isTargetInsideVoid : hasEditableTarget || isTargetInsideVoid`

I didn't do anything special to make my void element an editable target. Check out [this](https://codesandbox.io/s/slate-reproductions-forked-g486e7?file=/index.js) sandbox for what the element looks like. 

It _is_ curious though why the images element doesn't crash. 🤔 [Here's](https://codesandbox.io/s/slate-reproductions-forked-g486e7?file=/index.js) the image example that doesn't crash

ahoisl:
> EDIT: Aha, there's a `contentEditable={false}` in the div that contains the image. So the first clause returns false. I'm not sure if that's correct though. Thoughts?

Seems to be the correct way with `contentEditable={false}` according to the comment from cmmartin in #4886. If that fixes your error, I would be reluctant to integrate this MR, because of the change for non-readonly editors - but it's not my decision 😄

heyitsaamir:
Closing this because setting `contentEditable` to `false` prevents this issue:
https://github.com/ianstormtaylor/slate/issues/4886#issuecomment-1069699523

Issue #4900, opened 2022-03-20
Slate unmounts unchanged elements on `split_node`, `merge_node`


bryanph:
**Description**
When splitting a node the original split node gets unmounted instead of rerendered. Also on merging a node into a previous node, the previous node also gets unmounted and reinserted.
What happens is:
1. `Transforms.transform` modifies the parent (which is handled by immer so the reference changes)
2. This invalidates `NODE_TO_KEY`, so a new `NODE_TO_KEY` entry is generated for this node.
3. Then in `use-children` the key has changed for this element so the whole node will be unmounted and remounted.

Seems like a pretty major issue

**Sandbox**
Check https://codesandbox.io/s/slate-unmount-element-bug-mow9x?file=/index.js

To reproduce:
1. Put cursor at end of the line
4. press `enter`
5. check console, see that the element got unmounted
6. now press `backspace`
7. See both the merged node (correct) and the node to be merged into (not correct) got unmounted

**Expectation**
The component does not unmount but does rerender perhaps if need be.

**Context**
I debugged a little and this is where it seems to go wrong:
https://github.com/ianstormtaylor/slate/blob/947883435fe5534f566ca37adcebe48a4aaa3b0b/packages/slate-react/src/plugin/with-react.ts#L105-L108

A new key is generated for the unchanged node here which causes the element to remount

**Environment**
- Slate Version: latest
- Browser: Firefox, Chrome

PR #4901, opened 2022-03-20
Fix NODE_TO_KEY correction for split_node and merge_node

bryanph:
**Description**
Fixes https://github.com/ianstormtaylor/slate/issues/4900

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/4900

**Example**
The logic in `with-react` was not resetting the node keys for the right nodes for `split_node` and `merge_node` operations. This fixes that so that on `split_node` the node being split does not get remounted and on `merge_node` the node being merged into also does not get remounted.

**Context**
See https://github.com/ianstormtaylor/slate/issues/4900

**Checks**

### DIFF
--- a/.changeset/funny-humans-whisper.md
+++ b/.changeset/funny-humans-whisper.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fixes a bug where nodes remounted on split_node and merge_node
--- a/packages/slate-react/src/plugin/with-react.ts
+++ b/packages/slate-react/src/plugin/with-react.ts
@@ -60,42 +60,35 @@ export const withReact = <T extends Editor>(editor: T) => {
     }
   }
 
+  // This attempts to reset the NODE_TO_KEY entry to the correct value
+  // as apply() changes the object reference and hence invalidates the NODE_TO_KEY entry
   e.apply = (op: Operation) => {
     const matches: [Path, Key][] = []
 
     switch (op.type) {
       case 'insert_text':
       case 'remove_text':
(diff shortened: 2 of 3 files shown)


dylans (commented):
Thanks! I’d love to see a test for this though I know that’s not easy yet. Please add a changeset when you get a chance.

bryanph:
@dylans yeah a test here would be nice indeed, will see what I can do

nemanja-tosic:
@bryanph don't know about cypress, but here's a react test renderer version that fails against previous version but works with the PR:

```
it('should not unmount', async () => {
  const value = [{ children: [{ text: 'hello' }] }]
  const mounts = jest.fn<void, [Element]>()

  let el: ReactTestRenderer

  act(() => {
    el = create(
      <Slate editor={editor} value={value} onChange={() => {}}>
        <DefaultEditable
          renderElement={({ element, children }) => {
            useEffect(() => mounts(element), [])

            return children
          }}
        />
      </Slate>,
      { createNodeMock }
    )
  })

  // slate updates at next tick, so we need this to be async
  await act(async () =>
    Transforms.splitNodes(editor, { at: { path: [0, 0], offset: 2 } })
  )

  expect(mounts).toHaveBeenCalledTimes(2)

  // TODO: assertion against something specific to elements
})
```

bryanph:
@nemanja-tosic awesome, thanks! I pushed two tests for the particular cases being fixed here based on your example here. Let me know if there's something else I can assert here.

dylans (approved):

Issue #4902, opened 2022-03-21
Introduce Slate Test Utils to Slate-React


mwood23:
**Problem**
There has been discussion in the Slack community to improve the [test coverage](https://slate-js.slack.com/archives/CC58ZGGU9/p1647037393714159) and DX for slate-react.

**Solution**
Introduce [Slate Test Utils](https://www.npmjs.com/package/slate-test-utils) to the repo and write some tests on base editor functionality to reduce the number of regressions and increase confidence in package.

**Alternatives**
For now we've discussed using Jest (already in the package). There's talk of migrating to Vitest at some point in the future.

PR #4903, opened 2022-03-22
Add Slate Test Utils to Slate

mwood23:
**Description**
Introduces [slate-test-utils](https://github.com/mwood23/slate-test-utils) to run integration tests on the `/site/examples` editors. I looked at putting it directly into `slate-react`, but the library works better with true integration tests against running editors.

This PR stems from a conversation in Slack where @dylans and @horacioh wanted to see `slate-test-utils` adopted into Slate to encourage more test coverage.  I wrote a couple tests to get initial feedback.

These are some benefits of testing this way:
1. Tests both Slate and Slate React
2. Any node version, TS, or compiler changes will be tests against
3. It encourages writing tests like users interact with the editor
4. User centric tests like these shouldn't need to be updated with every API change. Only breaking changes should trigger warnings.

If all of the example editors get tests on them I think this will introduce more stability to Slate and help it towards a 1.0.

**Issue**
Fixes: #4902 

**Example**
image

**Context**
This is a library I wrote while working on Slate daily that has helped me wrangle Slate over the years.

**Checks**


packages/slate-hyperscript/src/index.ts
```diff
@@ -3,7 +3,16 @@ import {
   HyperscriptCreators,
   HyperscriptShorthands,
 } from './hyperscript'
-import { createEditor } from './creators'
+import {
```
mwood23, line 6: Exports a bunch of methods from Slate hyperscript that may be used for things like Slate Test Utils hyperscript

site/examples/richtext.snapshot.test.tsx
```diff
@@ -0,0 +1,225 @@
+/** @jsx jsx */
+
+import { assertOutput, buildTestHarness } from 'slate-test-utils'
+import RichTextEditor from './richtext'
+import { jsx } from '../test-utils'
+import { fireEvent } from '@testing-library/dom'
+
+it('snapshot = user triggers bold hotkey and types with a collapsed selection', async () => {
```
mwood23, line 8: This is an example of a snapshot test ran with Slate Test Utils. I'm not a huge fan of snapshot tests in general, but added it for living docs in case folks prefer them!

site/tsconfig.test.json
```diff
@@ -0,0 +1,20 @@
+{
+  "compilerOptions": {
+    "target": "es6",
```
mwood23, line 3: Needs `es6` otherwise the compiler breaks stuff. See for more details https://slate-js.slack.com/archives/C1RH7AXSS/p1647905084771509

site/hyperscript.d.ts
```diff
@@ -0,0 +1,19 @@
+declare namespace JSX {
```
mwood23, line 1: Adding for typesafe hyperscript in the tests!

site/examples/richtext.tsx
```diff
@@ -22,11 +22,20 @@ const HOTKEYS = {
 const LIST_TYPES = ['numbered-list', 'bulleted-list']
 const TEXT_ALIGN_TYPES = ['left', 'center', 'right', 'justify']
 
-const RichTextExample = () => {
-  const [value, setValue] = useState<Descendant[]>(initialValue)
+const RichTextExample = ({
```
mwood23, line 25: Change needed to make editor components test friendly. See https://github.com/mwood23/slate-test-utils#making-your-editor-test-friendly for more details on why

bryanph:
Nice work here, some questions/comments:

* From what I understand, these are run more like a unit test right? So keydown events and stuff are emulated?
* If we go this route we should add your code from `slate-test-utils` directly instead so it's easier to maintain the code

Overall my first impressions are that our efforts would be better spent fixing the cypress tests and perhaps creating a little layer of abstraction there to make them easier to write. Real E2E tests can help us test more subtle differences in browser behavior which some emulated tests will miss. Also introducing yet another testing library adds to the maintenance burden that's already fairly high.

horacioh:
hey @bryanph !

Basically this is the first step towards simplifying the testing setup of slate (more context here: https://slate-js.slack.com/archives/CC58ZGGU9/p1647037393714159)

The main goal is to give Slate users more stable options to test Slate in almost every app.

- Having a Cypress setup for E2E testing is a must
- Having a unit test setup is algo important too. (here's where slate-test-utils can help)
- I also have a PR not pushed that replaces Jest and Mocha with [vitest](https://vitest.dev/), which makes test run faster and with less config/tooling

Like I said, having a proper testing setup for Slate apps is benefitial not only for users, but for the overall library stability.

let us know what you think and how we can move forward with it :)
thanks!

bryanph:
@horacioh :wave: 

I very much agree that we need a better testing setup for slate-react as small changes can have big consequences especially due to the subtly different ways different browsers handle things. There's quite a bunch of custom behavior in slate-react for aligning different browser behaviors. The only way we can test that is running tests in an actual browser environment, which is what cypress does (and there are other good alternatives but I'm not very familiar with them). By running tests in `js-dom` we wouldn't be able to test these kinds of cross-browser behaviors. The browser part would be simulated which doesn't quite catch the cross-browser subtleties. 

So my thoughts here are that if we could run them in a cypress (or other E2E tool) kind of environment then we could test that as well. However we currently obviously don't have a very nice way of writing these cypress tests. But perhaps with some utilities we can make them easier to write. The API @mwood23 created here looks very nice and I wonder if it's possible to make it run in an actual browser environment. If that's the case it would be perfect.

Just my two cents :smile:

nemanja-tosic:
I don't really like using both mocha and jest like it is now, and this PR adds more jest code. I think one test runner (used loosely as there's more to jest than just a runner) should be used across the board. Jest was introduced because of slate-react, but mocha could've been used there as well. Cypress uses [mocha/sinon/chai](https://docs.cypress.io/guides/references/bundled-tools#Mocha), so this way we're forced to learn a couple of runners and mock/assertion libraries. I hoped as part of this PR we'd standardize on a runner at least.

I've always felt that there should be some core behavior that ought to be possible to spec out in a browser-agnostic way, where i feel that slate-test-utils make sense due to jsdom limitations, and then a set of integration tests that run against specific browsers. Maybe we're trying to throw technology at something that could be resolved by changing the design a bit? Encapsulating the behavior that differs across browsers communicates a lot and provides a pretty good target for integration tests. Anyway, when i saw the slate-test-utils conversation, it is the unit tests for this core behavior that i saw value in.

That being said, when talking about using cypress for slate-react unit tests - usually browser tests were slow and not very reliable, so i wonder if anyone knows what would be the time needed to run a large unit test suite in cypress? I know that the react testing library has tools that deal with asynchronicity, does this hold for cypress as well? I can see that cypress has a [component test runner](https://www.cypress.io/blog/2021/04/06/introducing-the-cypress-component-test-runner/) which might fit the bill here, but i'm curious if anyone has practical experience WRT speed and stability and also would it be possible to debug in an IDE? These are the criteria i would use to decide when putting together a testing strategy at least, while testing implementation specific behavior is possible to solve through design if all else fails.

bryanph:
@nemanja-tosic some good points there.

> I don't really like using both mocha and jest like it is now, and this PR adds more jest code. I think one test runner (used loosely as there's more to jest than just a runner) should be used across the board. Jest was introduced because of slate-react, but mocha could've been used there as well. Cypress uses [mocha/sinon/chai](https://docs.cypress.io/guides/references/bundled-tools#Mocha), so this way we're forced to learn a couple of runners and mock/assertion libraries. I hoped as part of this PR we'd standardize on a runner at least.

I agree, let's stick with one. Mocha or Jest, not both.

> I've always felt that there should be some core behavior that ought to be possible to spec out in a browser-agnostic way, where i feel that slate-test-utils make sense due to jsdom limitations, and then a set of integration tests that run against specific browsers. Maybe we're trying to throw technology at something that could be resolved by changing the design a bit? Encapsulating the behavior that differs across browsers communicates a lot and provides a pretty good target for integration tests. Anyway, when i saw the slate-test-utils conversation, it is the unit tests for this core behavior that i saw value in.

There might be a place for both indeed. I think the main problem has been the stability for the E2E tests. Also writing them isn't very nice at the moment. Both could be improved.

> That being said, when talking about using cypress for slate-react unit tests - usually browser tests were slow and not very reliable, so i wonder if anyone knows what would be the time needed to run a large unit test suite in cypress? I know that the react testing library has tools that deal with asynchronicity, does this hold for cypress as well? I can see that cypress has a [component test runner](https://www.cypress.io/blog/2021/04/06/introducing-the-cypress-component-test-runner/) which might fit the bill here, but i'm curious if anyone has practical experience WRT speed and stability and also would it be possible to debug in an IDE? These are the criteria i would use to decide when putting together a testing strategy at least, while testing implementation specific behavior is possible to solve through design if all else fails.

It is possible to parallelize cypress tests, described here https://docs.cypress.io/guides/continuous-integration/github-actions#Parallelization. They got an example app here (https://github.com/cypress-io/cypress-realworld-app) with a lot of tests. Although ironically the CI is failing in that example :laughing: which doesn't really inspire confidence. You can debug though the cypress browser thingy (`yarn cypress open`).

horacioh:
the current setup I have on my app for testing is:
- [vitest](https://vitest.dev/) for everything that does not involve rendering UI (unit test and model-based testing)
- [Cypress Component testing](https://docs.cypress.io/guides/component-testing/introduction) for everything that involves rendering UI (including slate interactions). this convo helped me not depend on testing-library for example.
- we use [Tauri](https://tauri.studio/) on our app, so E2E will involve something related to Tauri. we are not using Cypress for E2E for now (WIP at the moment)
- we are exploring using slate-test-utils for some very specific plugin behavior (compare *function(input)* equals to *output* stuff). We have a very specific tree structure and comparing the editor input/output is quite important. Please let me know if someone's doing this with *just* Cypress (I'm curious!)

having better tooling around Cypress for testing Slate sounds like most of us in this thread prefer over adding other tools?
Making Cypress plugins with a API similar to slate-test-utils maybe?

bryanph:
@horacioh 

> we are exploring using slate-test-utils for some very specific plugin behavior (compare function(input) equals to output stuff). We have a very specific tree structure and comparing the editor input/output is quite important. Please let me know if someone's doing this with just Cypress (I'm curious!)

For these kinds of tests I took the tests from the `slate` packages as a base and modified them to run in jest. Also extended the slate-hyperscript definitions so they support the elements used in our particular editor. These tests are not able to mock things like `keydown`, `beforeinput` events and stuff as this PR makes possible but are just for testing the main data manipulation as in the `slate` package tests.

A simple example:
```  
it("Inserting a text fragment at the start of a block should insert the fragment on the same line", () => {
    const input = (
      <editor>
        <title><text>A title</text></title>
        <paragraph>
          <text>
            <cursor />abc
          </text>
        </paragraph>
      </editor>
    )

    const fragment = (
      <fragment>Some text</fragment>
    )

    const output = (
      <editor>
        <title><text>A title</text></title>
        <paragraph>
          <text>
            Some text<cursor />abc
          </text>
        </paragraph>
      </editor>
    )

    assertRun(input, output, (editor) => {
      editor.insertFragment(fragment)
    })
  })
```

> having better tooling around Cypress for testing Slate sounds like most of us in this thread prefer over adding other tools?
Making Cypress plugins with a API similar to slate-test-utils maybe?

This would be an interesting experiment although I have no experience of how this scales in terms of number of tests. Maybe someone else can weigh in on that.

mwood23:
Thanks for all of the feedback! A lot of the feedback I see here is in regards to end to end versus slate test utils. I see them as complimentary approaches to testing Slate React. E2E tests suffer from the following in my experience:

1. They're slow
2. They're harder to maintain
3. They're flakey
4. Devs don't write/fix them because of a mixture of 1, 2, and 3

This is the genesis for why I went with the JSDOM approach for emulating events over using Cypress. I used Cypress for Slate React first, but the API made it to where tests weren't proactively written. What `slate-test-utils` can do for Slate React is increase confidence that code changes to either `slate` or `slate-react` won't cause regressions while keeping the burden of maintenance lower than an E2E test library like Cypress.

There is a need for E2E tests for Slate React, I'm not disagreeing. I'm advocating for more `slate-test-utils` integration tests and utilizing E2E tests for browser specific and smoke tests. My opinion is more tests yields higher confidence and increased development velocity so that active members in the community will feel empowered to put in the necessary changes to get Slate to a 1.0.

The Cypress component test runner is very interesting and brings together what I'd like to see in browser testing eventually. I have no working experience with it so I can't advocate for it one way or another.

As for jest/mocha I used what was already in the repo and agree it's annoying that there's multiple test runners. I'm not actively developing with Slate anymore so I don't have plans migrating `slate-test-utils` to run with mocha. I opened this at the request of some folks in the community and happy to see it in otherwise we can keep it out and folks can pull it into their projects from NPM.

I'm also open to seeing what it would look like to see `slate-test-utils` adopted into Slate as another package if that's what the community would prefer.

horacioh:
so what are the next actions? how can we help?

I'm in the middle of a testing refactor for my app too, and wanted to take advantage of it and contribute to Slate with the same :)

dylans:
> so what are the next actions? how can we help?
> 
> I'm in the middle of a testing refactor for my app too, and wanted to take advantage of it and contribute to Slate with the same :)

I don't mean to be the bottleneck on this, just thinking through all of the feedback and I don't have a clear answer on how to best proceed.

horacioh:
ok folks!

after reading all the great feedback in this thread and try some testing with Cypress component testing, I believe Cypress can be leverage to have a more unifed way for testing `slate` and `slate-react`.

Cypress Component testing can cover both unit and E2E tests. What we need to figure out is that there will be a need for simpler tests using another runner like mocha/chai/vitest.

I will create a branch where I show how I'm testing editor with Cypress component testing and we can go from there.

what y'all think?
and again, thanks y'all for all the feedback!!

horacioh:
I started moving all tests to Cypress component testing here: https://github.com/horacioh/slate/tree/feat/cypress-ct

You can checkout the tests are identical to the latest integration ones. You can see one example here: https://github.com/horacioh/slate/blob/feat/cypress-ct/site/examples/check-lists.spec.tsx

please share any feedback!!

bryanph:
@horacioh very nice, I assume the tests run quicker cause there's no need to start up the examples server and visit that page?

horacioh:
@bryanph indeed! I still need to setup some stuff for testing inside each package. and I will also push another branch where I migrate all tests to use vitest just for compaarison :)

looking forward to get more feedback!

mwood23:
I've moved on from Slate and going to close this since there hasn't been movement. The repo is still open sourced on my profile if people ever need the source.

Issue #3336, opened 2019-12-17
Firefox: wrong selection with double click on a text node


cvlmtg:
#### Do you want to request a _feature_ or report a _bug_?

bug

#### What's the current behavior?

double click on a word in an unformatted paragraph (i.e. made of just one `text` node) and apply some formatting. then deselect and select it again. even if the browser shows the word as selected, slate has selected the whole paragraph. you can see it because the next style is applied to the whole line and not just the word. on Chrome it seems to work correctly.

![Registrazione schermo 2019-12-17 alle 22 24 42](https://user-images.githubusercontent.com/6401008/71035749-71301980-211c-11ea-98f6-b38fd8c60184.gif)

Slate: 0.55.2
Browser: Firefox
OS: Mac

#### What's the expected behavior?

slate should select the word, not the whole paragraph

sirgru:
Just discovered this issue myself. In Firefox, anchor and cursor do not always return the path of the inner node when double clicking. Instead the paths seems to bleed into the previous/next node, What I've found is sometimes the selection works properly with manual drag.

cvlmtg:
@sirgru yes, that should be fixed by #3374

dylans:
This has a potential fix in #3374, but it was never finished. If someone wants to revisit I suggest starting there. Thanks.

PR #3374, opened 2019-12-21
Firefox: fix wrong selection with double click

cvlmtg:
#### Is this adding or improving a _feature_ or fixing a _bug_?

bugfix

#### What's the new behavior?

on firefox, when double clicking on a word which is a whole node, the selection is adjusted so that an inline style can be applied correctly

#### How does this change work?

on firefox, when the selection isn't collapsed, the range is adjusted so that:
1) if the anchor is on the end edge of a node, it is moved to the start edge of the next node
2) if the focux is on the start edge of a node, it is moved to the end edge of the previous node

this makes the range match what happens on chrome and avoid applying inline styles to unselected text

#### Have you checked that...?

- [x] The new code matches the existing patterns and styles.
- [x] The tests pass with `yarn test`.
- [x] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [x] The relevant examples still work. (Run examples with `yarn start`.)

#### Does this fix any issues or need any specific reviewers?

Fixes: #3336
Reviewers: @ianstormtaylor

PR #4908, opened 2022-03-23
Firefox: fix wrong text highlighting with double-click

Jawell:
**Description**
Firefox had an another response of `Document.getSelection()` method. After double-clicking on a word with some mark, you can't unmark this word.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/3336 https://github.com/ianstormtaylor/slate/pull/3374

**Example**
<details>
  <summary>Old behavior</summary>
Untitled
</details>

<details>
  <summary>New behavior</summary>
Untitled
</details>

**Context**
I also changed `toSlatePoint` interface. I don't see any use and reason of this method with old generics. I tried to make a cypress test, but I can't do it. Cypress doesn't support highlighting text by double-clicking. Maybe this is not true. But I see that cypress moves the text cursor to the end of the word that was double-clicked. And I can say that the test (`cypress/integration/examples/select.ts`) is not correct. Because cursor pointer is seated at the end of a paragraph after triple-click

**Checks**

### DIFF
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -471,10 +471,10 @@ export const ReactEditor = {
     editor: ReactEditor,
     domPoint: DOMPoint,
     options: {
-      exactMatch: T
-      suppressThrow: T
+      exactMatch: boolean
+      suppressThrow: boolean
     }
-  ): T extends true ? Point | null : Point {
+  ): Point | null {
     const { exactMatch, suppressThrow } = options
     const [nearestNode, nearestOffset] = exactMatch
       ? domPoint
@@ -642,16 +642,15 @@ export const ReactEditor = {
       )
     }
 
-    const anchor = ReactEditor.toSlatePoint(
-      editor,
-      [anchorNode, anchorOffset],
-      { exactMatch, suppressThrow }
-    )
+    let anchor = ReactEditor.toSlatePoint(editor, [anchorNode, anchorOffset], {
+      exactMatch,
+      suppressThrow,
+    })
     if (!anchor) {
       return null as T extends true ? Range | null : Range
     }
 
-    const focus = isCollapsed
+    let focus = isCollapsed
       ? anchor
       : ReactEditor.toSlatePoint(editor, [focusNode, focusOffset], {
           exactMatch,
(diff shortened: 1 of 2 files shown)


Jawell:
I see that the problem of integration:
```
  it('hovering toolbar appears', () => {
    cy.get('div')
      .eq(12)
      .should('not.exist')
```

I checked git history. That `.eq(12)` was `.eq(13)`. But I don't understand why it was used this indexes

dylans:
> I see that the problem of integration test in:
> 
> ```
>   it('hovering toolbar appears', () => {
>     cy.get('div')
>       .eq(12)
>       .should('not.exist')
> ```
> 
> I checked git history. That `.eq(12)` was `.eq(13)`. But I don't understand why it was used this indexes

It's just timing out in our CI environment. We need to improve our integration tests, but I don't think it's related to your change.

Jawell:
@dylans 
Hi, so, what about this PR? :)

dylans:
> @dylans Hi, so, what about this PR? :)

I've been traveling so I'm a bit behind on my reviews. I will get to it soon.

Jawell:
Hey, can we merge this?

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -471,10 +471,10 @@ export const ReactEditor = {
     editor: ReactEditor,
     domPoint: DOMPoint,
     options: {
-      exactMatch: T
-      suppressThrow: T
+      exactMatch: boolean
+      suppressThrow: boolean
     }
-  ): T extends true ? Point | null : Point {
+  ): Point | null {
```
BitPhinix, line 477: Not a huge fan of the signature change, it makes sense that it returns a non-nullable value when not using `suppressThrow` (although the current definition is wrong as `exactMatch` should be boolean and not `T` since it doesn't influence the 'nullability' of the return, as far as I can see). The same is the case for `toSlateRange`.

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -661,6 +660,46 @@ export const ReactEditor = {
...
+        const after = Editor.after(editor, anchor as Point)
+        // Editor.after() might return undefined
+        anchor = after || anchor
+      }
+
+      if (isStart) {
+        const before = Editor.before(editor, focus as Point)
+        focus = before || focus
+      }
+    }
```
BitPhinix, line 701: Maybe we should replace this with a `Editor.hangRange` helper

BitPhinix (changes requested):
Just 2 small suggestions, otherwise LGTM. I'm a bit worried that it will potentially break stuff in unexpected places, but in the worst case, we can revert this

dylans (changes requested):
Please see the feedback within the PR.

reza-ebrahimi:
@Jawell @dylans @BitPhinix Any progress on this PR?

dylans:
Superseded by #5580

Issue #4897, opened 2022-03-18
Triple click selection and copy&paste in read-only mode 


irynakoshynska:
**Description**
In Chrome in read-only mode, when we use triple-click selection and try to copy and paste: only one word is copied instead of the entire paragraph

**Recording**
https://user-images.githubusercontent.com/66263028/158982608-bada507e-2249-4b70-9c6c-906f687ee36a.mov
![Screen Recording 2022-03-18 at 10 59 34](https://user-images.githubusercontent.com/66263028/158985047-18f8c57d-0c58-4cef-bb2a-6c6630237fb7.gif)

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/read-only
2. Triple-click on the paragraph
3. Copy & Paste
5. Error: only one word is copied instead of entire paragraph

**Environment**
- Slate Version: 0.72.8
- Operating System:  iOS
- Browser: Chrome

sowes:
I too have encountered this problem! It's preventing our users from quickly copying text in our social feed. +1 for visibility!

Issue #3871, opened 2020-09-10
Triple clicking selects parts of the block below


evasteingrims:
#### Do you want to request a _feature_ or report a _bug_?

A bug

#### What's the current behavior?

Triple clicking for example a paragraph that has a heading below it also selects part of the heading. This can be seen by looking at the rich text menu, the heading format is selected. When copying and pasting the paragraph, an empty heading is also pasted below it.

This can be reproduced in the rich text demo.

![triple-click](https://user-images.githubusercontent.com/11168195/92759227-09db8900-f37f-11ea-8583-a47c6bf13e9e.gif)

Slate: 0.58.4
Browser: Chrome
OS: Mac

#### What's the expected behavior?

Only the block being triple clicked should be selected and copy/pasting it should not paste the empty heading

baijiahei:
maybe you should change focus when triple click!

mbaranovski:
As a workaround I've added this in my project to the Slate Editor component:
```
onMouseDown={(e: MouseEvent): void => {
            if (e.detail === 3) {
              e.preventDefault()
              e.stopPropagation()
            }
          }}
```

cezarsmpio:
This seems to be an issue still. Any news?

e1himself:
Is there a chance this will be fixed with #4908

sennpang:
> Is there a chance this will be fixed with #4908 ?

yes

PR #4914, opened 2022-03-25
Fixed Triple click selection and copy&paste in read-only mode

sennpang:
**Description**
Triple click selection and copy&paste in read-only mode

**Issue**
Fixes: #4897 #3871

**Example**
before:
https://user-images.githubusercontent.com/66263028/158982608-bada507e-2249-4b70-9c6c-906f687ee36a.mov
after:
https://user-images.githubusercontent.com/16861647/160055330-4df1837d-c364-4bc8-96e0-f4a5297c740e.mov

**Context**

**Checks**

### DIFF
--- a/.changeset/khaki-candles-serve.md
+++ b/.changeset/khaki-candles-serve.md
@@ -1,5 +1,5 @@
 ---
-"slate-react": patch
+'slate-react': patch
 "slate": patch
 ---
 
--- a/.changeset/khaki-candles-serve.md
+++ b/.changeset/khaki-candles-serve.md
@@ -1,0 +1,6 @@
+---
+"slate-react": patch
+"slate": patch
+---
+
+Fixed Triple click selection and copy&paste in read-only mode
--- a/.changeset/khaki-candles-serve.md
+++ b/.changeset/khaki-candles-serve.md
@@ -1,6 +1,6 @@
 ---
 'slate-react': patch
-"slate": patch
+'slate': patch
 ---
 
 Fixed Triple click selection and copy&paste in read-only mode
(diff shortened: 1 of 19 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -733,14 +733,25 @@ export const Editable = (props: EditableProps) => {
...
               if (
-                !readOnly &&
                 hasTarget(editor, event.target) &&
                 !isEventHandled(event, attributes.onClick) &&
                 isDOMNode(event.target)
               ) {
                 const node = ReactEditor.toSlateNode(editor, event.target)
                 const path = ReactEditor.findPath(editor, node)
 
+                if (event.detail === 3 && readOnly) {
```
dylans, line 743: In general, I prefer an enum so I don't have to remember what 3 is... even if it's just the line before, e.g. `const TRIPLE_CLICK = 3` and then `if (event.detail === TRIPLE_CLICK ...`

dylans (changes requested):
See my suggestion. Also please add a changeset. Thanks!

packages/slate-react/src/components/editable.tsx
sennpang: ok

.changeset/khaki-candles-serve.md
```diff
@@ -0,0 +1,6 @@
+---
+"slate-react": patch
```
dylans, line 2:
```suggestion
'slate-react': patch
```

.changeset/khaki-candles-serve.md
```diff
@@ -0,0 +1,6 @@
...
+"slate": patch
```
dylans, line 3:
```suggestion
'slate': patch
```

dylans (changes requested):

dylans (approved):

PR #4919, opened 2022-03-29
Restore user selection after beforeinput with target range

BitPhinix:
**Description**
Fixes macOS auto replacement and possibly issues with other IMES by restoring the intended user selection after beforeInput events with a target range.

**Example**
Before:
before

After:
after

**Context**
We currently select the before-input target range thus overwriting the actual intended user selection. The target range of the before-input event isn't meant to be used as the user selection. This pr stores the origin user selection before we select the target range and restores it once we have applied the change if no `set_selection` operation was applied on between.

There are definitely other ways to achieve to solve this, but I went with this approach as it was the simplest to do and shouldn't break any existing plugins. I don't really have a strong opinion about this.

Here is how the current behavior breaks the macOS auto replacement:

Typing "omw " results in the following events:

- Insert text "o" with target range from offset 0 to 0
- Selection change event to offset 1 to 1
- Insert text "m" with target range from offset 1 to 1
- Selection change event to offset 2 to 2
- Insert text "w" with target range from offset 2 to 2
- Selection change event to offset 3 to 3
- Insert text " " with target range from offset 3 to 3
- Selection change event to offset 4 to 4
- Insert text "On my way!" with target range from 0 to 3 (The actual replacement)

In the last step we currently select the target range and don't restore the actual user selection, so it ends up being directly after the "On my way!" before the space, while it should be after the space.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -8,6 +8,7 @@ import {
   Text,
   Transforms,
   Path,
+  RangeRef,
 } from 'slate'
 import getDirection from 'direction'
 import debounce from 'lodash/debounce'
@@ -51,6 +52,7 @@ import {
   IS_FOCUSED,
   PLACEHOLDER_SYMBOL,
   EDITOR_TO_WINDOW,
+  EDITOR_TO_USER_SELECTION,
 } from '../utils/weak-maps'
 
 type DeferredOperation = () => void
@@ -399,7 +401,14 @@ export const Editable = (props: EditableProps) => {
             })
 
             if (!selection || !Range.equals(selection, range)) {
+              const selectionRef =
+                editor.selection && Editor.rangeRef(editor, editor.selection)
+
               Transforms.select(editor, range)
+
+              if (selectionRef) {
+                EDITOR_TO_USER_SELECTION.set(editor, selectionRef)
+              }
             }
           }
         }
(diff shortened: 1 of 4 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -523,6 +532,17 @@ export const Editable = (props: EditableProps) => {
             break
           }
         }
+
+        // Restore the actual user section if nothing manually set it.
+        const toRestore = EDITOR_TO_USER_SELECTION.get(editor)?.unref()
```
bryanph, line 537: Instead of the `EDITOR_TO_USER_SELECTION` map, can you use a regular variable here as it's in the same scope? Or am I missing something

BitPhinix: It's used [here](https://github.com/ianstormtaylor/slate/blob/f564a1c16325e68662b740231b4997bdf0b052c7/packages/slate-react/src/plugin/with-react.ts#L78-L83) to unset the user selection in case of a 'set_selection' operation.

packages/slate-react/src/components/editable.tsx
```diff
@@ -1385,23 +1405,6 @@ const defaultScrollSelectionIntoView = (
   }
 }
 
-/**
```
dylans, line 1388: Was this intentionally removed? I’m not seeing the relevance of this change to the rest of the PR.

BitPhinix: It's dead code, not used anywhere, and not exported. Accidentally committed the removal but I think it's fine to keep it in the pr no?

dylans (approved):

PR #4920, opened 2022-03-31
Run <Slate />'s focus event listeners after <Editable />'s focus handlers in React >= 17

adri1wald:
**Description**
Run `<Slate />`'s focus event listeners after `<Editable />`'s focus handlers in React >= 17

**Issue**
Fixes: #4754

**Example**
See #4754 / #4755 - happy to provide more evidence if necessary

**Context**
In #4754, @jhurwitz noted that "the `useFocused` hook displays old values". I noticed this issue in my project as well, and the fix in #4755 solved the problem for me. However as noted by @bryanph, this change caused a regression when selecting between two different editors. Thus #4755 was reverted in #4874.

By chance, I noticed that running @jhurwitz's [code sandbox](https://codesandbox.io/s/stupefied-moore-qsmjh?file=/src/Editor.js) in slate's examples does not cause the "inverted" focus behaviour, and after some digging I realized that the difference is that the code sandbox uses React 17 whereas slate's examples use React 16. To validate this, I [cloned the sandbox](https://codesandbox.io/s/objective-grass-hcejne?file=/src/Editor.js) and downgraded to React 16.

After some debugging, I noticed that the order of execution of `<Editable />`'s onFocus handler and `<Slate />`'s focus event listener were reversed in React 17. This led me to look at the React changelog, and it became clear that all this comes down to a change in how `onFocus` and `onBlur` are implemented in React 17 (see https://github.com/facebook/react/pull/19186). As I understand it, `onFocus` and `onBlur` now listen to the `focusin` and `focusout` events during the bubbling phase.

Based on the above, I reasoned that listening to the `focusin` and `focusout` events without `useCapture` in `<Slate />` should fix the order of execution. I'm happy to report that this is indeed the case!

For backwards compatibility, I added a React version check, and only use the  `focusin` and `focusout` events if React >= 17. However, I've since tested using the `focusin` and `focusout` events no matter the React version and it doesn't seem to cause any issues (the focus behaviour in React 16 and 17 look identical and all tests still passing). Let me know whether it would be preferable to remove the version check.

Lastly, it would be good to get some input from @bryanph to verify that this doesn't cause the same issues as #4755 for them.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -9,7 +9,7 @@ import {
   SlateSelectorContext,
 } from '../hooks/use-slate-selector'
 import { EDITOR_TO_ON_CHANGE } from '../utils/weak-maps'
-import { IS_REACT_ABOVE_VERSION_17 } from '../utils/environment'
+import { IS_REACT_VERSION_17_OR_ABOVE } from '../utils/environment'
 import { useIsomorphicLayoutEffect } from '../hooks/use-isomorphic-layout-effect'
 
 /**
@@ -71,7 +71,7 @@ export const Slate = (props: {
 
   useIsomorphicLayoutEffect(() => {
     const fn = () => setIsFocused(ReactEditor.isFocused(editor))
-    if (IS_REACT_ABOVE_VERSION_17) {
+    if (IS_REACT_VERSION_17_OR_ABOVE) {
       document.addEventListener('focusin', fn)
       document.addEventListener('focusout', fn)
       return () => {
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -72,6 +72,9 @@ export const Slate = (props: {
   useIsomorphicLayoutEffect(() => {
     const fn = () => setIsFocused(ReactEditor.isFocused(editor))
     if (IS_REACT_VERSION_17_OR_ABOVE) {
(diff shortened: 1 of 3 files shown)


dylans (approved):

dylans (commented):
Please add a changeset and then I'll land this. Thanks @adri1wald!

bryanph:
Don't really have the time to test this thoroughly at the moment, but the reasoning makes sense to me. Nice catch @adri1wald and thanks for the detailed explanation!

bryanph (approved):

packages/slate-react/src/components/slate.tsx
```diff
@@ -70,16 +71,20 @@ export const Slate = (props: {
...
-    return () => document.removeEventListener('focus', fn, true)
-  }, [])
-
-  useIsomorphicLayoutEffect(() => {
-    const fn = () => setIsFocused(ReactEditor.isFocused(editor))
-    document.addEventListener('blur', fn, true)
-    return () => {
-      document.removeEventListener('focus', fn, true)
-      document.removeEventListener('blur', fn, true)
+    if (IS_REACT_VERSION_17_OR_ABOVE) {
```
bryanph, line 74: My only remark would be to add a little summary of your explanation in the PR here just to make it easier to figure out what this code does in the future

adri1wald: Have added a summary of my explanation!

PR #4924, opened 2022-04-03
Documents how to implement a Void Element.

DougReeder:
**Description**
Documents how to implement a Void Element and gives an example.

**Issue**
There are several non-obvious requirements, which are difficult to tease out of the examples.

### DIFF
--- a/docs/api/nodes/element.md
+++ b/docs/api/nodes/element.md
@@ -11,6 +11,7 @@ interface Element {
 - [Behavior Types](element.md#element-behavior-types)
   - [Block vs. Inline](element.md#block-vs-inline)
   - [Void vs Not Void](element.md#void-vs-not-void)
+    - [Rendering Void Elements](element.md#rendering-void-elements)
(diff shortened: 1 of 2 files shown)


docs/api/nodes/element.md
```diff
@@ -30,6 +31,25 @@ A "block" element can only be siblings with other "block" elements. An "inline"
 
 In a not "void" element, Slate handles the rendering of its `children` (e.g. in a paragraph where the `Text` and `Inline` children are rendered by Slate). In a "void" element, the `children` are rendered by the `Element`'s render code.
 
+#### Rendering Void Elements
+
+Void Elements must
```
dylans, line 36: Either remove must here or the start of each list item :)

dylans (commented):
Thanks! One quick suggestion

docs/api/nodes/element.md
DougReeder: Good catch!

dylans (approved):

Issue #4930, opened 2022-04-08
Easy repro crash when triple clicking


scottlu:
**Description**

Very easy to reproduce crash. Simply triple click in an area of Editable not occupied by a child element. See Context below for the lines of code that are causing it.

The bug was introduced very recently, 4/4/22, PR #4914, meant to fix issue #3871

**Recording**

https://user-images.githubusercontent.com/82301/162357857-514f3563-a256-45e2-ab11-21f6c019ab0c.mp4

**Sandbox**

https://codesandbox.io/s/slate-reproductions-forked-j0e6qq

**Steps**
To reproduce the behavior:
1. Open the sandbox
2. Ensure the sandbox is using slate 0.76.1, slate-react 0.76.1, and slate-history 0.66.0
3. Triple click in the Editable's padding area, not on the child element
4. Crash

**Expectation**

Not to crash.

**Environment**
- Slate Version: slate 0.76.1, slate-react 0.76.1, slate-history 0.66.0 
- Operating System: [Debian Bullseye]
- Browser: [Chrome]

**Context**

The problem is the TRIPLE_CLICK code merged on 4/4/22, PR #4914, in slate-react/src/components/editable.tsx, line ~1503. In the case of triple clicking on the Editable component itself, the node returned from toSlateNode is the editor node, and the path returned from findPath is of zero length. The code inside the TRIPLE_CLICK conditional appears to not take into account the the node is the editor, or that the path length may be zero.

benbrandt:
Also seeing this behavior

zijizhu:
Yeah I can also reproduce this. Any potential fix?

JosNun:
Is this fixed by #4944?

BitPhinix:
Yes, this can be closed

PR #4933, opened 2022-04-11
Fixed  #3871 (can reproduction when i click 4 times)

### LINKED ISSUE #3871: Triple clicking selects parts of the block below

sennpang:
**Description**
When i click 4 times, the selection is not expected, like #3871 

https://user-images.githubusercontent.com/16861647/163365630-22f55edf-625c-46ad-b575-c4001edc9288.mp4

fixed

https://user-images.githubusercontent.com/16861647/163365819-7f3e7115-cbcc-4bf2-bb40-ff84e5bb54c3.mp4

**Issue**
Fixes: #4930 #3871 

**Example**
https://user-images.githubusercontent.com/82301/162357857-514f3563-a256-45e2-ab11-21f6c019ab0c.mp4

**Context**
In the case of triple clicking on the Editable component itself, the node returned from toSlateNode is the editor node, and the path returned from findPath is of zero length. 

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -303,7 +303,13 @@ export const Editable = (props: EditableProps) => {
             exactMatch: false,
             suppressThrow: false,
           })
-          Transforms.select(editor, range)
+
+          if (state.isTripleClick) {
+            state.isTripleClick = false
+            Editor.unhangRange(editor, range)
+          } else {
+            Transforms.select(editor, range)
+          }
         }
       }
     }, 100),
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -765,6 +765,7 @@ export const Editable = (props: EditableProps) => {
                   const end = Editor.end(editor, [path[0]])
                   const range = Editor.range(editor, start, end)
                   setTimeout(() => {
+                    if (!ref.current) return
                     Transforms.select(editor, range)
                     Editor.unhangRange(editor, range)
                   }, 100)
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -144,6 +144,7 @@ export const Editable = (props: EditableProps) => {
       hasInsertPrefixInCompositon: false,
       isDraggingInternally: false,
       isUpdatingSelection: false,
+      isTripleClick: false,
       latestElement: null as DOMElement | null,
     }),
     []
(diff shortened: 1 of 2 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -760,11 +760,14 @@ export const Editable = (props: EditableProps) => {
...
-                if (event.detail === TRIPLE_CLICK) {
+                if (event.detail >= TRIPLE_CLICK && path.length) {
                   const start = Editor.start(editor, [path[0]])
                   const end = Editor.end(editor, [path[0]])
                   const range = Editor.range(editor, start, end)
-                  Transforms.select(editor, range)
+                  setTimeout(() => {
+                    Transforms.select(editor, range)
+                    Editor.unhangRange(editor, range)
+                  }, 100)
```
scottlu, line 770: Given the timeout, two potential issues: 1) the range could be invalid by the time the timer expires (editor content could have changed). Calling select() and unhangRange() with an invalid range could cause unexpected behavior including a crash. 2) the component could have unmounted before the timer expires, and these apis should not be called in that case. Making the code synchronous would remove these problems.

#1: I don't know of a fix for this one. It's not clear how you could know the range is valid when the timer fires.
#2: You can check if the component has been unmounted by setting a ref boolean in a useEffect cleanup callback. Then that ref's value can be checked in this setTimeout callback function to see if the component has unmounted. If it has, then don't make these api calls.

scottlu (changes requested):

packages/slate-react/src/components/editable.tsx
sennpang: add `if (!ref.current) return` in setTimeout callback

scottlu: The node, path, and range may not be right after 100ms if editor.children changed within that time. That could be solved by putting all the logic inside the setTimeout callback like this:

```
if (event.detail >= TRIPLE_CLICK) {
  setTimeout(() => {
    if (ref.current) {
      if (path.length) {
        const start = Editor.start(editor, [path[0]])
        const end = Editor.end(editor, [path[0]])
        const range = Editor.range(editor, start, end)
        Transforms.select(editor, range)
        Editor.unhangRange(editor, range)
      }
    }
  }, 100);
  return;
}
```

sennpang: putting all the logic inside the setTimeout callback have another problem, the event.target maybe not current target , it 
will set selection to prev target
`onDOMSelectionChange` function in `packages/slate-react/src/components/editable.tsx`
image
when i change to `Transforms.unhangRange(editor, range)`, triple and more click is fine, but i don't know  the potential issues

sennpang: @scottlu i delete the setTimeout function, and use ` state.isTripleClick ` to check 3 or more clicks, on `onDOMSelectionChange` event to decide how to set the selection, what do you think?

PR #4928, opened 2022-04-06
Slate Testing: refactor testing setup

horacioh:
**Description**
A clear and concise description of what this pull request solves. (Please do not just link to a long issue thread. Instead include a clear description here or your pull request will likely not be reviewed as quickly.)

**Issue**
Fixes: (link to issue)

**Example**
A GIF or video showing the old and new behaviors after this pull request is merged. Or a code sample showing the usage of a new API. (If you don't include this, your pull request will not be reviewed as quickly, because it's much too hard to figure out exactly what is going wrong, and it makes maintenance much harder.)

**Context**
If your change is non-trivial, please include a description of how the new logic works, and why you decided to solve it the way you did. (This is incredibly helpful so that reviewers don't have to guess your intentions based on the code, and without it your pull request will likely not be reviewed as quickly.)

**Checks**
- [x] The new code matches the existing patterns and styles.
- [x] The tests pass with `yarn test`.
- [x] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [x] The relevant examples still work. (Run examples with `yarn start`.)
- [ ] You've [added a changeset](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md) if changing functionality. (Add one with `yarn changeset add`.)

Issue #4378, opened 2021-07-15
Unified compatibility


JonasKruckenberg:
**Problem**
The unified ecosystem provides a vast amount of utility functions for tree traversal, transformation and validation. It also supports easy parsing of html and markdown out of the box. We (and I can image others too) would love to leverage those tools in combination with slate. However both trees are not compatible: Slate expects leaf nodes of the following shape:
```ts
interface Text {
   text: string
}
```
while leaf (literal) nodes defined in [unist](https://github.com/syntax-tree/unist) have the following shape:
```ts
interface Literal {
   type: "text",
   value: string
}
```

**Solution**
As you can see the only incompatibility is the name of the value key `text` vs `value` (And just renaming that across slate does the trick)

**Context**
In [my fork](https://github.com/JonasKruckenberg/slate/tree/fix-text) I have a version of slate that is fully compatible with unist-style trees. I was just wondering if support for this is something you're interested in and if I should therefore put the work into a PR. It would most definitely result in a breaking change so I can understand if you don't.

JonasKruckenberg:
@horacioh FYI

raitucarp:
I upvote this. Unified gives a good ecosystem. Hopefully, in the future, Slate will join the ecosystem.

dylans:
I personally am a fan of this change (not sure how others on the core slate team feel), though it's a breaking change and it would have significant impact on users that persist slate documents.

I can see a couple of paths towards getting this to work, assuming this is a change that the Slate team wants to support:

* Make the key configurable (with a default of text, and some future plan to change that default to value)
* Support text or value as the key and have the code base treat them interchangeably
* Change from text to value and introduce a breaking change at some point in the future (maybe 1.0)

cameronbraid:
I've come across this issue as I have written some code that converts the slate syntax tree into a uinst compatible syntax tree (and back again) to be able to use uinst transform, visit etc in my apps.

Is this proposal being seriously considered by the slate team ?

cameronbraid:
I should also note that its not only the `Text` interface that needs to be changed, as a unist `Node` requires a `type: string` field as well

dylans:
I think if we followed my proposed approach we could make it work, along with your note. Interested in raising a PR?

horacioh:
@JonasKruckenberg wdyt?

JonasKruckenberg:
That's super exciting! I had some issues in the past especially with making the types generic over the key, but that's something we can figure out for sure

[...]

PR #4934, opened 2022-04-11
refactor: use vitest instead of mocha

JonasKruckenberg:
**Description**
After chatting with @horacioh about slate testing I went ahead and converted the unit testing for all 4 packaged from mocha to vitest. Contrary to #4928 I used a script to actually rewrite and group up all the test files to a more "standard" approach using `describe` and `test`. While the speedup gained by vitest isn't significant on it's own (On my m1 macbook pro it was ~3.6s using vitest vs 4.8s using mocha) the live-retesting features offered by vitest make actually writing tests much nicer.
Vitest can detect changes in source-code and automatically re-run only affected tests for example. So during actual development and test writing to comparison is more like 200ms using vitest vs 4.8s using mocha on my machine.

The other advantage vitest brings to the table is testing without building. This means that tests can be run against the source-code, something that vitest actively encourages. This also tilts the scale in vitests favor as CI test runs can forego the build step. (I didn't touch the actual workflow files yet, though)

**Issue**
N/A

**Example**
N/A

**Context**
This PR was born out of prior with #4928 and my need to have variations of existing tests (i.e. multiple tests per file) for #4378.

**Checks**


dylans:
As a quick note, I'm following along but haven't said much yet. Testing is a challenge to get right. My perspective is mostly aligned with the comments in https://github.com/ianstormtaylor/slate/pull/4903#issuecomment-1076373954 and I'm just sort of waiting to see what emerges from the various PRs and trying to avoid my own bias.

TheSpyder (changes requested):
While I appreciate that a lot of work probably went into this, deleting the `index.js` files in favour of repeating boilerplate everywhere doesn't seem like a good change to me. I don't think the existing approach is particularly non-standard, both `describe` and `it` were used before, it's just a choice to use file convention over more detailed configuration.

Likewise, while using aliases to redirect import statements to the source code avoids the seemingly annoying bundle step, what we have was a very deliberate choice too. Mocha can be used this way too with the right configuration; running tests run on the bundled output is just something the original authors (and I) believe is important. With rollup watch mode the slowest part of testing is actually babel, not bundling.

As for your comments about not needing to build in CI, I very strongly disagree. CI should test that everything is ready to release not take shortcuts just to get faster feedback.

joshuawootonn:
Hey @JonasKruckenberg 👋🏼 

I am currently trying to use vitest to test a slate plugin I am creating. I am getting this error: 

CleanShot 2023-03-13 at 13 34 57@2x

whenever any of the actions I am running use `Transforms.XXXX`. I have the same tests working with jest. Did you happen to run into anything like this / do you have any input as to how to resolve it?

dylans:
I think the consensus that emerged has been to use Playwright for E2E tests, and testing-library + mocha for unit tests.

This has mostly meant not using Vite/vitest or Cypress. Closing this out for now in favor of what we've been using in main.

PR #4939, opened 2022-04-12
Fix pointRef leaks caused by not unref'ing

ahoisl:
May cause severe performance degradation if more and more point refs
need to be updated.

**Description**
PointRef instances are not cleaned up in some of the `Transform` functions and can cause performance degradation especially with large documents.

**Issue**
Fixes: no issue filed

**Checks**

### DIFF
--- a/.changeset/khaki-panthers-burn.md
+++ b/.changeset/khaki-panthers-burn.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Fix pointRef leaks caused by not unref'ing
--- a/packages/slate/src/transforms/node.ts
+++ b/packages/slate/src/transforms/node.ts
@@ -12,6 +12,7 @@ import {
   Ancestor,
 } from '..'
 import { NodeMatch, PropsCompare, PropsMerge } from '../interfaces/editor'
+import { PointRef } from '../interfaces/point-ref'
 
 export interface NodeTransforms {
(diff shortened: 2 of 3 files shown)


dylans (commented):
I need some time to wrap my head around this one. Any chance you have a simple test that shows the before and after, or profiles to compare before and after?

ahoisl:
> I need some time to wrap my head around this one. Any chance you have a simple test that shows the before and after, or profiles to compare before and after?

There are just a few lines changed, the diff just looks like more changes because I added a "try/finally", thus indenting all code lines in the "try". If you switch off showing whitespace changes, it should be rather easy to review.

Basically, there are just some places in the Transforms where point refs are not correctly unref-ed. You can easily see this by logging the size of the `Editor.pointRefs` set in between edits of any editor, e.g. when editing and deleting with backspace.

The easiest way to reproduce the performance decrease is:
1. open the "Huge document" example in Chrome.
2. copy & paste the whole document about 4-5 times (document should then be 4-5 times size)
3. just anywhere in the document, press and hold any key

-> the characters will appear laggy, i.e. in chunks and individually as usual

ahoisl:
For the "huge document" example, just editing a bit the size of the pointRefs increases from 0 to 1683 and never decreases again:
image

dylans (approved):

PR #4952, opened 2022-04-19
Fix positions iteration when starting in an inline void node

steve-codaio:
**Description**

The 'positions' iterator code has special logic to always return the 'start' of the first node in a block, when using a content based 'unit' such as 'character'. There's a flag to track whether we've returned that start point or not, in order to skip yielding the start point on subsequent nodes in the block. We also have an early continue on the loop for void nodes, to make sure we only ever yield the the first point of a void.

The bug was that if you start iteration inside a void, that early continue code executes, but does not reset the "I've returned the start point" flag, even though we have

**Example**
If you have 
```
<block>                  // path [0]
  <text text="hello"/>   // path [0, 0]
  <inline-void>          // path [0, 1]
    <text/>              // path [0, 1, 0]
  </inline-void> 
  <text text="world"/>   // path [0, 2]
</block>
```

and start iterating from `{path: [0, 1, 0], offset: 0}` with `unit: 'character'`, before this fix the second point returned will be `{path: [0, 2], offset: 0}` instead of `{path: [0, 2], offset: 1}`

**Checks**

### DIFF
--- a/.changeset/little-plums-behave.md
+++ b/.changeset/little-plums-behave.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Fix positions iteration when starting inside an inline void node
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -1341,6 +1341,9 @@ export const Editor: EditorInterface = {
         // then we will iterate over their content.
         if (!voids && editor.isVoid(node)) {
           yield Editor.start(editor, path)
+          // It's possible the start of the range we're iterating over is in a void, in which case
+          // we want to make sure we don't incorrectly yield the start of a subsequent text node for unit !== 'offset'
+          isNewBlock = false
           continue
         }
 
--- a/packages/slate/test/interfaces/Editor/positions/range/start-in-void.tsx
+++ b/packages/slate/test/interfaces/Editor/positions/range/start-in-void.tsx
@@ -1,0 +1,31 @@
+/** @jsx jsx */
+import { Editor } from 'slate'
+import { jsx } from '../../../..'
+
+export const input = (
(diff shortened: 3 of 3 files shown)


dylans (approved):

dylans:
@steve-codaio this one has introduced a pretty significant regression, specifically the right arrow skips past a void and moves to the next line, and inserting a void causes the cursor to move to a new line. We probably need to revert this unless you have a solution. Thanks.

steve-codaio:
Ah, in our implementation we already had special casing for 'leaving' a void via navigation to not use `unit: character`, since we're not traversing any text. 

The impetus for this fix was consistency. Even before my changes, if were to the left of the inline void and said "move forward 2 characters" you would end up on the next line, just like that bug you linked to. This only 'fixed' the case where you started on the void.

Maybe the fix is to go the other direction with consistency? i.e., with 'unit: character' we always step into and out of a void by one point? The only problem there is that kinda changes the semantics of a void to == 2 characters. 

So my ideas on options are: 
- We could fix the nav code to recognize the void cases and not use 'character' in the cases where the behavior doesn't make sense (i.e., leave the main 'positions' api as is)
- Change positions to _always_ go to the point right after an inline void, whether we start there or not
- Revert this, despite the inconsistency. (I'd be ok with this as our team can always work around it)

Any thoughts on the best option?

dylans:
@steve-codaio I think the approach I would favor would be:

* Revert it first
* Raise PRs for the proposed options and test them (I don't really know the right answer at this point)

dylans:
@steve-codaio let me know if you have time to raise a PR to revert in the next day or two. If not I'll do it over the weekend.

steve-codaio:
https://github.com/ianstormtaylor/slate/pull/5008

PR #4968, opened 2022-04-26
Added types for options and common string literals

### LINKED ISSUE #4966: Export standalone types for all nontrivial function parameters
**Problem**
There are quite a few `.d.ts` function types whose parameters have types that are more complex than just a primitive or union of primitives. For example, this is the start of `NodeTransforms` in `transforms/node.d.ts`:

```ts
export interface NodeTransforms {
    insertNodes: <T extends Node>(editor: Editor, nodes: Node | Node[], options?: {
        at?: Location;
        match?: NodeMatch<T>;
        mode?: 'highest' | 'lowest';
        hanging?: boolean;
        select?: boolean;
        voids?: boolean;
    }) => void;
```

Code that wants to refer to the type of `options` parameter has no easy way to do it. This is rather inconvenient in TypeScript if you have, say, a variable meant to be provided as `options`.

```ts
// What type is this?!
let options = {};

if (someCondition) {
  options.hanging = true;
}
```

**Solution**
Would you accept a PR that goes through and makes interfaces and union types for all these exported complex types, then exports them in the `slate` module?

**Alternatives**

Continuously using this kind of beast in userland:

```ts
import { NodeTransforms } from "slate/dist/transforms/node";

export type InsertNodesOptions = NonNullable<
  Parameters<NodeTransforms['insertNodes']>[2]
>;
```

JoshuaKGoldberg:
**Description**
Adds options types such as `EditorPathOptions` for each of the inline object literal types used as function parameters.

Also adds union types of string literals such as `LeafEdge` for ones commonly reused.

I have low confidence in any names, as this is my first PR in this repository. 🙂 

**Issue**
Fixes #4966

**Example**

Screenshot of an import for 'Options' autocompetion list with many options types from slate

**Checks**
- [x] ~You've [added a changeset](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md) if changing functionality. (Add one with `yarn changeset add`.)~

### DIFF
--- a/packages/slate/src/create-editor.ts
+++ b/packages/slate/src/create-editor.ts
@@ -14,6 +14,7 @@ import {
   Transforms,
 } from './'
 import { DIRTY_PATHS, DIRTY_PATH_KEYS, FLUSHING } from './utils/weak-maps'
+import { TextUnit } from './interfaces/types'
 
 /**
  * Create a new Slate `Editor` object.
@@ -121,7 +122,7 @@ export const createEditor = (): Editor => {
       }
     },
 
-    deleteBackward: (unit: 'character' | 'word' | 'line' | 'block') => {
+    deleteBackward: (unit: TextUnit) => {
       const { selection } = editor
 
       if (selection && Range.isCollapsed(selection)) {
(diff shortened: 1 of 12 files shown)


dylans:
Apologies @JoshuaKGoldberg , I landed this mistakenly so I reverted it. I will review tomorrow... if you could please reraise it with a changeset I'll try to land it on Tuesday or give feedback as appropriate.

BitPhinix:
Thanks for this PR! Incredibly useful 🙏

PR #4972, opened 2022-04-26
Add CodeQL Workflow

luigidimuzio:
### What’s this about
  Hey there,
   I was playing with the CodeQL workflow on some repos, trying to understand how it works.
    I got an alert when running the action on this project that I didn’t fully understand but felt like worth mentioning to you so that you can evaluate it yourself.    
      I think there’s a loose RegEx that could be dangerous if used to sanitise input but I’m not sure if it applies in this context.  
      
Screen Shot 2022-04-26 at 13 13 25
    
           I am sharing my original PR here that should trigger the action and fail with the same error: give it a look then feel free close this PR or keep the CodeQL action in your repo, maybe this was a false flag but the check it’s free and it runs very quick so you could find it useful for the future.  
      
      Cheers,
       Luigi
       
###          What’s in the PR       
       Just the introduction of a CodeQL .yml configuration file with auto-generated default settings. Adding this file as it is would trigger a CodeQL scan step on every push or pull request opened to /main.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,70 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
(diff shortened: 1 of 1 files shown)


dylans (approved):
Thanks I had been meaning to look into this.

PR #4979, opened 2022-04-29
Unset isComposing on keydown with isCompsing false

BitPhinix:
**Description**

 The composition end event isn't fired reliably in all browsers, so we sometimes might end up stuck in a composition state even though we / aren't composing anymore resulting in the selection not being updated:

https://user-images.githubusercontent.com/13185548/166024085-cda35d14-05d7-457a-952e-3637d9cb4fe7.mov

**Checks**

### DIFF
--- a/.changeset/breezy-lizards-impress.md
+++ b/.changeset/breezy-lizards-impress.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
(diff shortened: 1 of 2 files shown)


pzij (commented):
Using native event as a reliable state source, great.

PR #4981, opened 2022-05-02
Add ReactEditor.isComposing(editor)

BitPhinix:
**Description**

Moves the Editable `isComposing` state into the `isComposing` WeakMap used by the AndroidEditable and exposes the current state via `ReactEditor.isComposing(editor)`.

It's quite common to handle some events differently inside a user-defined handler based on the `isComposing` state (e.g. you don't want to trigger markdown shortcuts etc. while the user is composing as that often leads to undesirable behavior).

Currently, you have to keep track of the `isComposing` state on your own using your own, which I don't think is desirable.

**Checks**

### DIFF
--- a/.changeset/odd-cats-tie.md
+++ b/.changeset/odd-cats-tie.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Add `ReactEditor.isComposing(editor)` to get the current `isComposing` state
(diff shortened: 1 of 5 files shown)


.changeset/odd-cats-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'slate-react': patch
```
dylans, line 2: Probably needs to be minor since it is adding to the API.

BitPhinix: Oh yeah, you are right

dylans (approved):

PR #4983, opened 2022-05-04
Add caching to isElement

BitPhinix:
**Description**
Adds caching to the `isElement` method greatly improving performance on repeated `isElement` calls on deeply nested elements using the same approach we currently use for `Editor.isEditor`.

**Context**
`Element.isElement` can be quite slow (2ms+) on deeply nested elements due to it validating that all descendants of the element are valid. We call it repeatedly on the same elements inside the slate codebase and it's a commonly used utility in user-land. For our usage caching the result can shave off 20ms of the keystroke to paint time on longer notes.

This is definitely a trade-off since we make the assumption that elements are updated immutably, even if they are not part of the slate document, but I think that's a fair assumption to make since we currently make it for elements inside the document (and sort of for the editor objects as well).

**Checks**

### DIFF
--- a/.changeset/silver-taxis-drive.md
+++ b/.changeset/silver-taxis-drive.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Add caching to isElement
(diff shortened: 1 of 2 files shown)


SmilinBrian:
For clarity, in case someone comes up with a similar idea in the future, this was closed because `Node.isNodeList()` is already cached. The observed performance boost comes from avoiding the `isPlainObject()` call.

Issue #4959, opened 2022-04-21
Mobile Backspacing bugs


ReedRodgers:
**Description**

Pressing and holding backspace on mobile through the last character remaining in a block will stop the backspace, and the first letter will not be deleted. I need to lift my finger off the backspace key and hold it again to delete the first letter, but backspacing stops again before entering the next block.
Sometimes my keyboard will be removed as I'm backspacing too.
I was reading through the Android support issue and I saw it mentioned that we need an android plugin, but I'm not sure if that's still the case since there are only 3 commits in the android plugin repo from November 2020. If that's the case please let me know and I'll give that a try.

**Recording**
Note that I'm constantly holding down the backspace key in this video. When backspacing stops it is not because I have lifted my finger off the backspace key
https://user-images.githubusercontent.com/30154910/164515307-8b835bbc-0bab-4bb8-9541-ade7896ace98.mp4

**Sandbox**

[https://www.slatejs.org/examples/plaintext](https://www.slatejs.org/examples/plaintext)

**Steps**
To reproduce the behavior:
1. Go to [https://www.slatejs.org/examples/plaintext](https://www.slatejs.org/examples/plaintext) using an Android device
2. Copy and paste the example line a few times, pressing enter before each paste so each new line is a new node.
3. Select the sample document (anywhere will do, but it seems like the effect is most noticeable when blocks are shorter and more varied, and the document is longer)
4. Start holding backspace

**Expectation**
When tapping backspace, the character behind the caret should be deleted, the caret position should be moved forward one character, and the mobile keyboard should still be active.
When holding backspace, it should produce the same behaviour as when we tap backspace but characters should continuously be deleted.

**Environment**
- Slate Version: Whichever is used to power the plaintext example at the time of writing
- Google Pixel 4
- GBoard version: 11.4.07.422524001
- Chrome version: 100.0.4896.127
- Android version: 12
- Android build number: SP2A.220305.012

TrySpace:
I have similar issues on Android, it seems to have to do with fast input, similar issues arise when typing fast; It will skip characters and often revert the cursor back to a previous focus point, often where you started typing; Waiting a few seconds will often avoid it.

I believe it has to do with some artificial delay or debounce. I remember reading this somewhere in the issues here. There's some odd things about the android input that doesn't seem to be universal.

But typing fast does not feel smooth at all, let alone holding backspace.

tilemoon:
Same Problem when using Sogou input predictive typing on android device.

Issue #4954, opened 2022-04-20
Cursor keeps jumping to unexpected position and number is being replaced with unexpected text.


antonbakinowsky:
**Description**
Cursor jumps on wrong position when typing special symbols or numbers together ( **(**, **+**, **7** on the video).

- from 27 sec in the video - when typing **(+7** then it changes to **((+** and cursor is moved to after first bracket. It is difficult to see in the video so i gif where i pause/unpause to caught the issue
![slate-seven-replaced-with-bracket](https://user-images.githubusercontent.com/10063600/164206413-a01ca519-2a2e-4b9d-bf18-92bd964cef4e.gif)
- from 46 sec in the video - after typing **7** i press enter to move on the next line. Cursor have been moved to next line but **7** from prev line was deleted. Also attach gif with slower repro.
![slate-seven-is-deleted-after-moving-to-next-line](https://user-images.githubusercontent.com/10063600/164209266-693c3442-dc0b-4e02-92ac-5b884877cbca.gif)
- from 50 sec typing **hello 7** is chnaged to **hhello** and cursor is placed after first **h**
- from 53 sec manually moving cursor to the end of the line moved cursor back to is prev position

**Recording**

https://user-images.githubusercontent.com/10063600/164204891-b8835aea-c967-4f53-b3ee-1f8f59124589.mp4

**Sandbox**
[https://www.slatejs.org/examples/richtext](https://www.slatejs.org/examples/richtext)

**Steps**
Introduced in the video and gifs above.

**Expectation**
Typing **7** is not replaced with something else. Moving cursor to the end of the line actually moves cursor there instead of returning it back to its prev position.

**Environment**
- Slate Version: from slate.js examples
- Operating System: Android 12
- Browser: Chrome 100.0.4896.127

TrySpace:
I have similar issues with android input, especially when typing quickly.

For example typing Enter will often revert the cursor to the last position. Or sometimes pressing Enter will unfocus the editor/hide the keyboard...

antonbakinowsky:
fixed by #4988

Issue #4861, opened 2022-03-02
Android: Line is deleted when changing paragraph type


alex-fried:
**Description**
Create a new line, entering some characters and then changing the paragraph type removes the line/characters

**Recording**

https://user-images.githubusercontent.com/11149763/156387076-576bd9a9-0d8e-4c23-a4a3-d0da151dbe55.mp4

**Sandbox**
https://www.slatejs.org/examples/richtext

**Steps**

1. Goto https://www.slatejs.org/examples/richtext on an Android device
2. Create a new line on the very top
3. Enter some characters there
4. Scroll to the menu bar and select a paragraph type
5. Line is gone

Works when cursor is first moved away and back again (see video)

**Expectation**
The written text is still there and the paragraph type was applied, (as it is on desktop)

**Environment**
- Slate Version: The one in the example, but also 0.72.8 on my local setup
- Operating System: Android 10
- Browser: Chrome for Android 98.0.4758.101

**Context**

Issue #4853, opened 2022-02-24
ANDROID issue: Editing text in android is not optimal. 


rxxndy:
**Problem**
When users attempt to make an edit within their text post and return to the end of the line to finish their post the cursor automatically moves somewhere else. 

**Solution**
unknown

**Context**
See user screen recording of the issue here: https://twitter.com/constantGMI/status/1496814791280795649?s=20&t=H_cfisec069Y0N35En1xOA

Slapbox:
Where's the evidence they use Slate even?

TrySpace:
@Slapbox I've had exactly these types bug for about a year now (don't know if it occured before) with slate.js, often happens when typing fast or hitting enter, the cursor keeps reverting to aprevious position of the cursor.

I have no evidence that "Twetch" uses slate, but the exact issue occurs with slate on Android, so you can take it as such.

Slapbox:
I think we should close this issue.

PR #4988, opened 2022-05-14
Android input handing rewrite

BitPhinix:
# Description

This PR contains a re-write of the android input handling replacing the mutation reconciliation-based approach with a `beforeInput` based one. It also merges the AndroidEditable and EditableComponent into one as maintaining both adds a lot of overhead and results in the AndroidEditable not receiving all changes (see e.g. <https://github.com/ianstormtaylor/slate/issues/4621> for details)

The main issues with handling input on android compared to our 'normal' desktop handling are that beforeInput events on android aren't cancelable and that trying to modify the DOM with incorrect timing causes the keyboard to behave unexpectedly and 'fight' those changes.

### RestoreDom(-Manager)

The `beforeInput` events not being cancelable causes major issues with react as some changes modify the DOM tree quite heavily and result in the DOM not being in the state react expects it to be, thus crashing on the next render. To get around this we changed the key of the editable component in the current implementation which caused react to re-create and replace the entire DOM node. This is both really bad for performance on longer documents and potentially causes issues for some IMEs.

This PR replaces this approach with a `MutationObserver` based approach which tracks all user input caused mutations inside the editor and reverts them directly before react commits the new editable state to the dom. This ensures the DOM is always in the exact state react expects it to be and saves us from re-mounting the entire editable thus resulting in a big performance gain over the current approach.

### AndroidInputManager

Contains almost the entire input handling logic. We detect all changes based on `beforeInput` events. Since we can't apply them directly as they would cause a DOM modification/selection update which cancels composition and/or causes the IME to fight the dom changes we store the user-intended change either as a `diff` or an `action`.

`diff`s are simple text changes that don't require a re-render to visually match the state we would be in if we would apply those changes to the editor.

`action`s are changes that require us to perform DOM modifications to ensure the user sees a visually matching DOM state to the editor state with the changes applied or changes that result in a DOM state we can't get the slate selection from.

Generally, we apply actions as soon as it's safe to do so without interfering with the IME. This seems to be upon the next `input` event (or after a certain timeout as the input event isn't always fired reliably).

We wait until we apply pending text diffs to the editor state until the user stops composing since applying those diffs would cancel to composition thus interfering with typing compositing-based languages like Japanese, or composition would be canceled anyway (e.g. on re-rendering due to a remote change).

Since diffs are potentially long-lived with this approach we have to take care of transforming them by 'external' editor changes to not be invalid after performing e.g. changes by a remote user. This also comes with a lot of challenges of changes being based on the editor's state after applying those diffs without us being able to ensure that the state actually matches. I won't go into details here but most of this conflict resolution is done in the diff utils.

### Mark placeholders

This PR replaces the current concept of composition insert prefixes with a decoration-based alternative of 'mark placeholders'. This is mainly due to insert prefixes not playing well with collaboration and a way more frequent usage on android. They both solve the issue of making the current editor marks part of the actual DOM to ensure the user sees a visually matching state while composing. On Android, we don't have the luxury of being able to move the selection around on composition start without interfering with the IME so we always render them.

# Fixes:

- https://github.com/ianstormtaylor/slate/issues/5026
- https://github.com/ianstormtaylor/slate/issues/5019 (via schedule flush)
- https://github.com/ianstormtaylor/slate/issues/4994
- https://github.com/ianstormtaylor/slate/issues/4959
- https://github.com/ianstormtaylor/slate/issues/4954
- https://github.com/ianstormtaylor/slate/issues/4861
- https://github.com/ianstormtaylor/slate/issues/4853
- https://github.com/ianstormtaylor/slate/issues/4719
- https://github.com/ianstormtaylor/slate/issues/4693
- https://github.com/ianstormtaylor/slate/issues/4621
- https://github.com/ianstormtaylor/slate/issues/4602
- https://github.com/ianstormtaylor/slate/issues/4532
- https://github.com/ianstormtaylor/slate/issues/4531
- https://github.com/ianstormtaylor/slate/issues/4521
- https://github.com/ianstormtaylor/slate/issues/4400
- https://github.com/ianstormtaylor/slate/issues/4348
- https://github.com/ianstormtaylor/slate/issues/4223
- https://github.com/ianstormtaylor/slate/issues/5047
- https://github.com/ianstormtaylor/slate/issues/5052

and various other undocumented issues

# Remaining issues:

You tell me


BitPhinix:
CI passes locally

BitPhinix:
@vernon99 Thanks for the kind words 🙏. Let me know if you experience any issues when using the changes on this branch - they should be stable(-ish). I'm currently in the process of migrating our mobile editor at [Slite](slite.com) to this and probably will still find a few bugs/issues along the way. Still, this should be way more robust compared to the current implementation.

bryanph:
Thanks for the hard work @BitPhinix! I'll try and wrap my head around this over time, will at least give it a try in my environment.

Ping @thesunny, android head chef😄

thesunny:
@bryanph thanks for the ping.

@BitPhinix Wow! That's a lot of work you put in. Awesome!

I know how painful working on Android can be so thank you so much for this effort. :smiley: 

A few comments which are more just thoughts that I don't think should stop this code from making it into the main branch:

1.  It would be helpful to have a set of manual tests. It's a not as good as automated testing but they are easier to put together which might give us some of the benefits of testing without the work of figuring out how to automate the tests. A few benefits (1) allow others to test for regressions on Android (2) You can ask other people to run a manual test saving on the amount of work you need to do when testing (3) enable testing on other versions of physical devices. I found it was easy when fixing an issue, that it would break an old one, especially if a few months have passed. Being able to follow a set of instructions ("click at the end of the word 'hello', hit enter, press backspace" etc.) made it easier to reliably test functionality. 
2.  Long-term, there would be huge benefits to actual automated testing. I'm not sure if anybody has the bandwidth to set that up but there is this unfortunate pattern that over time, Android breaks.  Most recently, I don't think anybody knows where it started becoming unusable. This is the third major Android fix I'd say with only relatively short periods of time where it was usable. Android in some sort of CI would prevent or at least reduce the likelihood of that happening and your efforts becoming a waste.

Anyways, hope none of these comments take away from the amazing work you're doing!

tilemoon:
Nice PR，this fix many bugs i meet on android device

tilemoon:
SlateReact onChange not emit when typing chinese, any solution?

BitPhinix:
In this PR? Yeah, that's expected. Same as when composing, the changes are batched until the composition ends, so it isn't called on each change. But this PR introduces `ReactEditor.flushPendingChanges` to still allow stuff like markdown shortcuts to work. You can take a look at the markdown example on how to use it. Note that using `ReactEditor.flushPendingChanges` will cancel composition so you should only really trigger it on chars that you are sure you can cancel composition without any major consequences on (like ' ').

tilemoon:
> introduces

But it works fine in slate@0.81.0. so that is a downgrade in this PR.in my case, there is also no "compositionEnd" event in slate@0.81.0, but it works fine.

BitPhinix:
Well, that's why composition doesn't really work with the current android implementation - you can archive the same behavior if you call `ReactEditor.scheduleFlushPendingChanges` on each before input but that will kill support for all non-alphabetic languages and some keyboards

tilemoon:
> In this PR? Yeah, that's expected. Same as when composing, the changes are batched until the composition ends, so it isn't called on each change. But this PR introduces `ReactEditor.scheduleFlushPendingChanges` to still allow stuff like markdown shortcuts to work. You can take a look at the [markdown example](https://github.com/ianstormtaylor/slate/blob/a0779ae1b20f893c0c3b12a62c4414cdbccfa148/site/examples/markdown-shortcuts.tsx) on how to use it. Note that using `ReactEditor.scheduleFlushPendingChanges` will cancel composition so you should only really trigger it on chars that you are sure you can cancel composition without any major consequences on (like ' ').

When i just call `flush()` in `androidInputManager.handleInput()` , it works fine.but i don't know is there some problems.

BitPhinix:
That would work (you can achieve the same thing using `ReactEditor.scheduleFlushPendingChanges` and calling it on very before input) but it will mess with some IMEs and autocomplete. There is no way to avoid this sadly

Slapbox:
Is there a plan to merge this? It looks very promising.

BitPhinix:
Yes, there is! I just want to do some final testing and clean up once we have our mobile editor at slite running on this. It should be by mid-next week 🙏

jasontlouro:
Just checking in since it's Friday @BitPhinix - anything I can do to help?

RaulPROP:
Just so you know, all 3 of these bugs (#5052) are solved.

So, thank you and great work.

jasontlouro:
This is working GREAT so far. Thank you 👏

BitPhinix:
Thanks for reporting! It's caused by Swiftkey not being able to handle the non-contenteditable placeholder element. I fixed this before but must have broken it in the last few commits. Will create a PR with a fix for this (and some other small things) asap 😄

antonbakinowsky:
Example with mentions stopped working.
If i got it right then onChange is not being triggered on every change. But mentions example is relying on this.
https://github.com/ianstormtaylor/slate/blob/main/site/examples/mentions.tsx#L76




https://user-images.githubusercontent.com/10063600/182329273-7c4d7ccd-0395-4006-8367-657d2070e6dd.mp4

BitPhinix:
Yes, that is to be expected as the new implementation mimics how composition is handled on desktop. I'll update the example soon, you can use the pendingDiffs to make it work

kaya-altin:
In the latest version (0.82.1) such error occurs when focus or any key press.

image

kaya-altin:
When I focus or press any key, an error like the one below occurs. (0.82.1)

image

Slapbox:
@kaya-altin does it occur with an earlier version like `0.81.x`? Can you report the first affected version of Slate? And can you share what keyboard you're using and your Android version?

kaya-altin:
There is no problem in slate-react 0.81.0, but this error occurs after updating to slate-react 0.82.1. In addition, the error is not only available on android, but also on pc or mac.

Slapbox:
@BitPhinix thanks again for your great work. I noticed your project LiveBlocks is pinned to `"slate": "^0.82.0"`. I wonder how realistic it is for us to expect these changes to continue to work in newer versions? 

We have to decide between upgrading Slate from `0.47.x` or moving to Lexical, so I'm trying to gather whatever information I can.

PR #4993, opened 2022-05-16
Fix: better decorations handling and memoizing

jasonphillips:
**Description**
Decorations have undergone several changes recently, most notably from https://github.com/ianstormtaylor/slate/pull/4876 (which broke a few things fixed in https://github.com/ianstormtaylor/slate/pull/4910).

Prior to [4876](https://github.com/ianstormtaylor/slate/pull/4876), slate-react used a context to communicate the `decorate` function down the tree, with the purpose of notifying all elements of a change to the decorate function without passing it directly down the tree. However, this didn't work correctly ever since removing the extra component between parent and children, because when `useContext` is called from within another hook (`useChildren` in this case), it will always cause a rerender in the nearest parent component, regardless of the results of that outer hook. There is no way to avoid a changed context causing a rerender in the nearest parent of the `useContext` hook in React--so ever since the children wrapping component was removed, the entire tree would rerender on every `decorate` change even if the resulting decoration ranges themselves are compared and found equal.

In [4876](https://github.com/ianstormtaylor/slate/pull/4876), that was changed so that decorations are computed for the _entire tree_ up front, but there are problems with that PR in retrospect (not just the fix referenced above):

 - it's not clear if stale decorations (reason for that PR) were a real issue, because the test added by the PR was incorrect. [Here](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/test/index.spec.tsx#L82) the decorate function is updated in-place by using `mockImplementation`, but this doesn't change its reference--so the context isn't invalided and of course the nested elements wouldn't update their decorations in that case. Slate's [search highlighting](https://github.com/ianstormtaylor/slate/blob/main/site/examples/search-highlighting.tsx) example shows that anything which wants to trigger redoraction of the entire tree is expected to invalidate `decorate`, not change it in-place.
 - the android component [was not updated to match](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/android/android-editable.tsx#L326) the new approach of generating decorations at the top, so this caused a significant regression on android.
 - the old `useDecorate` context and provider was never cleaned up even though it no longer served any purpose after those changes 🤔 

More imporantly, performance is a major problem with the approach of that PR. Since decorations are generated at the top, every single element down the entire tree now calls `Range.intersection` on every decoration passed to it from above. This can explode in complexity in many cases, particularly if you have any intervening elements. 

For instance, see this example (variation on the code highlighting example), where there is a `code-container` wrapping many `code-line` elements (for the purposes of rendering line numbers). 

- [on slate-react@0.73](https://codesandbox.io/s/slate-react-0-73-decorations-with-depth-87rkr1?file=/src/App.js) it performs perfectly okay despite over 4,000 children of the code container and many more decorations.
- [on current slate-react](https://codesandbox.io/s/slate-react-0-79-decorations-with-depth-i5unon) (**warning, may freeze your browser for a bit**) the performance is basically broken and the editor is not usable. The reason for this is that decorations now have an exponential complexity when there are intervening elements, because (A) all the decorations are generated at the top, then (B) they are all (many thousands) passed down into the code container from above, then (C) that container calls `Range.intersection` on every combination of decoration and child. 

The fix in this PR is restore the context-based generation of decorations down the react tree (so that we don't have to run intersection checks all the way from the top every time), but then add a component that wraps this context and therefore and avoid rerenders if the decorations are in fact unchanged. The performance is tremendously improved this way.

There is also small change included to `Range.intersect` so that it doesn't unnecessarily spread every single range passed in. When the range is a child of the outer range being checked, we can return it directly and avoid a great deal of lost performance due to copying these objects all the way down the tree.   

**Checks**

### DIFF
--- a/.changeset/blue-dots-count.md
+++ b/.changeset/blue-dots-count.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Restore use of decorate context, wrap element and text with component to memoize.
--- a/.changeset/light-schools-switch.md
+++ b/.changeset/light-schools-switch.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Improve performance of Range.intersection by returning original range when possible.
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -600,9 +600,7 @@ export const Editable = (props: EditableProps) => {
     }
   }, [scheduleOnDOMSelectionChange])
 
-  const decorations = [...Node.nodes(editor)].flatMap(([n, p]) =>
-    decorate([n, p])
-  )
+  const decorations = decorate([editor, []])
 
   if (
     placeholder &&
--- a/packages/slate-react/src/components/element.tsx
+++ b/packages/slate-react/src/components/element.tsx
@@ -1,9 +1,10 @@
-import React, { Fragment, useRef } from 'react'
+import React, { Fragment, useRef, useMemo } from 'react'
 import getDirection from 'direction'
-import { Editor, Node, Range, Element as SlateElement } from 'slate'
+import { Editor, Path, Node, Range, Element as SlateElement } from 'slate'
 
 import Text from './text'
 import useChildren from '../hooks/use-children'
+import { useDecorate } from '../hooks/use-decorate'
 import { ReactEditor, useSlateStatic, useReadOnly } from '..'
 import { useIsomorphicLayoutEffect } from '../hooks/use-isomorphic-layout-effect'
 import {
@@ -26,17 +27,21 @@ import { IS_ANDROID } from '../utils/environment'
  * Element.
  */
 
-const Element = (props: {
+export interface ElementProps {
   decorations: Range[]
+  path: Path
   element: SlateElement
   renderElement?: (props: RenderElementProps) => JSX.Element
   renderPlaceholder: (props: RenderPlaceholderProps) => JSX.Element
   renderLeaf?: (props: RenderLeafProps) => JSX.Element
   selection: Range | null
-}) => {
+}
+
+const Element = (props: ElementProps) => {
   const {
     decorations,
     element,
+    path,
     renderElement = (p: RenderElementProps) => <DefaultElement {...p} />,
     renderPlaceholder,
     renderLeaf,
@@ -112,6 +117,7 @@ const Element = (props: {
           isLast={false}
           parent={element}
           text={text}
+          path={path.concat[0]}
         />
       </Tag>
     )
@@ -143,8 +149,23 @@ const Element = (props: {
   return content
 }
 
-const MemoizedElement = React.memo(Element, (prev, next) => {
-  return (
+const MemoizedElementWithDecorations = (props: ElementProps) => {
+  const decorate = useDecorate()
+  const ownDecorations = useMemo(() => decorate([props.element, props.path]), [
+    props.element,
+    props.path,
+    decorate,
+  ])
+  const decorations = useMemo(() => [...props.decorations, ...ownDecorations], [
+    props.decorations,
+    ownDecorations,
+  ])
+  return <MemoizedElement {...props} decorations={decorations} />
+}
+
+const MemoizedElement = React.memo(
+  Element,
+  (prev, next) =>
     prev.element === next.element &&
     prev.renderElement === next.renderElement &&
     prev.renderLeaf === next.renderLeaf &&
@@ -153,8 +174,7 @@ const MemoizedElement = React.memo(Element, (prev, next) => {
       (!!prev.selection &&
         !!next.selection &&
         Range.equals(prev.selection, next.selection)))
-  )
-})
+)
 
 /**
  * The default element renderer.
@@ -171,4 +191,4 @@ export const DefaultElement = (props: RenderElementProps) => {
   )
 }
 
-export default MemoizedElement
+export default MemoizedElementWithDecorations
--- a/packages/slate-react/src/components/text.tsx
+++ b/packages/slate-react/src/components/text.tsx
@@ -1,9 +1,10 @@
-import React, { useRef } from 'react'
-import { Range, Element, Text as SlateText } from 'slate'
+import React, { useRef, useMemo } from 'react'
+import { Range, Path, Element, Text as SlateText } from 'slate'
 
 import Leaf from './leaf'
 import { ReactEditor, useSlateStatic } from '..'
 import { RenderLeafProps, RenderPlaceholderProps } from './editable'
+import { useDecorate } from '../hooks/use-decorate'
 import { useIsomorphicLayoutEffect } from '../hooks/use-isomorphic-layout-effect'
 import {
   NODE_TO_ELEMENT,
@@ -18,14 +19,17 @@ import { IS_ANDROID } from '../utils/environment'
  * Text.
  */
 
-const Text = (props: {
+export interface TextProps {
   decorations: Range[]
   isLast: boolean
   parent: Element
   renderPlaceholder: (props: RenderPlaceholderProps) => JSX.Element
   renderLeaf?: (props: RenderLeafProps) => JSX.Element
   text: SlateText
-}) => {
+  path: Path
+}
+
+const Text = (props: TextProps) => {
   const {
     decorations,
     isLast,
@@ -78,14 +82,28 @@ const Text = (props: {
   )
 }
 
-const MemoizedText = React.memo(Text, (prev, next) => {
-  return (
+const MemoizedTextWithDecorations = (props: TextProps) => {
+  const decorate = useDecorate()
+  const ownDecorations = useMemo(() => decorate([props.text, props.path]), [
+    props.text,
+    props.path,
+    decorate,
+  ])
+  const decorations = useMemo(() => [...props.decorations, ...ownDecorations], [
+    props.decorations,
+    ownDecorations,
+  ])
+  return <MemoizedText {...props} decorations={decorations} />
+}
+
+const MemoizedText = React.memo(
+  Text,
+  (prev, next) =>
     next.parent === prev.parent &&
     next.isLast === prev.isLast &&
     next.renderLeaf === prev.renderLeaf &&
     next.text === prev.text &&
     isDecoratorRangeListEqual(next.decorations, prev.decorations)
-  )
-})
+)
 
-export default MemoizedText
+export default MemoizedTextWithDecorations
--- a/packages/slate-react/src/hooks/use-children.tsx
+++ b/packages/slate-react/src/hooks/use-children.tsx
@@ -48,11 +48,15 @@ const useChildren = (props: {
     const range = Editor.range(editor, p)
     const sel = selection && Range.intersection(range, selection)
 
-    const ds = decorations.reduce<Range[]>((acc, dec) => {
-      const intersection = Range.intersection(dec, range)
-      if (intersection) acc.push(intersection)
-      return acc
-    }, [])
+    const ds: Range[] = []
+
+    for (const dec of decorations) {
+      const d = Range.intersection(dec, range)
+
+      if (d) {
+        ds.push(d)
+      }
+    }
 
     if (Element.isElement(n)) {
       children.push(
@@ -60,6 +64,7 @@ const useChildren = (props: {
           <ElementComponent
             decorations={ds}
             element={n}
+            path={p}
             key={key.id}
             renderElement={renderElement}
             renderPlaceholder={renderPlaceholder}
@@ -78,6 +83,7 @@ const useChildren = (props: {
           renderPlaceholder={renderPlaceholder}
           renderLeaf={renderLeaf}
           text={n}
+          path={p}
         />
       )
     }
--- a/packages/slate-react/test/index.spec.tsx
+++ b/packages/slate-react/test/index.spec.tsx
@@ -1,12 +1,5 @@
 import React from 'react'
-import {
-  createEditor,
-  NodeEntry,
-  Node,
-  Range,
-  Element,
-  Transforms,
-} from 'slate'
+import { createEditor, NodeEntry, Range, Element, Transforms } from 'slate'
 import { create, act, ReactTestRenderer } from 'react-test-renderer'
 import {
   Slate,
@@ -46,20 +39,19 @@ describe('slate-react', () => {
         expect(decorate).toHaveBeenCalledTimes(3)
       })
 
-      it('should rerender the part of the tree that received an updated decoration', () => {
+      it('should rerender the nodes of the tree that received an updated decoration', () => {
         const editor = withReact(createEditor())
 
         const value = [
-          { type: 'block', children: [{ text: '' }] },
-          { type: 'block', children: [{ text: '' }] },
+          { type: 'block', children: [{ text: 'foo' }] },
+          { type: 'block', children: [{ text: 'bar' }] },
         ]
 
-        // initial render does not return
-        const decorate = jest.fn<Range[], [NodeEntry]>(() => [])
-
+        let decorate = jest.fn<Range[], [NodeEntry]>(() => [])
         const renderElement = jest.fn<JSX.Element, [RenderElementProps]>(
           DefaultElement
         )
+        const renderLeaf = jest.fn<JSX.Element, [RenderLeafProps]>(DefaultLeaf)
 
         const onChange = jest.fn<void, []>()
 
@@ -71,6 +63,7 @@ describe('slate-react', () => {
               <DefaultEditable
                 decorate={decorate}
                 renderElement={renderElement}
+                renderLeaf={renderLeaf}
               />
             </Slate>,
             { createNodeMock }
@@ -78,8 +71,12 @@ describe('slate-react', () => {
         })
 
         expect(renderElement).toHaveBeenCalledTimes(2)
+        expect(renderLeaf).toHaveBeenCalledTimes(2)
 
-        decorate.mockImplementation(([node]) => {
+        renderElement.mockClear()
+        renderLeaf.mockClear()
+
+        decorate = jest.fn<Range[], [NodeEntry]>(([node, p]) => {
           if (node !== value[0].children[0]) {
             return []
           }
@@ -87,7 +84,8 @@ describe('slate-react', () => {
           return [
             {
               anchor: { path: [0, 0], offset: 0 },
-              focus: { path: [0, 0], offset: 0 },
+              focus: { path: [0, 0], offset: 3 },
+              highlight: true,
             },
           ]
         })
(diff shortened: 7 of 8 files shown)


jasonphillips:
Looks like the markdown cypress test has been broken for a while

dylans:
@jasonphillips I'm getting an error on your second sandbox:

```
TypeError
_longsql.longSql is undefined
$csb$eval
/src/App.js:142:4

  139 | const initialValue = [
  140 |   {
  141 |     type: "code-container",
> 142 |     children: longSql.split("\n").map((codeLine) => ({
      |    ^
  143 |       type: "code-line",
  144 |       children: [{ text: codeLine }]
  145 |     }))

View compiled
▶ 4 stack frames were collapsed.
$csb$eval
/src/index.js:3

  1 | import ReactDOM from "react-dom";
  2 | 
> 3 | import App from "./App";
  4 | import "./styles.css";
  5 | 
  6 | const rootElement = document.getElementById("root");

View compiled
▶ 10 stack frames were collapsed.
```

jasonphillips:
@dylans Ah thanks, looks like I didn't save one of the files in the sandbox, try again now.

I'll get some key press measurements to add.

dylans:
Thanks, if you have full performance profiles that would be great as we want to make more sense of this and see what you're seeing more clearly.

We'll need a day or two to think through this as I think the issues you've identified are valid, but I'm concerned the proposed change breaks one of the scenarios we fixed that was causing a significant amount of re-renders for us.

jasonphillips:
> See the original discussion at https://github.com/BitPhinix/slate-yjs/discussions/293#discussioncomment-1285673

Thanks, that's helpful for additional context. 

Here's my understanding of what went on in that discussion, if I'm reading and checking the change history correctly:

 - at that time (ever since the wrapping `<Children />` component [was removed](https://github.com/ianstormtaylor/slate/pull/4152)), slate called `useDecorate` inside of `useChildren` without any intervening component. The negative consequence of this was that changing the top-level `decorate` function would _always_ rerender all children, even if you return the same ranges from your decoration. 
 - for that reason, the "official" advice of invalidating the decorate function in order to update exogenous decorations (as shown in the official search-highlighting example) was no longer performant at all--even if you use WeakMaps or anything else--because changing `decorate` now caused way too many rerenders no matter if the decorations returned for a node are identical or not
- that led to people trying to apply exogenous decorations without invalidating the decorate function, which then led to weird attempts to try and invalidate part of the tree with extra decorations
- PR #4876 was submitted in response to that problem, but taking the wrong approach and making performance actually much much worse, be always re-running all decorations and sending them down the tree by checking intersections

(By the way I checked with @BitPhinix and slate-yjs no longer uses decorations for cursors, although that's not relevant to the broader topic.)

But I'm going to write a couple more examples and measurements so we can continue to evaluate

nemanja-tosic:
> it's not clear if stale decorations (reason for that PR) were a real issue, because the test added by the PR was incorrect. [Here](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/test/index.spec.tsx#L82) the decorate function is updated in-place by using mockImplementation, but this doesn't change its reference--so the context isn't invalided and of course the nested elements wouldn't update their decorations in that case. Slate's [search highlighting](https://github.com/ianstormtaylor/slate/blob/main/site/examples/search-highlighting.tsx) example shows that anything which wants to trigger redoraction of the entire tree is expected to invalidate decorate, not change it in-place.

The test was written to represent the use of slate by various libraries, including plate and slate-yjs at the time, which used a stable decoration reference. It was not incorrect, it was supposed to represent a common usage pattern we hoped slate would support as it was causing issues for people.

It is an interesting discussion i think to say that whenever the decorations change we need to swap the decoration function, but i can argue that this is not semantically correct which lead to issues in various libraries in the first place - the function did not change in the clients as the same calculation is used (the function is supposed to represent that calculation in my mind), the only thing that changed was the state.

Forcing the client (a plugin framework like plate for example) to invalidate the decorate function might be problematic, and i believe it is the ease of use by the clients that is important here. In other words, this appears to be the discussion around are decorations supposed to be updated without altering the decorate function or not, and in my mind clients expected to be able to provide a stable decorate function.

It would also help a lot if we added a failing test for the initial implementation, as that way we could prevent performance issues such as these from happening again and profile the current code as right now we can only guess why things are slow - are there too many re-renders, could something be cached...?

As for Android, it was not updated, but the previous code could have been ported to Android as well (along with a long list of things that are currently missing on Android, unfortunately).

jasonphillips:
>  the function did not change in the clients as the same calculation is used (the function is supposed to represent that calculation in my mind), the only thing that changed was the state.

But to make sure we're on the same page... what changed is an _external_ state, correct? If anything _inside_ of slate's value changes, no matter how far down the tree, slate was always correctly updating it and didn't have any problem with stale decorations. 

Originally the way `decorate` was set up (when the `Children` component still existed), it seemed pretty clear in the design that there are only 2 way to invalidate decorations:
 - either a node changes inside slate (then, that part of the tree gets redecorated), or 
 - the function changes (everything gets redecorated). 
 
In the `search-highlight` example, the function changes since it relies on an external state, and in the `code-highlghting` example it works well by triggering naturally when slate nodes changes. 

But it seems that once the decorate became very inefficient (after removal of Children component, when suddenly changing decorate would rerender _all_ nodes no matter what) a new pattern emerged in some libraries, which started to use decorate without changing its signature--but that had some limitations since decorate wasn't necessarily designed that way to invalidate the tree without any change in nodes or function, and some hacks came about (creating decorations up the tree to invalide things intentionally) and then the fix which started generating them at the top. 

> Forcing the client (a plugin framework like plate for example) to invalidate the decorate function might be problematic, and i believe it is the ease of use by the clients that is important here. In other words, this appears to be the discussion around are decorations supposed to be updated without altering the decorate function or not, and in my mind clients expected to be able to provide a stable decorate function.

That's true, and it is a tricky pattern to work with... so it might be worth revisiting! Clients _can_ provide a stable decorate function when the decorations are derived from the changes to slate nodes, of course, like with syntax highlighting--but they only couldn't previously do that if it involves some _external_ state, because only invalidating the function tells slate that it needs to recompute everything. 

Perhaps you're right that it should change to make things easier for clients!  However, it caused a number of regressions for clients that rely on the older behavior. For instance, redecorating all nodes--even when neither the node nor the function changed--can be very expensive on long documents with mixed content. So, we have, say, 2000 paragraphs, and buried in there are some code blocks of various lengths; in the past, there's no reason that typing into one of those paragraphs should cause `decorate()` to be called again on _all_ the code blocks for syntax highlighting--but now that happens, because the old guarantee (stable nodes & stable function means no recalculation on a node) is removed, and documents can be slowed down by unnecessarily redecorating everything on every keypress.  Suddenly it becomes necessary for clients that were using the old behavior to start memoizing their decorations (weakmaps etc) which is a major unexpected change. 

> as right now we can only guess why things are slow - are there too many re-renders, could something be cached...?

Well right now, the main slow down is very clear and easy to demonstrate: decorations are generated at the top and all passed down the tree, but that means they have to be sent only to the nodes that intersect them, and so if you have element levels in between (like the code-highlighting example in my sandbox links) you get something like `number-of-children * number-of-total-decorations` complexity, and then the entire editor can basically freeze. We can't send intersections down the tree in cases where there might be 1000 children of some top-level block and each one has 2 decorations, because you get 1000 * 2000 checks on intersection. 

But I did try one other thing which might be a more amenable compromise, which I could push for comparison. One way to greatly improve the "decorations generated up-front" method is to not just add them all into one big array at the top, but instead associate them in a WeakMap with the node for which they were generated. So if `decorate()` was called on some node down the tree `path = [3,1,1]`, then the element with that path is the only one that gets those results from the weakmap and sends them to its own children by intersection--instead of checking intersections all the way from the top of the document, which can be very bad for any kind of nesting.

BitPhinix:
I'm in favor of the change in this PR. Re-decorating each node inside the editor on each change simply isn't sustainable.

While slate-yjs legacy indeed doesn't do it, I think the 'correct' way of doing cursor decorations would be to decorate the editor, not the text nodes themselves, which then would be updated on each render. If desirable I can adjust the implementation of the legacy version and publish a patch.

I can't really think of a lot of other common use-cases where you would want to redecorate nodes without changing their contents, but I would still be strongly in favor of adding an API to selectively re-render/invalidate nodes which would cover any potential future needs (and would be quite helpful in a lot of other cases as well). But that's outside the scope of this PR.

jasonphillips:
Using the large code-highlighting example sandbox from above, here's the quick comparison of typing 3 characters into a line:

Slate@.73 (before the PR being discussed) [sandbox](https://codesandbox.io/s/slate-react-0-73-decorations-with-depth-87rkr1?file=/src/App.js): 
image
[Profile-20220516T181818.json.zip](https://github.com/ianstormtaylor/slate/files/8703949/Profile-20220516T181818.json.zip)


Current version of Slate [sandbox](https://codesandbox.io/s/slate-react-0-79-decorations-with-depth-i5unon):
image
**performance profile is too large to upload; try the link above and generate one if interested**

Proposed changes here [sandbox](https://codesandbox.io/s/slate-react-proposed-changes-decorations-with-depth-zq7sqb?file=/src/App.js):
image
[Profile-20220516T181709.json.zip](https://github.com/ianstormtaylor/slate/files/8703943/Profile-20220516T181709.json.zip)

The changes in this PR basically just restore the old performance, and _some_ cases may marginally improve over the old behavior due to the improvement of Range.intersect.

nemanja-tosic:
My comment isn't about whether a change is needed, it is whether the decoration function should be stable, and the PR expects it not to be. How performance should be improved is based on the answer to this question. Based on the performance profiles i did, i haven't looked at the ones you provided but i guess they show the same, the problem isn't decorate, the problem is intersect. If we went with calculating all decorations, it would still be doable based on that.

I don't understand how we decorate the editor in order to add a cursor, as that doesn't translate into a leaf being changed as far as i understand how slate-react works. Unless we're talking about rendering something as an absolute position based on the range. I think it is much more complicated than updating decorations, but i can be persuaded otherwise.

If we were to go down this route, i'm curious why we don't use the useDecorate hook (although i'm very much in favor of another flavor of hook) on MemoizedTextWithDecorations only. I'm basing my assumptions on the fact that only Text nodes have the leaf calculation, i'm not sure if i'm correct though. IOW, if we have a context which can invalidate nodes deeply, why do we bother with passing down decorations?

jasonphillips:
> My comment isn't about whether a change is needed, it is whether the decoration function should be stable, and the PR expects it not to be. How performance should be improved is based on the answer to this question. Based on the performance profiles i did, i haven't looked at the ones you provided but i guess they show the same, the problem isn't decorate, the problem is intersect. If we went with calculating all decorations, it would still be doable based on that.

Yeah I think that if we wanted to calculate all the decorations at the top, we could still improve performance by using a mapping from node to decoration, rather than passing them all down. I tried this recently and it improved things, changing Editable.tsx to something roughly like: 

```typescript
for (const entry of Node.nodes(editor)) {
  const decorations = decorate(entry)
  NODE_TO_DECORATIONS.set(entry[0],  decorations)
}
```

Then this WeakMap can be used inside Element, so it would do something like `const ds = NODE_TO_DECORATIONS.get(node)` and concat that with any nodes passed from the parent. That helps a lot because the assumption is that any call to `decorate([node, path])` can only create decorations under that node, not outside it. Instead, in the latest slate code, for some reason _all_ the decorations are sent down from the top instead of only from the node fro which they were generated, which isn't necessary even if you do want to invalidate the whole tree. 

But whether `decorate` should be a stable reference is something that affects clients in both directions. Based on the old behavior and the setup of the examples, no one should expect a redecoration of _every_ node in the tree on every change, which can be a very big step backwards in performance if you're not expecting it (if you're not memoizing your decorate() because you didn't have to before) and really should be labeled as a breaking change. If the change were made to always redecorate, every use of decorations would now need to consider caching. For instance, the search-highlighting example would rerun the search on _all_ nodes on every keypress, which really isn't logical; you'd expect only the changed paths to ask for redecoration.   

> I don't understand how we decorate the editor in order to add a cursor, as that doesn't translate into a leaf being changed as far as i understand how slate-react works. Unless we're talking about rendering something as an absolute position based on the range. I think it is much more complicated than updating decorations, but i can be persuaded otherwise.

If I understand the suggestion correctly, the thing is that when decorating some node `n` in the tree during `decorate([node, path])`, you are always allowed to return decorations for anything under that node. So you might decorate the leaves of an embedded code block when decorate is called on the block itself, rather than when it is actually called on the leaves. That has always been the behavior. 

So when a user wants to generate decorations for things on every change, they can already do right now it if they create those decorations during `decorate(editor, [])`, instead of doing it at the lower levels. This means it's easy to do something like update a cursor on each change--that cursor would have to be passed down the whole tree with intersections, but you would normally do this with things that are "top-level" information like a cursor, so it makes sense. This style of usage respects the tree of the document.

> If we were to go down this route, i'm curious why we don't use the useDecorate hook (although i'm very much in favor of another flavor of hook) on MemoizedTextWithDecorations only. I'm basing my assumptions on the fact that only Text nodes have the leaf calculation, i'm not sure if i'm correct though. IOW, if we have a context which can invalidate nodes deeply, why do we bother with passing down decorations?

Well you can have decorations that span across text nodes though, yes? Or, ones that need to be generated based on the element above those.

BitPhinix:
> but releasing this would break Plate

Can you share what exactly would break when merging this change so we can better understand the use-case?

dylans:
> > but releasing this would break Plate
> 
> Can you share what exactly would break when merging this change so we can better understand the use-case?

IIRC the use cases where it is problematic are search-highlight, mention, etc.

But I think what was happening was that invalidating the decorator was invalidating all plugins and thus leading to a pretty big performance penalty. I could be misremembering and I'm sure @nemanja-tosic or @zbeyens can correct me if I'm wrong.

nemanja-tosic:
Plate doesn't update decorate unless plugins are invalidated so highlight wasn't working. Invalidating plugins used to lead to performance issues where everything was updated.

jasonphillips:
It does seem like the specific use case of "I want to recalculate all decorations for the entire document on every render" was already achievable, if plate did something like this in its [own decorate wrapper](https://github.com/udecode/plate/blob/d0ad6d7ed9d40062e6a3292a8a1a602e47826dcf/packages/core/src/utils/pipeDecorate.ts):

```typescript
export const pipeDecorate = <V extends Value>(
  editor: PlateEditor<V>,
  decorateProp?: TEditableProps<V>['decorate']
): TEditableProps<V>['decorate'] => {
  const decorates = editor.plugins.flatMap(
    (plugin) => plugin.decorate?.(editor, plugin) ?? []
  );
  if (decorateProp) {
    decorates.push(decorateProp);
  }
  if (!decorates.length) return;

  return (entry) => {
    // only run at the top
    if (entry[0] !== editor) return []

    let ranges: Range[] = [];

    const addRanges = (newRanges?: Range[]) => {
      if (newRanges?.length) ranges = [...ranges, ...newRanges];
    };

   // walk tree
    for (const node of Node.nodes(editor)) {
      decorates.forEach((decorate) => {
        addRanges(decorate(entry));
      });
    }

    return ranges;
  };
};
```

Of course, that always could have been implemented in user-space by plate plugins if they did the same thing and knew to decorate at the top in cases where they actually desire to recalculate everything on every render, rather than only recalculating the changed portions of the tree.

nemanja-tosic:
The change was not a regression for us and probably a lot of people given that updating the decorate function would rerender the entire tree.

The initial comment:

> So it still seems that the change was unambiguously a regression overall, even ignoring the way its expedited review / merge for such a significant change broke all decorations for a little whilea regression overall. I do think that we can improve further and incorporate solid ideas presented here but some sort of restoration is needed first.

is unwarranted in my mind, as we had a patch in place on our end for a while and given the performance boost the patch was contributed to slate. Nothing was expedited, there was no feedback on the change, nor tests to work against, nor anything.

jasonphillips:
Ah apologies, by "expedited" I mainly mean that, in retrospect, none of the perfectly reasonable scrutinity now being applied to this reversion (providing performance metrics, expectation of sandbox examples, explanations of use, comparisons of old & new behavior, detailed discussion of how the tree is applied on decorate, etc) was applied to the original PR, even though it had major implications. So it seems like a detailed discussion learning from all approaches is great now, but in the meantime I don't see any good argument for not completely reverting that change before we go forward, now that it is shown to be so significant.

dylans:
> Ah apologies, by "expedited" I mainly mean that, in retrospect, none of the perfectly reasonable scrutinity now being applied to this reversion (providing performance metrics, expectation of sandbox examples, explanations of use, comparisons of old & new behavior, detailed discussion of how the tree is applied on decorate, etc) was applied to the original PR, even though it had major implications. So it seems like a detailed discussion learning from all approaches is great now, but in the meantime I don't see any good argument for not completely reverting that change before we go forward, now that it is shown to be so significant.

I’m not sure how that helps. In general we try to keep things easy where possible, and slow down when it’s clear that something is not simple.

Rushing to revert a change that’s been live for 2 months doesn’t seem prudent after a day of discussions.

Let’s get this right rather than forcing it.

jasonphillips:
I pushed a more efficient approach here: https://github.com/ianstormtaylor/slate/pull/4997

> I'm not really deep into this part of the code right now but I did work a little on a decorations rework a while back here

At a glance, that looks like the Annotations that slate used to have. At one point, the concepts of Decoration and Annotation were distinct, where decorations are derived from the node and re-run whenever that particular node changes, whereas annotations were for adding some maintained range ref imperatively from the outside, identified by a key for lookups / replace / delete. There are definitely benefits to having annotations. I think we should finalize an implementation of decorations first but real annotations in slate-react would bring a great benefit.

bryanph:
I'm also in favor of reverting the previous decorations change for now as the performance implications right now are very big. For example the code highlighting example re-decorates all nodes on every selection change now as well.

In the longer term I think we need a solution that can do both "decorations" and "annotations" as they have historically been called in slate. I think the distinction between the two is not necessarily warranted and we can find a solution that has good performance for both use-cases.

dylans:
After reviewing both PRs, listening to feedback, and thinking more about this, I'm going to reverse my opinion and agree with the earlier suggestion to revert #4876, separately come up with a fix for Plate (perhaps what was suggested earlier), and then we'll have time to come up with something better.

dylans:
ok, so I think next steps on this would be to revert #4876 and possibly #4910 and then forge a path forward based on the feedback.

@jasonphillips if you have time please raise a PR for this, otherwise I'll get to it by the end of the weekend. 

@zbeyens in the short term this probably means doing something like https://github.com/ianstormtaylor/slate/pull/4993#issuecomment-1129118864 and @BitPhinix perhaps a fix is needed for slate-yjs 3.x?

BitPhinix:
Sure thing, I'll publish a new legacy slate-yjs release in the next few days updating the cursor decoration approach 👍 

But I really recommend switching to the rewrite as its better in almost every regard

jasonphillips:
No reason to leave this ticket open -- I have a separate [performance PR](https://github.com/ianstormtaylor/slate/pull/4997) related to decorations which makes no changes to the existing API or behavior, and then there is the open task of somehow better accomodating "annotations"-like entites if we want to pursue that separately.

Issue #4995, opened 2022-05-17
`scrollSelectionIntoView` doesn't allow to completely customize auto-scrolling behavior


e1himself:
**Description**

There is the `scrollSelectionIntoView` property on the `Editable` component that allows customizing the auto-scrolling behavior (introduced in #4037).

The problem is that it's usage is currently limited, as it's only called for specific use-cases, not being triggered on cursor jumping between paragraph text lines. This becomes an issue when one wants to tweak the auto-scrolling behavior, taking into account a sticky navigation bar, for example.

**Recording**

https://user-images.githubusercontent.com/370680/168827772-648b0317-a5a9-4dae-8f82-edfcd9851da1.mp4

**Sandbox**

https://codesandbox.io/s/slate-reproductions-forked-78jv4t?file=/index.js

**Steps**
To reproduce the behavior:
1. Go to https://78jv4t.csb.app/
2. Downsize your browser window so there's a scrollbar
3. Scroll content down, so the first lines of text are hidden below the sticky header 
4. Click on any visible line of text
5. Press <kbd>↑</kbd> on the keyboard multiple times, until the cursor disappears under the sticky header

**Expectation**
It is expected that the custom `scrollSelectionIntoView` function is called and the page content is being scrolled to keep the cursor visible. But it doesn't, because `Editable` does not invoke `scrollSelectionIntoView` for text-block selection changes.

**Environment**
- Slate Version: 0.77.2
- Operating System: Fedora Linux 35
- Browser: Firefox 101

pdchouse:
toolbar on my project is sticky too. do you have any improvement?

PR #4996, opened 2022-05-17
Call `scrollSelectionIntoView()` on every cursor position change

e1himself:
**Description**
Trigger `scrollSelectionIntoView` when cursor moves between text lines.

**Issue**
Fixes: #4995

**Example**

Both videos are recorded with this example codebase: https://codesandbox.io/s/slate-reproductions-forked-78jv4t?file=/index.js

With the patch applied, the passed custom `scrollSelectionIntoView` function is triggered on cursor position changes within the paragraph content, allowing to tweak auto-scrolling behavior during paragraph text keyboard navigation.

Before (custom `scrollSelectionIntoView` is not triggered, browser default behavior is in action):

https://user-images.githubusercontent.com/370680/168827772-648b0317-a5a9-4dae-8f82-edfcd9851da1.mp4

After (custom `scrollSelectionIntoView` is triggered, it takes into account the sticky header offset):

https://user-images.githubusercontent.com/370680/168835027-bcc09f4f-0323-47e9-a986-0dde3f565e06.mp4

**Checks**

### DIFF
--- a/.changeset/sixty-bears-march.md
+++ b/.changeset/sixty-bears-march.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Call `scrollSelectionIntoView` on every cursor position change
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -206,7 +206,9 @@ export const Editable = (props: EditableProps) => {
         try {
           const domRange = ReactEditor.toDOMRange(editor, selection)
           scrollSelectionIntoView(editor, domRange)
-        } catch {}
+        } catch {
+          // do nothing
+        }
 
         return
       }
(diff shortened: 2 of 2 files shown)


dylans (approved):

BitPhinix:
`scrollSelectionIntoView` causes a reflow so this might have unexpected performance implications, especially on long documents

BitPhinix:
Just confirmed it - it seems to add about 10ms to every keystroke when typing at the bottom of larger docs and is definitely noticeable.

Screenshot 2022-05-17 at 18 16 45

I don't think we should take over the default browser behavior here. If desired the `scrollIntoView` call can be added very easily in user-land.

dylans (changes requested):
Per the performance note from @BitPhinix we probably shouldn't land this as-is.

e1himself:
@BitPhinix @dylans 

Thank you for looking into this. I appreciate the quick response :+1: 

I'll do more research on it. But meanwhile, I'm closing this PR. Thanks!

Issue #3490, opened 2020-02-10
Cannot input chinese directly after the mention in mention demo


zhaoyao91:
#### Do you want to request a _feature_ or report a _bug_?

Bug

#### What's the current behavior?

![Kapture 2020-02-10 at 12 41 17](https://user-images.githubusercontent.com/3808838/74121701-982e3780-4c03-11ea-9521-7cbc3328c6a9.gif)

https://www.slatejs.org/examples/mentions

Slate: I don't know
Mac OS: 10.15.3
Chrome: 80.0.3987.87

#### What's the expected behavior?

When typing Chinese after the mention (void node), it could work fine, but it seems breaking the slate structure and the following typing is buggy.

tangzijun:
在safari中使用，输入中文后，光标始终在段落的最前面，这个问题十分困扰，不知道现在解决的进展如何了，我使用的是0.57.1的最新版本。

tangzijun:
It is used in safari. After entering Chinese, the cursor is always at the front of the paragraph. This problem is very disturbing. I don't know how the problem is solved. I use the latest version of 0.57.1.

pedily:
I can confirm similar issues with Microsoft IME whenever using IME input at offset 0 behind another element.

The IME widget will modify the wrong dom element.

My workaround is to insert a character without IME first, then use IME input, then deleting the character again

pedily:
I can confirm similar issues with Microsoft IME whenever using IME input at offset 0 behind another element.

The IME widget will modify the wrong dom element.

My workaround is to insert a character without IME first, then use IME input, then deleting the character again

pedily:
I can confirm similar issues with Microsoft IME whenever using IME input at offset 0 behind another element.

The IME widget will modify the wrong dom element.

My workaround is to insert a character without IME first, then use IME input, then deleting the character again

scil:
it's about IME. Chinsese can not be used well on slate right now. because IME related DOM event damage React Reconciliation.

#3824

hardywu:
confirm that it still appears on safari right now

xxxxxccc:
version 0.65 still has this problem, input chinese after a void node; use macos ，when u 'ArrowLeft' or 'ArrowRight' choose a chinese word ,the cursor incorrect

jim-kk-hc:
+1

dylans:
[...]

PR #5013, opened 2022-05-29
fix(mentions): correct spacer tag position

pubuzhixing8:
correct spacer tag position to avoid IME input being interrupted

**Issue**
Fixes:  #3490


dylans (changes requested):
This looks fine, but is failing a lint check.

dylans (changes requested):
Needs a changeset and a fix to pass linting and then we can accept it. Thanks @pubuzhixing8

pubuzhixing8:
ok

Issue #4884, opened 2022-03-11
Issue with Thai language on deleteBackward


tithanayut:
**Description**
A Thai word is typically composes from base consonant and vowels/tone marks above base consonant. Taking ปี่ for example, ป is base consonant and the other two are vowels/tone marks.

When deleting backward, **vowel/tone mark should be deleted one by one**. See the recording of Notepad on Windows for example.

https://user-images.githubusercontent.com/30551284/157884121-2c24f25b-0b56-4f4a-a041-14d211dbdc74.mp4

Slate, however, delete base consonant and vowels/tone marks above it altogether. See the recording below.

https://user-images.githubusercontent.com/30551284/157884550-5633baa6-cf77-4a70-9ead-a27c795d5199.mp4

Note that current Slate behavior is expected for deleteForward (where deleting base consonant should delete vowels/tone marks above it as well), but not for deleteBackward.

**Sandbox**
The demo page of Slate [www.slatejs.org/examples/richtext](https://www.slatejs.org/examples/richtext) exhibits this problem in action.

**Steps**
To reproduce the bug:
1. Go to [www.slatejs.org/examples/richtext](https://www.slatejs.org/examples/richtext)
2. Type in Thai words with above vowels/tone marks. For example, ปี่ปี่ปี่.
3. Press backspace to deleteBackward.
4. See the aforementioned bug.

**Expectation**
When a Thai word is deleting from backward, **vowel/tone mark should be deleted one by one**. See the first recoding on Notepad for example.

**Environment**
- Slate Version: 0.74.2
- Operating System: Windows, macOS
- Browser: Chrome, Firefox

tithanayut:
I believe this could have something to do with string utils function in https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/utils/string.ts

However, I cannot determine the cause yet.

tithanayut:
It seems that AndroidEditable doesn't have this problem while DefaultEditable is.

https://github.com/ianstormtaylor/slate/blob/2a620dc20d808942dc0b1866cbdaf58418156c46/packages/slate-react/src/index.ts#L7-L7

tithanayut:
I found that this issue has began since PR #4671.

It fixes the jump of cursor (to Thai language as well!). However, I believe the logic of moving cursor backward and deleting character is different. The former should ignore vowels/tone marks (currently works perfectly). The latter, however, should delete them one by one.

I will try if I can come up with a fix.

[...]

PR #5015, opened 2022-05-31
Fix deleteBackward behavior for Thai script

tithanayut:
**Description**
When Thai characters are deleting from backward, each code point should be considered as 1 unit for deletion regardless of their grapheme cluster. For example, deleting พี่ from backward should delete the last code point, leaving พี instead of deleting all entirely. 

**This PR fixes deleteBackward behavior for Thai script where deleting N character(s) backward should delete N code point(s) instead of an entire grapheme cluster.**

This change should not affect other languages as the regular expression `/[\u0E00-\u0E7F]+/` checks for the range of Thai script before performing any operation.

**Issue**
Fixes: #4884 

**Example**
Before (unexpected behavior):

https://user-images.githubusercontent.com/30551284/171138953-e72f8c20-17d9-466b-afa1-64080a5a433e.mp4

After (expected):

https://user-images.githubusercontent.com/30551284/171138969-1c6e1a53-b3a6-48f1-8390-566139c0b26d.mp4

**Checks**

### DIFF
--- a/.changeset/twelve-elephants-burn.md
+++ b/.changeset/twelve-elephants-burn.md
@@ -1,0 +1,6 @@
+---
+'slate': patch
+---
+
+Fix deleteBackward behavior for Thai script where deleting N character(s) backward should delete
+N code point(s) instead of an entire grapheme cluster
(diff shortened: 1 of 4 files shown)


dylans (approved):
This looks reasonable to me. Leaving it open a bit longer in case others have feedback.

PR #5017, opened 2022-06-04
Fix DOM selection sync and expose a `useSlateSelection` hook

bryanph:
**Description**
This change fixes syncing between a slate selection and the DOM selection in some cases when a lot of rerenders happen. Since the `useEffect` that's modified in this change was run on every rerender, it could sometimes be triggered when the slate selection didn't actually change which can have some unexpected side effects. This fixes that by using the newly added `useSlateSelection` to be able to add the selection as a dependency to this effect so it does not retrigger on every rerender of the component.

Besides that this change also introduces a new `useSlateSelection` convenience hook for getting the selection whenever it changes.

Could also rename `useSlateSelection` to `useSelection` if that makes more sense.

Also curious if this misses some subtle behavior as the `useIsomorphicEffect` function that's being modified here is kind of hard to parse for me

**Context**
This also fixes the problem that I alluded to in https://github.com/ianstormtaylor/slate/pull/4755#issuecomment-1021101415 which in hindsight probably was a side effect of the behaviour I described above here.

**Checks**


mlajtos:
This looks like it could solve some focus/blur-related issues I am seeing in Firefox. Any info on progress pls?

bryanph:
@mlajtos there were still some problems where the selection sync effect needed to be called but wasn't with this new code. I didn't put a whole lot of time into understanding why that was the case yet. This is one of the more brittle parts of the code base so it could be a bunch of things. Hopefully in the future I can look into this further 🤞 , I won't pursue this for now though.

mlajtos:
> @mlajtos there were still some problems where the selection sync effect needed to be called but wasn't with this new code. I didn't put a whole lot of time into understanding why that was the case yet. This is one of the more brittle parts of the code base so it could be a bunch of things. Hopefully in the future I can look into this further 🤞 , I won't pursue this for now though.

Thank you for the info, @bryanph ! :)

bryanph:
Opened https://github.com/ianstormtaylor/slate/pull/5041 for some of these changes, will get back to the main issue here at some point 🤞

Issue #3457, opened 2020-01-22
Pasted HTML content from Google Docs in wrong order and cursor is not at end


thallada:
#### Do you want to request a _feature_ or report a _bug_?

Report a bug.

#### What's the current behavior?

I have a very basic Google doc: https://docs.google.com/document/d/161uHSpY9kLytIf_izWSeTvVQ5xL5zAZo9D5HbQAxehY/edit?usp=sharing

<img width="227" alt="Screen Shot 2020-01-22 at 1 56 00 PM" src="https://user-images.githubusercontent.com/1505923/72924470-f95a9d80-3d1e-11ea-8557-1485bb180706.png">

When I select the text in the google doc there is a little bit of selectable whitespace at the end of the last sentence:

<img width="258" alt="Screen Shot 2020-01-22 at 1 22 08 PM" src="https://user-images.githubusercontent.com/1505923/72922161-8f3ff980-3d1a-11ea-9290-428bf550f944.png">

If I copy that and then paste it into the slate paste HTML example (https://www.slatejs.org/examples/paste-html), this is the result:

<img width="752" alt="Screen Shot 2020-01-22 at 1 22 43 PM" src="https://user-images.githubusercontent.com/1505923/72922180-97983480-3d1a-11ea-8264-65f3ae536e2e.png">

Note the extra space after the first paragraph and how the cursor ends up after the first paragraph after pasting instead of at the end of the entire pasted content.

If I exclude the extra whitespace at the end of the Google doc:

<img width="199" alt="Screen Shot 2020-01-22 at 1 22 23 PM" src="https://user-images.githubusercontent.com/1505923/72922196-9cf57f00-3d1a-11ea-824e-4ca9e0a00a6e.png">

And paste that in the example, even stranger things happen:

<img width="747" alt="Screen Shot 2020-01-22 at 1 23 05 PM" src="https://user-images.githubusercontent.com/1505923/72922211-a4b52380-3d1a-11ea-9c8a-f75fae878820.png">

The document was pasted out of order. Now the last paragraph is above the ordered list.

For that last example, the link seems to be an important. If I remove the link, pasting with or without the the extra whitespace leaves the paragraphs in the correct order but with the extra newlines and the cursor in the wrong place:

<img width="746" alt="Screen Shot 2020-01-22 at 1 48 18 PM" src="https://user-images.githubusercontent.com/1505923/72923883-e1cee500-3d1d-11ea-827e-e110c5a2f23b.png">

[...]

PR #5018, opened 2022-06-05
Fix mixed inline/block insertFragment.

ulion:
**Description**
When insert mixed fragments (inline1 + block + inline2) at an empty block end, the insert result is: inline1 + inline2 + block

**Issue**
Fixes: #3457

**Example**
insertFragment([{"text":"inline1\n"},{"type":"ul","children":[{"type":"li","children":[{"text":"block li1 \n"}]},{"type":"li","children":[{"text":"block li2 \n"}]},{"type":"li","children":[{"text":"block li3 \n"}]}]},{"text":"inline2\n"},{"text":"inline3\n"},{"text":"inline4"}])

should result:
> inline1
> - block li1 
> - block li2 
> - block li3 
> 
> inline2
> inline3
> inline4

but result:
> inline1
> inline2
> inline3
> inline4
> - block li1 
> - block li2 
> - block li3 

NOTICE: to get the correct input of the insertFragment, it depends on the deserializer used by the slate instance. so no directly online reproduce here, but if you call the insertFragment with the initial fragments input, the bug is just there.

**Context**
The reason of this bug is: middleRef is manually set to next blockPath when insert point is at blockEnd, but the endRef is not consider this case (it only calculate the next inline path and no split happens) so result the wrong ordered result.

**Checks**

### DIFF
--- a/.changeset/pretty-plums-grow.md
+++ b/.changeset/pretty-plums-grow.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Fix for insertFragment text/inline + block mixed fragments.
--- a/packages/slate/src/transforms/text.ts
+++ b/packages/slate/src/transforms/text.ts
@@ -372,15 +372,13 @@ export const TextTransforms: TextTransforms = {
 
       const middleRef = Editor.pathRef(
         editor,
-        isBlockEnd ? Path.next(blockPath) : blockPath
+        isBlockEnd && !ends.length ? Path.next(blockPath) : blockPath
       )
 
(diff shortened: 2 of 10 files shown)


dylans (changes requested):
Needs a changeset and tests are failing currently.

ulion:
> Needs a changeset and tests are failing currently.

changeset and tests are added.

dylans (approved):
Sorry this took so long, I've been traveling a lot this summer. This looks great and is definitely something I've found us needing.

PR #5020, opened 2022-06-07
Replace useState with useMemo in the "Installing Slate" walkthrough

zhouzi:
This PR replaces the `useState`s in the "Installing Slate" walkthrough with `useMemo`s to reflect what's recommended in the rest of the documentation.

### DIFF
--- a/docs/walkthroughs/01-installing-slate.md
+++ b/docs/walkthroughs/01-installing-slate.md
@@ -51,11 +51,11 @@ Of course we haven't rendered anything, so you won't see any changes.
 
 ```typescript
 // TypeScript users only add this code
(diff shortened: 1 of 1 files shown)


dylans (approved):

cameronbraid:
I think the move should be from useMemo to useState in these cases.. as react docs say that it may forget memoized values - probably not something you want to happen

https://reactjs.org/docs/hooks-reference.html#usememo

> You may rely on useMemo as a performance optimization, **not as a semantic guarantee**. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.

PR #5022, opened 2022-06-08
Replace useMemo with useState in the docs

zhouzi:
The documentation was previously using a mix of `useState` and `useMemo` for storing the editor. 90% of the times it was using `useMemo`.

As pointed by @cameronbraid in https://github.com/ianstormtaylor/slate/pull/5020#issuecomment-1149450318, React may forget about a previously memoized value and recalculate the useMemo.

So this PR suggests replacing all occurrences of `useMemo` to store the `editor` with `useState`.

### DIFF
--- a/docs/concepts/09-rendering.md
+++ b/docs/concepts/09-rendering.md
@@ -42,7 +42,7 @@ const MyEditor = () => {
 You don't have to use simple HTML elements, you can use your own custom React components too:
 
 ```javascript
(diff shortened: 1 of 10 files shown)


dylans (approved):

dylans (changes requested):
This change is fine, though it's failing the linter check.

dylans (approved):

PR #5039, opened 2022-06-30
Unhang a range even if the start point of the range is in the middle of a node

davisg123:
**Description**
We noticed that a range will only unhang if the start and end both have an offset of zero. This seems incorrect, we would expect the end range to be corrected even if the start point is within a node.

**Example**
```
Text <anchor>line 1
<focus>Text line 2
```
We would expect this range to be considered hanging and it should be corrected using `unhangRange`. The fix is to avoid an early return caused by an incorrect `start.offset` check

**Checks**

### DIFF
--- a/packages/slate/test/interfaces/Editor/unhangRange/block-hanging.tsx
+++ b/packages/slate/test/interfaces/Editor/unhangRange/block-hanging.tsx
@@ -5,8 +5,9 @@ import { jsx } from '../../..'
 export const input = (
   <editor>
     <block>
+      wo
       <anchor />
-      word
+      rd
     </block>
     <block>
       <focus />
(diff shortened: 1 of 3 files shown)


packages/slate/test/interfaces/Editor/unhangRange/block-hanging.tsx
```diff
@@ -5,8 +5,9 @@ import { jsx } from '../../..'
 export const input = (
   <editor>
     <block>
+      wo
```
dylans, line 8: Seems safer to keep the old test and add a new test? (so that there's a test for the start point at the start or in the middle?

davisg123:
This broke too many tests

PR #5041, opened 2022-07-06
Change how Slate context updates and introduce `useSlateSelection` hook

bryanph:
**Description**
This changes how `SlateContext` gets updated. This used to be done by replacing the context with a fresh array every time. This change uses an incrementing counter instead. This makes the code easier to read and has the added benefit of using the version counter to perform some logic such as preventing rerenders. I was drafting a solution to make DOM selection syncing more reliable in https://github.com/ianstormtaylor/slate/pull/5017 using this capability but it needs more work which is why I'm opening this PR instead for now.

This also introduces a `useSlateSelection` hook which you can use to get the latest slate selection (only triggers when it's actually updated).

**Checks**


dylans (changes requested):
I like this. It would be good to hear from anyone else before we land it. Also thinking this should be minor rather than patch?

bryanph:
@dylans I was thinking no breaking stuff, so just a patch. But adding the hooks might make it a minor?

dylans:
> @dylans I was thinking no breaking stuff, so just a patch. But adding the hooks might make it a minor?

Exactly, generally when new things are added it should be minor imo. Thanks!

dylans (approved):

PR #5043, opened 2022-07-07
Use `start-server-and-test` for cypress tests

bryanph:
**Description**
This is recommended in both next and cypress docs. Might help a little with cypress tests reliability

https://nextjs.org/docs/testing#running-your-cypress-tests
https://docs.cypress.io/guides/continuous-integration/introduction#Boot-your-server

**Checks**

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "test": "yarn run test:mocha && yarn run test:jest",
     "test:custom": "mocha --require ./config/babel/register.cjs ./packages/slate/test/index.js",
     "test:inspect": "yarn test --inspect-brk",
-    "test:integration": "run-p -r serve cypress:run",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
     "test": "yarn run test:mocha && yarn run test:jest",
     "test:custom": "mocha --require ./config/babel/register.cjs ./packages/slate/test/index.js",
     "test:inspect": "yarn test --inspect-brk",
-    "test:integration": "run-p -r serve cypress:run",
+    "test:integration": "start-server-and-test serve http://localhost:3000 cypress:run",
```
dylans, line 34: Is there a way to do this without hardcoding the URL and port number in here?

dylans (changes requested):

PR #5045, opened 2022-07-11
Don't native insert in elements with white-space="pre" containing tab chars

BitPhinix:
**Description**
Workaround for a obscure bug in chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1219139

**Example**
Behavior when adding whitespace="pre" styling to the code highlighting example and pasting tab chars:

Before:
before

After:
after

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -380,15 +380,36 @@ export const Editable = (props: EditableProps) => {
           const [node, offset] = ReactEditor.toDOMPoint(editor, anchor)
           const anchorNode = node.parentElement?.closest('a')
 
-          if (anchorNode && ReactEditor.hasDOMNode(editor, anchorNode)) {
-            const { document } = ReactEditor.getWindow(editor)
+          const window = ReactEditor.getWindow(editor)
 
+          if (
+            native &&
+            anchorNode &&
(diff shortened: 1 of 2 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -380,18 +380,39 @@ export const Editable = (props: EditableProps) => {
...
               native = false
             }
           }
+
+          // Chrome has issues with the presence of tab characters inside elements with whiteSpace = 'pre'
+          // causing abnormal insert behavior: https://bugs.chromium.org/p/chromium/issues/detail?id=1219139
+          if (
+            native &&
+            node.parentElement &&
+            window.getComputedStyle(node.parentElement).whiteSpace === 'pre'
```
dylans, line 405: Probably unnecessary, but for some reason I think I’d feel better with:

```suggestion
            window?.getComputedStyle(node.parentElement)?.whiteSpace === 'pre'
```

BitPhinix: I don't think this will make any difference indeed but it doesn't hurt either so 🤷‍♂️

BitPhinix:
Failing test is unrelated 👀

dylans (approved):

PR #5061, opened 2022-07-24
fix: update the useIsomorphicLayoutEffect dependencies in android-edi…

suilang:
**Description**
In android-editoable component, the useIsomorphicLayoutEffect's dependencies is incorrect. 

if the initial value of readOnly is true, the function of onDOMBeforeInput  will always invalid， even if the value of the variable later changes to false

**Checks**

### DIFF
--- a/.changeset/angry-files-kick.md
+++ b/.changeset/angry-files-kick.md
@@ -1,7 +1,0 @@
----
-'slate-react': minor
----
-
-- Introduces a `useSlateSelection` hook that triggers whenever the selection changes.
-- This also changes the implementation of SlateContext to use an incrementing value instead of an array replace to trigger updates
-- Introduces a `useSlateWithV` hook that includes the version counter which can be used to prevent re-renders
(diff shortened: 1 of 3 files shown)


dylans:
@suilang We appreciate the PR. That said, we have a substantial set of Android fixes landing shortly in #4988. Could you please verify that this fix has been made in that PR?

suilang:
@dylans Someone has already found this problem and fixed it in that PR, I will close this PR.  Also, due to the instability of the current Android editor, I import the DefaultEditable directly. I found that it worked well on higher versions of Android, but I didn't look into the differences. I wonder to know when the new fixes will be merged into the main branch

dylans:
> @dylans Someone has already found this problem and fixed it in that PR, I will close this PR. Also, due to the instability of the current Android editor, I import the DefaultEditable directly. I found that it worked well on higher versions of Android, but I didn't look into the differences. I wonder to know when the new fixes will be merged into the main branch

Thanks. To answer your last question, today!

PR #5071, opened 2022-08-03
Fix toolbar and mentions examples on android

BitPhinix:
**Description**
Updates the mention and hovering toolbar example so that they properly work on android.

This PR also contains a potentially breaking change: It memos the Editable component to make it easier to work with on android since renders cause a flush which will mess with the IME input.

I'm aware that there is still an issue when using swiftkey in the mention example. Swiftkey will duplicate the text inside the mention when pressing space directly after a mention component. This is (one of many) bugs with swiftkey itself and I haven't yet found a way to prevent it (also happens in a non-slate contenteditable, and every other editor out there that is based upon it)

**Checks**

### DIFF
--- a/.changeset/silent-carrots-pretend.md
+++ b/.changeset/silent-carrots-pretend.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Memo Editable component by default to make it easier to work with on android
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -127,7 +127,7 @@ export type EditableProps = {
  * Editable.
  */
 
-export const Editable = (props: EditableProps) => {
+const EditableComponent = (props: EditableProps) => {
   const {
     autoFocus,
     decorate = defaultDecorate,
@@ -142,6 +142,7 @@ export const Editable = (props: EditableProps) => {
     as: Component = 'div',
     ...attributes
   } = props
+
   const editor = useSlate()
   // Rerender editor when composition status changed
   const [isComposing, setIsComposing] = useState(false)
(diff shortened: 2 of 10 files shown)


BitPhinix:
@dylans, I've "refactored" the pending diff handling a bit; it should now be nicer to work with. Not a perfect implementation currently, but I'll clean it up if you like the approach here. We could use the same API for desktop composition, although another rabbit hole I haven't looked into yet.

bryanph:
@BitPhinix could you explain a little how the `shouldFlushPendingChanges` works and how it should be modified to make android work in different use cases? 🙏

BitPhinix:
Sure, there are mainly two use-cases for it:

1. Cancel composition to perform some action on the pending editor state while the user is composing (mainly relevant for auto-replace shortcuts like the markdown shortcut example)
2. Update some component/state to reflect the pending state while the user is composing (like in the mention example)

It's invoked with a "save" timing after each composition change, so you could modify the editor state inside it without interfering too much with the IME (I say too much because it still will if the render takes long/you type very fast). Forcing a flush manually by returning true will have these issues as well, so you should use it very sparingly and only where you have no other choice (like with markdown shortcuts).

We never had an API to deal with/update the editor state while the user is composing since the use-case is rare on desktop. We currently just insert the text on composition end. It's needed on android because composition is very long-lived (you can write an entire paragraph on android inside a single composition session).

BitPhinix:
Any objections/comments on this, or should I create a polished version of this PR?

dylans:
> Any objections/comments on this, or should I create a polished version of this PR?

My instinct is that I wish this didn't feel so complex, but I don't have a better suggestion. Let's go for the polished PR. Thanks!

piscessss:
I had the same problem.When will the content of this branch be officially released?Thanks.

Sunshine168:
Any update?

dylans:
We've had a few smaller PRs to fix parts of this. Closing in favor of a hopefully smaller PR later to address any remaining issues.

PR #5076, opened 2022-08-09
modify unhangRange logic

suilang:
**Description**
As shown in the test case，if there has two blocks, and each block only has one inline void element,  like <block><text/><inline void><text/></inline/><text/></block/>.  The anchor in first text of first line, the focus in the last text of last line.

When delete selected content, the function of unhangRange will change current range to the last text of the first line.

**Checks**

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -1630,7 +1630,7 @@ export const Editor: EditorInterface = {
 
     for (const [node, path] of Editor.nodes(editor, {
       at: before,
-      match: Text.isText,
+      match: n => !Editor.isBlock(editor, n) && !Editor.isEditor(n),
       reverse: true,
       voids,
     })) {
@@ -1639,8 +1639,8 @@ export const Editor: EditorInterface = {
         continue
       }
 
-      if (node.text !== '' || Path.isBefore(path, blockPath)) {
-        end = { path, offset: node.text.length }
+      if (!Text.isText(node) || Path.isBefore(path, blockPath)) {
+        end = Editor.end(editor, path)
         break
       }
     }
--- a/.changeset/angry-files-kick.md
+++ b/.changeset/angry-files-kick.md
@@ -1,0 +1,7 @@
+---
+'slate-react': minor
+---
+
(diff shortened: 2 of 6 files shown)


suilang:
There's a problem. I want to optimize this logic a little bit more, but several tests are going to fail anyway. But the current version is really error-free

dylans (approved):

packages/slate/src/interfaces/editor.ts
```diff
@@ -1630,7 +1630,7 @@ export const Editor: EditorInterface = {
 
     for (const [node, path] of Editor.nodes(editor, {
       at: before,
-      match: Text.isText,
+      match: n => !Editor.isBlock(editor, n) && !Editor.isEditor(n),
```
dylans, line 1633: Might be clearer if it was specific in what it is including, e.g.

Text.isText || Editor.isVoid || Editor.isInline ?

packages/slate/src/interfaces/editor.ts
```diff
@@ -1639,8 +1639,8 @@ export const Editor: EditorInterface = {
         continue
       }
 
-      if (node.text !== '' || Path.isBefore(path, blockPath)) {
-        end = { path, offset: node.text.length }
+      if (!Text.isText(node) || Path.isBefore(path, blockPath)) {
```
nemanja-tosic, line 1642: Question related to the comment on line 1633 - what are we targeting here? An inclusive list might be better. Alternatively, we should extract a variable that describes the conditional.

suilang: The conditions in the match function correspond to the following judgments. You're right, maybe we should revisit what does this function really want. Please do not deal with this PR for the time being. I will deal with the logic of this func again later if I have time

dylans: Meaning you would like us to land the PR and then improve this, or meaning just wait until you have more time?

I think it's a pretty important fix given some of the recent regressions, so I'll take a look soon if you don't have time.

suilang: I'm really sorry that I didn't have time to deal with these due to some unexpected things. I will close this PR first, and you can handle it first

PR #5084, opened 2022-08-17
Fix selection handling with slow flush in mark placeholders on android, fix auto-capitalize after placeholder

BitPhinix:
**Description**
Fixes 2 issues with the current android implementation:

1. Currently, writing in an empty editor with a visible placeholder won't auto-capitalize the first letter due to the IME picking up the text inside the non-contenteditable placeholder element. The most robust fix I found for this is to artificially add a line break to the end of it so the IME thinks the placeholder element is in the line above. Shoudn't have any visual impact on the rendered placeholder itself.

2. The current way we force a swiftkey update after inserting the first word after a mark placeholder to get around the incorrect insert offset causes issues if the editor flush takes a long time and/or you are typing fast because it causes the IME to drop the typed characters. I've experimented a lot around to find a more robust way to solve to issue and ended up with the implementation that seems to work robustly across all IMEs and manually corrects the insert offsets.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -1624,7 +1624,7 @@ export const DefaultPlaceholder = ({
   children,
 }: RenderPlaceholderProps) => (
   // COMPAT: Artificially add a line-break to the end on the placeholder element
-  // to prevent android IMEs to pick up it's content in autocorrect and to auto-capitalize the first letter
+  // to prevent Android IMEs to pick up its content in autocorrect and to auto-capitalize the first letter
   <span {...attributes}>
     {children}
     {IS_ANDROID && <br />}
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -222,7 +222,7 @@ export const Editable = (props: EditableProps) => {
           if (range) {
             if (
               !ReactEditor.isComposing(editor) &&
-              !androidInputManager?.hasPendingDiffs() &&
+              !androidInputManager?.hasPendingChanges() &&
               !androidInputManager?.isFlushing()
             ) {
               Transforms.select(editor, range)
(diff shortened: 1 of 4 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -1616,7 +1622,14 @@ export type RenderPlaceholderProps = {
 export const DefaultPlaceholder = ({
   attributes,
   children,
-}: RenderPlaceholderProps) => <span {...attributes}>{children}</span>
+}: RenderPlaceholderProps) => (
+  // COMPAT: Artificially add a line-break to the end on the placeholder element
+  // to prevent android IMEs to pick up it's content in autocorrect and to auto-capitalize the first letter
```
dylans, line 1627:
```suggestion
  // to prevent Android IMEs to pick up its content in autocorrect and to auto-capitalize the first letter
```

packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
```diff
@@ -537,41 +554,79 @@ export function createAndroidInputManager({
...
             end: end.offset,
             text,
-          })
+          }
+
+          // COMPAT: Swiftkey has a weird bug where the target range of the 2nd word
+          // inserted after a mark placeholder is inserted with a anchor offset off by 1.
+          // So writing 'some text' will result in 'some ttext'. Luckily all 'normal' insert
+          // text events are fired with the correct target ranges, only the final 'insertComposition'
+          // isn't, so we can just ignore it if we re 100% certain it's the incorrect insert.
```
dylans, line 569:
```suggestion
          // isn't, so we can just ignore it if we are 100% certain it's the incorrect insert.
```

dylans (approved):
Minor suggestions for a couple of the comments.

BitPhinix:
I need grammarly for comments inside vscode. Maybe a future side-project 😅

PR #5090, opened 2022-08-19
Add normalizeSelection api to editor

yanmao-cc:
**Description**
In the case of selecting some cells on the table, I want any node's change operation to be applied separately using the selection in each cell

table selection effect on example
image

I'm trying to implement rich text without using the contenteditable property(https://github.com/editablejs/editable).
This is the result after selecting the table

image

In fact, this does not change the selection, cell3 is still selected, but I want to reorganize the selection by adding the normalizeSelection api

**Example**

**Context**
Traversing the cells that need to be changed in the table plugin
```
const { normalizeSelection } = editor
editor.normalizeSelection = (fn) => {
  if (isTable) {
   const { selection } = editor
   for(const cell of cells) {
     fn(...cell selection)
   }
  // restore selection
  Transforms.select(editor,  selection)
  } else {
    normalizeSelection(fn)
  }
}
```
Call the normalizeSelection method within withoutNormalizing and modify the editor's selection property
```
withoutNormalizing(editor: Editor, fn: () => void): void {
    const value = Editor.isNormalizing(editor)
    Editor.setNormalizing(editor, false)
    try {
      editor.normalizeSelection(selection => {
        if (editor.selection !== selection) editor.selection = selection
        fn()
      })
    } finally {
      Editor.setNormalizing(editor, value)
    }
    Editor.normalize(editor)
  }
```

If you agree with this approach, I will improve the follow-up work and documentation

**Checks**

### DIFF
--- a/.changeset/long-buttons-refuse.md
+++ b/.changeset/long-buttons-refuse.md
@@ -1,0 +1,5 @@
+---
+'slate': minor
+---
+
+Added `normalizeSelection` API to split selection before applying node update
--- a/docs/api/nodes/editor.md
+++ b/docs/api/nodes/editor.md
@@ -13,6 +13,7 @@ interface Editor {
   isInline: (element: Element) => boolean
   isVoid: (element: Element) => boolean
   normalizeNode: (entry: NodeEntry) => void
+  normalizeSelection: (fn: (selection: Selection) => void) => void
   onChange: () => void
 
   // Overrideable core actions.
@@ -412,6 +413,10 @@ Check if a value is a void `Element` object.
 ### Normalize method
 
 #### `normalizeNode(entry: NodeEntry) => void`
+
+[Normalize](../../concepts/11-normalizing.md) a Node according to the schema.
+
+#### `normalizeSelection: (fn: (selection: Selection) => void) => void`
 
 [Normalize](../../concepts/11-normalizing.md) a Node according to the schema.
 
--- a/docs/concepts/07-editor.md
+++ b/docs/concepts/07-editor.md
@@ -13,6 +13,7 @@ interface Editor {
   isInline: (element: Element) => boolean
   isVoid: (element: Element) => boolean
   normalizeNode: (entry: NodeEntry) => void
+  normalizeSelection: (fn: (selection: Selection) => void) => void
   onChange: () => void
   // Overrideable core actions.
   addMark: (key: string, value: any) => void
--- a/docs/concepts/11-normalizing.md
+++ b/docs/concepts/11-normalizing.md
@@ -66,7 +66,7 @@ But what if the child has nested blocks?
 | cell4 | cell5 | cell6 |
 | cell7 | cell8 | cell9 |
 
-`Selection` is usually continuous, however in `Grid` data structure(such as a table) `Selection` is expected to be expressed by cell matrix. Take the matrix above as an example, when `anchor` is located at `cell1` and `focus` is located at `cell5`, `cell1` `cell2` `cell3` `cell4` and `cell5` will all be selected. In practice we expect only `cell1` `cell2` `cell4` `cell5` to be selected in matrix form(exclude `cell3`), and the execution of node changes will only applied to these 4 cells.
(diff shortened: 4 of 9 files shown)


dylans (changes requested):
This generally looks fine. I would want to see a few additional things before landing it:

* Unit test(s)
* Update to docs
* Changeset (see instructions in the PR)

yanmao-cc:
> I'm not a fan of this change tbh, as this can be archived in other ways and seems quite hackish (running stuff passed into `withoutNormalizing` multiple times is quite unexpected and will break a tone of stuff, even within slate itself). This also changes the signature/behavior of `Editor.marks` when not using `normalizeSelection` (it will never return null and always a new object, so you can't memo anything based upon it). It also changes the editor selection without using operations which will cause other unintended side-effects.
> 
> Even ignoring that, it's very counterintuitive what `normalizeSelection` does, even with the documentation (at least for me). This should be handled outside of slate and can be done in user-land, even though it's a bit more complex.

Yes, I'm currently working on it outside the slate. But if you have a data type like `grid`, it seems to be a must

```javascript
Editor.withoutNormalizing = (editor: Editor, fn: () => void) => { 
  if(Editable.isEditor(editor)) {
    editor.normalizeSelection(selection => {
      if(editor.selection !== selection) editor.selection = selection
      withoutNormalizing(editor, fn)
    })
  } else {
    withoutNormalizing(editor, fn)
  }
}

Editor.marks = (editor: Editor): Omit<Text, 'text' | 'composition'> => {
  let marks: Omit<Text, 'text' | 'composition'> = {}
  if(Editable.isEditor(editor)) {
    editor.normalizeSelection(selection => {
      const editorMarks = slateMarks({...editor, selection})
      marks = Object.assign(marks, editorMarks)
    })
  } else {
    return slateMarks(editor) ?? marks
  }
  return marks
}
```

BitPhinix:
The way we handle it inside our plugins is that we added an `editor.toggleMark` inside our "core plugin" which is used by keyboard shortcuts etc. to toggle marks. We intercept this `toggleMark` inside our table plugin to handle it based upon the "table selection". You can use a similar pattern for almost everything else without changing anything about `Editor.withoutNormalizing`/`Editor.marks`

yanmao-cc:
> The way we handle it inside our plugins is that we added an `editor.toggleMark` inside our "core plugin" which is used by keyboard shortcuts etc. to toggle marks. We intercept this `toggleMark` inside our table plugin to handle it based upon the "table selection". You can use a similar pattern for almost everything else without changing anything about `Editor.withoutNormalizing`/`Editor.marks`

This is possible in an ideal situation. But I have here not only toggleMark but also toggleList , toggleQuoteblock , toggleHeading and so on. And other plugins also use a lot of APIs such as setNodes wrapNodes unwrapNodes.

So I abstract things like tables into grid data types. Ideally, other non-grid plugins do not need to pay attention to whether they are in the grid. Each grid plugin will handle the selection and boundary by itself and pass it on.

BitPhinix:
I'm sorry, but again, I don't think this is something that should be handled inside slate itself. This is something you should handle inside your editor. You could add something similar to what you do to the `withoutNormalizing` in this pr. E.g., adding a `Editor.modifySelection(fn: (range: Range) => void)` inside your editor which is then called with the ranges you would get with your current `normalizeSelection` call (just thinking out loud, not sure about your exact implementation). But you can't assume that everything that is passed to `withoutNormalizing` can just be called again with a different selection because that would be a huge breaking change for almost all code bases using slate (and would probably break a tone of stuff inside slate internally).

yanmao-cc:
> I'm sorry, but again, I don't think this is something that should be handled inside slate itself. This is something you should handle inside your editor. You could add something similar to what you do to the `withoutNormalizing` in this pr. E.g., adding a `editor.modifySelection(fn: (range: Range) => void)` inside your editor which is then called with the ranges you would get with your current `normalizeSelection` call (just thinking out loud, not sure about your exact implementation). But you can't assume that everything that is passed to `withoutNormalizing` can just be called again with a different selection because that would be a huge breaking change for almost all code bases using slate (and would probably break a tone of stuff inside slate internally).

It doesn't matter, I just submitted this PR tentatively, it may be more of a discussion process. Because I really need a way to modify `selection` before executing api calls such as `setNodes` `wrapNodes` `unwrapNodes`. From the perspective of the currently provided API, `withoutNormalizing` is a good entry point

Issue #4144, opened 2021-03-27
Error in signature of withReact


ChildishForces:
**Description**
Attempting to start a new Typescript project on @next is causing interface issues from the 1st line.

`const editor = React.useMemo(() => withReact(createEditor()), []);`

Immediately complains because `createEditor()` returns a `BaseEditor`  and the definition for `withReact` requires a generic type `T` that extends from `ReactEditor` and then returns a union of `T & ReactEditor` which doesn't seem right.

**Recording**
(Definition: slate-react/dist/plugin/with-react.d.ts)

**Steps**
To reproduce the behavior:
1. On a fresh TS app write `const editor = React.useMemo(() => withReact(createEditor()), []);`

**Expectation**
I expect this is an error, and that withReact wants a generic extending `BaseEditor`

**Environment**
- Slate Version: [e.g. 0.60.15]
- Operating System: [N/A]
- Browser: [N/A]
- TypeScript Version: 4.1.5

mitchheddles:
@ChildishForces I had the same issue, but was able to fix it by defining a custom editor.

Check out the typescript docs https://github.com/ianstormtaylor/slate/blob/main/docs/concepts/11-typescript.md#defining-editor-element-and-text-types

lensbart:
> Check out the typescript docs https://github.com/ianstormtaylor/slate/blob/main/docs/concepts/11-typescript.md#defining-editor-element-and-text-types

Update: https://github.com/ianstormtaylor/slate/blob/main/docs/concepts/12-typescript.md

santialbo:
This is weird, the types are correct in the source
https://github.com/ianstormtaylor/slate/blob/fece5c157099b5ddef9f6dc8d070f2fdb3724b90/packages/slate-react/src/plugin/with-react.ts#L27

However on the generated type definitions they are wrong
```
export declare const withReact: <T extends ReactEditor>(editor: T) => T & ReactEditor;
```
should be
```
export declare const withReact: <T extends Editor>(editor: T) => T & ReactEditor;
```

abdessamadelhamdany:
I'm facing the same issue, seems like incompatible types, with some missing properties
as a work around I'm casting it to `ReactEditor` like so:
```
const editor = useMemo(() => withReact(createEditor() as ReactEditor), []);
```
as it doesn't seem to cause any compilation issues

iamdddev:
Same here, but I was able to resolve like the implementation of @abdessamadelhamdany

[...]

PR #5091, opened 2022-08-19
Fix - `withReact` type signature

e1himself:
**Description**
`withReact()` plugin type signature is inaccurate. 

In reality, it accepts any `editor` instance implementing `BaseEditor` and extends it with additional `ReactEditor` API.

But the type definitions require to pass an instance that is already a `ReactEditor`. 

The typical workaround is either to [cast the instance with `as ReactEditor`](https://github.com/ianstormtaylor/slate/issues/4144#issuecomment-1046097560) or [define the `Editor` custom type in the project](https://github.com/ianstormtaylor/slate/issues/4144#issuecomment-1075045383). However, both are band-aid solutions to the real problem: invalid type spec of the `withReact()` function.

**Issue**
Fixes: #4144

**Example**

Before the fix:

Screenshot from 2022-08-19 17-38-17

After the fix:

Screenshot from 2022-08-19 17-38-45

**Context**

n/a

**Checks**

### DIFF
--- a/.changeset/smooth-queens-tie.md
+++ b/.changeset/smooth-queens-tie.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fix `withReact()` function type definition
--- a/packages/slate-react/src/plugin/with-react.ts
+++ b/packages/slate-react/src/plugin/with-react.ts
@@ -1,5 +1,14 @@
 import ReactDOM from 'react-dom'
-import { Editor, Node, Operation, Path, Point, Range, Transforms } from 'slate'
+import {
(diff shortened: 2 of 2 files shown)


dylans (approved):

dylans:
So this looks fine to me. I do believe there was a deliberate reason why they deliberately kept ReactEditor in BaseEditor but I'm not remembering what that was. Leaving this open for a couple of days in case anyone remembers why.

e1himself:
> Leaving this open for a couple of days in case anyone remembers why.

:+1: 

Thanks.

PR #5105, opened 2022-08-31
Use callback-style ref in `Element` component

### LINKED ISSUE #5104: Cannot resolve a Slate node from DOM node when parent Element re-render child Element using React context
**Description**
I have a parent element passing down some of its properties using React context. When a child element hooks to the context and re-render to another DOM element, Slate can no longer resolve the Slate node from that DOM node.

**Recording**

https://user-images.githubusercontent.com/1330321/187599832-faa76b83-388a-4ffe-b29f-aa6371916112.mp4

**Sandbox**
https://codesandbox.io/s/slate-context-render-lm8don?file=/index.js

**Steps**
To reproduce the behavior:
1. Click at end of the text
1. Click the "Switch" button
2. Click at empty space right to the text

It's not clear why step 1 is required to 100% reproduce this issue, maybe otherwise some of slate-react components re-renders and updates DOM node refs.

**Expectation**
No error got thrown

**Environment**
- Slate Version: 0.82.1
- Browser: Microsoft Edge 106.0.1356.0
- TypeScript Version: N/A

yume-chan:
**Description**
When a parent Element re-render child Elements using React context, Slate can no longer track the DOM node to Slate node mapping.

**Issue**
Fixes #5104

**Example**
The same repro in #5104 with locally patched `slate-react` package:

https://user-images.githubusercontent.com/1330321/187604574-36b4e2ea-2616-42fe-a51e-73de17498563.mp4

**Context**
This PR changes `Element` to use callback-style ref, so every change in the ref target (including unmounting) will be observed.

`Editor` and `Leaf` doesn't need this treatment, because they attach `ref` to nodes that can only change when the component itself re-renders.

However, if we also change `Editor` and `Leaf`, then `ELEMENT_TO_NODE` and `NODE_TO_ELEMENT` maps can use normal `Map`s, instead of `WeakMap`s, which is 5x faster in reading, according to some random [benchmark](https://gist.github.com/jung-kim/d146b25ea754e73bf60a?permalink_comment_id=3811484#gistcomment-3811484).

**Checks**

### DIFF
--- a/.changeset/poor-jokes-chew.md
+++ b/.changeset/poor-jokes-chew.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Change `Element` component to use callback-style ref to reliably track DOM node of rendered custom elements
--- a/packages/slate-react/src/components/element.tsx
+++ b/packages/slate-react/src/components/element.tsx
@@ -1,11 +1,10 @@
-import React, { Fragment, useRef } from 'react'
+import React, { useCallback } from 'react'
 import getDirection from 'direction'
 import { Editor, Node, Range, Element as SlateElement } from 'slate'
 
 import Text from './text'
 import useChildren from '../hooks/use-children'
 import { ReactEditor, useSlateStatic, useReadOnly } from '..'
-import { useIsomorphicLayoutEffect } from '../hooks/use-isomorphic-layout-effect'
 import {
   NODE_TO_ELEMENT,
   ELEMENT_TO_NODE,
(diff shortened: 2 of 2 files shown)


dylans (changes requested):
I'm verifying that this won't break in some other way.

In the interim, please follow the instructions in the PR and add a changeset so that this ready to go once I've had time to review/verify.

yume-chan:
Changset added.

dylans (approved):

bryanph (approved):
Great change 👍

PR #5135, opened 2022-09-23
Allow void elements to receive marks

SmilinBrian:
**Description**
Some void elements are effectively stand-ins for text, such as with the mentions example, where the mention element renders the character's name. Users might want to format Void elements like this with bold, or set their font and size, so `editor.markableVoid` tells Slate whether or not to apply Marks to the text children of void elements.

- Adds `markableVoid()` as a schema-specific overrideable test.
- Changes `addMark` and `removeMark` so marks can apply to voids. Also changes behavior of collapsed selection so that if a markable Void is selected, the mark will be applied / removed.
- Shows how `markableVoid()` can work in the Mentions example

**Example**
image
The Mentions example, with R2-D2 in bold text

**Context**
I have achieved a similar effect by overriding `editor.addMark()` and `editor.removeMark()` in our own Slate use case, but we risk missing out on important fixes or changes by having such an override, and this seems generally useful.

**Checks**

### DIFF
--- a/.changeset/modern-cooks-grow.md
+++ b/.changeset/modern-cooks-grow.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Allow void elements to receive marks via markableVoid()
--- a/docs/api/nodes/editor.md
+++ b/docs/api/nodes/editor.md
@@ -12,6 +12,7 @@ interface Editor {
   // Schema-specific node behaviors.
   isInline: (element: Element) => boolean
   isVoid: (element: Element) => boolean
+  markableVoid: (element: Element) => boolean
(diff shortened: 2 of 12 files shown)


dylans (changes requested):
Could we please decouple the general documentation cleanup into a separate PR? This makes it easier should we need to revert the proposed change in the future.

SmilinBrian:
Removed the un-related documentation changes and rebased to combine the new feature into one clear commit.

Sorry this took so long - I was on a cruise ship for a couple weeks so internet was dodgy.

dylans (approved):

dylans:
Thanks! I’ll plan to land this over the weekend unless others have strong suggestions.

Issue #2288, opened 2018-10-21
Make some operations take a range of blocks as input


bryanph:
#### Do you want to request a _feature_ or report a _bug_?
Suggestion/discussion

#### What's the current behavior?
All operations currently operate on a single block. This can sometimes end up in a lot of operations when performing a large change like for example doing a `selectAll` + `delete()` and make them difficult to follow when for example moving a range of nodes.

#### What's the suggested behavior?
Make the appropriate operations take a range as input. 

I can see this potentially applying to the following operations:

* `move_node`: move a range of nodes to a new path
* `remove_node`: remove a range of nodes
* `set_node`: set new properties on a range of nodes

This also has the added benefit of simplifying the implementation of these changes and makes implementing something like #2075 a lot easier..

bryanph:
I don't really have the time to do a proper research but I will be reading about CKEditor and Prosemirror a bit and report my findings here.

bryanph:
It has been a long time but this has become relevant for me again. 

As for how prosemirror does this, it seems that there all transformations are done using the following kinds of operations (these are called steps in prosemirror):
* `ReplaceStep`: https://prosemirror.net/docs/ref/#transform.ReplaceStep
* `ReplaceAroundStep`: https://prosemirror.net/docs/ref/#transform.ReplaceAroundStep
* `AddMarkStep`: https://prosemirror.net/docs/ref/#transform.AddMarkStep
* `RemoveMarkStep`: https://prosemirror.net/docs/ref/#transform.RemoveMarkStep

The `ReplaceX` operations combine the ability to remove and add content in a single operation basically and both are operating on a range.

PR #5137, opened 2022-09-26
transform.delete and transform.insertFragment performance optimize

mainhanu:
**Description**
delete and insert large fragment can be very time consuming，the time complexity is `O(n2)`

**Example**
```ts
import { createEditor, Editor, Descendant, Transforms } from "slate";
import type { Descendant } from "slate";

const createP = (i: string = '') => ({
  type: "p",
  children: [
    { text: `${i} This is editable ` },
    { text: 'rich', bold: true },
    { text: ' text, ' },
    { text: 'much', italic: true },
    { text: ' better than a ' },
    { text: '<textarea>', code: true },
    { text: '!' },
  ],
})

// one thousand paragraphs
const largeFragment: Descendant[] = new Array(2000).fill(1).map((_, i) => createP(i + ''));

export const testDelete = () => {
  const editor = createEditor();
  editor.children = largeFragment;

  const sel = Editor.range(editor, []);
  Transforms.select(editor, sel);

  console.time("delete");
  Transforms.delete(editor);
  console.timeEnd("delete");
};

```

environment
* `MacBook Pro (13-inch, 2019) 2.4 GHz 4 Intel Core i5`
* `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/105.0.0.0 Safari/537.36`

result: `2000` pargagraphs
| seq       | time(millisecond) |
|---------|----------|
| 1       | 12210.78 |
| 2       | 12882.65 |
| 3       | 13567.65 |
| 4       | 13315.84 |
| average | 12994.23 |

but with `200` pargagraphs
| seq       | time(millisecond) |
|---------|----------|
| 1       | 191 |
| 2       | 179 |
| 3       | 177 |
| 4       | 180 |
| average | 181 |

**Context**
The reason is related to the number of **path refs** in the apply process. 

for `transform.delete`, The following code causes each **apply** to process the **path transform** of the first i-1 paragraphs.
```ts
for (const pathRef of pathRefs) {
    const path = pathRef.unref()!
    Transforms.removeNodes(editor, { at: path, voids })
}
```
we can remove nodes from last to first to avoid use path ref.
```ts
const middlePaths = pathRefs
  .reverse()
  .map(r => r.unref())
  .filter((r): r is Path => !!r)

for (const p of middlePaths) {
  Transforms.removeNodes(editor, { at: p, voids })
}
```

for `transform.insertFragment`, the `O(n2)` has something todo with the dirtyPath logic, each insertNode will add new dirtyPath to DIRTY_PATHS and subsquent `apply` will transform all previous dirtyPath。I hadn't figure out how to opitimize it。according to the logic, move large fragment may also has the same problem。

I just add a tiny optimization to `Path.transform`, array destruction is faster than `immer's produce`, and it also make remove 2000 paragraphs **3x faster**。

**after optimize**

delete 2000 paragraphs
| seq       | time(millisecond) |
|---------|----------|
| 1       | 896 |
| 2       | 843 |
| 3       | 853 |
| 4       | 850 |
| average | 860 |

delete 200 paragraphs
| seq       | time(millisecond) |
|---------|----------|
| 1       | 49 |
| 2       | 46 |
| 3       | 47 |
| 4       | 44 |
| average | 46 |

**Checks**

### DIFF
--- a/packages/slate/src/transforms/text.ts
+++ b/packages/slate/src/transforms/text.ts
@@ -187,9 +187,13 @@ export const TextTransforms: TextTransforms = {
         }
       }
 
-      for (const pathRef of pathRefs) {
-        const path = pathRef.unref()!
-        Transforms.removeNodes(editor, { at: path, voids })
+      const middlePaths = pathRefs
+        .reverse()
+        .map(r => r.unref())
+        .filter((r): r is Path => !!r)
+
+      for (const p of middlePaths) {
+        Transforms.removeNodes(editor, { at: p, voids })
       }
 
       if (!endVoid) {
--- a/packages/slate/src/transforms/text.ts
+++ b/packages/slate/src/transforms/text.ts
@@ -187,14 +187,11 @@ export const TextTransforms: TextTransforms = {
         }
       }
 
-      const middlePaths = pathRefs
+      pathRefs
         .reverse()
         .map(r => r.unref())
-        .filter((r): r is Path => !!r)
-
-      for (const p of middlePaths) {
-        Transforms.removeNodes(editor, { at: p, voids })
-      }
+        .filter((r): r is Path => r !== null)
+        .forEach(p => Transforms.removeNodes(editor, { at: p, voids }))
 
       if (!endVoid) {
         const point = endRef.current!
--- a/.changeset/new-needles-itch.md
+++ b/.changeset/new-needles-itch.md
@@ -1,0 +1,5 @@
+---
+'slate': minor
+---
+
+transform.delete and transform.insertFragment performance optimize
--- a/packages/slate/src/interfaces/path.ts
+++ b/packages/slate/src/interfaces/path.ts
@@ -1,4 +1,3 @@
-import { produce } from 'immer'
 import { Operation } from '..'
 import { TextDirection } from './types'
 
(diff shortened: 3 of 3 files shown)


packages/slate/src/transforms/text.ts
```diff
@@ -187,9 +187,13 @@ export const TextTransforms: TextTransforms = {
         }
       }
 
-      for (const pathRef of pathRefs) {
-        const path = pathRef.unref()!
-        Transforms.removeNodes(editor, { at: path, voids })
+      const middlePaths = pathRefs
```
nemanja-tosic, line 190: I think the idea of reversing the order of operation to remove without affecting `pathRefs` is a good one.

I'm curious why are we first creating an array using array functions and then using `for of`. What are the benefits of what you did here versus:

```ts
pathRefs
  .reverse()
  .map(r => r.unref())
  .filter((r): r is Path => r !== null) // note i would compare with null as per the unref interface
  .forEach(p => Transforms.removeNodes(editor, { at: p, voids })
```

Also, would it be possible to not use `pathRefs` at all given the fact that all `removeNodes` operations would work and the code before `removeNodes` doesn't seem to alter the tree. It is possible that not using `pathRefs` might be asking for trouble because something might get in between `matches` and `removeNodes` but seems like a possible performance gain (and performance gains are usually... less pretty).

mainhanu: 1. removed unnecessary `for...of`
2. Yes, pathRef can be removed, but currently there is only one remove_text, which has tiny impact on performance. Secondly, as you describe, using pathRef correctly expresses the relationship  between matches and removeNodes。

I think the importance is to let developers pay attention to the performance impact of `path/point/range ref` on `apply` Op. The mechanism based on location ref is prone to O(n2) operations, which will affect the experience of edit large fragment.

bryanph:
Ultimately we need operations like `insert_nodes` and `remove_nodes` that operate on a range to make this more performant I think

mainhanu:
> Ultimately we need operations like `insert_nodes` and `remove_nodes` that operate on a range to make this more performant I think (#2288)

I think current `Transform.insertFragment` and `Transform.delete` includes scenario like insert_nodes and removeNodes。we should optimize these methods，like how to optimze the cumulative dirtypath transform logic。

create new Operations will affect OT transform logic，which will case more problems to consider

dylans:
@bryanph any further thoughts after https://github.com/ianstormtaylor/slate/pull/5137#issuecomment-1261794212 ?

dylans:
> edit: I do have one question actually, do we know why this is so slow with immer? It's just operating on a flat array, seems like that should be comparable in performance to destructuring and applying changes to that?

They used to not recommend immer on arrays with thousands of items, but that doesn't seem to be part of their performance tips any more: https://immerjs.github.io/immer/performance#performance-tips

bryanph:
@dylans based on [this](https://immerjs.github.io/immer/performance/#pull-produce-as-far-up-as-possible) section it might just be that in these benchmarks`produce()` is called a whole lot of times which creates the overhead 🤷

bryanph (approved):

Issue #4862, opened 2022-03-04
Bad UX / bug when we triple click on text and type over it


nitishrddy:
**Description**
If we triple click on a block, the whole block gets selected (this is expected). If we start typing, instead of the text in the current block being replaced with the new text, the text in current block gets deleted and the new text appear at the start of the next block. The expectation is that the text appears in the same block instead of the next. 

One observation: At this point of time the anchor is {path: [x,0], offset: 0} and focus is {path: [x+1,0], offset: 0}. The same bug reproduces when we drag the mouse manually and select text until the anchor and focus are similar to the above anchor and focus or vice-vera.

This is reproducible in the Slate examples too. Sharing a recording of it below

**Recording**
(https://youtu.be/Xc7eMPxCNek)

**Sandbox**

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/richtext
2. Triple Click with mouse on any text in the first block
3. Start typing

**Expectation** 
Expected the text to appear in the first block instead of the second.

jameshfisher:
I agree that this is undesirable behavior, but I strongly disagree with the breaking change in https://github.com/ianstormtaylor/slate/pull/4868.

The behavior you're seeing here is actually Chromium's default contenteditable behavior; not a Slate bug. See demo: https://codepen.io/jameshfisher/pen/dyewjee

Note this does not happen in Firefox. I believe it's Firefox's behavior that we all expect and want.

The precise difference is that [Chrome triple-click creates a "hanging range"](https://docs.slatejs.org/api/nodes/editor#editor.unhangrange-editor-editor-range-range-options-greater-than-range). We should consider that the underlying bug, and push Chromium to fix the bug.

Relevant Chromium bug reports:

* https://bugs.chromium.org/p/chromium/issues/detail?id=862090
* https://bugs.chromium.org/p/chromium/issues/detail?id=59371

If we decide to fix this browser inconsistency in Slate, the correct fix cannot be to change the `insertText` behavior (as was done in https://github.com/ianstormtaylor/slate/pull/4868). This is a major breaking change, and is working around Chrome's bug at the wrong time. The correct fix would be to fix up the selection at triple-click time, presumably by using something like `unhangRange`.

PR #5154, opened 2022-10-12
Revert insertText breaking change that deletes fragment (fixes #5153)

### LINKED ISSUE #5153: Serious bug: selection is already deleted before insertText is called
**Description**
In `insertText`, the application can decide what to do in response to a user's request to insert some text. For example, `insertText` may decide that if the `editor.selection` is expanded, it will delete that text and replace it with the new text. However, `insertText` is also able to make other decisions and take other actions. In our application, we do the following check: if deleting the current selection would make the document invalid, then instead of invalidating the document by deleting the selection, we take no action. (We could do more sophisticated things, but the point is that `insertText` needs to decide whether to delete the current selection.)

As of slate-react 0.75.0, this is broken. It seems that the current selection is _already deleted_ before `insertText` is called! The application therefore has no opportunity to decide what to do with the current selection, because the decision has apparently already been made.

**Sandbox**
https://codesandbox.io/s/slatejs-bug-selection-is-deleted-before-inserttext-is-called-d5w1mu?file=/index.js

**Steps**
To reproduce the behavior:
1. Go to the sandbox above
2. Select some text
3. Type 'h'

Expected behavior: the text 'X' is inserted instead, due to the check in `insertText`.

Actual behavior: the text 'h' is inserted, because apparently the selected text was already deleted before `insertText` was called.

**Environment**
- Slate Version: Chrome with slate-react 0.75.0+; Firefox with slate-react 0.74.1+
- Operating System: Windows
- Browser: Chrome, Firefox

**Context**
I assume this change is an accidental bug, rather than deliberate: it breaks backwards compatibility with no record in the changelog, and violates Slate's goal of "completely customizing the editing experience without having to fight against the library's assumptions".

I can't think of any workaround for this bug. Therefore we are prevented from upgrading beyond slate-react 0.74.2.

jameshfisher:
**Description**
This reverts https://github.com/ianstormtaylor/slate/pull/4868.

The issue reported at https://github.com/ianstormtaylor/slate/issues/4862 is a valid complaint, but the change is not an appropriate fix. It is a serious breaking change, reported in https://github.com/ianstormtaylor/slate/issues/5153.

If we wish to fix Chromium's triple-click bug, we should do so by unhanging the selection when it is created; not by changing insertText behavior.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/5153

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -648,13 +648,6 @@ export const Editable = (props: EditableProps) => {
           case 'insertFromYank':
           case 'insertReplacementText':
           case 'insertText': {
-            const { selection } = editor
-            if (selection) {
-              if (Range.isExpanded(selection)) {
-                Editor.deleteFragment(editor)
-              }
-            }
-
             if (type === 'insertFromComposition') {
               // COMPAT: in Safari, `compositionend` is dispatched after the
               // `beforeinput` for "insertFromComposition". But if we wait for it


jameshfisher:
Thanks @dylans !

PR #5161, opened 2022-10-21
Adds descriptions to list of products

DougReeder:
**Description**
* Adds descriptions to list of products
* Removes broken link

**Issue**
It’s hard for readers to know which products to look at from the names alone.


dylans (approved):

dylans (commented):
If you have time, we should add Campfire and Coda to the list as well.

If not, I'll land this and add them later. Thanks for the efforts!

DougReeder:
campfirewriting.com ?  And which Coda?

PR #5168, opened 2022-10-26
fix: error getting parentEntry in selection range

i-artist:
**Description**
Wrong matching ancestor above a location in the document when selection range multiple nodes.

**Jsfiddle**
[demo](https://jsfiddle.net/Steven_zhong/wsoaL4zk/45/)

**Context**
The correctness of the method is guaranteed by different verification processing logic for points, paths or range.

**Checks**

### DIFF
--- a/.changeset/tame-melons-clap.md
+++ b/.changeset/tame-melons-clap.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Fixed above method that failed to get parentEntry when selection was range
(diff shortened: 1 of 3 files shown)


dylans (changes requested):
Ideally I'd like to see a test added that fails before this change and passes afterwards to avoid a future regression.

i-artist:
> Ideally I'd like to see a test added that fails before this change and passes afterwards to avoid a future regression.

Yeah, I agree with you.

i-artist:
before change:

image

after change:

image

dylans (approved):
Thanks!

could you add a quick changeset too and then I’ll land this over the weekend? (There’s instructions in the PR)

i-artist:
> Thanks!
> 
> could you add a quick changeset too and then I’ll land this over the weekend? (There’s instructions in the PR)

Sure, I'll add it later.

PR #5176, opened 2022-11-01
Adds an example using SlateNode.common() with the editor selection

DougReeder:
**Description**
The Slate API is difficult to grok.  This example shows how to use several calls together, and is useful besides.


dylans (approved):
Thanks!

Issue #5190, opened 2022-11-10
[Android] In some cases, the content pasted from the clipboard is empty.


YasinChan:
**Description**
Hi @BitPhinix !
I found a problem, I copied a text that contains \n, but I can't paste it in the editor, as you can see in the video below.

https://user-images.githubusercontent.com/22654945/200995108-973061ca-4c88-4378-bb8b-3a449c73009c.mp4

**Recording**

https://user-images.githubusercontent.com/22654945/200995130-ee71aece-ab0e-4810-a75c-fef3c516a948.mp4

I debugged with Chrome inspect and found that when triggered within the `handleDOMBeforeInput` method with type `insertText`, it would execute into this [logic](https://github.com/ianstormtaylor/slate/blob/6efe3d9a22ff7401e0b0dc02eca24dfb9eb574e2/packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts#L543) and execute `Editor.insertSoftBreak` causing  the entire text to be inserted as a break. What is the purpose of using `data.includes('\n')` to determine ?

**Environment**
- Slate Version: Latest
- Operating System: Android

PR #5195, opened 2022-11-11
fix: The content with `\n` pasted from the clipboard is empty.

YasinChan:
**Description**
The problem with pasting from the clipboard on Android has been located, and this PR resolved my problem on my phone. Maybe need @BitPhinix  to review this too.

**Issue**
Fixes: #5190 

**Context**

**Checks**

### DIFF
--- a/.changeset/lazy-files-repeat.md
+++ b/.changeset/lazy-files-repeat.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+Fix: The content with `\n` pasted from the clipboard is empty. #5190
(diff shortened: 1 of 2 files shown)


dylans (approved):
@BitPhinix any objections to this change?

BitPhinix:
Yes, sorry for the delay on this 🙇. The reason I've implemented it like this was definitely a shortcut, but there are some IMEs that will re-insert some content alongside the linebreak when pressing enter in the middle of a paragraph. So in those cases this PR would result in unexpected behaviour (alongside inserting '\n's inside the content). The ideal implementation would be to store the diff for the text changes without the '\n', figure out the point at which the '\n' should be inserted and schedule a line-break action there

Slapbox:
Perhaps related: https://github.com/ianstormtaylor/slate/issues/5273

YasinChan:
Fixed! #5359

Issue #4314, opened 2021-06-02
minHeight is no longer respected in slate-react 0.62+


hanford:
**Description**
This [commit](https://github.com/ianstormtaylor/slate/commit/00393bb18548976d4e79c8bc46d54b2125bb6701#diff-18ef6af284404b1670857e562c183f2da6774d81bb65ed39d82fe124f769787b) introduces a regression where slate editors min-height is now constantly being set.

**Recording**
![image](https://user-images.githubusercontent.com/2148168/120557004-20fd5380-c3b2-11eb-8ca9-7ac16043dba9.png)

**Sandbox**
https://stackblitz.com/edit/react-f4v5on

**Expectation**
I should be able to control the minHeight of my editor based on this value

**Environment**
- Slate Version: 0.62+

**Context**

I'm no longer able to specify a `minHeight` on my `<Editable />` and have it be respected, this is problematic because I have been relying on minHeight to specify the size my editors initialize with

gfohl:
Looks like line 43 in `packages/slate-react/src/components/leaf.tsx` might be overriding it, if there is a placeholder:
`editorEl.style.minHeight = `${placeholderEl.clientHeight}px`

aaron-swarm:
As a workaround:

```
useEffect(() => {
  const editorEl = document.querySelector<HTMLDivElement>(
    '[data-slate-editor="true"]',
  );
  editorEl.style.minHeight = `${rows * 24}px`;
}, []);
```

tom2strobl:
https://github.com/ianstormtaylor/slate/blob/25afbd43001cdee852af6386d2b701d943b788da/packages/slate-react/src/components/leaf.tsx#L32 is pretty dirty since it only works if there is only one slate instance on a page. the workaround of @aaron-swarm doesn't help in this case either since both fire `querySelector` from `document` which always targets the first slate instance on the page.

So right now it is completely impossible to have two slate instances, where the first one has a minHeight of 60px and the second one of 100px. No `renderPlaceholder` and no useEffect-hacky-business can do anything since the leaf.tsx hardcoded line will always pick up the first instance and then forcefully overwrite minHeight on that one. The first editor will always get the 100px, because https://github.com/ianstormtaylor/slate/blob/25afbd43001cdee852af6386d2b701d943b788da/packages/slate-react/src/components/leaf.tsx#L40 writes it to that instance and we can't bypass that.

hanford:
[...]

PR #5206, opened 2022-12-01
Use stylesheets to give Editable components default styles

kylemclean:
**Description**
This PR makes the default styles of an `Editable` component come from stylesheets added to the DOM rather than inline styles on the editor element.  This resolves an issue in which setting the `position`, `outline`, `white-space`, or `word-wrap` properties using the `className` prop would have no effect. It also resolves issue #4314 in which setting the `min-height` property on an editor would have no effect.

Additionally, a change was made to make the `min-height` of editors change when the size of their placeholder element changes.

**Issue**
Fixes: #4314

**Example**
Pictured is a modification of the custom placeholder example.

### Old
Screenshot from 2022-12-01 02-58-32

The `min-height` changes have no effect, and the change to `outline` only has an effect when passed as part of the `style` prop.

### New
Screenshot from 2022-12-01 02-59-10

The `min-height` and `outline` changes have an effect, regardless of whether they are set as part of a CSS class or on the `style` prop.

**Context**
Previously, `Editable` components used inline styles to set their `position`, `outline`, `white-space`, and `word-wrap` CSS properties to some chosen default values. These properties could be overridden by setting them with the `style` prop on the component, but attempting to override them by setting the `className` prop would not work because inline styles take precedence.

Additionally, as detailed in #4314, code in the `Leaf` component sets the `min-height` property on editors to the height of their placeholder to prevent the placeholder from overflowing the editor bounds. However, because this was done by directly setting an inline style on the editor element, users were unable to set the `min-height` of an editor using `className` or `style` props on the editor.

I decided to solve the first issue by introducing a global `<style>` element that gets added to the DOM on the first `Editable` mount and removed on the last `Editable` unmount. It contains the `position`, `outline`, `white-space`, and `word-wrap` properties that were previously passed to the `style` prop of each editor. The CSS rule in this stylesheet uses the `:where` pseudoclass to target editors, giving it 0 specificity and thus allowing user stylesheets to override its styles.

The second issue required a bit more effort. Fixing the first issue allowed the `min-height` property to be changed by the user using a stylesheet or inline styles, but it still needed to be possible for the `min-height` to default to the height of the placeholder element, and it needed to be done without using an inline style. There can be more than one editor on a page, each with its own unique placeholder height, so each editor needs its own default `min-height`. In order to have each editor element have a different `min-height` without giving them inline styles, the editor elements need some sort of unique identifier. To accomplish this, I added an `id` field to the `Editor` type. When an editor is created, it simply uses a counter of the number of `Editor` instances created as its `id`, then increments a counter.

Editor elements expose their identifier through their new `data-slate-editor-id` attribute. When an `Editable` component mounts, it will now create a `<style>` element that targets only itself using its `data-slate-editor-id` attribute value. `Leaf` components access that `<style>` element through a `WeakMap` and set the `min-height` of the editor on it depending on the placeholder height. Because this stylesheet also uses the `:where` pseudoclass, the `min-height` can be overridden by a user stylesheet.

During testing, I discovered an issue where sometimes the placeholder height that is measured will be inaccurate on initial page load, leading to the set `min-height` being incorrect until the component renders again. This was resolved by having `Leaf` components create a `ResizeObserver` that updates the `min-height` of the editor element whenever the size of the placeholder element changes.

**Checks**

### DIFF
--- a/.changeset/strange-pens-lie.md
+++ b/.changeset/strange-pens-lie.md
@@ -1,0 +1,6 @@
+---
+'slate': minor
+'slate-react': minor
+---
+
+Use stylesheet for default styles on Editable components
--- a/packages/slate-react/package.json
+++ b/packages/slate-react/package.json
@@ -30,6 +30,7 @@
     "@types/react": "^16.9.13",
     "@types/react-dom": "^16.9.4",
     "@types/react-test-renderer": "^16.8.0",
+    "@types/resize-observer-browser": "^0.1.7",
     "react": ">=16.8.0",
     "react-dom": ">=16.8.0",
     "react-test-renderer": ">=16.8.0",
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -55,6 +55,7 @@ import {
   EDITOR_TO_ELEMENT,
   EDITOR_TO_FORCE_RENDER,
   EDITOR_TO_PENDING_INSERTION_MARKS,
+  EDITOR_TO_STYLE_ELEMENT,
   EDITOR_TO_USER_MARKS,
   EDITOR_TO_USER_SELECTION,
   EDITOR_TO_WINDOW,
@@ -832,6 +833,16 @@ export const Editable = (props: EditableProps) => {
 
       if (mountedCount <= 0)
         document.querySelector('style[data-slate-default-styles]')?.remove()
+    }
+  }, [])
+
+  useEffect(() => {
+    const styleElement = document.createElement('style')
+    document.head.appendChild(styleElement)
+    EDITOR_TO_STYLE_ELEMENT.set(editor, styleElement)
+    return () => {
+      styleElement.remove()
+      EDITOR_TO_STYLE_ELEMENT.delete(editor)
     }
   }, [])
 
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -76,6 +76,9 @@ const Children = (props: Parameters<typeof useChildren>[0]) => (
   <React.Fragment>{useChildren(props)}</React.Fragment>
 )
 
+// The number of Editable components currently mounted.
+let mountedCount = 0
+
 /**
  * `RenderElementProps` are passed to the `renderElement` handler.
  */
@@ -802,6 +805,36 @@ export const Editable = (props: EditableProps) => {
     })
   })
 
+  useEffect(() => {
+    mountedCount++
+
+    if (mountedCount === 1) {
+      // Set global default styles for editors.
+      const defaultStylesElement = document.createElement('style')
+      defaultStylesElement.setAttribute('data-slate-default-styles', 'true')
+      defaultStylesElement.innerHTML =
+        // :where is used to give these rules lower specificity so user stylesheets can override them.
+        `:where([data-slate-editor]) {` +
+        // Allow positioning relative to the editable element.
+        `position: relative;` +
+        // Prevent the default outline styles.
+        `outline: none;` +
+        // Preserve adjacent whitespace and new lines.
+        `white-space: pre-wrap;` +
+        // Allow words to break if they are too long.
+        `word-wrap: break-word;` +
+        `}`
+      document.head.appendChild(defaultStylesElement)
+    }
+
+    return () => {
+      mountedCount--
+
+      if (mountedCount <= 0)
+        document.querySelector('style[data-slate-default-styles]')?.remove()
+    }
+  }, [])
+
   return (
     <ReadOnlyContext.Provider value={readOnly}>
       <DecorateContext.Provider value={decorate}>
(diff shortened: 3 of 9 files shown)


dylans (approved):
Thanks, this is long needed improvement, and I appreciate your approach and detailed explanation.

dylans (changes requested):
I believe you may be missing the commit to yarn.lock which is blocking the automated tests.

kylemclean:
Thank you.

I've added the yarn.lock change.

dylans (approved):
Thanks!

dylans:
As a note @kylemclean we accepted this and released it, but we do think there are things that could be done a bit more efficiently with respect to the observer logic that was added if you have time. Not urgent (unless the change causes a performance regression when more people upgrade).

barro32:
This causes `ReferenceError: Can't find variable: ResizeObserver` in Safari iOS < 13.4
https://caniuse.com/?search=resizeobserver

kylemclean:
> This causes `ReferenceError: Can't find variable: ResizeObserver` in Safari iOS < 13.4 https://caniuse.com/?search=resizeobserver

@dylans Should a polyfill be added for `ResizeObserver`? It's supported by all modern browsers since March 2020.

bryanph:
Yes I think it should be polyfilled

Also this whole placeholder min height business seems pretty convoluted and error-prone at this point. I haven't really needed it to implement placeholders for elements so I'm not quite sure when it becomes required? Perhaps just leaving it for user-land with some examples might be a better idea in the future?

kylemclean:
@bryanph
It seems that the min-height trick was implemented to prevent tall placeholders from overflowing the editor. There might be a better way of preventing that problem, but I haven't looked into it.

I'll try adding a polyfill for `ResizeObserver` and open a PR.

vshlos:
so all of the innerHtml calls destroyed content security policy. With this release, CSP doesn't really work with Slate unless we allow unsafe-inlinfe which is pretty bad.

kylemclean:
@vshlos Could you reply with the CSP you are using?

dylans:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for

Sorry, I forgot about the impact of innerHTML on CSP. We'll need to find another way and release an update.

kylemclean:
@dylans
If we wish to address the CSP issue directly, one possible solution would be to add a nonce to the `<style>` elements that Slate generates. This is the solution that CSS-in-JS libraries like JSS [1], styled-components [2], and Emotion [3][4] have used. I've successfully implemented this approach, but I'm not sure about it. To preserve security, a server would need to randomly generate the nonce for each request, which essentially requires CSP users to use SSR.

I have been reconsidering the changes made in this PR. I chose the method of dynamically injecting `<style>` elements as it allows users to style editors using a class name, which is generally the preferred method of styling. But before this change, it still would have been possible to do that using `!important`, and custom styling was mostly possible using inline styles. Not ideal, but ultimately doable with a minor amount of effort from the user.

Unfortunately, dynamically injecting `<style>` elements is complicated by SSR [5], and React is now recommending against it, mainly for performance reasons. [6] And of course, as we have been reminded, it isn't really compatible with CSP. So I think we might want to consider reverting most of the changes of this PR, and go back to using inline styles for default styles. The changes made regarding placeholder height should stay, though, as they fix a bug I was dealing with: #4314

To summarize, here are two possible paths (there might other options that are better):
- Keep using dynamically injected `<style>` elements and address the CSP issue as well as any potential issues that arise with SSR, performance, etc. While this option offers the user the most flexibility in terms of styling (allows users to easily use classes to style editors), it seems to be the most complex to maintain and prone to future problems.
- Go back to using inline styles. This would simplify the implementation, and would eliminate the CSP issue as well as other potential issues with dynamic `<style>` elements. It would be good to add information in the docs about how to properly style the editor (state that either inline styles or classes with `!important` must be used). Perhaps we could add a prop to the `Editable` component that disables the default inline styles if the user wants to use classes.

I'm leaning towards the latter option, so I might try implementing that along with the original fix to #4314 that I made.

[1] https://cssinjs.org/csp/?v=v10.9.2
[2] https://github.com/styled-components/styled-components/issues/887
[3] https://github.com/emotion-js/emotion/issues/403
[4] https://emotion.sh/docs/@emotion/cache#nonce
[5] https://github.com/reactwg/react-18/discussions/110
[6] https://beta.reactjs.org/reference/react/useInsertionEffect#injecting-dynamic-styles-from-css-in-js-libraries

dylans:
@kylemclean Agreed, sounds like the most straightforward path is the latter, and then if a better solution emerges we can pursue it separately. Thanks for your thorough response and explanation!

vshlos:
Also keep in mind we are not really doing any custom styling at all. So even if you keep this approach, it should not be injecting a custom style element unless styles are provided.

If you do want to keep supporting this, you can always use a nonce. Here is another library we are using that takes a nonce as a parameter and injects it into the style element that gets created: https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/content-security-policy.md

PR #5248, opened 2023-01-06
Switch from cypress to playwright.

gdborton:
**Description**
TLDR this PR:
 - [x] Switches existing integration tests from cypress to playwright
 - [x] Updates the version of yarn used to install dependencies
 - [x] Adds FF and Safari (when on mac) to the matrix of integration tests.

**Issue**
Fixes: (link to issue)

**Example**
N/A

**Context**
We're using slate for a project, but found that it has a couple of issues w/ Android (specifically around placeholders and auto-capitalize). I'm not sure I'll be able to make the change that fixes our issues, but I did want to try to contribute to the fix.

My ultimate goal here is to add an additional integration test for android that ensures autocapitalization is respected, and that the placeholder behaves as expected. To do this, I've switched the integration test runner from cypress to playwright (which has experimental android support).

As a neat side effect, the test suite is now MUCH faster. On my machine, it's come down from `7m 8s` to `1m 9s`

I've also updated the version of yarn being used as I noticed w/ the old version that CI would periodically fail.

**Checks**


bryanph:
+1 on moving to playwright, we can cover a lot more platforms that way.

Question: the integration tests say that they've run successful in the CI but when looking at the actual log output it seems a few failed?

playwright.config.ts
```diff
@@ -0,0 +1,90 @@
...
+    use: {
+      ...devices['Desktop Safari'],
+    },
+  })
+}
+
+const retries = process.env.PLAYWRIGHT_RETRIES
+  ? +process.env.PLAYWRIGHT_RETRIES
+  : process.env.CI
+  ? 5
```
gdborton, line 40: 5 retries in CI. If any test fails, it gets rerun and you get a flaky test report at the end of the run.

https://github.com/ianstormtaylor/slate/actions/runs/3859077663/jobs/6583635166#step:4:291

bryanph:
Ah I see thanks for the clarification. Yeah I do think some of them were flaky so that sounds about right 👍

PR #5261, opened 2023-01-15
bugfix: moves EDITOR_TO_ON_CHANGE.set(editor, onContextChange) into useEffect 

Cspeisman:
**Description**
With Strict Mode starting in React 18, whenever a component mounts in development, React will simulate immediately unmounting and remounting the component. This change in behavior results in the `EDITOR_TO_ON_CHANGE` map getting set to a `NoOp` function due to the unmount call and does not reset to the correct callback function until the next state change happens within the `Slate` component. 

The fix for this is to put `EDITOR_TO_ON_CHANGE.set(editor, onContextChange)` into `useEffect` so that the correct `onContextChange` callback gets added back into the `EDITOR_TO_ON_CHANGE` map when effects get re-created on a mounted component.

[Link to read more about this change in React 18.](https://reactjs.org/docs/strict-mode.html#ensuring-reusable-state)

**Example**

**This is what the `EDITOR_TO_ON_CHANGE` callbacks look like within the lifecycle before the fix:**

Screenshot 2023-01-15 at 9 14 06 AM

**After the fix:**

Screenshot 2023-01-15 at 9 14 39 AM

**Checks**

### DIFF
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -61,9 +61,9 @@ export const Slate = (props: {
     handleSelectorChange(editor)
   }, [onChange])
 
-  EDITOR_TO_ON_CHANGE.set(editor, onContextChange)
-
   useEffect(() => {
+    EDITOR_TO_ON_CHANGE.set(editor, onContextChange)
+
     return () => {
       EDITOR_TO_ON_CHANGE.set(editor, () => {})
       unmountRef.current = true


dylans:
Thanks @Cspeisman . Seemingly unrelated CI tests are failing, so I'm going to land this.

Cspeisman:
Thanks @dylans, any ideas on what's causing the flakey playwright tests? I saw a handful of PRs were running into the same issue but with different failing tests. I can take a look at some point but was curious if you all had a suspicion of where to start?

Thanks again!

etrepum:
Per #5289 the reason this causes an issue is that the `useEffect` needs to have `onContextChange` in its dependency array. This kind of bug can be avoided with React's eslint plugin https://reactjs.org/docs/hooks-rules.html#eslint-plugin

```diff
diff --git a/packages/slate-react/src/components/slate.tsx b/packages/slate-react/src/components/slate.tsx
index 0a03ed82..4a1b8829 100644
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -68,7 +68,7 @@ export const Slate = (props: {
       EDITOR_TO_ON_CHANGE.set(editor, () => {})
       unmountRef.current = true
     }
-  }, [])
+  }, [onContextChange])
 
   const [isFocused, setIsFocused] = useState(ReactEditor.isFocused(editor))
```

PR #5265, opened 2023-01-20
Improve compatibility for browsers that do not support `ResizeObserver` or `:where` selector

kylemclean:
**Description**
#5206 made Editable components use stylesheets for default styles instead of inline styles. However, it also inadvertently reduced compatibility with browsers that lack support for `ResizeObserver` and the `:where` CSS selector (mainly Safari versions released before September 2020). The result was that browsers that did not support `ResizeObserver` would throw a `ReferenceError` when rendering a placeholder element, and browsers that did not support `:where` would not see default styles for Editable components.

This PR adds a polyfill for `ResizeObserver` and a fallback for browsers that do not support `:where` to fix both of those problems.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/pull/5206#issuecomment-1384361788

**Context**
The latest major browser to add support for `ResizeObserver` is Safari in version 13.1, released March 2020. (https://caniuse.com/resizeobserver) On browsers that do not support it, a `ReferenceError` is thrown when a placeholder element is rendered. This problem is fixed by adding a polyfill for `ResizeObserver`: the npm package [`@juggle/resize-observer`](https://www.npmjs.com/package/@juggle/resize-observer). Note that the native `ResizeObserver` will be used if it exists.

The latest major browser to add support for `:where` is Safari in version 14, released September 2020. (https://caniuse.com/mdn-css_selectors_where) Browsers that do not support `:where` do not parse the default styles for Editable components and placeholder elements, and thus their default styles do not appear.

Default stylesheets using `:where` were introduced in #5206 to give the default styles for Editable components and placeholder elements a specificity of 0, which allows for user-defined stylesheets to override them by simply adding their own class to the Editable component. I do not know of a way to implement this behavior without support for `:where` or CSS layers, so I instead opted to just fix the default styles not being applied.

The fallback for the `:where` selector is implemented by checking support for it using a `@supports` rule. If there is no support, the default styles for Editable components and placeholder elements is set without `:where`. This means that after this change, browsers that do not support `:where` will see default styles, but the specificity of the default styles will be high enough that a user-defined stylesheet with a class for an Editable component will not override the default style. However, inline styles on the Editable component should still override the default styles.

I tested these changes on Firefox 68.9.0 ESR, which supports neither `ResizeObserver` nor `:where`, and both problems appear to be fixed.

Note: it is unclear to me which browser versions Slate intends to support, but [the FAQ](https://docs.slatejs.org/general/faq) only rules out support for IE11, and in the issue linked above at least one person has expressed a desire for Slate to support pre-13.1 Safari.

**Checks**

### DIFF
--- a/.changeset/witty-tips-work.md
+++ b/.changeset/witty-tips-work.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Improve compatibility for browsers that do not support ResizeObserver or :where selector
--- a/packages/slate-react/package.json
+++ b/packages/slate-react/package.json
@@ -14,6 +14,7 @@
     "dist/"
   ],
   "dependencies": {
+    "@juggle/resize-observer": "^3.4.0",
     "@types/is-hotkey": "^0.1.1",
     "@types/lodash": "^4.14.149",
     "direction": "^1.0.3",
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -66,6 +66,7 @@ import {
   NODE_TO_ELEMENT,
   PLACEHOLDER_SYMBOL,
 } from '../utils/weak-maps'
+import { whereIfSupported } from '../utils/where-if-supported'
 import { RestoreDOM } from './restore-dom/restore-dom'
 import { useAndroidInputManager } from '../hooks/android-input-manager/use-android-input-manager'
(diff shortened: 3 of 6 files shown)


dylans (changes requested):
This looks useful/reasonable.

We seem to be failing a few integration tests with the change, so that is worth looking into.

dylans (approved):

PR #5271, opened 2023-01-25
Compare only decorations offsets in MemoizedText. Code highlighting example improvements.

dsvgit:
**Description**
Initially, I tried to improve the Code Highlighting example on the Slate.js site. But while working on this, I didn't find a way to do it without changing the slate-react library code.

The Code Highlighting example improvements:
- Support syntax highlighting for **hierarchical tokens** that Prism.js generates. It is relevant for languages like HTML, JSX, TSX, etc...
- Tokenize the **whole code block at once**, instead of each line separately.
- Support **multiple code blocks** on one page. Show how to **combine paragraphs and code blocks**.
- Keep performance after these changes.
 
Slate-react changes:

- Only for keeping the right performance, I had to add changes to the `slate-react` library. These changes allow returning relative ranges from the `decorate` function. The `basePath` field is added to Point. So a range can have the following structure:
```javascript
const range = { anchor: { basePath: [0, 0], path: [], offset: 0 }, focus: { basePath: [0, 0], path: [], offset: 20 } }
```
- When the function `isDecoratorRangeListEqual` compares ranges via `Range.equals`, these fields aren't included in the comparison. It means that even if the `basePath` is changed, the component isn't re-rendered since other properties are the same.
- It is useful when users add or remove code lines above.
- It is still compatible with old code because `basePath` is optional. And if it isn't set everything works as before.

**Issue**

**Example**

- The example now illustrates multiple code blocks in one editor:

Before: 
Screenshot 2023-01-26 at 00 43 28

After:
Screenshot 2023-01-26 at 00 43 38

- Html, jsx, tsx syntax highlighted the right way:

Before:
Screenshot 2023-01-26 at 00 45 05

After:
Screenshot 2023-01-26 at 00 45 09

- Multiline tokens are supported. When it starts from one line and ends in another:

Before:
Screenshot 2023-01-26 at 00 46 18

After:
Screenshot 2023-01-26 at 00 46 22

- If some pieces affect the other, the other one is re-rendered:

Before:
Screen Recording 2023-01-26 at 00 47 37

After:
Screen Recording 2023-01-26 at 00 48 10

- If new elements are added, and they don't affect the elements below they aren't re-rendered. It is possible because now relative ranges are used on `slate-react` level:

Before:
Screen Recording 2023-01-26 at 00 51 55

After (the same result, leaves aren't re-rendered, performance is kept):
Screen Recording 2023-01-26 at 00 50 39

**Context**
Relative points in decorations are primarily intended to prevent re-renders when elements only move their position in the editor while keeping the structure of internal decorations.

**Checks**


site/examples/code-highlighting.tsx
```diff
@@ -1,237 +1,589 @@
...
+
+      // fill ranges map for decorate function
+      if (types.size) {
+        const key = ReactEditor.findKey(editor, editor.children[index])
+        if (!ranges.has(key.id)) {
+          ranges.set(key.id, [])
+        }
+
+        const anchor: Point = {
+          basePath: [index, 0],
```
dsvgit, line 295: here is an example of using the `basePath` property

site/examples/markdown-preview.tsx
```diff
@@ -1,13 +1,11 @@
 import Prism from 'prismjs'
+import 'prismjs/components/prism-markdown'
```
dsvgit, line 2: just minor refactoring: using prism js markdown implementation

packages/slate-react/src/custom-types.ts
```diff
@@ -10,5 +10,8 @@ declare module 'slate' {
     Range: BaseRange & {
       placeholder?: string
     }
+    Point: BasePoint & {
+      basePath?: Path
```
dsvgit, line 14: add `basePath` field in order to keep the `path` field relative

packages/slate-react/src/hooks/use-children.tsx
```diff
@@ -76,7 +76,7 @@ const useChildren = (props: {
     } else {
       children.push(
         <TextComponent
-          decorations={ds}
+          decorations={ds.map(toAbsoluteRange)}
```
dsvgit, line 79: using absolute ranges only in TextComponent component level

packages/slate-react/src/hooks/use-children.tsx
```diff
@@ -52,7 +52,7 @@ const useChildren = (props: {
     const ds = decorate([n, p])
 
     for (const dec of decorations) {
-      const d = Range.intersection(dec, range)
+      const d = Range.intersection(toAbsoluteRange(dec), range)
```
dsvgit, line 55: absolute ranges have to be calculated on checking decorations intersection as well

dsvgit:
Hey @dylans, great, thanks for taking a look at this

BitPhinix:
@jasonphillips this one might be interesting for you 👀

site/examples/code-highlighting.tsx
```diff
@@ -1,237 +1,589 @@
...
+    const id = ReactEditor.findKey(editor, element).id
+    const ranges = editor.ranges.get(id)
+
+    return ranges ? ranges.slice() : []
+  }, [])
+}
+
+// set editor.intervals with intervals of a continuous stream of identical elements
+// set editor.intervalsStarts with set of these intervals start index (for optimization)
+const ExtractIntervals = () => {
```
bryanph, line 138: could add this as a hook instead perhaps?

site/examples/code-highlighting.tsx
```diff
@@ -1,237 +1,589 @@
...
+  parent: null | PrismTokenNode
+
+  constructor(token: string | Prism.Token, parent: null | PrismTokenNode) {
+    this.token = token
+    this.parent = parent
+  }
+}
+
+// set editor.ranges with decorations to use them in decorate function then
+const ExtractRanges = () => {
```
bryanph, line 189: This one also could be a hook?

site/examples/code-highlighting.tsx
bryanph: So on every editor/selection change this loops over all (top-level only?) editor children. Could you explain what an interval is?

dsvgit: Hey @bryanph. Yes, it's just the contiguous elements intervals calculated on each editor/selection change And it loops only over all top-level editor children. It may have the following data:

```json
{ "paragraph": [[0, 0], [12,14]], "code-line": [[1, 11], [15, 25]] }
```

So the idea here is to extract additional information based on the current editor's children state. It should be always synced with editor children. That's why these "extractors" were placed inside Slate, but before Editable. It means that all components inside Editable can rely on `editor.intervals` as on `editor.children`. I think it actually can be reorganised with hooks in some way. I just wanted to keep it simple.

dsvgit: Actually, it is could be optimized by moving these calculations into

```javascript
useMemo(() => {...; editor.intervals = intervals; }, [editor.children])
```

hook to prevent calling these calculations when only selection is changed

bryanph:
Great work, a well needed change to the code example. I'm not too sure about the `basePath` change, maybe someone who has recently put work into the decorations code can add some input on that.

dsvgit:
@bryanph Thanks. I figured out how to keep the same performance without `basePath` change. I just changed the way how MemoizedText compares decorations. Since only offsets matter for text, I compare only offsets. It looks like just minor changes, that should improve performance but keep the same behaviour as before for other cases.

dsvgit:
Oh, I just clicked multiple times for review request. Didn't think it works this way

packages/slate-react/src/utils/range-list.ts
```diff
@@ -27,13 +37,39 @@ export const isDecoratorRangeListEqual = (
...
+  for (let i = 0; i < list.length; i++) {
+    const range = list[i]
+    const other = another[i]
 
+    // compare only offsets because paths doesn't matter for text
     if (
-      !Range.equals(range, other) ||
-      range[PLACEHOLDER_SYMBOL] !== other[PLACEHOLDER_SYMBOL] ||
-      !shallowCompare(rangeOwnProps, otherOwnProps)
+      range.anchor.offset !== other.anchor.offset ||
```
dsvgit, line 70: compare only offsets for text

packages/slate-react/src/components/text.tsx
```diff
@@ -79,7 +79,7 @@ const MemoizedText = React.memo(Text, (prev, next) => {
     next.isLast === prev.isLast &&
     next.renderLeaf === prev.renderLeaf &&
     next.text === prev.text &&
-    isDecoratorRangeListEqual(next.decorations, prev.decorations)
+    isTextDecorationsEqual(next.decorations, prev.decorations)
```
dsvgit, line 82: use the `isTextDecorationsEqual` function  that compare only offsets

dylans:
@jasonphillips do you have any feedback on this one before I land it?

jasonphillips:
> @bryanph Thanks. I figured out how to keep the same performance without `basePath` change. I just changed the way how MemoizedText compares decorations. Since only offsets matter for text, I compare only offsets. It looks like just minor changes, that should improve performance but keep the same behaviour as before for other cases.

This is a smart optimization! Slate already sends just the intersection of the decoration down into the text node so, as you said, the path shouldn’t matter for comparison. 

BasePath was confusing me a bit and I’m so wary about adding new concepts now that could affect someone’s use case, but this optimization makes perfect sense.

jasonphillips:
But looking at the updated code-highlighting example, I have to say that I'm concerned about having that much complexity in the example. 

Examples are meant to illustrate how to do the basics in slate, and it seems that all this additional logic of finding continuous sequences of elements at the top level of the editor etc is going to be very confusing and counterproductive as a code-highlighting example. Each example should ideally be basic and introduce only one concept, which is the feature being illustrated (in this case, just how to use the decorate function). 

I'm also unsure of why this approach is chosen (continuous lists of top-level elements) instead of using something like a nested container element, eg. why not have structure ` editor.children: [ { type: paragraph, ... }, { type: code-container, language: 'typescript', children: [ { type: code-line, ... }, { type: code-line, ... } ]}, ... ]`  so that the lines of code would be grouped under a parent element, which is where the language selection would be and where the decorate() function would operate. 

One of Slate's great advantages over many other editors is to handle truly nested content, after all; and making the rendering of an element depend on its _siblings_ rather than just parents is very tricky to get right--here, your codeline element depends upon checking whether it is the _first_ code line in sequence (prior sibling is not codeline) in order to decide whether to show the language selection. 

That said, don't get me wrong--we have cases in our own application that I work on where we likewise have to think of siblings and overall structure of the editor children, and your implementation here looks nice. But I'm concerned that this introduces so much complexity as an example, and it brings in some concepts that are not exactly anti-patterns but definitely not a great place for a beginner to start learning how to use Slate the standard way.

dsvgit:
@jasonphillips thank you for the feedback, most points make sense. I need some time to think about it.

dsvgit:
@dylans I created another PR only with `slate-react` changes in order to split code-highlighting example and MemoizedText decorations optimisation. https://github.com/ianstormtaylor/slate/pull/5285

dsvgit:
@jasonphillips , I updated the example, now it should be simpler

- For some cases it is a bit harder to support desired behaviour for nested elements, but with flat structure it's just enabled by default. For the current case I agree that it is better to use nested code block structure. I rewrote this part.

- With current decorations implementation, I didn't find any ways to achieve the right performance for cases when analysing the whole code block required. It works really slow if I use only decorate function, without ranges precalculations (that nodeToDecorations map provides).

- And the last complex thing is the Prism.js integration code. I replaced my code of tokens normalization with util function from `prism-react-renderer` library. It makes nested tokens flat and group them by lines. This is exactly what we need to map them to ranges for Slate.

site/examples/code-highlighting.tsx
```diff
@@ -1,237 +1,516 @@
...
+    const blockEntries = Array.from(
+      Editor.nodes(editor, {
+        at: [],
+        mode: 'highest',
+        match: n => Element.isElement(n) && n.type === CodeBlockType,
+      })
+    )
+
+    const nodeToDecorations = mergeMaps(
+      ...blockEntries.map(getChildNodeToDecorations)
```
dsvgit, line 245: precalculate editor.nodeToDecorations map to use it inside decorate function then
I didn't find a way to achieve good performance without this precalculation

site/examples/code-highlighting.tsx
```diff
@@ -1,237 +1,516 @@
...
+  token: string | Prism.Token
+  parent: null | PrismTokenNode
+
+  constructor(token: string | Prism.Token, parent: null | PrismTokenNode) {
+    this.token = token
+    this.parent = parent
+  }
+}
+
+const getChildNodeToDecorations = ([block, blockPath]: NodeEntry<
```
dsvgit, line 186: this function creates nodeToDecorations map for one `code-block` element

site/examples/code-highlighting.tsx
```diff
@@ -1,237 +1,516 @@
...
       {children}
     </span>
   )
 }
 
-const initialValue: Descendant[] = [
+const useDecorate = (editor: Editor) => {
+  return useCallback(([node, path]) => {
+    if (Element.isElement(node) && node.type === CodeLineType) {
+      const ranges = editor.nodeToDecorations.get(node) || []
```
dsvgit, line 168: just apply precalculated decorations

site/utils/normalize-tokens.ts
```diff
@@ -0,0 +1,117 @@
...
+  return types.concat(add)
+}
+
+// Takes an array of Prism's tokens and groups them by line, turning plain
+// strings into tokens as well. Tokens can become recursive in some cases,
+// which means that their types are concatenated. Plain-string tokens however
+// are always of type "plain".
+// This is not recursive to avoid exceeding the call-stack limit, since it's unclear
+// how nested Prism's tokens can become
+export const normalizeTokens = (
```
dsvgit, line 46: this function needs to apply Prism.js token the right way

dylans (commented):
Looks like a good round of improvements.

GitHub Actions are showing some playwright test failures (shown via files changed).

dsvgit:
@dylans I just fixed playwright tests. If you are ok with this PR we can just close https://github.com/ianstormtaylor/slate/pull/5285 because it contains the same changes.

dylans (approved):

PR #4703, opened 2021-11-30
Add tests for Editor.unhangRange()

e1himself:
**Description**

Add tests for `Editor.unhangRange()`

**Issue**

Related to #3683 

**Example**

n/a

**Context**

`Editor.unhangRange()` has zero tests. So it's really difficult to understand how it should work.
Also any changes to its behavior will not have obvious effect visible in tests expectations.

This PR documents the existing behavior of `Editor.unhangRange()` without modifying it.

**Checks**
- [x] The new code matches the existing patterns and styles.
- [x] The tests pass with `yarn test`.
- [x] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [x] The relevant examples still work. (Run examples with `yarn start`.)
- [x] You've [added a changeset](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md) if changing functionality. (Add one with `yarn changeset add`.)

PR #5275, opened 2023-01-29
Firefox: fix wrong text highlighting with double-click (revived PR)

### LINKED ISSUE #3336: Firefox: wrong selection with double click on a text node

### LINKED ISSUE #3840: Bug: In Firefox bold mark don't work properly after double click on a word
In Firefox, there is a bug with the selector. 

Steps to reproduce: 
1) Select a word and mark it as bold.
2) Click somewhere in the editor to remove the selector from the word.
3) Double click on the bolded word to select it again.

After these steps bold mark isn't detected and a user isn't able to remove bold from the selected word.

bug

Slate: 0.58.4
Browser: Firefox
OS: Mac

After double click on a word user should be able to remove the bold mark.

### LINKED ISSUE #4634: Double clicking on nodes with active marks don't work properly - Firefox
**Description**
double click on a word which has a different mark from the rest of the line, will consider it to be an active mark when testing in firefox.

**Recording** 
bold

**Sandbox**
[Test it here in the example](https://www.slatejs.org/examples/richtext)  

**Steps**
To reproduce the behavior:
1. Go to sandbox
2. Double click on a word
3. Mark the word with bold
4. select any normal text, it should let the mark inactive (that works 👍)
5. then double click again on the first word we used, the mark we used should be active (doesn't work)
6. if you clicked on the mark again, nothing changes (an error)

**Environment**
- Slate Version: [e.g. 0.62.0]
- Operating System: [e.g. Mac OS]
- Browser: [e.g. FireFox]

12joan:
**Description**
This is a revival of PR #4908, which itself is a revival of #3374, both of which are currently inactive.

This PR fixes the bug in Firefox where double-clicking a word creates a selection whose anchor is inside the preceeding text node. (This is similar to the concept of a "hanging range", but not an example of it as per #4852.)

**Issue**
Fixes #3336
Fixes #3840
Fixes #4634
Fixes [#1987](https://github.com/udecode/plate/issues/1987) (Plate)
Fixes [#1897](https://github.com/udecode/plate/issues/1897) (Plate)

**Example**
From #4908:

Example

**Context**
PR #4908 had two requested changes (left by @BitPhinix) which were never addressed.

1. [Relating to type signatures](https://github.com/ianstormtaylor/slate/pull/4908#discussion_r857711656)

I've fixed this and the resulting type errors. If anyone more experienced with TypeScript than myself knows a more elegant way of handling the casts on [lines 806 and 811](https://github.com/12joan/slate/blob/firefox_selection/packages/slate-react/src/plugin/react-editor.ts#L806), please let me know.

2. ["Maybe we should replace this with a `Editor.hangRange` helper"](https://github.com/ianstormtaylor/slate/pull/4908#discussion_r857713507)

There's still no clear concensus on what `Editor.unhangRange` should and should not be responsible for. In the comments on #4703, @e1himself suggested that it should be modified to support text nodes and cases where the starting point of the range needs adjusting.

Adding a separate `Editor.hangRange` helper would probably confuse the terminology even further. (If to "unhang" a range is to move the end point inside some text node, then to "hang" it should probably mean to reverse that, which isn't the intended behaviour here.)

My suggestion would be to first add support for text nodes to `Editor.unhangRange`, perhaps conditional on a `texts: boolean` option, and second to add a `side: 'start' | 'end' | 'both'` option, defaulting to `'end'` if compatibility is a concern, controlling whether the start or end of the range is adjusted.

This should certainly happen in a separate PR, however. For now, I'd suggest we leave the logic in `toSlateRange` as it is unless anyone can spot a flaw in it.

**Checks**

### DIFF
--- a/.changeset/thirty-starfishes-knock.md
+++ b/.changeset/thirty-starfishes-knock.md
@@ -1,0 +1,5 @@
+---
+'slate-react': major
+---
+
+Firefox: fix wrong text highlighting with double-click
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -535,9 +535,9 @@ export const ReactEditor = {
     domPoint: DOMPoint,
     options: {
       exactMatch: boolean
-      suppressThrow: boolean
+      suppressThrow: T
     }
-  ): Point | null {
+  ): T extends true ? Point | null : Point {
     const { exactMatch, suppressThrow } = options
     const [nearestNode, nearestOffset] = exactMatch
       ? domPoint
@@ -703,7 +703,7 @@ export const ReactEditor = {
     editor: ReactEditor,
     domRange: DOMRange | DOMStaticRange | DOMSelection,
     options: {
-      exactMatch: T
+      exactMatch: boolean
       suppressThrow: T
     }
   ): T extends true ? Range | null : Range {
@@ -797,18 +797,18 @@ export const ReactEditor = {
      */
 
     if (IS_FIREFOX && !isCollapsed && anchorNode !== focusNode) {
-      const isEnd = Editor.isEnd(editor, anchor, anchor.path)
-      const isStart = Editor.isStart(editor, focus, focus.path)
+      const isEnd = Editor.isEnd(editor, anchor!, anchor.path)
+      const isStart = Editor.isStart(editor, focus!, focus.path)
 
       if (isEnd) {
         const after = Editor.after(editor, anchor as Point)
         // Editor.after() might return undefined
-        anchor = after || anchor
+        anchor = (after || anchor!) as T extends true ? Point | null : Point
       }
 
       if (isStart) {
         const before = Editor.before(editor, focus as Point)
-        focus = before || focus
+        focus = (before || focus!) as T extends true ? Point | null : Point
       }
     }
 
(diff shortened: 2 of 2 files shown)


zbeyens (approved):

dylans (approved):

dylans:
Thanks @12joan . Agree on fixing the smaller issue here, and revisiting the API changes/flexibility separately. If you have interest let me know.

dylans:
GitHub actions is not cooperating tonight, will see if this is a momentary glitch or if something is wrong with our release script tomorrow.

```[HttpError]: You have exceeded a secondary rate limit. Please wait a few minutes before you try again.
    at /home/runner/work/_actions/changesets/action/v1/dist/index.js:192:1285
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  status: 403,
  response: {```

12joan:
https://github.com/ianstormtaylor/slate/actions/runs/4040806744/jobs/6946963489#step:6:68

Looks like a rate limit on the GitHub search API. Re-running the workflow should probably fix it?

zbeyens:
Happens in plate too.

PR #5285, opened 2023-02-08
Perf: Compare only offsets for MemoizedText decorations

dsvgit:
**Description**
Compare only offsets for MemoizedText decorations. It helps reduce text elements rerenders count when decorations paths are changed and offsets are the same. We can do that because on the Text component level decorations paths are never used and only offsets are taken into account.

**Issue**
Initial PR: https://github.com/ianstormtaylor/slate/pull/5271

**Example**

**Context**

**Checks**

### DIFF
--- a/.changeset/dull-eels-hammer.md
+++ b/.changeset/dull-eels-hammer.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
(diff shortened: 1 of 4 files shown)


dsvgit:
Close this PR because https://github.com/ianstormtaylor/slate/pull/5271 contains these changes

Issue #5222, opened 2022-12-13
Increase number of max normalization iterations


Jacfem:
**Problem**

The amount of max normalization iterations isn't enough for some large content in my app. It is set [here](https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/interfaces/editor.ts#L1060) as the number of dirty paths multiplied by 42. There's a comment next to the code that says `//HACK: better way?`. I'd love to learn more if bumping this multiplier is okay.

**Solution**

Either allow that `42` number to be configurable by the user, or increase that number. Locally, I've increased it to 100 and it fixes our use cases.

**Alternatives**
Open to suggestions!

**Context**
PR will be incoming.

I am curious if there are risks or concerns associated with this increase. I've applied the fix locally on my app and it fixes the issues. Thank you for your input!

zbeyens:
I think that error is generally dangerous for apps in production. We should add an editor method to override that behavior: `editor.getMaxNormalizationIterations(dirtyPaths)` so we choose to reuse `dirtyPaths`

PR #5295, opened 2023-02-15
More control on `editor.normalizeNode`

zbeyens:
**Description**

_not sure if the coding style fits as it looks to be the first time we have many parameters in an editor method, any suggestion is welcome_

Non-breaking changes:

- new `editor` method that can be overridden to control when the normalization should stop. Default behavior (unchanged) is to throw an error when it iterates over 42 times the dirty paths length.
```ts
shouldNormalize: ({
  iteration,
  dirtyPaths,
  operation,
}: {
  iteration: number
  dirtyPaths: Path[]
  operation?: Operation
}) => boolean
```
- `editor.onChange` signature change: `(options?: { operation?: Operation }) => void` where `operation` is triggering the function.
- `editor.normalizeNode` signature change: `(entry: NodeEntry, options?: { operation?: Operation }) => void` where `operation` is triggering the function.
- `EditorNormalizeOptions` new option `operation?: Operation` where `operation` is triggering the function.

**Issue**
Fixes: #5222

**Checks**

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -17,28 +17,28 @@ import {
   Text,
 } from '..'
 import {
-  DIRTY_PATHS,
+  getCharacterDistance,
+  getWordDistance,
+  splitByCharacterDistance,
+} from '../utils/string'
+import {
   DIRTY_PATH_KEYS,
+  DIRTY_PATHS,
   NORMALIZING,
   PATH_REFS,
   POINT_REFS,
   RANGE_REFS,
 } from '../utils/weak-maps'
-import {
-  getWordDistance,
-  getCharacterDistance,
-  splitByCharacterDistance,
-} from '../utils/string'
-import { Descendant } from './node'
 import { Element } from './element'
+import { Descendant } from './node'
 import {
   LeafEdge,
+  MaximizeMode,
+  RangeDirection,
   SelectionMode,
   TextDirection,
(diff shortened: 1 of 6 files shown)


dylans (commented):
Looks useful to me. Leaving this open for a few days to give others time to review and/or give feedback.

packages/slate/src/interfaces/editor.ts
```diff
@@ -1057,29 +1067,34 @@ export const Editor: EditorInterface = {
...
           if (Element.isElement(node) && node.children.length === 0) {
-            editor.normalizeNode(entry)
+            editor.normalizeNode(entry, { operation })
           }
         }
       }
 
-      const max = getDirtyPaths(editor).length * 42 // HACK: better way?
-      let m = 0
+      const dirtyPaths = getDirtyPaths(editor)
```
xhonato-rellify, line 1075: @zbeyens This constant is unused.

xhonato, line 1075: @zbeyens This constant is unused.

dylans (approved):

delijah:
@dylans @zbeyens Was this PR doiing any good? I've just tried to use the operation from within normalizeNode and it looks like this is always undefined since most transform operations are running withing `withoutNormalizing`...

zbeyens:
You're right. This option was experimental and can be removed since no use-case has been built upon it.

PR #5297, opened 2023-02-17
fix: remove memory leaks in Text and Editable

edhager:
**Description**
It was discovered that slate-react was leaving behind detached spans and divs as the edited text was being updated causing  memory leaks.

The cause of the leaks were in the weak maps used in `packages/slate-react/src/components/editable.tsx` and `packages/slate-react/src/components/text.tsx`.  DOM nodes stored in the maps were detached but the keys still had strong references.

This PR adds callback refs to remove DOM nodes from the maps when the refs are set to null.

**Checks**

### DIFF
--- a/.changeset/grumpy-pianos-sit.md
+++ b/.changeset/grumpy-pianos-sit.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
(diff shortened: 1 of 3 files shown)


dylans (commented):
Looks good, tests pass, etc.

We'll need to add a changeset for this when you get a chance (see instructions).

Issue #5183, opened 2022-11-07
Selecting an inline void element in Android platform doesn't evoke the keyboard


YaoKaiLun:
**Description**
Selecting an inline void element in Android platform doesn't evoke the keyboard, but works at iOS.

**Recording**
Android:

https://user-images.githubusercontent.com/11460856/200246609-869faf71-9984-436e-a1df-e997a6d13488.mp4

iOS

https://user-images.githubusercontent.com/11460856/200246704-a683ec3f-f88d-4a23-a75c-84a5516dffed.mp4

**Sandbox**
[slate mentions example](https://www.slatejs.org/examples/mentions)

**Steps**
To reproduce the behavior:
1. Click on a mention element

**Expectation**
Click on a mention element in Android platform the keyboard should evoke as iOS does.

**Environment**
- Slate Version: [e.g. 0.83.2]
- Android 9, Edge;   iOS 16.0 safari

u373693041:
Is there any solution to this problem? When the selection is on the inline void, it can also cause the keyboard to collapse

PR #5298, opened 2023-02-18
fix selecting mention element in android, can't evoke the keyboard

YaoKaiLun:
**Description**
In android, once selecting an inline void element, the keyboard doesn't be evoked, unless we set it's contenteditable to be true, like below

image

But it will make the children (a zero width node) to be contenteditable. So we should disable it's inputting at the beforeinput event.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/5183

**Example**

before:

https://user-images.githubusercontent.com/11460856/219867021-ca1b3937-3fc1-4986-8c4d-1744e567d846.mp4

after:

https://user-images.githubusercontent.com/11460856/219867027-4848f8dd-bc28-4c69-bb57-c189f3fb7158.mp4

**Checks**

### DIFF
--- a/packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
+++ b/packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
@@ -352,12 +352,25 @@ export function createAndroidInputManager({
       insertPositionHint = false
     }
 
+    let inVoidNode = false
     let [nativeTargetRange] = (event as any).getTargetRanges()
     if (nativeTargetRange) {
       targetRange = ReactEditor.toSlateRange(editor, nativeTargetRange, {
         exactMatch: false,
         suppressThrow: true,
       })
+
+      if (
+        nativeTargetRange.collapsed &&
+        nativeTargetRange.startContainer.nodeType === globalThis.Node.TEXT_NODE
+      ) {
+        const closestElement = nativeTargetRange.startContainer.parentElement.closest(
+          '[data-slate-node=element]'
+        )
+        inVoidNode = closestElement?.attributes.hasOwnProperty(
+          'data-slate-void'
+        )
+      }
     }
 
     // COMPAT: SelectionChange event is fired after the action is performed, so we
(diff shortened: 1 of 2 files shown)


YaoKaiLun:
@BitPhinix please review!

BitPhinix:
Not a huge fan of this approach. Having the nodes contenteditable will likely mess with autocomplete and unexpected behaviour especially on more complex void nodes. There are other hacks to prevent the keyboard from closing when selecting inline void nodes, like adding pseudo elements with (zero-width) spaces around them via css.

YaoKaiLun:
Allright, I'd try the pseudo elements method.

clauderic:
@BitPhinix can you provide an example of the approach you had in mind? I played with a few different variations using `::before` / `::after` pseudo elements with zero-width space content on void inline nodes, to no avail.

bryanph:
I use the following component for fixing this in general for inlines, not sure if this resolves your issue on android though. You use it by placing this before and after `children` in your inline component.

```
// Put this at the start and end of an inline component to work around this Chromium bug:
// https://bugs.chromium.org/p/chromium/issues/detail?id=1249405
// Necessary when the component has no padding on the sides
export const InlineChromiumBugfix = () => (
  <span
    contentEditable={false}
    css={css`
      font-size: 0;
    `}
  >
    ${String.fromCodePoint(160) /* Non-breaking space */}
  </span>
)
```

clauderic:
@bryanph I tried different variations of your suggestion on Android but haven't had any success so far. The only solution I've found that works so far is the one described in this PR, but I would love if we could find a solution that doesn't involve consumers having to handle Android-specific edge cases in the markup of their inline void nodes.

clauderic:
After a bit more digging I found the issue, it's because the zero-width spaces aren't rendered on Android: https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/string.tsx#L128

Removing this android specific condition solves the issue, but introduces other regressions, so I will look into those.

dylans:

Issue #5289, opened 2023-02-09
slate-react doesn't use most up-to-date onChange callback


endercrest:
**Description**
slate-react `Slate` component does not use the most up-to-date `onChange` callback provided to the component.

It looks like https://github.com/ianstormtaylor/slate/pull/5261 introduced this bug. It does not update the onChange callback so it is always utilizing the first time provided onChange value.

**Recording**

https://user-images.githubusercontent.com/2616075/217894479-329ec616-a837-40b6-9cab-b8b47545be8e.mov

**Sandbox**
https://codesandbox.io/s/onchange-incorrectly-updates-0-89-0-or-above-bqmcu1?file=/index.js

**Steps**
To reproduce the behavior:
1. Type into the editor in sandbox above. The console will print out 0 as that is the initial value of the counter.
2. Increment the counter using the button above the editor.
3. Type into the editor again and observe that the counter is still printing the initial counter value.

**Expectation**
onChange callback should be using the latest onChange prop value instead of the first-time provided value.

Going down to 0.88.2 of slate-react fixes the problem.

**Environment**
- Slate Version: slate-react 0.89.0 or above
- Operating System: MacOS Ventura
- Browser: Chrome

kaya-altin:
I switched back to version 0.88.2 for this reason. I am eagerly waiting for it to be fixed. :)

shannonjarcher:
I've encountered this issue as well.

PR #5310, opened 2023-02-23
fix: Add missing dependencies to useEffect and React.memo calls (#5289)

etrepum:
**Description**
The merge of #5261 caused a regression where the first value of onChange is cached indefinitely because it is not used as a dependency of the effect that is supposed to maintain it. This fixes that bug, and similar bugs in the React.memo comparisons for MemoizedElement and MemoizedText where a cached `renderPlaceholder` is passed down to the leaf.

**Issue**
Fixes: #5289

**Example**
See the recording in #5289 for an example of the bug. A similar one could be crafted with renderPlaceholder being cached if necessary, but it should be pretty clear from the code.

**Context**
There aren't really other ways to fix these issues in a localized manner. I did look into adding the eslint plugin that can help check hook dependencies but there are violations in the code that would require distracting workarounds or API changes so I opted to keep it concise.

**Checks**

### DIFF
--- a/.changeset/pretty-buses-stare.md
+++ b/.changeset/pretty-buses-stare.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fix to ensure that the latest versions of onChange and renderPlaceholder are used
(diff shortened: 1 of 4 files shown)


kaya-altin:
Hello, after then this version, there is a problem of not being able to focus on the first click. I think it's related to the placeholder, the issue doesn't occur when the placeholder is removed. @etrepum 

https://user-images.githubusercontent.com/4608228/222655358-5f54f5aa-2c64-4869-8015-50f5eaedc0e5.mov

etrepum:
@kaya-altin the only thing this PR can do is cause components to render more often. I think you are seeing a bug related to one of the other changes that were merged since the last version you tested.

PR #5311, opened 2023-02-23
fix: `normalizeNode`

zbeyens:
**Description**
The maximum number of iterations should be computed before the while loop, not inside.

**Checks**

### DIFF
--- a/packages/slate/src/interfaces/editor.ts
+++ b/packages/slate/src/interfaces/editor.ts
@@ -85,6 +85,7 @@ export interface BaseEditor {
     operation,
   }: {
     iteration: number
+    initialDirtyPathsLength: number
     dirtyPaths: Path[]
     operation?: Operation
   }) => boolean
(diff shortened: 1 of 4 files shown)


dylans (approved):

packages/slate/src/interfaces/editor.ts
```diff
@@ -1072,15 +1073,15 @@ export const Editor: EditorInterface = {
...
-      const dirtyPaths = getDirtyPaths(editor)
-
+      const initialDirtyPathsLength = getDirtyPaths(editor).length
       let iteration = 0
 
       while (getDirtyPaths(editor).length !== 0) {
         if (
           !editor.shouldNormalize({
-            iteration,
             dirtyPaths: getDirtyPaths(editor),
```
rcbevans, line 1082: Not sure what the performance overhead of `getDIrtyPaths` is, but it's not used by `shouldNormalize` and passing it here means it's done twice per iteration.

PR #5318, opened 2023-02-27
Implement type predicate for `operationCanTransformPath`

delijah:
**Description**
This pull request implements a type predicate for the function `operationCanTransformPath` which lives within the `Path` interface.
We have figured, it helps us in writing cleaner code if the function would use a type predicate.

**Example**
```
if (Path.operationCanTransformPath(operation)) {
    // ts would complain here without the type predicate
    console.log(operation.path);
}
```

**Checks**


delijah:
Shall we merge this? 😇

Issue #5319, opened 2023-02-27
uncaught exception when unmounting Slate Component


ThibautMarechal:
**Description**
When unmunting the Slate component directly after mounting it, (In my case it was the"key" props that change too quickly) some functions was still running after the component was unmounted. Raising uncaught execption.

**Recording**
![image](https://user-images.githubusercontent.com/20936586/221565623-126f952c-bdd4-4d0a-b2b2-e4fb467c3007.png)

**Sandbox**
- 

**Steps**
Change the key props quickly

**Expectation**
No uncaught exception raised.

**Environment**
- Slate Version: since 0.91.2
- Operating System: Windows
- Browser: Chrome

**Context**
-

PR #5320, opened 2023-02-27
Bugfix/cancel debounced and throttled function when unmounting component

ThibautMarechal:
**Description**
When unmunting the Slate component directly after mounting it, (In my case it was the"key" props that change too quickly) some functions was still running after the component was unmounted. Raising uncaught execption.

**Issue**
Fixes: [(link to issue)](https://github.com/ianstormtaylor/slate/issues/5319)

**Example**
image

**Context**

**Checks**

### DIFF
--- a/.changeset/cancel-throttled-and-debounced-functions-on-unmount.md
+++ b/.changeset/cancel-throttled-and-debounced-functions-on-unmount.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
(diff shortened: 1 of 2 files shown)


dylans:
Thanks for the PR. I saw #5322 first so this has already landed and been released.

ThibautMarechal:
Oh I didn't saw that ! Thanks !

Issue #5192, opened 2022-11-10
[Android Cursor Issue]The cursor automatically move while deleting text aside an inline embed void Element 


qiooi:
**Description**
Deleting the element's front text to empty,Cursor will be moved to the back of it,Deleting the back text to empty,Cursor will be moved to the front of it.

**Recording**
![动画](https://user-images.githubusercontent.com/69415032/201137252-0853d1aa-4cce-4150-a48e-cdb6a843f369.gif)

**Steps**
To reproduce the behavior:
1. Change to Android mode on the Chrome Console
2. Go to 'https://www.slatejs.org/examples/mentions'
3. Delete text

**Environment**
- user-agent: [Mozilla/5.0 (Linux; Android 8.0.0; SM-G955U Build/R16NW) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/87.0.4280.141 Mobile Safari/537.36]

clauderic:
I can confirm this is still an issue and I can consistently reproduce it 👍 Will try to spend a bit of time looking into it.

PR #5325, opened 2023-02-28
[Android] Fix edge-cases when text leaf nodes are deleted

clauderic:
#### Description
This PR fixes edge-cases in the Android input manager when text leafs are deleted.

#### Issue
Fixes: 
- https://github.com/ianstormtaylor/slate/issues/5192

**Context**
Previously, text leaf deletions were stored in pending diffs and only reconciled when flushing as a result of a selection change or an action (such as inserting a line break). This could result in a number of different edge cases, such as the browser not knowing where to place the selection after deleting a text node before or after a void node.

In order to fix these edge cases, we detect when an input event would result in a text leaf being deleted, and schedule an action to allow Slate to reconcile its state with the DOM and apply normalization fixes (such as inserting zero-width spacers next to inline void nodes) rather than storing these events as deferred text diffs.

##### Before

Before

https://user-images.githubusercontent.com/1416436/221979804-5ebfc519-f87d-401c-b03c-ab54129677ac.mov

##### After

https://user-images.githubusercontent.com/1416436/221979578-dbe437b5-73bc-4172-93b1-1d91d47376a3.mov

##### Note

There is a separate issue for the keyboard being closed when the selection moves right before void nodes, this PR does not solve that separate issue:

- https://github.com/ianstormtaylor/slate/issues/5183
- https://github.com/ianstormtaylor/slate/pull/5298

If we apply the fix described in the PR linked above, we can see what the final desired behaviour would eventually look like once we find a solution to https://github.com/ianstormtaylor/slate/issues/5183:

https://user-images.githubusercontent.com/1416436/221980680-73b9be67-69dd-4330-8b03-6e8668a2b26c.mov

#### Checks

### DIFF
--- a/packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
+++ b/packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
@@ -2,6 +2,7 @@ import { DebouncedFunc } from 'lodash'
 import { Editor, Node, Path, Point, Range, Text, Transforms } from 'slate'
 import { ReactEditor } from '../../plugin/react-editor'
 import {
+  applyStringDiff,
   mergeStringDiffs,
   normalizePoint,
   normalizeRange,
@@ -153,7 +154,7 @@ export function createAndroidInputManager({
       EDITOR_TO_PENDING_DIFFS.get(editor)
     )
 
-    let scheduleSelectionChange = !!EDITOR_TO_PENDING_DIFFS.get(editor)?.length
+    let scheduleSelectionChange = hasPendingDiffs()
 
     let diff: TextDiff | undefined
     while ((diff = EDITOR_TO_PENDING_DIFFS.get(editor)?.[0])) {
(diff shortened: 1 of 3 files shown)


packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
```diff
@@ -153,7 +154,7 @@ export function createAndroidInputManager({
       EDITOR_TO_PENDING_DIFFS.get(editor)
     )
 
-    let scheduleSelectionChange = !!EDITOR_TO_PENDING_DIFFS.get(editor)?.length
+    let scheduleSelectionChange = hasPendingDiffs()
```
clauderic, line 157: Unrelated to this PR, but thought I would clean this up while I was here, `hasPendingDiffs()` is equivalent to `!!EDITOR_TO_PENDING_DIFFS.get(editor)?.length`

packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
```diff
@@ -377,38 +378,76 @@ export function createAndroidInputManager({
...
+    // defer flushing them at a later point in time. We don't want to flush
+    // for every input event as this can be expensive. However, there are some
+    // scenarios where we cannot safely store the text diff and must instead
+    // schedule an action to let Slate normalize the editor state.
+    let canStoreDiff = true
 
-    if (Range.isExpanded(targetRange) && type.startsWith('delete')) {
-      if (Path.equals(targetRange.anchor.path, targetRange.focus.path)) {
+    if (type.startsWith('delete')) {
+      if (Range.isExpanded(targetRange)) {
```
clauderic, line 389: Just re-structuring the code paths here to avoid checking `type.startsWith('delete')` multiple times

packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
```diff
@@ -377,38 +378,75 @@ export function createAndroidInputManager({
...
-      const [start, end] = Range.edges(targetRange)
-      const leaf = Node.leaf(editor, start.path)
-
-      if (leaf.text.length === start.offset && end.offset === 0) {
-        const next = Editor.next(editor, { at: start.path, match: Text.isText })
-        if (next && Path.equals(next[1], end.path)) {
-          targetRange = { anchor: end, focus: end }
-        }
-      }
-    }
```
clauderic, line 390: This code block hasn't been deleted, it is just relocated on line 393

PR #5326, opened 2023-03-01
Fix local dev

juliankrispel:
**Description**
Currently I can't run the slate examples locally because more than one versions of react are installed at the same time. This pr fixes that by configuring "resolutions" in the workspace package.json

To be clear - this doesn't affect slate the library - just local dev

**Issue**
Fixes local dev environment

**Example**
https://www.loom.com/share/d701fcb9e4434e6699d32a2a2e7fdc4d


.nvmrc
```diff
@@ -0,0 +1 @@
+18
```
juliankrispel, line 1: added .nvmrc because it's been bugging me that I can't automatically switch to the version of node that's required to build thisi project

site/examples/code-highlighting.tsx
```diff
@@ -241,7 +241,7 @@ const SetNodeToDecorations = () => {
 
 const useOnKeydown = (editor: Editor) => {
   const onKeyDown: React.KeyboardEventHandler = useCallback(e => {
-    if (isHotkey('tab', e)) {
+    if (isHotkey('tab', e as any)) {
```
juliankrispel, line 244: Without this a type error happens when building

dylans (commented):
change looks fine/simple enough. Tests are failing so we need to track down the cause.

package.json
```diff
@@ -123,5 +123,9 @@
       "eslint --fix"
     ]
   },
+  "resolutions": {
```
ryanmitts, line 126: This shouldn't be needed? If you upgrade `react-error-boundary` too, then there is no conflict I think.

site/examples/code-highlighting.tsx
ryanmitts: This can be prevented I think by upgrading the `@types` packages for `react` and `react-dom`.

ryanmitts (changes requested):

package.json
ryanmitts: Looks like upgrading that library prevents the resolution, but the API on `react-error-boundary` has changed in the newer version.

The /site/pages/examples/[example].tsx can be changed to something like this to get it working.

```
const ExamplePage = ({ example }: { example: string }) => {
  const [showTabs, setShowTabs] = useState<boolean>()
  const EXAMPLE = EXAMPLES.find(e => e[2] === example)
  const [name, Component, path] = EXAMPLE
  return (
    <ErrorBoundary
      fallbackRender={({ error }) => {
        return (<Warning>
          <p>An error was thrown by one of the example's React components!</p>
          <pre>
            <code>
              <>
                {error}
                {'\n'}
                {error.stack}
              </>
            </code>
          </pre>
        </Warning>)
      }}
    >
      <div>
        <Head>
          <title>Slate Examples</title>
          <meta name="viewport" content="width=device-width, initial-scale=1" />
          <link rel="icon" href="/favicon.ico" />
          <link rel="stylesheet" href="/index.css" />
          <link
            rel="stylesheet"
            href="https://fonts.googleapis.com/css?family=Roboto:400,400i,700,700i&subset=latin-ext"
          />
          <link
            rel="stylesheet"
            href="https://fonts.googleapis.com/icon?family=Material+Icons"
          />
        </Head>
        <Header>
          <Title>Slate Examples</Title>
          <LinkList>
            <A href="https://github.com/ianstormtaylor/slate">GitHub</A>
            <A href="https://docs.slatejs.org/">Docs</A>
          </LinkList>
        </Header>
        <ExampleHeader>
          <TabButton
            onClick={e => {
              e.stopPropagation()
              setShowTabs(!showTabs)
            }}
          >
            <Icon>menu</Icon>
          </TabButton>
          <ExampleTitle>
            {name}
            <A
              href={`https://github.com/ianstormtaylor/slate/blob/main/site/examples/${path}.tsx`}
            >
              (View Source)
            </A>
          </ExampleTitle>
        </ExampleHeader>
        <TabList isVisible={showTabs}>
          {EXAMPLES.map(([n, , p]) => (
            <Link
              key={p as string}
              href="/examples/[example]"
              as={`/examples/${p}`}
              passHref
            >
              <Tab onClick={() => setShowTabs(false)}>{n}</Tab>
            </Link>
          ))}
        </TabList>
        <ExampleContent>
          <Component />
        </ExampleContent>
        <TabListUnderlay
          isVisible={showTabs}
          onClick={() => setShowTabs(false)}
        />
      </div>
    </ErrorBoundary>
  )
}
```

dylans:
I think this is stale and was resolved with the recent upgrade to React 18.

PR #5330, opened 2023-03-02
[Draft] Avoid `TextString` unmounting when rendering zero-width strings and line breaks

clauderic:
**Description**
This is a draft PR to refactor how zero-width strings and line breaks are rendered in the leaf `<String />` component. Currently, we conditionally render either the `<TextString>` component, or the `<ZeroWidthString>` component, which both ultimately render `<span>` elements to the DOM.

When we switch from rendering a `<TextString>` component to a `<ZeroWidthString>` component, React needs to first unmount the `<TextString>` component, which removes the text `<span>` from the DOM, and insert a new span when the `<ZeroWidthString>` component mounts.

This is wasteful from a performance point of view, and can also interrupt composition events on Android.

**Checks**


dylans:

PR #5354, opened 2023-03-10
Fix/suggested text closes keyboard

edhager:
As we all know, Android devices deal with content-editable divs in undesired ways.  There is a bug in `Editable` that causes the keyboard to close when you delete all of the text.  The same thing happens if you do the following while using the Samsung keyboard using one of the example pages:

1. Delete all of the text.
2. Type a word like "Engineering".
3. Delete some of that word leaving "Enginee".
4.  Select "Engineering" from the keyboard's suggested words.

The keyboard will hide.  This is because the Samsung keyboard is generating two input events.  First it deletes the "Enginee" and then it inserts "Engineering". Because of the keyboard hiding bug described above, the keyboard hides when you follow the test steps above as well.

When all of the text is deleted, the placeholder text is inserted into the editor as a span with `contentEditable` set to false.  Given that the only thing contained in the editable <div> is a <span> that can't be edited, the keyboard is hidden.

The goal of this PR is to move the placeholder text outside of the editable <div> to avoid the keyboard hiding problem.

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -29,7 +29,6 @@ import { useSlate } from '../hooks/use-slate'
 import { TRIPLE_CLICK } from '../utils/constants'
 import {
   DOMElement,
-  DOMNode,
   DOMRange,
   DOMText,
   getDefaultView,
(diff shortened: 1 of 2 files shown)


edhager:
This PR is still in draft.  I have enough working to show that moving the placeholder outside of the editable div works.  

I still have some clean up to do because there is code in `Leaf` dealing with placeholders that no longer needs to be there.  I also have a problem with the placeholder causing horizontal scrollbars to appear in some browsers.

edhager:
Closing because I don't like the side effects of this fix.  With the placeholder outside of the editable <div>, the placeholder does not participate in any custom styling that is applied to the contents.

PR #5361, opened 2023-03-15
test: remove useless element test case

ahungrynoob:
**Description**
Since `Node.elements` doesn't support `range` and `path` options, so these two test cases are no need.

**Checks**


dylans:

Issue #4998, opened 2022-05-19
Safari paste handling and Clipboard API


dylans:
**Description**

A recent Safari update seems to have blocked all of Slate's attempts to append our syntax tree to `application/x-slate-fragment`. Previously this worked but [would exclude files](https://github.com/ianstormtaylor/slate/issues/4465).

**Sandbox**/**Steps**

Any slate example shows an empty `application/x-slate-fragment` when pasting.

**Expectation**

We can control what gets pasted. Currently we're falling back to HTML which isn't great for void elements that have behavior not contained in html.

**Environment**
- Slate Version: Any
- Operating System: macOS, probably iOS
- Browser: Safari

**Context**

We probably need to switch to the newer clipboard api for Safari, though it's annoying in that it asks for user approval. See various links in https://github.com/ianstormtaylor/slate/issues/4465#issuecomment-904567668

IMSupperkaka:
same problem

jdmanabat:
Currently, we are having an issue when pasting bullet and number lists in Safari. Do you think it's related to this?

dylans:
> Currently, we are having an issue when pasting bullet and number lists in Safari. Do you think it's related to this?

What type of issue?

jdmanabat:
> > Currently, we are having an issue when pasting bullet and number lists in Safari. Do you think it's related to this?
> 
> What type of issue?

When pasting the copied bullet or numbered lists the HTML tags are removed.

Nikhil22:
When I paste plain text with line breaks in Safari, Slate does not honor the line breaks. End result is all one block of text.

dylans:
Safari has progressively made the old copy/paste API less usable, a strong to push to use the newer API (which requires user consent). We need to implement that in 2023 to solve a number of problems with copy/paste in Safari.

Nikhil22:
@dylans Sounds good. If you or the team need any help, I'm happy to contribute in any way I can.

12joan:
The Clipboard API doesn't require prompting for permission when writing to the clipboard from an event handler, only when reading it outside of a paste event. However, it won't work for copying `application/x-slate-fragment` data in Safari, since all major browsers only allow a limited set of data types when writing with `Clipbard.write()`. See https://webkit.org/blog/10855/.

I don't think there are any APIs that allow us to copy `application/x-slate-fragment` in Safari.

PR #5362, opened 2023-03-15
Directly paste ClipboardEvents in Safari

jason0x43:
**Description**
This PR changes updates the paste handling code to always use ClipboardEvents in Safari rather than preferring InputEvents as InputEvents won't include files or data items with custom mime types.

**Issue**
Fixes: #4998

**Example**
Log the `types` of the data property passed to `ReactEditor.insertData` when pasting and verify that with this PR they include `application/x-slate-fragment`, and without this PR they don't.

Ideally, if a document that has problems copying and pasting in Safari is available, such as one containing inline images, verify that content that normally has problems pastes properly with this PR.

**Context**
Pasting generates two events in Safari: a ClipboardEvent and an InputEvent. Currently, Slate generally ignores ClipboardEvents in Safari in favor of InputEvents, which flow through Slate's `onDOMBeforeInputHandler`. This is problematic because InputEvents generated by pasting in Safari are missing some key pieces of data: they won't include files or an `application/x-slate-fragment` data item. ClipboardEvents do have this data.

**Checks**


dylans (changes requested):
Wow, such a simple fix.

Could you please add a changeset (patch change level) and then I'll land it. Thanks!

jason0x43:
Changeset added

dylans (approved):

zbeyens:
Awesome, I think that deserves a playwright test.

PR #5363, opened 2023-03-16
enable eslint hooks rules

### LINKED ISSUE #3886: Runtime change of editor causes a crash

I am reporting a bug.

https://codesandbox.io/s/slate-reproductions-forked-er268?file=/index.tsx

To reproduce the issue click on Add Editor twice. The first editor is recreated when clicking on the button.
Slate: 0.58.0
Browser: This issue is not browser specific and is reproducible at least in Chromium Version 85.0.4183.102 (Official Build) Arch Linux (64-bit) and Firefox 80.0.1 (64-bit) for Arch Linux.
OS: Arch Linux

I expect changing the editor to work and not crash the app since React docs explicitly mention at https://reactjs.org/docs/hooks-reference.html#usememo that applications should work even if the memoized var is cleared:
> **You may rely on useMemo as a performance optimization, not as a semantic guarantee.** In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.

The issue seems to be caused by not including [editor] as a dependency in plenty of places.

aciccarello:
**Description**
A clear and concise description of what this pull request solves. (Please do not just link to a long issue thread. Instead include a clear description here or your pull request will likely not be reviewed as quickly.)

**Issue**
Fixes: (link to issue)

**Example**

**Context**

Closes #3886

**Checks**

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,7 +4,8 @@
     "plugin:import/typescript",
     "prettier",
     "prettier/@typescript-eslint",
-    "prettier/react"
+    "prettier/react",
+    "plugin:react-hooks/recommended"
   ],
   "plugins": ["@typescript-eslint", "import", "react", "prettier"],
   "parser": "@typescript-eslint/parser",
@@ -19,7 +20,7 @@
   "settings": {
     "import/extensions": [".js", ".ts", ".jsx", ".tsx"],
     "react": {
-      "version": "detect"
+      "version": "16"
     }
   },
   "env": {
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -69,6 +69,7 @@ import {
 import { RestoreDOM } from './restore-dom/restore-dom'
 import { useAndroidInputManager } from '../hooks/android-input-manager/use-android-input-manager'
 import { useTrackUserInput } from '../hooks/use-track-user-input'
+import { AndroidInputManager } from '../hooks/android-input-manager/android-input-manager'
 
 type DeferredOperation = () => void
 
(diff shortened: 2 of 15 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -181,11 +182,15 @@ export const Editable = (props: EditableProps) => {
     }
   }, [autoFocus])
 
+  let androidInputManager: AndroidInputManager | null
```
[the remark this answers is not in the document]
jason0x43: Weird, but good enough, I suppose. 🙂 

The intent might be a bit clearer with a ref, and then eslint wouldn't need it in the dependency list.

```tsx
const androidInputManager = useRef<AndroidInputManager>();

const onDOMSelectionChange = useMemo(() => throttle(() => {
  ... 
  // do something with androidInputManager.current
  ...
}), [editor, readonly, state]);

androidInputManager.current = useAndroidInputManager(...);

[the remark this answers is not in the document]
Comment (aciccarello) [reply packages/slate-react/src/components/editable.tsx]: I added a commit to implement this change. Let me know if the intent seems clear.

File: .eslintrc.json [hunk 2]
```diff
@@ -19,7 +20,7 @@
   "settings": {
     "import/extensions": [".js", ".ts", ".jsx", ".tsx"],
     "react": {
-      "version": "detect"
+      "version": "16"
```
aciccarello, line 23: I think yarn was causing detect to not work

packages/slate-react/test/index.spec.tsx
```diff
@@ -22,16 +22,16 @@ describe('slate-react', () => {
...
         let el: ReactTestRenderer
 
         act(() => {
           el = create(
             <Slate editor={editor} value={value} onChange={() => {}}>
               <Editable
-                renderElement={({ element, children }) => {
-                  React.useEffect(() => mounts(element), [])
+                renderElement={({ children }) => {
+                  useEffect(() => mounts(), [])
```
aciccarello, line 34: This is meant to be testing the mount call count (hence the empty dependency) but we were passing the element which eslint wants included in the dependency array. It's not needed for the test.

packages/slate-react/src/hooks/android-input-manager/use-android-input-manager.ts
```diff
@@ -26,6 +26,9 @@ export function useAndroidInputManager({
   node,
   ...options
 }: UseAndroidInputManagerOptions) {
+  /* eslint-disable react-hooks/rules-of-hooks */
```
jason0x43, line 29: Another option here would be to move the IS_ANDROID check outside of the function, like

```tsx
export const useAndroidInputManager = !IS_ANDROID ? () => null : ({
  node,
  ...options
}: UseAndroidInputManagerOptions) => {
```

packages/slate-react/src/hooks/use-slate-selector.tsx
```diff
@@ -113,16 +113,23 @@ export function useSlateSelector<T>(
  * Create selector context with editor updating on every editor change
  */
 export function getSelectorContext(editor: Editor) {
```
jason0x43, line 115: This is only used in one place. It's probably worth renaming it to `useSelectorContext` so we don't have to disable the hooks rule for a bunch of hook calls.

site/examples/code-highlighting.tsx
```diff
@@ -234,20 +237,23 @@ const SetNodeToDecorations = () => {
     )
 
     editor.nodeToDecorations = nodeToDecorations
-  }, [editor.children])
+  }, [editor])
```
jason0x43, line 240: We should be careful here, because `editor.children` might change even if `editor` doesn't.

aciccarello: Good catch. That was significant here. What's confusing is that this function shouldn't have to know about editor.children. This function is setting up a custom `nodeToDecorations` map to optimize some of the other logic. I think it should be triggered by something more explicit than `editor.children` changing.

I can make an eslint exception for now, but the implementation should be thought through. The simplest change that doesn't require an exception would be to drop useMemo.

dylans (commented):
Overall looks good, it looks like @jason0x43's feedback covers everything I would have suggested, so please let me know when you've addressed the other feedback. Thanks!

PR #5367, opened 2023-03-16
Fix `markdown-shortcuts.test.ts` for webkit

12joan:
**Description**
The `markdown-shortcuts.test.ts` Playwright test uses the `Home` key, which doesn't work in webkit. This PR changes the test to use `Meta+ArrowLeft` instead.

It also increases the default timeout from 10s to 20s, and 60s for the code highlighing tests. This was necessary to make the test suite pass consistently on my Mac.

**Checks**
- [X] ~~You've [added a changeset](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md) if changing functionality. (Add one with `yarn changeset add`.)~~

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -46,7 +46,7 @@ const retries = process.env.PLAYWRIGHT_RETRIES
 const config: PlaywrightTestConfig = {
   testDir: './playwright',
   /* Maximum time one test can run for. */
-  timeout: 10 * 1000,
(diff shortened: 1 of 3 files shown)


dylans (approved):
You might be able to avoid some of these longer timeouts by looking for versions of various calls that specifically wait for something to be present.

But for now let's just get you unstuck with the test suite running.

PR #5393, opened 2023-04-12
pastehtml-example:should paste plain text into code block

qirong77:
**Description**
When we copy code from editors such as vscode, we don't want to keep the original style (Vscode uses div to represent each line of code, so it will not be parsed correctly) but use plain text to paste into our code block,so before we insertData , We should judge whether the current selection is a code block(eg:'code-block,'code-line') before we use inertData.


dylans (changes requested):
I agree with the sentiment.

I don't think this check is sufficient though because you can have nested structures in code blocks and I believe this would have unexpected results.

Perhaps take a look at https://github.com/udecode/plate/tree/main/packages/nodes/code-block/src which is a more comprehensive code block implementation for inspiration?

dylans:
> > I agree with the sentiment.
> > I don't think this check is sufficient though because you can have nested structures in code blocks and I believe this would have unexpected results.
> > Perhaps take a look at https://github.com/udecode/plate/tree/main/packages/nodes/code-block/src which is a more comprehensive code block implementation for inspiration?
> 
> Thank you for your reply. I read the link you gave me and found that he used the deserializeHtmlCodeBlockPre method to process the copied code block. The main function of this method is to obtain the textContent of the node. I think his idea should be the same as mine. It is consistent. We need to perform additional processing on the copied code block, but the method may be different. However, there is no additional processing on the code block in this example, so it caused some problems when I learned and used it, but I'm clear now, thanks.

Right, I also meant more that just checking to see if the parent has type code in it might not be sufficient to determine if the paste is happening in a code-block.

PR #5396, opened 2023-04-18
Take previously added/removed nodes into account when fetching child …

Moerphy:
**Description**
When when a node has two child nodes that would violate the core constraints (i.e. two inline nodes as children of the Editor) this can cause other nodes to be deleted that do not violate any core constraints. 

**Example**
Input:
```
<editor>
    <inline>one</inline>
    <block>two</block>
    <inline>three</inline>
    <block>four</block>
</editor>
```
Normalized result before this change:
```
<editor>
    <block>four</block>
</editor>
```
Normalized result after this change:
```
<editor>
    <block>two</block>
    <block>four</block>
</editor>
```
See adapted unit tests for more examples.

**Context**
Disclaimer: I only know very little about Slates internal, so my understanding here might not be 100% correct. 

Looks to me like the current logic for iterating children in the normalization uses two sets of variables `node` & `i` and `currentNode` & `n`. I think `node` & `i` are referring to the state of the editor when `normalizeNode` was entered and `currentNode`& `n` are referring the "current" state, taking into account transforms from an earlier iteration.
The transforms that are executed when a constraint violation is detected are always using `n` to construct the path to the transformed node, but the constraint was checked on the child node at index `i` instead, which leads to the wrong nodes being transformed.

**Checks**

### DIFF
--- a/.changeset/tall-gorillas-deny.md
+++ b/.changeset/tall-gorillas-deny.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Correct core normalization that could cause wrong nodes to be removed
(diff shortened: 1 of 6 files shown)


dylans (approved):
Looks reasonable to me @Moerphy . We'll land it. If it breaks something else we can always revert.

Issue #5404, opened 2023-04-26
Incorrect return type for useSlateStatic


JarrodMFlesch:
**Problem**
`useSlateStatic` return type in docs and code are not aligned.

**Solution**
Change the return type from `Editor` to `ReactEditor`

See the [useSlateStatic doc section](https://docs.slatejs.org/libraries/slate-react/hooks#useslatestatic-reacteditor) it says it should return `ReactEditor` but in code it does not:

![CleanShot 2023-04-26 at 16 01 10](https://user-images.githubusercontent.com/30633324/234690155-5c40a0dd-3cec-418e-8afb-b7779cf6e00c.png)

PR #5405, opened 2023-04-26
chore: corrects return type for the use-slate-static hook

JarrodMFlesch:
**Description**
Aligns the return type listed in the docs with the return type used in code for the `useSlateStatic` hook.

**Issue**
Fixes: #5404 

**Example**
Docs: https://docs.slatejs.org/libraries/slate-react/hooks#useslatestatic-reacteditor

CleanShot 2023-04-26 at 16 01 10

**Checks**

### DIFF
--- a/.changeset/stupid-lies-search.md
+++ b/.changeset/stupid-lies-search.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+corrects return type for the useSlateStatic hook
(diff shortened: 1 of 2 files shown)


JarrodMFlesch:
I could not get tests to run locally but am open to any guidance there!

dylans:
> I could not get tests to run locally but am open to any guidance there!

It looks like we had a failure due to the changes in 0.94.0. Will get that fixed and then these should work again.

dylans (changes requested):
@JarrodMFlesch please take a look at the integration test failures that I believe are a consequence of fixing the type definition here. Thanks.

PR #5407, opened 2023-04-27
Error handling

### LINKED ISSUE #3243: Mentions example crashes when there's no text

_bug_

If you remove all the text in the mentions example and then type, it crashes. This is because it's Editor.before and Editor.after are returning `undefined`.

### LINKED ISSUE #3558: Error: NotFoundError: Failed to execute 'removeChild' on 'Node'
Bug.

The editor is crashing for some users, even at https://www.slatejs.org. Some extension or accessibility feature is modifying the DOM and then when the user types in the editor they get the following error:
"Error: NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node."

Most of the crashes on my website were due to users using Google Translate. I fixed that by adding the following attributes to the editor <div content=“notranslate” translate=“no” class=“notranslate”>. But still some users are getting the crash. One particular user is using Google Chrome and has no extensions installed.

### LINKED ISSUE #3568: Calling addMark in onDOMBeforeInput crashes Slate when selection is not collapsed

report a _bug_

Slate: 0.57.1
Browser: Chrome 80
OS: Mac

Adding an `editor.addMark` call to an `onDOMBeforeInput` callback reliably crashes Slate when the selection is not collapsed. The error is `Cannot resolve a DOM point from Slate point`, but stepping though the code, I noticed that the selection does not appear to be deleted correctly before the new text is inserted. Maybe because the `addMark` calls `setNode` and splits the block for non-collapsed selections? 

Here's a minimal reproduction of the issue:

https://codesandbox.io/s/little-rain-5089t

Select the first 'D', then type 'C' and it will crash.

In that example, after the insert_text command runs, `editor.childen[0].children` looks like 
```json
"[{"text":"ABC"},{"text":"D","bold":true},{"text":"DEF"}]"
```

having performed the following operations:

```json
[
  {
    "type": "split_node",
    "path": [
      0,
      0
    ],
    "position": 3,
    "target": null,
    "properties": {}
  },
  {
    "type": "split_node",
    "path": [
      0,
      0
    ],
    "position": 2,
    "target": null,
    "properties": {}
  },
  {
    "type": "set_selection",
    "properties": {
      "focus": {
        "path": [
          0,
          2
        ],
        "offset": 0
      }
    },
    "newProperties": {
      "focus": {
        "path": [
          0,
          1
        ],
        "offset": 1
      }
    }
  },
  {
    "type": "set_node",
    "path": [
      0,
      1
    ],
    "properties": {},
    "newProperties": {
      "bold": true
    }
  },
  {
    "type": "set_selection",
    "properties": {
      "anchor": {
        "path": [
          0,
          1
        ],
        "offset": 0
      },
      "focus": {
        "path": [
          0,
          1
        ],
        "offset": 1
      }
    },
    "newProperties": {
      "anchor": {
        "path": [
          0,
          0
        ],
        "offset": 2
      },
      "focus": {
        "path": [
          0,
          0
        ],
        "offset": 3
      }
    }
  },
  {
    "type": "remove_text",
    "path": [
      0,
      0
    ],
    "offset": 2,
    "text": ""    // <--- This looks like a bug.
  },
  {
    "type": "set_selection",
    "properties": {
      "anchor": {
        "path": [
          0,
          0
        ],
        "offset": 2
      }
    },
    "newProperties": {
      "anchor": {
        "path": [
          0,
          0
        ],
        "offset": 3
      }
    }
  },
  {
    "type": "insert_text",
    "path": [
      0,
      0
    ],
    "offset": 3,
    "text": "C"
  }
]
```

which demonstrates the text not being deleted correctly, and the mark applied to the wrong selection.

The crash eventually happens in the call to editor.onChange() within `apply`.

I believe the value after this input should look like:
```json
"[{"text":"AB"},{"text":"C","bold":true},{"text":"DEF"}]"
```

### LINKED ISSUE #3586: Bug: Cannot find DOMPoint after applying formatting keyboard shortcuts (onDOMBeforeInput->event.inputType) - also in stock examples

Bug

1. Go to https://www.slatejs.org/examples/hovering-toolbar
2. Replace all content with just 'abc'
3. Select the 'c'
4. Press `Cmd-B` or `Cmd-I`
5. The app crashes

Debugging locally it seems to `Editor.node(editor, point.path)` in `toDOMPoint` finds the wrong node when looking for the `domPoint` of `domFocus`.

- Slate: 0.57.1 (latest)
- Browser: Chrome/Electron (keybindings don't seem to work in Firefox / Safari)
- OS: Mac

It should bold or italicize the `c` — the same behavior as pressing the formatting buttons in the hovering toolbar.

Screen Shot 2020-04-05 at 17 23 17

Screen Shot 2020-04-05 at 17 35 09

### LINKED ISSUE #3641: We need to talk about... Slate's Draconian :) error handling

### LINKED ISSUE #3705: Slate history error
#### Error: Cannot apply an incomplete "set_selection" operation properties {"focus":{"path":[0,1],"offset":0}} when there is no current selection.

transform
src/interfaces/editor.ts:1441
```
  1438 |   selection = newProperties
  1439 | } else if (selection == null) {
  1440 |   if (!Range.isRange(newProperties)) {
  1441 |     throw new Error(
  1442 |       Cannot apply an incomplete "set_selection" operation properties ${JSON.stringify(
  1443 |         newProperties
  1444 |       )} when there is no current selection.
```

My application is breaking inconsistently when using undo/redo with Slate History by pressing ctrl+z/cmd+z.

### LINKED ISSUE #3708: insertText after addMark/removeMark will carsh

bug

insertText after addMark or removeMark in a React event handler, the editor will crash
 
editor

code example:
https://codesandbox.io/s/slate-addmark-leggp

Slate: 0.58.1
Browser: Chrome
OS: Mac

It may be caused by addMark function call editor.onChange sync and apply function call it in a microtask. conflict with lazy state update in React synthetic events
https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/create-editor.ts#L109

### LINKED ISSUE #3723: Cannot resolve a Slate point from DOM point
bug
Under the situation that if a normally `parent element`'s children are all `void element`,clicking the area(`contenteditable=false`) that belong to `parent element` and  multiple times will result in an error(`Cannot resolve a Slate point from DOM point: [object Text],17`)

here is the exmaple:
[example](https://codesandbox.io/s/cool-cache-xn6u5?file=/src/App.js)
Jun-05-2020 17-20-30
The bug doesn't happen each time. 

Slate: 0.58.1
Browser: Chrome 
OS: Mac 

No error occured!

### LINKED ISSUE #3777: Adding an accented character as the first letter on a new line crashes the editor in Firefox

Bug

Typing an accented character as the first letter on a line crashes the editor in firefox.

This can be reproduced on https://www.slatejs.org/examples/richtext

firefox

Slate: 0.58.4
Browser: Firefox
OS: Mac

That the accented character is inserted normally into the editor

### LINKED ISSUE #3811: [ Backspace + Space ] Leads to Repetition and continous [ Backspace ] leads to crash of Slate js Example on android device.
####  Backspace + Space ] Leads to Repetition and continous [ Backspace ] leads to crash of Slate js Example on android device.

ezgif com-video-to-gif

Slate: 0.58.X
Browser: Chrome
OS: Android

### LINKED ISSUE #3870: Editor crashes when clicking on initial void block or near its boundarie
Bug.

When I add inline void block, save value and reload editor, and when i click on void block or at its boundaries (beginning or end) it crashes. I get errors: `Uncaught Error: Cannot resolve a DOM point from Slate point: {"path":[0,0],"offset":33}` or `cannot find descedant.....` Also when I click somewhere at text, and try to add more text or delete, cursor just moves through text without changes. NOTE: this happens only if void node exists in initial value.

Slate version: 0.47
Here is gif: https://im4.ezgif.com/tmp/ezgif-4-29acae2ce4fd.gif
Codesandbox: https://codesandbox.io/s/slate-forked-843c2

It should behave normally as when I add simple text.

### LINKED ISSUE #3878: Editor's selection points to non-existing nodes when value is changed programmatically (Chrome, Safari)

Bug.

Peek 2020-09-16 19-15

https://codesandbox.io/s/slate-reproductions-forked-lzc7g?file=/index.js

1. Focus the last paragraph in the example
2. Click "Reset" button

`useSlate().selection` points to a non-existing location (location after performing step (1))

Slate: 0.58.4
Browser: Chrome, Safari
OS: Windows, Linux, Mac

`useSlate().selection` points to an existing location. 

Works fine in **Firefox**. 

Peek 2020-09-16 19-18

### LINKED ISSUE #3900: BUG in the example : Copy/paste html by triple-clicking

Report a bug

When I copy and paste a sentence in the editor (by triple clicking on it to select it), there is always a blank `p` at the end of my copied content and it makes my `withHtml` crashes.

```html
<meta charset="utf-8" />
<p
    data-slate-node="element"
    data-slate-fragment="JTVCJTdCJTIydHlwZSUyMiUzQSUyMmJsb2NrLXF1b3RlJTIyJTJDJTIyY2hpbGRyZW4lMjIlM0ElNUIlN0IlMjJ0ZXh0JTIyJTNBJTIyQSUyMHdpc2UlMjBxdW90ZS4lMjIlN0QlNUQlN0QlMkMlN0IlMjJ0eXBlJTIyJTNBJTIycGFyYWdyYXBoJTIyJTJDJTIyY2hpbGRyZW4lMjIlM0ElNUIlN0IlMjJ0ZXh0JTIyJTNBJTIyJTIyJTdEJTVEJTdEJTVE"
>
    <span data-slate-node="text">
        <span data-slate-leaf="true"><span data-slate-string="true">A wise quote.</span></span>
    </span>
</p>

<!-- This part is not supposed to be here-->
<p data-slate-node="element">
    <span data-slate-node="text">
        <span data-slate-leaf="true"><span data-slate-string="true"></span></span>
    </span>
</p>
<!----------------------------------------->
```

Strangely, it only works on the last paragraph but not on the others :

chrome-capture

The example : [https://www.slatejs.org/examples/paste-html](url)

Slate: 0.58.4
Browser: Chrome 
OS: Mac 

#### Possible fix

The code of the example : https://github.com/ianstormtaylor/slate/blob/master/site/examples/paste-html.tsx

It seems that it crashes when children is an empty array so I added a condition.

```js
  if (children.length) {
    if (nodeName === "BODY") {
      return jsx("fragment", {}, children);
    }

    if (HTML_ELEMENT_TAGS[nodeName]) {
      const attrs = HTML_ELEMENT_TAGS[nodeName](el);
      return jsx("element", attrs, children);
    }

    if (HTML_TEXT_TAGS[nodeName]) {
      const attrs = HTML_TEXT_TAGS[nodeName](el);
      return children.map((child) => jsx("text", attrs, child));
    }
  }
```

What do you think ? Any better suggestion ?

### LINKED ISSUE #3918: Slate does not react on delete button on page refresh - it just moves cursor accross letters
Bug.

When I navigate (i.e. click from navbar) to page containing slate editable text in my React app, everything works fine but when I'm at that page and refresh it, slate does not react anymore on delete button, it just moves cursor backwards accross letters. Or if I try to insert letters, I get error `Cannot resolve a DOM point from Slate point`. What could be the problem?

Slate: 0.47
Browser: Chrome
OS: Mac

GIF: https://gyazo.com/53d38d9fb90035dc276f4d5288f24048

### LINKED ISSUE #3948: Slate errors cannot be caught by error boundaries and crash the whole app
For some reason, currently, slate errors crash the whole app and cannot be caught by https://reactjs.org/docs/error-boundaries.html

### LINKED ISSUE #4001: 🐞 Bug: German Keyboard backtick isn't recognized in onChange event; crashes editor when using placeholder

🐞 Bug

The first keypress of a german backtic fails to register an onChange event which creates different slate and dom points.

From sandbox: https://codesandbox.io/s/slate-reproductions-forked-m8qpg?file=/package.json
- delete default text until placeholder displays
- make sure MacOS german keyboard is enabled
- type a backtick (on my mac u.s. keyboard this is the "=" key)
- then type any other character

**Editor crashes with error `Cannot resolve a Slate point from DOM point: [object Text],1`**

This can also be reproduced in [slate.js examples](https://www.slatejs.org/examples/richtext)

Slate: 0.59.0
Browser: Chrome Version 86.0.4240.198 (Official Build) (x86_64)
OS: Mac 11.01

- The first instance of typing a german backtick should register an onChange event 
- The editor should not crash when placeholder is enabled.

### LINKED ISSUE #4081: Error: Cannot find a descendant at path [0] in node - CRA live-reload
Bug

Create React App live editing error on every change. Using installation [instructions](https://docs.slatejs.org/walkthroughs/01-installing-slate)

To reproduce the error see this [repository](https://github.com/marcoSven/slate-test-one) with a "[workaround](https://codesandbox.io/s/falling-shadow-1k49x?file=/src/App.js:690-759)" to prevent the error.

Sandbox [URL](https://codesandbox.io/s/falling-shadow-1k49x?file=/src/App.js)
Edit the file and you should see the error. Uncomment the lines suggested above to prevent the error.

Slate: 0.59.0
Browser: Chrome
OS: Mac 

Live-editing should work without errors.

### LINKED ISSUE #4088: Mention example range logic throws Cannot resolve a DOM point from Slate point
**Description**
Throws error `Cannot resolve a DOM point from Slate point` after copying and pasting the logic used to find the string range in the mentions example.

**Sandbox**
https://codesandbox.io/s/slate-reproductions-forked-hz6fk?file=/index.js:1441-1484

**Steps**
To reproduce the behavior:
1. Go to sandbox link
2. type `@` followed by any char
3. See an error

**Expectation**
It would expect it to return the element.

**Environment**
- Slate Version: 0.59
- Operating System: Windows 10
- Browser: Chrome 88.0.4324.182 (Official Build) (64-bit) (cohort: Stable)

**Context**
Just trying to duplicate the mentioned example.

### LINKED ISSUE #4094: Crash on text update from outside of the editor when cursor to the right
When the editor content is updated from outside (e.g. by a database call) then the editor crashes 
- if the new text is shorter than the old text 
- and the cursor has been placed to the right (i.e. is outside of the new shorter text).

I just added a `useEffect` to the plain text example to show. Just start writing sth. behind the default text in the editor and it will crash.

See:
https://codesandbox.io/s/slate-reproductions-forked-qets6

added code:
`
  useEffect(() => {
    setInterval(() => setValue(initialValue), 2000);
  }, []);
 ;
`

Any idea how I can prevent / workaround this?

### LINKED ISSUE #4136: Crash if you enter multibyte characters.

### LINKED ISSUE #4323: Cannot resolve a DOM point from Slate point
I saw a couple of closed ticket with this error, but we think our issue is a bit different from those.
We have a scenario, where Slate is just a part of a document editor.
```
  const editorRef = useRef<ReactEditor>()
  if (!editorRef.current) {
    editorRef.current = withReact(createEditor())
  }
  const editor = editorRef.current
```
This document editor has it's own history, so we can not use slate-history.
Whenever this document editor undos or redus, we update the value of Slate and we get the error in the title.

As far as we can tell, the error happens when the position of the cursor will be invalid for the new value.
This is something that should be easy to fix, we just have to save the selector as well in the document state when onChange happens in Slate.

But the onChange does not provide a selector and the `editor.selector` does not represent the new state of the selector at the time when the onChange fires.

So either way slate should provide a way to get the proper selector for the change.
And also there should be a safe way to set a new selector too, since if we set the selector before the editor gets the new value, it crashes with the same error, but there is no event that slate fires after it sets its new values.

We tried to do something like this, but to no avail:
```
    editor.selection = {
      anchor: {
        path: [0],
        offset: 0
      },
      focus: {
        path: [0],
        offset: 0
      }
    };
    editor.children = value;
    editor.selection = selection;
```

If we were to miss something, please let us know how to save and set value and selector safely.

### LINKED ISSUE #4485: Cannot resolve a DOM point from Slate point: {"path":[36,0],"offset":7}
So basically when i type sometimes, randomly ui crashes with this error:
image

I'm using slate versions:
image

My code: 

```
import React, { useRef, useCallback, useEffect, useMemo, useState } from 'react'
import {useDispatch, useSelector} from 'react-redux'
import { Slate, Editable, withReact } from 'slate-react'
import { withHistory } from 'slate-history'
import { Transforms, createEditor } from 'slate'
import InlineSVG from 'svg-inline-react'
import { StyleSheet, css } from 'aphrodite'
import debounce from 'debounce'

import Theme from 'shared/assets/theme'
import * as Actions from 'state/actions'

import iconDelete from 'design/assets/delete-flow.svg'

const styles = StyleSheet.create({
  entry: {
    display: 'flex',
    flexDirection: 'row',
    position: 'relative',
    fontSize: '13px',
    ':hover .deleteExample': {
      opacity: 1
    },
    marginBottom: '8px'
  },
  field: {
    flex: '25%',
    alignSelf: 'flex-start',
    width: '25%',
    borderBottom: 'none',
    position: 'relative',
    backgroundColor: Theme.colors.fill.c1,
    maxWidth: '760px',
    overflow: 'hidden',
    webkitAppearance: 'none',
    border: 'none',
    color: Theme.colors.text.c1,
    background: Theme.colors.fill.c1,
    border: 'none',
    lineHeight: '38px',
    padding: '0 8px',
    borderRadius: '3px',
    marginRight: '4px',
    ':focus': {
      background: Theme.colors.fill.c0,
      color: Theme.colors.text.c0
    },
    '::-webkit-input-placeholder': {
      color: Theme.colors.text.c4
    },
    ':-moz-placeholder': {
      color: Theme.colors.text.c4
    },
    ':-ms-input-placeholder': {
      color: Theme.colors.text.c4
    },
    ':last-child': {
      marginRight: 0
    }
  },
  fieldId: {
    flex: '15%',
    width: '12.5%',
    paddingRight: '30px',
    alignSelf: 'flex-start',
    marginRight: 0
  },
  synonyms: {
    position: 'relative',
    flex: '60%',
    alignSelf: 'flex-start',
    flex: '1 0 62.5%',
    width: '62.5%',
    padding: '3px 8px 2px 8px',
    backgroundColor: Theme.colors.fill.c1,
    borderRadius: '3px',
    margin: '0 16px 0 4px',
    fontSize: '13px',
    fontWeight: '400',
    lineHeight: '32px',
    color: Theme.colors.text.c0,
    ':hover .deleteExample': {
      opacity: 1
    },
    ':focus-within': {
      background: Theme.colors.fill.c0,
      color: Theme.colors.text.c0
    }
  },
  synonymsEmpty: {
    ':after': {
      color: Theme.colors.text.c5,
      display: 'inline-block',
      lineHeight: '32px',
      position: 'absolute',
      userSelect: 'none',
      content: '"Synonyms…"',
      left: '10px',
      top: '4px'
    }
  },
  deleteButton:{
    position: 'absolute',
    zIndex: '10',
    right: '6px',
    padding: '3px',
    top: '7px',
    opacity: 0,
    transition: 'opacity 0.2s cubic-bezier(0.3,0,0,1.3)',
    cursor: 'pointer'
  },
  synonym: {
    color: Theme.colors.text.c1,
    boxShadow: `0 0 0 2px ${Theme.colors.black.c4}`,
    backgroundColor: Theme.colors.black.c4,
    borderRadius: '2px',
    lineHeight: '13px',
    display: 'inline-block',
    padding: '3px 1px 1px 1px',
    position: 'relative',
    top: '2px',
    display: 'inline-block',
    marginRight: '8px'
  }
})

const Entry = props => {

  const { 
    entry,
    entryIndex,
    isLast
  } = props

  const editor = useMemo(() => withHistory(withReact(createEditor())), [])
  const type = useSelector(state => state.agent.active?.type)
  const isProdAgent = type === 'PROD'
  const dispatch = useDispatch()

  const renderElement = editor => useCallback(props => <Element {...props} editor={editor} />, [])
  const renderLeaf = editor => useCallback(props => <Leaf {...props} editor={editor} />, [])

  const [keyword, setKeyword] = useState('')
  const [synonyms, setSynonyms] = useState([])
  const [identifier, setIdentifer] = useState('')
 
  const handlePastedText = useCallback(e => {
    const text = e.clipboardData.getData('Text') || ''

    if(text.length > 0) {
      
      const lines = text.split('\n')
      if(lines.length > 1) {

        const entries = []
        e.preventDefault()

        lines.forEach(line => {
          const words = line.split('\t').filter(n => n)
          entries.push({
            keyword: words.shift().trim(),
            synonyms: words || []
          })
        })
        dispatch(Actions.createEntries(entryIndex, entries))
      }
    }
  }, [])

  useEffect(() => {
    setSynonyms(serialize(entry.synonyms))
    if(keyword !== entry.keyword){
      setKeyword(entry.keyword)
    }
    if(identifier !== entry.identifier){
      setIdentifer(entry.identifier)
    }
  }, [entry])

  // Redux delayed voodoo magic
  // kudos to stack overflow
  const [persistedValue, setPersistedValue] = useState(false)
  const throttled = useRef(debounce(data => {
    if(data === false) {
      return
    }
    dispatch(Actions.updateEntry(entryIndex, data))
  }, 200))
  useEffect(() => throttled.current(persistedValue), [persistedValue])

  return (
    <div className={css(styles.entry)}>
      <input
        disabled={isProdAgent}
        type="text"
        placeholder="Keyword"
        className={css(styles.field)}
        value={keyword}
        onChange={event => {
          setKeyword(event.target.value)
          setPersistedValue({
            ...entry,
            keyword: event.target.value
          })
        }}
        onPaste={handlePastedText}
        autoFocus={!entry.keyword}
      />
      <div className={css(styles.synonyms)}>
        <Slate
          editor={editor}
          value={synonyms}
          onChange={newValue => {
            setSynonyms(newValue)

            if(newValue !== synonyms) {
              setPersistedValue({
                ...entry,
                synonyms: deserialize(newValue)
              })
            }
          }}
        >
          <Editable
            readOnly={isProdAgent}
            renderElement={renderElement(editor)}
            renderLeaf={renderLeaf(editor)}
            onKeyDown={event => {
              if(event.key == ',' || event.key == ';') {
                event.preventDefault()
                Transforms.insertNodes(
                  editor,
                  { 
                    type: 'entry',
                    children: [{
                      type: 'text',
                      text: ''
                    }]
                  },
                  {
                    select: true 
                  }
                )
              }
            }}
          />
        </Slate>
      </div>
      <input
        disabled={isProdAgent}
        type="text"
        placeholder="Key…"
        className={css(styles.field, styles.fieldId)}
        value={identifier}
        onKeyDown={event => {
          if(isLast && 
            event.key === 'Tab' && 
            !event.altKey && 
            !event.shiftKey && 
            !event.metaKey) {
            event.preventDefault()
            dispatch(Actions.createEntry())
          }
        }}
        onChange={event => {
          setIdentifer(event.target.value)
          setPersistedValue({
            ...entry,
            identifier: event.target.value
          })
        }}
      />
      <div 
        className={`${css(styles.deleteButton)} deleteExample`}
        contentEditable={false}
        style={{ userSelect: "none" }}
      >
        <InlineSVG
          src={iconDelete}
          onClick={event => {
            if(isProdAgent) return
            event.preventDefault()
            dispatch(Actions.removeEntry(entryIndex))
          }}
          title="Delete entry"
        />
      </div>
    </div>
  )
}

const serialize = values => {
  if(!values || !values.length) {
    return [{
      type: 'entry',
      children: [{ 
        type: 'text',
        text: ''
      }]
    }]
  }
  const nodes = values.map(synonym => ({
    type: 'entry',
    children: [{
      type: 'text',
      text: synonym
    }]
  }))
  return nodes
}

const deserialize = values => values.reduce((result, element) => {
  for (let i = 0; i < element.children.length; i++) {
    const child = element.children[i]
  
    const synonyms = child.text.split(',')
    for (let x = 0; x < synonyms.length; x++) {
      const synonym = synonyms[x]
      result.push(synonym)
    }
  }
  return result
}, [])

const Element = ({ attributes, children }) => (<span {...attributes}>{children}</span>)

const Leaf = ({ attributes, children, leaf }) => {
  if(!leaf || !leaf.text.length || leaf.isSeparator){
    return <span {...attributes}>{children}</span>
  }
  
  return <span className={css(styles.synonym)} {...attributes}>{children}</span>
}

export default Entry
```

### LINKED ISSUE #4561: Hot refresh causes an error
**Description**
```
Cannot find a descendant at path [0] in node: {"children":[],"operations":[],"selection":null,"marks":null,"history":{"undos":[],"redos":[]}}
```

**Sandbox**
A link to a sandbox where the error can be reproduced. (You can start from the base sandbox here: https://codesandbox.io/s/slate-js-hovering-tooolbar-i8xcw?file=/HoveringMenuExample.js

**Steps**
To reproduce the behavior:
1. Do anything that causes hot refresh, like adding space in the initial value
2. See the error

### LINKED ISSUE #4564: Cannot resolve a DOM point from Slate point: {"path":[0,0],"offset":3} when programmatically removing content
**Description**
I want to clear textarea on Enter. When I do so, slate throws an error:
`Cannot resolve a DOM point from Slate point: {"path":[0,0],"offset":3}`

**Recording**
slate

**Sandbox**
https://codesandbox.io/s/slate-0-5x-forked-kr8sc?file=/src/App.tsx

**Steps**
To reproduce the behavior:
1. Type something in the editor
2. Press Enter
3. Error is thrown

**Expectation**
I shouldn't get an error. I'm guessing it tries to preserve caret position across updates and fails because the new content is shorter.

**Environment**
- Slate Version: `slate@0.66.5`, `slate-react@0.66.4`
- Operating System: Windows 10
- Browser: Chrome 94.0.4606.61

### LINKED ISSUE #4581: Deleting a void element or text decoration on FireFox and inserting text throws an error
**Description**
On FireFox, when a text decoration or void element is created and deleted via backspace and text is typed in afterwards, the following error will throw:
`Error: Cannot resolve a Slate point from DOM point: [object Text],1`

**Recording**

https://user-images.githubusercontent.com/28989541/136637555-2a9e54c9-843a-43db-97d8-b71ad5eda26f.mov

https://user-images.githubusercontent.com/28989541/136637553-d67bbf8c-ed6b-46ba-b906-1e3592516026.mov

**Sandbox**
https://www.slatejs.org/examples/search-highlighting

**Steps**
To reproduce the behavior:
1. Go to /mentions demo
2. Delete entire contents of editor
3. Type in any text and insert a mention
4. Delete the mention with Backspace
5. Insert any text character
6. Validate that the editor crashes
7. Go to /search-highlighting demo
8. Delete entire contents of editor
9. Type in anything into search bar
10. Type in any text followed by text that matches the query in the search bar
11. Delete the highlighted text
12. INsert any text character
13. Validate that the editor crashes

**Expectation**
No error, should be able to insert text without a problem.

**Environment**
- Slate Version: 0.66.5
- Operating System: MacOS
- Browser: FireFox

**Context**
`slate@0.65.3` appears to be stable regarding this behavior. Tried replicating the same steps with `slate@0.66.4` and the same error is thrown. Deleting elements with `Cmd+Backspace` and inserting text or highlighting and deleting elements and inserting text don't throw this error, it is only when items are deleted one at a time with Backspace.

### LINKED ISSUE #4643: Invalid selection leads to uncatchable error: "Cannot resolve a DOM point from Slate point"
**Problem**
When dealing with complex replacements and selection operations, it's very common to get invalid ranges, which cause the editor to crash with an error `Cannot resolve a DOM point from Slate point`.

Example code:

```js
// Provide an invalid selection as the second argument.
Transforms.select(editor, { anchor: { path: [55,0] offset: 0 }, focus: { path: [55,0], offset: 10 } } );
```

I know this would be _partially_ the fault of an unreliable editor built on top of Slate, which could result in such invalid ranges, but I believe Slate could provide better ways to deal with these invalid ranges.

**Solution**
Suggested approaches:
- Throw a `console.warn` warning if the selection is invalid, and do nothing with the command. E.g if it's a select operation, don't perform the operation and throw a descriptive warning.
- Make the existing error easily catchable within a try/catch block.

**Alternatives**
Given Slate is still in beta, and there are some open issues around inconsistencies in ranges selecion/removal, I'd say the first option under the proposed solutions would be best - doing nothing in case there's an invalid range and throw a descriptive warning.

### LINKED ISSUE #4658: Custom table node allows entering text outside the table which leaves slate in a broken state
**Description**
On Google Chrome (tested Arch Linux and MacOS) and Safari, using a custom node to support html tables allows moving the cursor beyond the bounds of the table, when there are no nodes below the table. When entering any character in this position, the character is entered twice, one in the last cell of the table, and one below the table. Attempting to delete the character in the last cell deletes two characters. Moving the cursor below the table shows an error in the developer console:
> Error: Cannot resolve a Slate point from DOM point: [object Text],1
It is not possible to delete the character below the table. This issue does not effect Firefox (tested Arch Linux and MacOS). 

**Recording**
https://user-images.githubusercontent.com/239226/141764397-5abbaf88-2df5-4411-8aa2-abf5ba7a122f.mp4

**Sandbox**
https://www.slatejs.org/examples/tables

**Steps**
To reproduce the behavior:
1. Navigate to https://www.slatejs.org/examples/tables in latest Google Chrome (tested Arch Linux and MacOS) or Safari
2. Open the browsers developer console
3. Completely delete the last paragraph in the editor (cursor should now be next to the `9`)
4. Press the down arrow key
*Observe that the cursor is now to the right of the table*
5. Press any letter
*Observe that the letter has been inserted twice, one inside the last cell of the table and one below the table*
*Observe that the cursor has jumped into the last cell of the table, to the right of the newly inserted character*
6. Press the backspace key to delete the character
*Observe that _both_ characters in the cell were deleted*
7. Press the down arrow to navigate below the table
*Observe the error message in the developer console*
> Error: Cannot resolve a Slate point from DOM point: [object Text],1
8. Press the backspace key to delete the character below the table
*Observer that you cannot delete the character*

**Expectation**
Perhaps behaviour like Firefox, where the cursor won't go outside the bounds of the table. Or for the typed character to only be inserted in the last cell of the table.

**Environment**
- Slate Version: 0.70.0
- Operating System: Arch Linux and MacOS
- Browser: Chrome, Safari

**Context**
I've tried overriding the `insertText` function in a plugin, to detect if the cursor position is outside the the table, with the intention of cancelling user input. But slate crashes with an error if I early return without calling the original `insertText`:
> Error: Cannot resolve a Slate point from DOM point: [object Text],1
My next idea is to implement a normalisation rule to automatically insert a node below a table...I will report back with the outcome of this.

### LINKED ISSUE #4738: DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
**Description**
Slate crash when use mouse to select multiple line of text and typing at the same time

**Recording**
capture

**Sandbox**
it is reproducible on the `https://www.slatejs.org/examples/richtext`

**Steps**
To reproduce the behaviour:
1. type something
2. use mouse to select multiple line of text and keep typing
3. repeat until it crash

**Environment**
- Browser: Chrome

Error log:

framework.316b7589431ec1c0e3c8.js:1 
DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at vo (https://www.slatejs.org/_next/static/chunks/framework.316b7589431ec1c0e3c8.js:1:99968)
    at ku (https://www.slatejs.org/_next/static/chunks/framework.316b7589431ec1c0e3c8.js:1:113221)
    at t.unstable_runWithPriority (https://www.slatejs.org/_next/static/chunks/framework.316b7589431ec1c0e3c8.js:1:3950)
    at $l (https://www.slatejs.org/_next/static/chunks/framework.316b7589431ec1c0e3c8.js:1:58091)
    at wu (https://www.slatejs.org/_next/static/chunks/framework.316b7589431ec1c0e3c8.js:1:111079)
    at ou (https://www.slatejs.org/_next/static/chunks/framework.316b7589431ec1c0e3c8.js:1:107210)
    at https://www.slatejs.org/_next/static/chunks/framework.316b7589431ec1c0e3c8.js:1:58382
    at t.unstable_runWithPriority (https://www.slatejs.org/_next/static/chunks/framework.316b7589431ec1c0e3c8.js:1:3950)
    at $l (https://www.slatejs.org/_next/static/chunks/framework.316b7589431ec1c0e3c8.js:1:58091)
    at Xl (https://www.slatejs.org/_next/static/chunks/framework.316b7589431ec1c0e3c8.js:1:58327)

### LINKED ISSUE #4771: Crash when last node delete
**Description**
When click on element which responsible to delete node crash happen with error message `Cannot find a descendant at path` only happen when delete last node

**Sandbox**
https://codesandbox.io/s/slate-node-remove-bug-3qw26?file=/src/editor/index.js

**Steps**
To reproduce the behavior:
1. Click on trash icon

**Expectation**
No crash

**Environment**
- Slate Version: [e.g. 0.72.3]
- Operating System: [MacOS]
- Browser: [Chrome, Safari]

**Context**
I think it can be related to this issue [https://github.com/ianstormtaylor/slate/pull/4616](https://github.com/ianstormtaylor/slate/pull/4616) because fix with timeout works in case I described

### LINKED ISSUE #4784: Moving certain nodes causes crashing of app
**Description**
- The error occurs when trying to drag a block with children outside of another block with children after they have both been added to the editor i.e. I can move children about freely from my sidebar before they are added to the editor
- From debugging, the move node operation is called successfully (this also works if I split this into a removeNode and InsertNode operation). Following the operation, a console log of the editor shows the node successfully moved into the new position and deleted from its old position. The error is called in `node_modules/slate/src/interfaces/node.ts` on line 341
- Looking at other examples, the path causing the issue is the path of the deepest child node being moved i.e. the passage’s paragraph child node. However, from the error it appears to be looking for it in the new editor children instead of the old
- We have already tried the ref and useState solutions described here: https://github.com/ianstormtaylor/slate/issues/4081.
- I have also tried to use `withoutNormalizing`

**Recording**
[Recording of issue](https://www.notion.so/jackow98/Slate-moving-certain-nodes-issue-3d5129da04b04cd985f541961a508b30#31be0e5565c44b62a0264f1ca5b9ef94)

**Sandbox**
https://codesandbox.io/s/strange-liskov-xjhrc?file=/src/App.tsx

**Steps**
- Drag a question block to the editor, drag a passage block into this question, drag the passage block out of the question block. This will cause the error. See video below
- The `moveNodes` operation is called on line 194 in `EditorScreen`

**Environment**
    "slate": "^0.66.5",
    "slate-history": "^0.66.0",
    "slate-react": "^0.67.0",
    "typescript": "^4.5.4",
"react": "^17.0.2",
    "react-beautiful-dnd": "^13.1.0",
    "react-dom": "^17.0.2",
    "react-scripts": "5.0.0",

**Context**
See [https://www.notion.so/jackow98/Slate-moving-certain-nodes-issue-3d5129da04b04cd985f541961a508b30](more detail here)

### LINKED ISSUE #4847: Editor will crash after input Chinese in FireFox
**Description**

Editor will crash after input Chinese and then input something more in FireFox.

**Recording**

https://user-images.githubusercontent.com/22148349/154881576-b71d7c3c-d81d-4e52-83c1-1a0e3e616c28.mov

**Steps**

reproduce:
1. Input something in Chinese like 你好(ni hao) in https://www.slatejs.org/examples/richtext in FireFox.
2. Then what user input will be repeated.
3. Input something more in the end of the text in editor.
4. Then editor will crash

**Environment**

- Browser: FireFox

### LINKED ISSUE #4851: Error thrown with Katex library: Cannot resolve a DOM point
**Description**
When using Slate and Slate-react along with [Katex](https://katex.org/) library —to render math within blocks—, the whole editor breaks when cursor reachs the rendered formyla, throwing an error and breaking the whole editor: `Cannot resolve a DOM point from Slate point: {"path":[0,1],"offset":1}`

**Recording**
https://user-images.githubusercontent.com/55709791/155305908-16de0603-3f56-43b6-a9ca-9985e4e447b6.mov

**Sandbox**
[https://codesandbox.io/s/slate-reproductions-forked-xg95mz](https://codesandbox.io/s/slate-reproductions-forked-xg95mz)

**Steps**
To reproduce the behavior:
1. Place cursor on the start of the text
2. Move cursor to the right until you reach the formula
3. See how editor breaks when cursor reachs formula
4. Repeat backwards, starting from end of text to start

**Expectation**
Editor should not fail, or log an error without throwing and crashing whole editor

**Environment**
- Slate Version: 0.72.3
- Operating System: OSX, Ubuntu
- Browser: Chrome, Firefox and Safari

### LINKED ISSUE #5014: CJK will output duplicated and crash on Firefox.(same on slatejs.org)
**Description**
slatejs can't working correctly with CJK IME on Firefox.
May output duplicated and crash on Firefox

**Recording**
https://user-images.githubusercontent.com/6832868/170954162-ec267ece-5d5f-4bba-a9d7-cbfa8cb37a0e.mov

**Steps**
To reproduce the behavior:
1. Open https://www.slatejs.org/examples/richtext
2. Select all default text, and delete it.
3. Typing any CJK text
4. CJK text be duplicated.
5. Try to delete duplicated text (i.e. any operation that delete the dom that contain duplicated text)  

**Expectation**

**Environment**
- Slate Version: [e.g. 0.59] 
- Operating System: MacOS
- Browser: Firefox 100.0.2

**Context**
Error message:
```
Cannot resolve a DOM point from Slate point: {"path":[0,0],"offset":7}
```
Is there exists some issue when slate updating DOM data ?

### LINKED ISSUE #5025: Sanity check for `Transform.select`
**Problem**
`Transform.select(editor: Editor, target: Location)` will use whatever `target` it is provided without checking it actually exists or if it points to leaf nodes.

**Solution**
Adding code that checks whether the target location exists and it is indeed leaf node. If it is not leaf node, it will use `Editor.first(editor: Editor, at: Location)`.

**Alternatives**
Not adding code for checking. User might be surprised when s/he tries to select node at `[0]` and editor will crash or do something unexpected.

### LINKED ISSUE #5066: Chinese IME sogou，cause crash by editing text before a inline-element
**Description**
Chinese IME sogou，cause crash by editing text before a inline-element. 
select -> input by sogou IME -> crash

**Recording**
rec

**Sandbox**

This bug can be reproduce by official online demo.

**Steps**
To reproduce the behavior:
1. select all characters befor `@R2-D2`
2. start input with sogou IME
3. then the bug show up.
4. See error in console:
`Uncaught Error: Cannot resolve a Slate point from DOM point: [object Text],2
    at Object.toSlatePoint ([example]-5e4f16b701ecada0.js:1:99830)
    at Object.toSlateRange ([example]-5e4f16b701ecada0.js:1:100498)
    at HTMLDocument.<anonymous> ([example]-5e4f16b701ecada0.js:1:117948)
    at g (674-c731382ea76bd727.js:1:1377655)
    at b (674-c731382ea76bd727.js:1:1377711)
    at HTMLDocument.z (674-c731382ea76bd727.js:1:1378030)
    at g (674-c731382ea76bd727.js:1:1377655)
    at S (674-c731382ea76bd727.js:1:1377930)
    at w (674-c731382ea76bd727.js:1:1377820)`

**Context**
The official demo can work smooth with Microsoft's chinese IME, so I'm not sure if this is a bug. But sogou IME is widely used in china. It  would be very helpful if we can fix this.

### LINKED ISSUE #5229: Fail Gracefully
**Description**
Any time I run into an error like "Cannot resolve DOM point/node" or an IndexSizeError or anything else pertaining to Slate logic, my app just crashes. Is it not possible to catch these errors (internally) somehow? To fail gracefully?

As a workaround I'm wrapping all vulnerable slate logic in try/catch blocks ... but it doesn't feel right. I'd like to know how others handle this for production apps.

**Expectation**
A quiet failure, possibly returning null values, warning, or something else.

**Environment**
Slate Version: all
Operating System: all
Browser: all
TypeScript Version: all

### LINKED ISSUE #5273: Crash when deleting last paragraph ending in `\n`
**Description**

This crash occurs in Firefox when the editor has at least two paragraphs and the last paragraph ends in `\n`. When selecting all text and deleting it, the editor crashes with the following error message:

```
Uncaught (in promise) Error: Cannot resolve a DOM point from Slate point: {"path":[0,0],"offset":1}
    toDOMPoint https://zm8y60.csb.app/node_modules/slate-react/dist/index.es.js:508
    toDOMRange https://zm8y60.csb.app/node_modules/slate-react/dist/index.es.js:515
    setDomSelection https://zm8y60.csb.app/node_modules/slate-react/dist/index.es.js:2406
    Editable https://zm8y60.csb.app/node_modules/slate-react/dist/index.es.js:2419
    React 8
    unstable_runWithPriority https://zm8y60.csb.app/node_modules/scheduler/cjs/scheduler.development.js:255
    React 5
    onChange https://zm8y60.csb.app/node_modules/slate-react/dist/index.es.js:3677
    apply https://zm8y60.csb.app/node_modules/slate/dist/index.es.js:232
```

**Recording**

https://user-images.githubusercontent.com/4272090/214837667-36af22c8-da80-4ff4-bb57-f37ca0b171be.mov

**Sandbox**
https://codesandbox.io/s/slate-reproductions-forked-zm8y60

**Steps**
To reproduce the behavior:
1. Go to the sandbox
2. Select all text (command+a)
3. Press backspace

**Environment**
- Slate Version: 0.88.1
- Operating System: macOS
- Browser: Firefox only (no error in Safari or Chrome)

**Context**
This bug was introduced in `slate-react@0.82.0` by PR #4988. (I discovered this by testing timestamped NPM releases; `0.82.0-2022629111111` is the first one to contain the bug.)

### LINKED ISSUE #5355: Slate crashes when arrowing into or out of table with colgroup/col nodes
### Discussed in https://github.com/ianstormtaylor/slate/discussions/5221

<div type='discussions-op-text'>

<sup>Originally posted by **instadrew** December 12, 2022</sup>
Hello Slate Community,

I have run into what I think may be a Slate bug, but I'm not sure and thought I'd ask here first.  I'm trying to render a table in Slate with a `<colgroup>` and several `<col>` sub-elements (there are styles on the `<col>` sub-elements in what I'm really doing, but not in the code sandboxes cited below as they are not needed for demo purposes).  The rendering part works fine.  The problem I have run into is that if I place the cursor in the top left cell of the table, and then press the left arrow key repeatedly so as to move the cursor out of the table , the Slate editor crashes and I get the following:

```
Error
Cannot resolve a DOM node from Slate node: {"text":""}
```

The same thing occurs if I place the cursor before the table and then press the right arrow key repeatedly so as to move the cursor into the first table cell.

Another table without a colgroup, but otherwise the same structure as the above, does not produce this error.

I'm guessing that Slate is trying to map a text child node from one of the col nodes to an HTML node, doesn't find one, and then crashes.  I don't know what else to do with this since all nodes other than text nodes require a children property.

A demo of the above can is available here: https://codesandbox.io/s/silly-ptolemy-5ho3uz

Note that I have also tried defining the col nodes as text nodes, but this creates a DOM structure that  looks invalid to me (i.e. `<span data-slate-node="text"><col></span>` for each col node) and produces the same type of error anyway:

```
Error
Cannot resolve a DOM point from Slate point: {"path":[2,0,0],"offset":0}
```

A demo with the col nodes defined as text nodes can be found here: https://codesandbox.io/s/muddy-cdn-b098mm

Is this a Slate bug, or am I doing something wrong with how I have attempted to define and render the colgroup and col nodes?  Are there any workarounds to prevent Slate from crashing in these scenarios?</div>

### LINKED ISSUE #5375: Bug occurs when inputting Chinese with IME at the end of an inline element
**Edge desktop browser**
- Slate Version: 0.91.4
- Slate-react Version: 0.92.0
- Browser: Edge 111.0.1661.44 (Official build) (64-bit)

https://user-images.githubusercontent.com/111119927/227405680-29b8d8d6-8a82-4e92-a4e6-9d2ba113e8c3.mp4

In Edge, it just crashes when you input at the end of an inline element

**Firefox desktop browser**
- Slate Version: 0.91.4
- Slate-react Version: 0.92.0
- Browser: Firefox 111.0.1 (64-bit)

https://user-images.githubusercontent.com/111119927/227405840-04a215a4-f110-4b2d-a9a6-b62999dec22e.mp4

In Firefox, it's just unable to input inside an inline element when you're at the end of it

_Rich text editor is really tricky :)_

zbeyens:
**Description**
This pull request primarily focuses on enhancing error handling by introducing **`undefined`** checks, refining return types for functions, and implementing a global **`ErrorLogger`** interface. It also introduces an **`editor.onError`** function to manage errors in functions that depend on the **`editor`**.

**Issue**

Fixes #3243
Fixes #3558
Fixes #3568
Fixes #3586
Fixes #3641
Fixes #3705
Fixes #3708 
Fixes #3723
Fixes #3777
Fixes #3811
Fixes #3870
Fixes #3878
Fixes #3900
Fixes #3918
Fixes #3948
Fixes #4001
Fixes #4081
Fixes #4088
Fixes #4094
Fixes #4136
Fixes #4323
Fixes #4485
Fixes #4561
Fixes #4564
Fixes #4581
Fixes #4643
Fixes #4658
Fixes #4738
Fixes #4771
Fixes #4784
Fixes #4847
Fixes #4851
Fixes #5014
Fixes #5025
Fixes #5066
Fixes #5229
Fixes #5273
Fixes #5355
Fixes #5375

For a good part of these issues: this PR is only fixing the "crashing" part but not the underlying issue – the error will be stored in `editor.errors`. So it's **not** a complete fix, but I think it's important to close these issues until the description is updated (i.e. the editor does not crash anymore).

**Major Changes**

- Added **`editor.onError`** function to handle errors in functions that depend on **`editor`**.
- Introduced a global **`ErrorLogger`** interface to handle errors in functions not depending on **`editor`**.
- No errors are thrown by default, addressing community feedback from issue #3641.
- Updated return types of several functions to include **`| undefined`** and replaced **`throw new Error()`** statements with either **`editor.onError()`** or **`ErrorLogger.onError()`**.
- Implemented conditional checks for variables before accessing them to prevent crashes and improve code stability.

**Example**

```tsx
import { ErrorLogger } from 'slate'

ErrorLogger.addErrorHandler(error => {
  throw new Error(error.message)
})
```

You can also filter errors by type:

```tsx
ErrorLogger.addErrorHandler((error) => {
  // handle error
  if (error.type === 'Node.get') {
    throw new Error(error.message)
  }

  // access errors array
  console.log(ErrorLogger.getErrors());
  
  // reset errors array
  console.log(ErrorLogger.resetErrors());
});
```

**Context**
The motivation behind these changes is to improve code stability, maintainability, and robustness by providing more error control to users. By introducing the global **`ErrorLogger`** interface and the **`editor.onError`** function, this pull request aims to make error handling more flexible and manageable. Some behaviors that previously threw an error may now continue to work without crashing the editor, which could potentially cause regression.

# **Error Handling**

Error handling in Slate is designed to provide more control to users while maintaining code stability, maintainability, and robustness. Slate introduces a global **`ErrorLogger`** interface and an **`editor.onError`** function to handle errors in a more flexible and manageable way. Utilize these error handling mechanisms to tailor your error management based on your specific use case and requirements, ensuring a smoother and more stable experience for your users. For example, you may want to throw errors in development, but in production, you would prefer to have non-blocking errors for your users.

## `ErrorLogger`

The global **`ErrorLogger.onError`** function is used to handle errors in functions not depending on the **`editor`**, like `Node.get`. To set custom error handling behavior, call **`ErrorLogger.addErrorHandler()`**.

To throw errors on invalid operations:

```tsx
import { ErrorLogger } from 'slate'

ErrorLogger.addErrorHandler(error => {
  throw new Error(error.message)
})
```

You can also filter errors by type:

```tsx
ErrorLogger.addErrorHandler((error) => {
  // handle error
  if (error.type === 'Node.get') {
    throw new Error(error.message)
  }

  // access errors array
  console.log(ErrorLogger.getErrors());
  
  // reset errors array
  console.log(ErrorLogger.resetErrors());
});
```

## `editor.onError`

The **`editor.onError`** function is used to handle errors in functions that depend on the **`editor`**.

To set custom error handling behavior for the editor, override the **`editor.onError`** function.

To throw errors on invalid operations:

```tsx
editor.onError = error => {
  throw new Error(error.message)
}
```

You can also filter errors by type:

```tsx
// throw only for `move_node` operation
editor.onError = error => {
  if (error.type === 'move_node') {
    throw new Error(error.message)
  }
}
```

## `editor.errors`

By default, Slate pushes errors to the `editor.errors`, `ErrorLogger.getErrors()` arrays. You can actively manage this array to keep it empty by implementing your own error handling strategies. This allows you to effectively monitor and address errors that occur during the editor's operation, contributing to a stable user experience.

# Migration

If you want to keep the previous behavior, here is the quickest migration using non-null assertion (estimating to a couple of minutes):

Throw an error like before:

```tsx
import { ErrorLogger } from 'slate'

ErrorLogger.addErrorHandler(error => {
  throw new Error(error.message)
})

editor.onError = error => {
  throw new Error(error.message)
}
```

Here is the list of APIs that now return `| undefined`. Find usages for each of these and insert a non-null assertion. For example: `Path.next(...)` to `Path.next(...)!`

```tsx
Path.next
Path.parent
Path.previous
Path.relative
Editor.edges
Editor.end
Editor.first
Editor.last
Editor.leaf
Editor.node
Editor.parent
Editor.path
Editor.point
Editor.range
Editor.start
Node.ancestor
Node.child
Node.children
Node.common
Node.descendant
Node.first
Node.get
Node.last
Node.leaf
Node.parent
ReactEditor.findPath
ReactEditor.toDOMNode
ReactEditor.toDOMPoint
ReactEditor.toDOMRange
ReactEditor.toSlateNode
ReactEditor.findEventRange
```

The alternative is to wrap each of these into a function that does not return `| undefined` (alias type) but this would take more time to refactor.

**Checks**


docs/Summary.md
```diff
@@ -25,7 +25,7 @@
 - [Rendering](concepts/09-rendering.md)
 - [Serializing](concepts/10-serializing.md)
 - [Normalizing](concepts/11-normalizing.md)
-- [TypeScript](concepts/12-typescript.md)
+- [TypeScript](concepts/13-typescript.md)
```
SmilinBrian, line 28: Should add the link to `12-error-handling.md` here

zbeyens: Good catch, thanks!

SmilinBrian:
Wow, this should be a huge improvement!  🎉 Thank you! 🙏🏼 

It's a massive amount of work which I unfortunately do not have time to review in detail this week. What I've looked at so far looks good, although there are a few places/cases that used to crash which now "silently succeed" that I'll have to think carefully whether are worthy of passing to the error handler/logger or not… in some cases it is really hard to decide whether something is "reportable" or not.

zbeyens:
Thanks for the review @SmilinBrian ! This is the "possible regression" part I was thinking to. Unfortunately, we don't have unit tests that would help to identify these **critical errors**. The goal is to have an editor that never freezes on user inputs – e.g. even if the selection is buggy. One place that is obvious to throw is the slate-react hooks. One way of handling a (less) critical error is to early return. On the other hand, we could improve the user experience by avoiding early return for "soft errors" (array queries https://github.com/udecode/slate/blob/63350f8b3828108c70e21db1bedf28b8456193ab/packages/slate/src/interfaces/node.ts#L386). 

I think we can just wait for issues to emerge to identify the critical errors.

BrentFarese:
This is a good amount of code changes so thank you @zbeyens for putting in the effort!  I have 1 larger question for the community, which is this the preferred route for catching errors that are thrown by Slate?  There a couple downsides of this PR that I'm not too keen on:

1. The type changes that include `| undefined` are going to result in a lot of breaking changes for folks migrating.  The developer now needs to decide whether to throw errors in their own code, or just do nothing and "silently fail" if a given Slate func returns `undefined`.
2. Because funcs now return `| undefined` there can be silent failures that **_should_** be fixed in the codebase.  Is it better to loudly fail so the codebase gets better vs. silently fail?  Developers can still handle errors in their code properly w/o crashing their app...
3. Have any other lower-lift/backwards compatible ways of doing this been considered?

On (3), an idea I had (and something we do at [Aline](https://www.aline.co/)) is wrap all Slate queries/transforms in error handling logic, and then only use the wrapped queries/transforms in the app code base.  Here is some pseudocode example:

```
import { Transforms } from 'slate';

type SlateTransformType = typeof Transforms;
type WrappedSlateTransforms = {
  [K in keyof SlateTransformType]: (
    ...args: Parameters<SlateTransformType[K]>
  ) => ReturnType<SlateTransformType[K]> | undefined;
};

export const SlateTransformsWithErrorHandling = Object.keys(Transforms).reduce<WrappedSlateTransforms>(
  (acc, transformName) => {
    const transform = Transforms[transformName];

    return {
      ...acc,
      [transformName]: (...args: any[]) => {
        try {
          return transform(...args);
        } catch (e) {
          console.error(e);
        }
      },
    };
  },
  Transforms
);
```

Then any dev wanting errors to be caught could use `SlateTransformsWithErrorHandling` instead of `Transforms` directly.  This might be a decent approach, but haven't thought through all the ramifications.

Basically, changing the type of everything to `| undefined` I think is going to require a lot of code changes when users migrate and cause some silent errors that are not good, no?  It certainly would prevent Aline from migrating until we can dedicate time to make the changes needed in our codebase to handle `| undefined`.

Thanks for checking out the feedback!

zbeyens:
Thank you for your feedback and for expressing your concerns. I understand that introducing **`| undefined`** in return types may lead to a more challenging migration process for larger projects. However, I believe that including these in Slate core is essential for several reasons:

1. **Improved error handling**: By returning **`undefined`** instead of throwing errors, we provide developers with greater control over how their applications handle errors. They can decide whether to throw errors, handle them gracefully, or log them for further analysis. This flexibility enables more robust and customized error handling strategies in different environments (e.g., development vs. production).
2. **Enhanced code stability**: Previously, unhandled errors in Slate could cause the editor to crash or exhibit unexpected behavior (as evidenced by the numerous linked issues). Implementing **`| undefined`** and conditional checks improves code stability and reduces the likelihood of crashes, leading to a more reliable and stable user experience.
3. **Simplified debugging**: The introduction of **`| undefined`** encourages developers to consciously address potential errors in their code. By being aware of possible issues and handling them explicitly, developers can more easily identify and resolve bugs in their codebase. This approach contributes to better overall code quality and maintainability.

While I acknowledge your concerns, I believe that the benefits of this approach outweigh the drawbacks.

Your idea of wrapping Slate queries/transforms in error handling logic is an interesting alternative and could provide a helpful solution for developers who prefer to maintain their current error handling behavior without migrating to the new **`| undefined`** return types. As a way to move forward with the improvements mentioned above, we could also consider wrapping the new behavior so that all Slate queries do not return **`| undefined`** (could be just a type change). Silent errors would be the new default but could be reverted to error throwing by just overriding **`onError`** as mentioned in the changeset. What do you think?

12joan:
One thing that would improve both DX and UX is to fail loudly/silently depending on whether we're in development or production. In development, throwing errors is good because they can be fixed immediately, but production users don't wantt the app they're using crashing and potentially losing data, so failing silently (and ideally reporting via telemetry) is better.

Would this PR support the above pattern in its present state?

packages/slate/src/interfaces/error-logger.ts
```diff
@@ -0,0 +1,58 @@
...
+    _errorStore.onError(error)
+  },
+
+  resetErrors() {
+    _errorStore.errors = []
+  },
+  setOnError(onError) {
+    _errorStore.onError = onError
+  },
+}
```
12joan, line 58: An `addEventListener` pattern would work quite well here to allow developers to add multiple error handlers.

BrentFarese:
@zbeyens I think in general you are right that letting developers figure out how to handle the errors is better.  I can review the code in more detail later.  I think in general we just want to ensure `onError` or whatever handler is passed the "errors" Slate used to `throw`, and if we're doing that then at least we're not hiding anything/silently failing.  The errors are still discoverable.

zbeyens:
@BrentFarese Sounds good, every single `throw` has been replaced by `onError` so we should be good on that point. I've also added a few `onError` calls to `apply` – this pattern allows us to add much more logging without affecting the user experience.

--- 

Thanks to @12joan, I have updated `ErrorLogger` to be a class supporting multiple error handlers – description/doc/changeset updated.

bryanph:
First of all, thanks for tackling this, it seems like a great improvement.

I like the error interface, that seems to be a win. In general, if Slate doesn't have to throw, it's better to just have some way to recover for the user.

I'm a little torn on the changes to returning undefined everywhere. The main downside I can see is that you will start needing non-undefined checks everywhere now for typed code with this change when calling the modified functions. I guess the alternative would be to try/catch in the slate code and pass the error to the logger wherever we think the throw should be recoverable?

zbeyens:
Thanks for your review @bryanph.

This is indeed required to support `onError` handlers that are not throwing.

Here is the quickest migration using non-null assertion (estimating to a couple of minutes):

## Migration

Throw an error like before:

```tsx
import { ErrorLogger } from 'slate'

ErrorLogger.addErrorHandler(error => {
  throw new Error(error.message)
})

editor.onError = error => {
  throw new Error(error.message)
}
```

Here is the list of APIs that now return `| undefined`. Find usages for each of these and insert a non-null assertion. For example: `Path.next(...)` to `Path.next(...)!`

```tsx
Path.next
Path.parent
Path.previous
Path.relative
Editor.edges
Editor.end
Editor.first
Editor.last
Editor.leaf
Editor.node
Editor.parent
Editor.path
Editor.point
Editor.range
Editor.start
Node.ancestor
Node.child
Node.children
Node.common
Node.descendant
Node.first
Node.get
Node.last
Node.leaf
Node.parent
ReactEditor.findPath
ReactEditor.toDOMNode
ReactEditor.toDOMPoint
ReactEditor.toDOMRange
ReactEditor.toSlateNode
ReactEditor.findEventRange
```

The alternative is to wrap each of these into a function that does not return `| undefined` (alias type) but this would take more time to refactor.

bryanph:
@zbeyens My comment was more about another approach, namely keeping the throwing mechanics and instead of handling the undefined return, try/catching in the code and passing to the error logger instead from there inside slate.

However, I think returning `undefined` does make sense for the methods you've modified and I agree it can result in more type-safe code by requiring the user to handle the undefined case. So I'm all for this change 👍🎉

piotrrussw:
Hi @zbeyens any updates on this? `onError` will be very helpful!

zbeyens:
Based on the feedback we've received, we are leaning towards the idea that `onError` should throw by default, allowing the developer to override this behavior if needed. 

This approach should reduce the risk of silent failures that might result in a corrupted state and potential data loss, which is a key concern. Besides that, we agreed on:

- the overridable `onError` function, so that developers can provide their own error handling mechanisms if desired.
- returning `undefined` instead of throwing exceptions in Slate API.
- moving the `onError` handling from the Slate API to the core of Slate. That way, the error handling responsibility is on the consumer side, not the API itself. For instance, if a function like `Node.parent` cannot find a parent, it should return `undefined`, and the method caller should handle that edge case. This makes the API more intuitive for the end user.

I'll try to find some time in the coming weeks to proceed.

dscape:
As an app developer using Slate, merging this pull request has no impact on our app and requires no migration. However, Slate lacks error handling, leading to user confusion and delays in addressing even minor issues. 

This pull request is crucial for managing Slate and providing better error handling to users. We're considering using it as a patch due to Slate's current limitations in a production environment.

Kudos to @zbeyens for the valuable contribution.

skogsmaskin:
I too want to kudos you @zbeyens for this work! The impossibility to catch slate-react's toDOMPoint errors has been such a pain point for a [very long time](https://github.com/ianstormtaylor/slate/issues/3641#issuecomment-620581011) making it super hard to write a realtime collaborative applications using (especially) `slate-react`. `slate` in it self has been less problematic with this.

So looking forward to getting this merged. I would approve it as it is now, as @dscape said above the changes are non-breaking and adds a ton of value already.

BTW: I'm wondering if we should start using maturity indicators in the codebase (like tag them `@public`, `@beta`, `@alpha` tags), especially with the API suface.  This way new features (like this) could be merged earlier, and get some real world usage and experience before finally moving them into @public at some point.

dscape:
@ianstormtaylor any updates on this?

dylans:
> @ianstormtaylor any updates on this?

I was away for a couple of weeks on holiday, but I will review this asap.  Ian isn't currently active, so it's pretty much up to me to make sure this addresses the earlier concerns that were raised (there were conversations on Slack that led to the recent changes).

dscape:
Thanks @dylans appreciate it!

dscape:
@dylan Can you give an update?

Three months to merge such an important piece of work, why?

gdehmlow:
+1 on this. For production error logging, we don't want to leak potential PII from the editor content in the error message, and having to scrub this information at our top level error logger is a pain and error prone. Thank you for this work!!

dylans:
> @dylan Can you give an update?
> 
> Three months to merge such an important piece of work, why?

Slate is used by a lot of people and big changes take time to think through.

We had issues with the first version (felt it was squelching errors rather than actually solving the issue). I'm going to give it one more review and land it soon.

@zbeyens there's a small merge conflict to resolve after the recent range selection fix for Firefox.

dylans:
Ok, so I am definitely struggling with this one for a few reasons:

It's several types of changes intermixed, leading to a high chance of breaking everyone.

I am reminded of the comments in https://github.com/ianstormtaylor/slate/issues/3641#issuecomment-632869462 and https://github.com/ianstormtaylor/slate/issues/3641#issuecomment-632884427

In our app, we simply place an ErrorBoundary around the render function. So for example, if you have the Cannot resolve a DOM point from a Slate point issue, e.g. at https://codesandbox.io/s/inspiring-tristan-zq8lj4?file=/src/App.tsx

You might do an ErrorBoundary like this (ignore some of the older React patterns, just following the CodeSandbox example):

```<ErrorBoundary editor={editor}>
        <Slate editor={editor} value={value} onChange={onChange}>
          <Editable onDOMBeforeInput={onBeforeInput} />
        </Slate>
      </ErrorBoundary>

which would be implemented like this:

```import { Editable, Slate, withReact } from "slate-react";
import { Node, createEditor, Transforms } from "slate";
import React, { useCallback, useMemo, useState } from "react";
import { withHistory } from "slate-history";
class ErrorBoundary extends React.Component {
  static getDerivedStateFromError(error) {
    return {};
  }
  componentDidCatch(error, info) {
    // if selection issue then
    Transforms.deselect(this.props.editor);
  }
  render() {
    return this.props.children;
  }
}```

That said, this still wouldn't let you fully recover from the error per se.  So how do you recover from the state getting corrupted when you run into an error? ( I'm not sure the recovery attempt in this PR actually works).

As far as I can tell this doesn’t actually recover from errors, but might eat exceptions or stop the editor and leave it in a random state that cannot easily be recovered from, especially if you then allow edits to continue.

I would recommend that we start small... maybe we can start by relaxing toDomPoint or write better tests to reduce the number of scenarios where it breaks. So far example if deselect fails you might call:

```Transforms.select(this.props.editor, {
      anchor: { path: [0, 0], offset: 0 },
      focus: { path: [0, 0], offset: 0 }
    });```

Since you have no way to recover anyway.

So what I keep asking myself is what does this PR achieve that an ErrorBoundary does not achieve?

zbeyens:
The new `onError` approach seeks to provide a granular, customizable error-handling API that extends beyond what the `ErrorBoundary` approach can provide. While an `ErrorBoundary` indeed helps capture generic errors, it doesn't provide enough context for handling that error.

In contrast, the `onError` handler provides detailed information about the error context, including the unique error key, underlying error type, and additional operation data. This granular information allows developers to create targeted error recovery strategies that can **help** maintain a stable editor.

Moreover, the `onError` approach provides an optional `recovery` value (**experimental!**) to facilitate an immediate fix for many cases, such as `isEnd` returning `false`. This combined with the data attribute should offer enough flexibility for quick fixes without compromising the editor state. Then I expect many people patching slate from the fixes they could write thanks to the new API.

Remember, we want to keep the old behavior as the default until we are sure the new one works well (**`editor.strict = true`**). So there is no "high chance of breaking everyone" by default. This is surely a more dangerous strategy than crashing the editor and/or resetting its state, this is why we actually need this so we can make it less dangerous with the incoming patches.

I'm happy to hear more feedback and suggestions.

pgte:
As the CTO and developer of a product that has been using Slate for more than 2 years, I can say that this fix would improve the overall robustness of our code. This would allow us to recover from commonly seen errors where, right now, we can only react to by resetting the state, which can be an awful user experience for larger documents.

dscape:
This is how it feels to be a slatejs developer without error handling.

https://github.com/ianstormtaylor/slate/assets/8577/65b41dd5-eeaf-4afd-a9e8-b940f3f08f38

Lot's of users affected, incredibly limited options to tackle it. Most of the bugs are between browser support (and quirks), React, and Slate. Potentially also a plugin ecosystem like Plate. For instance, there's still some lack of standard about triple clicks handling in modern browsers. If there's an inconsistency between a browser and slate and slate has a bug, you won't be able to support your user (this is a real bug that was recently patched in slate)

Let's say you have a editor and do what @dylans suggested: You set an error boundary, but it lacks context. Because we don't know full cause, all we can do is re-render and check that we don't get into a loop. Eventually we need to tell the user to refresh the page.

From a user perspective, they see a "strange behaviour and re-render" and then an error page. "Ups, its broken" -- You will loose users, and there's little you can do about that. 

If we had error handling we could, as app developers, say: if its a selection error let's not crash the editor.

These behaviours are specific to the application, so they should be the application developer responsibility. With slate this is just not possible.

I have sympathy and kudos for the maintainers of slate. It's a very hard thing to maintain and create. It builds on between a very old technology in html and things that often aren't even standard, like triple clicks. In between there's an expectation that the browser must support editing similar to native apps, or better. This includes voids vs. non voids, cell re-renders, and lot's of complexity to maintain. In our case using CRDTs on top. We know this is hard, we know it goes across lot's of layers.

But imo not having error handling and letting applications be at the mercy of the bugs is not acceptable. It just doesn't allow us to respond to people using the tech, and therefor should be a big issue for people when they choose slate over an alternative.While I'm sure that this PR could be improved, and more will be learned once it is merged and it will evolve, i don't understand why this is being kept at bay. 

@zbeyens did high quality work here, and is open to feedback and recommendations. What else can be achieved here?

In the meantime, at Decipad we are considering having to maintain this as a patch on top of slate so we can support our users. I hope the maintainers reconsider their position here, or offer an alternative that empowers Slate users.

dylans:
@zbeyens and I are good collaborators and have worked together on Plate for a long time.

For something as big and widely used as Slate, I strive to err on the side of the smallest most incremental changes possible so that we can revert quickly if necessary. This is a big change that would not be easy to reverse in its current form.

Fundamentally what makes me uncomfortable is multiple types of changes on different levels via one large PR, and slate-react still doesn't have a good test library which makes me more conservative than I would like to be.

For example, there are some fixes in the PR that likely fix a few errors or make it clearer why something is an error. I would suggest a PR first for those and land those.

The change to `| undefined` on a lot of things is a pretty big change. I would do that as a separate PR. Having a decent test library for slate-react would make me happier as well.

Then I would sort of ask, could we make better use of throw new Error vs. onError to provide additional context as an intermediary step. I'm not opposed to onError, just wondering how much we can get out of the existing paradigms first.

It's clear that some can get more out of ErrorBoundary ( https://github.com/bugbakery/transcribee/pull/337/files ) at least for selection errors.

Where I remain skeptical is error recovery rather than fixing things that cause errors, but I have an open mind and am not opposed to it. But I am a lot more comfortable getting there incrementally, whether it's as proposed above or in some other form.

zbeyens:
I can understand this PR has a few risks we could optimize from Dylan’s review. I won’t have free time to split that PR for a few weeks so my recommendation is to either do a test release (Slate has a workflow to do that through a comment) so a few consumers could test it and give their feedback for the stable release. Or someone else could voluntarily split that PR.

dscape:
I don't have the tech skill to support this split, but at Decipad we would be happy to support a bounty if it helps the project.

@dylans appreciate the response. As someone that used to be involved and maintain a lot of open source projects I hope you understand we are not, in any way, not recognising the work. Just trying to raise the awareness of the issue, as your day to day using slate might be different from ours.

beorn:
Any update on this?  It would be great to have a way to catch errors in Slate.

zbeyens:
We need to split this PR. Since it's no longer my priority, anyone is welcome to take care of it.

dylans:

PR #5409, opened 2023-05-02
docs(walkthroughs): fix broken code example

drazik:
**Description**

In the Walkthroughs / Defining Custom Elements documentation page, the last code example didn't work for me: the type of a node didn't change when I pressed the Ctrl+\` key combination. I looked at the differences between this example and the previous one that was working, and found that the `match` condition of the last example for missing a part that was there in the example before.

I added it and it fixed the example.

**Issue**
It doesn't seem there is an existing issue for that.

**Example**
I don't think it is needed for this change

**Context**
I don't think it is needed for this change

**Checks**


dylans (approved):
Thanks @drazik !

PR #5415, opened 2023-05-05
Consistent insert* methods

### LINKED ISSUE #5414: Harmonize behavior of insert* methods
We currently have a fair bit of inconsistency and confusion with our insert methods, whether they are the version from Editor or Transforms.

I think we should make sure the following happens:

* All insert methods should optionally accept a location
* If no location is provided and there's no selection within the editor, we should either throw or insert at a consistent location (start or end of the document?)
* Update the documentation

zbeyens:
Fixes #5414


dylans (approved):
Perfect, thanks. Please add the changeset and we'll land and release.

Issue #4992, opened 2022-05-16
`value` should be renamed to `initialValue`


macrozone:
**Problem**

https://github.com/ianstormtaylor/slate/pull/4922 introduced a massive breaking change for consumers of `slate`, e.g. for https://github.com/react-page/react-page

Problem is that the meaning of `value` has changed, slate is unfortunatly no longer a controlled component (which was a far better design overall). 

**Solution**

`value` needs to be renamed to `initialValue` to avoid confusion. As slate is still on a 0.x version, this should be done in a breaking change as https://github.com/ianstormtaylor/slate/pull/4922 was already breaking.

this makes a clear cut and opens up better apis to partially restore the old, more powerful api.

**Alternatives**

keeping `value` as it is is not a good idea. This will just spread further confusion. When its named `value` it should be controlled, uncontrolled should not be named `value`. 

**Context**

I still have to figure out how to restore the old behavior. https://github.com/ianstormtaylor/slate/pull/4922 briefly mentions a way https://github.com/ianstormtaylor/slate/pull/4922/files#diff-641a49ba4a3c2cb032bc8ae94d51c85a40ad05dbe2c8488dc98dc8379f92c306R85 but its not enough to understand how to do it. Any help is appreciated

doei:
I'm new to Slate, and I'm guessing this is the reason why I can't get the onChange prop to work ?

I made a [simple codebox example](https://codesandbox.io/s/dazzling-bird-3yzvs2?file=/src/App.js) of what I'm trying to do.

Does this mean the deserialisation should always be done before rendering the editor ?

macrozone:
> I'm new to Slate, and I'm guessing this is the reason why I can't get the onChange prop to work ?
> 
> I made a [simple codebox example](https://codesandbox.io/s/dazzling-bird-3yzvs2?file=/src/App.js) of what I'm trying to do.
> 
> Does this mean the deserialisation should always be done before rendering the editor ?

kindof yes, its really sad that slate is now a uncontrolled component. That will lead to a lot of unexpected bugs.

In your codebox, this workaround works:

```


import { Slate, Editable, withReact, ReactEditor } from "slate-react";

// in your useEffect ...

    console.log("setting value", res);
        editor.children = res.result;
        ReactEditor.focus(editor);

```

macrozone:
This workaround not always works. Its really bad that slat changed to an uncontrolled Components. I think this is an antipattern in many case because its just a pitfall of failure.

[...]

PR #5421, opened 2023-05-12
Rename `<Slate>` component `value` prop to `initialValue`

e1himself:
**Description**

The Slate editor is in fact an uncontrolled component, but the public API deceives into thinking it's a controlled input. This is because of the `value` and `onChange` is a common React convention for controlled inputs, while uncontrolled inputs use `initialValue` instead.

**Issue**
Fixes https://github.com/ianstormtaylor/slate/issues/4992

**Example**

Before: 

```tsx
<Slate editor={editor} value={value} />
```

After:

```tsx
<Slate editor={editor} initialValue={value} />
```

**Checks**

### DIFF
--- a/.changeset/tiny-suns-lick.md
+++ b/.changeset/tiny-suns-lick.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+Rename `<Slate>` component prop from `value` to `initialValue` to emphasize uncontrolled nature of it
--- a/docs/concepts/12-typescript.md
+++ b/docs/concepts/12-typescript.md
@@ -53,7 +53,7 @@ const App = () => {
   const [editor] = useState(() => withReact(createEditor()))
 
   return (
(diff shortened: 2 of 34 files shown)


dylans (approved):

dylans:
Looks good and makes sense to me. Going to leave it open for a couple of days in case anyone has other feedback.

atanasov-deyan:
apologies for jumping into an already merged PR, but wouldn't this be a breaking change, as the API is changing? This is not explicitly highlighted anywhere, tbh. wdyt?

pchr-srf:
What @atanasov-deyan says, just spent more time than I care to admit on trying to find out why stuff wasn't working. Turns out a minor change only visible in https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/CHANGELOG.md#0950 was the culprit, fixed with a oneliner 🙈

Issue #5386, opened 2023-04-04
Placeholders not placed correctly on Safari 16.4


ashbork:
**Description**
Placeholders seem to be placed a line lower than they should be. This behavior didn't occur before 16.4

![image](https://user-images.githubusercontent.com/82605611/229818852-6d33c255-5701-487d-947a-627978bbb76d.png)

**Sandbox**
[One of Slate's examples](https://www.slatejs.org/examples/richtext)

**Steps**
To reproduce the behavior:
1. Go to ['the Slate playground](https://www.slatejs.org/examples/richtext)
2. Empty the text area
3. See the error

**Expectation**
The placeholder should be placed at the top of the text area.

**Environment**
- Slate Version: ^0.87.0 and current
- Operating System: macOS
- Browser: Safari 16.4

Noothan-am:
hey @ashbork is this issue fixed ?

ashbork:
You'd likely find an associated PR if it were, I can't check that now, you could just follow the reproduction steps

Noothan-am:
i tried to reproduce it and i got this, iam not able to figure out what is problem here could you please give me some more info about this issue ?

![image](https://user-images.githubusercontent.com/100135382/236182984-cc7bb1b6-36f2-490c-8520-479efe321693.png)

ashbork:
It's still reproducible for me and apparently 3 others, idk what to tell you - maybe the environment is different

horacioh:
> i tried to reproduce it and i got this, iam not able to figure out what is problem here could you please give me some more info about this issue ?
> 
> ![image](https://user-images.githubusercontent.com/100135382/236182984-cc7bb1b6-36f2-490c-8520-479efe321693.png)

the placeholder looks in the middle because you have the alignment set to center.
I see this issue but the placeholder renders below the cursor. Does this happens to anyone else?
<img width="948" alt="Screen Shot 2023-05-10 at 13 05 11" src="https://github.com/ianstormtaylor/slate/assets/725120/3ca7b71b-72e5-40cf-a6f8-30baf5315865">

ashbork:
That is the original issue I reported, yes

horacioh:
> That is the original issue I reported, yes

@ashbork the placeholder have some inline styles that makes it a block element, can you try to uncomment the block styles (and even the width) to see if the problem still persist?

ashbork:
Nope, removing `width` or `display:block` doesn't do anything to fix the alignment.

[...]

PR #5423, opened 2023-05-12
fix(slate-react): add top css to placeholder

horacioh:
**Description**
Placeholder position in Safari 16.4 is not correct. this PR aims to fix that. I also checked that in other versions of Safari there's no regression.

**Issue**
Fixes: #5386 

**Example**
Before:
image

After:
image

**Checks**

### DIFF
--- a/.changeset/few-queens-fold.md
+++ b/.changeset/few-queens-fold.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+fix placeholder position in Safari 16.x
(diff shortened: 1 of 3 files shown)


dylans (approved):

Arkellys:
Isn't this change too specific to Slate docs? It breaks my placeholders.

image

jasontlouro:
image

Breaks ours as well, completely ignores padding on the editor.

jasontlouro:
Sounds like a job for...

image

😅

heyitsaamir:
+1. Can we revert this? It's leading to a worse experience for _most_ cases.

Issue #5442, opened 2023-06-09
Wrong caret position when using chinese IME with async state change.


OldDream:
**Description**
In macos, when using chinese IME, caret jump to the start of selected text (I mena texts in a composition session ) when typing the first letter, but if we input the second letter, caret would jump to the right place --- the end of selected text.

The source of the bug is async code in onCompositionStartCapture, if we remove `setTimeout`, the bug would disappear.

**Recording**
https://recordit.co/gDZchKHAPa

When I press 'w', caret is on the left of 'w', like this '[caret]w',  **it should be on the right of 'w',  like this 'w[caret]'** .
Then  I type 'o', caret jump to the right place, like this 'wo[caret]'.

**Sandbox**
https://codesandbox.io/s/slate-caret-bug-s28fv6

**Steps**
To reproduce the behavior:
1. Go to https://codesandbox.io/s/slate-caret-bug-s28fv6
2. Switch to any chinese IME.
3. Focus the input area.
4. Type a word, like 'wo', and you will see the bug.

**Environment**
- Slate Version: "slate": "0.94.1", "slate-react": "0.95.0"
- Operating System: MacOS 13.4
- Browser: Chrome  113.0.5672.126, Safari 16.5

PR #5443, opened 2023-06-09
fix:  caret jump to the start of composition string

OldDream:
**Description**
Correct caret position during composition. 
When useIsomorphicLayoutEffect was triggered during composition, caret would jump to wrong position (jump to the start).
This commit fix the bug when using chinese IME.

More info about the bug is in the issue.

**Issue**
Fixes: [#5442](https://github.com/ianstormtaylor/slate/issues/5442)

### DIFF
--- a/.changeset/rare-glasses-tickle.md
+++ b/.changeset/rare-glasses-tickle.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+fix wrong caret position during composition.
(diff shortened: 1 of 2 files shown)


dylans (changes requested):
Looks reasonable, reminds me that we really need to make progress on a test suite.

@OldDream could you please add a changeset per the instructions and we'll land this in the next version?

dylans (approved):

PR #5446, opened 2023-06-10
include missed import in docs walkthrough

davi-alexandre:
**Description**
Imports Element from slate, as it is used in the given code sample of the walkthrough:
```ts
// docs/walkthroughs/03-defining-custom-elements.md
// line 156
  { match: n => Element.isElement(n) && Editor.isBlock(editor, n) }
```

Prevents the following TS error:
```
Property 'isElement' does not exist on type '{ new (): Element; prototype: Element; }'.ts(2339)
```


dylans (approved):
Thanks @davi-alexandre!

Issue #5406, opened 2023-04-27
In Chrome ending a backward selection in a Void and pressing delete make Slate crash


etienne-dldc:
**Description**
If you backward select some text and end the selection in a Void node, then press `delete` : Slate crashes with the following error:

```
Failed to execute 'setEnd' on 'Range': There is no child at offset 9.
```

**Recording**

https://user-images.githubusercontent.com/14336608/234879652-dc56f998-784c-492f-a654-5f1c78d1c66e.mov

**Sandbox**
https://codesandbox.io/s/slate-delete-void-crash-yymw29?file=/src/App.tsx

**Steps**
To reproduce the behavior:
1. Select some text under the void and end the selection inside the void node
2. Press delete

**Expectation**
Slate should not crash.

**Environment**
- Slate Version: 0.91.4
- Operating System: MacOS (was reproduced on Windows)
- Browser: Chrome (it works correctly in Firefox)

**Notes**
- Since the issue only occur in Firefox this might be related to #3148.
- Also the fact that the void node `useSelected` is `false` while the cursor is hovering the node feels like a bug to me.

PR #5451, opened 2023-06-13
Fixed triple click workaround breaking void elements selection

gtluszcz:
**Description**
#5343 Introduced a workaround in `toSlateRange` with a purpose to fix [chromium issue](https://bugs.chromium.org/p/chromium/issues/detail?id=608393)  with triple-clicking that sometimes moves cursor inside non content editable nodes `contentEditable="false"`. Unfortunately, the same workaround causes void elements to crash when being selected, since in the [official docs slates requires](https://docs.slatejs.org/api/nodes/element#rendering-void-elements) void nodes to also have `contentEditable="false"` attribute. Issue #5406 describes the problem in detail.

**Issue**
Fixes: #5406

**Proposed solution**
The original issue focused on fixing selection for `contentEditable="false"` nodes that **were not** void element root nodes at the same time. Also `toSlateRange` already can figure out the proper selection when domRange ends/starts in voids. I simply propose to check if a `focusedNode` is not a void element root node. In such case let Slate handle it the old way

**Example**
See  #5406

**Context**
I'd take a closer look at the workaround introduced in #5343. In my opinion, it causes more harm than profit. The algorithm for choosing `focusedNode` is very trivial. Just pick `anchorNode`. What if there are a couple of text nodes between anchorNode and a `contentEditable="false"` node? The selection collapses to a single node only even though the user selected multiple text nodes.

**Checks**

### DIFF
--- a/.changeset/popular-shrimps-give.md
+++ b/.changeset/popular-shrimps-give.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+Fixed occasional crashes when selecting void elements in Chrome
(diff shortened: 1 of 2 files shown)


dylans (approved):

DCzajkowski:
Good job @gtluszcz!

Issue #5450, opened 2023-06-13
Caret jump to wrong position after composition end.


OldDream:
**Description**
MacOS + chrome 113, caret  jump to the front of word after composition end. Which should stay after the word.
This bug didn't show up in MacOS +  Safari 16.5.

**Recording**
http://recordit.co/mNllqsP6BD

**Sandbox**
https://codesandbox.io/s/slate-caret-bug-forked-gvrzh4?file=/index.js

**Steps**
To reproduce the behavior:

Go to https://codesandbox.io/s/slate-caret-bug-forked-gvrzh4?file=/index.js
Switch to any chinese IME.
Focus the input area.
Type a word, like 'wokankan', select '我看看' from IME, then you will see the bug.
The bug looks like: [caret]我看看, and it should be: 我看看[caret]

**Environment**

Slate Version: "slate": "0.94.1", "slate-react": "0.96.0"
Operating System: MacOS 13.4
Browser: Chrome 113.0.5672.126

-----

I found the bug is related to this PR: https://github.com/ianstormtaylor/slate/pull/5437, remove  `!IS_WEBKIT &&` at line 1117 in `packages/slate-react/src/components/editable.tsx` can fix this. But I don't know if this would bring any side effect.

PR #5453, opened 2023-06-15
fix: regular expression for testing Webkit based browser

Shiba-ligo:
**Description**
Fix the regular expression for testing Webkit based browser. This bug was introduced in #5437 which changed `IS_SAFARI` environment constant to `IS_WEBKIT`, intending to apply all COMPATs for safari to Webkit based browsers like Chrome for ios. However, the regular expression for `IS_WEBKIT` is not correct, because it doesn't exclude blink based browser like Chrome for desktop. That makes some COMPATs for Chrome not work as expected.

**Issue**
Fixes: #5450 
Besides the problem mentioned in #5450, this bug breaks all user inputing behaviors using an IME which triggers a series of composition events in Chrome.

**Example**
**Video**
Chrome not working using an IME:
https://recordit.co/wYOq8Lj7RR
Safari works fine:
 https://recordit.co/auIGkmenlk
**Sandbox**
https://codesandbox.io/s/relaxed-poitras-ndspl8?file=/src/App.js

This bug breaks some COMPATs like
https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/editable.tsx#L1117-L1122

**Checks**

### DIFF
--- a/.changeset/khaki-toys-unite.md
+++ b/.changeset/khaki-toys-unite.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+fix regular expression for testing Webkit based browser.
--- a/packages/slate-react/src/utils/environment.ts
+++ b/packages/slate-react/src/utils/environment.ts
@@ -20,7 +20,7 @@ export const IS_FIREFOX =
   /^(?!.*Seamonkey)(?=.*Firefox).*/i.test(navigator.userAgent)
 
 export const IS_WEBKIT =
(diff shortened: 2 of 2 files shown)


dylans (approved):
Good catch, thanks!

dylans:
@Shiba-ligo could you please run prettier on the change and push and update and then I'll land and release. Thanks!

Shiba-ligo:
@dylans Thanks for replying. I've fixed it.

Issue #5457, opened 2023-06-19
wider space btw each line 


Noothan-am:
**Problem**
when you enter the new line, it gives more space than required
![image](https://github.com/ianstormtaylor/slate/assets/100135382/f3e0a950-eb38-4da4-8538-2561c0d9bcc3)

**Solution**
reduce the space btw lines figure 
![image](https://github.com/ianstormtaylor/slate/assets/100135382/98875ffe-9610-4fa6-90dc-37ff8e632c0d)

Noothan-am:
hey @ianstormtaylor, please assign it to me i want to work on this issue?

sws797:
use Shift+Enter to add a <br> in one <p>

Noothan-am:
i want to modify the css in this

Noothan-am:
hey @sws797, do you need that space when user presses enter. if not i will remove it from css? or if margin is required i will add Shift + enter --> <br/> feature?

sws797:
@Noothan-am It's not a matter for me, I will customize the style according to different products when necessary.

Noothan-am:
ok then, i will remove this part of the CSS which causes this issue
![image](https://github.com/ianstormtaylor/slate/assets/100135382/f02224fe-9d0d-41af-828b-1e4cda0c81af)

ApplyBoardDak:
this is standard behaviour in rich text editors, examples:
- [Summernote](https://summernote.org/)
- [TinyMCE](https://www.tiny.cloud/)
- [Froala](https://froala.com/wysiwyg-editor/demo/)

dylans:
Closing this because it does not look like a current Slate bug.

This is CSS or line-height confusion, not editor behavior.

If there is a minimal repro showing current Slate behavior is at fault, a narrower follow-up can be opened.

PR #5468, opened 2023-06-25
Fixed large Margin

Noothan-am:
**Description**
solves large space when we press Enter 

**Issue**
Fixes: #5457 

*ScreenShot*
image

**Checks**

### DIFF
--- a/site/public/index.css
+++ b/site/public/index.css
@@ -20,7 +20,7 @@ pre {
   white-space: pre-wrap;
 }
 
-:not(pre) > code {
+:not(pre)>code {
   font-family: monospace;
   background-color: #eee;
   padding: 3px;
@@ -75,8 +75,8 @@ iframe {
   border: 1px solid #eee;
 }
 
-[data-slate-editor] > * + * {
-  margin-top: 1em;
+[data-slate-editor]>*+* {
+  /* margin-top: 1em; */
 }
 
 .fancy {


dylans (changes requested):
I think a better approach for this and a few other things would be to make them configurable, or provide documentation to make it clear how to override them.

Not sure why this edit removed spaces in CSS and a commented out value isn't really a final solution.

Noothan-am:
> I think a better approach for this and a few other things would be to make them configurable, or provide documentation to make it clear how to override them.
> 
> Not sure why this edit removed spaces in CSS and a commented out value isn't really a final solution.

Hey @dylans, when we click enter, the margin used to get added, so I wanted to remove that space by commenting out the CSS responsible for this issue `margin-top: 1em`. The spaces in the CSS code are because of my code formatter that is not the actual change I made. I have only changed the margin-top by commenting it out. Should I remove this comment and add the feature discussed earlier?

dylans:
ok, I can land this once a changeset gets added and the linting issue are resolved.

site/public/index.css
```diff
@@ -20,7 +20,7 @@ pre {
   white-space: pre-wrap;
 }
 
-:not(pre) > code {
+:not(pre)>code {
```
ApplyBoardDak, line 23: why this change?

site/public/index.css
```diff
@@ -75,8 +75,8 @@ iframe {
   border: 1px solid #eee;
 }
 
-[data-slate-editor] > * + * {
-  margin-top: 1em;
+[data-slate-editor]>*+* {
```
ApplyBoardDak, line 78: same question here

dylans:
@Noothan-am do you have plans to address the feedback or should I close this PR?

dylans:

Issue #5513, opened 2023-09-20
The “move_node” operation triggers a node re-render


YaoKaiLun:
**Description**
The “move_node” operation triggers a node re-render. I see the PR https://github.com/ianstormtaylor/slate/pull/4120 had fixed move_node causing full re-render, but i think it is not enough.

**Recording**
A blockquote element includes paragraph elements

1. Pressing delete key in the start of the first paragraph, it would trigger the blockquote element re-render（the image is reloaded）
2. Pressing enter key in the last empty paragraph,  it also would trigger the blockquote element re-render

https://github.com/ianstormtaylor/slate/assets/11460856/9129d414-a04c-44a7-a82a-e5fc2a0753c1

**Sandbox**
Using the official example https://www.slatejs.org/examples/richtext can produce this problem. After deleting the paragraph behide the list element, the list element's key is changed

![企业微信截图_941c1881-d637-451e-810d-eaaf604f375d](https://github.com/ianstormtaylor/slate/assets/11460856/79d3ab49-b17a-40a3-90e1-1d06889fdc85)
![企业微信截图_612f7391-a8df-43f9-abc1-9c04d5967b07](https://github.com/ianstormtaylor/slate/assets/11460856/0382de40-cd4a-4a87-ae3a-9222d454e321)

the action will produce operations:
![image](https://github.com/ianstormtaylor/slate/assets/11460856/1314e0f2-94c3-4465-9ae4-8a019ea6a61b)

**Expectation**
the source parent element and the target parent element of the moved node, should not re-render

**Environment**
- Slate Version: [0.94.1]
- Operating System: [e.g. MacOS]
- Browser: [Chrome 116]

PR #5514, opened 2023-09-20
Fix move_node triggers nodes re-render (#5513)

YaoKaiLun:
**Description**
Reset the source parent element and the target parent element of the moved node to NODE_TO_KEY, for prevent them to re-render

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/5513

**Example**
Using the official example https://www.slatejs.org/examples/richtext . After deleting the paragraph behide the list element, the list element's key will not change anymore.
企业微信截图_617ae87a-93ce-49af-b10a-d419d8705827
企业微信截图_a0a55074-9833-4176-86ac-c3ab4597ddff

**Checks**

### DIFF
--- a/packages/slate-react/src/plugin/with-react.ts
+++ b/packages/slate-react/src/plugin/with-react.ts
@@ -6,6 +6,7 @@ import {
   Node,
   Operation,
   Path,
+  PathRef,
   Point,
   Range,
   Transforms,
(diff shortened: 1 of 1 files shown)


dylans (changes requested):
Needs a changeset please.

A few integration tests appear to be failing, but it's unclear if this is due to this change or not.

YaoKaiLun:
> Needs a changeset please.
> 
> A few integration tests appear to be failing, but it's unclear if this is due to this change or not.

The latest PR had same errors https://github.com/ianstormtaylor/slate/actions/runs/6186489569/job/16794248174

Issue #5169, opened 2022-10-26
ReactEditor.focus places the caret at the wrong position


mateuspiresl:
**Description**
`ReactEditor.focus` places the caret at the wrong position. Seems that Slate knows the correct position of the caret, at least internally, because a call to `editor.insertText` inserts the text at the correct position.

**Recording**

https://user-images.githubusercontent.com/6335577/198074164-7b025c9b-1831-49c5-a338-60a4ca73b0da.mov

**Sandbox**
Slate v0.84.0: https://codesandbox.io/s/slate-focus-bug-v0-84-0-z721fj
Slate v0.58.0: https://codesandbox.io/s/slate-focus-bug-v0-58-0-xtlz36

**Steps**
To reproduce the behavior:
1. Click at the end of the line to gain focus there
2. Click outside the editor to loose focus
3. Call `ReactEditor.focus` (by clicking on the button "Focus" in the sandbox)
4. See the caret at the beginning of the line

**Expectation**
The caret should have been placed at the end of the line.

**Environment**
- Slate Version: 0.84.0 (and also 0.58.0)
- Operating System: macOS Big Sur 11.6
- Browser: Chrome Version 106.0.5249.103 (Official Build) (arm64)

**Context**
It seems to be doable because calling focus on Draft.js would place the caret at the correct position. I'm currently switching from Draft.js to Slate.

mateuspiresl:
Is this focus behavior above expected on the `ReactEditor.focus` method? If so, I can create a PR for that. I still have to work on Firefox to find what's wrong with `editor.insertText`.

mateuspiresl:
The issue I mentioned about the `editor.insertText` is unrelated to this issue. The following sandbox fix both but I will create another issue for the unrelated one.

Sandbox: https://codesandbox.io/s/slate-focus-simple-alternative-9su7gr?file=/index.js.

Seems that Firefox doesn't clear its selection on blur, so changing the content of the contenteditable when it isn't focused (calling `editor.insertText`) makes the DOM fire a selection update. The problem is that this selection update is placed at the beginning of the document.

To fix that, I created a hook to listen to focus changes on the editor and clear the DOM selection on blur. It can be found at `./useBlurResolver` in the sandbox.

The solution for `ReactEditor.focus` is still the same and can be found at `./focusEditor.js` in the sandbox.

VictorBaron:
IMO it should indeed move the caret to the current selection.
Or at least have an option to chose that behavior.

[...]

PR #5516, opened 2023-09-21
Retain editor selection when using ReactEditor.focus()

josephmr:
**Description**
When calling `ReactEditor.focus()` the current editor selection is lost and the caret is placed at the beginning of the document. This change retains the current editor selection when giving the editor focus.

**Issue**
Fixes: [(link to issue)](https://github.com/ianstormtaylor/slate/issues/5169)

**Example**
Before:

https://github.com/ianstormtaylor/slate/assets/28637598/b68fc68d-90b5-4428-ae2d-e403988e74be

After:

https://github.com/ianstormtaylor/slate/assets/28637598/d55170ed-d200-4cf2-a6e1-77e534638091

**Context**
Before focusing the editor we set the DOM selection so that the corresponding `onDOMSelectionChange` handler in `editable.tsx` does not reset the editor selection when it fires after the `el.focus()` call.

**Checks**

### DIFF
--- a/.changeset/selfish-frogs-press.md
+++ b/.changeset/selfish-frogs-press.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+Retain editor selection when using ReactEditor.focus()
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -417,7 +417,7 @@ export const ReactEditor: ReactEditorInterface = {
     IS_FOCUSED.set(editor, true)
 
     if (root.activeElement !== el) {
-      if (root instanceof Document) {
+      if (editor.selection && root instanceof Document) {
         const domSelection = root.getSelection()
         const domRange = ReactEditor.toDOMRange(editor, editor.selection)
         domSelection?.removeAllRanges()
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -417,6 +417,12 @@ export const ReactEditor: ReactEditorInterface = {
     IS_FOCUSED.set(editor, true)
 
     if (root.activeElement !== el) {
+      if (root instanceof Document) {
+        const domSelection = root.getSelection()
+        const domRange = ReactEditor.toDOMRange(editor, editor.selection)
+        domSelection?.removeAllRanges()
+        domSelection?.addRange(domRange)
+      }
       el.focus({ preventScroll: true })
     }
   },


dylans (approved):
I have mild concerns this could break something else, but let's give it a try. Thanks!

zbeyens:
From this change, any `ReactEditor.focus(editor)` call after a transform will trigger that error https://github.com/udecode/plate/issues/2674

The workaround is to use `setTimeout`.

We need to either fix it so it can resolve the DOM node, or revert this @josephmr

12joan:
Is it possible we could add an option to `ReactEditor.focus` that controls this behaviour?

skogsmaskin:
What about something like this in the beginning of the static [focus method](https://github.com/josephmr/slate/blob/bcc40859c4e2980229321abe08233404570c017e/packages/slate-react/src/plugin/react-editor.ts#L414C1-L428C5):


```
focus: editor => {
  // If the editor has pending operations, focus should be requested
  // after those changes are applied. Retry in the next tick.
  if (editor.operations.length > 0) {
    setTimeout(() => {
      ReactEditor.focus(editor)
    })
    return
  }
 // ...the rest of the focus code
}
```

zbeyens:
@dylans @josephmr Many consumers having issues from that release https://github.com/udecode/plate/issues/2674. We should release @skogsmaskin's suggestion as a quick fix.

skogsmaskin:
I drafted a PR with my suggestion here - https://github.com/ianstormtaylor/slate/pull/5527

I haven't been able to confirm if this will actually fix your particular issue @zbeyens would you be able to confirm?

I'm not even sure if this is the right solution, but to me, it sounds like a good idea to not interfere with document selections, while the editor might potentially try to transform itself and potentially create a new one.

skogsmaskin:
There seems to be a similar issue with the `Editable` `onBlur` [event handler.] that only manifested itself with React `18.0.0` and Chrome (https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/editable.tsx#L984)

There it's also trying to map a DOM node with this function: `ReactEditor.hasSelectableTarget(editor, event.target)` which sometimes fails very similar to the bug we are discussing.

aodinok:
This change is also breaking our tests, slate is unable to resolve node on focus call.
Having ability to control that behaviour would be nice 🙏 

```
annot resolve a DOM point from Slate point: {"path":[0,0],"offset":20}
    at Object.toDOMPoint (react-editor.ts:598:13)
    at Object.toDOMRange (react-editor.ts:611:35)
    at Object.focus
```

sneridagh:
I am having the same issue since this PR was merged. When I try to set a link with a selection active.

```
index.es.js:644 Uncaught Error: Cannot resolve a DOM node from Slate node: {"text":"about"}
    at Object.toDOMNode (index.es.js:644:1)
    at Object.toDOMPoint (index.es.js:651:1)
    at Object.toDOMRange (index.es.js:723:1)
    at Object.focus (index.es.js:565:1)
    at Object.onChangeValue (index.js:82:1)
    at AddLinkForm.onSubmit (AddLinkForm.jsx:210:1)
```

`AddLinkForm` is my link plugin.

skogsmaskin:
I opened up my PR today, which should fix the `Cannot resolve a DOM node from Slate node` error.

Issue #5521, opened 2023-10-13
Derive `onSelectionChange` and `onValueChange` from `onChange`


zbeyens:
**Problem**
`onChange` currently triggers for all changes: value and selection. When I only want to listen to value changes, this is not optimal.

**Solution**
Keep `onChange` and add `onSelectionChange` to listen to selection change only and `onValueChange` to listen to value change only.

Bounty: https://github.com/udecode/plate/issues/2700

rishi-raj-jain:
@zbeyens 

`onValueChange` from useStateSelector where type of op is `set_selection` would the right logic, right?

rishi-raj-jain:
@zbeyens @12joan 

`onValueChange` from useStateSelector where type of op is `set_selection` would the right logic, right?

zbeyens:
The goal is to listen to value change only, all operations except `set_selection`. We want to avoid using an `equalityFn` like in `useSlateSelector`.

rishi-raj-jain:
Ah so totally a different hook.

zbeyens:
1. The ﻿`onSelectionChange` function should only listen to the ﻿`set_selection` operation, not all operations.
2. Conversely, the ﻿`onValueChange` function should only listen to the other operations, excluding ﻿`set_selection`.

The solution needs to be built exactly like `onChange` prop. No hook is expected, only `Slate` props.

PR #5523, opened 2023-10-17
feat/#5521/derive on selection change and on value change

ologbonowiwi:
- feat: trigger onSelectorChange and onValueChange on onChange
- docs: add comment
- docs: add changeset

**Description**
Keep onChange and add onSelectionChange to listen to selection change only and onValueChange to listen to value change only.

**Issue**
Fixes: #5521
/claim udecode/plate#2700

**Checks**

cc: @zbeyens

### DIFF
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -23,6 +23,7 @@ export const Slate = (props: {
   children: React.ReactNode
   onChange?: (value: Descendant[]) => void
 }) => {
+  // do we need to add onValueChange and onSelectorChange here?
   const { editor, children, onChange, initialValue, ...rest } = props
 
   const [context, setContext] = React.useState<SlateContextValue>(() => {
--- a/packages/slate-react/src/plugin/with-react.ts
+++ b/packages/slate-react/src/plugin/with-react.ts
@@ -53,7 +53,15 @@ export const withReact = <T extends BaseEditor>(
   clipboardFormatKey = 'x-slate-fragment'
 ): T & ReactEditor => {
   const e = editor as T & ReactEditor
-  const { apply, onChange, deleteBackward, addMark, removeMark } = e
+  const {
+    apply,
+    onChange,
+    onValueChange,
+    onSelectionChange,
+    deleteBackward,
+    addMark,
+    removeMark,
+  } = e
 
   // The WeakMap which maps a key to a specific HTMLElement must be scoped to the editor instance to
   // avoid collisions between editors in the DOM that share the same value.
@@ -364,6 +372,14 @@ export const withReact = <T extends BaseEditor>(
 
       if (onContextChange) {
         onContextChange()
+      }
+
+      switch (options?.operation?.type) {
+        case 'set_selection':
+          onSelectionChange(options)
+          break
+        default:
+          onValueChange(options)
       }
 
       onChange(options)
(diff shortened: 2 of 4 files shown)


ologbonowiwi:
@zbeyens let me know if it fills the expectations or there's anything that needs to be changed :smile:

packages/slate-react/src/plugin/with-react.ts
```diff
@@ -366,6 +374,14 @@ export const withReact = <T extends BaseEditor>(
         onContextChange()
```
zbeyens, line 374: You need to use this as stated in https://github.com/ianstormtaylor/slate/issues/5521#issuecomment-1762853454

zbeyens:
Please write unit tests for:
- `Slate` prop: `onValueChange`
- `Slate` prop: `onSelectionChange`
- editor method: `editor.onValueChange`
- editor method: `editor.onSelectionChange`

packages/slate-react/src/plugin/with-react.ts
ologbonowiwi: I'm not getting what you mean. What should I pass as parameter here?

zbeyens: We need 2 new props to `Slate`: `onSelectionChange` and `onValueChange`. These callbacks should be called in `onContextChange` like `onChange`.

packages/slate-react/src/components/slate.tsx
```diff
@@ -23,6 +23,7 @@ export const Slate = (props: {
   children: React.ReactNode
   onChange?: (value: Descendant[]) => void
 }) => {
+  // do we need to add onValueChange and onSelectorChange here?
```
ologbonowiwi, line 26: here is where I need to add the new handlers, @zbeyens?

and execute them inside [onContextChange](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/slate.tsx?plain=1#L51C1-L61)?

or do you want me to add a new parameter on [onContextChange](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/slate.tsx?plain=1#L51C1-L61), where we'll send the `onValueChange`/`onSelectorChange` (from [this line](https://github.com/ianstormtaylor/slate/pull/5523/files?plain=1#diff-7fe1360dfc4f25adbf2cb567cd2a6b3e655b3a62f07872dc072d64b5a80bcd82R374))

ologbonowiwi, line 26: here is where I need to add the new handlers, @zbeyens?

and execute them inside [onContextChange](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/slate.tsx#L51C1-L61)?

or do you want me to add a new parameter on [onContextChange](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/slate.tsx#L51C1-L61), where we'll send the `onValueChange`/`onSelectorChange` (from [this line](https://github.com/ianstormtaylor/slate/pull/5523/files#diff-7fe1360dfc4f25adbf2cb567cd2a6b3e655b3a62f07872dc072d64b5a80bcd82R374))

zbeyens: Whatever that makes it work :) Will be confirmed by unit tests.

zbeyens:
Another contributor has submitted a complete and satisfactory PR #5526. Therefore, this one can be closed. Thank you for your efforts.

PR #5526, opened 2023-10-19
Add `onSelectionChange` and `onValueChange` in Slate React component

jkcs:
**Description**
Add `onSelectionChange` and `onValueChange` in Slate React component

**Issue**
Fixes:  #5521 
/claim udecode/plate#2700

**Checks**

### DIFF
--- a/packages/slate-react/src/components/slate.tsx
+++ b/packages/slate-react/src/components/slate.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useEffect, useState } from 'react'
-import { Descendant, Editor, Node, Scrubber } from 'slate'
+import { Descendant, Editor, Node, Operation, Scrubber } from 'slate'
 import { FocusedContext } from '../hooks/use-focused'
 import { useIsomorphicLayoutEffect } from '../hooks/use-isomorphic-layout-effect'
 import { SlateContext, SlateContextValue } from '../hooks/use-slate'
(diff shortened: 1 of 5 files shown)


packages/slate-react/src/components/slate.tsx
```diff
@@ -22,8 +22,18 @@ export const Slate = (props: {
   initialValue: Descendant[]
   children: React.ReactNode
   onChange?: (value: Descendant[]) => void
+  onSelectorChange?: (value: Descendant[]) => void
```
zbeyens, line 25:
```suggestion
  onSelectionChange?: (selection: Editor['selection']) => void
```

packages/slate-react/test/index.spec.tsx
```diff
@@ -95,4 +95,49 @@ describe('slate-react', () => {
...
+          <Editable />
+        </Slate>,
+        { createNodeMock }
+      )
+    })
+    await act(async () =>
+      Transforms.select(editor, { path: [0, 0], offset: 2 })
+    )
+    expect(onSelectorChange).toHaveBeenCalled()
+  })
```
zbeyens, line 122: Could you add 
```ts
expect(onChange).toHaveBeenCalled()
expect(onValueChange).not.toHaveBeenCalled()
```

packages/slate-react/test/index.spec.tsx
```diff
@@ -95,4 +95,49 @@ describe('slate-react', () => {
...
+          onValueChange={onValueChange}
+        >
+          <Editable />
+        </Slate>,
+        { createNodeMock }
+      )
+    })
+    await act(async () => Transforms.insertText(editor, 'Hello word!'))
+    expect(onValueChange).toHaveBeenCalled()
+  })
```
zbeyens, line 142: Could you add 
```ts
expect(onChange).toHaveBeenCalled()
expect(onSelectionChange).toHaveBeenCalled()
```

And another test using `Transforms.setNodes` that should not change selection?

zbeyens (changes requested):
Great PR, I've added a few suggestions then it should be complete

jkcs:
> Great PR, I've added a few suggestions then it should be complete

Thank you for your suggestion. I will make the changes as soon as possible

packages/slate-react/test/index.spec.tsx
jkcs: I added a unit test case for setting bold using Transforms.setNodes.

zbeyens (approved):
LGTM, thanks!

PR #5527, opened 2023-10-19
Fix issue with ReactEditor.focus + tests

skogsmaskin:
**Description**
This will make sure we don't try to create a DOM range while the editor is still applying operations in the `ReactEditor.focus` function. And that it retries doing this until this the editor has flushed all it's changes.

If a client requests `ReactEditor.focus` from the UI, while the editor is updating the DOM, this can result in unrecoverable errors like: `Error: Cannot resolve a DOM node from Slate node: {"text":"hello"}`

This has been reported by several people since https://github.com/ianstormtaylor/slate/pull/5516 was merged (see discussion there).

[Have a look a these changes](https://github.com/ianstormtaylor/slate/pull/5527/files#diff-06bf833c7913658864514eb335e23b8ddb7a55888d004bd68d8d1293933103cbR414)

I have made tests for this function, and in order to properly test it I replaced `react-test-renderer` with `@testing-library/react` which has a very similar API, but fully supports `window`. This means we can test code that constructs DOM ranges against the actual `window.getSelection` API.

[This test](https://github.com/ianstormtaylor/slate/blob/649a36dddd761bb7d6e105863b579aa4a2f79af6/packages/slate-react/test/react-editor.spec.tsx#L44) will fail without my refactored code and provoke the above-mentioned error.

**Issue**
Fixes: see feedback on https://github.com/ianstormtaylor/slate/pull/5516

**Example**
See the new `react-editor.spec.tsx` tests

**Context**
See https://github.com/ianstormtaylor/slate/pull/5516

**Checks**


packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -412,6 +413,25 @@ export const ReactEditor: ReactEditorInterface = {
...
+      const handle = FOCUS_TIMER_HANDLE.get(editor)
+      if (handle) {
+        clearTimeout(handle)
+        FOCUS_TIMER_HANDLE.delete(editor)
+      }
+      FOCUS_TIMER_HANDLE.set(
+        editor,
+        setTimeout(() => {
+          // Flush changes before trying to focus again
+          editor.onChange()
```
skogsmaskin, line 428: Not sure if we should actually call onChange here, or just retry until there are no operations pending.

12joan: onChange should already have been called by this point, right? I can't think of any situations where manually flushing the changes would be beneficial, and it might result in some weird edge cases. If, somehow, there are still pending operations after the timeout, retrying is probably safest.

12joan: Do we still need a weak map if we're not calling onChange? I can't think of any adverse side effects of running the timeout callback twice if a consumer happens to call ReactEditor.focus twice.

12joan: The only other thing that occurs to me is what happens if something goes wrong and editor.operations never gets emptied? That might happen if we call ReactEditor.focus after the editor has been unmounted, for example.

Perhaps we should add an `options: { retries?: number }` argument that defaults to 5 and is decremented on each attempt?

skogsmaskin: @12joan - thank you for your comments. I've been looking a bit closer at this now and I'm unsure if this is actually the right fix, though it seems to help in some cases.

I think we might be able to just drop the whole selection concern in this function, as the selection change event handler should already have taken care of this. When we call `focus()` on the element, it should trigger a selection change event that will restore the proper selection.

I'll check in the new code which has this code removed, and that would definitely be the best if we don't actually need it here.

skogsmaskin: > onChange should already have been called by this point, right? I can't think of any situations where manually flushing the changes would be beneficial, and it might result in some weird edge cases. If, somehow, there are still pending operations after the timeout, retrying is probably safest.

Yes, I have refactored this now. It will now call `onChange`, but only because it will try to set a selection in the top of the document if the editor currently isn't selected.

> Do we still need a weak map if we're not calling onChange? I can't think of any adverse side effects of running the timeout callback twice if a consumer happens to call ReactEditor.focus twice.

I agree. I have removed it - also one less WeakMap :)

> The only other thing that occurs to me is what happens if something goes wrong and editor.operations never gets emptied? That might happen if we call ReactEditor.focus after the editor has been unmounted, for example.
> Perhaps we should add an options: { retries?: number } argument that defaults to 5 and is decremented on each attempt?

Yeah, good thinking. I have implemented a retry (default 5) and then throw an error if the editor is still stuck with operations.

skogsmaskin:
I have updated this PR quite a bit today. Please have another look.

skogsmaskin:
I can make a separate PR on the test upgrades if you'd prefer. If not, this PR should be rebased on top of `main` to keep those commits separate.

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -411,19 +411,44 @@ export const ReactEditor: ReactEditorInterface = {
...
         domSelection?.removeAllRanges()
         domSelection?.addRange(domRange)
       }
+      // Create a new selection in the top of the document if missing
+      if (!editor.selection) {
+        Transforms.select(editor, Editor.start(editor, []))
+        editor.onChange()
+      }
       el.focus({ preventScroll: true })
+      IS_FOCUSED.set(editor, true)
```
skogsmaskin, line 451: Moved this to the end to ensure that we were able to set everything properly first.

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -411,19 +411,44 @@ export const ReactEditor: ReactEditorInterface = {
...
         const domSelection = root.getSelection()
         const domRange = ReactEditor.toDOMRange(editor, editor.selection)
         domSelection?.removeAllRanges()
         domSelection?.addRange(domRange)
       }
+      // Create a new selection in the top of the document if missing
+      if (!editor.selection) {
+        Transforms.select(editor, Editor.start(editor, []))
+        editor.onChange()
+      }
```
skogsmaskin, line 449: This seems like it's missing. Added it as the tests showed that the editor would be stuck with a `null` selection after focus was set.

packages/slate-react/test/react-editor.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <Slate editor={editor} initialValue={initialValue}>
+              <Editable />
+            </Slate>
+          )
+        })
+
+        await act(async () => {
+          Transforms.removeNodes(editor, { at: [0] })
+          Transforms.insertNodes(editor, propagatedValue)
+          ReactEditor.focus(editor) // Note: calling focus in the middle of these transformations.
```
skogsmaskin, line 68: This focus call here will provoke that `Error: Cannot resolve a DOM node from Slate node: {"text":"hello"}`

dylans (approved):

PR #5549, opened 2023-11-01
Fix crash when deleting paragraph ending `\n` in Firefox

12joan:
**Description**
Due to a [Firefox browser quirk](https://github.com/ianstormtaylor/slate/issues/5273#issuecomment-1789494609), the DOM selection sometimes includes the additional `\n` added when `isTrailing` is true in `TextString`. This results in an invalid focus point, causing a crash during delete operations.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/5273

**Example**
Before:

https://github.com/ianstormtaylor/slate/assets/4272090/5195ec18-a630-48ef-8e8d-579ba471b832

After:

https://github.com/ianstormtaylor/slate/assets/4272090/32c15c50-52af-4c1a-92e5-43a50d6dfe10

**Context**
It's possible that the logic for the check I added might not be perfect. However, I'm 99% sure that any new bugs it introduces will be harder to reproduce accidentally than the bug it fixes.

**Checks**

### DIFF
--- a/.changeset/chatty-chefs-sneeze.md
+++ b/.changeset/chatty-chefs-sneeze.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Firefox compat: Fix incorrect focus.offset when text node ends with \n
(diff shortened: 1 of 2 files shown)


dylans (changes requested):
GitHub actions is providing a linting warning that should be addressed. Please review that and then we can land this.

Issue #3309, opened 2019-12-12
decorated text cannot be selected (#3118 re-open)


jege31:
Issue #3118 was unfortunately not corrected by #3093, so I open this new issue, as suggested.

I wrote a new [index.jsx](https://github.com/ianstormtaylor/slate/files/3957896/index.txt) (renamed index.txt because github does not accept .jsx) compliant with the new architecture of Slate.
The behaviour remains the same : after rendering a decoration (that looks ok in the browser) the text cannot be selected anymore. The cursor automatically moves to the end of the text. (see screen record #3118)
Moreover, when the cursor reaches the end of the text, the script raises an error : "Error: Cannot resolve a DOM point from Slate point: {"path":[0,0],"offset":33}"

The browser is Firefox and Slate is 0.53.0

stevez86:
I was just about to post an issue related to this, not sure if its related to selecting decorated text. I get the following errors very frequently:

`Cannot resolve a Slate point from DOM point: [object HTMLLIElement],0`
`Cannot resolve a DOM point from Slate point: {"path":[0,1,0],"offset":106}`

It **sometimes** happens after a user clicks in the editor. I know this from my error reporting which reports clicks. 

I can't reproduce yet, (only noticing in analytics) but it seems to happen more frequently (but not entirely) with mobile Android devices, (including Chrome Mobile 77.0.3865, Amazon Silk 73.7.5, Mobile Safari 13.0.3, Chrome Mobile 78.0.3904). 

I've had this occur on 0.50, 0.52 and 0.54. (I keep updating hoping this issue will resolve)

I'll keep trying to reproduce and update as I find more information.

jege31:
I think there are 2 issues.
1 issue concerning the selection of decorated text. This issue was already existing before the slate refactoring.
1 issue concerning the raised error "Cannot resolve a DOM point from Slate point". This one is new since the slate refactoring.

Since the Slate refactoring, the 1st issue when selecting of decorated text, also leads to raise the 2nd issue.

jege31:
I've inserted a GIF.
There was already a sandbox : [index.jsx](https://github.com/ianstormtaylor/slate/files/3957896/index.txt)

mattapperson:
Also having this issue

jege31:
I found a work-around, setting "contentEditable={false}" for the decoration elements.

```
const Leaf = ({ attributes, children, leaf}) => {
  const style = { backgroundColor: "lightcoral" };

[...]

PR #5571, opened 2023-11-30
add `suppressThrow` to `ReactEditor`s `toDOMRange` and `toDOMPoint`

markacola:
**Description**
When the editor content is set externally, the DOM may not have the focus points the editor attempts to resolve. This is already handled for when [the current selection is null](https://github.com/ianstormtaylor/slate/blob/03125312fbbd36da629978d9da92c244ffc836af/packages/slate-react/src/components/editable.tsx#L389).

This is an implementation of the first suggestion noted [here](https://github.com/ianstormtaylor/slate/issues/3309#issuecomment-1832964997)

**Issue**
Fixes: Various issues that result from the DOM content changing under the editor and I believe [3309](https://github.com/ianstormtaylor/slate/issues/3309)

**Example**
Here is an example of the external value being reset to "Text" programmatically, with the cursor at the end of the input. Previously, this would have errored.
[Screencast from 30-11-23 14:29:54.webm](https://github.com/ianstormtaylor/slate/assets/11350169/17df3319-7b37-4b9e-8cf0-820da1047446)

**Context**
I have replicated the `suppressThrow` pattern on other methods and enabled it when reprocessing focus via the useLayoutEffect in the Editable component.

**Checks**

### DIFF
--- a/.changeset/twelve-mails-warn.md
+++ b/.changeset/twelve-mails-warn.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Suppressed error throwing on selection processing for externally updated DOM content
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -384,7 +384,13 @@ export const Editable = (props: EditableProps) => {
       state.isUpdatingSelection = true
 
       const newDomRange: DOMRange | null =
-        selection && ReactEditor.toDOMRange(editor, selection)
+        selection &&
+        ReactEditor.toDOMRange(editor, selection, {
+          // Even when we have a selection in Slate, the DOM might have moved
+          // underneath us, so we have to check that the anchor and focus nodes
+          // still exist.
+          suppressThrow: true,
+        })
 
       if (newDomRange) {
         if (ReactEditor.isComposing(editor) && !IS_ANDROID) {
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -217,7 +217,11 @@ export interface ReactEditorInterface {
   /**
    * Find a native DOM selection point from a Slate point.
    */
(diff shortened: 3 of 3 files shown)


markacola:
Here you go: https://replit.com/@MarkCola/Slate-js-5571#src/App.tsx

Run this, type some extra content beyond the existing and you will see this error when the user focus is outside the updated/reset content:
image

dylans:
@markacola still thinking through this one. I'm a bit conflicted on this. I understand what the documentation says and your use case. I know that in our app we manage to avoid this scenario by instead mostly working with operations rather than the approach you are following (which isn't a bad thing, just why I've not really had this issue).

markacola:
@dylans Thank you for considering the changes - I implemented the PR under the assumption that the desired behaviour is to avoid throwing an error.

The only functional difference with these changes would be the focus blurring instead of throwing an error; I haven't considered the implementation details, but perhaps an alternative to this behaviour is attempting to recover an approximate selection.

Would that be preferable? Or would it be better to:
1. update the docs outlining the risks of implementing the naive `editor.children = externalValue` approach, and
2. document a way to consume external values via operations.

azvoncov-smartling:
The same problem occurs when changing the editor structure synchronously (like wrapNodes), and setSelection is set asynchronously - sometimes the same problem occurs

Only decision:
```js
const isPointExist = (editor, point) => {...};
// ...
editor.apply = (operation) => {
         if (operation.type === "set_selection" && operation.newProperties) {
             if (operation.newProperties.anchor && !isPointExist(editor, operation.newProperties.anchor)) return;
             if (operation.newProperties.focus && !isPointExist(editor, operation.newProperties.focus)) return;
         }

         apply(operation);
     };
};

bryanph:
instead of suppressing the error everywhere I think it would make more sense to introduce a specific `editor.reset` method that also resets the selection appropriately. Alternatively you can just reset the selection to `[0,0]` before setting `editor.children`.

dylans:

Issue #5574, opened 2023-11-30
Attempting to remove a mark after deselecting and reselecting the marked text does not work.


bradcypert:
**Description**
Selecting text, adding a mark, deselecting the text, then reselecting the text and trying to remove the mark does not seem to work. This is the case in both Slate 0.101.1 for my local development and I was able to recreate the issue with Slate's Rich Editor demo application.

**Recording**
![slate-bug(2)](https://github.com/ianstormtaylor/slate/assets/1455979/1e0f5c48-2007-4cfb-bb8f-86029087026d)

**Sandbox**
You are able to reproduce this issue here as well: https://www.slatejs.org/examples/richtext

**Steps**
To reproduce the behavior:
1. Open the link to the rich text example: https://www.slatejs.org/examples/richtext
2. double click to select a word with no marks (such as the 5th word "text").
3. Click on the "Bold" button to add a bold mark to the text.
4. Click on "Bold" button again to remove the mark
5. Observe that this behavior works as expected.
6. Click on "Bold" button again to add the mark once more.
7.  Deselect the currently selected text
8. Double click to previously selected text to select it once more.
9. Click on "Bold" button again and observe that the mark is not removed this time.

**Expectation**
I'd expect for the mark to be removed since it's already applied to the selection.

**Environment**
- Slate Rich Text Editor Demo Application (link above)
- Additionally, Slate 0.101.1

**Context**
I'm still learning how Slate works but if someone can point me in the right direction regarding the internals here, I'd be happy to take a look at fixing this.

Issue #5560, opened 2023-11-16
Regression: Firefox selection by double click


k0ka:
**Description**
Selecting a block in Firefox via double click doesn't allow to remove formatting.

**Recording**

https://github.com/ianstormtaylor/slate/assets/203597/a217e537-8060-4b5f-89f6-8382f4519217

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/richtext
2. Double click on a word to select it.
3. Click on the "bold" button.
4. Deselect.
5. Double click on the same word again to select it.
6. Click on the "bold" button.

**Expectation**
The bold button must be active and the font must get to normal after clicking on the "bold" button again.

**Environment**
- Slate Version: 0.98.0, 0.98.4, 0.101.0
- Operating System: MacOS, Windows 10
- Browser: Firefox

**Context**
This bug was reintroduced in slate-react 0.98.0. It was working properly in slate-react 0.97.2

I think it is the same bug as https://github.com/ianstormtaylor/slate/issues/3336 (or https://github.com/ianstormtaylor/slate/issues/4793) which was fixed in https://github.com/ianstormtaylor/slate/pull/5275 and reintroduced in https://github.com/ianstormtaylor/slate/pull/5486

Kaporos:
+1 with slate version 0.101.1 and firefox 120.0

12joan:
This can be closed as completed: #5580

PR #5486, opened 2023-07-25
Fix firefox disconnected selection api usage

WcaleNieWolny:
**Description**
This PR fixes the invalid usage of the selection API in firefox. Firefox allows disconnected ranges that cause problems with slate. I originally started to work on this to fix . [this issue](https://github.com/udecode/plate/issues/560)

**Issue**
Fixes: (I do not think this fixes any slate issue, however I am not sure. I did not checked)

**Example**
Before my changes:

https://github.com/ianstormtaylor/slate/assets/50914789/54a02c0a-4211-4d87-b5bc-197433eafd62

After my changes:

https://github.com/ianstormtaylor/slate/assets/50914789/84c69ec1-3088-480f-a75e-7e73ab93bab7

**Context**
1:\
`anchorNode`, `anchorOffset`, `focusOffset`. Do not behave the same in chrome and in firefox. In firefox if there is more then 1 range said values would be invalid. I fixed this by using the `getRangeAt` function (see `toSlateRange` changes).
2:\
I reversed [this pr](https://github.com/ianstormtaylor/slate/pull/5275/files) as it was interfering with my  changes (I believe this PR relies on broke selection api in firefox)
3:\
I removed the usage of `setDomSelection`. This is because `setBaseAndExtent` removed all ranges leaving only 1 range (this removed the ability to select more cells)
4:\
I removed the `el.focus()` because it relied on the `setDomSelection`
5:\
The slight refactoring was at some point done by eslint

**Checks**
- [x] The new code matches the existing patterns and styles.
- [x] The tests pass with `yarn test`.
- [x] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [x] The relevant examples still work. (Run examples with `yarn start`.)
- [ ] You've [added a changeset](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md) if changing functionality. (Add one with `yarn changeset add`.) (I will soon, this will be the next commit)

PR #5580, opened 2023-12-09
marks on double click selection using firefox fix

Kaporos:
**Description**

This PR contains a quick-fix of marks function using Firefox. This was already fixed, but the fix was removed inside of #5486. I put it back but this time only in marks code instead of selection handling one,  so it should not affect anything else. 

**Issue**
Fixes: #5574 #5560 

**Example**
Before the fix:

Recording 2023-12-09 at 15 47 57

When fixed:
Recording 2023-12-09 at 15 46 07

**Context**

This is related to firefox weird selection range when double clicking.
Explanation from original fix: 
    * suppose we have this document:
     *
     * { type: 'paragraph',
     *   children: [
     *     { text: 'foo ' },
     *     { text: 'bar' },
     *     { text: ' baz' }
     *   ]
     * }
     *
     * a double click on "bar" on chrome will create this range:
     *
     * anchor -> [0,1] offset 0
     * focus  -> [0,1] offset 3
     *
     * while on firefox will create this range:
     *
     * anchor -> [0,0] offset 4
     * focus  -> [0,2] offset 0

**Checks**

### DIFF
--- a/packages/slate/src/editor/marks.ts
+++ b/packages/slate/src/editor/marks.ts
@@ -4,6 +4,11 @@ import { Range } from '../interfaces/range'
 import { Path } from '../interfaces/path'
 import { Text } from '../interfaces/text'
 import { Element } from '../interfaces/element'
+import {Point} from "../interfaces";
+
+export const IS_FIREFOX =
+  typeof navigator !== 'undefined' &&
+  /^(?!.*Seamonkey)(?=.*Firefox).*/i.test(navigator.userAgent)
 
 export const marks: EditorInterface['marks'] = (editor, options = {}) => {
   const { marks, selection } = editor
@@ -11,14 +16,24 @@ export const marks: EditorInterface['marks'] = (editor, options = {}) => {
   if (!selection) {
     return null
   }
+  let { anchor, focus } = selection
 
   if (marks) {
     return marks
   }
 
   if (Range.isExpanded(selection)) {
-    const [match] = Editor.nodes(editor, { match: Text.isText })
-
+    if (IS_FIREFOX) {
+      const isEnd = Editor.isEnd(editor, anchor, anchor.path)
+      if (isEnd) {
+        const after = Editor.after(editor, anchor as Point)
+        // Editor.after() might return undefined
+        anchor = (after || anchor)
+      }
+    }
+    const [match] = Editor.nodes(editor, { match: Text.isText, at: {
+      anchor, focus
+    } })
     if (match) {
       const [node] = match as NodeEntry<Text>
       const { text, ...rest } = node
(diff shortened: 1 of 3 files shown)


packages/slate/src/editor/marks.ts
```diff
@@ -4,21 +4,36 @@ import { Range } from '../interfaces/range'
...
+  let { anchor, focus } = selection
 
   if (marks) {
     return marks
   }
 
   if (Range.isExpanded(selection)) {
-    const [match] = Editor.nodes(editor, { match: Text.isText })
-
+    if (IS_FIREFOX) {
```
12joan, line 26: I think we can safely remove this check in favour of cross-browser consistency

12joan (changes requested):
Nice! Can you add a unit test for the case that affects Firfox?

packages/slate/src/editor/marks.ts
Kaporos: isn't the issue only firefox-related ?

Kaporos:
> Nice! Can you add a unit test for the case that affects Firfox, and add a changeset?

I tried to add one, i hope i did it the right way :D

12joan:
I think this change should be a `patch` rather than a `minor`, since it's a bug fix. Minor usually means non-breaking feature changes.

packages/slate/src/editor/marks.ts
```diff
@@ -4,21 +4,36 @@ import { Range } from '../interfaces/range'
 import { Path } from '../interfaces/path'
 import { Text } from '../interfaces/text'
 import { Element } from '../interfaces/element'
+import {Point} from "../interfaces";
```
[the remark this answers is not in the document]
Kaporos: done !

packages/slate/src/editor/marks.ts
Kaporos: i tried, it should work on both without the if, i removed it :D

12joan:
Thanks for making those changes. Did you want to add a unit test?

Kaporos:
> Thanks for making those changes. Did you want to add a unit test?

IDK how to make them the right way..

12joan:
If you grant me commit access to [your fork](https://github.com/Kaporos/slate), I can add one later today. 🙂

Kaporos:
> If you grant me commit access to [your fork](https://github.com/Kaporos/slate), I can add one later today. 🙂

done !

12joan:
Thanks. I've added the test.

12joan (approved):

12joan:
The integrations test failures are unrelated to this PR and can be safely ignored

bradcypert:
Wow, awesome! Thank you for working on a fix for my issue. I had really wanted to use Slate but this was a blocker for me, so I started exploring alternatives -- now I can go back to slate! :D

dylans (approved):

PR #5621, opened 2024-03-28
chore: add shouldMergeNodesRemovePrevNode

felixfeng33:
**Description**
see https://slate-js.slack.com/archives/CC58ZGGU9/p1711614468642409

**Issue**
Fixes: (link to issue)

**Example**

**Context**

**Checks**
- [x ] The new code matches the existing patterns and styles.
- [x ] You've [added a changeset](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md) if changing functionality. (Add one with `yarn changeset add`.)


packages/slate/src/editor/index.ts
```diff
@@ -52,3 +52,4 @@ export * from './start'
 export * from './string'
 export * from './unhang-range'
 export * from './without-normalizing'
+export * from './shouldMergeNodesRemovePrevNode'
```
12joan, line 55: This should probably be kebab-case to match the other files here

packages/slate/src/transforms-node/merge-nodes.ts
```diff
@@ -133,12 +133,7 @@ export const mergeNodes: NodeTransforms['mergeNodes'] = (
     // prevent losing formatting when deleting entire nodes when you have a
     // hanging selection.
     // if prevNode is first child in parent,don't remove it.
-    if (
-      (Element.isElement(prevNode) && Editor.isEmpty(editor, prevNode)) ||
-      (Text.isText(prevNode) &&
-        prevNode.text === '' &&
-        prevPath[prevPath.length - 1] !== 0)
-    ) {
+    if (editor.shouldMergeNodesRemovePrevNode(prev, current)) {
```
12joan, line 136: Since we've got this on the Editor interface, should we use that here?

```suggestion
    if (Editor.shouldMergeNodesRemovePrevNode(editor, prev, current)) {
```

12joan (changes requested):

packages/slate/src/transforms-node/merge-nodes.ts
felixfeng33: You're right ,This is my first PR on here. Thank for your guidance

.changeset/early-mayflies-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+'slate': patch
+---
+
+add interface shouldMergeNodesRemovePrevNode
```
12joan, line 5:
```suggestion
---
'slate': minor
---

Add a `shouldMergeNodesRemovePrevNode` editor method to control when `Transforms.mergeNodes` should remove the previous node rather than carrying out a merge operation. 
```

packages/slate/src/transforms-node/merge-nodes.ts
```diff
@@ -133,12 +133,7 @@ export const mergeNodes: NodeTransforms['mergeNodes'] = (
     // prevent losing formatting when deleting entire nodes when you have a
     // hanging selection.
     // if prevNode is first child in parent,don't remove it.
-    if (
-      (Element.isElement(prevNode) && Editor.isEmpty(editor, prevNode)) ||
-      (Text.isText(prevNode) &&
-        prevNode.text === '' &&
-        prevPath[prevPath.length - 1] !== 0)
-    ) {
+    if (Editor.shouldMergeNodesRemovePrevNode(editor, prev, current)) {
```
12joan, line 136: We can remove the comment before this line since it's duplicated

12joan (changes requested):

12joan (approved):

Issue #3895, opened 2020-09-29
Request: Documentation should have a user option to display either TypeScript or JavaScript


timbuckley:
#### Do you want to request a _feature_ or report a _bug_?

Request

#### What's the current behavior?

Currently there is no way to toggle between JS and TS examples on the documentation ([example](https://docs.slatejs.org/concepts/01-interfaces)), it only shows TS examples.

#### What's the expected behavior?

User should see a toggle above each example (or at the top of all pages) to see a TS or JS version of example usage.

Walsker:
I'd like to tackle this, however I'd like to clarify something. Some snippets only include things like [interfaces](https://docs.slatejs.org/concepts/06-editor) or [types](https://docs.slatejs.org/concepts/03-locations#path), which don't directly translate back to JS. How should that be represented when toggled to JS, or should that not be an option for these kinds of snippets?

thesunny:
@Walsker @timbuckley I wonder if the right approach is to use the TypeScript to generate the JavaScript. I don't know if the JavaScript will appear natural, but I feel like it's worth a try. Then we don't have to keep two separate code bases (one in TS and one in JS) in sync.

With respect to `interface`s and `type`s, I don't think those have a JS equivalent anyways so they should be ignored.

RavenColEvol:
@thesunny @ianstormtaylor is it still a good first issue? I would like to hop in and take this issue.

thesunny:
@RavenColEvol feels like a good first issue to me!

Since it affects documentation and not Slate itself, it is simultaneously easy to confirm that it works (does the code look right) and also shouldn't introduce any issues into Slate.

Issue #5612, opened 2024-03-02
examples are not 100% type safe


0xMALVEE:
**Problem**
Examples are not 100% typesafe after I copy code I get a lot of type errors. 
![image](https://github.com/ianstormtaylor/slate/assets/35369563/422c8a2f-9a60-4b7d-84e2-d77e5b017890)

**Solution**
Improve type safety 

**Alternatives**
No alternative

**Context**
Typescript improvements of types in examples.

dylans:
It is indeed difficult to keep the examples safe as every version of TS potentially introduces changes that were not caught previously.

PRs are always welcome to update them.

alexander-larsson:
Sorry if this is not 100% related to this issue but I just started using Slate and got really confused by this when trying to get some code working based on the examples. Is it just the types that are wrong or does the `type` for example in fact not exist on BaseElement so the example code is wrong and based on some older version of Slate?

KarsonJo:
In fact most type errors stem from not defining the editor types. This issue is actually addressed [in the documentation](https://docs.slatejs.org/concepts/12-typescript). You can also get the type definition from [inside the example](https://github.com/ianstormtaylor/slate/blob/f3be9f12925e85a38f27168419dd78d55784ba51/site/examples/custom-types.d.ts#L126). But it would be great if this information could be shown in the comments of the examples.

Issue #4915, opened 2022-03-25
Property 'type' does not exist on type 'Node' when trying to apply custom formatting


jonahallibone:
**Description**
From the example (which isn't in TS, but I am doing my best to translate as I follow along), using match seemed to work with custom types. This is my code for toggling Custom Formatting:
```
const [match] = Editor.nodes(editor, {
  match: (n: Node) => {
    return n.type === "code";
  },
});
```
However, I get the warning:
```
Property 'type' does not exist on type 'Node'.
  Property 'type' does not exist on type 'CustomText'
```

**Recording**
<img width="817" alt="Screen Shot 2022-03-25 at 3 25 26 AM" src="https://user-images.githubusercontent.com/8886318/160074073-e7399316-ffcb-49b8-a744-1db4a91c6236.png">

**Environment**
- Slate Version:
```
"slate": "^0.76.0",
"slate-history": "^0.66.0",
"slate-react": "^0.76.0",
```
- Operating System: macOS

- TypeScript Version:
`"typescript": "^4.6.2"`

**Context**
I'll add the relevant code below

```
declare module "slate" {
  interface CustomTypes {
    Editor: BaseEditor & ReactEditor;
    Element: CustomElement;
    Text: CustomText;
  }
}

type CustomText = { text: string };
type CustomElement =
  | { type: "paragraph"; children: CustomText[] }
  | { type: "code"; children: CustomText[] };

const RichEditor = (): ReactElement => {
  const editorRef = useRef<Editor>();

  if (!editorRef.current) {
    editorRef.current = withReact(withHistory(createEditor()));
  }
  const editor = editorRef.current;

  const initialValue: Descendant[] = [
    {
      type: "paragraph",
      children: [{ text: "A line of text in a paragraph." }],
    },
  ];

  const [value, setValue] = useState<Descendant[]>(initialValue);

  const customRenderer = useCallback((props) => {
    if (props.element.type === "paragraph") {
      return <Text {...props} />;
    } else if (props.element.type === "code") {
      return <Code {...props} />;
    }
  }, []);

  const formatCodeBlock = () => {
    const [match] = Editor.nodes(editor, {
      match: (n: Node) => {
        console.log(n);

        return n.type === "code";
      },
    });

    Transforms.setNodes(
      editor,
      { type: match ? "paragraph" : "code" },
      { match: (n) => Editor.isBlock(editor, n) }
    );
  };

  return (
    <Box>
      <Heading>Editor</Heading>
[...]

Issue #4395, opened 2021-07-27
Docs: Typescript examples are faulty or missing


schlenger:
**Problem**
I tried to setup a sample project by reading the docs and stumbled upon some missing dependencies and wrongly formatted code which is hard to fix if someone tries to implement the initial setup as a beginner.

**Solution**
The sections in https://docs.slatejs.org/walkthroughs/01-installing-slate as well as https://docs.slatejs.org/concepts/12-typescript need formatting updates as well as a check if things work if you stick to the documentation.

**Context**
I can take care about this, if it is helpful.

schlenger:
Sidenote: When implementing https://docs.slatejs.org/walkthroughs/05-executing-commands in TS, there is an issue, that `bold` is used for spans in prior and in this example for the complete blocks

ilovecommits:
I found the docs hard to follow too when using TS. The best solution would be to split every page into two separate JS and TS sections.

Andrewnt219:
@ilovecommits In the `01-installing-slate`, the TypeScript code for setting up the component has a compilation error. `CustomElement` should be an array.
```ts
// Correct version
const App = () => {
  const initialValue: CustomElement[] = []
  //...
}
```
![image](https://user-images.githubusercontent.com/52666982/130633824-1357e342-402c-402c-a018-346e38e0befd.png)

ilovecommits:
> @ilovecommits In the `01-installing-slate`, the TypeScript code for setting up the component has a compilation error. `CustomElement` should be an array.
> 
> ```ts
> // Correct version
> const App = () => {
>   const initialValue: CustomElement[] = []
>   //...
> }
> ```
> 
> ![image](https://user-images.githubusercontent.com/52666982/130633824-1357e342-402c-402c-a018-346e38e0befd.png)

Not sure why you tagged me, but just submit a PR.

TrySpace:
Same, getting:

```
Subsequent property declarations must have the same type.  Property 'Element' must be of type 'CustomElement', but here has type 'CustomElement'.
```
When trying the first example

BrentFarese:
> Sidenote: When implementing https://docs.slatejs.org/walkthroughs/05-executing-commands in TS, there is an issue, that `bold` is used for spans in prior and in this example for the complete blocks

@TrySpace were you able to resolve this?  I am getting a similar error in my codebase.

maral:
[...]

PR #5635, opened 2024-04-14
Add TypeScript code variant to walkthroughs

maral:
**Description**
One of the frequent issues here seems to be the lack of TypeScript example code in the documentation (see the issues linked) - it's upon the user to guess them correctly. I happened to struggle through that a bit too and decided to do something about it instead of complaining.

**This PR is a draft at the moment** - before converting the whole walkthrough, I would like to get a second opinion first, so I know I am going about it right the first time.

**Issue**
Fixes: #3895, #4290, #4395, #4915, #5612

**Context**
The approach I've chosen is to use GitBook's [Tabs](https://docs.gitbook.com/content-editor/blocks/tabs). For each code snippet (except those explicitly TypeScript-only), I added a tabs wrapper. Even if the snippets are the same for TS and JS, for the consistency I added the tabs there.

**Potential problems:**
- The markdown gets a bit verbose and it's not as nice to manage in the long run.
- TypeScript API might still be changing until the 1.0 version is released.
- I don't know how to preview the docs - I [asked about it on Discord](https://slate-js.slack.com/archives/CC58ZGGU9/p1710430945452449), I didn't get any helpful reply.
- As far as I found, changing one tab from JS to TS will not change any other tabs, so folks wanting TypeScript (which I hope is the majority) will end up having to switch manually every time they want to copy the code - not ideal. I created [an issue](https://github.com/GitbookIO/gitbook/issues/2278) in Gitbook's repo, so we'll see if I get any answer there about this.

**Checks**

### DIFF
--- a/docs/walkthroughs/01-installing-slate.md
+++ b/docs/walkthroughs/01-installing-slate.md
@@ -28,6 +28,10 @@ import { Slate, Editable, withReact } from 'slate-react'
 
 Before we use those imports, let's start with an empty `<App>` component:
 
+{% tabs %}
+
+{% tab title="JavaScript" %}
+
 ```jsx
 // Define our app...
 const App = () => {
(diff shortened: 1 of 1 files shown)


dylans:
This looks useful and is appreciated @maral !

I wouldn't worry too much about things changing with respect to TypeScript, that's basically always a risk with TS.

It would be nice if GitBook has a solution for the issue you mentioned.

dylans:

PR #5670, opened 2024-06-28
fix: site example(Search Highlighting) range

WindrunnerMax:
**Description**
When I was testing the 'decorations' feature on the[ official website (search-highlighting)](https://www.slatejs.org/examples/search-highlighting), I noticed that there were some issues with supporting content that had been split into multiple formats.

**Issue**
none

**Example**
When I search for "adds", it works pretty well. However, when I perform a search that spans nodes, it doesn't highlight the content very effectively.

20240628191020_rec_

**Context**
I think it's necessary to tackle this issue. Initially, I planned to start with [Range.intersection](https://github.com/ianstormtaylor/slate/blob/97c88d/packages/slate-react/src/hooks/use-children.tsx#L55), by calculating the overlapping ranges at the parent node and judging the intersection of multiple nodes to address this issue. However, I realized that this might impose a burden on the rendering performance of Slate, as it may need to constantly traverse nodes for searching.

Therefore, I resolved this issue here by modifying the example of the site. By clearly identifying the content to be searched and the boundaries of the split nodes to obtain smaller values to form a range, even if the content being searched is very long, it will only save the search content within its own range.

20240628192046_rec_

Additionally, I noticed that there's no option for `site` changes in the changeset config. Therefore, I believe that this change doesn't require a changeset.

**Checks**

### DIFF
--- a/site/examples/search-highlighting.tsx
+++ b/site/examples/search-highlighting.tsx
@@ -13,22 +13,39 @@ const SearchHighlightingExample = () => {
     ([node, path]) => {
       const ranges = []
 
-      if (search && Text.isText(node)) {
-        const { text } = node
-        const parts = text.split(search)
-        let offset = 0
-
-        parts.forEach((part, i) => {
-          if (i !== 0) {
-            ranges.push({
-              anchor: { path, offset: offset - search.length },
-              focus: { path, offset },
-              highlight: true,
-            })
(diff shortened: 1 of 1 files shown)


dylans (approved):
Thanks @WindrunnerMax.

I agree with your sentiment as well (updating the example, for now).

Issue #5672, opened 2024-07-04
ReactEditor.focus triggers onValueChange unexpectedly


gwy15:
**Description**
Playground: https://codesandbox.io/s/api-focus-triggers-onvaluechange-dlsfxx?file=/src/App.tsx:962-975

for a simple `<Slate />`, mannually focus / blur using cursor does not trigger its `onValueChange` event.
However if using `ReactEditor.focus(editor);`, `onValueChange` event is triggered (and only for the first call, weirdly).

**Recording**
![20240704170143_rec_](https://github.com/ianstormtaylor/slate/assets/23229760/cb6e4b04-d481-4d3b-aad4-43cac1a493fb)

**Sandbox**
https://codesandbox.io/s/api-focus-triggers-onvaluechange-dlsfxx?file=/src/App.tsx:962-975

**Steps**
To reproduce the behavior:
1. Go to [Sandbox](https://codesandbox.io/s/api-focus-triggers-onvaluechange-dlsfxx?file=/src/App.tsx:962-975)
2. Click on the button
3. a onValueChange event is triggered.

**Expectation**
A consistent behavior (i.e., either both mannually focus and api focus triggers the onValueChange event, or neither does, which I guess is more reasonable.

**Environment**
- Slate Version: 0.103
- Operating System: macOS
- Browser: Chrome and Safari
- TypeScript Version: 3.7.5

WindrunnerMax:
In the current implementation, focusing (by ReactEditor.focus) from editor.selection being null to a specific value will inevitably trigger OnChange. 

Maybe handling it during the apply phase, [like here](https://github.com/WindrunnerMax/DocEditor/blob/master/packages/core/src/event/index.ts#L22), would be more appropriate.

WindrunnerMax:
Before this issue is resolved, I think you can use `editor.apply` to dispatch events, something [like this](https://github.com/WindrunnerMax/DocEditor/blob/4adc9b8/packages/core/src/event/index.ts#L22).

PR #5677, opened 2024-07-07
fix: unexpected event triggered when using ReactEditor.focus

WindrunnerMax:
**Description**
When opening the document for the first time, using `ReactEditor.focus` to focus on the document unexpectedly triggers the `onValueChange` event, and there will also be an additional `onChange `event triggered.

**Issue**
Fixes:  #5672

**Example**
The debug code is shown below, and the autoFocus property of Editable is turned off.

https://github.com/ianstormtaylor/slate/blob/631a911ca96e5ac55b80bb260e1e8d114ba76337/site/examples/richtext.tsx#L30-L50

The effect when calling ReactEditor.focus is shown in the image, where you can see the unexpected event triggers.

bad59c70-1387-4948-8fd3-1e28937daeb4

**Context**

Actually, calling `onChange` in `ReactEditor.focus` isn't necessary. If needed, `Transforms.select` will handle triggering the event.

img_v3_02ci_bd5439c3-9fc0-4e56-9759-36ce9e8ab98g

**Checks**

### DIFF
--- a/.changeset/mighty-trainers-juggle.md
+++ b/.changeset/mighty-trainers-juggle.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+fix unexpected event triggered when using `ReactEditor.focus`
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -445,7 +445,6 @@ export const ReactEditor: ReactEditorInterface = {
       // Create a new selection in the top of the document if missing
       if (!editor.selection) {
         Transforms.select(editor, Editor.start(editor, []))
-        editor.onChange()
       }
       // IS_FOCUSED should be set before calling el.focus() to ensure that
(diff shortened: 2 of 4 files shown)


WindrunnerMax:
I think I need to add a unit test triggered by a focus event. Please hold on a moment.

dylans (approved):
Thanks @WindrunnerMax . I'm going to approve and land this, but it's possible I'm forgetting a reason why that onChange call was in there. If there was a valid reason we may need to revert, we'll see!

WindrunnerMax:
@dylans If necessary, I am more than willing to rectify or assist in reverting this issue again and add the corresponding unit tests. 😄

[fc0818](https://github.com/ianstormtaylor/slate/commit/fc081816e08ade6838d05a96f84088de9f2734ce#diff-06bf833c7913658864514eb335e23b8ddb7a55888d004bd68d8d1293933103cbR446-R449) This is the commit and unit tests that were introduced at the time. I believe they can serve as a reference.

PR #5685, opened 2024-07-20
fix: prevent IME input interruption on MacOS

WindrunnerMax:
**Description**

In the current [Mentions Example](https://www.slatejs.org/examples/mentions), the issue with #3490 still persists. Although I noticed it was fixed in #5013, it was reverted back in #5360 .

So this seems to be a compatibility issue with MacOS. Therefore, we check the environment here and only modify the spacer tag position on MacOS.

**Issue**
Fixes: #3490

**Example**

The current issues.

20240720192010_rec_

After making adjustments.

20240720192658_rec_

 It's also running smoothly on my Android device.

**Checks**

### DIFF
--- a/site/examples/mentions.tsx
+++ b/site/examples/mentions.tsx
@@ -1,4 +1,11 @@
-import React, { useMemo, useCallback, useRef, useEffect, useState } from 'react'
+import React, {
+  useMemo,
+  useCallback,
+  useRef,
+  useEffect,
+  useState,
+  Fragment,
+} from 'react'
 import { Editor, Transforms, Range, createEditor, Descendant } from 'slate'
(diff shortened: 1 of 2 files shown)


dylans (changes requested):
@WindrunnerMax Thanks for addressing this, the fix looks reasonable.

Please add a changeset and then we'll land this. Thanks!

WindrunnerMax:
@dylans I'm happy to do this, but my changes are limited to the Mention Example. According to the current .changeset configuration, it shouldn't be necessary to set a changeset.md for packages.

dylans (approved):
You're right, I missed that it was only a fix for the example.

PR #5696, opened 2024-08-14
Add withMerging

felixfeng33:
**Description**
Currently, Slate lacks an API to merge a series of history records to the top of the stack. In some cases, the withMerging function is very useful, for example the upload status from pending to success and AI generate a series of text  withMerging can make sure to remove it in one times redo.

**Issue**
Fixes: (link to issue)

**Example**

**Context**

**Checks**

### DIFF
--- a/.changeset/honest-rice-doubt.md
+++ b/.changeset/honest-rice-doubt.md
@@ -1,0 +1,5 @@
+---
+'slate-history': minor
+---
+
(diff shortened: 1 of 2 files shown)


dylans (approved):
This looks fine. Ideally we would add something to the docs as well, but I won't block the PR for that.

PR #5716, opened 2024-09-09
Fix selections with non-void non-editable focus

### LINKED ISSUE #5714: Slate selection doesn't update when dragging browser selection focus into `contentEditable={false}` element
## Description

When I drag my browser selection focus into a `contentEditable={false}` DOM element that's being rendered by a non-void Slate element renderer, Slate's selection doesn't get updated. At times, this causes non-highlighted text to be deleted when pressing <kbd>Backspace</kbd> (first recording below). At other times, this causes a <kbd>Backspace</kbd> press to crash the page (second recording below).

## Recording

A demonstration of the bug in Slate's checklists example, in which pressing Backspace causes non-highlighted text to be deleted

A demonstration of the bug in Slate's checklists example, in which pressing Backspace causes the page to crash

## Sandbox

https://www.slatejs.org/examples/check-lists

## Expectation

In the above recordings, that only the visually highlighted text would be deleted on <kbd>Backspace</kbd> and that the page would not crash.

## Environment

- Slate Version: `main` (confirmed locally with 7e77a93)
- Browser: Chrome, Firefox

## Context

I'm in the process of working on a fix, since this bug affects my project (we have a block type that renders a larger `contentEditable={false}` element above its text content, so it's a bit easier to trigger). I'll submit a PR once that's ready; just wanted to give y'all a heads-up here first.

TyMick:
## Description

Currently, when I drag my browser selection focus into a `contentEditable={false}` DOM element that's being rendered by a non-void Slate element renderer, Slate's selection doesn't get updated. At times, this causes non-highlighted text to be deleted when pressing <kbd>Backspace</kbd> (first example below). At other times, this causes a <kbd>Backspace</kbd> press to crash the page (second example below).

The fix in this PR is to set the focus point to the edge of the nearest Slate leaf when the selection focus is in a non-void, non-editable node, instead of disregarding such selections.

## Issue

Fixes #5714

## Example

These examples use https://www.slatejs.org/examples/check-lists, where the “non-void, non-editable” node is [this `<span>`](https://github.com/ianstormtaylor/slate/blob/7e77a932f0489a9fff2d8a1957aa2dd9b324aa78/site/examples/check-lists.tsx#L153-L170).

#### Before

A demonstration of the bug in Slate's checklists example, in which pressing Backspace causes non-highlighted text to be deleted

#### After

A demonstration of this PR's fix of the incorrect deletion, in which pressing Backspace only deletes the highlighted text

#### Before

A demonstration of the bug in Slate's checklists example, in which pressing Backspace causes the page to crash

#### After

A demonstration of this PR's fix of the page crash, in which pressing Backspace only deletes the highlighted text

## Checks

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -255,9 +255,9 @@ export const Editable = forwardRef(
               ReactEditor.hasEditableTarget(editor, anchorNode) ||
               ReactEditor.isTargetInsideNonReadonlyVoid(editor, anchorNode)
 
-            const focusNodeSelectable = ReactEditor.hasTarget(editor, focusNode)
+            const focusNodeInEditor = ReactEditor.hasTarget(editor, focusNode)
 
-            if (anchorNodeSelectable && focusNodeSelectable) {
+            if (anchorNodeSelectable && focusNodeInEditor) {
               const range = ReactEditor.toSlateRange(editor, domSelection, {
                 exactMatch: false,
                 suppressThrow: true,
@@ -277,7 +277,7 @@ export const Editable = forwardRef(
             }
 
             // Deselect the editor if the dom selection is not selectable in readonly mode
-            if (readOnly && (!anchorNodeSelectable || !focusNodeSelectable)) {
+            if (readOnly && (!anchorNodeSelectable || !focusNodeInEditor)) {
               Transforms.deselect(editor)
             }
           }
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -255,9 +255,7 @@ export const Editable = forwardRef(
               ReactEditor.hasEditableTarget(editor, anchorNode) ||
               ReactEditor.isTargetInsideNonReadonlyVoid(editor, anchorNode)
 
-            const focusNodeSelectable =
-              ReactEditor.hasEditableTarget(editor, focusNode) ||
-              ReactEditor.isTargetInsideNonReadonlyVoid(editor, focusNode)
+            const focusNodeSelectable = ReactEditor.hasTarget(editor, focusNode)
 
             if (anchorNodeSelectable && focusNodeSelectable) {
               const range = ReactEditor.toSlateRange(editor, domSelection, {
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -691,7 +691,7 @@ export const ReactEditor: ReactEditorInterface = {
       searchDirection?: 'forward' | 'backward'
     }
   ): T extends true ? Point | null : Point => {
-    const { exactMatch, suppressThrow, searchDirection } = options
+    const { exactMatch, suppressThrow, searchDirection = 'backward' } = options
     const [nearestNode, nearestOffset] = exactMatch
       ? domPoint
       : normalizeDOMPoint(domPoint)
@@ -796,17 +796,32 @@ export const ReactEditor: ReactEditorInterface = {
       } else if (nonEditableNode) {
         // Find the edge of the nearest leaf in `searchDirection` from the
         // non-editable node
-        const leafNodes = Array.from(
-          editorEl.querySelectorAll(
-            // This editor's leaves (not those of a nested editor)
-            '[data-slate-leaf]:not(:scope [data-slate-editor] [data-slate-leaf])'
-          )
+        const getLeafNodes = (node: DOMElement | null | undefined) =>
+          node
+            ? node.querySelectorAll(
+                // Exclude leaf nodes in nested editors
+                '[data-slate-leaf]:not(:scope [data-slate-editor] [data-slate-leaf])'
+              )
+            : []
+        const elementNode = nonEditableNode.closest(
+          '[data-slate-node="element"]'
         )
-        leafNode =
-          (searchDirection === 'forward'
-            ? leafNodes.find(leaf => isAfter(nonEditableNode, leaf))
-            : leafNodes.findLast(leaf => isBefore(nonEditableNode, leaf))) ??
-          null
+
+        if (searchDirection === 'forward') {
+          const leafNodes = [
+            ...getLeafNodes(elementNode),
+            ...getLeafNodes(elementNode?.nextElementSibling),
+          ]
+          leafNode =
+            leafNodes.find(leaf => isAfter(nonEditableNode, leaf)) ?? null
+        } else {
+          const leafNodes = [
+            ...getLeafNodes(elementNode?.previousElementSibling),
+            ...getLeafNodes(elementNode),
+          ]
+          leafNode =
+            leafNodes.findLast(leaf => isBefore(nonEditableNode, leaf)) ?? null
+        }
 
         if (!leafNode) {
           offset = 1
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -794,8 +794,7 @@ export const ReactEditor: ReactEditorInterface = {
           })
         }
       } else if (nonEditableNode) {
-        // Find the edge of the nearest leaf in `searchDirection` from the
-        // non-editable node
+        // Find the edge of the nearest leaf in `searchDirection`
         const getLeafNodes = (node: DOMElement | null | undefined) =>
           node
             ? node.querySelectorAll(
--- a/packages/slate-react/src/plugin/react-editor.ts
+++ b/packages/slate-react/src/plugin/react-editor.ts
@@ -823,9 +823,7 @@ export const ReactEditor: ReactEditorInterface = {
             leafNodes.findLast(leaf => isBefore(nonEditableNode, leaf)) ?? null
         }
 
-        if (!leafNode) {
-          offset = 1
-        } else {
+        if (leafNode) {
           textNode = leafNode.closest('[data-slate-node="text"]')!
           domNode = leafNode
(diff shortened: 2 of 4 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -255,9 +255,7 @@ export const Editable = forwardRef(
               ReactEditor.hasEditableTarget(editor, anchorNode) ||
               ReactEditor.isTargetInsideNonReadonlyVoid(editor, anchorNode)
 
-            const focusNodeSelectable =
-              ReactEditor.hasEditableTarget(editor, focusNode) ||
-              ReactEditor.isTargetInsideNonReadonlyVoid(editor, focusNode)
+            const focusNodeSelectable = ReactEditor.hasTarget(editor, focusNode)
```
TyMick, line 258: I'm not sure `focusNodeSelectable` is still the best name for this variable, since the non-void non-editable nodes still aren't selectable. `ReactEditor.toSlateRange` just changes the focus to a node that _is_ selectable now.

TyMick: Also, this PR only fixes non-selectable _focus_ points now, but should non-selectable _anchor_ points be fixed as well? It wouldn't be necessary for my project, but conceivably, if a non-void non-editable node contained non-interactive text, it would be easy to make a browser selection with the anchor inside that node and the focus outside it.

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -978,18 +1022,6 @@ export const ReactEditor: ReactEditorInterface = {
...
-    // inside a `contenteditable="false"` DOM node following the word, which
-    // will cause `toSlatePoint` to throw an error. (2023/03/07)
-    if (
-      'getAttribute' in focusNode &&
-      (focusNode as HTMLElement).getAttribute('contenteditable') === 'false' &&
-      (focusNode as HTMLElement).getAttribute('data-slate-void') !== 'true'
-    ) {
-      focusNode = anchorNode
-      focusOffset = anchorNode.textContent?.length || 0
-    }
```
TyMick, line 991: No longer needed with the changes below and in `ReactEditor.toSlatePoint`.

dylans:
Thanks @TyMick I will look at this one sometime this week, and try to answer your open questions as well.

TyMick:
Thanks, Dylan!

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -778,6 +793,35 @@ export const ReactEditor: ReactEditorInterface = {
...
+          editorEl.querySelectorAll(
+            // This editor's leaves (not those of a nested editor)
+            '[data-slate-leaf]:not(:scope [data-slate-editor] [data-slate-leaf])'
+          )
+        )
+        leafNode =
+          (searchDirection === 'forward'
+            ? leafNodes.find(leaf => isAfter(nonEditableNode, leaf))
+            : leafNodes.findLast(leaf => isBefore(nonEditableNode, leaf))) ??
+          null
```
TyMick, line 809: I've thought of a possible optimization here. Currently, `leafNodes` is a collection of all leaf nodes in the entire editor. But since all Slate elements contain leaves, the furthest we should have to search from the `nonEditableNode` is one element away. I'm going to edit this so that `leafNodes` contains only the leaf nodes in the current Slate element and the element next to it in `searchDirection`.

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -778,6 +793,35 @@ export const ReactEditor: ReactEditorInterface = {
...
+
+        if (!leafNode) {
+          offset = 1
+        } else {
```
TyMick, line 813: I copied this from the `else if (voidNode)` block, which prefixes it with `// COMPAT: In read-only editors the leaf is not rendered.` But I'm thinking that may not be applicable here. I'm considering replacing this with

```suggestion
        if (leafNode) {
```

TyMick:
I tried adding a Jest regression test for this, but my attempts to create a selection range with Testing Library's `UserEvent.pointer` weren't able to affect `editor.selection`, which remained `null` at the end of this test:

```tsx
test('fixes DOM selections that end in non-editable nodes', async () => {
  const user = userEvent.setup()
  const initialValue = [
    {
      type: 'check-list-item',
      checked: false,
      children: [{ text: 'Test item' }],
    },
  ]

  render(
    <>
      <div>Something outside the editor</div>
      <Slate editor={editor} initialValue={initialValue}>
        <Editable renderElement={CheckListItemElement} />
      </Slate>
    </>
  )

  await user.pointer([
    // Mouse down in the middle of the checklist item text
    { keys: '[MouseLeft>]', target: screen.getByText('Test item'), offset: 4 },
    // Move mouse outside editor
    { target: screen.getByText('Something outside the editor'), offset: 2 },
    // Move mouse to the checkbox
    { target: screen.getByRole('checkbox') },
    // Mouse up
    { keys: '[/MouseLeft]' },
  ])

  await waitFor(() => expect(editor.selection).toBeTruthy())
})

function CheckListItemElement({ attributes, children, element }: any) {
  const { checked } = element
  return (
    <div {...attributes}>
      <span contentEditable={false}>
        <input type="checkbox" checked={checked} />
      </span>
      <span suppressContentEditableWarning>{children}</span>
    </div>
  )
}
```

I think this may not be testable without adding an end-to-end framework like Playwright.

packages/slate-react/src/components/editable.tsx
TyMick: Fixing non-selectable anchor points would add extra complexity in that we'd have to decide what to do when a selection starts and ends in the same non-selectable node... This PR is already an improvement on the existing behavior as-is, so maybe fixing anchors can be saved for a future PR? 😅

packages/slate-react/src/plugin/react-editor.ts
TyMick: Though I only say this out of intuition; I haven't tested the Chrome triple-clicking situation described in the comment.

packages/slate-react/src/plugin/react-editor.ts
TyMick: Done in https://github.com/ianstormtaylor/slate/pull/5716/commits/9534c218343c72f535c45af5718037c8c42d9e89

packages/slate-react/src/components/editable.tsx
TyMick: Changed the variable name to `focusNodeInEditor` with https://github.com/ianstormtaylor/slate/pull/5716/commits/98630d11ebf14e40fdcfb59c080530e69c0bfb5e.

packages/slate-react/src/plugin/react-editor.ts
TyMick: Done in https://github.com/ianstormtaylor/slate/pull/5716/commits/1f368a9748298dac1ed8870031d2268ae14011e0

packages/slate-react/src/plugin/react-editor.ts
TyMick: I should find the PR that added this block and figure out how to replicate the situation to make sure it's still working...

dylans:
> I tried adding a Jest regression test for this, but my attempts to create a selection range with Testing Library's `UserEvent.pointer` weren't able to affect `editor.selection`, which remained `null` at the end of this test:
> 
> ```tsx
> test('fixes DOM selections that end in non-editable nodes', async () => {
>   const user = userEvent.setup()
>   const editor = withReact(createEditor())
>   const initialValue = [
>     {
>       type: 'check-list-item',
>       checked: false,
>       children: [{ text: 'Test item' }],
>     },
>   ]
> 
>   render(
>     <>
>       <div>Something outside the editor</div>
>       <Slate editor={editor} initialValue={initialValue}>
>         <Editable renderElement={CheckListItemElement} />
>       </Slate>
>     </>
>   )
> 
>   await user.pointer([
>     // Mouse down in the middle of the checklist item text
>     { keys: '[MouseLeft>]', target: screen.getByText('Test item'), offset: 4 },
>     // Move mouse outside editor
>     { target: screen.getByText('Something outside the editor'), offset: 2 },
>     // Move mouse to the checkbox
>     { target: screen.getByRole('checkbox') },
>     // Mouse up
>     { keys: '[/MouseLeft]' },
>   ])
> 
>   await waitFor(() => expect(editor.selection).toBeTruthy())
> })
> 
> function CheckListItemElement({ attributes, children, element }: any) {
>   const { checked } = element
>   return (
>     <div {...attributes}>
>       <span contentEditable={false}>
>         <input type="checkbox" checked={checked} />
>       </span>
>       <span suppressContentEditableWarning>{children}</span>
>     </div>
>   )
> }
> ```
> 
> I think this may not be testable without adding an end-to-end framework like Playwright.

We do actually have Playright in there, just not too many tests yet.

TyMick:
Nope. I have spent hours trying to

1. mouse down in the first checklist item's text
2. move mouse into the paragraph text above it
3. move mouse into the checkbox
4. mouse up
5. press <kbd>Backspace</kbd>

in a Playwright test, but the behavior is far too flaky to be able to reliably create the same text selection by dragging the mouse. And I can't reproduce the bug with just the keyboard. Can't think of any other way to test it.

packages/slate-react/src/plugin/react-editor.ts
TyMick: Alright, found it was added in #5343, and I found a way to replicate that Chrome triple-click situation. I've confirmed that it still works after this PR's changes. 👍🏼

dylans:
@TyMick I appreciate you trying. These interactions are not easy to test.

dylans (approved):

PR #5736, opened 2024-10-08
fix: chromium interrupt IME when moving cursor

WindRunnerMax:
**Description**
I found an issue in Chromium: when a node (A) is set to contenteditable=false and is followed by a text node (B), activating the IME between A and B and then holding the left arrow key to adjust input causes the IME to disappear and leads to unexpected state issues, such as composition events. 

This problem does not occur in Firefox and Safari, only in browsers with the Chromium engine. 

This issue is also mentioned in https://github.com/ianstormtaylor/slate/issues/3490#issuecomment-2241179871.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/3490#issuecomment-2241179871

**Example**
The current issues.

After making adjustments.

**Context**
Here's the minimal demo for this issue: when you activate the IME before the "!" character and then hold the left arrow key, the input method gets interrupted and disappears.

```html
<div contenteditable="true"><span contenteditable="false" style="background: #eee;">Void</span><span>!</span></div>
```

To fix this issue, you need to create the following DOM structure. The key is that the outer span tag has a display:inline-block style, and the child div tag has a contenteditable=false attribute. This is magic.

```html
<div contenteditable="true"><span contenteditable="false" style="background: #eee; display: inline-block;"><div contenteditable="false">Void</div></span><span>!</span></div>
```

**Checks**

### DIFF
--- a/site/examples/js/mentions.jsx
+++ b/site/examples/js/mentions.jsx
@@ -220,18 +220,22 @@ const Mention = ({ attributes, children, element }) => {
       data-cy={`mention-${element.character.replace(' ', '-')}`}
       style={style}
     >
-      {IS_MAC ? (
-        // Mac OS IME https://github.com/ianstormtaylor/slate/issues/3490
-        <Fragment>
-          {children}@{element.character}
-        </Fragment>
-      ) : (
-        // Others like Android https://github.com/ianstormtaylor/slate/pull/5360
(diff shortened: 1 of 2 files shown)


dylans (approved):

dylans:
Thanks @WindRunnerMax !

PR #5741, opened 2024-10-16
fix: Prevent ReactEditor.toDOMRange crash in setDomSelection

AdrienPoupa:
**Description**
Somehow related to https://github.com/ianstormtaylor/slate/pull/5407 and https://github.com/ianstormtaylor/slate/pull/5571

I debugged an issue in our application where using the [TextExpander](https://textexpander.com/) extension on a Slate input would cause the editor to crash with `Cannot resolve a DOM point from Slate point` coming from `setDomSelection`'s call.

I was able to prevent the crash in our codebase using an ErrorBoundary as suggested [here](https://github.com/ianstormtaylor/slate/pull/5407#issuecomment-1673337395), but wondered why it was crashing to begin with. 

As far as I can tell, although the TextExpander text is not expanded, the editor recovers fine if we discard the error and the existing code supports having a null `newDomRange` so I figured I could let it be null if it can't calculate a `domPoint`.

This pattern is already used in `ensureDomSelection`:

```typescript
          const ensureDomSelection = (forceChange?: boolean) => {
            try {
              const el = ReactEditor.toDOMNode(editor, editor)
              el.focus()

              setDomSelection(forceChange)
            } catch (e) {
              // Ignore, dom and state might be out of sync
            }
          }
```

Please let me know if I missed something!

**Checks**

### DIFF
--- a/.changeset/serious-eels-pay.md
+++ b/.changeset/serious-eels-pay.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fix ReactEditor.toDOMRange crash in setDomSelection
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -333,7 +333,7 @@ export const Editable = forwardRef(
         const focusNode = domSelection.focusNode
         let anchorNode
 
-        // COMPAT: In firefox the normal seletion way does not work
+        // COMPAT: In firefox the normal selection way does not work
         // (https://github.com/ianstormtaylor/slate/pull/5486#issue-1820720223)
         if (IS_FIREFOX && domSelection.rangeCount > 1) {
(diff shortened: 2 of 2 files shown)


dylans:
In general I think as a project we try to avoid squashing errors/failures because they might hide an issue for someone else which can make debugging a real pain. Let me think about this a bit. In general I think slate would benefit from an option to configure how to handle failures, but then I always end up back at "add an ErrorBoundary" which isn't a great answer for many users.

AdrienPoupa:
I think ultimately we need better error management as in https://github.com/ianstormtaylor/slate/pull/5407 but until we do this could be an easy stop gap. If I understand why this PR got stale, it's because we'd want to
1. Create a first PR that changes the return types to be `| undefined`
2. Introduce error management
I'm willing to take a stab at the former, but I feel we'd still benefit from the current PR to begin with.

Issue #5694, opened 2024-08-07
onDOMSelectionChange causes crash during concurrent updates


gblaketx:
**Description**
The editor can crash if the user is typing while other updates are being applied (for instance, applying remote user updates or automated edits). The user's typing triggers an `onDOMSelectionChange` call, but the selection update from this function is stale, which can cause the cursor to jump or crash the editor.

**Recording**
https://github.com/user-attachments/assets/a68453b1-df47-4324-a542-9313569958ca

**Sandbox**
https://codesandbox.io/s/slate-selection-change-repro-ztssmp

**Steps**
To reproduce the behavior:
1. Go to https://codesandbox.io/s/slate-selection-change-repro-ztssmp
2. In Chrome developer tools, under the Performance tab, enable CPU throttling. This approximates the behavior of a large document and makes the issue easier to repro, although it still occurs without throttling.
3. Click the "Simulate ops" button, which repeatedly adds new elements to the middle of the document.
4. Type continuously in the last line of the document
5. A "Cannot resolve a DOM point from Slate point" error is triggered

**Expectation**
Basic concurrent updates shouldn't crash the editor. In the example posted, the user should be able to continue to type on the last line while new lines are added above.

**Environment**
- Slate Version: 0.103.0
- Operating System: Mac
- Browser: Chrome

**Context**
I suspect the root cause of this issue is a stale `NODE_TO_INDEX` weak map. Because that map is updated after nodes rerender, if the editor content changes and a selection event fires before a re-render, the map can be stale.

Looking at the code, the sequence of events is:
1. A new line is inserted in the editor. React does not yet re-render
2. The user's typing triggers the [onDOMSelectionChange](https://github.com/ianstormtaylor/slate/blob/b8bf92dc7e42f7547128fcfe1c63e89dffb29032/packages/slate-react/src/components/editable.tsx#L208C11-L208C31) handler
3. The handler attempts to convert the user's selected DOM node to a Slate range in [toSlateRange](https://github.com/ianstormtaylor/slate/blob/b8bf92dc7e42f7547128fcfe1c63e89dffb29032/packages/slate-react/src/components/editable.tsx#L262), which calls [findPath](https://github.com/ianstormtaylor/slate/blob/b8bf92dc7e42f7547128fcfe1c63e89dffb29032/packages/slate-react/src/plugin/react-editor.ts#L839) on the resulting Slate Node
[...]

PR #5746, opened 2024-10-18
fix: invalidate node maps when nodes change in-between paints #5694.

DustinMackintosh:
**Description**
This fixes an editor crash where the user types quickly after earlier nodes have been added, split or deleted, but the node maps have not yet been regenerated on React render. 

**Issue**
Fixes: [(link to issue)](https://github.com/ianstormtaylor/slate/issues/5694)

**Example**
Recording (from issue linked above)
https://github.com/user-attachments/assets/a68453b1-df47-4324-a542-9313569958ca

Sandbox (from issue linked above)
https://codesandbox.io/s/slate-selection-change-repro-ztssmp

**Context**
The `NODE_TO_INDEX` WeakMap is built on each React render as part of the `useChildren` hook. Both `onDOMBeforeInput` and `onDOMSelectionChange` use these maps to set the selection. This change introduces a new `WeakMap` called `IS_NODE_MAP_DIRTY` to track whether the underlying nodes have changed (eg: `remove_node` or `insert_node`), allowing those methods to skip attempts to `setSelection` that could crash the component. 

Recommend you compare with "Ignore white space" selected.

**Checks**

### DIFF
--- a/.changeset/bright-fishes-protect.md
+++ b/.changeset/bright-fishes-protect.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Invalidate node maps when nodes change until next react paint
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -57,6 +57,7 @@ import {
 } from '../utils/environment'
 import Hotkeys from '../utils/hotkeys'
 import {
+  IS_NODE_MAP_DIRTY,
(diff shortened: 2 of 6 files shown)


Calyhre:
We have the same issue on our side, and this patch seems to solve it 👍

dylans (approved):
Ok, this looks reasonable. Obviously if it causes any major regressions we may need to revert it.

PR #5752, opened 2024-10-28
fix: Prevent domSelection.collapseToEnd crash in setDomSelection

leondt1:
**Description**
At the beginning of the paragraph, composing text and then deleting all unconfirmed characters causes the value of <code>domSelection.type</code> to be <code>None</code>, leading to an error.

iShot_2024-10-28_20 31 19

image

**Issue**
Fixes: (link to issue)

**Example**

**Context**

**Checks**

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -416,7 +416,11 @@ export const Editable = forwardRef(
           selection && ReactEditor.toDOMRange(editor, selection)
 
         if (newDomRange) {
-          if (ReactEditor.isComposing(editor) && domSelection.type !== 'None' && !IS_ANDROID) {
+          if (
+            ReactEditor.isComposing(editor) &&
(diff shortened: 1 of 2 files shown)


packages/slate-react/src/components/editable.tsx
```diff
@@ -416,7 +416,7 @@ export const Editable = forwardRef(
           selection && ReactEditor.toDOMRange(editor, selection)
 
         if (newDomRange) {
-          if (ReactEditor.isComposing(editor) && !IS_ANDROID) {
+          if (ReactEditor.isComposing(editor) && domSelection.type !== 'None' && !IS_ANDROID) {
```
Fukiii, line 419: eslint error
```
if (
    ReactEditor.isComposing(editor) &&
    hasDomSelection &&
    !IS_ANDROID
)
```

leondt1: fixed

PR #5755, opened 2024-10-29
Fix: Handle deleteText for more than Thai language

seanghay:
**Description**

Currently, it deletes a grapheme cluster at a time for Khmer, Lao, and Burmese. Add the same functionality as Thai language.

**Checks**


dylans (changes requested):
Please fix the prettier/linting rules and add a changeset and then I will land it.

PR #5763, opened 2024-11-08
Use extended `Editor` type in `useSlateWithV` return type

TyMick:
## Description

The `useSlateWithV` return object's `editor` property is currently typed as just a `ReactEditor`, which causes type errors in our app when we try to use it as the extended `Editor` type. This change brings `useSlateWithV`'s return type in line with those of `useSlate` and `useSlateStatic`.

## Checks

### DIFF
--- a/.changeset/eleven-cycles-kneel.md
+++ b/.changeset/eleven-cycles-kneel.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Use extended `Editor` type in `useSlateWithV` return type
--- a/docs/libraries/slate-react/hooks.md
+++ b/docs/libraries/slate-react/hooks.md
@@ -28,15 +28,15 @@ Get the current `selected` state of an element.
 
 ### Editor hooks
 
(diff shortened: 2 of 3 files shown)


TyMick:
Should we add a release note that this may be a breaking type change if someone isn't [defining a custom `Editor` type](https://docs.slatejs.org/concepts/12-typescript#defining-editor-element-and-text-types) that includes `ReactEditor`? Probably an uncommon situation, since anyone who hasn't extended `Editor` with `ReactEditor` would also have problems with `useSlate` and `useSlateStatic`.

dylans:
> Should we add a release note that this may be a breaking type change if someone isn't [defining a custom `Editor` type](https://docs.slatejs.org/concepts/12-typescript#defining-editor-element-and-text-types) that includes `ReactEditor`? Probably an uncommon situation, since anyone who hasn't extended `Editor` with `ReactEditor` would already have problems with `useSlate` and `useSlateStatic`.

We could, though I think the description is probably sufficient already, so I'm going to land it.

Issue #5771, opened 2024-11-29
Interference between Select and Anchor Operations in Collaboration


electroluxcode:
**Description**
Hello, I've noticed that there seem to be several issues in the user collaboration. So I'd like to discuss and communicate about them with you.
Issue One: Suppose there are two users, namely me and him. We are currently editing the same section of a document normally. He edits at a very fast speed, specifically at 50 QPS. Then I find that during this process, my anchor will be preempted and my selection doesn't take effect immediately. I believe this is not in line with what was expected. 

Issue Two: When we are collaborating on a section of text simultaneously, when I select the whole text, in my view, this text shouldn't continue to expand. However, when he continues to edit this text, the whole code block will still be fully selected.

it will reproduction when click `click me trigger`  button， it mock 50 qps 

The video reproduction and the relevant code are as follows.  You can notice that in the video, 

if you don't edit the same code block（`with slate` area）, there will be no situation of the selection being occupied. However, when you are editing the same code block（`33333333333333` area， and the selection can be made successfully when entering the same code block for the first time.）, the situation where the selection can't be made properly will occur.

and you can notice that I double-clicked on the code block, but the code block of the collaborating party remains in the fully selected state all the time.

**Recording**

video

https://github.com/user-attachments/assets/6e3b8a89-422f-418d-b0a3-a1ca4b8198c2

code
```tsx

import React, { useMemo, useCallback } from 'react'
import {
  Slate,
  Editable,
  withReact,
} from 'slate-react'
import {
  createEditor,
  Descendant,
} from 'slate'
import { withHistory } from 'slate-history'

const initialValue: Descendant[] = [
  {
    type: 'paragraph',
    children: [
      {
        text: 'With Slater',
      },
    ],
  },
  {
    type: 'paragraph',
    children: [{ text: '' }],
  },
]

const CheckListsExample = () => {
  const renderElement = useCallback(props => <Element {...props} />, [])
  const editor = useMemo(
    () => withHistory(withReact(createEditor())),
    []
  )
  window.editor = editor
  let offset = 0
  const onClick = ()=>{
    setInterval(() => {
      window.editor.apply({
        type: 'insert_text',
        path: [1, 0],
        enableSelectionOp: true,
        offset,
        text: '3',
      })
      offset++
    }, 20);
  }
  return (
   <>
   <button onClick={onClick}>click me trigger</button>
    <Slate editor={editor} initialValue={initialValue}>
      <Editable
[...]

PR #5772, opened 2024-11-29
feat:  add prop enableSelectionOp and onDOMSelectionChangeThrottleTime

electroluxcode:
**Description**
the issue of the anchor, it was discovered that the function for setting the anchor was filtered out during `throttle`. Currently, the default value for this is 100 milliseconds. If it is changed to 0 milliseconds, this problem can be solved.

As for the issue of selection, it can be considered to add a `new type` in the operation name `enableSelectionOp`. This type can be used to determine whether the selection of the collaborator needs to be applied to the co-collaborator.

This pull request mainly adds the passing of two parameters to solve this issue if developer want to resolve the problem like me. and It will not interfere with the original logic.

Finally, the achieved effect is as follows. It can be seen that both the selection and the anchor are as expected.

**Issue**
fix https://github.com/ianstormtaylor/slate/issues/5771

**Example**

https://github.com/user-attachments/assets/86bf9233-e39f-459d-a810-5031c7429cd0

```tsx

import React, { useMemo, useCallback } from 'react'
import {
  Slate,
  Editable,
  withReact,
} from 'slate-react'
import {
  createEditor,
  Descendant,
} from 'slate'
import { withHistory } from 'slate-history'

const initialValue: Descendant[] = [
  {
    type: 'paragraph',
    children: [
      {
        text: 'With Slater',
      },
    ],
  },
  {
    type: 'paragraph',
    children: [{ text: '' }],
  },
]

const CheckListsExample = () => {
  const renderElement = useCallback(props => <Element {...props} />, [])
  const editor = useMemo(
    () => withHistory(withReact(createEditor())),
    []
  )
  window.editor = editor
  let offset = 0
  const onClick = ()=>{
    setInterval(() => {
      window.editor.apply({
        type: 'insert_text',
        path: [1, 0],
        enableSelectionOp: false,
        offset,
        text: '3',
      })
      offset++
    }, 20);
  }
  return (
   <>
   <button onClick={onClick}>click me trigger</button>
    <Slate editor={editor} initialValue={initialValue}>
      <Editable
        onDOMSelectionChangeThrottleTime={0}
        scrollSelectionIntoView={()=>{
          console.log('scrollSelectionIntoView')
        }}
        renderElement={renderElement}
        placeholder="Get to work…"
        spellCheck
        autoFocus
      />
    </Slate>
   </>
  )
}
const Element = props => {
  const { attributes, children, element } = props

  switch (element.type) {
    default:
  }
}

export default CheckListsExample


```

### DIFF
--- a/.changeset/lovely-penguins-begin.md
+++ b/.changeset/lovely-penguins-begin.md
@@ -1,5 +1,5 @@
 ---
-"slate-react": patch
+'slate-react': patch
 'slate': patch
 ---
 
--- a/.changeset/lovely-penguins-begin.md
+++ b/.changeset/lovely-penguins-begin.md
@@ -1,6 +1,6 @@
 ---
-"slate-react": minor
-"slate": minor
+"slate-react": patch
+"slate": patch
 ---
 
-feat:  add prop enableSelectionOp and onDOMSelectionChangeThrottleTime
+add prop enableSelectionOp and onDOMSelectionChangeThrottleTime
--- a/.changeset/lovely-penguins-begin.md
+++ b/.changeset/lovely-penguins-begin.md
@@ -1,6 +1,6 @@
 ---
 "slate-react": patch
-"slate": patch
+'slate': patch
 ---
 
 add prop enableSelectionOp and onDOMSelectionChangeThrottleTime
--- a/.changeset/lovely-penguins-begin.md
+++ b/.changeset/lovely-penguins-begin.md
@@ -1,0 +1,6 @@
+---
+"slate-react": minor
+"slate": minor
+---
+
+feat:  add prop enableSelectionOp and onDOMSelectionChangeThrottleTime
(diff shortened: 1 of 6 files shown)


dylans (changes requested):
Nice improvement and thanks for the explanation.

I don't see an easy way to write an automated test for this, so I'll let that pass for now.

Could you please add a changeset and then I'll land it shortly?

.changeset/lovely-penguins-begin.md
```diff
@@ -0,0 +1,6 @@
+---
+"slate-react": patch
```
dylans, line 2:
```suggestion
'slate-react': patch
```

.changeset/lovely-penguins-begin.md
```diff
@@ -0,0 +1,6 @@
...
+"slate": patch
```
dylans, line 3:
```suggestion
'slate': patch
```

electroluxcode:
I have some new ideas for issue1 about https://github.com/ianstormtaylor/slate/issues/5771 . Could you please take a look at this draft? https://github.com/ianstormtaylor/slate/pull/5777.

dylans:

PR #5776, opened 2024-12-07
feat: add optional merge method to decoration object

yf-yang:
**Description**
Currently, when decorations overlap with each other and they are merging with each other, `Object.assign` is used.
However, if two decorations with the same key has different values, one of the decoration will overwrite the other.

``` js
// text content: `them are happy`
// decoration 1:
{
  anchor: {
    path: [0],
    offset: 0,
  },
  focus: {
    path: [0],
    offset: 1,
  },
  message: '`t` should be upper case.'
}

// decoration 2:
{
  anchor: {
    path: [0],
    offset: 0,
  },
  focus: {
    path: [0],
    offset: 4,
  },
  message: 'The correct word is `they`.'
}
```

In this example, two messages cannot be shown together due to `Object.assign` overwriting behavior.

**Example**

After adding a function `merge` to the decoration, two messages can be merged together:

``` js
// text content: `them are happy`

const merge = (leaf, decoration) => {
  const { messages, ...rest } = decoration;
  leaf.messages = [...(leaf.messages ?? []), ...messages];
  Object.assign(leaf, rest);
}

// decoration 1:
{
  anchor: {
    path: [0],
    offset: 0,
  },
  focus: {
    path: [0],
    offset: 1,
  },
  merge,
  messages: ['`t` should be upper case.']
}

// decoration 2:
{
  anchor: {
    path: [0],
    offset: 0,
  },
  focus: {
    path: [0],
    offset: 4,
  },
  merge,
  messages: ['The correct word is `they`.']
}
```

**Checks**

### DIFF
--- a/.changeset/tiny-cheetahs-ring.md
+++ b/.changeset/tiny-cheetahs-ring.md
@@ -1,0 +1,5 @@
+---
+'slate': minor
+---
+
+Add `merge` optional function to decorations and change related type signatures to `DecoratedRange`. Now developers can specify how two decoration object with the same key but different value are merged together if they overlap"
--- a/docs/api/nodes/text.md
+++ b/docs/api/nodes/text.md
@@ -26,7 +26,7 @@ If a `props.text` property is passed in, it will be ignored.
 
 If there are properties in `text` that are not in `props`, those will be ignored when it comes to testing for a match.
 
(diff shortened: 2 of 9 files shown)


dylans (approved):
Nice improvement, thank you!

PR #5777, opened 2024-12-07
fix: selectionchange render when user trigger

electroluxcode:
The PR is to completely solve issue1 in https://github.com/ianstormtaylor/slate/issues/5771 by controlling the triggering behavior of users.

This PR is inherited from https://github.com/ianstormtaylor/slate/pull/5772 . Due to uncertainty about the side effects brought by this PR, it is temporarily submitted as a draft PR. In this https://github.com/ianstormtaylor/slate/pull/5772, in order to solve issue1 inside, I added a variable named `onDOMSelectionChangeThrottleTime`. 

Then, I found that when this variable is set to the extremely small value of 0, there is still a 20-30% probability that the selection area is invalid when the QPS is 50, and a second click is required to select the selection area. 

Therefore, referring to the event triggering of quill, https://github.com/slab/quill/blob/20f02f55e30ca87c0c539632b36c6c52ca4670eb/packages/quill/src/core/selection.ts#L54-L68

The onClick event was changed to onMouseUp and onMouseDown (the main reason is that the timing of the onClick event is too late. When the onClick event is triggered, the selection area has already changed). And these two events generate a variable mouseDown, which is used to control the triggering of selectionchange. This ensures that only user actions can trigger selectionchange, and API actions cannot actively trigger it. The `setTimeout` in https://github.com/ianstormtaylor/slate/blob/5a1c728c62bcc6bafe60b9598946bb3adde21cdf/packages/slate-react/src/components/editable.tsx#L459-L462 seems to cause abnormal behavior of the selection area. Changing it to synchronously call state.isUpdatingSelection as in the PR will not cause abnormalities. What is this design for?

### DIFF
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -861,7 +861,11 @@ export const Editable = forwardRef(
       // (2019/11/04) https://github.com/facebook/react/issues/5785
       window.document.addEventListener(
         'selectionchange',
-        scheduleOnDOMSelectionChange
+        ()=>{
+          if(mouseDown.current){
+            scheduleOnDOMSelectionChange()
+          }
+        }
       )
 
       // Listen for dragend and drop globally. In Firefox, if a drop handler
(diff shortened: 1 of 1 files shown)


dylans:
@electroluxcode  Overall it seems reasonable and consistent with what I know about event timing, but it does appear to possibly be trigger a series of test failures (it's entirely possible the test suite is wrong and needs to be adjusted), or could be a flaky test.

dylans:

PR #5781, opened 2024-12-20
Update package.json: add `"slate-dom": "workspace:*",` to help IDE identify `import xx from "slate-dom"`

sws797:
add `"slate-dom": "workspace:*",` to help IDE identify `import xx from "slate-dom"`


dylans (changes requested):
This change is fine (and appreciated)... I think we need to also commit the updated lockfile for this to pass CI.

Issue #5782, opened 2024-12-20
Error with documentation regarding how to use renderPlaceholder


ianleeduckworth:
**Problem**
The documentation around renderPlaceholder appears to be wrong, specifically the example usage. I used the prescribed example exactly and the placeholder behavior seemed incorrect. I noticed that it is because the attributes object that comes in has a style tag that must be applied to the div element. As such I believe the right way to use this functionality is to extend style instead of overriding it

**Solution**
This is the recommended approach to make a custom placeholder per the docs referenced below:
```
<Editable
  renderPlaceholder={({ attributes, children }) => (
    <div {...attributes} style={{ fontStyle: 'italic', color: 'gray' }}>
      {children}
    </div>
  )}
/>
```
However this appears not to work because `attributes` that comes in has styles already attached to it. If these styles are overridden as in the example above, you get some undesirable behavior and some errors. I got it to work by modifying the example slightly as such:
```
<Editable
  placeholder="Enter text here..."
  renderPlaceholder={({ attributes, children }) => {
    const styledAttributes = {
      ...attributes,
      style: {
        ...attributes.style,
        color: controlPlaceholderColor,
        opacity: 1,
      },
    };
    return <div {...styledAttributes}>{children}</div>;
  }}
</Editable>
```
The above example does two things:

1. It respects the existing styles and only extends; this means that as long as you don't override critical properties like the position etc it will still work fine
2. It actually adds a placeholder which is what will be rendered as `children` in the `renderPlaceholder` function

**Alternatives**
I can't really see any alternatives; the functionality works fine. If I am correct it's just the docs that need to be updated

**Context**
Documentation I'm referencing: https://docs.slatejs.org/libraries/slate-react/editable.

PR #5783, opened 2024-12-22
Issue 5782

ianleeduckworth:
**Description**
Fixes issue where instructions for renderProps in the react-slate documentation, if followed properly, produce an undesirable effect. This PR updates instructions such that a user can re-style the placeholder while still maintaining desirable default behavior

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/5782

**Example**
```
<Editable
  placeholder="Enter text here..."
  renderPlaceholder={({ attributes, children }) => {
    const styledAttributes = {
      ...attributes,
      style: {
        ...attributes.style,
        color: 'gray',
      },
    };
    return <div {...styledAttributes}>{children}</div>;
  }}
/>
```

**Context**
It was noticed that, when following the docs exactly, for renderProps within react-slate that the placeholder was behaving strangely. It was then noticed that in the docs the style prop that came in from `attributes` was being completely overridden. This was creating a situation where the placeholder exhibited some strange and undesirable behavior. The documentation has been updated so that the `attributes` prop that comes in is extended upon instead of overwritten entirely. This allows a user to do things like change the color, opacity, etc without overwriting critical style related props.

**Checks**
No checks necessary; only a documentation change

### DIFF
--- a/docs/libraries/slate-react/editable.md
+++ b/docs/libraries/slate-react/editable.md
@@ -165,13 +165,21 @@ An example usage might look like:
 
 ```jsx
 <Editable
-  renderPlaceholder={({ attributes, children }) => (
-    <div {...attributes} style={{ fontStyle: 'italic', color: 'gray' }}>
-      {children}
-    </div>
-  )}
+  placeholder="Enter text here..."
+  renderPlaceholder={({ attributes, children }) => {
+    const styledAttributes = {
+      ...attributes,
+      style: {
+        ...attributes.style,
+        color: 'gray',
+      },
+    };
+    return <div {...styledAttributes}>{children}</div>;
+  }}
 />
 ```
+
+Note that the `attributes` prop that comes in will contain a `style` object already. This object contains important styling properties which will make the placeholder behave like a placeholder. As such, it is advisible to extend styles and to focus on things like changing colors, opacity etc. Changing positioning, for example, could cause undesirable behavior.
 
 #### `scrollSelectionIntoView?: (editor: ReactEditor, domRange: DOMRange) => void`
 


dylans (approved):

dylans (changes requested):
@ianleeduckworth any chance you could run prettier on this real quick and then I'll land it? Thanks!

docs/libraries/slate-react/editable.md
```diff
@@ -165,14 +165,22 @@ An example usage might look like:
...
+        ...attributes.style,
+        color: 'gray',
+      },
+    };
+    return <div {...styledAttributes}>{children}</div>;
+  }}
 />
 ```
 
+Note that the `attributes` prop that comes in will contain a `style` object already. This object contains important styling properties which will make the placeholder behave like a placeholder. As such, it is advisible to extend styles and to focus on things like changing colors, opacity etc. Changing positioning, for example, could cause undesirable behavior.
```
dylans, line 182:
```suggestion
Note that the `attributes` prop that comes in will contain a `style` object already. This object contains important styling properties which will make the placeholder behave like a placeholder. As such, it is advisable to extend styles and to focus on things like changing colors, opacity etc. Changing positioning, for example, could cause undesirable behavior.
```

docs/libraries/slate-react/editable.md
```diff
@@ -165,14 +165,22 @@ An example usage might look like:
...
+  placeholder="Enter text here..."
+  renderPlaceholder={({ attributes, children }) => {
+    const styledAttributes = {
+      ...attributes,
+      style: {
+        ...attributes.style,
+        color: 'gray',
+      },
+    };
+    return <div {...styledAttributes}>{children}</div>;
```
dylans, line 177:
```suggestion
    return <div {...styledAttributes}>{children}</div>
```

dylans (changes requested):
These are the two changes needed to make this pass.

dylans (approved):

docs/libraries/slate-react/editable.md
```diff
@@ -165,14 +165,22 @@ An example usage might look like:
...
-  )}
+  placeholder="Enter text here..."
+  renderPlaceholder={({ attributes, children }) => {
+    const styledAttributes = {
+      ...attributes,
+      style: {
+        ...attributes.style,
+        color: 'gray',
+      },
+    };
```
dylans:
```suggestion
    }
```

PR #5788, opened 2024-12-26
docs: document `props.onValueChange` and `props.onSelectionChange`

christianhg:
**Description**

Documents `Slate` props added in #5526.


dylans (approved):
Thanks @christianhg !

Issue #5791, opened 2025-01-12
Selection reverts when button click in void element causes input to appear


12joan:
**Description**
Due to a regression introduced in Chrome 127, clicking a button inside a void element that causes an input element to appear causes the selection to incorrectly revert to its previous location.

**Recording**

https://github.com/user-attachments/assets/bfcd6d45-7d9a-48f3-8cf8-09b406197bad

**Sandbox**
https://codesandbox.io/p/sandbox/p553rr

**Steps**
To reproduce the behavior:
1. Open the sandbox in Chrome 127 or newer
2. Click anywhere in block 1 to select it (indicated by grey background)
3. Click the button in block 2 (which is a void block element)
4. The selection incorrectly reverts to block 1

**Expectation**
Ordinarily, clicking a button inside a void element causes that void element to be selected. When clicking the button in block 2, block 2 should be selected, which is what happens in other browsers.

**Environment**
- Slate Version: all versions
- Operating System: tested in macOS, Windows and Android
- Browser: Chrome 127+

**Context**
I've reported this bug to the Chromium issue tracker: https://issues.chromium.org/issues/389368412

The behaviour in Chrome 127 that causes this bug is that appending to the DOM an input element with a non-empty value causes a `selectionchange` event to be sent. However, the selection according to `document.getSelection()` remains the same as it was before the click. This confuses slate-react's selection change and click handling logic and causes the selection to revert incorrectly.

PR #5795, opened 2025-01-16
Ignore `selectionchange` events originating from `<input>` and `<textarea>` elements

12joan:
**Description**
In Chrome, `selectionchange` events can fire when `<input>` and `<textarea>` elements are appended to the DOM, causing `editor.selection` to be overwritten in some circumstances. This PR addresses this issue by ignoring `selectionchange` events originating from these elements.

**Issue**
Fixes: #5791

**Example**
Before:

https://github.com/user-attachments/assets/1e550014-0ece-4354-b768-13b23da8b93b

After:

https://github.com/user-attachments/assets/49113ad1-1046-4cfc-8b3f-94efb6ef3250

**Context**
Although Chrome's behaviour is apparently [compliant with the spec](https://issues.chromium.org/issues/389368412#comment9), other browsers do not behave this way, and the spec may at some point be changed to match other browsers' behaviour. See the [Chromium issue](https://issues.chromium.org/issues/389368412) for more information.

I'm fairly sure that ignoring `selectionchange` events from `<input>` and `<textarea>` elements will not cause other issues, since actual text selections in these elements do not affect `editor.selection`. However, if this PR causes a regression, feel free to revert it, since any bug it causes is likely to be more common than the one it fixes.

**Checks**

### DIFF
--- a/.changeset/empty-pillows-exercise.md
+++ b/.changeset/empty-pillows-exercise.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Ignore selectionchange events originating from input and textarea elements (addresses Chrome bug https://issues.chromium.org/issues/389368412)
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -856,15 +856,25 @@ export const Editable = forwardRef(
     useIsomorphicLayoutEffect(() => {
       const window = ReactEditor.getWindow(editor)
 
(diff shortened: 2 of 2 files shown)


dylans (changes requested):
@12joan could you please verify the one failing integration test (see the changes under file changed) isn't caused by this change? thanks!

12joan:
@dylans It looks like that test is just flaky, unrelated to this change. I ran it a few times locally on both this branch and main, and it seems to fail just as frequently on both, on roughly 7% of runs.

PR #5800, opened 2025-01-27
feat: implement findPath without relying on DOM

yf-yang:
**Description**
This PR aims at replacing `ReactEditor.findPath` with `Editor.findPath`

Currently, two weakmaps `NODE_TO_PARENT` and `NODE_TO_INDEX` are set when rendering the React component tree. These two weakmaps help compute a node's path.

However, if we are going to support virtualization, then not all components will be rendered, which means some nodes' information cannot be collected. Therefore, we'll not rely on React to compute the path of each element.

In this PR, we add multiple internal data structures to editor (`editor.internal`).

The first one is a map `descendantToParent`, which is a map from each non-editor node to its parent node, it performs the same functionality with current `NODE_TO_PARENT` weak map. This map should always be updated.

We patch operations that will affect element path, which are the same with `Path.operationCanTransformPath`: `insert_node`, `remove_node`, `merge_node`, `split_node`, `move_node`. When applying those operations, `descendantToParent` will be updated.

`descendantToParent` will not exist initially, that's because editor.initialValue could be set after it is created. Before each call to write/read `descendantToParent`, we'll check if it is initialized yet. If not, all the current editor descendants will be iterated to build initial store of `descendantToParent`.

Another two data structures are `pathCache` (map from descendant to path) and `parentChildIndexCache` (map from parent to another map, the map caches a descendant's index in its parent).

When calling `editor.findPath` to find path of a node `N`, we will first search `pathCache`. If it does not exist, we will find `N`'s parent, then find its index in its parent from the entry in `parentChildIndexCache`. If `N` does not exist in the cache entry, then we'll do the following:
1. Get the cache entry's size `S`, parent's first S children's indices are already cached
2. Starting from No.`S` child of parent's children, record it's index in the cache entry until we find the node
3. Leave other children uncalculated, they will be calculated when we need to do so.
Once we get the index, we are able to build the whole path by iterate all of its ancestors.

Those two caches should be cleared when they become invalid. Specifically, when those operations are applied, the whole pathCache would be cleared, dirty parent's `parentChildIndexCache` will be cleared.

**Checks**


packages/slate/src/editor/find-path.ts
```diff
@@ -0,0 +1,58 @@
...
+  const nodeToParent = getNodeToParent(editor)
+
+  const parent = nodeToParent.get(node)
+  if (!parent) {
+    throw new Error(
+      `Unable to find the path for Slate node (parent not found): ${Scrubber.stringify(
+        node
+      )}`
+    )
+  }
```
12joan, line 20: This could occur during normal use if `editor.children` is modified in a way that bypasses `normaliseNode`, such as when setting `editor.children` directly or when finding the path of a newly added node inside `withoutNormalizing`. I'd suggest traversing the full tree to repair the cache in this case, and only throw an error if the parent still cannot be found. In addition, the `withoutNormalizing` case can be mitigated by updating the cache in `editor.apply` instead of `normalizeNode`.

qigemingzizenmezhemekunnan:
Can slate be virtualized?

Issue #5292, opened 2023-02-10
`cmd+backspace` not always deleting the full line


francisashley:
**Description**
When hitting `cmd+backspace` to delete a line, occasionally a letter is missed. When this happens, hitting cmd+backspace again will not do anything. This behaviour can be seen in both Chrome and Firefox on macOS.

**Recording**

https://user-images.githubusercontent.com/12685308/218127272-deb423e8-b32c-4cdf-93b3-d79656b26af2.mov

**Steps to reproduce the issue**
1. Head to https://www.slatejs.org/examples/richtext
2. copy and paste the paragraph in the middle a couple of times
3. move the cursor to the end of the paragraph
4. keep hitting `cmd+backspace`
5. one of the lines will not be fully removed with a trailing character left behind

**Expectation**
`cmd+backspace` deletes the entire line.

**Environment**
- Slate Version: 0.90
- Operating System: macOS Monterey
- Browser: Chrome, Firefox

francisashley:
Hi @yupeilin123, I have removed the codesandbox link from the ticket as it is running on an old version.

Both the demo in the slate docs (https://www.slatejs.org/examples/richtext) and my own code are running on the latest 0.91.4 version and both have this issue.

PR #5827, opened 2025-03-19
Fix delete line being out by 1 char

louisfnz:
**Description**
Code was introduced [in this PR](https://github.com/ianstormtaylor/slate/pull/3364) that adds the ability to delete a line backwards using cmd + backspace. However, it is always short by one character. 

In standard paragraphs of text that span many lines without a break the issue is not immediately obvious as after deleting a line the character/word left over often jumps back to the available space on the previous line. We first noticed the issue in our Slate project in a code block element where it was more apparent.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/5292

**Example**
You can see in this video that unless it's the first line of an element (in this case the range is returned before hitting the problem algorithm), deleting a line always misses the first character. (Note that as there is often a single character remaining it jumps to the space in the previous line, but is clearly not deleted):

https://github.com/user-attachments/assets/75e3ece9-d570-4f96-aac1-5ceae42e6b77

Here's the fix branch with the first character getting deleted:

https://github.com/user-attachments/assets/594a5829-4b5b-4752-8eb5-d8335acbe043

**Context**
[This algorithm](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-dom/src/utils/lines.ts#L58-L74) always reduces the positions to within 1 character - `left` and `middle` are always the same value and `right` is always this value + 1. The returned range uses the position at the `right` index up to the parent boundary, but it should be using the `left` index.

**Checks**

### DIFF
--- a/packages/slate-dom/src/utils/lines.ts
+++ b/packages/slate-dom/src/utils/lines.ts
@@ -71,5 +71,5 @@ export const findCurrentLineRange = (
     middle = Math.floor((left + right) / 2)
   }
 
-  return Editor.range(editor, positions[right], parentRangeBoundary)
+  return Editor.range(editor, positions[left], parentRangeBoundary)
 }


dylans:
I was in a hurry and forgot to add a changeset. I need to do that before I do a release, otherwise this won't get included in the next release.

PR #5837, opened 2025-04-03
fix: add try/finally block in withoutSaving method to ensure state re…

ederzz:
…storation

**Description**
When i perform some operations and put them in the history stack, then perform some other operations without putting them into the history stack, and then trigger an undo action, this action might fail - for example, it can not find the previous element.The error can cause the restoration of the `SAVING` state to fail, which in turn leads to the failure of subsequent historyEditor operations to record history.

**Issue**
Fixes: (link to issue)

**Example**
image

**Context**
I think we should  catch the error from the callback of `withoutSaving`, make it won't affect the restoration of the saving state.

**Checks**

### DIFF
--- a/.changeset/rotten-poems-collect.md
+++ b/.changeset/rotten-poems-collect.md
@@ -1,0 +1,5 @@
+---
+'slate-history': patch
+---
+
+fix: add try/finally block in withoutSaving method to ensure state restoration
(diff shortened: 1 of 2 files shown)


dylans (changes requested):
Thanks for the PR, please add a changeset as it's a small change in behavior.

ederzz:
> Thanks for the PR, please add a changeset as it's a small change in behavior.

Fine, got it done!

dylans (approved):

PR #5850, opened 2025-04-28
Add `Leaf` with `start` and `end` props

zbeyens:
**Description**
Add optional `position` property to leaves returned by `Text.decorations()`. This makes it easier to track the exact position and boundary of each leaf within its parent text node, especially useful when you need to render elements relative to text nodes that have been split by decorations.

**Example**
Before:
```js
Text.decorations({ text: 'abc', mark: 'bold' }, [...])
// Returns:
[
  { text: 'a', mark: 'bold' },
  { text: 'b', mark: 'bold', decoration: 'highlight' },
  { text: 'c', mark: 'bold' }
]
```

After:
```js
Text.decorations({ text: 'abc', mark: 'bold' }, [...])
// Returns leaves with position info only when split by decorations:
[
  { 
    text: 'a', 
    mark: 'bold',
    position: { start: 0, end: 1, isFirst: true, isLast: false }
  },
  { 
    text: 'b', 
    mark: 'bold', 
    decoration: 'highlight',
    position: { start: 1, end: 2, isFirst: false, isLast: false }
  },
  { 
    text: 'c', 
    mark: 'bold',
    position: { start: 2, end: 3, isFirst: false, isLast: true }
  }
]

// Single leaf = no position info needed:
[{ text: 'abc', mark: 'bold' }]
```

**Context**
When working with decorated text nodes, it's often necessary to know the exact position of each leaf within the original text. A common use case is rendering an element exactly once after a text node, even when that text node has been split into multiple leaves by decorations.

For example, if you want to render a tooltip after a text node, using `renderLeaf` directly would cause the tooltip to render multiple times (once per leaf) because `renderLeaf` is called for each decorated segment. With the new `position` property, you can now reliably check:

```jsx
const renderLeaf = ({ leaf, text, ...props }) => {
  return (
    <>
      <span {...props.attributes}>{props.children}</span>
      {/* Render tooltip only once, after the last leaf of this text node */}
      {leaf.position?.isLast && <Tooltip />}
    </>
  )
}
```

The `position` property is only added when a text node is actually split by decorations, keeping the leaf objects clean and simple in the common case where no splitting occurs. Potential naming conflicts with user-defined `position` mark could be resolved by referring to the `Text` node instead.

### DIFF
--- a/.changeset/slate-react.md
+++ b/.changeset/slate-react.md
@@ -2,4 +2,4 @@
 'slate-react': minor
 ---
 
-- Update `RenderLeafProps` to use the new `Leaf` type from `slate`, that includes `start` and `end` properties representing the offset range within the original `Text` node.
+- Update `RenderLeafProps` to receive the `Leaf` type from `slate`, which may include an optional nested `position: { start, end, isFirst, isLast }` property if the text node was split by decorations. Useful to render something only in a single leaf per text node.
--- a/.changeset/slate-react.md
+++ b/.changeset/slate-react.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+- Update `RenderLeafProps` to use the new `Leaf` type from `slate`, that includes `start` and `end` properties representing the offset range within the original `Text` node.
--- a/.changeset/slate.md
+++ b/.changeset/slate.md
@@ -1,0 +1,6 @@
+---
+'slate': minor
+---
+
+- Update `Text.decorations` to return `Leaf[]` instead of `Text[]`.
+- Add new `Leaf` type, which includes `start` and `end` properties representing the offset range within the original `Text` node.
--- a/.changeset/slate.md
+++ b/.changeset/slate.md
@@ -2,5 +2,6 @@
 'slate': minor
 ---
 
-- Update `Text.decorations` to return `Leaf[]` instead of `Text[]`.
-- Add new `Leaf` type, which includes `start` and `end` properties representing the offset range within the original `Text` node.
+- Update `Text.decorations` to return `Leaf[]`.
+- Add `position?: { start, end, isFirst, isLast }` property to `Leaf` type.
+- The `position` property is only added if decorations cause the text node to be split into multiple leaves.
--- a/docs/api/nodes/text.md
+++ b/docs/api/nodes/text.md
@@ -26,9 +26,15 @@ If a `props.text` property is passed in, it will be ignored.
 
 If there are properties in `text` that are not in `props`, those will be ignored when it comes to testing for a match.
 
-#### `Text.decorations(node: Text, decorations: DecoratedRange[]) => Text[]`
+#### `Text.decorations(node: Text, decorations: DecoratedRange[]) => Leaf[]`
 
 Get the leaves for a text node, given `decorations`.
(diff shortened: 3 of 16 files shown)


12joan:
I wonder if it would be better to put this information in a separate property of `RenderLeafProps` from `leaf`? That would prevent unintuitive behaviour like `Node.extractProps` returning `position` for decorated text nodes when called from `renderLeaf`, and it would keep the interface of `leaf` consistent in all cases.

The property could be named to make it clearer that it refers to the position of the leaf within a decoration, rather than its position within the parent element, for example.

zbeyens:
Good point @12joan, I've refactored to use `leafPosition` instead. I've also added `renderText` that would be an alternative solution to our use-case.

packages/slate-react/src/components/text.tsx
```diff
@@ -65,17 +78,32 @@ const Text = (props: {
...
     </span>
   )
+
+  if (renderText) {
+    return renderText({
+      text,
+      children: textContent,
+      attributes: {
+        'data-slate-node': 'text',
+        ref: callbackRef,
```
12joan, line 94: We probably don't want to apply `data-slate-node="text"` to multiple different layers of the DOM hierarchy, and we certainly don't want to apply the ref twice. I would probably omit `attributes` from `renderText`, or possibly leave it as an empty object if we want to preserve the ability to add a distinct `ref` later without it being a breaking change. I can't think of any reason why that would be necessary, though.

12joan:
> Good point @12joan, I've refactored to use `leafPosition` instead. I've also added `renderText` that would be an alternative solution to our use-case.

Neat solution with `renderText`!

packages/slate-react/src/components/text.tsx
zbeyens: I took those attributes from the default behavior for backward compatibility.

Unless I'm missing something, there can't be nested renderText.

12joan: `textContent` also has those attributes, and is passed as `children` to `renderText`.

```tsx
  const textContent = (
    <span data-slate-node="text" ref={callbackRef}>
      {children}
    </span>
  )

  if (renderText) {
    return renderText({
      text,
      children: textContent,
      attributes: {
        'data-slate-node': 'text',
        ref: callbackRef,
      },
    })
  }
  ```

If `renderText` is implemented as `({ attributes, children }) => <span {...attributes}>{children}</span>`, then the resulting JSX is:

```tsx
<span data-slate-node="text" ref={refCallback}>
  <span data-slate-node="text" ref={refCallback}>
    {children}
  </span>
</span>
```

zbeyens: Good catch, I meant to use children instead of textContent there. Updating...

12joan: Got it. Since that's your intent, perhaps we could use the same pattern as `renderElement` to provide a default `renderText` that just renders a span with those attributes? That would eliminate the need for an if statement here.

https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/element.tsx#L44

zbeyens: Done

PR #5859, opened 2025-05-03
Optimize `isText`, `isElement`, `isNodeList` and `isEditor`

12joan:
**Description**
Removes the `isPlainObject` dependency and changes `isElement`, `isNodeList` and `isEditor` to perform a shallow check by default. Adds an option `{ deep: true }` to enable the old behaviour. These changes result in a roughly 70% performance improvement to the core `insertText` method when called in large documents.

**Context**
In documents containing tens of thousands of top-level blocks, I've found that the normalizations that run following a keystroke can take around 20ms, about half of which is spent in the `isPlainObject` check used by `isText` and `isElement`, specifically in the following check:

```ts
Object.prototype.toString.call(o) === '[object Object]'
```

In addition, the recursive behaviour of `isElement` and `isEditor` is probably unnecessary in most cases, so I've disabled it by default.

**Performance stack chart (before, 25ms)**
image

**Performance stack chart (after, 7ms)**
image

By itself, this change won't have a significant effect on the performance of `slate-react`, since 95% of the time spent per keystroke is spent by React rather than by core `slate`. However, I'm working on some potential optimisations to `slate-react` that will make optimising `slate` much more relevant, and this was a low-hanging fruit.

Related PR: #4983

**Checks**


.changeset/tough-pianos-train.md
```diff
@@ -0,0 +1,8 @@
+---
+'slate': patch
+'slate-history': patch
+'slate-hyperscript': patch
+'slate-react': patch
+---
+
+Optimize `isElement`, `isText`, `isNodeList` and `isEditor` by removing dependency on `is-plain-object` and by performing shallow checks by default. To perform a full check, including all descendants, pass the `{ deep: true }` option to `isElement`, `isNodeList` or `isEditor`.
```
dpolugic, line 8: Love to see it! 🎉 

As a side note, I noticed that [`is-plain-obj`](https://github.com/sindresorhus/is-plain-obj) made a similar optimization in https://github.com/sindresorhus/is-plain-obj/pull/16 by removing the `.toString()` check.

packages/slate/src/interfaces/operation.ts
```diff
@@ -210,21 +209,21 @@ export const Operation: OperationInterface = {
       case 'set_node':
         return (
           Path.isPath(value.path) &&
-          isPlainObject(value.properties) &&
-          isPlainObject(value.newProperties)
+          typeof value.properties === 'object' &&
+          typeof value.newProperties === 'object'
```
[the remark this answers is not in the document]
12joan: I don't think a null check is necessary here, but it wouldn't hurt, and you're right that a custom function would make the code easier to read and review. I'll do that. Thanks!

packages/slate/src/interfaces/element.ts
```diff
@@ -56,24 +65,46 @@ export interface ElementInterface {
...
  * Shared the function with isElementType utility
  */
-const isElement = (value: any): value is Element => {
-  return (
-    isPlainObject(value) &&
-    Node.isNodeList(value.children) &&
-    !Editor.isEditor(value)
-  )
+const isElement = (
+  value: any,
```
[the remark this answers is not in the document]
12joan: I think `any` is the right choice for `isElement` and similar functions. If we used `unknown`, then even with the object check TypeScript will complain when we try to access arbitrary properties on what it thinks of as the `object` type.

image

.changeset/tough-pianos-train.md
12joan: Interesting! I thought about using a more robust plain object check, but I saw that [someone in 2021](https://github.com/ianstormtaylor/slate/discussions/4665) encountered a problem when trying to use a non-plain object intentionally.

The only thing the plain object check might be useful for is as a defence in depth measure against prototype pollution, but that would only apply in very niche cases and only then if `isNodeList(value, { deep: true })` were used on the input value before consuming it.

packages/slate/src/interfaces/element.ts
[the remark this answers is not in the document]
dpolugic: Yeah, that part is a bit annoying. :-( It's possible to handle by checking for `'hello' in x`, but that can be a bit annoying.

If we have a custom `isPlainObject` function, we could potentially have it narrow the type to `Record<string, unknown>`, which should let us access arbitrary properties without errors.

Examples ([TS Playground](https://www.typescriptlang.org/play/?ts=5.8.0-pr-60718-25#code/CYUwxgNghgTiAEYD2A7AzgF3gDwFzwFcUBrFJAdxQCgqBLAM3gAoMBPABxCUe3gF4B8AORIARgCtwGIfABksnPACEglAQgQAlPADeVePAD0h+AFEYMJDHwAFS5xhthACxAakM4EhBp4ZLCDYtJjwqPBsnMJikmDSAHRMAEwAzMkAnJr6OHGu7lQAvjQMzBFcPPyCIhJSMvKKKnx+6hByCkK5EB7wtCg42noGxvAA8gDSADTwogRYUChkrD0A5uFI8OQwtBgI7W6dMj04Wdg5e0gFNPREsbRhwTbQPcPVsUx4hCRklJr4vMHwACVwFZgAAeTCbFBLSZEUgUFAAPl0WTgGAIMF6pW4igEjSqMWkrXqqmaFzojCY90eKGeBLemn6WSGY2OpzyhSAA)):
```
declare const x: unknown

if (typeof x === 'object' && x !== null) {
  // Error: Property 'hello' does not exist on type 'object'.(2339)
  x.hello
}

if (typeof x === 'object' && x !== null && 'hello' in x) {
  // No error, but annoying to write `'hello' in x`
  x.hello
}

function isPlainObject(x: unknown): x is Record<string, unknown> {
  return typeof x === 'object' && x !== null
}

if (isPlainObject(x)) {
  // No error
  x.hello
}
```

I think keeping the type as `any` is also fine though! ⭐ Just wanted to bring it up.

.changeset/tough-pianos-train.md
dpolugic: Yeah, that sounds reasonable! 👍

packages/slate/src/interfaces/element.ts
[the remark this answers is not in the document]
12joan: I'll keep it as `any` for now, but that's a nice solution to keep in mind in case we decide to change it later.

packages/slate-hyperscript/package.json
12joan, line 1: Since `slate-hyperscript` now depends on `slate`'s `isObject` export, the version of the `slate` peer dependency needs to be increased to the upcoming release. I'm not sure what version number this will be, though.

The same goes for `slate-history`.

dylans: I would assume 0.115.0 ?

12joan: @dylans It turned out to be 0.114.3 since I used a patch changeset. (Perhaps it should have been minor due to the new export, but I created the changeset before adding `isObject` and I forgot to update it.) Would you be able to commit the `peerDependencies` changes for `slate-hyperscript` and `slate-history` directly on the [Version Packages](https://github.com/ianstormtaylor/slate/pull/5860) PR?

PR #5861, opened 2025-05-07
fix: add missing `RenderTextProps` export

christianhg:
`renderText` was introduced in #5850, but the prop types didn't get exported with it.

### DIFF
--- a/.changeset/spicy-chefs-relate.md
+++ b/.changeset/spicy-chefs-relate.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+fix: add missing `RenderTextProps` export
--- a/packages/slate-react/src/index.ts
+++ b/packages/slate-react/src/index.ts
@@ -3,6 +3,7 @@ export {
   Editable,
   RenderElementProps,
   RenderLeafProps,
+  RenderTextProps,
   RenderPlaceholderProps,
   DefaultPlaceholder,
 } from './components/editable'


12joan:
@christianhg Thanks! Please could you add a changeset using `yarn changeset` so that merging this PR will release a new version of `slate-react`?

```md
---
"slate-react": patch
---

Add missing `RenderTextProps` export
```

christianhg:
> @christianhg Thanks! Please could you add a changeset using `yarn changeset` so that merging this PR will release a new version of `slate-react`?
> 
> ```
> ---
> "slate-react": patch
> ---
> 
> Add missing `RenderTextProps` export
> ```

Hi @12joan ! I _just_ did that 😄 

I wrote:

> fix: add missing `RenderTextProps` export

But perhaps the `fix: ` prefix is redundant in the changeset? It's been a while since I used these.

12joan:
> But perhaps the `fix: ` prefix is redundant in the changeset? It's been a while since I used these.

All good! It probably autofilled the changeset from your commit message. So long as a person reading the changelog can understand what the changes are, it doesn't matter too much how you phrase it.

skorenb:
seems all fine, but what is the problem here stands for in CI?
@christianhg will it go away if rebase your branch on the baseline and try again?

i just want to get rid of the workaround 😊
`type RenderTextProps = Parameters<Parameters<typeof Editable>['0']['renderText']>['0'];`

Issue #5814, opened 2025-03-06
`onValueChange` not triggered when selecting text from right to left then paste


gwy15:
**Description**
For slate react, `onValueChange` is not triggered when 
- select all content from **right to left**, then do a ctrl+v/cmd+v paste

**Recording**

![Image](https://github.com/user-attachments/assets/bebfaedd-19ea-474b-8c1d-2113845eb84a)

**Sandbox**
https://codesandbox.io/p/devbox/slate-reproductions-forked-jc6pq2?workspaceId=ws_Pbq92KWBCGm3cg9fGwjBFV

**Steps**
To reproduce the behavior:
1. Go to https://codesandbox.io/p/devbox/slate-reproductions-forked-jc6pq2?workspaceId=ws_Pbq92KWBCGm3cg9fGwjBFV
2. Copy some random text in your clipboard
3. Select all text in slate editor from **right to left**,
4. Do a cmd+v/ctrl+v paste
5. The onValueChange callback is not triggered!
 
**Expectation**
Since pasting value does change the value, onValueChange should be triggered.

As a matter of fact, 
- select text from left to right, then paste
- cmd+a select all, then paste
works fine, only select from right to left then paste does not work XD

**Environment**
- Slate Version: latest (0.112)
- Operating System: macOS
- Browser: Chrome

doonazif:
This also happens to me.

Seems like the, in [this `switch`](https://github.com/ianstormtaylor/slate/blob/7a8ab18c527c97aebcafe5b88032da38aa1664b0/packages/slate-react/src/components/slate.tsx#L65-L71), the `set_selection` is passed as the `options.operation.type` when typing after selecting from **right to left**.

I didn't check it thoroughly but it can easily be checked by:
* listening to both `onSelectionChange` and `onValueChange`
* select **right to left**, then type a character
* the `onSelectionChange` is called when typing the 1st character
* now select from **left to right**, then type a character
* the  `onValueChange` is called when typing the 1st character

I can use `onChange` instead of `onValueChange` but `onChange` is called profusely (selecting, moving cursor, typing, etc) and I tried to avoid that because of the many side-effects that my component will create when a text change.

PR #5863, opened 2025-05-08
fix issue #5814. onValueChange not triggered when selecting text from right to left then

virgil1996:
onValueChange not triggered when selecting text from right to left

✅ Closes: #5814

**Description**
when selecting text from right to left, it will trigger set_selection, remove_text, insert_text at the same time, cause onValueChange not trigger.

**Issue**
Fixes: [(link to issue)](https://github.com/ianstormtaylor/slate/issues/5814)

**Example**
image

### DIFF
--- a/packages/slate/src/core/apply.ts
+++ b/packages/slate/src/core/apply.ts
@@ -1,7 +1,7 @@
 import { PathRef } from '../interfaces/path-ref'
 import { PointRef } from '../interfaces/point-ref'
 import { RangeRef } from '../interfaces/range-ref'
-import { FLUSHING } from '../utils/weak-maps'
+import { FLUSHING, FLUSHING_SELECTION } from '../utils/weak-maps'
 import { Path } from '../interfaces/path'
 import { Transforms } from '../interfaces/transforms'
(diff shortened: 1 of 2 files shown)


dylans (changes requested):
@virgil1996 this looks reasonable. Could you please add a changeset as a patch and I'll land this.

virgil1996:
> @virgil1996 this looks reasonable. Could you please add a changeset as a patch and I'll land this.

done.

nabbydude:
With how the flushing flag is separated, this will trigger `onChange` twice in one frame, so any generic behavior will trigger twice. (the second will also be called while `editor.operations` is empty)

seems like strange behavior to have in core `slate`--since `onValueChange` and `onSelectionChange` are `slate-react` specific concepts.

I think it would make more sense if the `slate-react` `onChange` handler could see all the operations in a frame and decide to run one or both of the sub-events with one call.

zbeyens:
Agree with @nabbydude. Migrating from `operation` to `batch: Operation[]` (or `operations`) should enable the solution.

Issue #5838, opened 2025-04-05
Slate-vue3 has been released and all test cases have passed. Welcome to try it out


Guan-Erjia:
[Portal：](https://guan-erjia.github.io/slate-vue3)

dylans:
Closing this because it does not look like a current Slate bug.

This is third-party ecosystem announcement noise, not a Slate issue.

If there is a minimal repro showing current Slate behavior is at fault, a narrower follow-up can be opened.

PR #5865, opened 2025-05-09
feat(docs): add another member for slate (#5838)

Guan-Erjia:
Hi, I hope to add slate-vue3 to the documentation resources, which would be very useful for those who want to use slate in Vue projects. At present, the functionality of slate-vue3 has reached consistency with slate-react.

You can view demos in [slate-vue3](https://guan-erjia.github.io/slate-vue3/examples/rich-text)


dylans (approved):
Thank @Guan-Erjia !

PR #5871, opened 2025-05-15
Experimental chunking optimisation and other performance improvements

12joan:
> [!NOTE]
> Try it here: https://deploy-preview-5871--slatejs.netlify.app/examples/huge-document

## Summary

This PR makes various improvements to the performance of `slate-react`, reducing the latency when typing in very large documents (between 30,000 and ~100,000 blocks) by around 90% in Firefox and around 99.7% in Chrome.

It includes the following major changes:

- **Chunking** is the most significant and complex optimisation in this PR. Large children arrays are split into nested "chunks", each of which is rendered as a separate memoised React component and (optionally) a separate DOM element. This optimisation is disabled by default and can be enabled by overriding `editor.getChunkSize`. If enabled, chunking is also used to optimise setting the `NODE_TO_PARENT` and `NODE_TO_INDEX` weak maps for each child, since weak map operations (both getting and setting) can be extremely slow in large numbers.
  - Relevant files:
    - [`/packages/slate-react/src/chunking`](https://github.com/12joan/slate/tree/performance/chunking/packages/slate-react/src/chunking)
    - [`/packages/slate-react/src/hooks/use-children.tsx`](https://github.com/12joan/slate/blob/performance/chunking/packages/slate-react/src/hooks/use-children.tsx)
- **Immer** has been removed from most of `slate`. When applying operations, vanilla JS is used instead to replace modified nodes and their ancestors with new versions while keeping unmodified descendants and siblings unchanged.
  - Relevant files: mainly [`/packages/slate/src/interfaces/transforms/general.ts`](https://github.com/12joan/slate/blob/performance/chunking/packages/slate/src/interfaces/transforms/general.ts)
- **`useSlate`** no longer uses a context provider in the `Slate` component whose value is modified on each change, which was causing minor performance issues for unknown reasons. Instead, `useSlate` works in the same way as `useSlateSelector`, subscribing to `onChange` to re-render the component when Slate's value changes.
  - Relevant files:
    - [`/packages/slate-react/src/components/slate.tsx`](https://github.com/12joan/slate/blob/performance/chunking/packages/slate-react/src/components/slate.tsx)
    - [`/packages/slate-react/src/hooks/use-slate.tsx`](https://github.com/12joan/slate/blob/performance/chunking/packages/slate-react/src/hooks/use-slate.tsx)
- **`useSelected`** no longer uses a context. Instead, it uses `useSlateSelector` to compare `editor.selection` to the current element's range. This change was mainly because the old context-based approach is difficult to do efficiently alongside chunking due to the added complexity in `useChildren`.
  - Relevant files: [`/packages/slate-react/src/hooks/use-selected.ts`](https://github.com/12joan/slate/blob/performance/chunking/packages/slate-react/src/hooks/use-selected.ts)
- **`useElement`** has been added to get the current element. This is mainly to support the new `useSelected` solution.
  - Relevant files: [`/packages/slate-react/src/hooks/use-element.ts`](https://github.com/12joan/slate/blob/performance/chunking/packages/slate-react/src/hooks/use-element.ts)
- **Decorations** now work slightly differently:
  - The `decorate` function from `Editable` is now provided to nodes using a selector-like solution, so that if `decorate` is replaced, only nodes whose decorations are affected by the change will be re-rendered.
  - Decorations are no longer recomputed when a node's parent re-renders, only when the node itself re-renders or when the `decorate` function is changed. **This is a breaking change.**
  - As with `useSelected`, the main reason for this change is to work better with chunking.
  - Relevant files:
    - [`/packages/slate-react/src/hooks/use-decorations.ts`](https://github.com/12joan/slate/blob/performance/chunking/packages/slate-react/src/hooks/use-decorations.ts)
    - [`/packages/slate-dom/src/utils/range-list.ts`](https://github.com/12joan/slate/blob/performance/chunking/packages/slate-dom/src/utils/range-list.ts)
    - [`/packages/slate-react/src/hooks/use-children.tsx`](https://github.com/12joan/slate/blob/performance/chunking/packages/slate-react/src/hooks/use-children.tsx)
- **The Huge Document example** is now an interactive playground to explore the effects of various factors on Slate's performance, including changing the number of blocks, enabling/disabling chunking, changing the chunk size, and setting `content-visibility: auto` on each chunk or each element.

## Remaining tasks

- [x] Examine the effect of the `placeholder` prop on performance
- [x] Investigate Playwright test failures
- [x] Investigate flaky code highlighting Playwright test on Chromium
- [x] Check for IME regressions
- [x] [Check for regressions on Android](https://slate-js.slack.com/archives/C1RH7AXSS/p1747382533204319?thread_ts=1747379246.729299&cid=C1RH7AXSS)
- [x] Check for regressions on iOS
- [x] Check for regressions on Windows
- [x] Test the compatibility of these changes with existing Slate apps, such as Plate
  - [X] ~~Plate bug: `Cannot update a component while rendering a different component` when creating a new paragraph above a suggestion and clicking the suggestion.~~ Error on Plate's end
  - [X] ~~Plate bug: Code highlighting does not update when language is changed~~ Covered by the breaking decorations change
- [X] Clean up the implementation of `useSlate` (currently hacked together using `useSlateSelector`)
- [x] Do something about `useSlateWithV` - The `v` doesn't exist anymore, but we might want to polyfill it for compatibility?
- [x] Add changesets
- [x] Write a documentation page about performance, including instructions to enable chunking and other optimisations that must be enabled manually, such as `content-visibility: auto`
- [x] Update JS examples to match TS examples

## Performance comparison

https://github.com/user-attachments/assets/9ab92476-e184-479a-a7e5-117f6f13b4c8

The following table shows the average delay per keystroke when typing in a 50,000 block document. In Chrome and Safari, the durations include the time taken to paint the DOM after each keystroke. The durations in Firefox do not include painting time, but this is generally negligible in Firefox.

| Configuration                                                | Firefox (ms) | Chrome (ms)   | Safari (ms)                                                   |
| ------------------------------------------------------------ | ------- | -------- | -------------------------------------------------------- |
| https://slatejs.org                                          | 376   | 25,176 | 344<br>(initial render took several minutes)              |
| This branch with no optional features enabled                | 284   | 22,488 | 354<br>(initial render took several minutes)              |
| `content-visibility: auto` on each element                   | 219   | 485    | (initial render took about 30 minutes, too slow to test) |
| Chunking enabled                                             | 29    | 26,840 | 166<br>(relatively fast initial render)                   |
| Chunking enabled and `content-visibility: auto` on each chunk | 27    | 75     | 60<br>(relatively fast initial render)                    |

This data and my own observations indicate that:

- The non-optional optimisations on this branch provide negligible performance improvements on their own (although they are much more significant when combined with chunking and `content-visibility: auto`).
- The biggest optimisation in Chrome comes from enabling `content-visibility: auto`, especially if it is applied on each chunk instead of each element. This CSS rule prevents the browser from painting off-screen DOM nodes. However, enabling it on too many nodes causes additional overhead as the browser must determine whether each one is visible individually; this is why enabling it on each chunk (each of which contains 1000 Slate nodes) is more effective.
- In Safari, the editor is very slow to render without `content-visibility: auto`, but enabling it on all 50,000 nodes makes it completely unusable due to the overhead. Enabling it on each chunk is the only viable solution.
- In browsers that do not have painting issues, such as Firefox, enabling chunking reduces the delay per keystroke by approximately 90%.

## Chunking

Chunking is the technique of splitting a node's children into nested chunks. The size of each chunk does not matter much; I get almost identical results regardless of whether the chunk size is 10 or 1000. Chunking happens inside the `useChildren` hook and uses a complex algorithm to keep a mutable "chunk tree" object in sync with the `children` array (analogous to React's virtual DOM). The `editor.children` value is not affected.

Once assigned to a chunk, a Slate node is never reassigned to a different chunk, since this would result in remounting its React component. When blocks are added to the editor one by one, a greedy algorithm is used to assign them to chunks. Blocks that are already present in the editor (and large groups of blocks inserted all at once) are chunked using a more optimal algorithm since the total number of nodes is known in advance.

Each chunk is rendered as a separate memoised React component. This significantly reduces the amount of JSX that React must process on each change, which I believe is the main reason why the chunking optimisation is effective.

By default, chunking does not affect the DOM so that users can enable chunking for their editors without breaking their HTML structure. Optionally, users can pass a `renderChunk` prop to `Editable` that renders a DOM node for each chunk, on which styles such as `content-visibility: auto` can be applied. Most users will need to customise their editor's CSS to account for the nested DOM structure; I recommend setting the chunk size to a small number such as 3 during development so that any issues can be identified and resolved quickly.

## Feedback welcome

If you have any questions or feedback about the design or implementation of these changes, please let me know.


lagupa:
@12joan I haven’t had the time to dive deeper yet, but I just wanted to quickly thank you and commend your effort. The video is excellent — it explains the performance improvement very clearly and effectively. Much appreciated!

.github/workflows/ci.yml
```diff
@@ -34,3 +34,11 @@ jobs:
...
         env:
           CI: true
+
+      - name: Upload Playwright test results
+        if: ${{ !cancelled() && matrix.command == 'test:integration' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: test-results
+          path: test-results
+          retention-days: 30
```
12joan, line 44: **Explanation:** This uploads the `test-results/` directory produced by Playwright, which includes `trace.zip` files for the first failure of each test. These can be uploaded to https://trace.playwright.dev/ to see exactly what failed.

docs/libraries/slate-react/hooks.md
```diff
@@ -1,19 +1,19 @@
 # Slate React Hooks
 
-- [Check hooks](hooks.md#check-hooks)
-- [Editor hooks](hooks.md#editor-hooks)
-- [Selection hooks](hooks.md#selection-hooks)
```
12joan, line 5: **Explanation:** I'm not sure these hook categories are useful. `useElement` doesn't really fit into any of them, and `useSlateSelector` was miscategorised as a selection hook.

packages/slate-react/src/components/editable.tsx
```diff
@@ -999,10 +1020,12 @@ export const Editable = forwardRef(
       })
     })
 
+    useFlushDeferredSelectorsOnRender()
```
12joan, line 1023: **Explanation:** Selectors created using `useSlateSelector` may now be "deferred", which means they do not run until after React has re-rendered the `Editable`. This is used by `useSelected` to get the element's updated path.

The `useFlushDeferredSelectorsOnRender` hook is what tells these deferred selectors when to run.

packages/slate-react/src/hooks/use-generic-selector.tsx
```diff
@@ -0,0 +1,92 @@
...
+ *   return addEventListener(update)
+ * }, [addEventListener, update])
+ *
+ * return state
+ */
+
+export function useGenericSelector<T>(
+  selector: () => T,
+  equalityFn: (a: T | null, b: T) => boolean
+): [state: T, update: () => void] {
```
12joan, line 25: **Explanation:** This logic was previously part of `useSlateSelector`. I've refactored it out since I needed to use very similar logic in `useDecorations` to recompute decorations without re-rendering unless the decorations change.

packages/slate-react/src/hooks/use-slate.tsx
```diff
@@ -1,47 +1,61 @@
...
   }
 
-  return context
+  return v
+}
+
+/**
+ * Get the current editor object and its version, which increments on every
+ * change.
+ */
```
12joan, line 55: Should we mark this as `@deprecated` using JSDoc, which will show a warning in some IDEs when using it?

packages/slate-react/src/plugin/react-editor.ts
```diff
@@ -1,10 +1,18 @@
...
  */
 
-export interface ReactEditor extends DOMEditor {}
+export interface ReactEditor extends DOMEditor {
+  /**
+   * Determines the chunk size used by the children chunking optimization. If
+   * null is returned (which is the default), the chunking optimization is
+   * disabled.
+   */
+  getChunkSize: (node: Ancestor) => number | null
```
12joan, line 14: **Explanation:** Since the chunking optimisation is only applicable to `slate-react`, I've put the relevant method here instead of on `DOMEditor`.

packages/slate/src/interfaces/range.ts
```diff
@@ -1,4 +1,3 @@
-import { produce } from 'immer'
```
12joan, line 1: **Explanation:** Same situation as `point.ts`. Aside from the indentation change, the only real difference is:
```diff
-    r.anchor = anchor
-    r.focus = focus
+    return { anchor, focus }
```

playwright.config.ts
```diff
@@ -76,8 +76,8 @@ const config: PlaywrightTestConfig = {
     /* Base URL to use in actions like `await page.goto('/')`. */
     // baseURL: 'http://localhost:3000',
 
-    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: 'on-first-retry',
+    /* Collect trace if the first attempt fails. See https://playwright.dev/docs/trace-viewer */
+    trace: 'retain-on-first-failure',
```
12joan, line 80: **Explanation:** This lets us get proper traces for flaky tests, not just tests that fail consistently.

dylans (commented):
This looks really nice and thanks for the detailed explanations. I need at least a few days to digest it and read through it, and to give anyone else interested time to give feedback, but it looks like a huge win and a good way to simplify some things. We'll need to socialize the breaking changes.

dylans (approved):
ok, I've spent a lot of time reading through it... and I think now that I landed 0.115.0, we just need to land it and see what happens.

Do we need to increase the minimum in the docs to 0.116.0 since that will be the version number?

12joan:
@dylans Thanks for the review! Now seems like a good time to merge it. If there are any problems, people can downgrade to 0.115.x, and I can spend some time this week getting any problems resolved.

> Do we need to increase the minimum in the docs to 0.116.0 since that will be the version number?

Yep, I've updated it in `slate-react`'s changeset and `package.json` file.

Issue #4189, opened 2021-04-11
Tests that should do deep compare do shallow compare


beorn:
Adding an issue to accompany PR #4188.

A few tests do a `export const output = input`, and then in the test compares that input is equal to output, which will always be the case.  These tests need to do a deepCopy instead of just assigning to the same object reference, e.g., `export const output = cloneDeep(input)`.

PR #4188 uses lodash to do this, and reveals that 4 tests that previously passed now fails:

- slate-history / undo / delete_backward / block-text
- slate-history / undo / delete_backward / custom-prop
- slate-history / undo / delete_backward / inline-across
- slate / interfaces / Node / get / root

The PR doesn't attempt to fix these problems revealed by the tests (so we now have failing tests), it just fixes the tests themselves to do a deepClone.

PR #5872, opened 2025-05-15
fix: Closing skip logic for some unit tests in slate history section

Guan-Erjia:
**Description**
After turning off the skip logic, the unit test can pass
image

### DIFF
--- a/packages/slate-history/test/undo/delete_backward/custom-prop.tsx
+++ b/packages/slate-history/test/undo/delete_backward/custom-prop.tsx
@@ -18,5 +18,4 @@ export const input = (
     </block>
   </editor>
 )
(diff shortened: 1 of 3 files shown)


ahoisl:
This would I guess at least partly fix #4189 ?

Guan-Erjia:
> This would I guess at least partly fix #4189 ?
> 
> Have you also tried to remove the `skip` from: `slate/test/interfaces/Node/get/root.tsx` `slate/test/interfaces/Node/getIf/root.tsx` These skips are marked with the same link to #4188

I tried deleting these two skips, but an error occurred. It seems that the issues with these two unit tests were not fixed in previous versions.

PR #5875, opened 2025-05-18
Fix pasting exactly two blocks

### LINKED ISSUE #5775: Function `Transforms. insertFragment ` is not working properly when insert a fragment with two paragraphs
**Description**
When this function is called to insert a fragment with two paragraphs, the two paragraphs will be merged and inserted as inline nodes.

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/paste-html
2. Select the last two paragraphs and copy the content
image
3. Put the cursor in the **middle of one paragraph** (in this example it's before text`of doing exactly that`)
image
4. paste from clipboard
5. Easy to notice the copied two paragraphs are merged into one and inserted in the cursor
image

**Expectation**
It should create a new paragraph
image

**Environment**
- Slate Version: 0.110.2
- Operating System: macOS
- Browser: Chrome
- TypeScript Version: N/A

**Context**
The root cause of this issue is the incorrect behavior of `Transforms. insertFragment` when the fragment is two paragraphs.

Here is a new test case which can reproduce this issue, please create it in the repo to reproduce.

`packages/slate/test/transforms/insertFragment/of-blocks/block-middle-two-blocks.tsx`
```typescript
/** @jsx jsx */
import { Transforms } from 'slate'
import { jsx } from '../../..'

export const run = (editor, options = {}) => {
  Transforms.insertFragment(
    editor,
    <fragment>
      <block>one</block>
      <block>two</block>
    </fragment>,
    options
  )
}
export const input = (
  <editor>
    <block>
      wo
      <cursor />
      rd
    </block>
  </editor>
)
export const output = (
  <editor>
    <block>woone</block>
    <block>
      two
      <cursor />
      rd
    </block>
  </editor>
)
```

12joan:
**Description**
Although `insertFragment` works correctly when inserting one block or three or more blocks, it fails to split the destination block when inserting exactly two blocks. This PR fixes this by always splitting the block if the fragment contains text nodes in multiple distinct blocks.

I've also simplified the code slightly and added comments, since I had a very hard time understanding what the first part of `insertFragment` was doing and why.

**Issue**
Fixes #5775 

**Example**

https://github.com/user-attachments/assets/eeb7344b-f264-43a7-beb2-28a830697e97

**Checks**

### DIFF
--- a/packages/slate/test/transforms/insertFragment/of-lists/merge-lists.tsx
+++ b/packages/slate/test/transforms/insertFragment/of-lists/merge-lists.tsx
@@ -5,10 +5,12 @@ import { jsx } from '../../..'
 export const run = (editor, options = {}) => {
   Transforms.insertFragment(
     editor,
-    <block>
-      <block>3</block>
-      <block>4</block>
-    </block>,
+    <fragment>
+      <block>
+        <block>3</block>
+        <block>4</block>
+      </block>
+    </fragment>,
     options
   )
 }
@@ -33,4 +35,3 @@ export const output = (
     </block>
   </editor>
 )
-export const skip = true


packages/slate/src/transforms-text/insert-fragment.ts
```diff
@@ -77,66 +77,95 @@ export const insertFragment: TextTransforms['insertFragment'] = (
     const isBlockStart = Editor.isStart(editor, at, blockPath)
     const isBlockEnd = Editor.isEnd(editor, at, blockPath)
     const isBlockEmpty = isBlockStart && isBlockEnd
-    const mergeStart = !isBlockStart || (isBlockStart && isBlockEnd)
-    const mergeEnd = !isBlockEnd
```
12joan, line 81: I'm not sure why the logic for `mergeStart` was written like this when it can be simplified to `!isBlockStart || isBlockEnd`.

In addition, the `mergeStart` and `mergeEnd` variables are unused in the case where `isBlockEmpty` (i.e. `isBlockStart && isBlockEnd`) is true. See the following truth table:

| `isBlockStart` | `isblockEnd` | `isBlockEmpty` | `mergeStart`               | `mergeEnd`                  |
| -------------- | ------------ | -------------- | -------------------------- | --------------------------- |
| false          | false        | false          | true                       | true                        |
| false          | true         | false          | true                       | false                       |
| true           | false        | false          | false                      | true                        |
| true           | true         | true           | true (this case is unused) | false (this case is unused) |

As a result, we can simplify `mergeStart` to just `!isBlockStart` without affecting the functionality at all. Then we get:

```ts
const mergeStart = !isBlockStart
const mergeEnd = !isBlockEnd
```

I've removed these variables and included `!isBlockStart` and `!isBlockEnd` directly in the conditions where they were used, along with comments to explain more clearly what they're for.

dylans (approved):

PR #5878, opened 2025-05-19
Add way to prevent data loss in `normalizeNode`

ahoisl:
**Description**
normalizeNode ensures that all children of a node are either all blocks or inline/text nodes.
In #5768 the code was changed to unwrap blocks rather than removing them, thus keeping the data in the block.

This change now also allows to keep data in inline/text nodes by specifying a `wrapperElement` creation callback in the `normalizeNode` options.
If a inline/text node would need to be removed and `wrapperElement` is specified, the created block element is used to wrap the the element and thus ensure normalization.
The `normalizeNode` would be overridden in editors providing the `wrapperElement` function.

**Issue**
Fixes: no issue created yet

**Example**
See the changed unit tests. We override the editor `normalizeNode` and specify a `wrapperElement` callback that creates a block that is used to wrap inline/text nodes that would otherwise be removed.

```
const { normalizeNode } = editor
editor.normalizeNode = (entry, options) => {
    normalizeNode(entry, { ...options, wrapperElement: () => ({ /* empty block element */} ) })
}
```

**Context**
This change improves resiliency on invalid documents, e.g. if documents can be changed manually, uploaded by an API, another normalization function ill-behaves, or some other way exists to create invalid documents.
Without the change, on the first edit action in the editor, the document would be normalized and data is lost without an easy way for a user to cope with the behavior (you cannot copy & paste part of the editor because this would by default result in the same problem, the pasted part would be normalized and thrown away).

**Checks**


12joan (commented):
Good idea! I wonder if we can find more descriptive names than `wrapperElement` (and `withWrapping` in the test files). Perhaps `defaultElement` or `wrapInvalidInlines`?

ahoisl:
Thanks for the review @12joan !

> Good idea! I wonder if we can find more descriptive names than `wrapperElement` (and `withWrapping` in the test files). Perhaps `defaultElement` or `wrapInvalidInlines`?

Your are right, probably not very descriptive. The correct naming would probably be `createElementToWrapInvalidTextAndInlines` but that's a bit long 😄 
`fallbackElement` would be another suggestion.

> Or perhaps `defaultElement` should be a method on `editor`. That would make overriding it a little easier, let us use it in other places in the future if necessary, and avoid the problem of the `wrapperElement` option getting lost if some normalizers haven't been updated to pass `options` yet.

I'm fine with putting it directly into the Editor interface if we see potential other (more general?) usage. I just wanted to keep it to normalization for now, but it's true that the `options` argument is prone to being missed when overriding the `normalizeNode`.

12joan:
Good point

Issue #5031, opened 2022-06-23
Is Editor.above right when current selection is range?


jpuncle:
**Description**

Is Editor.above right when current selection is range

**Steps**

```javascript
export const input = (
  <editor>
    <block1>
      <block2>on<anchor>e <mark>tw</focus>o</mark></block2>
    </block1>
  </editor>
)

export const test = editor => {
  return Editor.above(editor, {
    match: n => Editor.isBlock(editor, n),
    mode: 'lowest',
  })
}

export const output = [<block1>...</block1>, [0]]
```

**Expectation**

```javascript
export const output = [<block2>...</block2>, [0, 0]]
```

PR #5880, opened 2025-05-21
Optimize editor#above and allow passing a location that doesnt exist as long as its parent exists

nabbydude:
**Description**
Currently calling `Editor#above` will error if the node that its input argument resolves to doesn't exist--even though it can never return that node and doesn't depend on it at all.
This fixes that and also reduces the amount of repeated/wasted checks in the implementation. (its probably faster now; I didn't check)

**Issue**
(just a personal issue 🙂)
In my usecase I want to filter/modify certain operations in `Editor#apply` depending on whether they are within a certain kind of block. A problem arises with `insert_node` (and `move_node`) operations: if I call `editor.above(operation.path)`, it errors because there is no node at `operation.path` yet. This seems silly because `editor.above` can never return the node at `operation.path`, the lowest node it can return is its parent. I've had to use a cumbersome workaround:
```ts
let entry = editor.parent(operation.path);
if (!matchesWhatIWant(node)) { // manually check direct parent
  entry = editor.above({ at: entry[1], match: n => matchesWhatIWant(n) });
}
```

**Example**
```xml
<editor>
  <block>
    <block>some text</block>
    <!-- path points here -->
  </block>
</editor>
```
```ts
editor.above({ at: [0, 1], match: n => Element.isElement(n) && editor.isBlock(n) })
// Old -> Error: Cannot find a descendant at path [0,1] in node: {...}
// New -> [<block><block>some text</block></block>, [0]]
```

**Context**
I took all checks out of the for-loop:
- `Text.isText(n)`: we now always pass an ancestor into the iterator. (`Path.equals(at.focus.path, at.anchor.path)` is used to test whether it would be text without having to look up the node!)
- `Path.isAncestor(p, at.anchor.path) && Path.isAncestor(p, at.focus.path)`: Unless I'm missing something, this was always true even in the old implementation.
- `!Path.equals(path, p)`: we use `Path.parent` to always start above `editor.path(at)`.

Also I added a test to prevent regressions of this behavior.

Alternatively (or additionally) to this change I could see an argument for some sort of method like `atOrAbove`--a version of above that *can* return its original input, or even just an option on `above`. This way made more sense for my specific usecase but I wouldnt be against it, my workaround above is essentially `editor.atOrAbove(Path.parent(operation.path))`.

If there are any changes I should make or if I should go in another direction like mentioned above just let me know 😄

**Checks**

### DIFF
--- a/packages/slate/src/editor/above.ts
+++ b/packages/slate/src/editor/above.ts
@@ -1,4 +1,5 @@
 import { Editor, EditorInterface } from '../interfaces/editor'
+import { Range } from '../interfaces'
 import { Path } from '../interfaces/path'
 
 export const above: EditorInterface['above'] = (editor, options = {}) => {
@@ -13,15 +14,18 @@ export const above: EditorInterface['above'] = (editor, options = {}) => {
     return
   }
 
-  const path = Editor.path(editor, at)
-  const reverse = mode === 'lowest'
+  let path = Editor.path(editor, at)
 
-  if (path.length === 0) {
-    return
+  // start with the direct ancestor -- unless its already the common ancestor of a cross-node range
+  if (!Range.isRange(at) || Path.equals(at.focus.path, at.anchor.path)) {
+    if (path.length === 0) return
+    path = Path.parent(path)
   }
 
+  const reverse = mode === 'lowest'
+
   for (const entry of Editor.levels(editor, {
-    at: Path.parent(path),
+    at: path,
     voids,
     match,
     reverse,
--- a/packages/slate/src/editor/above.ts
+++ b/packages/slate/src/editor/above.ts
@@ -25,6 +25,11 @@ export const above: EditorInterface['above'] = (editor, options = {}) => {
 
   const reverse = mode === 'lowest'
 
-  const [firstMatch] = Editor.levels(editor, { at: path, voids, match, reverse });
+  const [firstMatch] = Editor.levels(editor, {
+    at: path,
+    voids,
+    match,
+    reverse,
+  })
   return firstMatch // if nothing matches this returns undefined
 }
--- a/packages/slate/src/editor/above.ts
+++ b/packages/slate/src/editor/above.ts
@@ -25,12 +25,6 @@ export const above: EditorInterface['above'] = (editor, options = {}) => {
 
   const reverse = mode === 'lowest'
 
-  for (const entry of Editor.levels(editor, {
-    at: path,
-    voids,
-    match,
-    reverse,
-  })) {
-    return entry
-  }
+  const [firstMatch] = Editor.levels(editor, { at: path, voids, match, reverse });
+  return firstMatch // if nothing matches this returns undefined
 }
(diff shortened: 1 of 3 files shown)


nabbydude:
After some due diligence into related issues I think #5031 can be closed with or without merging this. Expected output there was the behavior before I got here--I know because I reintroduced same bug and failed [the `slate/interfaces/Editor/above/range` test](https://github.com/ianstormtaylor/slate/blob/7f5f9e1c626e61ec476087212b22ee9ff86538e4/packages/slate/test/interfaces/Editor/above/range.tsx) mid-development. 😅

packages/slate/src/editor/above.ts
```diff
@@ -15,27 +14,22 @@ export const above: EditorInterface['above'] = (editor, options = {}) => {
...
-      ) {
-        return [n, p]
-      }
-    } else {
-      if (!Path.equals(path, p)) {
-        return [n, p]
-      }
-    }
+    return entry
   }
```
12joan, line 34: Can this be replaced with `return Editor.levels(...)`?

packages/slate/src/editor/above.ts
```diff
@@ -15,27 +14,22 @@ export const above: EditorInterface['above'] = (editor, options = {}) => {
...
 
-  const path = Editor.path(editor, at)
+  let path = Editor.path(editor, at)
+
+  // start with the direct ancestor -- unless its already the common ancestor of a cross-node range
+  if (!Range.isRange(at) || Path.equals(at.focus.path, at.anchor.path)) {
+    if (path.length === 0) return
+    path = Path.parent(path)
+  }
+
```
12joan, line 24: I had to dig into `Editor.path` to figure out what was going on here. Perhaps if you add a comment like:

```ts
  // If `at` is a range that includes multiple paths, `Editor.path` already
  // returns the common ancestor of those paths. Otherwise, start with the
  // direct ancestor.
```

12joan (changes requested):
Nice! This makes `Editor.above` a lot simpler.

packages/slate/src/editor/above.ts
12joan: Oh, never mind. I forgot it's only returning the first match.

nabbydude: It could be `const [firstMatch] = Editor.levels(...); return firstMatch`

packages/slate/src/editor/above.ts
nabbydude: yeah I had to do the same to understand it lol. Updated the comment to be a bit more verbose

packages/slate/src/editor/above.ts
12joan: Good idea. That would make it a bit clearer what's going on.

12joan (approved):
Looks good!

packages/slate/src/editor/above.ts
nabbydude: done

PR #5881, opened 2025-05-23
Fix Firefox IME bug caused by placeholder

### LINKED ISSUE #5766: In Firefox ，When entering Japanese characters, the first letter will not be displayed after selection
**Description**

When inputting Japanese characters, the first letter does not display after selection. In Firefox, using the Japanese input method, inputting sinn and selecting up and down to 🆕 does not display the corresponding character on the page, and selecting other characters up and down also does not select
If you only see a drop-down option, you can press the left key and then the down key to display more options.

https://github.com/user-attachments/assets/78efe776-f088-4919-abf2-f0cb85c1979d

12joan:
**Description**
When composition starts and the placeholder is showing, Slate hides the placeholder. Due to a [bug in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1967203), this causes macOS's IME to receive outdated information about the start offset of the composition. This PR works around this issue by rendering the placeholder after the cursor instead of before.

**Issue**
Fixes #5766

**Example**

https://github.com/user-attachments/assets/41382b2f-6d0e-4f34-9727-c3f7daac64be

**Checks**

### DIFF
--- a/.changeset/mean-carpets-tap.md
+++ b/.changeset/mean-carpets-tap.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fix IME issues in Firefox caused by placeholder
--- a/packages/slate-react/src/components/leaf.tsx
+++ b/packages/slate-react/src/components/leaf.tsx
@@ -144,8 +144,8 @@ const Leaf = (props: {
 
     children = (
       <React.Fragment>
-        {renderPlaceholder(placeholderProps)}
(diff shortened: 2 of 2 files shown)


12joan:
The Firefox bug that made this workaround necessary will be resolved in Firefox 141. Noting this here so that anyone reading this in the future knows that this change is now safe to revert if it causes any unwanted side-effects.

PR #5882, opened 2025-05-24
FIX: Unexpected auto-scrolling behavior when refocus the editor (issue: 5826)

### LINKED ISSUE #5826: Unexpected auto-scrolling behavior when refocus the editor
**Description**
The editor has rendered a long-text, when i make it blured, and refocus the editor it occurs a auo-scroll behavior.It is unexpected.It seems the dom-selection hasn't change when enter the code logic: [code link](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/editable.tsx#L442)

**Recording**

Image

**Sandbox**
https://codesandbox.io/p/sandbox/n8tm84

**Steps**
To reproduce the behavior:
1. Go to sanbox
2. Click on top line of editor
3. Click outer, make editor blured.
4. Scroll down to the end line of editor
5. Click on editor, make it focused, then occurs auto-scroll

**Expectation**
I think the auto-scroll behavior shouldn't occur, because i have selected a new postion.

**Environment**
- Slate Version: latest
- Operating System: MacOs
- Browser: Chrome

**Context**
After reading the source code, i found that when the code runs into function [setDomSelection](https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/src/components/editable.tsx#L331), neither the Slate selection nor the DOM selection updates to where I clicked.

renasami:
**Description**
Fixed an issue where the editor would scroll to the previous focus position when rendering long text.

**Issue**
Fixes: [5826](https://github.com/ianstormtaylor/slate/issues/5826)

**Example**
Current behaviors

https://github.com/user-attachments/assets/fbd3ff52-8d9b-4df8-88f9-3405d193c3be

Fixed behaviors

https://github.com/user-attachments/assets/cdc50a69-6480-41a2-852e-f9d9b87ea31a

**Context**
It was missing scroll logic to the new caret position.

**Checks**


virgil1996:
I don’t think this is a safe fix.

Calling `el.focus({ preventScroll: true })` inside `onBlur` effectively cancels the blur. In practice, this makes the editor unable to lose focus, and when the user tries to focus another input element, focus is immediately stolen back by the editor.

That is a significant behavioral regression: clicking outside the editor, focusing toolbar inputs, dialogs, forms, or any other focusable control should be allowed to move focus away from the editor.

Issue #5886, opened 2025-05-31
Backspace doesn't work with combined characters in BEN


joblolin95:
**Description**
Some languages, such as Bengali, form combined characters by typing multiple characters together in sequence.

For example, `ক` (k on a Qwerty keyboard) and ` ি` (f on a Qwerty keyboard) combine to form character `কি`.

Backspacing the `কি` character removes the characters altogether, rather than just removing the ` ি` character to revert the character back to `ক`.

**Recording**
https://github.com/user-attachments/assets/be40c566-32a0-4e9c-9522-a53f4a043f72

**Sandbox**
https://www.slatejs.org/examples/richtext

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/richtext
2. Clear all contents from editor.
3. Install Bengali standard Qwerty keyboard on client.
4. Set Bengali Qwerty keyboard as the active keyboard.
3. Press k on Qwerty keyboard to input `ক` character.
4. Press f on Qwerty keyboard to input ` ি` character.
5. Backspace.
6. See that characters have been removed altogether.

**Environment**
- Slate Version: 0.114.2
- Operating System: Mac/Windows
- Browser: Chrome/Firefox

**Context**
While in Chrome, Firefox, or Microsoft Word, you can backspace the `কি` character and it does not remove it altogether - the text is reverted back to `কি`. This appears to be expected behavior for the Bengali language, and others like it.

12joan:
Would extending this logic to include languages other than Thai solve this problem?

https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/transforms-text/delete-text.ts#L172-L186

joblolin95:
> Would extending this logic to include languages other than Thai solve this problem?
> 
> https://github.com/ianstormtaylor/slate/blob/main/packages/slate/src/transforms-text/delete-text.ts#L172-L186

That's an excellent question and, considering the similarities between Thai and Bengali in relation to combined characters, I think that might solve this problem.

12joan:
Here it is: https://github.com/ianstormtaylor/slate/pull/5887

joblolin95:
> Sorry, I should have mentioned. You can test the PR using the automated Netlify build here: https://deploy-preview-5887--slatejs.netlify.app/examples/richtext

Ran a few tests with Bengali in this build, and it appears to now be functioning as desired.

Thank you for hard work on this!

12joan:
No problem.

[...]

PR #5890, opened 2025-06-04
Fix backspace for more grapheme clusters

joblolin95:
**Description**
This PR further resolves the problem detailed in issue #5886.

That issue records behavior wherein the user might create a grapheme cluster using multiple characters, but then when backspacing the entire grapheme cluster would be removed rather than just the last character that was input.

PR #5887 resolved this issue for Bengali. This PR resolves this for a wider set of languages where this would also be problematic.

The changes in this PR are specifically for the following languages:
- Burmese
- Hindi
- Khmer
- Malayalam
- Oriya
- Punjabi
- Tamil
- Telugu

**Issue**
Fixes: #5886.

**Example**
https://github.com/user-attachments/assets/255d7056-8d78-4b8a-afcf-a6c68ec96f46

**Checks**

### DIFF
--- a/packages/slate/src/transforms-text/delete-text.ts
+++ b/packages/slate/src/transforms-text/delete-text.ts
@@ -169,17 +169,25 @@ export const deleteText: TextTransforms['delete'] = (editor, options = {}) => {
       })
     }
 
-    // For certain scripts, deleting N character(s) backward should delete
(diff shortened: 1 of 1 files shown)


dylans (approved):
Thanks @joblolin95!

PR #5902, opened 2025-06-13
Fix autoscroll to top

zbeyens:
See changeset


kpkonghk01:
> `toDOMRange` should be patched once a fix is found

The same scroll to top problem occurs when the cursor is moving to the position of `\n` by using arrow keys. Is this behavior related to the same issue? Should I create another issue ticket for it?

zbeyens:
Could you reproduce it in https://platejs.org/ ?

PR #5908, opened 2025-06-19
Fix android linebreak deletion bug

### LINKED ISSUE #5730: Android delete forward at end of block causes selection to jump
**Description**
When you are at the end of a block and hit the `DEL` key to trigger a `deleteForward`, the cursor jumps to the next block, deletes a character, then jumps back to the end of the block. it deletes the first character on the next block but leaves that block intact; it takes another key press on the `delete` key to merge the blocks.

**Recording**
Android Delete Forward

**Sandbox**
https://www.slatejs.org/examples/richtext

**Steps**
To reproduce the behavior:
1. On an Android device
2. Go to `https://www.slatejs.org/examples/richtext`
3. Tap on the end of any block, that has a following block with content.
4. Press the `DEL` key once
5. Notice the first character of the next block is deleted, but they aren't merged

**Expectation**
The first character of the following block should be deleted and the two blocks should be merged into one.

**Environment**
- Slate Version: 0.110.1
- Operating System: Android
- Browser: Chrome

nabbydude:
**Description**
On android (or emulating it with devtools), when deleting a linebreak, the android-input-manager places the selection just after the break before deleting. This results in unexpected behavior when deleting forward, effectively skipping the linebreak and deleting the first character of the following line. Changed this behavior to check for that and place the selection just before instead.

**Issue**
Fixes #5730

**Example**
https://github.com/user-attachments/assets/248174a7-d434-4b4d-a18c-f715fd11da84

**Context**
Originally I wanted to delete this whole snippet. The browser sends a target range which starts at the end of the first line and ends at the start of the next--spanning only the break. This functions just fine in all cases...
*Except* when the previous line is empty. In that case the end and the start of the first block is the same, and if a range is from the start of one block to the start of the next its a hanging range, and gets unhung in the `Editor#deleteText` function--which takes the break we want to delete out of the range.

Fighting with magic behavior like this doesn't feel very great but fixing overarching issues with hanging ranges is out of my scope (for now) so I just fixed the existing behavior to not assume a backward delete.

**Checks**
  - [x] (also tested with `/android-tests` endpoint)


packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
```diff
@@ -396,25 +396,32 @@ export function createAndroidInputManager({
     let canStoreDiff = true
 
     if (type.startsWith('delete')) {
-      if (Range.isExpanded(targetRange)) {
-        const [start, end] = Range.edges(targetRange)
-        const leaf = Node.leaf(editor, start.path)
+      const direction = type.endsWith('Backward') ? 'backward' : 'forward'
+      let [start, end] = Range.edges(targetRange)
+      let [leaf, path] = Editor.leaf(editor, start.path)
```
nabbydude, line 401: Just moved these up instead of having separate definitions inside the branch and after it. We just make sure to update them if we make any changes

dylans (commented):
Thanks @nabbydude . We'll try this, but we may need to revert, as it's possible it could break IME behavior.

Issue #5912, opened 2025-07-04
How can we use this library on browsers that don't support composition events?


LLwill:
**Description**
composition event no support for windows pc in wechat browser.
In Wechat Browser open https://www.slatejs.org/examples/inlines, I input Chinese characters, but I can't delete them. The placeholder doesn't disappear. It feels like I didn't input them.

**Recording**

![Image](https://github.com/user-attachments/assets/bd70f1c0-b295-4092-8b8e-52b01b450ea0)

**Environment**
- Slate Version: 0.117.0
- Operating System: windows
- Browser: Wechat Browser

PR #5913, opened 2025-07-05
fix: Determine the higher version of WeChat browser for Windows

LLwill:
**Description**
Now the latest WeChat browser for Windows has used a newer version of Chrome

**Issue**
Fixes: [https://github.com/ianstormtaylor/slate/issues/5912](url)

### DIFF
--- a/packages/slate-dom/src/utils/environment.ts
+++ b/packages/slate-dom/src/utils/environment.ts
@@ -52,8 +52,8 @@ export const IS_UC_MOBILE =
 export const IS_WECHATBROWSER =
   typeof navigator !== 'undefined' &&
   /.*Wechat/.test(navigator.userAgent) &&
-  !/.*MacWechat/.test(navigator.userAgent) // avoid lookbehind (buggy in safari < 16.4)
-
+  !/.*MacWechat/.test(navigator.userAgent) && // avoid lookbehind (buggy in safari < 16.4)
+  (IS_CHROME ? IS_CHROME_LEGACY : true) // wechat and low chrome is real wechat
 // Check if DOM is available as React does internally.
 // https://github.com/facebook/react/blob/master/packages/shared/ExecutionEnvironment.js
 export const CAN_USE_DOM = !!(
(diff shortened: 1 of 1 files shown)


packages/slate-dom/src/utils/environment.ts
```diff
@@ -52,8 +52,8 @@ export const IS_UC_MOBILE =
 export const IS_WECHATBROWSER =
   typeof navigator !== 'undefined' &&
   /.*Wechat/.test(navigator.userAgent) &&
-  !/.*MacWechat/.test(navigator.userAgent) // avoid lookbehind (buggy in safari < 16.4)
-
+  !/.*MacWechat/.test(navigator.userAgent) && // avoid lookbehind (buggy in safari < 16.4)
+  (IS_CHROME ? IS_CHROME_LEGACY : true) // wechat and low chrome is real wechat
```
12joan: This is equivalent to `!IS_CHROME || IS_CHROME_LEGACY`. If that's what you intended, can we replace the condition with this?

Is it possible for the `!IS_CHROME` case to arise if the user agent contains `Wechat` (i.e. are there WeChat versions that don't mention `Chrome` in their user agents)? If not, we can simplify to `IS_CHROME_LEGACY`.

You said in the PR description:

>  Now the latest WeChat browser for Windows has used a newer version of Chrome

So it sounds like `IS_WECHATBROWSER` shouldn't be dependent on `IS_CHROME_LEGACY`. Do we actually want to rename this constant `IS_WECHAT_LEGACY`?

LLwill: The main concern is that the old version still uses the lower version of the Chrome kernel, so I added a bottom-line judgment. In fact, it is not necessary for the latest version at present.
Now UA:

> Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36 NetType/WIFI MicroMessenger/7.0.20.1781(0x6700143B) WindowsWechat(0x63090c33) XWEB/13639 Flue

12joan: That makes sense. It sounds like the best thing to do would be to rename the constant to `IS_WECHAT_LEGACY` to clarify that special behaviour is only needed on the old version, and to simplify the boolean condition.

If the old version's user agent always contains the string `Chrome`, then `&& IS_CHROME_LEGACY` should be sufficient; no need to check `IS_CHROME`.

Otherwise, I would recommend `&& (!IS_CHROME || IS_CHROME_LEGACY)`.

LLwill: You are right, this looks more elegant, I will modify it.

dylans (approved):

12joan:
@dylans This was merged without a changeset. One will need to be added before #5917 is merged to ensure `slate-dom` gets released.

Issue #5883, opened 2025-05-27
Composition interrupted in empty text nodes on Android IME


12joan:
**Description**
When using a Japanese IME on Android (I'm using Gboard), composition ends early when typing in empty leaves (including empty text nodes and leaves created by `editor.marks`).

**Recording**

https://github.com/user-attachments/assets/684322d9-aabc-4e7f-aa21-5583333a4783

**Steps**
To reproduce the behavior:
1. Enable the Japanese QWERTY keyboard
2. Go to https://www.slatejs.org/examples/custom-placeholder or any other empty Slate editor (the placeholder isn't required)
3. Press the "h" key - the underline under "h" disappears after a few milliseconds, indicating composition has ended
5. Press the "a" key
6. The editor now contains "hあ"

**Expectation**
Composition should not end, and the editor should contain "は" (the result of composing "h" and "a").

**Environment**
- `slate-react`: 0.114.2
- Operating System: Android (reproducible in all versions, tested up to 15.0)
- Browser: Chrome
- Keyboard: Gboard 14.2

**Context**
Potentially related: #4400

~~This problem happens because the `restoreDOM` method of `RestoreDOMManager` removes the DOM text node the user is currently typing in, replacing it with the `<br>` from the previous render. Disabling `RestoreDOM` fixes the problem, although the correct solution would probably involve checking some condition inside `restoreDOM` to determine whether to skip undoing the mutation.~~ I was mistaken. `RestoreDOM` does not seem to be the (sole) cause of this issue after all.

PR #5921, opened 2025-07-14
Fix IME composition being ignored in the first character in a line

bibixx:
**Description**
When using an IME, the first character's composition gets interrupted.

**Root Cause:**
**1. Immediate Selection Updates During Composition**
When insertCompositionText events occurred, the code would:
1. ✅ Correctly store the composition text as a pending diff with storeDiff()
2. ❌ Immediately call `scheduleAction()` to update selection
3. ❌ `scheduleAction()` would set `actionTimeoutId = setTimeout(flush)`
4. ❌ This caused pending diffs to be flushed almost immediately

**2. Selection-Triggered Flush Timeouts**
During IME composition, selection changes are common as the cursor position updates. The handleUserSelect function would:
1. Detect selection changes during composition
2. Schedule a 200ms flush timeout with `setTimeout(flush, FLUSH_DELAY)`
3. This would "auto-accept" the composition before the user could continue typing

**Issue**
Fixes: #5883

**Example**

**Context**

**Checks**

### DIFF
--- a/.changeset/shaggy-points-suffer.md
+++ b/.changeset/shaggy-points-suffer.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fix IME composition being ignored in the first character in a line
--- a/packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
+++ b/packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
@@ -688,6 +688,15 @@ export function createAndroidInputManager({
             insertPositionHint = false
           }
 
+          if (
+            type === 'insertCompositionText' ||
+            type === 'deleteCompositionText'
+          ) {
+            storeDiff(start.path, diff)
+            // Don't schedule selection updates during composition
+            return
+          }
+
           if (canStoreDiff) {
             const currentSelection = editor.selection
             storeDiff(start.path, diff)
@@ -761,7 +770,7 @@ export function createAndroidInputManager({
       insertPositionHint = false
     }
 
-    if (pathChanged || hasPendingDiffs()) {
+    if ((pathChanged || hasPendingDiffs()) && !IS_COMPOSING.get(editor)) {
       flushTimeoutId = setTimeout(flush, FLUSH_DELAY)
     }
   }


12joan:
Thanks for finding a solution to this one! I couldn't get to the bottom of it when I tried.

To make sure this doesn't cause any regressions, please could you go to `/examples/android-tests` and try out each of the test cases? Best to try it on a phone that uses composition for English (indicated by the underline while typing; recent versions of Gboard don't use it anymore).

12joan:
Actually, could you also add a new test case to `android-tests.tsx` explaining how others in the future can verify that this still works?

I've got plans to introduce a suite of automated Android tests using Appium and BrowserStack, although I haven't had time to work on that yet. In the meantime, adding to the suite of manual Android tests is the best way of making sure issues like this stay fixed.

bibixx:
> Thanks for finding a solution to this one! I couldn't get to the bottom of it when I tried.
> 
> To make sure this doesn't cause any regressions, please could you go to `/examples/android-tests` and try out each of the test cases? Best to try it on a phone that uses composition for English (indicated by the underline while typing; recent versions of Gboard don't use it anymore).

Yup verified all of those already on a physical device (OnePlus 10T 5G w/ Android 15).

> Actually, could you also add a new test case to `android-tests.tsx` explaining how others in the future can verify that this still works?
> 
> I've got plans to introduce a suite of automated Android tests using Appium and BrowserStack, although I haven't had time to work on that yet. In the meantime, adding to the suite of manual Android tests is the best way of making sure issues like this stay fixed.

Will do tomorrow 🫡 

---

Also one of our QAs found one more problem around this in the internal build. I haven't verified it yet though, so I need to check if it is ours of Slate's issue

12joan:
> I've got plans to introduce a suite of automated Android tests using Appium and BrowserStack

I've made some progrss with the automated Android tests, and I tried them out on the code in this PR.

For some reason, it doesn't fix the issue in the Android Studio emulator I'm using (Gboard 12.4.05.482060964, Android 14, Pixel 3a). Composition still ends after typing the first character.

In addition, it causes a regression. Before, the Slate value was updated immediately on every `compositionupdate`. Now, it's only updated when composition finishes.

12joan:
Actually, it looks like this does fix it, but only when the editor is non-empty.

12joan:
Here are the tests I'm running: https://github.com/12joan/slate-android-tests/blob/a43e787fe74ba962e7b538560afc73a8850710b0/appium/ime.spec.tsx

Before this PR, `types a simple sentence` and `updates the Slate value during composition` pass and the other two tests fail.

With the code in this PR, `composes correctly at the start of a block` now passes, `composes correctly at the start of the editor` continues to fail, and `updates the Slate value during composition` newly fails.

Issue #5932, opened 2025-08-09
`Cannot resolve a Slate point from DOM point` when dropping image on first position


delijah:
**Description**
We render a drop area on images. When this image is rendered on the first line of the editor and a user drops an image from the local file system on it, slate will throw an error: `Cannot resolve a Slate point from DOM point`. This happens only, if the image is rendered on the first line.

**Recording**
![Image](https://github.com/user-attachments/assets/987caf35-45f4-4911-8d81-c0dc9b6dbb7c)

**Sandbox**
https://codesandbox.io/p/sandbox/slate-image-drop-chv4xz

**Steps**
To reproduce the behavior:
1. Drag image from your filesystem
2. Drop on image in slate editor

**Expectation**
The image should be dropped and displayed. If you change the `initialValue` to have a paragraph rendered on the first line, and therefore the image will be rendered on the 2nd line, no error will happen.

**Environment**
- Slate Version: 0.118.0
- Operating System: iOS 12.6.1
- Browser: Chrome Version 138.0.7204.184 (Official Build) (arm64)

**Context**
I wonder if we structure our dom tree for the image wrong, or if this is a slate bug.

delijah:
I found, that if i set `searchDirection` to `'forward'` for this function call, it works fine: https://github.com/ianstormtaylor/slate/blob/7fdf77a56f9530e7b8eec55267649033abe40ff4/packages/slate-dom/src/plugin/dom-editor.ts#L1043

But it's hard to asses the consequences of such a change... 😓

PR #5936, opened 2025-08-14
fix: Search backward and forward for leaf nodes in non contenteditable elements

delijah:
**Description**
This PR change the `toSlatePoint` function in a way, that if the selection is within a `contentEditable={false}` node, it will search backward and forward for valid slate leafs. Before it was only looking backward by default.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/5932

**Example**
You can find a GIF and a codesandbox link in the linked issue. Here you can see how it behaves after the fix:
slate-drop

**Context**
I am aware, that this is a very sensible spot to make changes. Help appreciated!

**Checks**

### DIFF
--- a/.changeset/real-clouds-accept.md
+++ b/.changeset/real-clouds-accept.md
@@ -1,0 +1,5 @@
+---
+'slate-dom': patch
+---
+
+Search backward and forward for leaf nodes in non contenteditable elements inside `toSlatePoint`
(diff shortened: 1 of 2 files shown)


delijah:
How shall we proceed here @12joan ?

12joan:
> How shall we proceed here @12joan ?

Apologies if I gave you the wrong impression in the other threads where we've talked, but I'm not actually a Slate maintainer. I just comment on issues and PRs from time to time when I feel my input could be helpful. 😃 

@dylans is the maintainer of Slate, and will likely review this PR in due course.

dylans (approved):

delijah:
> Apologies if I gave you the wrong impression in the other threads where we've talked, but I'm not actually a Slate maintainer. I just comment on issues and PRs from time to time when I feel my input could be helpful. 😃
> 

I am sorry, there was probably some desperation involved, tagging you 🙈 Thx for the reply though @12joan and thx for proceeding @dylans !

PR #5937, opened 2025-08-15
fix: Fixed with pasting a void block in a paragraph middle path

kinaluo:
**Description**
Inserting a void block (such as an image) in the middle of a paragraph would incorrectly insert it before or after the paragraph. This change will cause it to be inserted at the cursor or selection position.

**Issue**
Fixes: (link to issue)

**Example**
befor:

https://github.com/user-attachments/assets/978db57c-1960-48d8-ba35-29e16755c64e

after:

https://github.com/user-attachments/assets/f8ce025f-ec71-4a54-847a-e0241ddd445e

**Context**

**Checks**

### DIFF
--- a/packages/slate/src/transforms-text/insert-fragment.ts
+++ b/packages/slate/src/transforms-text/insert-fragment.ts
@@ -180,19 +180,26 @@ export const insertFragment: TextTransforms['insertFragment'] = (
     const isInlineStart = Editor.isStart(editor, at, inlinePath)
     const isInlineEnd = Editor.isEnd(editor, at, inlinePath)
 
+    // When inserting block-level nodes into the middle, we still anchor to the
+    // current block path; after split Slate keeps the right half at the same
+    // path, so inserting at that path puts middles between halves.
+    const insertingIntoBlockMiddle = !isBlockStart && !isBlockEnd
     const middleRef = Editor.pathRef(
       editor,
       isBlockEnd && !ends.length ? Path.next(blockPath) : blockPath
     )
-
     const endRef = Editor.pathRef(
       editor,
       isInlineEnd ? Path.next(inlinePath) : inlinePath
     )
 
-    // If the fragment contains inlines in multiple distinct blocks, split the
-    // destination block.
-    const splitBlock = ends.length > 0
+    // Split the destination block if:
+    // - the fragment has trailing inlines to merge (original behavior), or
+    // - we're inserting block-level nodes into the middle of a block so they
+    //   can appear between the two halves of the destination block.
+    const shouldSplitForMiddleBlocks =
+      middles.length > 0 && insertingIntoBlockMiddle
+    const splitBlock = ends.length > 0 || shouldSplitForMiddleBlocks
 
     Transforms.splitNodes(editor, {
       at,
@@ -223,7 +230,9 @@ export const insertFragment: TextTransforms['insertFragment'] = (
       batchDirty,
     })
 
-    if (isBlockEmpty && !starts.length && middles.length && !ends.length) {
+    const shouldDeleteEmptyBlock =
+      isBlockEmpty && !starts.length && middles.length && !ends.length
+    if (shouldDeleteEmptyBlock) {
       Transforms.delete(editor, { at: blockPath, voids })
     }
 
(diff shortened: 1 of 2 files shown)


12joan:
My concern with this is it might make it harder to modify the behaviour of pasting of void blocks at locations where these blocks are not valid.

**List items example**
For example, in many list implementations, lists are represented using a UL > LI > LIC > text structure. The LIC (sometimes called list-item-text, or just a paragraph) acts as a block wrapper for text so that the LI can also have a nested list as its second child:

- UL
  - LI
    - LIC
      - [text]
    - OL
      - ...

Generally, no block elements other than LI, LIC, OL and UL are allowed anywhere in the list structure.

It looks like with this PR, pasting in the middle of the text of a LIC will cause that LIC to be split and the image inserted between the two LICs.

- LI
  - LIC
  - Image
  - LIC

A normalizer might clean this up by removing the image or relocating it to before or after the list, but this would still leave us with two LICs in the same LI, which isn't valid. Existing normalizers might handle the two-LICs case by splitting the LI.

- LI
  - LIC
- LI
  - LIC

But this doesn't make sense since nothing was actually pasted at the point where the LIs were split. Normalizers might instead handle the two-LICs case by re-merging the LICs, but this might require making additional code changes, such as when splitting a LIC intentionally using the enter key or when pasting text containing multiple lines.

**Conclusion**
I'm not opposed to us making this change, but we should make sure there's still an easy way of handling edge cases like this.

kinaluo:
> My concern with this is it might make it harder to modify the behaviour of pasting of void blocks at locations where these blocks are not valid.
> 
> **List items example** For example, in many list implementations, lists are represented using a UL > LI > LIC > text structure. The LIC (sometimes called list-item-text, or just a paragraph) acts as a block wrapper for text so that the LI can also have a nested list as its second child:
> 
> * UL
>   
>   * LI
>     
>     * LIC
>       
>       * [text]
>     * OL
>       
>       * ...
> 
> Generally, no block elements other than LI, LIC, OL and UL are allowed anywhere in the list structure.
> 
> It looks like with this PR, pasting in the middle of the text of a LIC will cause that LIC to be split and the image inserted between the two LICs.
> 
> * LI
>   
>   * LIC
>   * Image
>   * LIC
> 
> A normalizer might clean this up by removing the image or relocating it to before or after the list, but this would still leave us with two LICs in the same LI, which isn't valid. Existing normalizers might handle the two-LICs case by splitting the LI.
> 
> * LI
>   
>   * LIC
> * LI
>   
>   * LIC
> 
> But this doesn't make sense since nothing was actually pasted at the point where the LIs were split. Normalizers might instead handle the two-LICs case by re-merging the LICs, but this might require making additional code changes, such as when splitting a LIC intentionally using the enter key or when pasting text containing multiple lines.
> 
> **Conclusion** I'm not opposed to us making this change, but we should make sure there's still an easy way of handling edge cases like this.

Indeed, this behavior might occur, but I understand that in this method, we should focus solely on the behavior of block splitting and connection during content insertion, keeping his role as singular as possible. The normalizer process should be handled by the corresponding plugin implementation, or we could add a parameter in options to let users control whether to split paragraphs instead of accommodating the boundary behavior of different types of tags in this method. Or maybe I've misunderstood insertFragment. If you have other ideas for how to do this, I'd love to hear your suggestions.

12joan:
I suppose list plugins could override `insertFragment` and filter the tree of nodes about to be pasted into a list or other structured element. That would probably be the cleanest way of avoiding this problem.

kinaluo:
> I suppose list plugins could override `insertFragment` and filter the tree of nodes about to be pasted into a list or other structured element. That would probably be the cleanest way of avoiding this problem.

This is indeed a solution, but it requires the user to handle the corresponding plugins themselves. If this change can be merged, all we can do is indicate the impact of this change in the changelog

12joan:
I don't think there's any way of handling this in the core `insertFragment` itself. Slate doesn't have any API for determining in advance whether a particular void is valid in a particular location, such as an image inside a list item. Normalization logic can address this after the fact, but as I described above, normalization alone won't be enough in this case due to the split when inserting a void block.

One thing we could do is add options to `insertFragment` to make overriding it to handle this case easier. For example, it would be useful to have a `filter` option that controls whether each individual node is eligible for insertion. (Either that or we could introduce a `Node.filter` method on the Node interface.)

However, for this to be useful, the override code would also need a way of determining which exact block the nodes will be inserted into. The logic for determining this is non-trivial, since it involves [`getDefaultInsertLocation`](https://github.com/ianstormtaylor/slate/blob/896bc9c7c75e929e861e341fc4d7cc7395e8e82a/packages/slate/src/transforms-text/insert-fragment.ts#L18) (not currently exported), as well as [performing a deletion and checking the result of a point ref](https://github.com/ianstormtaylor/slate/blob/896bc9c7c75e929e861e341fc4d7cc7395e8e82a/packages/slate/src/transforms-text/insert-fragment.ts#L32-L40) if `at` is an expanded range.

This is quite a lot of logic for override code to duplicate, so maybe instead the `blockMatch` could be exposed as an argument on the `filter` option:

```ts
const { insertFragment } = editor

editor.insertFragment = (fragment, options) => {
  const { filter = () => true } = options
  const filteredNodes: Node[] = []

  insertFragment(fragment, {
    ...options,
    filter: (node, filterOptions) => {
      if (!filter(node, filterOptions)) return false

      // Prevent splitting a LIC to insert a void block
      if (
        filterOptions.blockEntry[0].type === 'lic' &&
        Element.isElement(node) &&
        Editor.isBlock(node) &&
        Editor.isVoid(node)
      ) {
        filteredNodes.push(node)
        return false
      }

      return true
    },
  })

  // Insert filteredNodes before or after the list
  // ...
}
```

I think it's reasonable to expect list plugins to do this, but as you say we should describe this in the changeset.

Issue #5938, opened 2025-08-21
`DOMEditor.findPath` returns no or wrong path


delijah:
**Description**
When using `DOMEditor.findPath` within the `onChange` event handler, it either throws an error, or returns a wrong path, after manipulating the document. The reason is, `NODE_TO_INDEX` and `NODE_TO_PARENT` only updates on render, but `onChange` is fired before the changes have been rendered.

**Sandbox**
https://codesandbox.io/p/sandbox/fyslvy

**Steps**
To reproduce the behavior:
1. Visit the codesandbox link
2. Click within text on the first line
3. Press enter
4. See error / wrong paths in console

**Expectation**
`DOMEditor.findPath` should return the correct path at any point in time.

**Environment**
- Slate Version: 0.118.0
- Operating System: macOS
- Browser: Chrome

**Context**
This became a problem while using `useEditorSelector` from plate.js. `useEditorSelector` is fired directly after `onChange`, if we want to retrieve a path of a node in there, it fails. Example:

```ts
import { type EditorSelection, type Path, PathApi } from 'platejs';
import { useEditorSelector } from 'platejs/react';

const isPathFocused = (selection: EditorSelection, path?: Path) =>
    Boolean(path && selection?.focus.path && PathApi.isCommon(path, selection.focus.path));

export const useEditorNodeFocused = (node: TNode, options?: Options) => {
    const focused = useEditorSelector((editor) => {
        const { selection } = editor;
        const path = DOMEditor.findPath(editor, node);

        return isPathFocused(selection, path);
    }, []);

    return focused;
};
```

12joan:
This is indeed an unusual quirk of how paths are computed. There have been efforts in the past to compute paths in the core `slate` package instead, but those efforts seem to have stalled. See #5800.

As a workaround, you can use Slate's `useSlateSelector` instead of `useEditorSelector` and pass `{ deferred: true }` as an option, which tells it to only run the selector after the editor has finished re-rendering, when the new path will be available and correct. As a side-effect of this, the result of the selector might be out of sync for one very brief render pass, but this probably isn't an issue for checking if a node is focused.

delijah:
Ok. At the moment we are using a custom `findPath` function. Of course, this is not an option for slate, there we would need an index, for better performance.

```ts
const findNodePath = (editor: PlateEditor, needle: TNode, options: Options = {}) => {
    const { depth = Infinity } = options;
    const needleKey = editor.api.findKey(needle);

[...]

PR #5939, opened 2025-08-21
fix: Sync `NODE_TO_INDEX` and `NODE_TO_PARENT` on `apply` instead on render

delijah:
**Description**
This PR will keep `NODE_TO_INDEX` and `NODE_TO_PARENT` in sync right after applying operations, instead of on render. This will make the `DOMEditor.findPath` function more reliable.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/5938

**Context**
Please have a look at the issue for more context.

**Checks**

### DIFF
--- a/packages/slate-dom/src/plugin/with-dom.ts
+++ b/packages/slate-dom/src/plugin/with-dom.ts
@@ -35,6 +35,8 @@ import {
   EDITOR_TO_USER_MARKS,
   EDITOR_TO_USER_SELECTION,
   NODE_TO_KEY,
+  NODE_TO_INDEX,
+  NODE_TO_PARENT,
 } from '../utils/weak-maps'
 import { DOMEditor } from './dom-editor'
 
@@ -222,6 +224,15 @@ export const withDOM = <T extends BaseEditor>(
 
     for (const [path, key] of matches) {
       const [node] = Editor.node(e, path)
+
+      NODE_TO_INDEX.set(node, path.length ? path[path.length - 1] : 0)
+
+      if (path.length) {
+        const [parent] = Editor.parent(e, path)
+
+        NODE_TO_PARENT.set(node, parent)
+      }
+
       NODE_TO_KEY.set(node, key)
     }
 
--- a/packages/slate-react/src/components/element.tsx
+++ b/packages/slate-react/src/components/element.tsx
@@ -9,8 +9,6 @@ import {
   EDITOR_TO_KEY_TO_ELEMENT,
   ELEMENT_TO_NODE,
   NODE_TO_ELEMENT,
-  NODE_TO_INDEX,
-  NODE_TO_PARENT,
 } from 'slate-dom'
 import {
   RenderChunkProps,
@@ -138,9 +136,6 @@ const Element = (props: {
         />
       </Tag>
     )
-
-    NODE_TO_INDEX.set(text, 0)
-    NODE_TO_PARENT.set(text, element)
   }
 
   return renderElement({ attributes, children, element })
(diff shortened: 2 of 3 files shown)


12joan:
The main problem with updating `NODE_TO_INDEX` and `NODE_TO_PARENT` on every `apply` is performance. Due to the way weak map keys are implemented in browsers, both reading from and writing to weak maps is very slow, and doing so on every operation is very likely to have a measurable impact on performance.

On of the main focuses of my [performance PR](https://github.com/ianstormtaylor/slate/pull/5871) was reducing the number of weak map operations, which is why `getChunkTreeForNode` is used when chunking is enabled to avoid setting the weak maps unnecessarily.

You could probably do the same thing at the `apply` level by figuring out which nodes' parents and indices are changed by each operation and updating the minimum number of weak map entries to account for this.

delijah:
Ah. I assumed inside `matches` we already have only the nodes that were affected by the operation?

12joan:
> Ah. I assumed inside `matches` we already have only the nodes that were affected by the operation?

`matches` doesn't exactly equal the set of nodes whose paths will change. For example, it a node is inserted or removed, all future sibling nodes' `NODE_TO_INDEX` must be updated. These sibling nodes are not currently included in `matches`.

Ideally, we would handle each operation type individually to determine whether `NODE_TO_INDEX` and/or `NODE_TO_PARENT` needs updating. This would also require very careful testing, since it can be unintuitive in some cases. For example, an `insert_text` operation, although it doesn't cause any node's path to change, still requires updating the weak maps for the affected leaf node and all of its ancestors, since these nodes' objects will be immutably replaced with fresh objects.

Regarding that last point, it may be a good idea if we used `NODE_KEY_TO_INDEX` and `NODE_KEY_TO_PARENT_KEY` instead, where the "key" is the one from `NODE_TO_KEY`. Keys are stable for the lifetime of a node, so the weak maps for computing paths would need updating much less frequently. Alternatively, we could simply store the entire path in a single weak map `NODE_TO_PATH`.

delijah:
> Regarding that last point, it may be a good idea if we used `NODE_KEY_TO_INDEX` and `NODE_KEY_TO_PARENT_KEY` instead, where the "key" is the one from `NODE_TO_KEY`. Keys are stable for the lifetime of a node, so the weak maps for computing paths would need updating much less frequently. Alternatively, we could simply store the entire path in a single weak map `NODE_TO_PATH`.

In the same time you wrote your answer, i've create a new commit. Were you talking about something like that?

Inpired by https://webreflection.medium.com/a-basic-weakmap-performance-hint-e33c14908dff

12joan:
Broadly, yes, although using `Key['id']` as a key would lead to a memory leak. It needs to be a weak map where the key is the Key object itself.

delijah:
Nice. Ok, added the changes. But yes, the complex part will be finding out which paths are affected and also testing this logic. Will not have time for now to take care about this 😢

delijah:
> For example, an `insert_text` operation, although it doesn't cause any node's path to change, still requires updating the weak maps for the affected leaf node and all of its ancestors, since these nodes' objects will be immutably replaced with fresh objects.

Not sure if i understand you correctly. What needs to be updated on `insert_text`? You're saying the node keys change for the affected leaf node and all of its ancestors? Therefore we have to set the paths for the new node keys?

12joan:
> > For example, an `insert_text` operation, although it doesn't cause any node's path to change, still requires updating the weak maps for the affected leaf node and all of its ancestors, since these nodes' objects will be immutably replaced with fresh objects.
> 
> Not sure if i understand you correctly. What needs to be updated on `insert_text`? You're saying the node keys change for the affected leaf node and all of its ancestors? Therefore we have to set the paths for the new node keys?

That part of my comment was assuming that the weak maps were keyed on the node object, not the node key object. Since node objects are completely replaced when their or their descendant's value changes, any weak maps keyed on the node object need to be updated. That isn't necessary when keying on the key object.

PR #5946, opened 2025-08-26
fix: keyboard accessibility for example site

hiteshshetty-dev:
**Description**
This PR fixes keyboard accessibility issues on example site.

**Example**
Before:
https://github.com/user-attachments/assets/535572b9-ab9c-49fd-bab4-e6b892b3dda7

After:
https://github.com/user-attachments/assets/dad10596-c2e2-41c9-a157-80a3ceacd248

**Context**
Earlier, Tab indexes was not working as expected for the menu and menu-items in example sites. Fix converts the menu dom from span to button and make sures that menu-items are rendered conditionally with proper aria tags. Additionally, adds keydown event on menu to open and close menu through "Enter" and "Escape" key respectively.

**Checks**


site/examples/ts/components/index.tsx
```diff
@@ -25,6 +25,8 @@ export const Button = React.forwardRef(
     <span
       {...props}
       ref={ref}
+      tabIndex={0}
+      role="button"
```
12joan: Adding these will make the button focusable and tell assistive technology that the element is a button, but since it's a `<span>`, keyboard users will be unable to activate it using the enter or space keys.

In addition, most places where this button is used (editor toolbar buttons) are using the unfortunate pattern of:
```ts
onMouseDown={event => {
  event.preventDefault()
  actualClickHandler()
}}
```
This means that even if the `<span>` were switched to a button, keyboard users would be unable to activate it. 

There's actually a good reason for preventing the mousedown event, since it prevents mouse users' focus from leaving the editor when they click the button. This has to happen on mousedown/pointerdown since it's too late to prevent focus loss by the time the click event is fired. However, it would be much better to use:
```ts
onPointerDown={event => event.preventDefault()}
onClick={actualClickHandler}
```
This achieves the same thing without breaking keyboard accessibility.

12joan: There are quite a few places where this needs to be changed, and then tested to make sure nothing's broken. I don't blame you if you don't want to tackle all of this right now, but in that case I would suggest removing the `tabIndex={0}` and `role="button"` so we're not giving readers the false impression that this is sufficient.

site/pages/examples/[example].tsx
```diff
@@ -148,6 +151,7 @@ const TabList = ({
       white-space: nowrap;
       max-height: 70vh;
       z-index: 3; /* To appear above the underlay */
+      visibility: ${isVisible ? 'visible' : 'hidden'};
```
Comment (12joan) [inline site/pages/examples/[example].tsx:156]: This causes the menu to close instantly rather than animating closed. I think it's fine to keep it this way, but I wanted to double-check if using `visibility: hidden` in addition to `width: 0` and `aria-hidden` is necessary. We might also want to add an `inert` attribute to prevent keyboard focus inside the menu while it's closed?

site/pages/examples/[example].tsx
```diff
@@ -171,10 +175,17 @@ const TabListUnderlay = ({
 )
 
 const TabButton = (props: React.HTMLAttributes<HTMLSpanElement>) => (
-  <span
+  <button
     {...props}
+    aria-label="Toggle examples menu"
+    aria-expanded={props['aria-expanded']}
```
Comment (12joan) [inline site/pages/examples/[example].tsx]: This is already covered by `{...props}`.

```suggestion
```

12joan (commented):
Nice work!

site/pages/examples/[example].tsx
Comment (hiteshshetty-dev) [reply site/pages/examples/[example].tsx]: Good catch, updated the code.

site/pages/examples/[example].tsx
Comment (hiteshshetty-dev) [reply site/pages/examples/[example].tsx:156]: Thanks for pointing this out! Initially, I tried returning `null` early based on the `isVisible` flag, but that removed the opening animation. To preserve the transition, I switched to using the `visibility` property together with `width`. In that case, `aria-hidden` becomes somewhat redundant and can likely be removed.

Comment (hiteshshetty-dev) [reply site/pages/examples/[example].tsx:156]: I see following possible approaches here:

1. Keep `visibility` and animate it:
`transition: width 0.2s, visibility 0.2s;`
2. Drop visibility and use inert: 
`inert` is now broadly supported (April 2023 onward). However, React 18 doesn’t have built-in support yet, so we’d need a small workaround to apply it:
`inert={isVisible ? undefined : ""}`
Additionally, an update to the TS declaration.

Comment (hiteshshetty-dev) [reply site/pages/examples/[example].tsx:156]: Personally I lean toward option 1 for simplicity, but I’m fine with whichever direction you prefer.

Comment (12joan) [reply site/pages/examples/[example].tsx:156]: ~~The `inert` solution sounds like a good idea to me. I didn't know React 18 didn't support it. Would `true` or `'true'` work as an easier to understand alternative to the empty string in the hidden case?~~

Comment (12joan) [reply site/pages/examples/[example].tsx:156]: Actually, I've just tried the `transition: width 0.2s, visibility 0.2s` solution and it works very well. I wasn't aware until I checked MDN just now that `visibility` adjusts the tab order too, so that removes the need for `inert`.

Comment (hiteshshetty-dev) [reply site/pages/examples/[example].tsx:154]: `visibility: hidden` already handles focus loss, so we might not need `inert`.

site/examples/ts/components/index.tsx
hiteshshetty-dev: Thanks for the detailed feedback!
I have implemented suggested changes:
- Button component: span → button
- Event handling: onPointerDown + onClick pattern
- Updated across all examples with proper TS types


https://github.com/user-attachments/assets/301a1cc9-eba2-4616-b79e-7e83cf3d3e0f

12joan (approved):
Awesome! ✨

PR #5948, opened 2025-08-28
test: fix `select.test.ts` after button component styling updates

hiteshshetty-dev:
**Description**
The `select.test.ts` was using a dynamically generated CSS class selector (`.css-1vdn1ty`) from Emotion CSS-in-JS library, which broke after recent button component updates in PR #5946 .

**Context**
- The button component uses `@emotion/css` which generates hash-based class names
- When the button styling was updated, the CSS class hash changed from `.css-1vdn1ty` to a new value
- The test continued to reference the old, non-existent class name

Updates:
- Updated `richtext.tsx` to add test-id for BlockButton.
- Updated `select.test.ts` to use `page.getByTestId()` instead of CSS class selector.

**Checks**
- [x] The relevant e2e test cases are passing.

**Apologies for missing this E2E update in the original PR. This ensures our test suite remains stable against future styling changes.**
Screenshot 2025-08-28 at 1 29 12 PM

### DIFF
--- a/playwright/integration/examples/select.test.ts
+++ b/playwright/integration/examples/select.test.ts
@@ -12,11 +12,7 @@ test.describe('selection', () => {
     for (let i = 0; i < 3; i++) {
       await page.locator(slateEditor).nth(1).click()
     }
-    await page.pause()
-
-    // .css-1vdn1ty is the gray, unselected button
-    expect(await page.locator('.css-1vdn1ty').nth(6).textContent()).toBe(
-      'format_quote'
-    )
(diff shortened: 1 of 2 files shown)


hiteshshetty-dev:
We currently have actionTimeout set to 0, which disables timeouts for actions. Combined with the use of locator.textContent() in tests, this caused our E2E suite to run indefinitely.
Should we update the timeout to 2 minutes instead?

hiteshshetty-dev:
These changes are already merged with #5950

PR #5953, opened 2025-09-06
Example site performance improvements and optimization

hiteshshetty-dev:
**Description**
This PR implements performance improvements and optimizations for the example site. The changes include dynamic imports for better bundle splitting, improved accessibility.

**Example**

**Before:**
- Single monolithic example page with all components loaded upfront
- Build size: 
Sample Page Sizes before

analyze/client before

- Lighthouse score: 
Lighthouse report before

**After:**

- Individual dynamic pages with code splitting
- Build size:
Sample Page Sizes after

analyze/client after

- Lighthouse source
Lighthouse report before

**Context**
The example site was experiencing performance issues due to loading all example components upfront on each route, which created large bundle sizes and slow initial page loads. Additionally, the older version of @faker-js/faker (v8.2.0) was not supporting tree-shaking, significantly increasing the build size.

**Checks**


site/pages/examples/android-tests.tsx
12joan, line 1: I think this could still be done using a single `[example].tsx` file, mapping the example name to the dynamic import path, page title and description, couldn't it?

hiteshshetty-dev: Thanks for the feedback! You're absolutely right that a single `[example].tsx` file could handle this with a mapping approach.
I initially implemented it [that](https://github.com/ianstormtaylor/slate/pull/5953/commits/aef162be4d5b4d35b6b6e3eb1d8648eda512d149) way, but then explored individual pages thinking it would provide better bundle optimization. However, after your comment, I realize the performance benefits might not be as significant as I initially thought, especially since we're already using dynamic imports.
I'd be happy to revert to the single `[example].tsx` approach - it would definitely be cleaner and easier to maintain. The main optimization we achieved (moving the common layout to `_app.tsx`) would still be preserved.

hiteshshetty-dev: Additional context: The benefit of moving the layout to _app.tsx is that it gets included in the initial HTML, and only the RTE examples would be loaded dynamically. This should provide a better user experience with faster initial page loads.

Before:
before moving layout to _app.tsx
After:
after moving layout to _app.tsx

12joan: Nice! I think keeping the `[example].tsx` would be better. Since these are just examples, code simplicity is more important than performance, and having a single pair of TSX and JSX files per example would help with that.

hiteshshetty-dev: Thanks, I’ve reverted the changes to `[example].tsx`

site/components/ExampleLayout.tsx
```diff
@@ -0,0 +1,235 @@
...
+  ['Paste HTML', 'paste-html'],
+  ['Plain Text', 'plaintext'],
+  ['Read-only', 'read-only'],
+  ['Rendering in iframes', 'iframe'],
+  ['Rich Text', 'richtext'],
+  ['Search Highlighting', 'search-highlighting'],
+  ['Shadow DOM', 'shadow-dom'],
+  ['Styling', 'styling'],
+  ['Tables', 'tables'],
+]
```
12joan: Is there an easy way of deduplicating this list with the one in `[example].tsx`?

hiteshshetty-dev: Yes, I missed this while migrating changes to `ExampleLayout`. I’ve now created a constants file with all the examples listed there, and both `Layout` and `[example].tsx` now use the same file

Issue #5962, opened 2025-10-16
Editor crashing when placed inside ShadowDOM, and "enter" is pressed on Android


iperzic:
**Description**
The editor crashes when "enter" is pressed on Android, and editor is placed inside a ShadowDOM.

**Recording**

![Image](https://github.com/user-attachments/assets/69b41cad-f1fd-4406-9cec-8d634ec7bf07)

**Sandbox**
No need for a sandbox, the error can be reproduced in [Shadow DOM example](https://www.slatejs.org/examples/shadow-dom)

**Steps**
To reproduce the behavior:
1. Go to https://www.slatejs.org/examples/shadow-dom
2. Choose an Android phone in responsive mode in browser devtools
4. Focus the editor
5. Press "enter"
6. See error "Cannot resolve a Slate node"

**Note:** Same thing happens on a real Android phone, this is just easier to reproduce.

**Expectation**
User should be able to type regardless of the platform used.

**Environment**
- Slate Version: 0.118.1
- Operating System: Android
- Browser: Chrome

**Context**
The problem is that some of the methods used to handle inputs on Android use native Web API, that doesn't pierce through Shadow DOM boundary, hence DOM elements aren't found.

PR #5963, opened 2025-10-16
Fix crashing editor when placed in ShadowDOM on Android

iperzic:
**Description**
This PR fixes the issue with using editor inside Shadow DOM, on Android phones. Also, it enables a "mobile" project in Playwright config, so that the test suite can catch bugs like these in the future.

**Issue**
Fixes: https://github.com/ianstormtaylor/slate/issues/5962

**Example**
CleanShot 2025-10-16 at 16 49 50

**Context**
I think that comments in the code are pretty self-explanatory as to what they do and how they fix the issue. If this isn't necessarily the case, I will update the PR description with more context.

**Checks**

### DIFF
--- a/playwright/integration/examples/shadow-dom.test.ts
+++ b/playwright/integration/examples/shadow-dom.test.ts
@@ -25,7 +25,7 @@ test.describe('shadow-dom example', () => {
 
     // Clear any existing text and type new text into the textbox
     await textbox.click()
-    await page.keyboard.press('Meta+A')
+    await page.keyboard.press('ControlOrMeta+A')
 
     await page.keyboard.press('Backspace')
     await page.keyboard.type('Hello, Playwright!')


dylans (changes requested):
The code is mostly ok, but there are a lot of comments for obvious things. Let's trim those up please, to just have comments for things that are not obvious.

packages/slate-dom/src/utils/dom.ts
```diff
@@ -355,3 +355,79 @@ export const isAfter = (node: DOMNode, otherNode: DOMNode): boolean =>
...
+  while (current) {
+    // Check if current element matches the selector
+    if (current.matches && current.matches(selector)) {
+      return current
+    }
+
+    // Try to go to the parent element
+    if (current.parentElement) {
+      current = current.parentElement
+    } else if ((current.parentNode as any)?.host) {
```
dylans: Do these two any types need to be any, or can we make the type more specific?

iperzic: Sorry, leftover. Added proper types in [c7b8139](https://github.com/ianstormtaylor/slate/pull/5963/commits/c7b8139762a5ac45b91900dac0de6b788905cb41)

playwright/integration/examples/shadow-dom.test.ts
```diff
@@ -24,11 +24,43 @@ test.describe('shadow-dom example', () => {
...
+    const pageErrors: Error[] = []
+    page.on('pageerror', error => {
+      pageErrors.push(error)
+    })
+
+    const outerShadow = page.locator('[data-cy="outer-shadow-root"]')
+    const innerShadow = outerShadow.locator('> div')
+    const textbox = innerShadow.getByRole('textbox')
+
+    // Add text in a new line
```
dylans:
```suggestion
```

iperzic:
There was a failing test on CI, can you please re-run the CI workflow?

PR #5968, opened 2025-11-06
Scroll to focus point of expanded selections

TyMick:
Currently, `defaultScrollSelectionIntoView` only scrolls collapsed selections into view, citing the following concern:

https://github.com/ianstormtaylor/slate/blob/c6137cdc20ecceb1f68b98b9d5bddefd6bb0c141/packages/slate-react/src/components/editable.tsx#L1954-L1955

This causes a bug when undoing the deletion of a range of text. To reproduce:

1. Go to https://www.slatejs.org/examples/richtext
2. Add enough text so that you can't see the first paragraph when you scroll to the bottom of the page
3. Select the whole first paragraph and delete it
4. Scroll down to the bottom of the page
5. Press <kbd>Ctrl</kbd>+<kbd>Z</kbd> or <kbd>⌘</kbd><kbd>Z</kbd> to undo
6. Note that the page does not automatically scroll back to the restored text
7. Scroll back manually to the top and confirm that the deleted text is restored

https://github.com/user-attachments/assets/8a82d13e-45c2-4936-9b80-19f29738a6e7

This PR updates `defaultScrollSelectionIntoView` to allow scrolling when the selection is expanded, but it specifically scrolls to the _focus point_ of the selection (rather than trying to scroll to the entire selected range). This fixes the undo-after-delete bug:

https://github.com/user-attachments/assets/71e3d1c5-a088-46cf-be95-9b7f9cccf51e

And scrolling to the focus point appears to avoid the issues with "the selection of multiple blocks and dragging behavior" that the original comment mentions. I've experienced those issues in development when allowing scrolling to expanded selections if I'm still trying to scroll to the entire `domRange`. But we've made this `domFocusPoint` change in our project's custom `scrollSelectionIntoView`, and I can't find any undesired behaviors when creating large selections, using either the mouse or the keyboard.

## Checks

### DIFF
--- a/.changeset/violet-moons-hope.md
+++ b/.changeset/violet-moons-hope.md
@@ -1,0 +1,5 @@
+---
+'slate-react': minor
+---
+
+Scroll to focus point of expanded selections in `defaultScrollSelectionIntoView`
--- a/packages/slate-react/src/components/editable.tsx
+++ b/packages/slate-react/src/components/editable.tsx
@@ -1951,18 +1951,17 @@ export const defaultScrollSelectionIntoView = (
   editor: ReactEditor,
   domRange: DOMRange
 ) => {
-  // This was affecting the selection of multiple blocks and dragging behavior,
-  // so enabled only if the selection has been collapsed.
-  if (
-    domRange.getBoundingClientRect &&
-    (!editor.selection ||
-      (editor.selection && Range.isCollapsed(editor.selection)))
-  ) {
-    const leafEl = domRange.startContainer.parentElement!
+  // Scroll to the focus point of the selection, in case the selection is expanded
+  const isBackward = !!editor.selection && Range.isBackward(editor.selection)
+  const domFocusPoint = domRange.cloneRange()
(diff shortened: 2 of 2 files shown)


dylans (changes requested):
For the most part this looks reasonable, but it does cause one of the chunking integration tests to fail. @TyMick can you please take a look at that and see if it's the test or your change causing the failure?

12joan:
Hi @dylans, it looks like the issue you're seeing on the Playwright integration test was a random flake, probably caused by the page taking slightly longer than usual to load. The same test succeeded on its second attempt. I'm pretty sure it's unrelated to this PR.

https://github.com/ianstormtaylor/slate/actions/runs/19149811365/job/54736850063?pr=5968#step:4:1328

TyMick:
Yeah the chunking test's flakiness seems unrelated to my changes. When I run the test locally, it's failing at the same rate on both my feature branch and the `main` branch.

dylans (approved):

Issue #5945, opened 2025-08-25
Slow Pasting of Large Text Content in Slate.js


electroluxcode:
Compared to Tiptap, Slate seems to perform poorly when pasting large amounts of data.  
I generated test data using:  
```javascript
let text = [];
for (let i = 0; i < 10000; i++) {
  text.push(`${i} this is a test demo. \n`);
}
let res = text.join('');
console.log(res);
```  
Then I pasted the generated text into Slate (https://www.slatejs.org/examples/plaintext) and Tiptap (https://tiptap.dev/docs/examples/basics/default-text-editor) respectively.  
Slate took about 7 seconds to parse the content, while Tiptap only took 1 second.  

<img width="2308" height="906" alt="Image" src="https://github.com/user-attachments/assets/1430eee2-584f-473a-8f5d-536800747b37" />

Are there any ways to optimize this?

12joan:
That's quite a big difference. I'm not sure how Tiptap handles pasting internally, but here's how Slate handles this case: `editor.insertTextData` splits the text into lines and alternates calling `editor.insertText(line)` and `Transforms.splitNodes(editor, { always: true })`.

88% of the total time in `insertData` is spent normalizing the editor after each split and insertion. It's possible that deferring normalizations until the end of `insertData` will improve things a little.

<img width="1531" height="706" alt="Image" src="https://github.com/user-attachments/assets/bd645a46-f01a-4d1b-952b-fe645b183689" />

Another interesting result is that around 50% of the time spent in `insertData` is spent inside `isEditor`, a utility that's supposed to check if a given object is a Slate editor. Specifically, most of this time is spent inside `isOperationList`, which checks if `editor.operations` is a correctly formed list of operations.

<img width="1533" height="135" alt="Image" src="https://github.com/user-attachments/assets/8cb073d9-27c5-42b6-9db7-ca08a91903d3" />

Since every operation needs to be checked, and the number of operations increases with each new line of text that has been inserted, this causes the overall time spent in `isOperationList` to have time complexity O(n^2).

I don't see why we should bother validating the exact structure of `editor.operations` just to determine that an object is a genuine editor. Checking for a property called `operations` should be more than sufficient.

@electroluxcode Do you want to try experimenting with optimising this?

electroluxcode:
[...]

PR #5970, opened 2025-11-16
Optimize Editor#normalizeNode

nabbydude:
**Description**
normalizeNode makes some expensive checks that really stack up when a lot of nodes change at once

**Issue**
Addresses #5945, thank you to the people in that thread for the research & analysis

**Example**
Running the test described in the above issue, pasting 10000 lines into the plaintext demo:

main:
image
this PR:
image

Current version takes ~11.9 seconds on my machine to handle the paste event once dispatched by React, ~9.7s of which is spent in `normalizeNode`. This PR shrinks those to ~1.9s and ~0.4s respectively.

**Context**
as @12joan pointed out, most of the time is taken up by `isEditor`, but there's only one editor in any given tree and its passed as an argument to the function so checking if a node is an editor is as simple as `node === editor.`

This implementation assumes any children that needed normalizing have been normalized already and are thus valid nodes. It also assumes its being called with normalizing disabled and so it can predict eg. that exactly one node will be removed when `removeNodes` is called on a path.

Checking for inline/blockiness of children is disentangled into two separate loops.

Adding `children: []` to a node that is neither text nor element to coerce it into an element now happens when you call `normalizeNode` on that node, not on its parent.

My regression testing has consisted of the mocha/jest suite and anecdotal testing with examples but this seems like the sort of area that could cause unexpected issues, especially mingling with custom extensions, but from what I can tell it seems fine.

**Checks**


packages/slate/src/core/normalize-node.ts
```diff
@@ -1,123 +1,134 @@
...
+  // Determine whether the node should have only block or only inline children.
+  // - The editor should have only block children.
+  // - Inline elements should have only inline children.
+  // - Elements that begin with a text child or an inline element child should have only inline children.
+  // - All other elements should have only block children.
+  const shouldHaveInlines =
+    !(element === editor) &&
+    (editor.isInline(element) ||
+      Text.isText(element.children[0]) ||
+      editor.isInline(element.children[0]))
```
nabbydude, line 54: this might be worth factoring into its own function, would allow implementers to easily enforce certain blocks to only allow block children (eg. lists -- currently if a list somehow ends up with a text node as its first child, its items will all become unwrapped)

packages/slate/src/core/normalize-node.ts
```diff
@@ -1,123 +1,134 @@
...
 
-  for (let i = 0; i < node.children.length; i++, n++) {
-    const currentNode = Node.get(editor, path)
-    if (Text.isText(currentNode)) continue
-    const child = currentNode.children[n] as Descendant
-    const prev = currentNode.children[n - 1] as Descendant
-    const isLast = i === node.children.length - 1
```
nabbydude, line 49: old implementation used `i` as an index into the snapshot of the node children passed into `normalizeNode` and `n` as an index into the children of the live node as its changed--this would be useful if we could use `i` to avoid `Node.get` calls by reading from the cached `node` whenever possible, but needing to know the prev child, which might be a newly added node, creates complexity.

packages/slate/src/core/normalize-node.ts
```diff
@@ -1,123 +1,134 @@
...
+          } else if (Text.equals(child, prev, { loose: true })) {
+            Transforms.mergeNodes(editor, { at: path.concat(n), voids: true })
+            element = Node.get(editor, path) as Element
+            n--
+          }
         }
       } else {
-        Transforms.unwrapNodes(editor, { at: path.concat(n), voids: true })
-      }
-      n--
```
nabbydude, line 70: pretty sure there's an invisible bug if this runs. we add some number of new child nodes and step back to read them--but we still only run the loop a number of times equal to the number of original children (everywhere else we add a child we skip over it and num iterations stay the same)

I dont think it caused issues due to the multi-pass nature that `normalizeNode` gets called but we can avoid running it more times than we need to by accounting for it.

packages/slate/src/core/normalize-node.ts
```diff
@@ -1,123 +1,134 @@
...
+          if (prev == null || !Text.isText(prev)) {
+            const newChild = { text: '' }
+            Transforms.insertNodes(editor, newChild, {
+              at: path.concat(n),
+              voids: true,
+            })
+            element = Node.get(editor, path) as Element
+            n++
+          }
+          if (n === element.children.length - 1) {
```
nabbydude, line 97: using if instead of else-if here so we can make the before and after nodes in the same pass--this could cause issues if any wrappers around `normalizeNode` rely on doing something in between but I can't imagine why anything reasonably would

dylans (commented):
I haven't been meaning to let this sit this long, just haven't had a solid block of time to think through it. I hope to look at it soon.

nabbydude:
this can land after. I'll update it with the new functions once theyre merged

nabbydude:
good to go 👍

Issue #4111, opened 2021-03-05
Editor selection is always null in Internet Explorer 11 (IE11) even with additional polyfills


marko-hologram:
**Description**
Editor selection is always null in Internet Explorer 11 (IE11) and therefore no editing can be applied to any text. No error is observed in developer tools console during this operation, it just fails silently.

**Recording**
IE11 behavior:
https://user-images.githubusercontent.com/22904204/110108060-ffd59880-7dab-11eb-9832-0d3e0dd3b509.mp4

Chrome/Firefox etc.
https://user-images.githubusercontent.com/22904204/110108332-55aa4080-7dac-11eb-9f36-38b7e2f9659c.mp4

**Sandbox**
Unfortunately codesandbox doesn't work in IE11 so I cannot really use these online environments to create a minimal example. I created this small repository instead:
https://github.com/marko-hologram/issue-slate-ie11

**Steps**
To reproduce the behavior:
1. Run the project as described in README
2. Select any text inside the editor
3. Press BOLD button or use `Ctrl + B` shortcut for example
4. Observe that `editor` logged inside the console has `selection: null` all the time and bold hasn't been applied

**Expectation**
- editor selection should not be `null` after making a selection
- bold styling should be applied after clicking BOLD button with a desired text selection

**Environment**
- Slate Version: 0.59.0
- Operating System: Windows 10
- Browser: Internet Explorer 11 (IE11)

**Sidenote**
This project uses `next-transpile-modules` to transpile `slate` and `slate-react` packages in order for them to be used in IE11 because otherwise the code fails.

Polyfills used are from this comment https://github.com/ianstormtaylor/slate/issues/3800#issuecomment-758472391, just without the TS types and adapted for Nextjs to run only in browser, not on server.

---

I tried poking around some code in order to try to figure out where the problem happens, but I didn't come up with anything. I checked some code where some Selection logic is being triggered and as far as I've seen this useEffect doesn't run after doing a text selection, it only runs on initial mount.
https://github.com/ianstormtaylor/slate/blob/513771c82ad3c464bee3c7e3b73bbce189c6cf6e/packages/slate-react/src/components/editable.tsx#L149

[...]

PR #5971, opened 2025-11-17
Remove immer, making the core `slate` package dependency-free

### LINKED ISSUE #5644: Update to immer@10+ breaks react native + hermes builds
**Description**
The update to `"immer": "^10.0.3"` in https://github.com/ianstormtaylor/slate/pull/5528 prevents Slate from building in react native projects that utilize the hermes engine. The newer version of `immer` utilizes some features which are not implemented in [hermes yet](https://github.com/facebook/hermes/issues/685#issuecomment-1784172071).

We utilize Slate in react native to build and normalize documents that match our web application even though we don't use it to control an actual input element.

**Steps**
To reproduce the behavior:
1. Add Slate as a dependency to a React Native project
2. Enable hermes engine
3. Attempt a release build of the application

**Expectation**
I'm unsure if Slate should take any action here. We are working around the issue by pinning `immer` to `9.0.21` and forcing slate to use that version. This currently works because slate hasn't updated any usages of `immer` so it is still compatible with `9.0.21`.

This is largely an FYI issue for others. Additionally, resisting any `immer` 10+ backwards incompatible code might want to be avoided to continue supporting react native and older browser usages until strictly necessary.

**Environment**
- Slate Version: 0.102

nabbydude:
**Description**
After #5871, the only usage of `immer` left is in `Node.fragment` for creating a fragment object. This PR switches it to using the existing non-immer functions. 

**Issue**
Mainly decreases bundle size for the `slate` package
Fixes #5644 by circumvention, could affect a solution posted for #4111

**Context**
I couldn't find any actual unit tests for fragments, but manual testing has yielded identical results. If I have the time and patience I might add some next week

also this removes `tiny-warning` which was entirely unused already

**Checks**

### DIFF
--- a/.changeset/purple-tigers-leave.md
+++ b/.changeset/purple-tigers-leave.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Removed the last traces of immer and eliminated it as a dependency


dylans (approved):
I guess immer is not really always. 😂

This looks fine to me, going to land another fix to hopefully clean up the failing tests.

Issue #5973, opened 2025-11-18
#5963 regression: "containsShadowAware" is not exported by slate-dom


philicious:
**Description**
#5963 introduced `containsShadowAware` in slate-dom and slate-react makes use of it. However the PR didnt bump the required peer-dep version to `0.119.0`, only the dev dep.
https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/package.json#L46

For that reason builds are failing after update to slate-react 0.119.0 unless you explicitly install slate-dom@0.119.0

> error during build:
node_modules/slate-react/dist/index.es.js (7:800): "containsShadowAware" is not exported by "node_modules/slate-dom/dist/index.es.js", imported by "node_modules/slate-react/dist/index.es.js".
file: /app/node_modules/slate-react/dist/index.es.js:7:800

semimikoh:
@dshelters 

I can take this issue if no one is working on it.

The fix looks straightforward — updating the `slate-react` peer dependency to `slate-dom@0.119.0` should resolve the regression.  
Is it okay if I open a PR for this?

12joan:
@semimikoh That would be great, thanks! Generally, you can just go ahead and work on things like this. The chances of someone else working on it at the same time are quite small.

PR #5976, opened 2025-11-20
fix(slate-react): update slate-dom peer dependency to 0.119.0

semimikoh:
## Description
This pull request fixes a regression introduced in slate-react@0.119.0.

In this release, slate-react began importing containsShadowAware from slate-dom. However, the peerDependencies entry for slate-dom was not updated to require the corresponding version (>=0.119.0) where containsShadowAware was added.

As a result, users installing slate-react@0.119.0 alongside an older slate-dom version encountered a build-time error:

"containsShadowAware" is not exported by node_modules/slate-dom/dist/index.es.js

Updating the peerDependency resolves this issue.

## Issue
Fixes: https://github.com/ianstormtaylor/slate/issues/5973

## Example
Minimal reproduction:

1. Install slate-react@0.119.0 with slate-dom@0.118.1
2. Run a build
3. Build fails with: "containsShadowAware is not exported by slate-dom"

After this change, slate-react@0.119.0 works correctly when paired with slate-dom >= 0.119.0.

This PR introduces no UI changes or runtime behavior changes — it only corrects the dependency specification.

## Context
The regression occurred because PR #5963 introduced containsShadowAware in slate-dom, and slate-react began using it, but the peer dependency version range was not updated.

devDependencies already pointed to slate-dom ^0.119.0, so the issue only affected downstream users relying on peerDependency resolution.

This PR aligns the expected version and prevents build failures.

The fix is intentionally minimal: updating the peerDependencies.slate-dom field in slate-react/package.json.

## Checks
- [x] The new code matches existing patterns and styles.
- [x] Tests pass with yarn test.
- [x] The linter passes with yarn lint.
- [x] Relevant examples still work (yarn start).
- [x] A changeset has been added (yarn changeset add).

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -13743,7 +13743,7 @@ __metadata:
     react: ">=18.2.0"
     react-dom: ">=18.2.0"
     slate: ">=0.114.0"
-    slate-dom: ">=0.116.0"
+    slate-dom: ">=0.119.0"
   languageName: unknown
   linkType: soft
 


12joan:
Thanks! Please could you add a patch changeset to `slate-react` so that the new version with the fixed peer dependency is released?

semimikoh:
@12joan Thanks! I've added the patch changeset for slate-react.

.changeset/fix-slate-dom-peer-dependency.md
```diff
@@ -0,0 +1,5 @@
+---
+"slate-react": patch
```
dylans:
```suggestion
'slate-react': patch
```

dylans (approved):

dylans (changes requested):
You need to commit the change to yarn.lock please.

semimikoh:
@dylans  I've committed the yarn.lock changes. Please review when you have a chance. Thanks!

Issue #5979, opened 2025-11-21
Handwriting on Android prematurely commits initial strokes


aiden-art:
**Description**

On Android devices, using handwriting input (e.g., Sogou Input Method) causes the initial stroke of a character to be committed prematurely. For example, when writing "我", the first stroke "一" is inserted immediately, leading to a final incorrect result of "一我".

This is reproducible on the official Slate examples.

**Recording**

https://github.com/user-attachments/assets/0e2e5d27-c7d1-4dbf-bc7c-7cae1ea6bccc

https://github.com/user-attachments/assets/e33f53c6-abc7-4505-bada-da47f082f534

**Sandbox**

Reproducible on the official Rich Text example: https://www.slatejs.org/examples/richtext

**Steps to Reproduce**

1.  On an Android device, open Chrome.
2.  Go to the Slate Rich Text example link above.
3.  Use a handwriting keyboard (e.g., Sogou).
4.  Write a multi-stroke character like "我".
5.  Observe that the first stroke is inserted incorrectly.

**Expected Behavior**

The editor should only insert the final, complete character ("我") after the user finishes writing and selects it.

**Environment**

-   **Slate Version:** `0.102.0`
-   **OS:** Android 11
-   **Browser:** Chrome
-   **Input Method:** Sogou Input Method (Handwriting)

**Context**

We have tried several client-side workarounds without success, including:
-   Using `event.preventDefault()` in `onDOMBeforeInput`.
-   Implementing state locks with `useRef` to manage composition events.
-   Debouncing `onChange` during composition sessions.

**Question**

This appears to be a core issue with how Slate handles Android's handwriting IME events. Is there a known workaround or an official solution for this?

PR #5981, opened 2025-11-27
fix(slate-react): prevent premature flush during composition for Andr…

aiden-art:
## Description

On Android devices, using handwriting input (e.g., Sogou Input Method) causes the initial stroke of a character to be committed prematurely. For example, when writing "我", the first stroke "一" is inserted immediately, leading to a final incorrect result of "一我".

The root cause is that Android handwriting IME triggers `compositionStart`/`compositionEnd` events per stroke, unlike keyboard input which only fires these events at the beginning and end of the entire input session. During composition, multiple code paths (`handleInput()`, `scheduleAction()`, `handleUserSelect()`) were calling `flush()`, which applies pending diffs to the editor via `Editor.insertText()`. This DOM modification causes the browser to terminate the current composition session, and the next stroke starts a new composition that appends to the already-committed text rather than replacing it.

The fix adds `IS_COMPOSING` checks in `flush()`, `scheduleAction()`, and `handleInput()` to prevent flushing during an active composition session. The pending diffs now properly accumulate and are only flushed after `compositionEnd` fires.

## Issue

Fixes: https://github.com/ianstormtaylor/slate/issues/5979

**Checks**

### DIFF
--- a/.changeset/brave-dragons-write.md
+++ b/.changeset/brave-dragons-write.md
@@ -1,0 +1,10 @@
+---
+'slate-react': patch
+---
+
+Fix Android handwriting input where first stroke commits prematurely on empty lines (e.g., '我' becomes '一我').
+
+**Cause:** When writing the first character of an empty line, Android handwriting triggers childList mutations (creating new DOM structure). These mutations get restored by RestoreDOM during React re-render, which interrupts the IME composition session.
+
+**Fix:** Skip flush during composition only when at the first character of an empty line. This targeted fix preserves normal onChange behavior for other scenarios while preventing the handwriting issue.
+
--- a/packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
+++ b/packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
@@ -123,7 +123,35 @@ export function createAndroidInputManager({
     action.run()
   }
 
+  // Helper: Check if cursor is at the first character of an empty line
+  const isAtEmptyLineFirstChar = (): boolean => {
+    const { selection, children } = editor
+    if (!selection) return false
+    
+    const { anchor } = selection
+    // offset should be 0 or 1 (first char)
+    if (anchor.offset > 1) return false
+    
+    // Get the block node at current path
+    const blockPath = anchor.path.slice(0, 1) // Get top-level path
+    const blockNode = children[blockPath[0]]
+    if (!blockNode) return true
+    
+    // Check if this line/block is empty or has only one character
+    const text = Node.string(blockNode)
+    return text.length <= 1
+  }
+
   const flush = () => {
+    // FIX: Only skip flush during composition for empty line first char scenario.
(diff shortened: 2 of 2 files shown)


12joan:
@aiden-art What are the implications of this for the timing of `onChange` and `insertText`? I think previously, `onChange` was called continuously during composition; is this still the case?

Can you also check if the [mentions example](https://www.slatejs.org/examples/mentions) still works with this change? The combobox should update itself while composition is still in progress. Many Android devices also use composition for writing English text, so this could break a lot of features if we're not careful.

aiden-art:
@12joan Thanks for the thoughtful review!

Yes, this change does affect `onChange` timing during composition:

- **Before**: `onChange` was called continuously during composition
- **After**: `onChange` is deferred until `compositionEnd` + 25ms 

I've tested the [mentions example](https://deploy-preview-5981--slatejs.netlify.app/examples/richtext) and it works correctly. I'll attach a video demonstrating this.

https://github.com/user-attachments/assets/d7064a16-39b0-41e2-97b4-ef10645d0351

Here's a video demonstrating the fix for Android handwriting input:

https://github.com/user-attachments/assets/f5c80b63-9497-435a-996b-aa7d1a81fa4a

12joan:
@aiden-art Thanks for confirming that about `onChange`.

It looks like in your video where you try out the mentions example, your device isn't using composition for English text, which I believe is common in Gboard and perhaps other keyboard applications running on recent versions of Android. 

Samsung Keyboard and older versions of Gboard also use composition for English, which you can tell by the underline while typing. I've just tried your PR on my older Samsung phone, and the combobox doesn't show up until I explicitly end composition.

I'm seeing the same thing on a Samsung Galaxy S23 running on BrowserStack Live.

https://github.com/user-attachments/assets/7a5b6cd1-8531-408d-a550-e22d286e9f6f

Ideally, it would be great if we could fix this problem without causing any breaking changes. If changing the way Slate works during composition is unavoidable, then we should make sure that existing patterns have valid workarounds, and these should be clearly communicated in the changeset and probably the docs too.

aiden-art:
Thanks for your feedback and testing!

After detailed log analysis, I found that the issue only occurs when writing **the first character of an empty line** (including new lines). 

**Root cause:** When writing on an empty line, Android handwriting triggers `childList` mutations (creating new DOM structure). These mutations get restored by `RestoreDOM`, which interrupts the IME composition session. On lines with existing content, only `characterData` mutations occur, which are already skipped by RestoreDOM.

**Current fix:** Skip flush during composition only when at the first character of an empty line:

if (IS_COMPOSING.get(editor) && isAtEmptyLineFirstChar()) {
  return
}However, this is a very targeted fix that may not cover all edge cases. Since this doesn't seem applicable to most scenarios (like the Samsung keyboard composition issue you mentioned), I'll close this PR.

I hope this analysis and approach can help others who encounter similar handwriting input issues on Android. The key insight is that the problem is related to `childList` vs `characterData` mutation types and how RestoreDOM handles them during composition.

PR #5982, opened 2025-12-04
Avoid overly-cautious type guards in favor of more performant type discriminators

### LINKED ISSUE #5980: Avoid overly-cautious type guards for nodes
**Problem**
There are many parts of the codebase where a function receives a Node and needs to determine whether its a Text, Element, or Editor, which roughly works as follows:
 - Is it a Text?
    - Does it have a `text` property?
        - Then yes, its a Text. Done 😄
- Is it an Element?
    - Does it have a `children` property?
        - Then yes, its a --oh wait it could still be an Editor, gotta make sure its not that first.
- Is it an Editor?
    - I'm going to assume this is a radioactive value and check every single property type and each of those properties' property types... (the current `isEditor` check [doesn't even properly check for every property](https://github.com/ianstormtaylor/slate/blob/49f28e5067b213431bd16e423524d608cd819a80/packages/slate/src/editor/is-editor.ts#L16-L35) of [the current Editor object](https://github.com/ianstormtaylor/slate/blob/49f28e5067b213431bd16e423524d608cd819a80/packages/slate/src/interfaces/editor.ts#L44-L178) but I digress)
        - Okay, whew, that was a lot but yes this is definitely an editor 👍

The current hack to prevent a bloated `isEditor` subcall for every single `isElement` call is just to check for the `apply` function and assume anything with `apply` is an editor (and therefore not an Element).
https://github.com/ianstormtaylor/slate/blob/49f28e5067b213431bd16e423524d608cd819a80/packages/slate/src/interfaces/element.ts#L82-L84

**Solution**
I have a few related suggestions:
1. ~~Codify this check in the schema, that an Element is an object with a `children` property and *without* an `apply` property.~~ (Elements are always JSON, so `apply` being a function only for editors is already part of the spec, whoops!)
2. Implement a discriminator function(s) that can assume its input is a Node and only needs to choose between them. This saves on checking types of children, and editor can be checked by just checking the `apply` property.
    I'm thinking something like:
    ```ts
    function nodeIsText(node: Node): node is Text
    ```
    as opposed to:
    ```ts
    function isText(node: any): node is Text
    ```
3. Avoid using `isEditor` whenever possible, by either comparing the node to a known editor object or checking the length of its path.
3. (bonus) properly check for all properties of `Editor` in `isEditor`

**Alternatives**
~~The specific property isn't crucial for my proposal, we could just as easily check for the (non)existence of `onChange` or `operations` or `shouldMergeNodesRemovePrevNode`. It also can be more specific, like specifically checking if its a function.~~ (I really wrote that last part forgetting that it's already the case since Elements are JSON only...)

There are other ways this could be done, like turning `Editor` into a class or creating some new property/symbol for this check, but to me this way seemed the most in line with the current slate ethos.

**Context**
Happy to implement this myself, just wanted to hear other thoughts before running off on my own.

nabbydude:
**Description**
Adds `Editor.isEditorNode`, `Element.isElementNode`, and `Text.isTextNode` that take a `Node` argument and check quickly if it's a specific node type.

Also fixes classic `isEditor` to actually check for all properties, actually respect deepness settings (a bug I only noticed because I was editing the file 😅), and also respect those deepness settings for its operation list.

This also removes any calls to `Element.isAncestor`, since it was only ever used where `!Text.isTextNode` would function just fine. It doesn't remove the function itself since its user accessible. Rigourously testing for any of two node types from `any` like this is something I cant think of a real use case for and it might be worth deprecating.

**Issue**
Fixes #5980
Basically checking if something is an object all the time and checking it has every single function its supposed to is a waste of performance, 99% of the time we can just quickly check if its definitely not one of the other two types and move on.

**Context**
This touches a lot of files, since it replaces almost every call of  `isEditor`, `isElement`, and `isText` in the `slate` package.
I've removed a few redundant calls and cleaned up a few spaghetti boolean expressions.

`Editor.isEditor` previously only passed its first argument to the function it wrapped, which meant deepness options were ignored. this PR just adds that function directly to the object without wrapping it to avoid that happening again. `isEditorNode` is included in the same way.

I have another open PR (#5970) that changes `normalizeNode`, which this will conflict with. If both end up approved I have a merging commit ready.

**Alternatives**
I chose these names mostly to be close to the originals in alphabetical order, and to keep the `isX` paradigm. other options could be `nodeIsEditor`, `isNodeEditor`, `Node.isEditor`. I can change them if something else is preferred.

We could overwrite the old functionality and migrate the cautious type guard to `verifyElement` or some other name but that might break existing user code.

**Future**
- [ ] Add docs
- [x] Use in other packages in repo
- [x] Use in examples

**Checks**

### DIFF
--- a/site/examples/js/inlines.jsx
+++ b/site/examples/js/inlines.jsx
@@ -2,13 +2,7 @@ import { css } from '@emotion/css'
 import { isKeyHotkey } from 'is-hotkey'
 import isUrl from 'is-url'
 import React, { useMemo } from 'react'
-import {
-  createEditor,
-  Editor,
-  Element as SlateElement,
-  Range,
-  Transforms,
-} from 'slate'
+import { createEditor, Editor, Node, Range, Transforms } from 'slate'
 import { withHistory } from 'slate-history'
 import { Editable, useSelected, useSlate, withReact } from 'slate-react'
 import * as SlateReact from 'slate-react'
@@ -143,24 +137,24 @@ const insertButton = editor => {
 }
 const isLinkActive = editor => {
   const [link] = Editor.nodes(editor, {
-    match: n => SlateElement.isElementNode(n) && n.type === 'link',
+    match: n => Node.isElement(n) && n.type === 'link',
   })
   return !!link
 }
 const isButtonActive = editor => {
   const [button] = Editor.nodes(editor, {
-    match: n => SlateElement.isElementNode(n) && n.type === 'button',
+    match: n => Node.isElement(n) && n.type === 'button',
   })
   return !!button
 }
 const unwrapLink = editor => {
   Transforms.unwrapNodes(editor, {
-    match: n => SlateElement.isElementNode(n) && n.type === 'link',
+    match: n => Node.isElement(n) && n.type === 'link',
   })
 }
 const unwrapButton = editor => {
   Transforms.unwrapNodes(editor, {
-    match: n => SlateElement.isElementNode(n) && n.type === 'button',
+    match: n => Node.isElement(n) && n.type === 'button',
   })
 }
 const wrapLink = (editor, url) => {
--- a/site/examples/js/check-lists.jsx
+++ b/site/examples/js/check-lists.jsx
@@ -1,13 +1,6 @@
 import { css } from '@emotion/css'
 import React, { useCallback, useMemo } from 'react'
-import {
-  Editor,
-  Point,
-  Range,
-  Element as SlateElement,
-  Transforms,
-  createEditor,
-} from 'slate'
+import { Editor, Node, Point, Range, Transforms, createEditor } from 'slate'
 import { withHistory } from 'slate-history'
 import {
   Editable,
@@ -85,8 +78,7 @@ const withChecklists = editor => {
     const { selection } = editor
     if (selection && Range.isCollapsed(selection)) {
       const [match] = Editor.nodes(editor, {
-        match: n =>
-          SlateElement.isElementNode(n) && n.type === 'check-list-item',
+        match: n => Node.isElement(n) && n.type === 'check-list-item',
       })
       if (match) {
         const [, path] = match
@@ -96,8 +88,7 @@ const withChecklists = editor => {
             type: 'paragraph',
           }
           Transforms.setNodes(editor, newProperties, {
-            match: n =>
-              SlateElement.isElementNode(n) && n.type === 'check-list-item',
+            match: n => Node.isElement(n) && n.type === 'check-list-item',
           })
           return
         }
--- a/site/examples/js/code-highlighting.jsx
+++ b/site/examples/js/code-highlighting.jsx
@@ -11,7 +11,7 @@ import 'prismjs/components/prism-sql'
 import 'prismjs/components/prism-tsx'
 import 'prismjs/components/prism-typescript'
 import React, { useCallback, useState } from 'react'
-import { Editor, Element, Node, Transforms, createEditor } from 'slate'
+import { Editor, Node, Transforms, createEditor } from 'slate'
 import { withHistory } from 'slate-history'
 import {
   Editable,
@@ -101,14 +101,14 @@ const CodeBlockButton = () => {
       editor,
       { type: CodeBlockType, language: 'html', children: [] },
       {
-        match: n => Element.isElementNode(n) && n.type === ParagraphType,
+        match: n => Node.isElement(n) && n.type === ParagraphType,
         split: true,
       }
     )
     Transforms.setNodes(
       editor,
       { type: CodeLineType },
-      { match: n => Element.isElementNode(n) && n.type === ParagraphType }
+      { match: n => Node.isElement(n) && n.type === ParagraphType }
     )
   }
   return (
@@ -135,7 +135,7 @@ const renderLeaf = props => {
 }
 const useDecorate = () => {
   return useCallback(([node, path]) => {
-    if (Element.isElementNode(node) && node.type === CodeBlockType) {
+    if (Node.isElement(node) && node.type === CodeBlockType) {
       return decorateCodeBlock([node, path])
     }
     return []
--- a/site/examples/js/forced-layout.jsx
+++ b/site/examples/js/forced-layout.jsx
@@ -1,11 +1,5 @@
 import React, { useCallback, useMemo } from 'react'
-import {
-  Editor,
-  Node,
-  Element as SlateElement,
-  Transforms,
-  createEditor,
-} from 'slate'
+import { Editor, Node, Transforms, createEditor } from 'slate'
 import { withHistory } from 'slate-history'
 import { Editable, Slate, withReact } from 'slate-react'
 
@@ -34,7 +28,7 @@ const withLayout = editor => {
         let type
         const slateIndex = childPath[0]
         const enforceType = type => {
-          if (SlateElement.isElementNode(child) && child.type !== type) {
+          if (Node.isElement(child) && child.type !== type) {
             const newProperties = { type }
             Transforms.setNodes(editor, newProperties, {
               at: childPath,
--- a/site/examples/js/markdown-preview.jsx
+++ b/site/examples/js/markdown-preview.jsx
@@ -2,7 +2,7 @@ import { css } from '@emotion/css'
 import Prism from 'prismjs'
 import 'prismjs/components/prism-markdown'
 import React, { useCallback, useMemo } from 'react'
-import { Text, createEditor } from 'slate'
+import { Node, createEditor } from 'slate'
 import { withHistory } from 'slate-history'
 import { Editable, Slate, withReact } from 'slate-react'
 
@@ -11,7 +11,7 @@ const MarkdownPreviewExample = () => {
   const editor = useMemo(() => withHistory(withReact(createEditor())), [])
   const decorate = useCallback(([node, path]) => {
     const ranges = []
-    if (!Text.isTextNode(node)) {
+    if (!Node.isText(node)) {
       return ranges
     }
     const getLength = token => {
--- a/site/examples/js/markdown-shortcuts.jsx
+++ b/site/examples/js/markdown-shortcuts.jsx
@@ -1,13 +1,5 @@
 import React, { useCallback, useMemo } from 'react'
-import {
-  createEditor,
-  Editor,
-  Element as SlateElement,
-  Node as SlateNode,
-  Point,
-  Range,
-  Transforms,
-} from 'slate'
+import { createEditor, Editor, Node, Point, Range, Transforms } from 'slate'
 import { withHistory } from 'slate-history'
 import { Editable, ReactEditor, Slate, withReact } from 'slate-react'
 
@@ -37,15 +29,14 @@ const MarkdownShortcutsExample = () => {
           if (!diff.text.endsWith(' ')) {
             return false
           }
-          const { text } = SlateNode.leaf(editor, path)
+          const { text } = Node.leaf(editor, path)
           const beforeText = text.slice(0, diff.start) + diff.text.slice(0, -1)
           if (!(beforeText in SHORTCUTS)) {
             return
           }
           const blockEntry = Editor.above(editor, {
             at: path,
-            match: n =>
-              SlateElement.isElementNode(n) && Editor.isBlock(editor, n),
+            match: n => Node.isElement(n) && Editor.isBlock(editor, n),
           })
           if (!blockEntry) {
             return false
@@ -79,7 +70,7 @@ const withShortcuts = editor => {
     if (text.endsWith(' ') && selection && Range.isCollapsed(selection)) {
       const { anchor } = selection
       const block = Editor.above(editor, {
-        match: n => SlateElement.isElementNode(n) && Editor.isBlock(editor, n),
+        match: n => Node.isElement(n) && Editor.isBlock(editor, n),
       })
       const path = block ? block[1] : []
       const start = Editor.start(editor, path)
@@ -95,8 +86,7 @@ const withShortcuts = editor => {
           type,
         }
         Transforms.setNodes(editor, newProperties, {
-          match: n =>
-            SlateElement.isElementNode(n) && Editor.isBlock(editor, n),
+          match: n => Node.isElement(n) && Editor.isBlock(editor, n),
         })
         if (type === 'list-item') {
           const list = {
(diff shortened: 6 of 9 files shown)


packages/slate/src/transforms-node/merge-nodes.ts
```diff
@@ -9,20 +9,13 @@ import { Scrubber } from '../interfaces/scrubber'
...
-  } else {
-    return true
-  }
+  return (
+    node !== editor &&
+    (Text.isTextNode(node) ||
+      Editor.isVoid(editor, node) ||
+      (node.children.length === 1 &&
+        hasSingleChildNest(editor, node.children[0])))
+  )
```
nabbydude, line 18: same functionality, rearranged for readability.
Basically its a nest if its a leaf or its only child is a leaf--and editor doesn't count.

packages/slate/src/transforms-node/wrap-nodes.ts
```diff
@@ -70,16 +71,14 @@ export const wrapNodes: NodeTransforms['wrapNodes'] = (
...
-      })
-    )
+    const roots = editor.isInline(element)
+      ? Editor.nodes(editor, {
+          at,
+          match: n => Element.isElementNode(n) && Editor.isBlock(editor, n),
+          mode: 'lowest',
+          voids,
+        })
+      : [[editor, []] as NodeEntry]
```
nabbydude, line 81: `Editor.nodes` with `match: n => Editor.isEditor(n)` only ever returns the same one node: the editor.

packages/slate/src/interfaces/node.ts
```diff
@@ -322,19 +322,19 @@ export const Node: NodeInterface = {
...
     for (const [node, path] of Node.nodes(root, options)) {
-      if (Element.isElement(node)) {
+      if (Element.isElementNode(node)) {
         yield [node, path]
       }
     }
   },
 
   extractProps(node: Node): NodeProps {
-    if (Element.isAncestor(node)) {
```
12joan, line 332: Would it be worth adding an `Element.isAncestorNode` to make it easier to convert existing code like this?

packages/slate/src/transforms-node/unwrap-nodes.ts
```diff
@@ -49,7 +50,7 @@ export const unwrapNodes: NodeTransforms['unwrapNodes'] = (
 
       Transforms.liftNodes(editor, {
         at: range,
-        match: n => Element.isAncestor(node) && node.children.includes(n),
```
12joan, line 52: Same here. I find `isAncestor` more expressive, as it essentially means "has children".

packages/slate/src/transforms-node/wrap-nodes.ts
```diff
@@ -70,16 +71,14 @@ export const wrapNodes: NodeTransforms['wrapNodes'] = (
...
```
12joan: I find the prior version of this a little easier to read, even if it's more indirect in `match`ing just the editor. Are you able to write the new version without a type assertion?

.changeset/rich-cheetahs-battle.md
```diff
@@ -0,0 +1,7 @@
+---
+'slate': minor
+---
+
+Add `Editor.isEditorNode`, `Element.isElementNode`, and `Text.isTextNode` as alternative type guards for when we already know the object is a node.
+Use these new functions instead of `isEditor`, `isElement`, and `isText` whenever possible, the classic functions are only necessary for typechecking an entirely unknown object.
```
12joan: Nice! Should we also change the docs to refer to the new type guards by default?

packages/slate/src/editor/is-editor.ts
```diff
@@ -13,26 +13,95 @@ export const isEditor: EditorInterface['isEditor'] = (
...
+    typeof value.withoutNormalizing === 'function' &&
+    typeof value.wrapNodes === 'function' &&
     (value.marks === null || isObject(value.marks)) &&
     (value.selection === null || Range.isRange(value.selection)) &&
-    (!deep || Node.isNodeList(value.children)) &&
-    Operation.isOperationList(value.operations)
+    (deep
+      ? Node.isNodeList(value.children) &&
+        Operation.isOperationList(value.operations)
+      : Array.isArray(value.children) && Array.isArray(value.operations))
```
12joan, line 104: Nice fix!

packages/slate/src/editor/is-editor.ts
```diff
@@ -13,26 +13,95 @@ export const isEditor: EditorInterface['isEditor'] = (
   }
 
   const isEditor =
```
12joan, line 15: I'm not sure whether it's worth including this many `typeof value.x === 'function'` checks. Each one carries a small overhead, and there's a big difference between having 16 of them versus 86 (apologies if I miscounted).

Here's a benchmark I ran to compare these numbers of `typeof` checks. Since the numbers don't mean much in isolation, I also included a [`WeakMap.prototype.set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap/set) call for comparison. Setting and getting weak map values tends to be relatively slow, so anything slower than it should generally be avoided if possible.

https://jsbm.dev/xM6mIrc116gyW - Longer bars are faster.

|Firefox|Chrome|Safari|
|-|-|-|
|image|image|image|

12joan:
> This only changes the slate package (mostly because the packages depend on the live versions of eachother and typescript complained if I tried to call the newly added methods) the other packages can be changed once this is merged, and until then will continue to operate gracefully. Same for any user code.

The other packages don't use the live versions; I believe they use the transpiled versions of the packages found in `dist`. There should be a command you can run to automatically keep these `dist` versions up to date. `yarn watch`, I think.

Bear in mind that if you do update the other packages to depend on the new functions, you'll also need to bump the peer dependency versions.

12joan:
> We could overwrite the old functionality and migrate the cautious type guard to verifyElement or some other name but that might break existing user code.

I would be in favour of this personally, but I'd be interested to hear what others think.

packages/slate/src/interfaces/node.ts
nabbydude: If we did it would probably end up as just a wrapper around `!Text.isTextNode` (I'd assume checking string typing is faster than checking array typing although tbf I haven't checked)

I'm not personally a fan of those sorts of inverse functions-- "is not a leaf" p directly translates to "has children" in my mind. That being said it doesn't really get in my way and I don't hate it enough to boycott it if it were added.

packages/slate/src/transforms-node/wrap-nodes.ts
nabbydude: Does `as const` count? 😄

It's definitely not pretty, I'll see if I can take a crack at de-jankifying it

.changeset/rich-cheetahs-battle.md
nabbydude: I definitely think so, user code should only need the old functions for serializing/deserializing etc, basically call match functions etc are better off with the new ones.

Just didn't want to cement them in too deep before getting feedback on the naming. Thankfully old code still functions, just inefficiently (equally as inefficiently as before, to be precise)

packages/slate/src/editor/is-editor.ts
nabbydude: Yeah, my intention would be that `Editor.isEditor` is basically never used. The only current use case I can imagine in the current codebase is *maybe* the sanity check that `slate-react`'s `Slate` component does when its first passed an `editor` prop but really that should be unnecessary in a typescripted world.

As I'm typing though I do see the perspective of any user code that updates its dependency without tweaking its calls getting an unexpected performance hit. 🤔

nabbydude:
> The other packages don't use the live versions; I believe they use the transpiled versions of the packages found in `dist`. There should be a command you can run to automatically keep these `dist` versions up to date. `yarn watch`, I think.

Well that'll show me assuming repos other people create are as jank as my own. In that case I can add the other packages changes to this PR

packages/slate/src/transforms-node/wrap-nodes.ts
nabbydude: I can see a possible nicer structure but it'd require warping more of the function around it so I'll save it for its own PR, reverted to just the guard changes for now.

site/examples/js/inlines.jsx
```diff
@@ -370,8 +367,8 @@ const ToggleEditableButtonButton = () => {
   return (
     <Button
       active
-      onMouseDown={event => {
-        event.preventDefault()
+      onPointerDown={event => event.preventDefault()}
+      onClick={() => {
```
nabbydude, line 365: these were changed by `yarn tsc:examples`, they already existed in the tsx files and weren't synced

nabbydude:
replaced the calls in other packages as well. A few were left in intentionally:

https://github.com/nabbydude/slate/blob/2f75743c89b7561d95b1b5cd8cd5a88f7f5dc925/packages/slate-react/src/components/slate.tsx#L47-L51
This check gets called when a `<Slate>` component gets first mounted. I'd be in favor of this check being deleted entirely because if typescript is being respected it never ever happens, but it isn't a hot path so for now I'm just assuming it's here on purpose for baby-proofing.

https://github.com/nabbydude/slate/blob/2f75743c89b7561d95b1b5cd8cd5a88f7f5dc925/packages/slate-hyperscript/src/creators.ts#L35
This is being received as `any` by the surrounding function and I don't have enough familiarity with slate-hyperscript to write more sensible type guards so I'll assume it could be necessary and leave it for now.

I also didn't touch any of the tests, since rigorous checks have purpose there.

dylans:
@12joan anything else you'd like to see changed with this PR?

packages/slate/src/interfaces/node.ts
12joan: I'm thinking there are more reasons to include it than to exclude it, even if it's a wrapper for `!Text.isTextNode`.

12joan:
@dylans No, overall it looks pretty good to me.

On the subject of naming, I like the idea of keeping the existing function names but changing their implementations and type signatures to expect a `Node` argument. This should make for an easier migration, since most places won't require any code changes, and it doesn't introduce the confusion of two similarly named families of functions.

We could maybe keep the existing implementations as a new `verify` family of functions, but I'm not 100% sure this is worth doing. If developers need to recursively check that a Slate value is correctly structured, it isn't difficult for them to do so themselves, and a custom solution could also check that custom properties are correct.

Perhaps to make migration easier, we could keep `Node.isNode`'s argument type as `any` and encourage developers to use `Node.isNode(anyObject) && Element.isElement(anyObject)` in the rare cases that node and non-node values are mixed.

12joan:
Also, don't forget that if we do end up changing `slate-react` and other packages to depend on new function names, we'll also need to bump their peer dependencies on `slate` prior to publishing them.

nabbydude:
> On the subject of naming, I like the idea of keeping the existing function names but changing their implementations and type signatures to expect a Node argument [...]

FWIW I agree with this if we can do it, most uses I imagine are in match functions that only take nodes anyways. In a strict sense it's a breaking change but in practice v few people would have to change any code--and any that do on typescript would see it immediately.

On `verify`-style functions I agree they probably aren't needed. The only thing they'd be doing differently for descendants is making sure they're non-null objects, so for receiving foreign data (clipboard data, collaborator operations, etc) it would be a trivial additional check for the user, and probably want to live within some code that checks for user specific schema anyways. Editors on the other hand are full of functions, and wouldn't (or shouldn't) get passed around like hostile unknown values.

If we do want to just use the existing names just lmk and I can squash the commits to keep the git-blames cleaner

> Perhaps to make migration easier, we could keep Node.isNode's argument type as `any` [...]

(`isNode` uses `isEditor` et al under the hood so we'd just have to be sure to rewrite it if we do this.)

> Also, don't forget that if we do end up changing `slate-react` and other packages to depend on new function names, we'll also need to bump their peer dependencies on `slate` prior to publishing them.

Is this something I should edit manually in the PR or is it best done after things are merged? (either way I will hold off until after naming is settled)

packages/slate/src/interfaces/node.ts
nabbydude: Understandable, I'll wait until names are settled and then add it if these names stay

12joan:
> Is this something I should edit manually in the PR or is it best done after things are merged? (either way I will hold off until after naming is settled)

It should be fine to do it in this PR, but only when it's relatively close to being merged. Otherwise, if `slate`'s version is bumped by other PRs, the peer dependency in this PR would need to be updated again.

Alternatively, @dylans could commit the `peerDependencies` change directly to the version packages PR branch after this is merged, which is a bit safer since it's that PR which actually bumps the package versions, so there's no uncertainty as to which peer dependency version to use.

dylans:
@nabbydude please resolve the conflict and handle the version number change. I'll do a release with just this change so that there won't be any issues with the peerDependencies hopefully.

nabbydude:
@dylans Can do 🫡
Any thoughts on naming/replacing the existing functions?
- Replacing the existing would be ideal imo but could potentially break some code--any match functions would be fine as is, and most technically-type-unsafe things (like plate calling it with their `At` type, which is either a node or location), would also work just fine as is, but anything relying on it to gracefully handle nullish values for them could throw.
- `Node.isEditor` etc is probably my preferred alternative looking back after the fact and if the above option is off the table and there are no objections I'm going to change it to this

dylans:
> @dylans Can do 🫡 Any thoughts on naming/replacing the existing functions?
> 
>     * Replacing the existing would be ideal imo but could potentially break some code--any match functions would be fine as is, and most technically-type-unsafe things (like plate calling it with their `At` type, which is either a node or location), would also work just fine as is, but anything relying on it to gracefully handle nullish values for them could throw.
> 
>     * `Node.isEditor` etc is probably my preferred alternative looking back after the fact and if the above option is off the table and there are no objections I'm going to change it to this

I don't have a strong opinion so I would go with your preferred alternative. We'll break someone regardless of what we do. I guess as I type this out loud, I tend to prefer when the breaking changes are obvious rather than subtle if possible. 😂

nabbydude:
Switched to `Node.isX` since its terse and descriptive, and it doesn't risk any subtle breakages downstream. I'd love to deprecate+remove/rename the classic functions but I'll save that until I (or you, random bystander reading this right now 👁️) add some proper guidance in the docs in favor of these new functions.

Also added `Node.isAncestor` as an alias of `!Node.isText` as requested by @12joan

peerDependencies updated, should be all good to merge now 👍

12joan:
> I tend to prefer when the breaking changes are obvious rather than subtle if possible

@dylans I hadn't considered this before, but this is a pretty good rule for making changes.

PR #5986, opened 2025-12-09
Switch from JSX.Element to React.JSX.Element

ebroder:
**Description**

The global JSX namespace has been deprecated for around two and a half years now in favor of using React.JSX, and as of the type declarations for React 19, it's been fully removed. By switching to the new namespace, this package maintains compatibility with React 19.

Without this change, attempting to use React 19 with `slate-react` and TypeScript results in typechecking errors:

```
node_modules/slate-react/dist/components/text.d.ts:8:59 - error TS2503: Cannot find namespace 'JSX'.

8     renderPlaceholder: (props: RenderPlaceholderProps) => JSX.Element;
                                                            ~~~

node_modules/slate-react/dist/components/text.d.ts:9:47 - error TS2503: Cannot find namespace 'JSX'.

9     renderLeaf?: ((props: RenderLeafProps) => JSX.Element) | undefined;
                                                ~~~

node_modules/slate-react/dist/components/text.d.ts:10:47 - error TS2503: Cannot find namespace 'JSX'.

10     renderText?: ((props: RenderTextProps) => JSX.Element) | undefined;
                                                 ~~~
```

**Issue**

N/A

**Example**

**Context**

The `React.JSX` namespace was introduced and global `JSX` was deprecated both in DefinitelyTyped/DefinitelyTyped#64464, which was published as v18.2.6 of `@types/react` (n.b.: version numbers in `@types/` packages are only loosely connected to upstream version numbers).

Since TypeScript is structurally-typed, the types `React.JSX.Element` and `JSX.Element` should be functionally equivalent to the typechecker, so as long as users have a version of `@types/react` that includes both declarations, this should not cause a backwards compatibility issue.

Ideally, we would enforce that users had v18.2.6 or later of `@types/react`, but given the nature of type dependencies, that's not possible. (There's no way to have a non-dev dependency introduce new dev dependencies). But presumably anyone who is proactively updating this package is also updating their type declarations as well.

**Checks**

### DIFF
--- a/.changeset/bright-turtles-reply.md
+++ b/.changeset/bright-turtles-reply.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Update TypeScript declarations for React v19 by switching from JSX.Element to React.JSX.Element
--- a/docs/libraries/slate-react/editable.md
+++ b/docs/libraries/slate-react/editable.md
@@ -1,6 +1,6 @@
 # Editable component
 
-## `Editable(props: EditableProps): JSX.Element`
+## `Editable(props: EditableProps): React.JSX.Element`
 
 The `Editable` component is the main editing component. Note that it must be inside a `Slate` component.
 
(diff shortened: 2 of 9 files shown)


ebroder:
Is there anything I can do on my end to encourage merging and releasing this? AFAIK it's the last blocker for us to upgrade.

PR #5988, opened 2025-12-10
Fix rendering issues when both chunking and React's strict mode are enabled

12joan:
**Description**
If React's strict mode and the experimental chunking optimisation are both enabled, this caused rendering issues inside chunks that prevented `slate-react` from synchronising changes to the editor value with the DOM. This PR resolves these issues by adjusting the timing of when the chunk tree's `modifiedChunks` set is cleared.

**Example**
Recording of the bug as reported by z2devil on Slack:

> When the editor chunk is initialized in React StrictMode, text input inside this chunk stops working and the cursor disappears.

https://github.com/user-attachments/assets/a3f6db85-c7b6-4a61-997f-89d5e2219172

**Context**
The chunking optimisation works by iteratively mutating a nested "chunk tree" object as part of `useChildren` every time the parent element renders. On each iteration, the new children array is "reconciled" with the chunk tree, analogous to how React reconciles the DOM with its virtual DOM. This reconciliation can result in chunks being added, removed or modified.

Since each individual chunk in the chunk tree is mutable, React cannot automatically determine whether the component corresponding to each chunk needs to be re-rendered. For this reason, the chunk tree maintains a set of `modifiedChunks`, which are chunks that have been modified since the last render. If a given chunk appears in this set, we bypass React's memoisation and re-render the chunk's component.

Prior to this PR, this `modifiedChunks` set was cleared at the start of each iteration (i.e. at the start of each render), so that the set included only those chunks that were modified as part of the current iteration. However, this caused issues with React's strict mode, in which each component's render function is called twice prior to that component actually being rendered.

This meant that on the first render pass, `modifiedChunks` was correct but unused. On the second render pass (caused by strict mode), `modifiedChunks` had already been cleared, and the chunk tree had already been updated. This meant that `modifiedChunk` was empty when `React.memo` checked whether each chunk component needed to be re-rendered, resulting in the chunk component failing to update in response to changes to the editor value.

This PR fixes this by only clearing `modifiedChunks` inside a `useEffect`, by which time its value has already been used.

**Checks**

### DIFF
--- a/site/examples/js/huge-document.jsx
+++ b/site/examples/js/huge-document.jsx
@@ -1,5 +1,5 @@
 import { faker } from '@faker-js/faker'
-import React, { useCallback, useEffect, useState } from 'react'
+import React, { StrictMode, useCallback, useEffect, useState } from 'react'
 import { createEditor as slateCreateEditor, Editor } from 'slate'
 import { Editable, Slate, withReact, useSelected } from 'slate-react'
 
@@ -41,6 +41,7 @@ const initialConfig = {
     'chunk'
   ),
   showSelectedHeadings: parseBoolean('selected_headings', false),
+  strictMode: parseBoolean('strict', false),
 }
 const setSearchParams = config => {
   if (searchParams) {
@@ -54,6 +55,7 @@ const setSearchParams = config => {
       'selected_headings',
       config.showSelectedHeadings ? 'true' : 'false'
     )
+    searchParams.set('strict', config.strictMode ? 'true' : 'false')
     history.replaceState({}, '', `?${searchParams.toString()}`)
   }
 }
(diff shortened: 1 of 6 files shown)


site/examples/js/huge-document.jsx
```diff
@@ -41,6 +41,7 @@ const initialConfig = {
     'chunk'
   ),
   showSelectedHeadings: parseBoolean('selected_headings', false),
+  strictMode: parseBoolean('strict', false),
```
12joan, line 44: Adds a checkbox to the huge document example to enable strict mode for testing. Due to the nature of strict mode, this only works when running the examples locally.

image

site/examples/js/richtext.jsx
12joan, line 1: This file was changed as a result of `yarn tsc:examples`

PR #6004, opened 2026-01-21
fix(slate-dom): respect `suppressThrow` in `toSlatePoint` when `findPath` fails

christianhg:
When `toSlatePoint` is called with `suppressThrow: true`, internal calls to `findPath` should not cause errors to propagate. This is particularly important during component unmount when node references may become stale.

The throttled selection change handler in slate-react uses `toSlateRange` with `suppressThrow: true`, but if `findPath` throws during unmount, the error was not being caught, causing "Unable to find the path for Slate node" errors after the component unmounts.

This fix wraps the two `findPath` calls in `toSlatePoint` with try-catch blocks that respect the `suppressThrow` option.

**Checks**

### DIFF
--- a/packages/slate-dom/src/plugin/dom-editor.ts
+++ b/packages/slate-dom/src/plugin/dom-editor.ts
@@ -888,10 +888,16 @@ export const DOMEditor: DOMEditorInterface = {
 
       if (node && DOMEditor.hasDOMNode(editor, node, { editable: true })) {
         const slateNode = DOMEditor.toSlateNode(editor, node)
-        let { path, offset } = Editor.start(
-          editor,
-          DOMEditor.findPath(editor, slateNode)
-        )
+        let nodePath
+        try {
+          nodePath = DOMEditor.findPath(editor, slateNode)
+        } catch (e) {
+          if (suppressThrow) {
+            return null as T extends true ? Point | null : Point
+          }
+          throw e
+        }
+        let { path, offset } = Editor.start(editor, nodePath)
 
         if (!node.querySelector('[data-slate-leaf]')) {
           offset = nearestOffset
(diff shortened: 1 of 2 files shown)


packages/slate-dom/src/plugin/dom-editor.ts
```diff
@@ -888,10 +888,18 @@ export const DOMEditor: DOMEditorInterface = {
...
+        let nodePath
+        try {
+          nodePath = DOMEditor.findPath(editor, slateNode)
+        } catch {
+          if (suppressThrow) {
+            return null as T extends true ? Point | null : Point
+          }
+          throw new Error(
+            `Cannot resolve a Slate point from DOM point: ${domPoint}`
+          )
```
12joan: This will mask the real error thrown by `findPath`. Can we throw the original error instead?

packages/slate-dom/src/plugin/dom-editor.ts
```diff
@@ -914,7 +922,17 @@ export const DOMEditor: DOMEditorInterface = {
...
+    let path
+    try {
+      path = DOMEditor.findPath(editor, slateNode)
+    } catch {
+      if (suppressThrow) {
+        return null as T extends true ? Point | null : Point
+      }
+      throw new Error(
+        `Cannot resolve a Slate point from DOM point: ${domPoint}`
+      )
```
12joan: Same here

12joan (changes requested):

packages/slate-dom/src/plugin/dom-editor.ts
christianhg: Ah, yes! Fixed! Thank you 🙇

12joan (approved):
Looks good to me

christianhg:
@dylans , any chance we could get this fix in? Thanks in advance 🙇

PR #6006, opened 2026-01-27
Passing `force` property from `normalize` to `normalizeNode` method

skorenb:
It might be helpful when you have custom normalizations and you need to react accordingly to `editor.normalize({ force: true })`


dylans (approved):

dylans:
Note, I landed this, but for some reason our automated release workflow didn't include it in the version packages PR above. I don't have time to sort that out today, but if anyone does, I'll get this into a release soon.

Issue #6011, opened 2026-02-10
Bug: Stale `editor` in focus listener effect in `Slate` component


changlin-cn:
# Bug: Stale `editor` in focus listener effect in `Slate` component

## Description

In `packages/slate-react/src/components/slate.tsx`, the `useIsomorphicLayoutEffect` that subscribes to document `focusin`/`focusout` (or `focus`/`blur`) uses a callback that reads `editor`, but the effect has an **empty dependency array** `[]`. That causes a stale closure: the listener `fn` always refers to the `editor` from the first render. If the parent ever passes a different `editor` instance (e.g. after remount or when `editor` is recreated), the effect does not re-run, so the document-level listeners keep calling `ReactEditor.isFocused(editor)` with the **old** editor. That can lead to wrong `isFocused` state in context and, in edge cases, to using an unmounted or replaced editor.

## Code

```tsx
// slate.tsx, lines 94–114
useIsomorphicLayoutEffect(() => {
  const fn = () => setIsFocused(ReactEditor.isFocused(editor))
  // ... addEventListener('focusin', fn) etc.
  return () => { /* removeEventListener ... */ }
}, [])  // ← bug: `editor` is used inside but not in deps
```

## Expectation

- The effect should depend on `editor`, so when `editor` changes we remove the old listeners and attach new ones that use the current editor.
- Same pattern as the existing `useEffect` above (lines 88–90), which correctly has `[editor, onContextChange]`.

## Suggested fix

Include `editor` in the dependency array:

```tsx
}, [editor])
```

So the effect re-subscribes whenever the `editor` instance changes, and the listener always calls `ReactEditor.isFocused(editor)` with the current editor.

## Environment

- Slate Version: (current from repo)
- Any OS / browser where focus events are used

## Context

This is a dependency-array bug: the effect closes over `editor` but does not list it as a dependency. In typical usage the same `editor` instance is reused for the lifetime of the component, so the bug may not surface often; it appears when the `editor` prop changes (e.g. remount, conditional editor creation, or tests that swap editors).

PR #6012, opened 2026-02-10
Fix Slate component to properly handle editor updates by adding `edit…

changlin-cn:
**Description**
In packages/slate-react/src/components/slate.tsx, the useIsomorphicLayoutEffect that subscribes to document focusin/focusout (or focus/blur) uses a callback that reads editor, but the effect has an empty dependency array []. That causes a stale closure: the listener fn always refers to the editor from the first render. If the parent ever passes a different editor instance (e.g. after remount or when editor is recreated), the effect does not re-run, so the document-level listeners keep calling ReactEditor.isFocused(editor) with the old editor. That can lead to wrong isFocused state in context and, in edge cases, to using an unmounted or replaced editor.

**Issue**
Fixes: [Bug: Stale editor in focus listener effect in Slate component](https://github.com/ianstormtaylor/slate/issues/6011)

### DIFF
--- a/.changeset/modern-crabs-try.md
+++ b/.changeset/modern-crabs-try.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fix Slate component to properly handle editor updates by adding `editor` as a dependency in the useEffect hook.


12joan:
It looks like an accidental omission to me, especially since the `useEffect` above has no purpose other than to support the `editor` being replaced.

```ts
  const [isFocused, setIsFocused] = useState(ReactEditor.isFocused(editor))

  useEffect(() => {
    setIsFocused(ReactEditor.isFocused(editor))
  }, [editor])
```

jsauca:
Here is quick fix for me:


```
  // Force Slate remount when the editor instance changes
  // slate-react's focusin/focusout listener has [] deps, capturing the initial
  // editor forever — remounting resets the closure with the correct editor.
  const slateKeyRef = useRef(0);
  const prevEditorRef = useRef(editor);
  if (editor !== prevEditorRef.current) {
    prevEditorRef.current = editor;
    slateKeyRef.current += 1;
  }

<Slate
    key={slateKeyRef.current}
  />
```

changlin-cn:
> This may be on purpose, but I'm not remembering. Maybe @12joan remembers?
> 
> Also this will need a changeset to get released if it is needed.

If this is intentional, it will cause the useFocused hook to return an incorrect value after the editor instance is recreated. Additionally, the changeset has been updated.

Issue #5983, opened 2025-12-05
When the content is empty, the first voice input on Android repeats


mayingjie123:
**Description**
On Android devices, when using Baidu Input Method or Sogou Input Method for voice input, if there is no content in the slate input box, the first voice input will be repeated. It's normal if there is content in the input box (even if it's just one space)

Every time the voice input is "hello", but when the input box is empty, "hello hello" appears. It works normally when there is content in the input box (even if there is only one space)

**Recording**

https://github.com/user-attachments/assets/79088039-da75-41f4-9288-31643ec4d6b6

**Sandbox**
Reproducible on the official Rich Text example: https://www.slatejs.org/examples/richtext

**Steps**
1. On an Android device, open Chrome.
2. Go to the Slate Rich Text example link above.
3. Press and hold the space bar on the keyboard for voice input (e.g., Sogou).
4. It's normal to say "hello" when there is content in the input box
5. When there is no content in the input box and you say "hello", a bug occurs: "hello" is repeated

**Expectation**
The content input by long-pressing the voice button in the Android input method cannot be repeated

**Environment**
- Slate Version: 0.120.0
- OS: android 13
- Browser: Chrome
- Input Method: Sogou Input Method、Baidu Input Method

PR #6020, opened 2026-03-11
fix: prevent Android IME first character duplication on empty nodes

joyboyguo-sketch:
**Description**
Fixes a bug introduced with https://github.com/ianstormtaylor/slate/issues/5983 

When the cursor is positioned at an empty text node, Android IME treats the transition from empty to non-empty as a structural change and restarts composition, causing the first character to be duplicated.

Fix by inserting a zero-width space (U+200B) at the start of composition when the current text node is empty. This makes the node appear non-empty to the IME, preventing the composition restart. The placeholder is automatically cleaned up from the AST after compositionEnd, keeping the document state clean.

Fixes: Android IME first character duplication issue

**Issue**
Fixes: (link to issue)
https://github.com/ianstormtaylor/slate/issues/5983

### DIFF
--- a/packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
+++ b/packages/slate-react/src/hooks/android-input-manager/android-input-manager.ts
@@ -34,6 +34,9 @@ const RESOLVE_DELAY = 25
 // Time with no user interaction before the current user action is considered as done.
 const FLUSH_DELAY = 200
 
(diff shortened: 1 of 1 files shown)


tomdyqin:
good！it works for me！

12joan:
This fixes the issue for me on Gboard, but it's still present on Samsung Keyboard 3.3.23.33.

Issue #6023, opened 2026-03-20
Normalization can hang when merging text under a non-selectable element


juliankrispel:
**Description**
`Editor.normalize(editor, { force: true })` can hang indefinitely when normalization tries to merge adjacent text nodes inside a custom non-selectable element.

The minimal repro is a plain Slate editor with a custom `editor.isSelectable` override that returns `false` for `collapsible-content` and delegates otherwise. During normalization, Slate reaches the adjacent-text merge path, but `mergeNodes` looks up the previous node via cursor-style traversal. Because `Editor.positions` consults `editor.isSelectable`, traversal skips the non-selectable subtree and the merge fails to make progress.

**Recording**
Not applicable in the usual UI sense because this is a core-only programmatic hang. The repro is a direct `createEditor()` + `Editor.normalize()` call shown below.

**Sandbox**
No public sandbox yet. Minimal programmatic repro:

```ts
const editor = createEditor()
const originalIsSelectable = editor.isSelectable

editor.isSelectable = element =>
  element.type === 'collapsible-content'
    ? false
    : originalIsSelectable(element)

editor.children = [
  {
    type: 'paragraph',
    children: [{ text: 'Before the collapsible.' }],
  },
  {
    type: 'collapsible',
    children: [
      {
        type: 'collapsible-summary',
        children: [{ text: 'Summary' }],
      },
      {
        type: 'collapsible-content',
        children: [
          {
            type: 'paragraph',
            children: [{ text: 'is' }, { text: ' here' }],
          },
        ],
      },
    ],
  },
]

editor.selection = null
Editor.normalize(editor, { force: true })
```

**Steps**
1. Create a plain Slate editor.
2. Override `editor.isSelectable` so `collapsible-content` returns `false`.
3. Seed the editor with adjacent text nodes under that non-selectable element.
4. Call `Editor.normalize(editor, { force: true })`.
5. Observe that normalization does not return.

**Expectation**
Normalization should complete and merge the adjacent text nodes into a single text node (`"is here"`).

**Environment**
- Slate Version: current `main` as of Mar 20, 2026
- Operating System: macOS
- Browser: N/A
- TypeScript Version: not TypeScript-specific

**Context**
I traced this to the interaction between:
- `packages/slate/src/core/normalize-node.ts`
- `packages/slate/src/transforms-node/merge-nodes.ts`
- `packages/slate/src/editor/previous.ts`
- `packages/slate/src/editor/before.ts`
[...]

PR #6024, opened 2026-03-20
Fix normalization hanging under non-selectable elements

juliankrispel:
**Description**
Fix a Slate core regression where normalization can hang while merging adjacent text nodes under a non-selectable element.

**Issue**
Fixes: #6023

**Example**
Minimal repro:

```ts
const editor = createEditor()
const originalIsSelectable = editor.isSelectable

editor.isSelectable = element =>
  element.type === 'collapsible-content'
    ? false
    : originalIsSelectable(element)

editor.children = [
  {
    type: 'paragraph',
    children: [{ text: 'Before the collapsible.' }],
  },
  {
    type: 'collapsible',
    children: [
      {
        type: 'collapsible-summary',
        children: [{ text: 'Summary' }],
      },
      {
        type: 'collapsible-content',
        children: [
          {
            type: 'paragraph',
            children: [{ text: 'is' }, { text: ' here' }],
          },
        ],
      },
    ],
  },
]

editor.selection = null
Editor.normalize(editor, { force: true })
```

This PR also adds:
- a normalization regression test that reproduces the hang
- a narrower `Transforms.mergeNodes` test covering the structural merge path directly

**Context**
The regression comes from `Editor.positions` honoring `editor.isSelectable`, while `mergeNodes` was using cursor-style traversal to find the previous merge partner even in the default path-based merge case.

For structural merges, that traversal is the wrong abstraction: when `mergeNodes` is called with the default sibling match at a concrete path, it should use the direct previous sibling path instead of `Editor.previous`. This keeps normalization from skipping non-selectable subtrees while still preserving the `isSelectable` behavior in cursor movement APIs.

**Checks**

Targeted checks run locally:
- `yarn build:rollup`
- `yarn test:custom --grep "(merge-adjacent-non-selectable-ancestor|non-selectable-ancestor)"`
- `yarn test:custom --grep "text-across"`

### DIFF
--- a/packages/slate/src/transforms-node/merge-nodes.ts
+++ b/packages/slate/src/transforms-node/merge-nodes.ts
@@ -31,8 +31,12 @@ export const mergeNodes: NodeTransforms['mergeNodes'] = (
       return
     }
 
+    const isPathMerge = Location.isPath(at)
+    const pathAt = isPathMerge ? (at as Path) : null
+    const usesDefaultSiblingMatch = match == null && isPathMerge
+
     if (match == null) {
-      if (Location.isPath(at)) {
+      if (isPathMerge) {
         const [parent] = Editor.parent(editor, at)
         match = n => parent.children.includes(n)
       } else {
@@ -60,7 +64,13 @@ export const mergeNodes: NodeTransforms['mergeNodes'] = (
     }
 
     const [current] = Editor.nodes(editor, { at, match, voids, mode })
-    const prev = Editor.previous(editor, { at, match, voids, mode })
+    const previousPath =
+      usesDefaultSiblingMatch && pathAt && Path.hasPrevious(pathAt)
+        ? Path.previous(pathAt)
+        : null
+    const prev = previousPath
+      ? Editor.node(editor, previousPath)
+      : Editor.previous(editor, { at, match, voids, mode })
 
     if (!current || !prev) {
       return
--- a/packages/slate/test/normalization/text/merge-adjacent-non-selectable-ancestor.ts
+++ b/packages/slate/test/normalization/text/merge-adjacent-non-selectable-ancestor.ts
@@ -1,0 +1,74 @@
+import { createEditor } from 'slate'
+
+export const input = createEditor() as any
+
+const { isSelectable, shouldNormalize } = input
+
+input.isSelectable = (element: any) => {
+  return element.type === 'collapsible-content' ? false : isSelectable(element)
+}
+
+input.shouldNormalize = (options: any) => {
+  if (options.iteration > 20) {
+    throw new Error(
+      'Normalization likely stalled while merging text under a non-selectable element.'
+    )
+  }
+
+  return shouldNormalize(options)
+}
+
+input.children = [
+  {
+    type: 'paragraph',
+    children: [{ text: 'Before the collapsible.' }],
+  },
+  {
+    type: 'collapsible',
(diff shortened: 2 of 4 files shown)


packages/slate/src/transforms-node/merge-nodes.ts
```diff
@@ -31,8 +31,12 @@ export const mergeNodes: NodeTransforms['mergeNodes'] = (
       return
     }
 
+    const isPathMerge = Location.isPath(at)
+    const pathAt = isPathMerge ? (at as Path) : null
+    const usesDefaultSiblingMatch = match == null && isPathMerge
```
juliankrispel, line 36: Why this setup exists: apps can override `editor.isSelectable` to mark specific element types as non-selectable. In this PR, `collapsible-content` is one of those app-defined non-selectable containers. This branch only changes the merge behavior for the simple case where `mergeNodes` is given a concrete path and is merging with the default sibling match.

packages/slate/src/transforms-node/merge-nodes.ts
```diff
@@ -60,7 +64,13 @@ export const mergeNodes: NodeTransforms['mergeNodes'] = (
     }
 
     const [current] = Editor.nodes(editor, { at, match, voids, mode })
-    const prev = Editor.previous(editor, { at, match, voids, mode })
+    const previousPath =
+      usesDefaultSiblingMatch && pathAt && Path.hasPrevious(pathAt)
+        ? Path.previous(pathAt)
+        : null
+    const prev = previousPath
```
juliankrispel, line 71: This is the actual fix. Before, `mergeNodes` always looked for the previous merge partner through `Editor.previous`, which uses selection-style traversal. That traversal skips nodes that the app marks as non-selectable, so the merge could fail to find the real previous sibling. Here we use the direct previous sibling path in the default path-based merge case, so structural merges do not depend on selection traversal.

packages/slate/test/normalization/text/merge-adjacent-non-selectable-ancestor.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const input = createEditor() as any
+
+const { isSelectable, shouldNormalize } = input
+
+input.isSelectable = (element: any) => {
+  return element.type === 'collapsible-content' ? false : isSelectable(element)
+}
+
+input.shouldNormalize = (options: any) => {
```
juliankrispel, line 11: This test reproduces the bug through normalization. The key detail is that the editor is customized so `collapsible-content` is non-selectable, meaning the app wants selection traversal to skip that container. The guard on `shouldNormalize` is only there to turn an infinite loop into a fast test failure with a clear error.

packages/slate/test/transforms/mergeNodes/path/non-selectable-ancestor.ts
```diff
@@ -0,0 +1,70 @@
...
+            },
+          ],
+        },
+      ],
+    },
+  ]
+
+  editor.selection = null
+
+  Transforms.mergeNodes(editor, { at: [1, 1, 0, 1], voids: true })
```
juliankrispel, line 41: This test checks the same problem one layer lower, directly at `Transforms.mergeNodes`. It shows that even when an app marks `collapsible-content` as non-selectable, a direct sibling text merge should still work, because this is a structural tree operation, not a caret-movement operation.

PR #6039, opened 2026-03-31
perf: add a batch-aware apply engine

### LINKED ISSUE #6038: Perf: repeated tree updates need a batch-aware apply engine
# Repeated tree updates need a batch-aware apply engine

## Summary

While benchmarking Plate against Slate on huge documents, I ran into a large transform cliff in Slate for repeated exact-path node updates.

The original trigger on the Plate side was fixed separately, but the Slate benchmark showed a more general problem:

- repeated exact-path updates on wide sibling arrays are dominated by repeated immutable branch rewrites
- `normalize` is not the main cost in that workload
- path/range ref transforms are not the main cost either
- the main cost is redoing shared ancestor work over and over in the per-op apply path

That turned this from a narrow `setNodes` observation into a broader batch-engine problem.

## Requirements

Any upstream solution needs to preserve a few important semantics:

- `editor.apply(op)` remains the single public/plugin seam
- plugin authors should not need to learn a second override path just to stay correct in batches
- code that calls downstream `apply(op)` and then immediately inspects `editor.children` should still see correct state
- previously published node references should remain immutable

## Current direction

The direction implemented in PR #6039 is:

- `Editor.withBatch(editor, fn)` as an explicit transaction boundary
- `Transforms.applyBatch(editor, ops)` as sugar over batched execution
- a private batch draft for tree changes
- accessor-backed `editor.children`, so committed state and staged state can be separated cleanly
- an optimized exact-path `set_node` path inside the batch executor
- generic batched semantics for other tree operations, even where they are not yet highly optimized

This keeps one public seam (`editor.apply(op)`) while moving batching into the engine underneath it.

## Benchmarks

Using:

```bash
yarn node -r ./config/babel/register.cjs ./packages/slate/test/perf/set-nodes-bench.js --blocks=5000 --group-size=50 --repeats=3
```

Current branch results:

- flat exact `set_node`
  - `Transforms.applyBatch(...)`: `6.03 ms`
  - `editor.apply(set_node)` loop inside `Editor.withoutNormalizing(...)`: `95.86 ms`
  - `Transforms.setNodes(...)` inside `Editor.withoutNormalizing(...)`: `92.84 ms`
  - `Transforms.setNodes(...)` with full per-call normalize: `1311.22 ms`
- grouped exact `set_node`
  - `Transforms.applyBatch(...)`: `6.88 ms`
  - `editor.apply(set_node)` loop inside `Editor.withoutNormalizing(...)`: `13.26 ms`
  - `Transforms.setNodes(...)`: `42.21 ms`
- mixed exact `set_node` batch plus one tail `insert_node`: `9.91 ms`
- pure `insert_node` batch on an empty document
  - `Transforms.applyBatch(...)`: `3152.11 ms`
  - replay-ish `editor.apply(insert_node)` loop inside `Editor.withoutNormalizing(...)`: `3709.12 ms`

So the batch engine now solves the original repeated exact-path `set_node` cliff and provides the right execution model for broader batching, but further op-family optimizations still need to be justified by benchmarks.

## Why I’m rewriting this issue

The original issue body focused on a `set_node` microbenchmark and possible exact-path bulk-update helpers.

At this point the underlying conclusion is clearer:

- the root problem is the per-op apply engine on repeated tree updates
- the durable solution is a batch-aware engine, not a public one-off `setNodesBatch` API
- exact-path `set_node` is the first optimized family, not the entire end state

## Tracking

- Draft implementation: #6039

zbeyens:
Fixes #6038

## Summary

This PR adds an explicit batch-aware apply engine:

- `editor.apply(op)` stays the single-operation/plugin seam.
- `Editor.withBatch(editor, fn)` adds an explicit transaction boundary.
- `Transforms.applyBatch(editor, ops)` batches ordinary operations and flushes once.
- `editor.children` is now accessor-backed so batching can stage private draft state without mutating the committed snapshot.
- exact-path `set_node` batches use an optimized path directly inside the batch executor.

Current benchmark highlight:
- flat exact `set_node`: `7.08 ms` for `applyBatch`, about `11.3x` faster than the per-op `editor.apply(...)` loop and about `556x` faster than fully normalized `Transforms.setNodes(...)`.
- grouped exact `set_node`: `8.84 ms` for `applyBatch`, about `1.4x` faster than the per-op `editor.apply(...)` loop and about `13.7x` faster than grouped `Transforms.setNodes(...)`.
- empty-doc `insert_node`: `14.62 ms` for `applyBatch`, about `200x` faster than the per-op `editor.apply(...)` loop.

## Semantics

The rewrite keeps the constraints that came up in issue discussion:

- plugin authors still override one thing: `editor.apply(op)`
- `applyBatch` still routes through `editor.apply(op)` overrides
- read-after-downstream-apply behavior is preserved for custom wrappers
- previously published node references stay immutable during batched exact-path updates
- mixed batches promote staged exact `set_node` work into generic draft children instead of committing early

## Supporting Changes

A few diffs outside the batch executor are still part of the batching feature and are intentional:

- `packages/slate-history/src/with-history.ts`
  - batching can stage operations and can also invalidate prior history when `editor.children` is replaced directly, so history now tracks the post-batch selection and clears stale undo/redo state after direct tree replacement.
  - this keeps undo/redo replay equivalent to the non-batched path; it is not a separate history feature.
- `packages/slate/src/core/normalize-node.ts`
  - batching can flush a single structural operation with explicit provenance, so normalization now takes an operation-aware fast path for direct-child block validation.
  - the old full scan remains the fallback for forced normalization or multi-op cases, so legacy behavior is preserved.
- `packages/slate/src/interfaces/transforms/general.ts`
  - `Transforms.applyBatch(...)` is the public entry point for the feature.
  - the same-parent `move_node` branch keeps batched child-array updates replay-equivalent to the existing transform semantics instead of doing unnecessary remove/insert churn.
- `packages/slate/src/create-editor.ts`
  - the accessor-backed `children` storage has to be installed at editor construction time so batches can stage draft children without mutating the committed tree.
  - the initial `editor.children = []` assignment is only seeding that backing store on a fresh editor instance.
- `packages/slate-dom/src/plugin/dom-editor.ts`
  - `focus()` may synthesize a selection when one is missing; under batching that temporary `set_selection` can otherwise trip the pending-op retry path and leave the DOM selection behind Slate's selection.
  - the DOM selection sync now runs after the focus check instead of only inside `root.activeElement !== el`, which fixes the stale-caret case where the editor is already focused but `focus()` just created or changed the Slate selection.
  - this is still batching support, not a separate focus feature: without it, batching can leave Slate selection and browser selection out of sync.
- `packages/slate-dom/src/plugin/with-dom.ts`
  - batching can leave Android/input bookkeeping stale after text ops or direct `editor.children = ...` replacement, so the DOM plugin now clears pending diffs, pending selection/action state, and dirty node-map state on those paths.
  - this is there to avoid replaying stale DOM-side state across a batched tree rewrite, not to add a separate DOM feature.
- `packages/slate/src/editor/normalize.ts`, `packages/slate/src/editor/add-mark.ts`, and `packages/slate/src/editor/remove-mark.ts`
  - normalization requests and mark-only `onChange` flushes now route through the shared batch scheduler so they do not fire in the middle of a batch.
  - outside a batch the behavior is unchanged; inside a batch they are deferred to the single final flush.
- `packages/slate/src/interfaces/editor.ts` and `packages/slate/src/core/index.ts`
  - `Editor.withBatch(...)` and `Transforms.applyBatch(...)` are the intended public API surface for this feature, and batching now threads `initialDirtyPathsLength` through `shouldNormalize(...)` because dirty paths may be accumulated before the final flush.
  - `wrapApply` and `isWritingBatchInternals` are re-exported from `slate` so sibling packages can integrate with batching through the package surface instead of reaching into `slate/src/...`.
- `packages/slate/src/transforms-node/merge-nodes.ts`
  - batching/path-based transforms need an explicit-path merge path so replay merges the requested node with its previous sibling directly instead of re-resolving through a fresh iterator pass.
  - the default sibling-match behavior is unchanged; this just keeps the path-targeted case replay-equivalent under batching.

None of the above are meant as standalone improvements. They are the supporting changes needed to keep batching replay-equivalent to `main` instead of introducing separate runtime semantics.

## Benchmarks

Using:

```bash
yarn node -r ./config/babel/register.cjs ./packages/slate/test/perf/set-nodes-bench.js --blocks=5000 --group-size=50 --repeats=3 --include=setnodes-flat,setnodes-flat-no-normalize,apply-flat-no-normalize,apply-batch-flat,setnodes-grouped,apply-grouped-no-normalize,apply-batch-grouped,apply-batch-flat-mixed-insert-tail,apply-insert-empty-no-normalize,apply-batch-insert-empty
```

Results on the current branch:

- flat exact `set_node`
  - `Transforms.setNodes(...)`: `3939.20 ms`
  - `Transforms.setNodes(...)` inside `Editor.withoutNormalizing(...)`: `77.21 ms`
  - `editor.apply(set_node)` loop inside `Editor.withoutNormalizing(...)`: `79.82 ms`
  - `Transforms.applyBatch(...)`: `7.08 ms`
  - speedup vs per-op `editor.apply(...)`: about `11.3x`
  - speedup vs `setNodes(...)` inside `withoutNormalizing(...)`: about `10.9x`
  - speedup vs fully normalized `setNodes(...)`: about `556x`
- grouped exact `set_node`
  - `Transforms.setNodes(...)`: `121.35 ms`
  - `editor.apply(set_node)` loop inside `Editor.withoutNormalizing(...)`: `12.36 ms`
  - `Transforms.applyBatch(...)`: `8.84 ms`
  - speedup vs per-op `editor.apply(...)`: about `1.4x`
  - speedup vs grouped `Transforms.setNodes(...)`: about `13.7x`
- mixed exact `set_node` batch plus one tail `insert_node`
  - `Transforms.applyBatch(...)`: `8.69 ms`
  - overhead vs the flat exact-only batch: about `1.23x` total, so the exact-path fast path still mostly holds when one simple structural op is appended
- pure `insert_node` batch on an empty document
  - `editor.apply(insert_node)` loop inside `Editor.withoutNormalizing(...)`: `2929.88 ms`
  - `Transforms.applyBatch(...)`: `14.62 ms`
  - speedup vs per-op `editor.apply(...)`: about `200x`

So the batch engine is no longer just about exact-path `set_node`: that remains the clearest targeted fast path, but the current empty-document `insert_node` path is also dramatically faster than replaying per-op inserts.

## Coverage

Added/updated coverage includes:

- batch lifecycle and semantics in `packages/slate/test/with-batch.js`
- exact-path `set_node` batch semantics in `packages/slate/test/apply-batch-exact-set-node.js`
- generic tree-op replay equivalence in `packages/slate/test/apply-batch-generic-tree-ops.js`
- committed-children accessor semantics in `packages/slate/test/children-accessor.js`
- history integration in `packages/slate-history/test/apply-batch-exact-set-node.js`
- React integration in `packages/slate-react/test/apply-batch.spec.tsx`

I also tightened the Mocha entrypoint so CI no longer globs helper `.ts/.tsx` files as runnable tests under Node 22.

## Verification

Ran locally on Node 22:

- `yarn build`
- `yarn lint:prettier`
- `yarn lint:eslint`
- `yarn lint:typescript`
- `yarn test`

I also ran the integration-local Playwright lane long enough to confirm browser install/startup, but I did not use local macOS integration output as the gating signal for this PR because the CI job runs on Linux and the first run was noisy.


packages/slate-dom/src/plugin/with-dom.ts
```diff
@@ -119,7 +120,7 @@ export const withDOM = <T extends BaseEditor>(
 
   // This attempts to reset the NODE_TO_KEY entry to the correct value
   // as apply() changes the object reference and hence invalidates the NODE_TO_KEY entry
-  e.apply = (op: Operation) => {
+  wrapApply(e, apply => (op: Operation) => {
```
12joan: Can we do this using traditional overrides instead of a `wrapApply` function in non-batching-related code?

packages/slate-react/test/apply-batch.spec.tsx
```diff
@@ -0,0 +1,40 @@
+import { createEditor, Transforms, wrapApply } from 'slate'
+import { withReact } from '../src'
+
+describe('withReact', () => {
```
12joan: I'm not sure why `withReact` would need new tests or behaviour, since as a plugin it should ideally have no knowledge of batching.

packages/slate/src/core/apply.ts
```diff
@@ -1,36 +1,81 @@
...
 
-export const apply: WithEditorFirstArg<Editor['apply']> = (editor, op) => {
+export const transformOperationPathRefs = (
+  editor: Editor,
+  op: Parameters<Editor['apply']>[0]
+) => {
   for (const ref of Editor.pathRefs(editor)) {
     PathRef.transform(ref, op)
   }
+}
```
12joan: I think this would be easier to read if the internal functions weren't exported. Then it would be clearer what the intended entrypoints are. Limiting exports also keeps the public API of the module smaller, which makes refactoring easier.

Also, is it possible to put these helper functions at the bottom?

packages/slate/src/core/apply.ts
```diff
@@ -1,36 +1,81 @@
...
 
+export const transformOperationPointRefs = (
+  editor: Editor,
+  op: Parameters<Editor['apply']>[0]
```
12joan:
```suggestion
  op: Operation
```

packages/slate/src/core/apply.ts
```diff
@@ -1,36 +1,81 @@
...
     updateDirtyPaths(editor, editor.getDirtyPaths(op), transform)
   }
+}
 
+export const transformOperationTree = (
+  editor: Editor,
+  op: Parameters<Editor['apply']>[0]
+) => {
   Transforms.transform(editor, op)
+}
```
12joan: Does this really need to be a function?

packages/slate/src/core/apply.ts
```diff
@@ -40,14 +85,62 @@ export const apply: WithEditorFirstArg<Editor['apply']> = (editor, op) => {
   if (op.type === 'set_selection') {
     editor.marks = null
   }
+}
+
+export const applyOperationPhases = (
```
12joan: Is there a better name we can use for this function than applying the operation phases?

packages/slate/src/editor/add-mark.ts
```diff
@@ -43,9 +43,7 @@ export const addMark: EditorInterface['addMark'] = (editor, key, value) => {
       }
 
       editor.marks = marks
-      if (!FLUSHING.get(editor)) {
-        editor.onChange()
-      }
+      scheduleOnChange(editor)
```
12joan, line 46: What is the impact of delaying the `onChange` call here?

packages/slate/src/editor/with-batch.ts
```diff
@@ -0,0 +1,6 @@
+import { withBatch as runWithBatch } from '../core/batch'
+import { EditorInterface } from '../interfaces/editor'
+
+export const withBatch: EditorInterface['withBatch'] = (editor, fn) => {
+  runWithBatch(editor, fn)
+}
```
12joan: What's the purpose of wrapping `withBatch` from `../core/batch` in a function with the same name and arguments?

packages/slate/src/create-editor.ts
```diff
@@ -184,7 +185,10 @@ export const createEditor = (): Editor => {
     wrapNodes: (...args) => wrapNodes(editor, ...args),
     shouldMergeNodesRemovePrevNode: (...args) =>
       shouldMergeNodesRemovePrevNode(editor, ...args),
-  }
+  } as Editor
```
12joan: Is there a way to avoid this type assertion?

packages/slate/test/perf/set-nodes-bench.js
12joan, line 1: Did you mean to use plain `js` files for some of these?

package.json
```diff
@@ -34,7 +34,7 @@
     "test:integration": "playwright install --with-deps && run-p -r serve playwright",
     "test:integration-local": "playwright install && run-p -r serve playwright",
     "test:integration-docker": "./playwright/docker/run-tests.sh",
-    "test:mocha": "mocha --require ./config/babel/register.cjs ./packages/{slate,slate-history,slate-hyperscript}/test/**/*.{js,ts}",
+    "test:mocha": "mocha --require ./config/babel/register.cjs ./packages/slate/test/*.js ./packages/slate-history/test/**/*.js ./packages/slate-hyperscript/test/index.js",
```
12joan, line 38: What's the purpose of this change?

packages/slate-dom/src/plugin/with-dom.ts
zbeyens: > Can we do this using traditional overrides instead of a `wrapApply` function in non-batching-related code?

Done. `withDOM` and `withReact` both use plain `editor.apply` overrides now; `wrapApply` is back to batching/tests/perf-only use.

packages/slate-react/test/apply-batch.spec.tsx
zbeyens: $> I'm not sure why  would need new tests or behaviour, since as a plugin it should ideally have no knowledge of batching.\n\nAgreed. I dropped the React-only batch spec and kept the coverage in the wrapper-stack matrix instead.

packages/slate/src/core/apply.ts
zbeyens: $> I think this would be easier to read if the internal functions weren't exported.\n\nDone. The phase helpers moved into , and  only re-exports the real public entrypoint now.

packages/slate/src/core/apply.ts
zbeyens: > suggestion `op: Operation`

Took that cleanup with the extraction. The phase helpers use `Operation` directly now.

packages/slate/src/core/apply.ts
zbeyens: $> Does this really need to be a function?\n\nI kept the phase boundary, but moved it out of  into . That keeps the dispatcher smaller while still giving the perf harness a stable hook.

packages/slate/src/core/apply.ts
zbeyens: $> Is there a better name we can use for this function than applying the operation phases?\n\nYep.  is gone; the extracted helper is just  now.

packages/slate/src/editor/add-mark.ts
zbeyens: $> What is the impact of delaying the  call here?\n\nIt just makes mark-only cursor changes use the same scheduled flush path as the rest of batching. Outside a batch it still flushes on the next microtask.

packages/slate/src/editor/with-batch.ts
zbeyens: $> What's the purpose of wrapping  from  in a function with the same name and arguments?\n\nThere wasn't a good one.  is a straight re-export now.

packages/slate/test/perf/set-nodes-bench.js
zbeyens: > Did you mean to use plain `js` files for some of these?

Yes. The batch matrix and perf helpers are plain `.js` so Mocha only runs real JS entrypoints instead of helper TS/TSX files.

package.json
zbeyens: > What's the purpose of this change?

It narrows `test:mocha` to the real `.js` test entrypoints. That fixes the Node 22/Mocha issue where helper `.ts` and `.tsx` files were being treated like runnable tests.

docs/api/nodes/editor.md
```diff
@@ -349,6 +349,10 @@ Normalize any dirty objects in the editor.
 
 Options: `{force?: boolean; operation?: Operation}`
 
+#### `Editor.withBatch(editor: Editor, fn: () => void) => void`
+
+Call a function, deferring normalization and `onChange` flush until it completes.
+
```
12joan, line 361: This description and the one for `Transforms.applyBatch` seem incomplete to me, since deferring normalizaiton and `onChange` can both also be achieved using `withoutNormalizing`.

packages/slate/src/interfaces/transforms/general.ts
```diff
@@ -113,6 +123,23 @@ export const GeneralTransforms: GeneralTransforms = {
 
         const node = Node.get(editor, path)
 
+        const truePath = Path.transform(path, op)!
+
+        if (Path.equals(Path.parent(path), Path.parent(truePath))) {
```
12joan, line 142:
```suggestion
        // PERF: If we're moving a node to a different position in the same
        // parent, do so using a single modifyChildren.
        if (Path.equals(Path.parent(path), Path.parent(truePath))) {
```

packages/slate-dom/src/plugin/with-dom.ts
```diff
@@ -204,6 +205,10 @@ export const withDOM = <T extends BaseEditor>(
 
     apply(op)
 
+    if (op.type === 'insert_text' || op.type === 'remove_text') {
+      EDITOR_TO_PENDING_DIFFS.delete(e)
+    }
```
12joan, line 211: Can you add a brief comment explaining what this is doing?

packages/slate-history/src/with-history.ts
```diff
@@ -14,15 +21,26 @@ import { HistoryEditor } from './history-editor'
...
+    e.history.redos = []
+  }
 
   e.redo = () => {
     const { history } = e
     const { redos } = history
 
     if (redos.length > 0) {
-      const batch = redos[redos.length - 1]
+      const batch = redos[redos.length - 1] as Batch
```
12joan: Why are we adding type casts here when `undos` and `redos` are already `Batch[]`?

packages/slate-history/src/with-history.ts
```diff
@@ -66,14 +87,19 @@ export const withHistory = <T extends Editor>(editor: T) => {
     }
   }
 
-  e.apply = (op: Operation) => {
+  wrapApply(e, apply => (op: Operation) => {
     const { operations, history } = e
     const { undos } = history
-    const lastBatch = undos[undos.length - 1]
+    const lastBatch =
+      undos.length > 0 ? (undos[undos.length - 1] as Batch) : undefined
```
12joan: What is the `undos.length > 0` check for?

packages/slate-history/src/with-history.ts
```diff
@@ -112,10 +145,56 @@ export const withHistory = <T extends Editor>(editor: T) => {
...
+    }
+  })
+
+  const childrenDescriptor = Object.getOwnPropertyDescriptor(e, 'children')
+
+  if (childrenDescriptor?.get && childrenDescriptor?.set) {
+    const getChildren = childrenDescriptor.get
+    const setChildren = childrenDescriptor.set
+
+    Object.defineProperty(e, 'children', {
```
12joan: I think a neater way of doing this would be to have one central `defineProperty` call that links `editor.children` to `editor.getChildren` and `editor.setChildren` methods, which can be overridden in the traditional manner.

packages/slate-history/src/with-history.ts
```diff
@@ -154,10 +233,45 @@ const shouldMerge = (op: Operation, prev: Operation | undefined): boolean => {
...
  */
 
-const shouldSave = (op: Operation, prev: Operation | undefined): boolean => {
+const shouldSave = (op: Operation, _prev: Operation | undefined): boolean => {
   if (op.type === 'set_selection') {
     return false
   }
 
   return true
 }
```
12joan: Might as well simplify this to:
```suggestion
const shouldSave = (op: Operation) => op.type !== 'set_selection'
```

packages/slate-history/src/with-history.ts
```diff
@@ -154,10 +233,45 @@ const shouldMerge = (op: Operation, prev: Operation | undefined): boolean => {
...
+  if (selection == null) {
+    if (editor.selection != null) {
+      Transforms.deselect(editor)
+    }
+
+    return
+  }
+
+  Transforms.select(editor, selection)
+}
```
12joan, line 234: I would remove the `undefined` type for `selection`, since it doesn't make sense to restore an `undefined` selection, and this just complicates the valid case of setting a `null` selection. You're already checking for `undefined` in the `restoreSelection(e, batch.selectionAfter)` case.

Then this can be refactored to:

```suggestion
const restoreSelection = (
  editor: Editor,
  selection: Editor['selection']
): void => {
  if (selection) {
    Transforms.select(editor, selection)
  } else if (editor.selection) {
    Transforms.deselect(editor)
  }
}
```

packages/slate-history/src/with-history.ts
```diff
@@ -154,10 +233,45 @@ const shouldMerge = (op: Operation, prev: Operation | undefined): boolean => {
...
+  if (Object.prototype.hasOwnProperty.call(batch, 'selectionAfter')) {
+    batch.selectionAfter = selection
+    return
+  }
+
+  Object.defineProperty(batch, 'selectionAfter', {
+    value: selection,
+    enumerable: false,
+    writable: true,
+  })
```
12joan: Why not just set `batch.selectionAfter` in the traditional manner?

I don't think a `hasOwnProperty` check is necessary here, but if it is you can use `Object.hasOwn` instead, which doesn't require reaching into the prototype and is safer than calling `hasOwnProperty` on the object.

packages/slate-react/test/react-editor.spec.tsx
```diff
@@ -29,16 +29,7 @@ describe('slate-react', () => {
...
-
-        await act(async () => {
-          const windowSelection = ReactEditor.getWindow(editor).getSelection()
-          expect(windowSelection?.focusNode?.textContent).toBe('test')
-          expect(windowSelection?.anchorNode?.textContent).toBe('test')
-          expect(windowSelection?.anchorOffset).toBe(
-            testSelection.anchor.offset
-          )
-          expect(windowSelection?.focusOffset).toBe(testSelection.focus.offset)
-        })
```
12joan, line 41: Why are we removing this part of the test?

packages/slate/src/core/batching/direct-text-merge-batch-children.ts
```diff
@@ -0,0 +1,57 @@
...
+) => {
+  if (ops.length === 0) {
+    return children
+  }
+
+  const nextChildren = children.slice()
+  let hasChanges = false
+
+  for (const op of ops) {
+    const [parentIndex, textIndex] = op.path
```
12joan, line 23: Can you add a `typeof` check to make sure these are numbers? Otherwise malicious paths might cause unexpected behaviour.

packages/slate/src/core/batching/direct-text-split-batch-children.ts
```diff
@@ -0,0 +1,60 @@
...
+) => {
+  if (ops.length === 0) {
+    return children
+  }
+
+  const nextChildren = children.slice()
+  let hasChanges = false
+
+  for (const op of ops) {
+    const [parentIndex, textIndex] = op.path
```
12joan, line 24: Same here, and anywhere else you're using data from operations as indices or keys.

packages/slate/src/core/batching/direct-text-split-batch-children.ts
```diff
@@ -0,0 +1,60 @@
...
+      children: replaceChildren(
+        parent.children,
+        textIndex,
+        1,
+        {
+          ...node,
+          text: before,
+        },
+        {
+          ...(op.properties as Partial<Text>),
```
12joan: After https://github.com/ianstormtaylor/slate/pull/6043 is merged, please could you add equivalent logic here to prevent setting dangerous properties using batched `split_node` operations?

If you can think of a neat way to refactor out that logic between `set_selection`, `set_node`, `split_node` and the batched versions thereof, that would be great!

12joan (commented):
Partial review of the first 20 changed files (95 total).

docs/api/nodes/editor.md
zbeyens: > "This description and the one for `Transforms.applyBatch` seem incomplete..."

Addressed: clarified both docs to distinguish these batch APIs from `withoutNormalizing`. They now call out that batch state stays live across the whole callback/ops array, with the final normalize and `onChange` flush happening once at the end.

packages/slate/src/interfaces/transforms/general.ts
zbeyens: > "If we're moving a node to a different position in the same parent, do so using a single modifyChildren."

Addressed: added the PERF comment on that same-parent fast path.

packages/slate-dom/src/plugin/with-dom.ts
zbeyens: > "Can you add a brief comment explaining what this is doing?"

Addressed: added a comment explaining that direct `children` replacement invalidates DOM bookkeeping derived from the old tree, while internal batch writes stay inside the current batch lifecycle.

packages/slate-history/src/with-history.ts
zbeyens: > "one central `defineProperty` call that links `editor.children` to `editor.getChildren` and `editor.setChildren` methods"

Not addressing: I left this out on purpose. That would add a second children-access seam on top of the accessor-backed `editor.children` surface this PR is trying to keep central. Here the wrappers are small and only reset plugin-local state, so I’d rather not widen the editor API just to dedupe two setter interceptors in this change.

packages/slate-history/src/with-history.ts
zbeyens: > "I would remove the `undefined` type for `selection`..."

Addressed: `restoreSelection` now takes `Editor['selection']`, and the `undefined` case stays at the call site where `selectionAfter` is optional.

packages/slate/src/core/batching/direct-text-merge-batch-children.ts
zbeyens: > "Can you add a `typeof` check to make sure these are numbers?"

Addressed: added explicit numeric index guards for the direct text merge helper.

packages/slate/src/core/batching/direct-text-split-batch-children.ts
zbeyens: > "Same here, and anywhere else you're using data from operations as indices or keys."

Addressed: added numeric index/position guards to the direct text split helper, and folded the disallowed split-property guard in here too so the batched path matches the normal `split_node` semantics.

packages/slate-history/src/with-history.ts
zbeyens: > "Why are we adding type casts here when `undos` and `redos` are already `Batch[]`?"

Addressed: the current file imports the shared `Batch` type and reads `undos[undos.length - 1]` / `redos[redos.length - 1]` directly. Those casts are gone.

packages/slate-history/src/with-history.ts
zbeyens: > "What is the `undos.length > 0` check for?"

Addressed by explanation: I kept the empty-stack guard so `undo` / `redo` stay cheap no-ops when there is no saved batch, instead of reading the last entry first and branching later.

packages/slate-history/src/with-history.ts
zbeyens: > "Might as well simplify this to: `const shouldSave = (op: Operation) => op.type !== 'set_selection'`"

Addressed: `shouldSave` is now exactly that logic.

packages/slate-history/src/with-history.ts
zbeyens: > "Why not just set `batch.selectionAfter` in the traditional manner?"

Partially addressed: I switched the guard to `Object.hasOwn`. I kept the non-enumerable `defineProperty` write because `selectionAfter` is redo metadata, not part of the public batch shape I want showing up in enumerable history snapshots.

packages/slate-react/test/react-editor.spec.tsx
zbeyens: > "Why are we removing this part of the test?"

Addressed: restored the removed DOM-selection assertions in the focus test, and kept the extra `ReactEditor.isFocused(editor)` assertion.

packages/slate/src/core/batching/direct-text-split-batch-children.ts
zbeyens: > "After #6043 is merged, please could you add equivalent logic here to prevent setting dangerous properties using batched `split_node` operations?"

Addressed: the batched `split_node` path now rejects dangerous `then` setters, and the batch-only helpers now validate non-numeric path indexes too.

zbeyens: > "After #6043 is merged, please could you add equivalent logic here to prevent setting dangerous properties using batched `split_node` operations?"

Addressed: the batched `split_node` path now rejects dangerous `then` setters, and the batch-only helpers now validate non-numeric path indexes too.

packages/slate-history/src/with-history.ts
zbeyens: > "What is the `undos.length > 0` check for?"

Addressed by explanation: I kept the empty-stack guard so `undo` / `redo` stay cheap no-ops when there is no saved batch, instead of reading the last entry first and branching later.

packages/slate-history/src/with-history.ts
12joan: Can you add a comment explaining this, and anything else that someone wouldn't guess from reading the code?

packages/slate/src/core/batching/direct-text-split-batch-children.ts
12joan: I wouldn't mark this as resolved just yet, since the full logic from #6043 hasn't been included yet and we might forget otherwise.

docs/api/nodes/editor.md
```diff
@@ -349,6 +349,13 @@ Normalize any dirty objects in the editor.
 
 Options: `{force?: boolean; operation?: Operation}`
 
+#### `Editor.withBatch(editor: Editor, fn: () => void) => void`
+
+Call a synchronous function as one logical batch. Operations still go through
+`editor.apply` one at a time, but batch state stays live until the callback
+completes, and normalization plus the final `onChange` flush run once at the
+end.
```
12joan: Can we say something like this so that readers get a general understanding of how batching works and the benefits of it (enough so to make a decision on whether or not to use it)?

```suggestion
Call a synchronous function as one logical batch. During a batch, operations
are applied by deeply mutating a draft `editor.children` structure, which is
converted back to an immutable `editor.children` structure when the callback
completes, or when code tries to read `editor.children` during a batch. When
applying large numbers of similar operations synchronously, this can be more
performant than applying the operations outside a batch.
```

packages/slate-dom/src/plugin/with-dom.ts
```diff
@@ -234,6 +240,25 @@ export const withDOM = <T extends BaseEditor>(
     }
   }
 
+  wrapSetChildren(e, setChildren => children => {
```
12joan: I'm not fond of this `wrapApply`/`wrapSetChildren`/`wrapGetChildren` pattern. I don't think using symbols is necessary for getting and setting children, as it prevents code from using the idiomatic Slate pattern of overriding editor methods in the traditional way.

packages/slate-history/src/with-history.ts
```diff
@@ -112,10 +142,44 @@ export const withHistory = <T extends Editor>(editor: T) => {
...
+
+    if (
+      isWritingBatchInternals(e) ||
+      (operationsLengthBeforeApply !== undefined &&
+        e.operations.length === operationsLengthBeforeApply)
+    ) {
+      return
+    }
+
+    resetHistoryAfterChildrenAssignment()
```
12joan: Since this is only used here and the function logic is simple, I would move the assignments here and document using a comment rather than a function name.

packages/slate/src/core/batching/direct-text-merge-batch-children.ts
```diff
@@ -0,0 +1,63 @@
...
+  const nextChildren = children.slice()
+  let hasChanges = false
+
+  for (const op of ops) {
+    validateOperationPathIndexes(
+      op.path,
+      `Cannot apply batched merge_node operations at path [${op.path}]`
+    )
+
+    const [parentIndex, textIndex] = op.path
```
12joan, line 23: Can you add a `typeof` check to make sure these are numbers? Otherwise malicious paths might cause unexpected behaviour.

(Duplicate of [this comment](https://github.com/ianstormtaylor/slate/pull/6039/changes#r3031870609), but the auto-responder incorrectly marked this as resolved and I'm unable to unresolve.)

packages/slate/src/core/batching/direct-text-split-batch-children.ts
```diff
@@ -0,0 +1,91 @@
...
+  const nextChildren = children.slice()
+  let hasChanges = false
+
+  for (const op of ops) {
+    validateOperationPathIndexes(
+      op.path,
+      `Cannot apply batched split_node operations at path [${op.path}]`
+    )
+
+    const [parentIndex, textIndex] = op.path
```
12joan, line 24: Same here

packages/slate/src/core/batching/direct-text-split-batch-children.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    if (typeof op.position !== 'number') {
+      throw new Error(
+        `Cannot apply batched split_node operations at path [${op.path}] because the split position must be a number.`
+      )
+    }
```
12joan: This check isn't necessary for security reasons (at the moment) because `position` is never used directly as an array index or object key. I was referring to `parentIndex` and `textIndex` in my comment above.

packages/slate/src/core/batching/direct-text-split-batch-children.ts
```diff
@@ -0,0 +1,91 @@
...
+        `Cannot apply batched split_node operations at path [${op.path}] because the target is not a text node.`
+      )
+    }
+
+    const before = node.text.slice(0, op.position)
+    const after = node.text.slice(op.position)
+    const nextNode: Descendant = { text: after }
+    const mutableNextNode = nextNode as unknown as Record<string, unknown>
+
+    for (const key in op.properties) {
```
12joan, line 47: This will need to be updated with changes from #6043 once that PR is merged

packages/slate/src/core/batching/exact-set-node-children.ts
```diff
@@ -0,0 +1,160 @@
...
+    throw new Error('Cannot set properties on the root node!')
+  }
+
+  validateOperationPathIndexes(
+    op.path,
+    `Cannot apply batched set_node operations at path [${op.path}]`
+  )
+
+  for (const key in op.newProperties) {
+    if (key === 'children' || key === 'text') {
```
12joan: This needs updating with #6040

packages/slate/src/core/batching/executor.ts
```diff
@@ -0,0 +1,866 @@
...
+  BatchSegment,
+  batchNeedsSegmentPlanning,
+  isSameParentInsertBatch,
+  planOperationBatchSegments,
+  shouldPreferWholeBatchExecution,
+} from './planner'
+import { withInternalBatchReads } from '../batch'
+
+// The executor owns staged draft mutation and dirty-path batching once a
+// segment shape has already been chosen by planner.ts.
```
12joan: Transforms/queries might be a less confusing pair of terms than executor/planner, which seem to convey a level of agency to these modules. If they were classes in an OO code base, it might make more sense to call them `Executor` and `Planner`, although even then I would prefer more descriptive names.

packages/slate/src/core/batching/apply-operation-batch.ts
```diff
@@ -0,0 +1,866 @@
...
+      nextPath = Path.transform(nextPath, op)
+
+      if (!nextPath) {
+        return null
+      }
+    }
+  }
+
+  return nextPath
+}
```
12joan: If `Path.operationCanTransformPath` were included in `Path.transform`, this could be simplified to:

```suggestion
const transformPathThroughOps = (path: Path, ops: Operation[]) =>
  ops.reduce((path, op) => path && Path.transform(path, op), path);
```

packages/slate/src/core/batching/executor.ts
```diff
@@ -0,0 +1,866 @@
...
+        for (const op of ops) {
+          editor.apply(op)
+        }
+
+        updateDirtyPaths(editor, newDirtyPaths, transformDirtyPath)
+      })
+    },
+    () => {}
+  )
+}
```
12joan: Functions like `applyDirtyPathBatchedOperations` and `batchOperationDirtyPaths` that use positional arguments whose meaning can't be inferred without reading the argument name generally benefit from using an options object as an argument.

packages/slate/src/core/batching/executor.ts
```diff
@@ -0,0 +1,866 @@
...
+  for (const op of ops) {
+    const targetIndex = op.path[op.path.length - 1]
+
+    if (targetIndex > order.length) {
+      throw new Error(
+        `Cannot apply an "insert_node" operation at path [${op.path}] because the destination is past the end of the node.`
+      )
+    }
+
+    const token = `inserted:${insertedTokenIndex++}`
```
12joan: Why are we using these tokens to populate the `order` array rather than adding the nodes to `order` directly?

packages/slate/src/core/batching/executor.ts
```diff
@@ -0,0 +1,866 @@
...
+  }
+
+  for (const path of Path.levels(parentPath)) {
+    addDirtyPath(path)
+  }
+
+  order.forEach((token, position) => {
+    if (typeof token === 'number') {
+      return
+    }
```
12joan: I'm not keen on `number` and `string` having magic meanings here. I don't think we're ever using the actual value of a `number` array element on `order`, so maybe `null` is what we mean here?

12joan (commented):
Another partial review of the first 20 files

packages/slate/src/core/batching/direct-text-merge-batch-children.ts
zbeyens: I already applied this more broadly through validateOperationPathIndexes(...) and now use that in the direct text merge/split helpers, plus the other batch-only helpers that use operation-derived path segments as array indexes or keys.

packages/slate/src/core/batching/apply-operation-batch.ts
zbeyens: I kept this explicit. The suggested reduce(...) form depends on changing Path.transform so it absorbs Path.operationCanTransformPath, which would be a broader Path contract change outside this PR. I’d rather keep that guard local here unless we want to make that API change deliberately.

packages/slate/src/core/batching/executor.ts
zbeyens: I renamed them to batch-segments.ts and apply-operation-batch.ts to match Slate’s Operation terminology without the awkward operation-batch-segments name.

packages/slate/src/core/batching/executor.ts
zbeyens: I changed both `applyDirtyPathBatchedOperations` and `batchOperationDirtyPaths` to take an options object instead of positional arguments.

packages/slate/src/core/batching/direct-text-split-batch-children.ts
zbeyens: I removed that `position` check and kept the guards on `parentIndex` / `textIndex`.

packages/slate-dom/src/plugin/with-dom.ts
zbeyens: I removed the symbol-backed `wrapSetChildren` / `wrapGetChildren` and switched this back to the normal Slate pattern: one central `children` accessor calling ordinary `editor.getChildren()` / `editor.setChildren(children)` methods, with plugins overriding those methods directly.

packages/slate-history/src/with-history.ts
zbeyens: I inlined the reset at the call site and replaced the helper name with a comment explaining why direct tree replacement clears both history stacks.

packages/slate-history/src/with-history.ts
```diff
@@ -14,76 +22,91 @@ import { HistoryEditor } from './history-editor'
...
     }
 
     if (save) {
       if (merge == null) {
         if (lastBatch == null) {
           merge = false
-        } else if (operations.includes(lastOp)) {
+        } else if (
+          lastBatch.operations.length === 0 &&
+          operations.length !== 0
```
nabbydude: Unless I'm missing something this is a regression of #5866
```suggestion
          operations.includes(lastOp)
```

packages/slate/src/core/apply-operation.ts
```diff
@@ -0,0 +1,120 @@
...
+  editor,
+  fn,
+  applyDirtyPaths,
+}: {
+  editor: Editor
+  fn: () => void
+  applyDirtyPaths: () => void
+}) => {
+  batchDirtyPaths(editor, fn, applyDirtyPaths)
+}
```
nabbydude: this function imports a function from another file and just calls it with the same three parameters, and only exists to be called from a third file. I really dont think it needs to exist.

packages/slate/src/core/batching/apply-operation-batch.ts
```diff
@@ -0,0 +1,859 @@
...
+
+    if (nextIndex == null) {
+      return transformPathThroughOps(path, ops)
+    }
+
+    return parentPath.concat(nextIndex, ...path.slice(parentPath.length + 1))
+  }
+}
+
+const calculateDirtyPathsAfterBatch = (editor: Editor, ops: Operation[]) => {
```
nabbydude: unused, dupe of function of same name in another file

nabbydude:
To be brutally honest I struggled following the file structure and the functions that do nothing but call other functions, but after finding the third function that is name-and-body duplicated between multiple files I think I'm going to give up on a full review for now and just focus on what I can discern of the overall structure.

From what I can tell the strategy here is a hodgepodge of different bespoke optimizations:
- grouping consecutive `set_node` operations and resolving them via depth-first recursion, only rebuilding parent nodes after all its children have been processed
- doing the same for consecutive chunks of `insert_text` and `remove_text`.
  - It's unclear to me why this is separate to `set_node`, so that eg [set, insert, set, remove] can't be processed together.
- grouping consecutive `insert_node` operations that insert into the same parent using fenwick trees.
  - I'm nearly certain a simpler data structure could be used to this effect, and I'm somewhat skeptical that all this javascript outmaneuvering will outpace regular engine-optimized `splice` calls.
- grouping `split_node` calls but only for text nodes
  - and only at `path.length == 2`? not sure I'm reading this right but it seems like `applyDirectTextSplitBatchToChildren` only reads the first two indices of `op.path` and is only ever called with some variation of the root children array.
- ditto for `merge_node`
- `applyBatch` seems to split its list of ops into segments matching patterns and manually prefab new dirtyPaths + a dirtyPaths transformer to avoid calculating that all manually.
  - I know some of these patterns are seen in commonly-used builtin transforms but I think we could save that pattern-matching overhead by just calling `batchDirtyPaths` manually in transforms like `insertNodes` does.

Overall this really doesn't feel like a robust solution IMO, I might be missing something but it seems like it falls apart as soon as there are heterogeneous operations.

zbeyens:
Please be brutal here. Most of this was generated with Codex, so I’m treating review feedback as input. My read is that Slate’s current apply/normalize pipeline does not have a clean native place for this kind of transactional tree batching; a truly clean version probably wants a deeper engine rethink. So this PR is very much test/benchmark-driven.

I pushed a cleanup pass that directly addresses a few of these points:

- removed the Fenwick tree from same-parent `insert_node` batching
  - I replaced it with a much simpler copy + `splice` loop
  - the structural insert benchmarks stayed fast enough, so the Fenwick version was just not worth it
- pulled the child-array helper idea from #6050 into the normal non-batched helpers
  - this improved the normal `set_node` path quite a bit
- centralized the duplicated dirty-path replay helpers into `batching/dirty-paths.ts`
- cut `apply-operation-batch.ts` down substantially
- added `batching/README.md` to explain which paths are intentional fast paths and which are just correctness support
- narrowed the required benchmark contract so split/merge are not treated as core perf lanes anymore

On split/merge specifically: I tried removing those draft paths entirely. That broke replay-equivalence tests for cases like split/merge followed by later text ops under deferred normalization. So I agree they are not convincing as perf optimizations, but right now they seem to be carrying correctness for deferred normalization rather than just being optional speed paths. I documented that distinction.

On text ops: I also tried removing the separate text batching path. That regressed the `insert_text` batch benchmark from roughly `26ms` to roughly `3400ms`, so that one is unfortunately load-bearing.

The current intent is narrower than “optimize every op family”:

- exact `set_node` batches
- text leaf batches
- same-parent `insert_node`
- same-parent `insert_node + move_node`
- same-parent `move_node`

Everything else should either replay normally or only exist where tests show it is needed for replay-equivalent deferred normalization.

nabbydude:
have you compared control benchmarks to the main branch? it seems like non-batched behavior (specifically the `setnodes-flat` benchmark) is being slowed around 30x on my machine in this branch vs main unless my methodology is faulty

zbeyens:
Good catch. I pushed a fix for that.

Where `#6039` still seems to have real value is structural and mixed batches. I reran a focused matrix locally against current `#6039` and a `#6050` shim using the same basic shape as your latest comment (`whileMutablyBatching` + dirty-path recompute after the loop, wrapped in `withoutNormalizing`).

Median ms, same machine, 3 repeats:

| lane | `#6039` | `#6050` shim |
|---|---:|---:|
| pure exact `set_node` flat | `13.45` | `10.21` |
| pure exact `set_node` grouped | `9.54` | `7.96` |
| pure `move_node` flat | `8.46` | `541.56` |
| prepend `insert_node` into empty doc | `8.29` | `822.76` |
| interleaved `insert_node + move_node` | `8.68` | `551.44` |
| mixed `set_node + move_node` | `12.04` | `1078.94` |
| mixed `set_node + move_node + split_node` | `858.57` | `3342.79` |

So I think the right narrower claim for this PR is:

- `#6050` is the better answer for exact repeated `set_node`
- `#6039` is only worth defending for structural / heterogeneous batching
- if that broader batching story is not something Slate wants in core, then `#6039` probably shouldn't land in anything like its current size

I also reran the public non-batch harness on current head (`packages/slate/test/perf/public-nonbatch-bench.js`) and the earlier 30x non-batch regression is no longer there.

Public non-batch repro:
```bash
yarn node -r ./config/babel/register.cjs \
  ./packages/slate/test/perf/public-nonbatch-bench.js \
  --blocks=5000 \
  --group-size=50 \
  --insert-count=2000 \
  --move-count=2000 \
  --text-count=2000 \
  --structural-count=1000 \
  --repeats=3

nabbydude:
https://github.com/zbeyens/slate/blob/31ca576e4711c8e7d85bb712d08ecc42707cb549/packages/slate/src/interfaces/node.ts#L246-L253
I'm not a fan of how complex it is to properly get the children of a node. Are users expected to write code like this to avoid performance issues?

I have two big issues with this `getChildren` paradigm:
1. it's a big opaque wall around what used to be a transparent property of the editor, and I think it's a big sacrifice to the simplicity of slate.
2. if `Editor#children`, `Node.get`, and `Node.nodes` all call `getChildren`, it seems like even in a batch any transform that reads the tree (basically all of the existing ones) is going to cause flushing right away and negate the whole point of batching.
    - It seems like `applyBatch` works around this but it requires pre-queuing all of your operations (which includes accounting for how paths will look after the nth operation when dealing with structural changes) which is another big change to how slate transforms would need to be written to be performant

zbeyens:
You're right. I'm closing this as a limitation of current Slate, following up #6050.

I'm also in the middle of [a Slate rewrite](https://github.com/udecode/slate/tree/v2), that would avoid this kind of complex scheme. That alone would not justify a rewrite, `slate` itself is in pretty good shape. My main interest is to optimize `slate-react` but also unlock more features (e.g. decorations/annotations, invisible nodes, virtualization) - right now it's a dead end without changing the Slate runtime underneath it.

zbeyens:
@nabbydude here is the draft proposal: https://beta.platejs.org/docs/slate/why-this-fork. Would love to hear your thoughts.

PR #6040, opened 2026-04-02
Harden property accessors against untrusted keys

12joan:
**Description**
It's sometimes possible to cause mischief in JavaScript by supplying unexpected keys to square bracket property accessors. These unexpected keys can come from unexpected places. For example, `path.length` might be a non-number if a user is able to supply a path of `{ length: 'Untrusted value' }` as part of JSON input, such as serialised  operations.

I've gone through and hardened every property accessor I could find where it's remotely possible that an untrusted key might be used with harmful consequences.

As part of this, I've banned `set_node` operations from shadowing any property on the Object prototype (e.g. `__proto__`, `toString`, `hasOwnProperty`), since overwriting these default properties may cause unexpected behaviour.

The list of banned properties is currently: `toString`, `toLocaleString`, `valueOf`, `hasOwnProperty`, `isPrototypeOf`, `propertyIsEnumerable`, `__defineGetter__`, `__defineSetter__`, `__lookupGetter__`, `__lookupSetter__`, `__proto__`, `constructor`. This list is subject to change automatically as new properties are added to `Object.prototype`, but I think the risk of a collision between a future prototype property and a genuine user-supplied property name is very small.

**Checks**

### DIFF
--- a/.changeset/pink-jobs-kneel.md
+++ b/.changeset/pink-jobs-kneel.md
@@ -1,0 +1,7 @@
+---
+'slate-dom': patch
+'slate': patch
+---
+
+- Harden property accessors against untrusted keys
+- Fix incorrect argument types for the `compare` and `merge` options of `Transforms.setNodes`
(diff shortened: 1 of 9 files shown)


zbeyens:
@codex review (just trying)

dylans (approved):

PR #6065, opened 2026-05-29
Introduce `HyperscriptPointRef` and `HyperscriptRangeRef`

12joan:
**Description**
This PR adds two new concepts to `slate-hyperscript`: `HyperscriptPointRef` and `HyperscriptRangeRef`. These can be used to reduce the need for hardcoded points and ranges in tests.

**Example**
A `HyperscriptPointRef` can be used with the newly introduced `<point />` tag to construct any arbitrary point in the editor, similar to `<cursor />` but without affecting the selection.

```tsx
const pointRef = new HyperscriptPointRef()

<editor>
  <paragraph>
    Hello, <point ref={pointRef} />world!
  </paragraph>
</editor>

// Throws an error if the point hasn't been defined yet
const point = pointRef.point()

console.log(point) // => { path: [0, 0], offset: 7 }
```

A `HyperscriptRangeRef` can be used with the existing `<anchor />` and `<focus />` tags to construct any arbitrary range in the editor. `<anchor />` and `<focus />` tags do not affect `editor.selection` when used in this manner.

```tsx
const rangeRef = new HyperscriptRangeRef()

<editor>
  <paragraph>
    Hello, <anchor ref={rangeRef} />world!</focus ref={rangeRef} />
  </paragraph>
</editor>

// Throws an error if the range ref hasn't been used with both <anchor /> and <focus /> yet
const range = rangeRef.range()

console.log(range) // => {
//   anchor: { path: [0, 0], offset: 7 }
//   focus: { path: [0, 0], offset: 13 }
// }
```

**Context**
I wanted to add a `HyperscriptPathRef` too that could be used with the `ref` prop of any node's JSX tag, but this is less trivial since elements don't have their own token objects to add metadata to without polluting the node itself. Perhaps associating the node and the ref using a weak map as part of `createElement` would work best?

**Checks**

### DIFF
--- a/packages/slate-hyperscript/src/creators.ts
+++ b/packages/slate-hyperscript/src/creators.ts
@@ -2,11 +2,14 @@ import { Element, Descendant, Node, Range, Text, Editor } from 'slate'
 import {
   AnchorToken,
   FocusToken,
+  PointToken,
   Token,
   addAnchorToken,
   addFocusToken,
-  getAnchorOffset,
-  getFocusOffset,
+  addPointToken,
+  getAnchors,
+  getFoci,
+  getPoints,
 } from './tokens'
 
 /**
@@ -16,10 +19,10 @@ import {
 
 const STRINGS: WeakSet<Text> = new WeakSet()
 
-const resolveDescendants = (children: any[]): Descendant[] => {
+function resolveDescendants(children: any[]): Descendant[] {
   const nodes: Node[] = []
 
-  const addChild = (child: Node | Token): void => {
+  function addChild(child: Node | Token | string) {
     if (child == null) {
       return
     }
@@ -59,6 +62,8 @@ const resolveDescendants = (children: any[]): Descendant[] => {
         addAnchorToken(n, child)
       } else if (child instanceof FocusToken) {
         addFocusToken(n, child)
+      } else if (child instanceof PointToken) {
+        addPointToken(n, child)
       }
     } else {
       throw new Error(`Unexpected hyperscript child object: ${child}`)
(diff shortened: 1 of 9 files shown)


packages/slate-hyperscript/src/creators.ts
```diff
@@ -16,10 +19,10 @@ import {
 
 const STRINGS: WeakSet<Text> = new WeakSet()
 
-const resolveDescendants = (children: any[]): Descendant[] => {
+function resolveDescendants(children: any[]): Descendant[] {
   const nodes: Node[] = []
 
-  const addChild = (child: Node | Token): void => {
+  function addChild(child: Node | Token | string) {
```
12joan, line 25: On an unchanged line in this file, we're calling `addChild('')`, which for some reason TypeScript didn't complain about even though `string` wasn't explicitly listed in the types the `child` argument.

I think this was because the `Token` was previously equivalent to `{}` which TypeScript implicitly treats as `any`. As part of this PR, I've refactored `Token` to be non-empty, so `string` must now be explicitly listed as an argument type.

packages/slate-hyperscript/src/creators.ts
```diff
@@ -236,17 +253,23 @@ export const createEditor =
     // Search the document's texts to see if any of them have tokens associated
     // that need incorporated into the selection.
     for (const [node, path] of Node.texts(editor)) {
-      const anchor = getAnchorOffset(node)
-      const focus = getFocusOffset(node)
+      for (const { ref = selection, offset } of getAnchors(node)) {
+        if (offset != null) {
+          ref.anchor = { path, offset }
+        }
+      }
```
12joan, line 260: Previously, `getAnchorOffset` returned a single `[offset: number, AnchorToken]` tuple. If there were multiple `<anchor />`s used, the last one would be returned.

Since multiple `<anchor />`s are now a valid use case (when one or more is used with a `ref` prop), I've renamed the function to `getAnchors`. It now returns an `AnchorToken[]`, where the `offset` is stored on the object itself. As before, the last ref-less `<anchor />` tag always wins.

The same applies to `<focus />` and `<point />` tags.

dylans (approved):

PR #6073, opened 2026-06-27
fix(slate-react): prevent useSelected from throwing when element is removed

### LINKED ISSUE #6053: useSelected error when remove myself
I get an error in `useSelected()` when I try to delete myself.

```ts
const Options = React.forwardRef<any, { element: Element }>((props, ref) => {
  const selected = useSelected()
  React.useEffect(() => {
    if (!selected) {
      try {
        const path = ReactEditor.findPath(editor, props.element)
        deleteOptionsWhenEmpty(editor, path)
      } catch (error) {
        Sentry.captureException(error)
      }
    }
  }, [selected])
  return <OptionsStyled selected={selected} {...props} ref={ref} />
})
```
I got error `Error: Cannot find a descendant at path [0,1] in node`

My local solution is copy pase useSlected and wrap ` const range = Editor.range(editor, path)` in try catch

I proposle add it to lib

```ts
export const useSelectedMy = (): boolean => {
  const element = useElementIf()

  // Breaking the rules of hooks is fine here since `!element` will remain true
  // or false for the entire lifetime of the component this hook is called from.
  // TODO: Decide if we want to throw an error instead when calling
  // `useSelected` outside of an element (potentially a breaking change).
  if (!element) return false

  // eslint-disable-next-line react-hooks/rules-of-hooks
  const selector = React.useCallback(
    (editor: Editor) => {
      if (!editor.selection) return false
      try {
        const path = ReactEditor.findPath(editor, element)
        const range = Editor.range(editor, path)
        return !!Range.intersection(range, editor.selection)
      } catch (errro) {
        console.log(errro)
      }
    },

    [element],
  )

  return useSlateSelector(selector, undefined, { deferred: true })
}

```

NikunjSonigara:
## Fixes
Fixes #6053

## Summary
This PR prevents `useSelected` from throwing when the referenced element has been removed from the editor while still being referenced by a mounted React component.

Previously, when an element was deleted, `ReactEditor.findPath` or `Editor.range` could throw:

> Cannot find a descendant at path [...]

This would cause runtime crashes in components still holding a reference to the removed element.

## Solution
The path resolution and range calculation are now wrapped in a `try/catch`. If the element can no longer be resolved (because it has been removed from the editor tree), `useSelected` safely returns `false` instead of throwing.

## Why this works
When the element is removed, it is no longer part of the editor's document tree, so selection state should be treated as `false` rather than attempting to compute a range on a non-existent node.

## Test coverage
A regression test was added to ensure:
- A component still referencing a removed element does not crash
- `useSelected()` returns `false` after the element is removed
- Behavior is consistent with and without chunking enabled

## Notes
This fixes a real-world crash scenario where stale element references exist during React render cycles after DOM updates in Slate.

### DIFF
--- a/.changeset/fix-use-selected-removed-element.md
+++ b/.changeset/fix-use-selected-removed-element.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+fix: prevent `useSelected` from throwing when its element has been removed from the editor
--- a/.changeset/fix-use-selected-removed-element.md
+++ b/.changeset/fix-use-selected-removed-element.md
@@ -1,5 +1,5 @@
 ---
-'slate-react': patch
+'slate-react': minor
 ---
 
-fix: prevent `useSelected` from throwing when its element has been removed from the editor
+Add a `suppressThrow` option to `useSelected` to return `false` instead of throwing when the element can no longer be found in the editor.
--- a/packages/slate-react/src/hooks/use-selected.ts
+++ b/packages/slate-react/src/hooks/use-selected.ts
@@ -21,9 +21,17 @@ export const useSelected = (): boolean => {
   const selector = useCallback(
     (editor: Editor) => {
       if (!editor.selection) return false
-      const path = ReactEditor.findPath(editor, element)
-      const range = Editor.range(editor, path)
-      return !!Range.intersection(range, editor.selection)
+
+      try {
+        const path = ReactEditor.findPath(editor, element)
+        const range = Editor.range(editor, path)
+        return !!Range.intersection(range, editor.selection)
+      } catch (e) {
+        // The element may have been removed from the editor while a component
+        // referencing it is still mounted, in which case its path can no
+        // longer be resolved. Treat it as not selected rather than throwing.
+        return false
+      }
     },
     [element]
   )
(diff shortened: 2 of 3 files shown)


dylans (changes requested):
The general pattern we have been using to stop throwing is to make this opt-in. Some people prefer that things throw rather than failing silently. There are a few recent PRs that have made error throwing configurable, so I would prefer that approach. For example #6072.

Also there are a lot of comments in here that I would consider to be over-explaining by AI, please clean those up.

NikunjSonigara:
Thanks for the review @dylans, I have updated the PR:

* **Made it opt-in instead of silently swallowing**. `useSelected` now accepts a `suppressThrow` option and **throws by default** (unchanged behavior). Callers opt in with `useSelected({ suppressThrow: true })` to get `false` when the element can no longer be resolved. This follows the `suppressThrow` pattern from #6072 
* **Cleaned up the comments** - removed the over-explaining ones in both the hook and the test

Since the default still throws, users hitting #6053 opt in via `useSelected({ suppressThrow: true })`. Happy to adjust the option name or default if you'd prefer something different.

dylans (approved):

dylans (approved):

Issue #3556, opened 2020-03-21
 Unable to find the path for Slate node:


johnpaul89:
Hey Guys?

I'm trying to serialize this HTML as defined [here in the docs](https://docs.slatejs.org/concepts/09-serializing) but I keep getting this error

This error occurs when I have this component defined

``` 
             <Editable
                    renderElement={renderElement}
                    renderLeaf={renderLeaf}
                    placeholder="Enter some rich text…"
                    spellCheck
                    autoFocus
                    onKeyDown={event => {
                        for (const hotkey in HOTKEYS) {
                            if (isHotkey(hotkey, event)) {
                                event.preventDefault()
                                const mark = HOTKEYS[hotkey]
                                toggleMark(editor, mark)
                            }
                        }
                    }}
                />
```

Uncaught Error: Unable to find the path for Slate node: {"children":{"children":[{"type":"paragraph","children":[{"text":"This is editable "},{"text":"rich","bold":true},{"text":" text, "},{"text":"much","italic":true},{"text":" better than a "},{"text":"<textarea>","code":true},{"text":"!"}]},{"type":"paragraph","children":[{"text":"Since it's rich text, you can do things like turn a selection of text "},{"text":"bold","bold":true},{"text":", or add a semantically rendered block quote in the middle of the page, like this:"}]},{"type":"block-quote","children":[{"text":"A wise quote."}]},{"type":"paragraph","children":[{"text":"Try it out for yourself!"}]}]}}

Anyone knowing how to fix this?

johnpaul89:
I figured it out guys, on your initial value you should use 

```
    [
        {
            children: [{ text: '' }],
        }
    ]
```

Instead of the one provided in the [Docs](https://docs.slatejs.org/concepts/09-serializing) because that one in the docs is the one causing the error above.

fosterer:
Does not work for me...

cormacncheese:
@johnpaul89 is a temporary fix but I'm still unable to add an image from url. Same error as before

isachivka:
Opened a PR that fixes the `beforeinput` slice of this: https://github.com/ianstormtaylor/slate/pull/6080

[...]

PR #6080, opened 2026-07-02
Suppress findPath throw when translating beforeinput target ranges in Editable (#3556)

isachivka:
### Description

`Editable.onDOMBeforeInput` has two `ReactEditor.toSlateRange` call sites that hard-code `suppressThrow: false` — the `event.getTargetRanges()` branch and the WebKit ShadowRoot branch. When the browser's target range points at a DOM node that is no longer resolvable to a Slate node (stale NODE_MAP entry after a re-render, IME composition, clicking between editors), `findPath` throws `Unable to find the path for Slate node: …` **out of the DOM event handler**, reaching `window.onerror` and crashing the host app. This is the long-standing #3556 family.

Every other `toSlateRange` call site in this file already passes `suppressThrow: true` and null-checks the result — this PR aligns the last two:

- **`getTargetRanges` branch:** `suppressThrow: true`; when the range is unmappable, fall back to slate's synthetic handling at the current selection (`native = false`) — native handling can't be trusted with a range we couldn't map. Resolvable ranges behave exactly as before.
- **WebKit ShadowRoot branch:** `suppressThrow: true`; on `null` skip only the `Transforms.select`, still `preventDefault` (letting WebKit native-edit a shadow-root range we couldn't map would diverge the DOM from the model).

### Context / evidence

In production (slate-react 0.124.2) this is a steady session-killing crash — ~8 users/day, all with the identical stack:

```
Error: Unable to find the path for Slate node: {...}
    at Object.findPath
    at Object.toSlatePoint
    at Object.toSlateRange
    at HTMLDivElement.<anonymous>   ← Editable.onDOMBeforeInput, getTargetRanges branch
```

`toSlatePoint`/`toSlateRange` already honor `suppressThrow` around their `findPath` calls (returning `null`), so the change is confined to the two call sites and their degrade paths.

### Checks

- `yarn test:jest` after `yarn build:rollup`: 94/94 pass.
- `yarn lint:typescript`: no new errors (the conditional return type `T extends true ? Range | null : Range` makes the null-narrowing type-safe).
- Changeset added (`slate-react` patch).

### Issue

Relates to https://github.com/ianstormtaylor/slate/issues/3556


packages/slate-react/src/components/editable.tsx
```diff
@@ -563,13 +563,21 @@ export const Editable = forwardRef(
           newRange.setStart(range.startContainer, range.startOffset)
           newRange.setEnd(range.endContainer, range.endOffset)
 
-          // Translate the DOM Range into a Slate Range
+          // Translate the DOM Range into a Slate Range. The target range can
```
dylans: Walls of comments don't belong here please.

It would be better if it reads the value for suppressThrow as configured rather than just flipping the switch (meaning I agree the hard setting of it isn't great, but switching to the other hard setting isn't ideal either). Especially during dev/testing, it's difficult to track issues down if you suppress all errors.

packages/slate-react/src/components/editable.tsx
```diff
@@ -689,12 +697,22 @@ export const Editable = forwardRef(
             const [targetRange] = (event as any).getTargetRanges()
 
             if (targetRange) {
+              // The browser's target range can point at a DOM text node whose
```
dylans: This wall of comments either belongs in the docs or cut completely.

dylans (changes requested):

isachivka:
Thanks for the review — pushed a commit that cuts the multi-line comment blocks at both sites down to a single line.

On `suppressThrow`: there isn't an editor-level configuration for it in the codebase today — every `toSlateRange`/`toSlatePoint` call passes it explicitly, and all seven DOM-event call sites (`editable.tsx` ×5, `android-input-manager.ts` ×2) hard-code `suppressThrow: true`, so this change just brings these two `onDOMBeforeInput` sites in line with the rest of the file rather than introducing a new setting.

Worth noting `suppressThrow` is narrow: it only suppresses the specific "Unable to find the path for Slate node" resolution throw inside `toSlatePoint`/`toSlateRange` (returning `null`), not arbitrary errors — so it doesn't blanket-hide errors during dev/testing, and the handlers degrade explicitly on the `null` (fall back to synthetic handling / skip the selection move).

If you'd prefer an actual configurable opt-out (e.g. an editor-level flag read at these call sites), I'm happy to add that as a separate change across all the sites — I just didn't want to introduce new public API inside this bugfix. WDYT?

dylans:
> If you'd prefer an actual configurable opt-out (e.g. an editor-level flag read at these call sites), I'm happy to add that as a separate change across all the sites — I just didn't want to introduce new public API inside this bugfix. WDYT?

It would be useful, as a separate PR is fine.

dylans (approved):

PR #6084, opened 2026-07-19
fix(slate-react): keep the DOM in sync when a native insertText is a no-op

### LINKED ISSUE #5152: If insertText does nothing, text is inserted into the DOM anyway, causing DOM/Slate desync (as of slate-react 0.75.0)
**Description**
It is valid for `insertText` to return without doing anything. The expected behavior is that the document is left unchanged. For example, you might want to disallow typing uppercase characters -- you can do this by implementing `insertText`, and if the text is uppercase, do nothing. With slate-react 0.74.2 and below, this works correctly. But with slate-react 0.75.0 and above, it is broken: the text is inserted into the DOM anyway. However, it is (correctly!) NOT inserted into Slate's document -- this means the DOM and the Slate document are then out of sync, leading to further bugs.

**Sandbox**

Example: https://codesandbox.io/s/inserttext-should-do-nothing-if-it-returns-wzc647?file=/index.js

**Steps**
To reproduce the behavior:
1. Go to the sandbox above
2. In the editor, type "X"
3. Note that the character "X" is incorrectly inserted into the DOM
4. Type "x"
5. Note that the previous "X" has disappeared, and an incorrect character has been replaced with "x". (This is presumably due to desynchronization between DOM and Slate document)

**Expectation**
If my `insertText` does nothing, nothing should happen.

**Environment**
- Slate Version: slate-react 0.75.0 and above
- Operating System: Windows
- Browser: Chrome, Firefox

a-y-ibrahim:
**Description**

`Editable` handles single-character insertion natively for performance: it skips `preventDefault()` on the `beforeinput` event and lets the browser insert the character directly into the DOM, deferring the corresponding `Editor.insertText` call until the following `input` event.

If a custom `insertText` override intentionally does nothing for that character (e.g. blocking uppercase letters), the deferred call never changes the Slate document. Since the document didn't change, React never re-renders the affected `<Text>`, so `<TextString>`'s layout effect (the only mechanism that reconciles this span's DOM text against the model) never runs. The character the browser already inserted is left permanently stuck in the DOM even though the model correctly stayed unchanged.

This PR makes `Editable` undo that stray browser-side insertion directly, once it confirms the deferred `Editor.insertText` left the model unchanged. It captures the exact DOM text node and offset the browser is about to insert into (before the insertion happens), and if the model comes back unchanged, deletes exactly that one character at that exact position. Because it targets a specific DOM node rather than a whole leaf's text, this also correctly handles text nodes split into multiple spans by marks or decorations, not just the plain single-leaf case.

**Issue**

Fixes: #5152

**Example**

```tsx
const withNoUppercase = (editor: Editor) => {
  const { insertText } = editor

  editor.insertText = text => {
    if (/[A-Z]/.test(text)) return // block uppercase entirely
    insertText(text)
  }

  return editor
}
```

Before this fix, typing an uppercase letter into an editor using this plugin left the character visibly stuck in the DOM (e.g. `abc` → `abcX`) even though `editor.children` correctly still read `abc`. After this fix, the DOM and the model stay in sync (`abc` stays `abc`).

Two example pages reproduce this and are included in this PR:
- `/examples/insert-text-noop` — the plain, single-leaf case from the original report.
- `/examples/insert-text-noop-decorated` — the same bug for text split into multiple leaves by a decoration, which the initial version of this fix didn't yet cover (see Context).

**Context**

The first version of this fix compared the whole leaf's expected text against its `[data-slate-string]` span's `textContent`, replacing it wholesale on divergence, and explicitly bailed out whenever a text node was split into more than one span (marks/decorations), reasoning that the native fast path couldn't reach that case. That reasoning was wrong: the native-insertion gate only checks `editor.marks`, never decorations, so a decorated text node can still take the native path and hit the same stray-character bug with no correction applied.

The current version fixes that by working at the exact DOM position instead of the whole leaf: `ReactEditor.toDOMPoint` already resolves a Slate point to the correct specific span/offset regardless of how many leaves a text node is split into, so capturing `[domNode, domOffset]` there and surgically deleting one character is leaf-agnostic by construction. It also removes the need to reconstruct each leaf's expected string (dropping the `getLeafDomText` export this fix originally added to `string.tsx`, which is no longer needed), and mutating the DOM text node in place with `deleteData` is less disruptive than replacing `textContent` (which destroys and recreates the node, always resetting the caret).

Verified against React's built-in test suites (`yarn test:jest`, `yarn test:mocha`) and the full Playwright example suite, plus new tests for: the plain case, the decorated/multi-leaf case, and several consecutive/interleaved native insertions in a row (to rule out the single-slot tracking ref losing an earlier correction to a later one — it doesn't, since each keystroke's `beforeinput`/`input` cycle completes before the next one starts).

One known, pre-existing limitation this PR does **not** address: Android routes all input through a separate `AndroidInputManager` path that bypasses this correction entirely (both the native-insertion gate in `onDOMBeforeInput` and the correction in `onInput` return early when `androidInputManagerRef.current` is set). I confirmed by rebuilding from the commit before this fix and re-testing that the same stray-character bug reproduces identically on Android with zero changes applied, so this is an existing gap in `AndroidInputManager`'s own reconciliation, independent of this change and out of scope here.

**Checks**
- [x] The linter passes with `yarn lint`.
- [x] The relevant examples still work. (Verified via Playwright against a running `yarn serve`, not just manually.)
- [x] Added a changeset (`patch`, since this fixes a bug with no public API change).


dylans:
@12joan let me know if you have opinions on this one.

12joan:
@dylans Sure! I'll take a look later today.

site/constants/examples.ts
```diff
@@ -11,6 +11,8 @@ export const EXAMPLE_NAMES_AND_PATHS = [
   ['Huge Document', 'huge-document'],
   ['Images', 'images'],
   ['Inlines', 'inlines'],
+  ['Insert Text Noop', 'insert-text-noop'],
+  ['Insert Text Noop Decorated', 'insert-text-noop-decorated'],
```
12joan, line 15: Perhaps we should add these to `HIDDEN_EXAMPLES`, since they might confusing to people seeing them in the sidebar without explanation?

site/examples/ts/insert-text-noop-decorated.tsx
```diff
@@ -0,0 +1,84 @@
...
+      }
+    >
+      {children}
+    </span>
+  )
+}
+
+const InsertTextNoopDecoratedExample = () => {
+  const editor = useMemo(() => withNoUppercase(withReact(createEditor())), [])
+  const decorateCallback = useCallback(decorate, [])
```
12joan, line 61: There's no need to use `useCallback` to wrap a constant function. It's only necessary for functions that are recreated as part of the render function.

site/examples/ts/insert-text-noop-decorated.tsx
12joan, line 1: Would it do any harm to combine these into a single example where decoration is always enabled? At the very least, it would be good to combine the two Playwright tests since they're so similar.

packages/slate-react/src/components/editable.tsx
```diff
@@ -200,6 +200,25 @@ export const Editable = forwardRef(
...
+    // the deferred `Editor.insertText` has been flushed - that Slate's
+    // document actually changed. If a custom `insertText` ignored the
+    // character (see https://github.com/ianstormtaylor/slate/issues/5152),
+    // no Slate operation is applied, so no re-render happens to correct the
+    // DOM via <TextString>'s layout effect; we correct it manually instead,
+    // by undoing the browser's own single-character insertion at the exact
+    // DOM position it happened. This is leaf-agnostic (works regardless of
+    // how many marks/decorations split the text node into separate spans),
+    // since `domNode`/`domOffset` (from `ReactEditor.toDOMPoint`) already
+    // resolve to the specific leaf span the insertion landed in.
```
12joan, line 214: Having some explanation of this is good since it's pretty unintuitive otherwise, but please could you make this comment a bit more concise so that it pertains only to the `lastNativeInsertion` ref itself?

Also, I'm not sure `COMPAT` is quite right. It looks like it's used in a few existing comments with no clear pattern, but ideally I would expect it to appear only on code addressing browser-specific quirks (compatibility issues).

packages/slate-react/src/components/editable.tsx
```diff
@@ -829,6 +848,30 @@ export const Editable = forwardRef(
                 // Only insertText operations use the native functionality, for now.
                 // Potentially expand to single character deletes, as well.
                 if (native) {
+                  if (selection) {
+                    try {
+                      const { path } = selection.anchor
+                      const [node] = Editor.node(editor, path)
+
+                      if (Text.isText(node)) {
```
12joan, line 856: How about `const [node, path] = Editor.leaf(editor, selection.anchor)`?

This removes the need to pull out the path and node separately, or to check that `node` is a text node. The path of a point should always refer to a text node, but if not (such as if the selection is invalid), the error thrown by `Editor.leaf` will be caught by the try...catch anyway.

packages/slate-react/src/components/editable.tsx
```diff
@@ -829,6 +848,30 @@ export const Editable = forwardRef(
...
+                        const [domNode, domOffset] = ReactEditor.toDOMPoint(
+                          editor,
+                          selection.anchor
+                        ) as [DOMText, number]
```
12joan, line 860: Adding an `if (domNode instanceof DOMText)` might be safer than a type assertion here.

packages/slate-react/src/components/editable.tsx
```diff
@@ -1126,6 +1169,59 @@ export const Editable = forwardRef(
...
+                    // turned out to be a no-op (e.g. a custom `insertText` ignored
+                    // the character), the browser has already mutated the DOM, but
+                    // since Slate's document didn't change, no re-render happens to
+                    // correct it via <TextString>'s layout effect. Undo the
+                    // browser's own single-character insertion directly, at the
+                    // exact DOM position it happened - this works regardless of
+                    // how many leaves (marks, decorations) the surrounding text
+                    // node is split into, since `domNode`/`domOffset` already
+                    // identify the specific leaf span the insertion landed in.
+                    // https://github.com/ianstormtaylor/slate/issues/5152
```
12joan, line 1182:
```suggestion
                    // If a native insertion's deferred `Editor.insertText` did
                    // nothing, undo the browser's native insertion to remove the
                    // character from the DOM.
```

packages/slate-react/src/components/editable.tsx
```diff
@@ -1126,6 +1169,59 @@ export const Editable = forwardRef(
...
+                    const nativeInsertion = lastNativeInsertion.current
+                    lastNativeInsertion.current = null
+
+                    if (nativeInsertion) {
+                      try {
+                        const [node] = Editor.node(editor, nativeInsertion.path)
```
12joan, line 1188: `Node.get` (or `Node.getIf` to avoid throwing an error, or `Node.leaf` to check if it's a text node) returns a node rather than a node entry, which would avoid the need to destructure it here.

packages/slate-react/src/components/editable.tsx
```diff
@@ -1126,6 +1169,59 @@ export const Editable = forwardRef(
...
+                              editor.selection.anchor
+                            )
+                            domSelection?.collapse(domNode, domOffset)
+                          }
+                        }
+                      } catch {
+                        // The path may no longer point to a valid node (e.g.
+                        // it was affected by some other operation) - nothing
+                        // to correct in that case.
+                      }
```
12joan, line 1222: We might not need a try...catch if `Node.getIf` is used instead of `Editor.node`.

12joan (changes requested):
Nice work! The fix itself seems pretty robust. I tried changing the example to insert a different character to that which was typed, or to insert the character elsewhere in the text, and everything I tried behaved correctly.

I've left a few suggestions, mainly to improve code quality.

packages/slate-react/src/components/editable.tsx
12joan: Or, if possible, it might be better to avoid the need for a try...catch entirely.

PR #6091, opened 2026-08-04
Implement custom editor command helpers

bye-shuvo:
Added custom helper functions for editor commands.

**Description**
A modular approach to use the custom editor functionalities for any of the blocks or marks of the editor.

**Example**
In the previous doc it only shows a custom method for each of the marks. In the added doc it shows a way of implementing modular marks.

### DIFF
--- a/docs/walkthroughs/05-executing-commands.md
+++ b/docs/walkthroughs/05-executing-commands.md
@@ -251,35 +251,37 @@ We can make the code more modular by separating the logic within a few helper fu
 // Below code is written for typescript users
 const CustomEditor = {
     isMarkActive(editor: Editor, mark: string) {
-      const marks = Editor.marks(editor) as Record<string, unknown> | null;
-      return marks ? (marks as Record<string, boolean>)[mark] === true : false;
+      const marks = Editor.marks(editor)
+      return marks ? marks?.[mark] === true : false
     },
 
-    isBlockActive(editor: Editor, block: string) {
+    isBlockActive(editor: Editor, block: Element["type"]) {
       const [match] = Editor.nodes(editor, {
         match: (n) => Element.isElement(n) && n.type === block,
-      });
+      })
 
-      return match;
+      return !!match
     },
-
+    
     toggleMark(editor: Editor, mark: string) {
-      if (this.isMarkActive(editor, mark)) {
-        editor.removeMark(mark);
+      if (CustomEditor.isMarkActive(editor, mark)) {
+        editor.removeMark(mark)
       } else {
-        editor.addMark(mark, true);
+        editor.addMark(mark, true)
       }
     },
 
-    toggleBlock(editor: Editor, block: Element['type']) {
+    toggleBlock(editor: Editor, block: Element["type"]) {
       const isActive = CustomEditor.isBlockActive(editor, block)
       Transforms.setNodes(
         editor,
-        { type: isActive ? null : block },
-        { match: n => Element.isElement(n) && Editor.isBlock(editor, n) },
+        { type: isActive ? undefined : block },
+        {
+          match: (n) => Element.isElement(n) && Editor.isBlock(editor, n),
+        },
       )
     },
-  };
+  }
 
 const initialValue = [
   {
@@ -318,13 +320,13 @@ const App = () => {
           switch (event.key) {
             case '`': {
               event.preventDefault()
-              CustomEditor.toggleBlock(editor, "code");
+              CustomEditor.toggleBlock(editor, "code")
               break
             }
 
             case 'b': {
               event.preventDefault()
-              CustomEditor.toggleMark(editor, "bold");
+              CustomEditor.toggleMark(editor, "bold")
               break
             }
           }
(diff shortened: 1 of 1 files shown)


docs/walkthroughs/05-executing-commands.md
```diff
@@ -244,4 +244,97 @@ const App = () => {
 
 That's the benefit of extracting the logic.
 
+We can make the code more modular by separating the logic within a few helper functions using the below code:
```
12joan, line 247: Since this code is very similar to the existing snippet above, it might be better to replace that example rather than adding a new example.

12joan: Also note that Slate's code doesn't use semicolons, so we should probably remove them here for consistency.

docs/walkthroughs/05-executing-commands.md
```diff
@@ -244,4 +244,97 @@ const App = () => {
...
+
+```jsx
+// Define our own custom set of helpers.
+// Below code is written for typescript users
+const CustomEditor = {
+    isMarkActive(editor: Editor, mark: string) {
+      const marks = Editor.marks(editor) as Record<string, unknown> | null;
+      return marks ? (marks as Record<string, boolean>)[mark] === true : false;
+    },
```
12joan, line 256: `Editor.marks` returns `Omit<Text, 'text'> | null` by default, so its type will already be correct if a custom `Text` type is used. Maybe it would be better to avoid any type assertions and instead specify the type of the `mark` argument as `Exclude<keyof Text, 'text'>`?

In any case, I think the `marks as Record<string, boolean>` type assertion is definitely unnecessary. Perhaps just `return Boolean(marks?.[mark])`?

docs/walkthroughs/05-executing-commands.md
```diff
@@ -244,4 +244,97 @@ const App = () => {
...
+
+    isBlockActive(editor: Editor, block: string) {
+      const [match] = Editor.nodes(editor, {
+        match: (n) => Element.isElement(n) && n.type === block,
+      });
+
+      return match;
+    },
```
12joan: The type of `block` could optionally be tightened to `Element['type']`.

12joan: Also, we should be returning `!!match` rather than just `match`, as in the previous example.

docs/walkthroughs/05-executing-commands.md
```diff
@@ -244,4 +244,97 @@ const App = () => {
...
+
+    toggleMark(editor: Editor, mark: string) {
+      if (this.isMarkActive(editor, mark)) {
+        editor.removeMark(mark);
+      } else {
+        editor.addMark(mark, true);
+      }
+    },
```
12joan, line 272: I recommend using `CustomEditor` explicitly rather than `this`, mainly because `this` can become incorrect if the `toggleMark` function is passed as a variable before being called.

Also, I think `Editor.removeMark(editor, mark)` is more idiomatic than `editor.removeMark(mark)`, although it doesn't make much of a difference.

docs/walkthroughs/05-executing-commands.md
```diff
@@ -244,4 +244,97 @@ const App = () => {
...
+    toggleBlock(editor: Editor, block: string) {
+      const isActive = this.isBlockActive(editor, block);
+      Transforms.setNodes(
+        editor,
+        { type: isActive ? undefined : block },
+        {
+          match: (n) => Element.isElement(n) && Editor.isBlock(editor, n),
+        },
+      );
+    },
```
12joan:
```suggestion
    toggleBlock(editor: Editor, block: Element['type']) {
      const isActive = CustomEditor.isBlockActive(editor, block)
      Transforms.setNodes(
        editor,
        { type: isActive ? null : block },
        { match: n => Element.isElement(n) && Editor.isBlock(editor, n) },
      )
    },
```

docs/walkthroughs/05-executing-commands.md
bye-shuvo:
```jsx
    isMarkActive(editor: Editor, mark: string) {
      const marks = Editor.marks(editor);
      return Boolean(marks ?? marks?.[mark])
    },
```

docs/walkthroughs/05-executing-commands.md
bye-shuvo:
```jsx
  isBlockActive(editor: Editor, block: Element['type']) {
      const [match] = Editor.nodes(editor, {
        match: (n) => Element.isElement(n) && n.type === block,
      });

      return !!match;
    },

```

This should work fine

docs/walkthroughs/05-executing-commands.md
dylans: While I use semicolons in all of my code, we don't in Slate, so you should update to the current Slate standard and then we can merge this. Thanks.

bye-shuvo (commented):
Maintained zero ; in the code supporting slate standard

bye-shuvo:
Thanks for the clarifications.

PR #6092, opened 2026-08-05
Deeply compare nested array data structures

12joan:
**Description**
This PR extends the logic for `isDeepEqual` to deeply compare array structures. Previously, although objects were compared to unlimited depth, any arrays were only compared shallowly using `===` on their items.

**Example**
Before, these two text nodes would not have been merged:

```js
{ text: 'a', data: [{}] },
{ text: 'b', data: [{}] },
```

Now, they're treated as identical by `Text.equals(a, b, { loose: true })`, and therefore will be merged by Slate's built-in normalisation.

**Context**
Here's the history of how the text node comparison logic has changed up until now:

1. Initially, top-level node properties were shallowly compared using `===`
2. #3532 - @timbuckley added a deep equality check using Lodash
3. #4245 - @JonasKruckenberg replaced the Lodash dependency with `fast-deep-equal` to reduce the bundle size
4. #4276 - @TheSpyder replaced `fast-deep-equal` with a specialised custom equality check to fix a [regression](https://github.com/ianstormtaylor/slate/pull/3532#issuecomment-844461006) and avoid a third-party dependency
   - This is when support for deeply comparing array contents was dropped.
   - According to the comment added at this time, this was intentional, but I'm unable to find a documented reason for this in any code comment, GitHub conversation or unit test.

**Performance**
I haven't tested the performance of the new code, but I believe in simple cases where there are no nested data structures, this will only add one extra `isArray` check and two extra `isObject` checks to per `Text.equals` call. This is because `isDeepEqual` no longer assumes that its arguments are objects in order to simplify recursion.

**Checks**

### DIFF
--- a/.changeset/funny-shoes-allow.md
+++ b/.changeset/funny-shoes-allow.md
@@ -1,0 +1,5 @@
+---
+'slate': patch
+---
+
+Deeply compare nested data structures inside array properties on nodes for the purpose of merging identical text nodes. Previously, items in array properties were shallowly compared using `===`.
--- a/packages/slate/src/utils/deep-equal.ts
+++ b/packages/slate/src/utils/deep-equal.ts
@@ -1,45 +1,39 @@
 import { isObject } from './is-object'
 
-/*
-  Custom deep equal comparison for Slate nodes.
-
-  We don't need general purpose deep equality;
-  Slate only supports plain values, Arrays, and nested objects.
-  Complex values nested inside Arrays are not supported.
-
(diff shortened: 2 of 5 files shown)


timbuckley (approved):

timbuckley:
To other maintainers:

I think this is a straightforward improvement with little possible downside that will collapse more unnecessary duplicates of nodes, reducing confusion. I will let it sit for a bit to give time to others to approve.

cc: @ianstormtaylor

dylans (approved):

PR #5901, opened 2025-06-13
Fix Android cursor jumping to word start after autocorrect

bibixx:
**Description**
Fixes some of the cursor jumping on Android. There seems to be some kind of a race condition between the native `onDOMSelectionChange` and the actions that get run with Slate.

I tried to debug it more properly and find the underlying issue (eg. what causes the selection to be set with offset: 1) but I couldn't really find anything.

I have tested this with all of the android testcases and it seems to be working correctly.

**Issue**
Fixes #5893, fixes #5892

**Example**

| Before | After |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/31687b3c-a424-4cab-b20c-d8cf2324eb31"> | <video src="https://github.com/user-attachments/assets/f5d9d408-186e-40b4-8ce0-ebae9373e913"> |

| Before | After |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/d7f73982-494c-4ba5-b8cd-fee2a288cdae"> | <video src="https://github.com/user-attachments/assets/03ba246c-227b-45d8-b9db-16a1f3cd8304"> |

**Context**
The fix modifies the `handleDOMBeforeInput` method in `android-input-manager.ts` to properly maintain cursor position after storing text diffs. When `canStoreDiff` is true and we're storing a diff for text insertion, the code now:

1. Stores the current selection before applying the diff
2. Calculates the new cursor position based on the inserted text length  
3. Uses `scheduleAction` to ensure the selection update happens after the diff is applied
4. Updates the selection to the correct position at the end of the inserted text

This approach ensures that after autocorrect applies changes through the text diff system, the cursor remains in the expected position rather than jumping to unexpected locations.

**Checks**
- [x] The new code matches the existing patterns and styles.
- [x] The tests pass with `yarn test`.
- [x] The linter passes with `yarn lint`. (Fix errors with `yarn fix`.)
- [x] The relevant examples still work. (Run examples with `yarn start`.)
- [x] You've [added a changeset](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md) if changing functionality. (Add one with `yarn changeset add`.)

PR #6096, opened 2026-08-09
Fix IME composition being cancelled on the first character in an empty leaf (Android)

### LINKED ISSUE #5883: Composition interrupted in empty text nodes on Android IME

melodysdreamj:
Fixes #5883.

On Android, composing the first character into an empty leaf cancels the composition, leaving that character behind on its own. In Korean, typing `안녕` produces `ㅇ안녕` or `ㅇㅏㄴ녕`; in Japanese, `ha` produces `hあ` (the report in #5883); pinyin and Gboard English autocorrect break the same way. It affects the start of the editor and the start of any empty block, so on a chat-style app every message a Korean user types starts with a stray jamo.

## Why it happens

An empty leaf does not render a text node — it renders `ZeroWidthString`, which on Android is a `<span data-slate-zero-width><br/></span>`. When the IME composes the first character there, **the browser creates the text node**, and the composition lives inside a node React does not know about. Anything that replaces that node cancels the composition, and the IME then starts a fresh session, orphaning what it had composed so far.

Four separate paths do exactly that:

1. **The flush scheduled after every `insertCompositionText`.** #5901 added a `scheduleAction` that only re-selects, but `scheduleAction` sets `actionTimeoutId = setTimeout(flush)`, so a flush lands on the next task — about 10 ms after the first keystroke. Flushing applies the pending diff, the leaf stops being empty, and the re-render swaps `ZeroWidthString` for `TextString`, removing the composing node. This is what made the breakage deterministic rather than timing-dependent.
2. **The `FLUSH_DELAY` timer.** Same re-render, 200 ms after a keystroke, so it fires for anyone typing at a relaxed pace. This is the pre-#5901 version of the bug.
3. **`RestoreDOM`.** The browser-created text node is an untracked childList mutation, so any mid-composition re-render (the `compositionstart` state update, a placeholder resize) reverts it away. `restoreDOM` already skips `characterData` mutations *"because this interrupts the composition"* — childList mutations interrupt it just as much when the composition lives in a node that was just added.
4. **`TextString`'s layout effect.** It force-rewrites `textContent` whenever the DOM differs from the value. During composition the DOM is *supposed* to run ahead of the value — that is what the pending-diff design is for — so this replaces the composing text node on any render that happens mid-composition.

## The fix

1. `android-input-manager`: replace #5901's `scheduleAction` with `EDITOR_TO_PENDING_SELECTION`. The caret still lands where #5901 intended (applied on the next flush) without forcing a flush onto the next task.
2. `android-input-manager`: in `flush`, defer while `IS_COMPOSING` **and** the pending diffs target a leaf that is still empty in the value. That is the only case where applying a diff changes DOM *structure*; composing into a leaf that already has text only updates `textContent`, so it is untouched.
3. `android-input-manager`: apply those deferred diffs synchronously (`flushSync`) in `handleCompositionEnd`. IMEs that compose a syllable at a time start the next composition in the same tick, so an async render would replace a node the new composition has already started in.
4. `restore-dom-manager`: skip reverting childList mutations that contain the node the IME is composing in — the same rule already applied to `characterData` mutations.
5. `string.tsx`: skip the `textContent` rewrite while the IME composes inside that span (Android only).

Points 1–4 are inside Android-only code paths; point 5 is explicitly `IS_ANDROID`-gated. Nothing changes for other platforms.

## How this differs from #5921

#5921 addressed the same issue but @12joan found it fixed only the non-empty-editor case, still failed `composes correctly at the start of the editor`, and newly failed `updates the Slate value during composition` because the value stopped updating on every `compositionupdate`.

The deferral here is narrowed to *pending diffs whose target leaf is empty in the value*. In `updates the Slate value during composition` the caret sits after `Type here: `, so the leaf is not empty, no deferral happens, and the value still updates on every `compositionupdate`. I reproduced that test's shape (caret after existing text, value read mid-composition) and the value contains the composed text before composition ends.

I could not run [12joan/slate-android-tests](https://github.com/12joan/slate-android-tests) directly (it drives Appium/BrowserStack), so I reproduced each of the four cases by hand against the manual suite instead. Happy to re-verify if you can run the automated suite against this branch.

## Verification

**Emulator, real Gboard** — Pixel API 34 AVD, Android 14, **Gboard 12.4.05.482060964** (the same Gboard build @12joan reported), Chrome 113, typing on the actual on-screen keyboard at human pace into `/examples/android-tests`:

| Case | before | after |
|---|---|---|
| `Empty` — type 안녕 (Korean 2-bulsik) | `ㅇㅏㄴ녕`, Gboard's suggestion strip desynced to `ㅏㄴ녕` | `안녕`, suggestions `안녕 / 안뇽 / 안녕하세요` |
| `Autocorrect` — type `Cant`, space, then `i` | `Can't i` | `Can't i` (caret stays after the corrected word, #5901 still fixed) |

**Chromium composition pipeline via CDP** (`Input.imeSetComposition` / `Input.insertText` under an Android UA), against `/examples/plaintext` emptied first, driving Gboard's exact event stream:

| Case | before | after |
|---|---|---|
| Korean `안녕`, 350 ms/key | `ㅇ안녕` | `안녕` |
| Korean `안녕`, 120 ms/key | `ㅇ안녕` | `안녕` |
| Composing English `hello` | `hhello` | `hello` |
| First character in a block created by Enter | broken | correct |
| Value read mid-composition, caret after existing text | updates | updates (no regression) |
| Desktop UA control | `안녕` | `안녕` |

`yarn test:jest` passes (96/96), `lint:eslint` and `lint:prettier` are clean. `lint:typescript` and `test:mocha` fail identically on `main` in my environment (pre-existing, unrelated to this change), with no errors in the files touched here.

## Also included

A manual `IME first character` case in `android-tests`, as @12joan asked for on #5921, covering both the start of the editor and the start of an empty block, with the expected-vs-broken output spelled out.

## Downstream

Found while fixing this in Cinny, a Matrix client, where it makes every Korean message start with a stray jamo: cinnyapp/cinny#3066 carries these same changes as a `patch-package` patch until this lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/lucky-pandas-shave.md
+++ b/.changeset/lucky-pandas-shave.md
@@ -1,0 +1,5 @@
+---
+'slate-react': patch
+---
+
+Fix IME composition being cancelled on the first character typed into an empty leaf on Android, which left the first character behind as a separate, uncomposed character (typing 안녕 in Korean produced ㅇ안녕).
--- a/packages/slate-react/src/components/restore-dom/restore-dom-manager.ts
+++ b/packages/slate-react/src/components/restore-dom/restore-dom-manager.ts
@@ -1,6 +1,6 @@
 import { RefObject } from 'react'
 import { ReactEditor } from '../../plugin/react-editor'
-import { isTrackedMutation } from 'slate-dom'
+import { IS_COMPOSING, isTrackedMutation } from 'slate-dom'
 
 export type RestoreDOMManager = {
   registerMutations: (mutations: MutationRecord[]) => void
@@ -32,10 +32,37 @@ export const createRestoreDomManager = (
 
   function restoreDOM() {
     if (bufferedMutations.length > 0) {
+      // The node the IME is currently composing in, if any. Restoring a
+      // mutation that contains it removes the node out from under the
+      // composition and cancels it, in the same way characterData mutations do
+      // below. This happens when the first character is typed into an empty
+      // leaf, where the browser adds a text node next to the leaf's `<br>`.
+      let composingNode: Node | null = null
+
+      if (IS_COMPOSING.get(editor)) {
+        try {
+          composingNode =
+            ReactEditor.getWindow(editor).getSelection()?.anchorNode ?? null
+        } catch {
+          composingNode = null
+        }
+      }
+
+      const containsComposingNode = (node: Node) =>
+        !!composingNode &&
+        (node === composingNode || node.contains(composingNode))
+
       bufferedMutations.reverse().forEach(mutation => {
         if (mutation.type === 'characterData') {
           // We don't want to restore the DOM for characterData mutations
           // because this interrupts the composition.
+          return
+        }
+
+        if (
+          containsComposingNode(mutation.target) ||
+          Array.from(mutation.addedNodes).some(containsComposingNode)
+        ) {
           return
         }
 
--- a/packages/slate-react/src/components/restore-dom/restore-dom-manager.ts
+++ b/packages/slate-react/src/components/restore-dom/restore-dom-manager.ts
@@ -1,6 +1,6 @@
 import { RefObject } from 'react'
 import { ReactEditor } from '../../plugin/react-editor'
-import { IS_COMPOSING, isTrackedMutation } from 'slate-dom'
+import { isTrackedMutation } from 'slate-dom'
 
 export type RestoreDOMManager = {
   registerMutations: (mutations: MutationRecord[]) => void
@@ -32,37 +32,10 @@ export const createRestoreDomManager = (
 
   function restoreDOM() {
     if (bufferedMutations.length > 0) {
-      // The node the IME is currently composing in, if any. Restoring a
-      // mutation that contains it removes the node out from under the
-      // composition and cancels it, in the same way characterData mutations do
-      // below. This happens when the first character is typed into an empty
-      // leaf, where the browser adds a text node next to the leaf's `<br>`.
-      let composingNode: Node | null = null
-
-      if (IS_COMPOSING.get(editor)) {
-        try {
-          composingNode =
-            ReactEditor.getWindow(editor).getSelection()?.anchorNode ?? null
-        } catch {
-          composingNode = null
-        }
-      }
-
-      const containsComposingNode = (node: Node) =>
-        !!composingNode &&
-        (node === composingNode || node.contains(composingNode))
-
       bufferedMutations.reverse().forEach(mutation => {
         if (mutation.type === 'characterData') {
           // We don't want to restore the DOM for characterData mutations
           // because this interrupts the composition.
-          return
-        }
-
-        if (
-          containsComposingNode(mutation.target) ||
-          Array.from(mutation.addedNodes).some(containsComposingNode)
-        ) {
           return
         }
 
--- a/packages/slate-react/src/components/string.tsx
+++ b/packages/slate-react/src/components/string.tsx
@@ -3,7 +3,7 @@ import { Editor, Text, Path, Element, Node } from 'slate'
 
 import { ReactEditor, useSlateStatic } from '..'
 import { useIsomorphicLayoutEffect } from '../hooks/use-isomorphic-layout-effect'
-import { IS_ANDROID } from 'slate-dom'
+import { IS_ANDROID, IS_COMPOSING } from 'slate-dom'
 import { MARK_PLACEHOLDER_SYMBOL } from 'slate-dom'
 
 /**
@@ -61,6 +61,7 @@ const String = (props: {
  */
 const TextString = (props: { text: string; isTrailing?: boolean }) => {
   const { text, isTrailing = false } = props
+  const editor = useSlateStatic()
   const ref = useRef<HTMLSpanElement>(null)
   const getTextContent = () => {
     return `${text ?? ''}${isTrailing ? '\n' : ''}`
@@ -81,6 +82,20 @@ const TextString = (props: { text: string; isTrailing?: boolean }) => {
     const textWithTrailing = getTextContent()
 
     if (ref.current && ref.current.textContent !== textWithTrailing) {
+      // On Android the DOM intentionally runs ahead of the value while the IME
+      // composes, since input is buffered as pending diffs. Writing
+      // `textContent` here would replace the text node the composition lives in
+      // and cancel it; the pending diffs reconcile this span when the
+      // composition ends.
+      if (IS_ANDROID && IS_COMPOSING.get(editor)) {
+        const composingNode =
+          ReactEditor.getWindow(editor).getSelection()?.anchorNode
+
+        if (composingNode && ref.current.contains(composingNode)) {
+          return
+        }
+      }
+
       ref.current.textContent = textWithTrailing
     }
 
--- a/packages/slate-react/src/components/string.tsx
+++ b/packages/slate-react/src/components/string.tsx
@@ -150,9 +150,15 @@ export const ZeroWidthString = (props: {
   // be because accepting an IME suggestion when at the start of a block (no
   // preceding \uFEFF) removes one or more DOM elements that `toSlateRange`
   // depends on. (https://github.com/ianstormtaylor/slate/issues/5703)
+  // A leaf with no text node at all forces the IME to create one when the
+  // first character is composed there, and no re-render can then touch the
+  // leaf without cancelling the composition. Rendering the zero-width space on
+  // Android too gives the composition a React-owned text node to live in; the
+  // wrapper keeps its data-slate-zero-width attributes, so selection mapping
+  // is unchanged.
   return (
     <span {...attributes}>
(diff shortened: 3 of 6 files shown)


melodysdreamj:
Note for anyone reading later: the Codex review thread on this PR is now collapsed as outdated, because the fix for it rewrote the exact line the comment was anchored to. Summarising it here so it is not lost.

**Codex flagged (P1):** `import { flushSync } from 'react-dom'` breaks the UMD production build, since rollup cannot statically discover named exports from react-dom's CommonJS entry and `config/rollup/rollup.config.js` only registers `findDOMNode` under `namedExports`.

**It was right.** `NODE_ENV=production yarn build:rollup` failed with `"flushSync" is not exported by react-dom/index.js`. My earlier build ran without `NODE_ENV=production`, which skips the UMD targets entirely, so I had missed it.

**Fixed in cad63b60** by using the default import (`ReactDOM.flushSync`) instead of adding another name to `namedExports` — this matches how `with-react.ts` already reaches `unstable_batchedUpdates` and leaves the shared rollup config untouched.

After the fix: `NODE_ENV=production yarn build:rollup` completes, `test:jest` passes 96/96, eslint and prettier are clean, and the Korean composition check on the Gboard emulator still produces 안녕.

melodysdreamj:
Confirmed on a physical Android device, in a real chat app rather than a test page.

A user hitting this bug ran an instrumented build on their own phone (Korean 2-bulsik keyboard) and typed 안녕 into an empty composer. Trace, times in ms:

```
54881 compositionstart
54884 compositionupdate  ㅇ
54891 DOM+  #text"ㅇ"                                   <- browser creates the text node
54893 DOM-  <br>                                        <- inside the empty leaf
54895 text  "ㅇ"
55048 compositionupdate  아
55263 compositionupdate  안
55432 compositionupdate  안ㄴ
55588 compositionupdate  안녀
55659 compositionupdate  안녕
56021 compositionend     안녕
56034 DOM-  <span data-slate-zero-width>"안녕"           <- leaf swap happens once,
56036 DOM+  <span data-slate-string>"안녕"                  at the composition boundary
```

Result: `안녕`.

The important part is what is *absent*. Between the first character at 54891 and `compositionend` at 56021 there is no structural DOM change — only `characterData` updates. The zero-width leaf survives the whole composition, so the IME keeps its session and all six jamo land in one syllable pair.

Without this change the same sequence swaps `data-slate-zero-width` for `data-slate-string` about 10 ms after the first character, which cancels the composition; the IME then opens a fresh session and the first jamo is orphaned as `ㅇ안녕`.

melodysdreamj:
Two follow-ups since the device confirmation.

**Closed a hole in the deferral.** The deferral had exactly one exit — `compositionend` — and Slate's own comment in `Editable` notes that event is not fired reliably in every browser. A composition that got stuck therefore left the typed text out of the value indefinitely while re-arming a 200 ms timer. The visible case is tapping a send button (no `keydown`, so the existing stuck-composition safeguard on keydown never runs) while the first character is still composing: the app reads the value and the character is missing. My earlier submit-race check pressed Enter, which *is* a keydown, so it did not cover this.

A composition now stops holding flushes back once it has lost focus, or once it has been idle for 5 s. The idle window is deliberately generous so that pausing mid-word in a long Japanese or pinyin composition is not cut short.

**Coverage beyond Korean.** The PR claims this fixes #5883, which was reported in Japanese, so verifying only Korean was not enough. On the same Pixel/Android 14/Gboard emulator, into an empty leaf:

| input | result |
|---|---|
| Japanese: `た` then the dakuten key | `だ` — one character |
| Chinese pinyin: `n`, `i` | `ni` as a single composing run, candidates `你 尼 泥 逆` |
| Korean: 안녕 (re-run after this change) | `안녕` |

The Japanese case is the decisive one: dakuten modifies the character already being composed, so a composition broken at the first character would leave `た゛` as two characters instead of `だ`. For pinyin, a break after `n` would have restarted composition at `i` and offered candidates for `i` alone.

`test:jest` still passes 96/96, eslint and prettier are clean, and both the dev and `NODE_ENV=production` rollup builds complete.

melodysdreamj:
Dropped one hunk: replacing #5901's `scheduleAction` was both unnecessary and harmful.

A tester on a physical device reported that after committing a syllable and pressing space, the caret sat a character behind for a moment before jumping forward. Sampling `editor.selection` every 45 ms after the space shows it plainly:

| | caret offset over 540 ms |
|---|---|
| with the `EDITOR_TO_PENDING_SELECTION` swap | `1 1 1 1 1 2 2 2 2 2 2 2` — wrong for ~225 ms |
| with #5901 left alone | `2 2 2 2 2 2 2 2 2 2 2 2` |

That is #5901's own bug coming back, which makes sense: it fixed the caret by scheduling the selection as an *action*, and I had turned it into a pending selection that only lands on the next flush.

The swap was also redundant. Its purpose was to stop `scheduleAction`'s `setTimeout(flush)` from landing mid-composition — but the flush gate added here already refuses that flush. Verified by re-running the whole matrix with the hunk removed: Korean (slow and fast), composing English, first character in a block created by Enter, the submit race, and the mid-composition value check all still pass, and 안녕 still composes correctly on the Gboard emulator.

So `android-input-manager` now only adds the flush gate, the synchronous flush at `compositionend`, and the staleness check; `scheduleAction` is untouched. `test:jest` 96/96, eslint and prettier clean, dev and production rollup builds both complete.

melodysdreamj:
Found and fixed a regression in this PR during device testing, and the fix simplified the whole approach (df2fc346).

**The regression:** backspacing away the only composing character (type `ㅇ` into an empty composer, press backspace) left the character in the DOM and could close the Android keyboard. The IME reports that as an *empty* `insertCompositionText` — a cancellation, not a delete — and the deferred text existed in neither the value nor anywhere slate could delete it from, so every path out of that state was wrong in some way.

**The underlying problem** was one level deeper than this PR originally went. On Android an empty leaf renders no text node at all (`<br>` only), so the first composition lives in a node the *browser* creates — a node no re-render can preserve, which is what made RestoreDOM/childList protection necessary in the first place.

Three changes replace it:

1. `ZeroWidthString` now renders the `﻿` text node on Android too, same as every other platform. Composition then lives in a React-owned node and the browser only ever edits `characterData` — which `restoreDOM` already skips. The wrapper keeps its `data-slate-zero-width` attributes, so `toSlatePoint`/`toDOMPoint` are unaffected. (I could not find a reason for the Android exclusion in history; if there is a known Gboard issue with it, very interested to hear.)
2. An empty `insertCompositionText` drops the deferred diffs instead of applying them. The value never contained the cancelled text, so it is already in the desired state; applying-then-re-deleting churned the DOM mid-composition, which is what closed the keyboard.
3. When a cancelled composition ends, one forced render lets `RestoreDOM` reconcile whatever the IME did to the leaf's DOM, and the placeholder is only re-shown once the composition is over (a `contenteditable=false` element appearing next to the caret mid-composition is a known keyboard-closer).

With composition in a stable node, the `restoreDOM` childList guard from earlier in this PR became unnecessary and is reverted — the diff is smaller than before.

Re-verified on the Gboard emulator (Android 14, Gboard 12.4.05): first character composes (`안녕`), backspacing the last composing character deletes it with the keyboard staying open (3/3 runs), typing again immediately after a cancel composes cleanly, and the CDP matrix (Korean slow/fast, composing English, submit race, new-block-after-Enter, mid-composition value updates with a non-empty leaf, caret-after-space, desktop control) all passes. `test:jest` 96/96, eslint/prettier clean, dev and production rollup builds complete.

dylans (approved):
Thanks @melodysdreamj !

PR #6102, opened 2026-08-31
Handle numbers and booleans in `slate-hyperscript`

12joan:
**Description**
Previously, trying to include a number or boolean as a child in `slate-hyperscript` resulted in an error like `Error: Unexpected hyperscript child object: <value>`.

This PR updates the child handling logic to make number and boolean handling consistent with React JSX. Numbers are converted to strings, and booleans are ignored.

**Example**
```jsx
<element>{4}</element> // Element containing { text: '4' }
<element>{true}{false}</element> // Empty element, since true and false are ignored
```

**Context**
Since JSX is not type-safe, it's quite common to accidentally pass a number instead of a string as a child of a JSX tag when creating JSX dynamically.

Ignoring booleans lets us safely use patterns like `{condition && <element />}`.

**Checks**

### DIFF
--- a/docs/walkthroughs/05-executing-commands.md
+++ b/docs/walkthroughs/05-executing-commands.md
@@ -262,7 +262,7 @@ const CustomEditor = {
 
       return !!match
     },
-    
+
     toggleMark(editor: Editor, mark: string) {
       if (CustomEditor.isMarkActive(editor, mark)) {
         editor.removeMark(mark)
--- a/site/examples/js/android-tests.jsx
+++ b/site/examples/js/android-tests.jsx
@@ -177,7 +177,7 @@ const TEST_CASES = [
     id: 'autocorrect',
     name: 'Autocorrect',
     instructions:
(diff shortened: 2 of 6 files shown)


docs/walkthroughs/05-executing-commands.md
```diff
@@ -262,7 +262,7 @@ const CustomEditor = {
 
       return !!match
     },
-    
+
```
12joan, line 265: From `yarn fix`

site/examples/js/android-tests.jsx
```diff
@@ -177,7 +177,7 @@ const TEST_CASES = [
     id: 'autocorrect',
     name: 'Autocorrect',
     instructions:
-      'Type "Cant", then press space to autocorrect it. Make sure the cursor position is correct (after the autocorrected word)',
+      'Type "Cant" (make sure to misspell it), then press space to autocorrect it. Make sure the cursor position is correct (after the autocorrected word)',
```
12joan, line 180: From `yarn fix`

dylans (approved):

